PR #27, opened 2016-07-23
DB/Characters: update DB structure

ShinDarth:
Imported: https://github.com/TrinityCore/TrinityCore/commit/f8230a59a9e878a6f54f421d6621f1595bd93861

in order to make the characters DB similar to the TrinityCore one

### DIFF
--- a/data/sql/base/characters.sql
+++ b/data/sql/base/characters.sql
@@ -533,22 +533,22 @@ DROP TABLE IF EXISTS `character_aura`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `character_aura` (
   `guid` int(10) unsigned NOT NULL DEFAULT '0' COMMENT 'Global Unique Identifier',
-  `caster_guid` bigint(20) unsigned NOT NULL DEFAULT '0' COMMENT 'Full Global Unique Identifier',
-  `item_guid` bigint(20) unsigned NOT NULL DEFAULT '0',
+  `casterGuid` bigint(20) unsigned NOT NULL DEFAULT '0' COMMENT 'Full Global Unique Identifier',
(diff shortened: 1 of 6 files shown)


Bodeguero:
@ShinDarth Update for TrinityCore Rev: 61?

ShinDarth:
@Bodeguero this is to make characters database structure similar to TrinityCore, it would be nice to have the same structure

Yehonal:
Could you update it with latest changes please?

ShinDarth:
@Yehonal done

data/sql/updates/characters/2016_MM_DD_XX.sql
```diff
@@ -0,0 +1,49 @@
...
+CHANGE `remaintime`       `remainTime`      int(11) NOT NULL DEFAULT '0',
+CHANGE `remaincharges`    `remainCharges`   tinyint(3) unsigned NOT NULL DEFAULT '0';
+
+-- `character_glyphs`
+ALTER TABLE `character_glyphs`
+CHANGE `spec` `talentGroup` tinyint(3) unsigned NOT NULL DEFAULT '0';
+
+-- `character_talent`
+ALTER TABLE `character_talent`
+CHANGE `specMask` `talentGroup` tinyint(3) unsigned NOT NULL DEFAULT '0';
```
Yehonal: specMask cannot be converted to talentGroup since it is a totally different field of TC.

AzerothCore is using a mask, developed by SunwellCore, where:

1 -> spec 1
2 -> spec 2
3 -> spec 1 and 2

TrinityCore instead use 0 for first spec, and 1 for second spec...if a talent is present on both spec, it is present twice

specMask is also used on character_spell

I think they cannot be never synced with TC

PR #41, opened 2016-07-27
Core/Fixes: Fix quest Convocation at Zol'Heb

Bodeguero:
* Update Quest

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,48 @@
 # AzerothCore
 
-[![Gitter](https://badges.gitter.im/azerothcore/azerothcore.svg)](https://gitter.im/azerothcore/azerothcore?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+## Status
+
+AzerothCore | Chat
+:------------: | :------------:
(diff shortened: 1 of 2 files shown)


ShinDarth:
https://github.com/azerothcore/azerothcore-wotlk/pull/41/commits

You should use a separate branch for each PR, otherwise we'll have unrelated commits

Bodeguero:
@ShinDarth I sent separate PR - one was the Brach - Update General Informations and other Quest just do not know why this shit here going along I always sent them separate PR and will have here is this shit!

ShinDarth:
@Bodeguero I think we need a guideline for submitting PRs, I will write one if needed

Bodeguero:
@ShinDarth It will not be necessary because I know very well send Pull Request the source - I do separate branch to send and do not know what the hell goes along only here - already sent to other sources and never gave this problem already here ta well - more okay as I mentioned above I will only remove things fixed this source and implement the new revision of TrinityCore! Thanks for sharing this great source!

Grateful!

ShinDarth:
If you successfully import anything from TrinityCore, it would be nice to have a PR to import it here as well

PR #87, opened 2016-08-14
Core/Spells: Fix Hellscream's Warsong & Strength of Wrynn buffs.

DevRival:
Changes proposed: Fix Hellscream's Warsong & Strength of Wrynn buffs

- Currently, ICC uses 10% buff, but we have a version 3.3.5a and it is 30%.

Target branch(es): 1.x

Tests performed: (Does it build, tested in-game, etc)
Tested in-game , works as intended.

Known issues and TODO list: No issue.

### DIFF
--- a/data/sql/updates/2016_08_14_01.sql
+++ b/data/sql/updates/2016_08_14_01.sql
@@ -1,0 +1,8 @@
+ALTER TABLE world_db_version CHANGE COLUMN 2016_08_14_00 2016_08_14_01 bit;
+
+-- Hellscream's Warsong
+UPDATE `spell_area` SET `spell`='73822'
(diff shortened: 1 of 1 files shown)


ShinDarth:
So basically you are replacing:

http://wotlk.openwow.com/spell=73818 with http://wotlk.openwow.com/spell=73822

and 

http://wotlk.openwow.com/spell=73824 with http://wotlk.openwow.com/spell=73828

why you think it should be 30% instead of 10%?

DevRival:
By Rokoma (893 – 1·7·23) on 2010/08/04 (Patch 3.3.5)		
Horray for the 30% ICC buff! Now, instead of actually working to get your kills, you get your tank to whack them upside the head and they fall to the floor with loot comming out the unmentionables.

PR #164, opened 2016-10-01
Adding character deleted command from trinitycore

Kalhac:
**Changes proposed:**
Add command `.character deleted`, which have the subcommands `list`,`delete`,`restore` and old.

**Target branch(es):** Master

**Issues addressed:** None

**Tests performed:** The subcommands `list`, `delete` and `restore` have been tested and they worked. The subcommand `old` needs to be tested.

**Known issues and TODO list:** Nothing


HeartWell:
Good work, i test this!

ShinDarth:
good job, but the SQL part is missing, please add it

@HeartWell  thanks for the feedback, did you also test the ```old``` subcommand ?

HeartWell:
```sql
DELETE FROM command WHERE name IN ('character delete', 'character deleted list', 'character deleted restore', 'character deleted delete', 'character deleted old', 'character erase');
INSERT INTO command (name, security, help) VALUES
('character erase',4,'Syntax: .character erase $name\r\n\r\nDelete character $name. Character finally deleted in case any deleting options.'),
('character deleted delete', 4, 'Syntax: .character deleted delete #guid|$name\r\n\r\nCompletely deletes the selected characters.\r\nIf $name is supplied, only characters with that string in their name will be deleted, if #guid is supplied, only the character with that GUID will be deleted.'),
('character deleted list', 3, 'Syntax: .character deleted list [#guid|$name]\r\n\r\nShows a list with all deleted characters.\r\nIf $name is supplied, only characters with that string in their name will be selected, if #guid is supplied, only the character with that GUID will be selected.'),
('character deleted old', 4, 'Syntax: .character deleted old [#keepDays]\r\n\r\nCompletely deletes all characters with deleted time longer #keepDays. If #keepDays not provided the  used value from mangosd.conf option ''CharDelete.KeepDays''. If referenced config option disabled (use 0 value) then command can''t be used without #keepDays.'),
('character deleted restore', 3, 'Syntax: .character deleted restore #guid|$name [$newname] [#new account]\r\n\r\nRestores deleted characters.\r\nIf $name is supplied, only characters with that string in their name will be restored, if $guid is supplied, only the character with that GUID will be restored.\r\nIf $newname is set, the character will be restored with that name instead of the original one. If #newaccount is set, the character will be restored to specific account character list. This works only with one character!');

File: src/scripts/Commands/cs_character.cpp [hunk 1]
```diff
@@ -506,7 +826,7 @@ class character_commandscript : public CommandScript
             do{
                 uint64 _guid = result->Fetch()[0].GetUInt64();
                 char buff[20];
-                sprintf(buff,"%I64u", _guid);
+                sprintf(buff,"%I64u", (uint)_guid);
```
HeartWell: Change (uint)_guid)->(uint32)_guid) and %I64u->%u, this right.

Kalhac:
I added the sql file and I made some changes as suggested.

HeartWell:
Good, anyone else test?

Krath:
Did you test it? It doesn't seem to work for me. Every subcommand I try, it shows me the syntax, even if I type it good. Can anyone else confirm?

Kalhac:
I tested all the commands and subcommands except the `character deleted old` command and they worked. I use linux by the way.

ShinDarth:
I confirm

sucofog:
I confirm also, tested.

falzard:
So no need to test it but it's not working?

ShinDarth:
I guess we need someone to actually finish the job here

Yehonal:
@Kalhac any news?

Kalhac:
I can't work on it at the moment. I'll try to make some tests in the weekend

Franklampardst:
Tested and worked all commands tnx

Ladrek1989:
everything works but .character delete old

talamortis:
Thank you for testing @Ladrek1989

Issue #299, opened 2016-12-28
Quest: The Ancient Armor of the Kvaldir


Irome:
** ** Description: http://www.wowhead.com/quest=11567/the-ancient-armor-of-the-kvaldir

** ** Current behavior: If you try to get on the harrys-bomber, when you try to click on the vehicle does not allow you to fly. Same problem happens if you try to go back. The quest can be completed without using the vehicle.

** ** Expected behavior: The goal is to recover the Ancient Armor of the Kvaldir, to reach the place you should get on the harrys-bomber (http://www.wowhead.com/npc=32682/harrys-bomber) it should bring in the right place.

This video does better understand what happens: http://tinypic.com/player.php?v=1tx89d&s=9#.WGQM1DVjZ0w

**AC hash/commit**:  7025bad000e31ed1bd058e4f94b19fe832f9d28a

starrheld:
GIVE ME FEEDBACK
------

Fix:

```SQL
UPDATE `creature_template` SET `AIName`='SmartAI' WHERE `entry` IN (32682, 27932);
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (32682, 27932) AND `source_type`=0;
UPDATE `creature_template` SET `InhabitType`=4, `speed_run`=2.5 WHERE `entry` IN (32682, 27932);
INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
(32682, 0, 0, 0, 27, 0, 100, 1, 0, 0, 0, 0, 53, 1, 32682, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0,0,'Harry''s Bomber- On passeger rided - Start wp'),
(27932, 0, 0, 0, 27, 0, 100, 1, 0, 0, 0, 0, 53, 1, 27932, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0,0,'Harry''s Bomber- On passeger rided - Start wp'),
[...]

Issue #327, opened 2017-01-14
Scalding Mornbrew Delivery


Flerex:
**Description**: Mission _Scalding Mornbrew Delivery_ from the initial Dwarf/Gnome zone gets completed even when the timer runs out of time.

**Current behaviour**: It gets auto-completed when the timer runs out of time

**Expected behaviour**: It should mark it as failed.

**Branch(es)**: master

**AC hash/commit**:  e30ae1d4c329b8a203ed0d75b6eccccef30cba66

**Operating system**:  MacOS

**Modules**:  Nope

**Other Customizations**: Nope

ShinDarth:
missing revision commit hash, please read:

https://github.com/azerothcore/azerothcore-wotlk/wiki/Contribute#how-to-open-an-issue

and update your report adding the hash

Flerex:
Fixed

starrheld:
**Can someone confirm?**
-----------------------------
From **WOWHEAD**

> By Dregothalom (135 – 1) on 2009/12/24 (Patch 3.3.0)		
No longer timed as of patch 3.3.

Source: http://www.wowhead.com/quest=3364/scalding-mornbrew-delivery

If so
- Fix:

```SQL
UPDATE `quest_template` SET `LimitTime` = 0 WHERE `id` IN(3364);
```

PR #385, opened 2017-02-19
Minor SQL fixes

starrheld:
Closes: #299 and #327


talamortis:
tested and both working :)

PR #389, opened 2017-02-19
[NPC][VANILLA] Rotting Agam'ar

starrheld:
Rotting Agam'ar should not be in the air, should be in Razorfen Kraul.


talamortis:
Tested and working

PR #393, opened 2017-02-21
Minor SQL Fix

starrheld:
[NPC] Kruban Darkblade <Darkmoon Faire Barker> should only be present while Darkmoon Faire event is active.


talamortis:
Tested and working :) 👍

PR #423, opened 2017-03-11
Core/Command: Added disable commands.

HeartWell:
**Changes proposed:**

-  Added disable commands.

**Target branch(es):** 1.x

**Tests performed:** Build and tested

### DIFF
--- a/src/scripts/ScriptLoader.cpp
+++ b/src/scripts/ScriptLoader.cpp
@@ -601,7 +601,7 @@ void AddCommandScripts()
     AddSC_cast_commandscript();
     AddSC_character_commandscript();
     AddSC_debug_commandscript();
-	AddSC_disable_commandscript();
+    AddSC_disable_commandscript();
     AddSC_event_commandscript();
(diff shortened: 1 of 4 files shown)


src/scripts/ScriptLoader.cpp
```diff
@@ -600,6 +601,7 @@ void AddCommandScripts()
     AddSC_cast_commandscript();
     AddSC_character_commandscript();
     AddSC_debug_commandscript();
+	AddSC_disable_commandscript();
```
ShinDarth: indentation

HeartWell: I'll fix it later when I'm at the PC.

HeartWell: And fix the error travis, I forgot that you do not have debug_mmaps.

Viste:
Tested work fine, but need to fix .reload disables(not work now need restart server to apply changes in table even if use command .disable)

PR #425, opened 2017-03-14
Fix bad SQL

Rochet2:
Restructured and reuploaded an SQL that kept destroying travis and local sql update.

How the restructuring was done:
- Download original SQL
- Remove the transactions and version stuffs (only preserve table deletion, creation and data insertions)
- Split INSERT INTO statements into small enough ones that they can be executed to local DB
- After successful import of the SQL to local DB, make an export with heidisql
  - The export had 1024 KB for each insert into statement, so now there are almost 50 of them instead of 3 like before http://i.imgur.com/Fmx7iy8.png
- Add the transaction and version code back in again

However splitting the insert into statements is not enough. There is too much data inserted to be in a transaction (enabled by the stored procedure). **This is why, for now, I have removed the transaction/stored procedure.**
See the commit that removes the transaction: https://github.com/azerothcore/azerothcore-wotlk/pull/425/commits/77d08db26c76cdd11db8f9de619aa678c00ab9d6

**I am not entirely sure I did that last part right or what you would prefer instead of removing it.**


Rochet2:
@mayaren888 It would seem that at one point in the conversion the encoding has changed and screwed the document. Will take a look later today if I have time.

mayaren888:
@Rochet2  You are our God.

Rochet2:
Should be fixed in https://github.com/azerothcore/azerothcore-wotlk/commit/31f43a6cd166035d0c5b976cdcd5d534084285b2
@mayaren888

PR #491, opened 2017-05-22
auto-attack for warbear matriarch

PolarCookie:
**Changes proposed:**
Borrowed from TC issue#13905

The Warbear Matriarch vehicle for The Warm-Up (12996) and Into the Pit (12997) is not doing any auto-attacks, neither does an auto-attack button appear on the action bar.

This update fixes the first issue, and adjust the auto attack damage to be the same as the maul ability, as that's what it seems to be from this video: https://youtu.be/CJ28xVAeh6M?t=41s

**Target branch(es):** master

**Tests performed:** Tested it in-game

Fix the action bar to include auto-attack, currently you have to right click the target to start auto-attacking.


Yehonal:
PR imported on staging-branch, ready to be tested on public demo server

Jagermaister95:
I tested the signal on the azerothcore demo server and the fix works properly.

PR #562, opened 2017-08-17
Remodel a XxRenderxX  PR

Ercules76:
**Changes proposed:**
Added a character creations handler

Master

Build: OK
Test in game: OK


Yehonal:
@Inifield spell_generic should contain only scripts based on SpellScriptLoader class

So i think that script is good for now.

src/scripts/World/character_creation.cpp
```diff
@@ -0,0 +1,56 @@
...
+            // Can be modified based on personal needs;
+            return;
+        }
+    }
+};
+
+void AddSC_character_creation()
+{
+    new CharacterCreationProcedures();
+}
```
ShinDarth: please add the new line at the end of file

Ercules76: Next fix i implement the new line.

PR #574, opened 2017-08-22
Fix LFG last player disband party

### LINKED ISSUE #468: [CORE] Looking For Group
**Description**: Group Not remove for Last Player on LFG

**Current behaviour**: after LFG last player always on party and not remove

**Expected behaviour**: need to disband party.

i open PR but can't work with that . really looool

You need go file on  src\game\DungeonFinding\LFGMgr.cpp and after `player->TeleportToEntryPoint();` add this lines : 

        // we need disband group for last player (i know this is not blizzlike)
        if (group->GetMembersCount() == 1)
        {
            group->Disband();
	        sLog->outDebug( LOG_FILTER_LFG, "Player %s is last in lfggroup so we disband the group.", player->GetName().c_str());
        }

this is not blizzlike but if you have another way we can improve it :)

Ercules76:
**Changes proposed:**

**Target branch(es):** 
- Master

**Issues addressed:** Closes #468

- Build OK
- Test in local: OK


src/game/Scripting/ScriptMgr.h
```diff
@@ -1290,13 +1290,15 @@ class ScriptMgr
...
+        //void OnAllCreatureUpdate(Creature* creature, uint32 diff);
         void Creature_SelectLevel(const CreatureTemplate *cinfo, Creature* creature);
 
     public: /* AllMapScript */
 
-        void OnPlayerEnterMapAll(Map* map, Player* player);
-        void OnPlayerLeaveMapAll(Map* map, Player* player);
+        //listener functions are called by OnPlayerEnterMap and OnPlayerLeaveMap
+        //void OnPlayerEnterAll(Map* map, Player* player);
+        //void OnPlayerLeaveAll(Map* map, Player* player);
```
ShinDarth: please clear those commented lines :D

Yehonal: @ShinDarth we kept it because they could be used if we want to split the currently used function

So don't worry :)

Ercules76: Those functions could serve, better to have an explanatory comment

Issue #529, opened 2017-07-26
Trash spawn infinite (The Shattered Halls)


Siid974:
**Description**: The Shattered Halls (Trash spawn infini)

**Current behaviour**: After first boss in the long corridor trash no spawn in wave but spawn when you get to the end of the corridor

**Expected behaviour**: Normally the trash spawn by wave arrived at the beginning of the corridor after killing the non-elite trash that spawn we must kill the trash elite of the bottom

**Steps to reproduce the problem**:

1. Killed the first boss
2. Go to the bottom of the corridor
3. And Find the bug 

**Branch(es)**: Master

**Modules**:  Anticheat / Transmog

**Video**: https://youtu.be/hXWud6JoJwo

sucofog:
Confirmed

Yehonal:
@Siid974 i didn't understand if the issue is about a "cosmetic" since spawns are not at correct position and do not follow a path....or the bug is related to the fact that they spawn unlimited mobs and it shouldn't happen.

please be more clear about it otherwise we don't know what we've to fix exactly.

Yehonal:
Moreover please give us links to creatures from here: https://wowgaming.altervista.org/aowow

Yehonal:
We've fixed following this video: https://youtu.be/Tgok9bDH9N4?t=282

PR incoming ;)

PR #579, opened 2017-08-25
[Script/SAI] Fixed Shattered Hand Blood Guard Event

Yehonal:
**Changes proposed:**

Fixed Shattered Hand Blood Guard Event

expecially:
- [x] fixed arrows and spawns events
- [x] avoid infinite spawns
- [x] added voice audio

How it worked before:
https://www.youtube.com/watch?v=hXWud6JoJwo

How it works now:
https://www.youtube.com/watch?v=mhquiJv1I_M

**Target branch(es):** master

**Issues addressed:** Closes https://github.com/azerothcore/azerothcore-wotlk/issues/529

**Tests performed:** tested on local, it works as in video

In Retail server, the areatriggers seems to start in ~150 yards ...instead in our case it's about ~70 yards, but it's not an huge issue. 
I think we need to create a table where fix this kind of special areatrigger instead of read directly from dbc (or use an npc trigger instead)

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1503630648814585000.sql
+++ b/data/sql/updates/pending_db_world/rev_1503630648814585000.sql
@@ -1,0 +1,21 @@
+INSERT INTO version_db_world (`sql_rev`) VALUES ('1503630648814585000');
+
+-- fix arrows and spawn script, must be started after talk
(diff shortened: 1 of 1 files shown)


Yehonal:
tested again on our demo server

PR #589, opened 2017-08-29
Calendar/GameEvents: Store + Send holiday data to calendar

sucofog:
Core/Events: Some Holidays don't show in calendar
Darkmoon Faire event does not show in calendar.

(Tell us what should happen instead)
Darkmoon Faire event should show in calendar, like all other world events.
This issue is also on Azerothcore latest revision.

Fix by @Gerhood port by Sucofog for AC

**Changes proposed:**

Copypasta from https://github.com/TrinityCore/TrinityCore/pull/20117

Info for all holidays (world events) will be sent to client to have them show in the calendar. To synchronize game event times with the dates shown in the calendar the time of game events linked to a holiday will be adjusted to the holiday time (also, there is no more need to update game_event data for every occurrence). There is a new table for additional holiday dates that are missing in the 3.3.5 client (e.g. Lunar Festival, Noblegarden, Darkmoon Faire).

**Target branch(es):** master

**Issues addressed:** Closes #Calendar Events (none addressed on AC)

**Tests performed:**

Tested, builds without errors and problems.
Working as intended, tested on live realm 60-120 players realm.

**TODO list:**

- Maybe not save time on packed db, instead using unix timestamp (core handled)?
- Maybe load from World::SetInitialWorldSettings instead of sGameEventMgr->LoadHolidayDates(); ?


data/sql/updates/pending_db_world/rev_1504044222560696700.sql
```diff
@@ -0,0 +1,193 @@
...
+RETURNS INT UNSIGNED DETERMINISTIC
+RETURN (yy << 24) | ((mm - 1) << 20) | ((dd - 1) << 14);
+
+INSERT INTO holiday_dates VALUES
+-- (181, 0, 120700928),
+-- (181, 1, 136675328),
+-- (181, 2, 154550272),
+-- (181, 3, 170967040),
+-- (181, 4, 188071936),
+-- (181, 5, 204587008),
```
ShinDarth: why those commented lines?

sucofog: Those commented lines probably here to have a reference, probably wrong date value, but kept here

Gerhood: Those values are already present in client so we dont have to send them. Kept them just for information

ShinDarth: there is no point in keeping them as comments in the SQL update file because anyway there won't be any trace of them after a new DB version is released. Better to create a page in the wiki to keep notes if you really want to save information.

sucofog: Ill remove of this commented lines on next commit, anyways, did you test?

sucofog:
@Gerhood 

There is a trace on console about loading:

image

Significative?

sucofog:
283 284 285 353 400 420 Did check and all are Calls to Arms of the battlegrounds

Gerhood:
Should not happen. Probably some differences between our systems outside this PR maybe I can check it when I have time

ShinDarth:
The best way to test a PR is to do it in a clean environment, using the latest `master` branch without any customisations But if you test it in a live environment with real players, it also helps to hear the feedback.

sucofog:
Tested without any custom content, probably there is some mismatch between the code and dbc start date events

Yehonal:
Please merge updates from master branch, we need to run a travis build to check about warnings.

StygianTheBest:
Tested: Calendar looks fine, but Brewfest is enabled and running (able to get quests, all buildings, etc) today server time 9/19 @ 2:15pm when the event is supposed to start at 9/20 12:01am.

Gerhood:
Brewfest see https://github.com/TrinityCore/TrinityCore/pull/20117/commits/629ddcb2522d751c225e7726c9d1e03f06477914

ShinDarth:
@Gerhood can you open a PR please?

Gerhood:
This is the PR. I opened one for TrinityCore, you'll have to do the porting yourselves

PR #622, opened 2017-09-12
Remove Unneeded code

GeForceBG:
**Changes proposed:**
Removing this will make sure that anything you equip that has a proc will work instantly. So you don't have to wait 30 seconds after equipping to proc.
-  
-  
Example: Equip Shadowmourne it won't proc untill 30 seconds have passed. Same thing for Solace of the fallen and other items that have procs.
-  

- [ ] 
- [ ] 

### DIFF
--- a/src/game/Entities/Player/Player.cpp
+++ b/src/game/Entities/Player/Player.cpp
@@ -23161,14 +23161,6 @@ void Player::ApplyEquipCooldown(Item* pItem)
         if (!spellData.SpellId)
             continue;
 
-        // xinef: apply hidden cooldown for procs
-        /*if (spellData.SpellTrigger == ITEM_SPELLTRIGGER_ON_EQUIP)
-        {
-            // xinef: uint32(-1) special marker for proc cooldowns
-            AddSpellCooldown(spellData.SpellId, uint32(-1), 30*IN_MILLISECONDS);
-            continue;
-        }*/
-
         // wrong triggering type (note: ITEM_SPELLTRIGGER_ON_NO_DELAY_USE not have cooldown)
         if (spellData.SpellTrigger != ITEM_SPELLTRIGGER_ON_USE)
             continue;
--- a/src/game/Entities/Player/Player.cpp
+++ b/src/game/Entities/Player/Player.cpp
@@ -23162,12 +23162,12 @@ void Player::ApplyEquipCooldown(Item* pItem)
             continue;
 
         // xinef: apply hidden cooldown for procs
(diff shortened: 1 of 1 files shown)


src/game/Entities/Player/Player.cpp
```diff
@@ -23162,12 +23162,12 @@ void Player::ApplyEquipCooldown(Item* pItem)
...
 
         // xinef: apply hidden cooldown for procs
-        if (spellData.SpellTrigger == ITEM_SPELLTRIGGER_ON_EQUIP)
+        /*if (spellData.SpellTrigger == ITEM_SPELLTRIGGER_ON_EQUIP)
         {
             // xinef: uint32(-1) special marker for proc cooldowns
             AddSpellCooldown(spellData.SpellId, uint32(-1), 30*IN_MILLISECONDS);
             continue;
-        }
+        }*/
```
ShinDarth: if this code is really unneeded, then it should be actually **removed** and not just commented

GeForceBG: Not sure why xinef would add that it's best to comment it out.

StygianTheBest:
I suggest making this an option in the config.

sucofog:
image

image

GeForceBG:
@StygianTheBest Why make it an option? Imagine Warriors / Paladins in Pvp constantly switching between Shadowmourne and 1H / Shield It will make the proc non existent.

sucofog:
@GeForceBG With a config option, it's a better choice, so we can enable/disable from config, i'm with @StygianTheBest

StygianTheBest:
@GeForceBG Any change that alters the original gameplay should be an option. The power to make these changes should be up to the sysop and not baked into the core.

Yehonal:
Indeed with @StygianTheBest 

Thanks for your contribution @GeForceBG but you need to set it as an optional configuration, or a module if you want to play with them.

Please do it if you can

Yehonal:
closed for now, i'll reopen it when code will be improved

StygianTheBest:
Just out of curiosity, how would you improve this?

Yehonal:
The inplementation of config option is not finished. 
@Sucofog posted screen about how it should be implemented. 

However...are those 30 seconds blizzlike? If they are not the config should be disabled by default

StygianTheBest:
@Yehonal It should absolutely be false. Just making sure I didn't miss something code-wise since I made a recent mod that does the same type of thing. However, I did include a config update.

PR #658, opened 2017-09-22
Fix CMake custom script

BarbzYHOOL:
**Changes proposed:**

-  Fixed the custom script compilation process
-  Added an example of CMakeLists.txt in the Readme.md

**Target branch(es):** master

**Tests performed:** Added a .cpp and .h file in the Custom directory and CMake detected them and the compilation begun, sadly the script in question was not working so it never made to the end, so someone should test this.


Yehonal:
Thank you very much for your contribution!

ShinDarth:
good job

PR #680, opened 2017-10-13
Huge directory Restructuring

### LINKED ISSUE #672: [ENHANCEMENT] Rewrite directory structure for AC 2.X
In AzerothCore 1.x we used first version of DIRS (a proposed standard that can be used to build modular applications)

Since then i've released a new (beta) version of DIRS that is more detailed and introduces some rules to improve software architecture. You can read the documentation [HERE](https://github.com/HW-Core/directory-structure)

For AzerothCore we can use a subset of DIRS to fit our needs. The proposed new structure will be:

- apps: all tools used by this repository, such as db_assembler, installer etc.

- bin: only executable files in a flat structure (no sub-directories here).
         we can put .sh files that runs specific applications.

- conf: configuration files related to this context, in few words for scripts and cmake (not server)
 
- data: all static resources such as documentation, sql files, media etc.

- deps: pre-installed dependencies. 3rd party modules (such as acelib, gsoap, jmalloc etc.) but also standalone azerothcore generic modules that are essential for this system, should be placed in this folder. If a dependency has an external repository, it can be handled by git subtree.

- env: this is the space of sub-environments related to this repository
  - user: gitignored user environment (you can put here anything)
  - dist: can be used as default folder where the server will be installed (if you don't want to install in a custom/system directory)
    - bin
    - lib
    - conf
    - var
      - logs

- modules: in this context will be used for azerothcore server addons

- src: of course the sources to be compiled strictly related to wotlk project

- var: variable/temporary files related to repository
  - build: we can store here the build files generated by the compiler

Note: this kind of structure allows us to have a more flat and detailed structure even restoring original position of deps dir.
Moreover we can use modules folder only for azerothcore extension reorganizing acore, worldengine etc. Under apps/ , deps/ and src/

Yehonal:
**Changes proposed:**

Directory restructuring 

**Target branch(es):** master

**Issues addressed:** Closes #672

**Tests performed:** 
Currently build

- [ ] finish to restructure src/ subdirectories, we should create a more meaningful structure where a developer can easily detect where are sources for:
 - generic purpose libraries developed by AC
 - sources related to mmo framework
 - sources related to wotlk


fmrmx:
So delightful

Re3os:
pleas update log transfer commit on TK.

PR #709, opened 2017-11-16
Fixing NearPoint calculation and fishing height issue #127

### LINKED ISSUE #127: Fishing Error
**Description**: can not fish from a ledge  

**Current behaviour**: Go to bootybay fish. doesnt work u have to be on the same hight as the water. also can not loot grey items. (can from dalaran fountain)

**Expected behaviour**: should be able to fish from ledges.

**Branch(es)**:  1.x 
**AC hash/commit**:  

**Operating system**:  Windows 10
##

Yehonal:
**Changes proposed:**

it should fix this issue https://github.com/azerothcore/azerothcore-wotlk/issues/127

But we must check that it doesn't break something else since there were an hack to skip los in collisions.
We must try sindragosa's tombs for example because this bug could appear: https://github.com/TrinityCore/TrinityCore/issues/6541#issuecomment-147715990

**Target branch(es):** master

**Issues addressed:** Closes #127

**Tests performed:** compiles and fishing works, but need more tests

- [x] test sindragosa's tomb
- [ ] test other similar cases
- [x] maybe import this fix: https://github.com/TrinityCore/TrinityCore/pull/15807

**this PR is already on Staging server if someone want to test without compile/run by themself**


data/sql/updates/db_world/2017_11_16_00.sql
```diff
@@ -0,0 +1,26 @@
+-- DB update 2017_11_14_01 -> 2017_11_16_00
```
Yehonal: Next time the import.sh must be executed only after PR has been merged, to avoid date time conflicts

d3tr0l:
Works fine, tested on my local. Have not found any issues YET after this fix.
Btw i do not understand the need of the sql update. Why is necessary to do
+UPDATE `creature_model_info` SET `BoundingRadius`=7.5, `CombatReach`=6 WHERE `DisplayID`=30890;--
for fixing this issue?

PR #724, opened 2017-12-01
Realmlist returns 0 instead of 1 characters after you deleted all characters

### LINKED ISSUE #629: [CORE][DATABASE] wrong count characters on server list
**Description**: When we create a character, we play and one day we want to erase it or even play God create one to erase it! The server always counts 1 character even if we have already erased everything.

**Current behaviour**: Tell us what happens
The server always counts 1 character even if we have already erased everything.

**Expected behaviour**: Tell us what should happen instead
Counts 0 character and put population to " new players" instead of "low"

1. create character (only one an realm)
2. delete 
3. look your server list and see 1 character
wowscrnshot_091217_140842

**Branch(es)**: Master

**AC hash/commit**:  9d2395e

**Operating system**:  Ubuntu server 16.04.03 LTS

**Modules**:  Are you using modules? if yes, list them

**Other Customizations**: Are you using any core/db patches or other customizations? If yes please specify them here.

Kaev:
**Changes proposed:**
- Removed an unneeded GROUP BY to prevent empty return sets
- Added accountId as a fallback value (SELECT statement would return NULL because there is no entry for your account after you deleted all your characters)

**Target branch(es):** 
master

**Issues addressed:** 
Closes #629

It builds and i tested it in-game without any problems.

### DIFF
--- a/modules/worldengine/nucleus/src/Database/Implementation/CharacterDatabase.cpp
+++ b/modules/worldengine/nucleus/src/Database/Implementation/CharacterDatabase.cpp
@@ -136,7 +136,7 @@ void CharacterDatabaseConnection::DoPrepareStatements()
     PrepareStatement(CHAR_DEL_ACCOUNT_INSTANCE_LOCK_TIMES, "DELETE FROM account_instance_times WHERE accountId = ?", CONNECTION_ASYNC);
     PrepareStatement(CHAR_INS_ACCOUNT_INSTANCE_LOCK_TIMES, "INSERT INTO account_instance_times (accountId, instanceId, releaseTime) VALUES (?, ?, ?)", CONNECTION_ASYNC);
     PrepareStatement(CHAR_SEL_MATCH_MAKER_RATING, "SELECT matchMakerRating, maxMMR  FROM character_arena_stats WHERE guid = ? AND slot = ?", CONNECTION_SYNCH);
-    PrepareStatement(CHAR_SEL_CHARACTER_COUNT, "SELECT IFNULL(account, ?), COUNT(guid) FROM characters WHERE account = ?", CONNECTION_ASYNC);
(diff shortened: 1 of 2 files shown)


sucofog:
So it's tested and working?

Kaev:
Yes :)

talamortis:
can anyone else test please, and if it is working as intended I will merge

Yehonal:
It's now available on our staging server, can be tested there (instruction on discord)

Yehonal:
Ok tested and i'm getting this error during worldserver startup:

```
[ERROR]: In mysql_stmt_prepare() id: 102, sql: "SELECT IFNULL(account, ?), COUNT(guid) FROM characters WHERE account = ?"
[ERROR]: In aggregated query without GROUP BY, expression #1 of SELECT list contains nonaggregated column 'ac_wotlk_chars.characters.account'; 
```

Am i missing something?

Yehonal:
@Kaev read my previous (updated) comment

Kaev:
Will update it as soon as i'm at home, thank you for testing it.
Also going to update my development MySQL server. :p

Inifield:
using mysql 5.7 is not recommended for a live server

Yehonal:
Azerothcore works great on mysql 5.7, i ve it running from months. The fact
is that some query must follow new rules so you cannot break such
constraint.


Il 04 Dic 2017 12:13, "Iny" <notifications@github.com> ha scritto:

Kaev:
Updated, Yehonal's query worked for me without any problems

Yehonal:
image

this is an example of account without characters....is this the behaviour?

Inifield:
I know it work, I added support for it but when I was still using AC I noticed mysql 5.7 did not perform as good as 5.6, I don't have the data anymore but it was tested on 50 players online

Kaev:
@Yehonal Yes, this is the correct behaviour

Yehonal:
@Inifield i'm using percona server based on mysql 5.7 and it's very fast

the reason is that percona just optimize tables and configuration, an optimization that you should do manually on 5.7 and other mysql version ofc.

Issue #679, opened 2017-10-13
[DB] Remove all server loading warnings


Yehonal:
Edit by **Shin** (@FrancescoBorzi):

As of:  https://github.com/azerothcore/azerothcore-wotlk/commit/b1b48b92df68a2d55340d2795a320c24cc45b32f

Current startup errors:

```
2018-12-26 15:46:10 >>  Loaded 0 Players data!
2018-12-26 15:46:14 Table 'mail_loot_template' entry 90 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 119 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 124 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 140 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 180 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 183 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 216 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 217 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 218 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table 'mail_loot_template' entry 270 isn't mail template id and not referenced from loot, and then useless.
2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 4989 Type: 19), ignored.
2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 4988 Type: 19), ignored.
2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5620 Type: 73), ignored.
2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5632 Type: 73), ignored.
2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5619 Type: 73), ignored.
[...]

PR #733, opened 2017-12-09
Fixed warning at server start related to transports

Kaev:
**Changes proposed:**
Fixed the following warning related to transports during server server start:
Warning: GetUInt32() on non-(medium)int field. Using type: SHORT.

MapEntryId is uint32 in DBC and code - Changed smallint to int

**Target branch(es):** master

**Issues addressed:** Part of #679


ShinDarth:
please read:

https://github.com/azerothcore/azerothcore-wotlk/wiki/Dealing-with-SQL-files

also always leave an empty line at the end of source files, especially for SQLs

Kaev:
Like this?
(Somehow GitHub doesn't show the empty line here, if you check the file in the repo it is there :s)

ShinDarth:
Yes, that should be fine. Waiting for some tester feedback

Rochet2:
This should be revertedAnyone remember what this fixed exactly?
Most of the code in core assumes uint16 for mapid.

So either all core types need to be changed for handling the map or just the one that caused the error attempted to fix in this PR.

BarbzYHOOL:
The bug was ignored by everybody because you need to use "-DWITH_COREDEBUG=1" in cmake to notice a change ingame and in the console

So with that flag, it was really horrible bug.

PR #748, opened 2017-12-29
Implement data storage and add it to Player class

Rochet2:
**Problem:**
Modules need to store data for a specific player.
They do so by making a map that maps player guid to some data.
This is not safe when hooks such as PvP kill can be called from two threads at same time and would require locks for every access to the map.

Example module that has the issue: https://github.com/azerothcore/mod-transmog/blob/master/src/Transmogrification.h#L42-L43
Btw I dont think null mutex actually acts as a lock in ace. Could be wrong though.

**Solutions:**
Each module would need to use locks, but they would reduce concurrency by doing so.
Better concurrency is achieved by storing the data directly to the Player class. However that means module makers need to edit Player.h.

**This solution:**
Test online: https://repl.it/repls/ShadyMajesticOcelot
Code: https://pastebin.com/mtimZJHG

The solution is to implement a storage to player class that can store data without knowing the type of it, but can safely return a pointer to it with the correctly requested type. This allows modules to store any data to player without editing player files.

Basically we make a container of pointers and use dynamic cast to convert them to correct user defined type.
Failing the casting or not finding an element results in nullptr.
Memory is managed automatically for the stored data so you can just create, store and forget.
Compile time checks are in place to ensure that set/get objects are derived from required base class. (without this dynamic cast could result in undefined behavior or we could see crash/ub during runtime)

Example of usage:
```c++
#include "DataMap.h"

class MyClass : public DataMap::Base
{
    // anything here
};

// store
player.DataMap.Set("modulename", new MyClass());
// get
if (MyClass* myclass = player.DataMap.Get<MyClass>("modulename"))
    myclass->stuff(); // Do stuff with myclass
// deleted when player logs out
```

The same system could be used per WorldObject for example to allow creature, gameobject, player specific storages and for maps too.

Compiles

- [ ] Test / use it with something
- [ ] Evaluate if it is wanted
- [ ] How to get ppl to use it / know about it?
- [ ] Use pimpl http://en.cppreference.com/w/cpp/language/pimpl

### DIFF
--- a/src/common/Utilities/DataMap.h
+++ b/src/common/Utilities/DataMap.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2008-2017 TrinityCore <http://www.trinitycore.org/>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _DATA_MAP_H_
+#define _DATA_MAP_H_
+
+#include "Define.h"
+#include <string>
+#include <unordered_map>
+#include <memory>
+
+class TC_COMMON_API DataMap
+{
+public:
+    class Base
+    {
+    public:
+        virtual ~Base() = default;
+    };
+
+    template<class T> T* GetCustomData(std::string const & k) const {
+        static_assert(std::is_base_of<Base, T>::value, "T must derive from Base");
+        auto it = Container.find(k);
+        if (it != Container.end())
+            return dynamic_cast<T*>(it->second.get());
+        return nullptr;
+    }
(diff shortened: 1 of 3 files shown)


src/common/Utilities/DataMap.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license: http://github.com/azerothcore/azerothcore-wotlk/LICENSE-GPL2
```
ShinDarth: new files don't require to be released under GPL v2. You can use AGPL here

Yehonal:
It's amazing Idea @Rochet2 ! i was trying to create custom uint32 vectors for world objects to store custom information to be reused by modules. But it's more advanced ofc.  

i'll test with some use-cases. And we should create a documentation about it then.

Yehonal:
@Rochet2 i would accept this PR and starting to test it on custom XP rate module, but before....could you move `DataMap DataMap;` , from Player.h to Object.h please?

Rochet2:
A related commit: https://github.com/azerothcore/azerothcore-wotlk/commit/d85e55dd9b521a24f901d93c34f736ff69b89af5

Rochet2:
A lot of people asked how to use this. Here is an example.

However it also shows that some convenience functions should be implemented.
For example Erase instead of Set with a nullptr and GetDefault to get an object that is default constructed if one doesnt exist. (similar behavior to std::map, but we want type safety)

https://gist.github.com/Rochet2/8fd91f301a255865cb35a5d12341f7a2

Rochet2:
I would like to note that I changed the API slightly, documented the code and implemented the convenience functions in https://github.com/azerothcore/azerothcore-wotlk/commit/0a3be94815675f8a12a5e04824d749c7db8041f0

PR #756, opened 2018-01-14
Implemented guild info command

Nefertumm:
**Changes proposed:**

-  Implemented guild info command.

**Target branch(es):** master

**Issues addressed:** None.

**Tests performed:** Tested and working as intended.
Proof: [https://imgur.com/a/1TvYC](https://imgur.com/a/1TvYC)

**Known issues and TODO list:** None.

### DIFF
--- a/src/server/scripts/Commands/cs_guild.cpp
+++ b/src/server/scripts/Commands/cs_guild.cpp
@@ -184,47 +184,26 @@ public:
 
     static bool HandleGuildInfoCommand(ChatHandler* handler, char const* args)
     {
-        Player* target;
-        uint32 guildId;
-        std::string guildName;
-        std::string guildMasterName;
-        Guild* guild;
+        Guild* guild = nullptr;
 
-        if (!*args)
+        if (args && args[0] != '\0')
         {
-            // Look for the guild of the selected player or ourselves
-            if (target = handler->getSelectedPlayerOrSelf())
-                guild = target->GetGuild();
+            if (isNumeric(args))
+                guild = sGuildMgr->GetGuildById(strtoull(args, nullptr, 10));
             else
-                // getSelectedPlayerOrSelf will return null if there is no session
-                // so target becomes nullptr if the command is ran through console
-                // without specifying args.
-                return false;
+                guild = sGuildMgr->GetGuildByName(args);
         }
-        else if (guildId = atoi(args)) // Try searching by Id
-            guild = sGuildMgr->GetGuildById(guildId);
-        else
-        {
-            // Try to extract a guild name
-            char* tailStr = *args != '"' ? strtok(nullptr, "") : (char*)args;
-            if (!tailStr)
-                return false;
-
-            char* guildStr = handler->extractQuotedArg((char*)args);
-            if (!guildStr)
-                return false;
-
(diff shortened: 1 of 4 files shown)


Nefertumm:
Do i need to satisfy Travis? :thinking:

src/server/scripts/Commands/cs_guild.cpp
```diff
@@ -180,6 +181,65 @@ class guild_commandscript : public CommandScript
...
         return targetGuild->ChangeMemberRank(targetGuid, newRank);
     }
+
+    static bool HandleGuildInfoCommand(ChatHandler* handler, char const* args)
+    {
+        Player* target;
+        uint32 guildId;
+        std::string guildName;
+        std::string guildMasterName;
+        Guild* guild;
```
Rochet2: Having variables at the top is in general not the preferred way in C++.
Please move them as close to where they are used as possible.
Also this may fix travis from complaining as you would do `if (Type* var = GetSomething())` and it would explicitly tell that you are doing an assignment and not possibly some comparing operation.
Also if you need something like guild to be at the top for the code then you should probably initialize it with nullptr when there is no value.

Nefertumm: Thanks @Rochet2 :)

src/server/scripts/Commands/cs_guild.cpp
```diff
@@ -180,6 +181,65 @@ class guild_commandscript : public CommandScript
...
+            guild = sGuildMgr->GetGuildByName(guildName);
+        }
+
+        if (!guild)
+            return false;
+
+        // Display Guild Information
+        handler->PSendSysMessage(LANG_GUILD_INFO_NAME, guild->GetName().c_str(), guild->GetId()); // Guild Id + Name
+        if (sObjectMgr->GetPlayerNameByGUID(guild->GetLeaderGUID(), guildMasterName))
+            handler->PSendSysMessage(LANG_GUILD_INFO_GUILD_MASTER, guildMasterName.c_str(), guild->GetLeaderGUID()); // Guild Master
```
Rochet2: Guid is uint64 and the format in DB is `| Guild Master: %s (GUID: %u)`. This means that you should instead print lowguid in the message. That is uint32.

src/server/scripts/Commands/cs_guild.cpp
```diff
@@ -180,6 +181,65 @@ class guild_commandscript : public CommandScript
...
+                // so target becomes nullptr if the command is ran through console
+                // without specifying args.
+                return false;
+        }
+        else if (guildId = atoi(args)) // Try searching by Id
+            guild = sGuildMgr->GetGuildById(guildId);
+        else
+        {
+            // Try to extract a guild name
+            char* tailStr = *args != '"' ? strtok(nullptr, "") : (char*)args;
```
Rochet2: `strtok(nullptr)` expects a previous invocation of `strtok` with some pointer. You do not do this. Or am I missing it?
http://en.cppreference.com/w/cpp/string/byte/strtok

Nefertumm: Yes, it's better other way. I will commit again for a better version of this.

PR #766, opened 2018-01-24
Fix Unbound Plague, credit to Nefertumm

### LINKED ISSUE #763: Professor Putricide Unbound Plague
**Description**: Professor Putricide Unbound Plague's mechanic is broken.

**Current behaviour**: Unbound Plague is being cast totally randomly. Sometimes not even once during the whole fight, or it is simply cast once at a random Phase (notice that it's always cast on the Tank for some unknown reason)

**Expected behaviour**: Unbound Plague should be cast during the whole fight, from the first phase to the last. It may also happen that you have to handle it during interphases.

1.  Engage Professor Putricide and notice that Unbound Plague is never cast or once randomly in the fight (when it is cast, it pops out on the current Tank, should be a random raid target)

**Branch(es)**: Master

**AC hash/commit**:  29f4a1e090338bc033ebdecb0d71ee45cde6f66f

**Operating system**:  Linux

Gargarensis:
**Changes proposed:**

-  Fix Putricide's Unbound Plague

**Issues addressed:** Closes #763 

**Tests performed:** Not tested by me, maybe @Nefertumm has done it.

**NOTE** 
Fixed by @Nefertumm

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
@@ -173,15 +173,15 @@ public:
     bool operator()(WorldObject* object) const
     {
         if (!object)
-            return true;
+            return false;
         if (Player* p = object->ToPlayer())
         {
             if (p == _source->GetVictim() || p->GetExactDist(_source) >= 45.0f)
-                return true;
+                return false;
 
-            return false;
+            return true;
         }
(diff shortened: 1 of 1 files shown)


ShinDarth:
please set the proper author in the commit:

https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

Nefertumm:
I tested them, and they're good. :smile:

Issue #754, opened 2018-01-12
module translations table


Rochet2:
**Description**:
One thing I think could possibly be useful is having a DB table with this structure:
`modulename, key, english, trans1, trans2, ...`
all columns are string.
And it would be accessible from core.

**Current behaviour**: Tell us what happens
Adding translations to C++ is not really good (code and chinese doesnt mix well).
Existing tables use integer keys and cause collisions and maintenance issues.

**Expected behaviour**: Tell us what should happen instead
Support translations for modules with no collision problems.
Also since the module name is a separate column it is easy to find/modify/delete translations belonging to specific module.

**Steps to reproduce the problem**:

**Branch(es)**: 0.x / 1.x / master (Specify the branch(es) affected by this issue) 

**AC hash/commit**:  
90ed2548c071c1b6a0733042806aa356364ab674

**Operating system**:  

**Modules**:  Are you using modules? if yes, list them

**Other Customizations**: Are you using any core/db patches or other customizations? If yes please specify them here.

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

Nefertumm:
Sounds good

talamortis:
This sounds like an interesting idea maybe another this in this table is instead of using conf to alter options maybe we could have another table

`module_name`,`id`,`option1`,`option2`... this way it will all be database driven

talamortis:
![image](https://user-images.githubusercontent.com/6143065/35126892-5fc1967e-fca7-11e7-9d3e-66eea00ac43c.png)
Working on this as we speak will do a PR if people want it to be merged :)

talamortis:
funny thing is, this is also going to be a module so it gives the players the option :) making hooks and that to make the module work correctly.

BarbzYHOOL:
Nice ideas

PR #774, opened 2018-01-29
Added Hook for modules

talamortis:
**Changes proposed:**

-  Added hook
-  
-  

tested hook and is working

- [  ]
- [ ] 

### DIFF
--- a/src/server/game/Scripting/ScriptMgr.cpp
+++ b/src/server/game/Scripting/ScriptMgr.cpp
@@ -446,6 +446,11 @@ void ScriptMgr::OnBeforeConfigLoad(bool reload)
     FOREACH_SCRIPT(WorldScript)->OnBeforeConfigLoad(reload);
 }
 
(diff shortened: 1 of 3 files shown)


Yehonal:
Please use "new standard" way to naming hook that is composed in this way:

On[Before/After][ActionName]

You should use "before" when it's placed Before that action, and After when it's placed  after the action itself.

In your case should be something like:

OnAfterScriptsLoading

also, what does it means: 

sLog->outString("Loading Modules..."); 

it doesn't load any module

talamortis:
https://github.com/azerothcore/azerothcore-wotlk/issues/754

Yehonal:
Ok but please update hook name and can be merged

Il 30 Gen 2018 00:51, "Lee" <notifications@github.com> ha scritto:

BarbzYHOOL:
@talamortis this never got merged?

PR #780, opened 2018-02-03
Update Creature.cpp

### LINKED ISSUE #777: Aggro is broken
**Description**:
Aggro is not working as intended.

**Current behavior**: Tell us what happens
Mobs pull and link at greater distances and with more sensitivity then they should.  (The configuration file setting is ignored too:  (Line 1652 worldserver.conf: Rate.Creature.Aggro = 1)

**Expected behavior**: Tell us what should happen instead
Aggro should have a maximum range of 45 according to source code Entities -> Creature -> Creatures.cpp line nuber 2706 " // The maximum Aggro Radius is capped at 45 yards (25 level difference)"
And a minimum 20 according to source (cited above) line number 2710 "// The base aggro radius for mob of same level"

1.  A good example is to run  the stockade dungeon
2. While in the hallway mobs will aggro from all rooms instead of none.
3. The issue is not isolated to dungeons it is world wide
4. Try to edit Line 1652 worldserver.conf: "Rate.Creature.Aggro = 1" to a 0.5 or 0 (It won't matter, nothing will fix the aggro)

**Branch(es)**: 0.x / 1.x / master (Specify the branch(es) affected by this issue) 
master

**AC hash/commit**:  
403ed26

**Operating system**:  
Windows 10 x64

**Modules**:  Are you using modules? if yes, list them
None

**Other Customizations**: Are you using any core/db patches or other customizations? If yes please specify them here.
No Customization

Trystanosaurus:
Multiply aggroRadius by setting in config file (which is currently unused).

**Changes proposed:**

-  Multiply aggro radius for creatures by the setting in the config file
-  
-  

**Target branch(es):** master

**Issues addressed:** Closes #777

- [ ] 
- [ ] 

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2728,7 +2728,9 @@ float Creature::GetAggroRange(Unit const* target) const
     float minRange = IsPet() ? 10.0f : 5.0f;
     if (aggroRadius < minRange)
         aggroRadius = minRange;
-
+	
+	aggroRadius *= sWorld->getRate(RATE_CREATURE_AGGRO); // Multiply aggroRadius by setting in config file
+    
     return aggroRadius;
 }
 
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2728,8 +2728,8 @@ float Creature::GetAggroRange(Unit const* target) const
     float minRange = IsPet() ? 10.0f : 5.0f;
     if (aggroRadius < minRange)
         aggroRadius = minRange;
-	
-	aggroRadius *= sWorld->getRate(RATE_CREATURE_AGGRO); // Multiply aggroRadius by setting in config file
+    
+    aggroRadius *= sWorld->getRate(RATE_CREATURE_AGGRO); // Multiply aggroRadius by setting in config file
     
     return aggroRadius;
 }
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2699,6 +2699,10 @@ float Creature::GetAggroRange(Unit const* target) const
     // Determines the aggro range for creatures
     // Based on data from wowwiki due to lack of 3.3.5a data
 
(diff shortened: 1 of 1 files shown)


Yehonal:
Please use 4 spaces instead of 1 tab

Rochet2:
We could maybe introduce the editorconfig that TC has. It is supposed to be a general guide for IDEs on what kind of settings the project uses. VS, VScode, atom, github online editor and many others are able to make use of it directly.
This way there will maybe be less "spaces pls" - comments.

Trystanosaurus:
Should be 4 spaces now. Sorry about that, it's my first go at PR's and things.

dunjeon:
Maybe we should follow the format from other cores? I've tested it and it works as intended.

	float aggroRate = sWorld->getRate(RATE_CREATURE_AGGRO);
	if (aggroRate == 0)
		return 0.0f;
 
	return (aggroRadius * aggroRate);

Trystanosaurus:
Functionally it shouldn't make any difference but their solution is a bit neater and should save a few calculations if aggroRate is 0, so I've made those changes to the PR.

dunjeon:
The following adjustmetns and it appears that AC is in line with the current state of other top cores

Line 2730 add:
// "Minimum Aggro Radius for a mob seems to be combat range (5 yards)"
if (aggroRadius < 5)
    aggroRadius = 5;
// Youfie <Nostalrius> Baron Rivendare
if ((GetEntry() == 10440))
    aggroRadius = 2;

Trystanosaurus:
Hi Dunjeon, 

  The first bit I don't think is needed, AC already has min radius in, it's a little different from other cores because it gives pets a slightly higher minimum radius.

```
    // Minimum Aggro Radius for a mob seems to be combat range (5 yards)
    // hunter pets seem to ignore minimum aggro radius so we'll default it a little higher
    float minRange = IsPet() ? 10.0f : 5.0f;
    if (aggroRadius < minRange)
        aggroRadius = minRange;
```

  I'm sure there's a good reason Nostalrius give Baron Rivendare a hard coded 2 yard minimum radius but it seems a bit odd having the range in there, better would a database solution. Looking at the aggro range function giving him an aura that reduces his aggro range would work.

dunjeon:
The Nostalrius bit does feel like a "hack" solution --  a hotfix fora wip that was likely forgotten.

PR #805, opened 2018-03-03
Remove temporary fix and document possible null in hook

Rochet2:
**Changes proposed:**

-  Remove temporary crash fix (not a real solution)
-  Document possible null in hook which caused the crash to occur

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -71,8 +71,6 @@ Object::Object() : m_PackGUID(sizeof(uint64)+1)
     m_inWorld           = false;
     m_objectUpdated     = false;
 
-    CustomData.Set("base",new DataMap::Base); // avoid crash when access not initialized CustomData
-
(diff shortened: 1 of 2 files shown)


Yehonal:
Has been tested on linux?

Il 03 Mar 2018 20:50, "Rochet2" <notifications@github.com> ha scritto:

Rochet2:
@Yehonal No it is not tested on linux.

BarbzYHOOL:
How to test this?

Rochet2:
I think the problem fixed originally was that using attacker->CustomData.GetDefault<Something>(); was crashing the core. However the reason for it crashing was that attacker was NULL and not that CustomData was somehow faulty.
This commit comments that attacker is null and removes the custom data temporary fix.

The test should basically be calling CustomData.GetDefault for a creature and a player with some custom data class that contains something. This could be done in ModifyPeriodicDamageAurasTick as that is where the original issue was found - except that you should check that attacker is not NULL, but I dont think its mandatory to test it there.

PR #857, opened 2018-05-10
Change determination of pet type

Stoabrogga:
Allow modules like NPC Beastmaster to handle hunter pets for all classes without regression to Azerothcore.

**Changes proposed:**
Change determination of pet type: Instead of evaluating the player class only check if the pet is tameable. Neither demons, elementals nor undead are tameable, so this would be an improvement for the NPC Beastmaster module without regression to Azerothcore.

**Target branch(es):** master

**Tests performed:** tested build, tested in-game

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -671,20 +671,18 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
 
     //Determine pet type
     PetType petType = MAX_PET_TYPE;
+
     if (IsPet() && m_owner->GetTypeId() == TYPEID_PLAYER)
     {
-        if (m_owner->getClass() == CLASS_WARLOCK ||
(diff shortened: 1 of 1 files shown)


AyaseCore:
I've been thinking about a question before. Should we change the hardcode of the core to be compatible with the module?But I don't think it's a good idea.
I wanted to do the ahbot module before, but ahbot needs to change the core.
But not all people will use this module. I should not affect other people because of this module.

Should we change one way:
Provide  *. patch file in the module.
Users who need to use this module apply the patches themselves.

Deku:
It's either applying the patch manually in your fork of the core as @AyaseCore said, or propose a workaround using a hook.

This PR could break some stuff, so it shouldn't be accepted.

Stoabrogga:
Ok, somehow I've expected this response. I too don't have a good feeling about it, but was interested in the feedback. The best way would be to add a patch file to the NPC Beastmaster module which contains the changes to Pet.cpp. People who use the module are not interested in blizzlike behaviour anyway, so they can apply the patch on their own risk.

Issue #711, opened 2017-11-18
Long wait on login


36li:
**Description**:I can't login. I need to wait for tens of minutes

**Current behaviour**: Console message:"Instance ID reset occurred, sending updated calendar and raid info to all players!"

**Expected behaviour**: Getting  list of characters,Message box disappears after tens of minutes.Console message:"Instance ID reset occurred, sending updated calendar and raid info to all players!"

**Steps to reproduce the problem**:

1. Install a new server and start the server

**Branch(es)**: master

**AC hash/commit**:  AzerothCore rev. c4622096e973+ 2017-11-17 00:26:19 +0100 (master branch)

**Operating system**:  Windows 10 x64

**Modules**:  nothing

**Other Customizations**: nothing

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

Yehonal:
cannot confirm, maybe you have some issues on your network?

that console message is not related

sucofog:
Can't confirm

36li:
After a long time, I can play the game.But the same problem happens when you rebuild a characters database

talamortis:
Can not confirm try a recompile closing issue unless someone else can confirm.

Furrazzlhf:
Hello , i have the same problem , i'v to wait a long time during characters login and haven't got any problem on my network ...

talamortis:
are you using any modules?

Furrazzlhf:
I'm using the transmog module

Trystanosaurus:
I know this issue has been closed but I've had the same issue using commit dd463b30b613

Heidi SQL shows one of Azeroth Cores threads trying to run the query "instance_reset SET resettime = ? WHERE mapid = ? AND difficulty = ?" which doesn't seem to want to run but that's using up the thread so the client ends up stuck on retrieving character list.

I'm using SQL 5.6.28, InnoDB.

Inifield:
ok I confirmed the issue, i'll try to fix it soonish

Yehonal:
I can confirm it, please try truncating instance_reset table and restart worldserver

Kron4ek:
Confirm it. Problem disappears after truncating instance_reset.

BarbzYHOOL:
[...]

PR #880, opened 2018-05-20
fix: Stuck at "Retrieving character list".

### LINKED ISSUE #877: Stuck at "Retrieving character list" for 5 minutes
**Description**:

Irregularly, but often enough, when we launch AzerothCore we are stuck at "retrieving character list" just after login on our account. This concerns all the people logging in (players).

**Current behaviour**:

We are stuck during 2 minutes to 5 minutes (maybe more, sometimes we are stuck for more and it times out - this is not verified). Once we are able to choose our characters and enter the game, it never happens again (see **Remarks** below).

**Expected behaviour**:

This should never happen

**Remarks**:

- This has been noticed by MANY people in the discord, by me personally, by AyaseCore too.
- @AyaseCore told me this didn't occur on SunwellCore. This need to be verified (same for mangos, cmangos, trinitycore etc)
- It happened to me only when I didn't start my server for months. Or switched to another unused for a while database (like an older characters db). **So I think this is related to some MySQL cache or something like that** - need opinions on this.

I don't know, it seems like people here have a similar issue (not sure if it's the same) https://github.com/azerothcore/azerothcore-wotlk/issues/711#issuecomment-362535892

**Branch(es)**: 0.x / 1.x / master (Specify the branch(es) affected by this issue) 

**AC hash/commit**:  

**Operating system**:  
On windows or on linux, same thing

PS: I started writing this issue and noticed the other issue afterwards so I let this one to refresh the problem https://github.com/azerothcore/azerothcore-wotlk/issues/711#issuecomment-362535892

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58465793-stuck-at-retrieving-character-list-for-5-minutes?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

AyaseCore:
**Changes proposed:**

**Target branch(es):**  master

**Issues addressed:**  Closes #877 #711 

**Tests performed:** tested in-game

------

If you need to verify it.
Before applying this fix.Running the SQL script in characters DB:
```
update instance_reset set resettime = 12345;
```
Restart worldserver.Login account. You will find stuck at "Retrieving character list".
Apply this fix. Repeat the above steps. You can enter the game.

### DIFF
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -516,8 +516,8 @@ void InstanceSaveManager::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, b
         SetExtendedResetTimeFor(mapid, difficulty, next_reset + period);
         ScheduleReset(time_t(next_reset-3600), InstResetEvent(1, mapid, difficulty));
 
-        //if diff time < 2 reset cycles. not executed the SQL command.
-        if ((now - resetTime) < (period + diff) * 2)
+        //if diff time < 1.1 reset cycles. not executed the SQL command.
+        if ((now - resetTime) < (period + diff) * 1.1)
         {
             // update it in the DB
             PreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_UPD_GLOBAL_INSTANCE_RESETTIME);
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -290,6 +290,12 @@ void InstanceSaveManager::LoadResetTimes()
             SetResetTimeFor(mapid, difficulty, t);
             CharacterDatabase.DirectPExecute("INSERT INTO instance_reset VALUES ('%u', '%u', '%u')", mapid, difficulty, (uint32)t);
         }
+        else
+        {
+            // next reset should be in future. If its not, skip to future.
+            while (t < now)
+                t = uint32(((t + MINUTE) / DAY * DAY) + period + diff);
+        }
         SetExtendedResetTimeFor(mapid, difficulty, t + period);
 
         // schedule the global reset/warning
@@ -512,26 +518,6 @@ void InstanceSaveManager::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, b
             period = DAY;
 
         uint32 next_reset = uint32(((resetTime + MINUTE) / DAY * DAY) + period + diff);
-        uint32 previous_reset = resetTime;
-        uint8 event_type = 1;
-        while (time_t(next_reset - 3600) < now) // while next_reset in past, skip to next reset time
-        {
-            bool warn = event_type < 5;
-            if (warn)
-            {
-                ++event_type;
-                next_reset = next_reset - ResetTimeDelay[event_type - 1];
-            }
-            else
-            {
-                event_type = 1;
-                previous_reset = next_reset;
-                next_reset = uint32(((next_reset + MINUTE) / DAY * DAY) + period + diff);
-            }
-        }
-        if (previous_reset != resetTime && event_type != 1) // if previous_reset is not current reset and next_reset is scheduled from warn, use last non-warning reset time
-            next_reset = previous_reset;
-
         SetResetTimeFor(mapid, difficulty, next_reset);
         SetExtendedResetTimeFor(mapid, difficulty, next_reset + period);
         ScheduleReset(time_t(next_reset-3600), InstResetEvent(1, mapid, difficulty));
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -512,8 +512,25 @@ void InstanceSaveManager::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, b
             period = DAY;
 
         uint32 next_reset = uint32(((resetTime + MINUTE) / DAY * DAY) + period + diff);
-        while (time_t(next_reset - 3600) < now) // while next reset in past, skip to next reset time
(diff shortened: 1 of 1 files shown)


AyaseCore:
In fact, I think this issue is rare.
Only worldserver is a long time not run.This issue will arise.
It needs to update resttime long ago to the present time.
So we have many solutions.

1. Clean up too old data before reading the instance_reset table.
2.Outside the loop. Execute the SQL command.
3.Increase judgment. Reduce SQL command execution.
4.Modified algorithm. For too old data.So let's calculate the final number.Then execute the SQL command.

BarbzYHOOL:
Ok I tried this ```update instance_reset set resettime = 12345;``` on the character DB, and when you connect, you can look at instance_reset and it keeps updating little by little (refresh the view each time).

Of course it takes so long that my server crashes because of the anti stuck configuration "Starting up anti-freeze thread (30 seconds max stuck time)..."

Stoabrogga:
Could someone please explain to me why this should work? Those are all integer values:
next_reset
period
diff
MINUTE (60)
DAY (86400)

Division by integer has no fractional part, e.g. for initial value of 1527088446 for next_reset:
1527088446 + MINUTE = 1527088506
1527088506 / DAY = 17674 (no fractional part)
17674 * DAY = 1527033600

So the calculation

((next_reset + MINUTE) / DAY * DAY)

will always lead to the same value because adding a minute doesn't change the first division (would only make a difference if working with floats). Or am I totally mislead in this case?

AyaseCore:
I guess:
if next_reset = 1527119962(2018/5/24 7:59:22)
1527119962/ DAY = 17674
17674 * DAY = 1527033600(2018/5/23 8:0:0)

1527119962+ MINUTE = 1527120022(2018/5/24 8:0:22)
1527120022/ DAY = 17675
17675 * DAY = 1527120000(2018/5/24 8:0:0)

Stoabrogga:
Sure, but if you have another value for next_reset which lies in the past and is not exactly at the edge of another day you're stuck in the while loop because the value stays always the same.

Rochet2:
> the calculation
> ((next_reset + MINUTE) / DAY * DAY)
> will always lead to the same value because adding a minute doesn't change the first division

The +MINUTE does shift the time calculation a bit in the edge case where time is near the end of a day as ayase noted, but why that would be needed is not clear to me either.
Further inspection shows that the code is 10 years old and can only be traced back to the SVN -> git transition back in 2008.
https://github.com/TrinityCore/TrinityCore/commit/9b1c0e006f20091f28f3f468cfcab1feb51286bd

> you're stuck in the while loop because the value stays always the same

Only if period and diff would be 0. Since they should not be, the time will increase by some amount of seconds and will not be infinite. But it seems that indeed +MINUTE alone does not increase time in pretty much all cases, if that was your point.



I am not familiar with how raid resets are supposed to work. In the code I posted I simply copied the existing logic, but it partially ignores logic in InstanceSaveManager::Update where `bool warn = event.type < 5;` is true. I guess that should be added in at least to keep consistency with the reset times, I cant see if there is other relevant logic I would have missed.

BarbzYHOOL:
I have people who can testify that I wanted solution 4!! :rofl: 

I tested and effectively, it only modified each row once (so only one query per row, only one calculation I guess) and freezing was very short.

Thank you @AyaseCore and @Rochet2 

If there is something else to add or edit, tell me before we merge it

Stoabrogga:
@Rochet2 I think I've got it, thanks. It will only work if the sum of period and diff is greater than or equal to 1 day, which should always be the case.

Hmm, you could also calculate the value directly instead of the while loop. This here should also work:

```
if (time_t(next_reset - HOUR) < now)
{
    uint32 skip_minutes = uint32((now - resetTime + HOUR + MINUTE) / MINUTE);
    next_reset = uint32(((next_reset + (skip_minutes * MINUTE)) / DAY * DAY) + period + diff);
}
```

BarbzYHOOL:
Just retested and it's longer than 40ms because I could refresh 3 times before all the rows were filled :P

AyaseCore:
Either solution 3 or solution 4.
The goal is to reduce unnecessary SQL command execution.
Personally, I prefer solution 3, because it may only take one line of code to solve the issue. :grinning:

BarbzYHOOL:
@Rochet2 said this can be merged. I tested and it's working very well.

Thanks a lot to Rochet2 and @AyaseCore for starting this!

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -290,6 +290,12 @@ void InstanceSaveManager::LoadResetTimes()
             SetResetTimeFor(mapid, difficulty, t);
             CharacterDatabase.DirectPExecute("INSERT INTO instance_reset VALUES ('%u', '%u', '%u')", mapid, difficulty, (uint32)t);
         }
+        else
+        {
+            // next reset should be in future. If its not, skip to future.
+            while (t < now)
+                t = uint32(((t + MINUTE) / DAY * DAY) + period + diff);
```
ShinDarth: is `DAY * DAY` meant to do some rounding?

Stoabrogga: Not the `DAY * DAY` but the `/ DAY * DAY`. The first division by DAY leads to the number of days without fractional part. Multiplicating with DAY results in the exact number of seconds for x days.

ShinDarth: yeah that's what I meant, thanks for clarifying my doubt!

BarbzYHOOL:
Better fix here https://github.com/azerothcore/azerothcore-wotlk/commit/c4a6445a232a3cd8fe8d1e636f91b98c68900943

PR #885, opened 2018-05-23
how_to_make_a_module.md

BarbzYHOOL:
This file is there to warn people to read the wiki and use our skeleton-module (template) so they gain time and they follow the best practice advices

**Changes proposed:**

-  Add a file "tutorial"
-  I don't know if the name is good or not. README.md might not be read 
-  

**Issues addressed:**

Prevents people from wasting time and making half complete modules

- [ ] Add more info


BarbzYHOOL:
I merge it even if it can probably be improved, because more and more people make modules and they don't use the skeleton module.

PR #867, opened 2018-05-13
Bugfix quest "The Endless Hunger": Wrong source for second talk text.

Stoabrogga:
**Changes proposed:**
Currently the second talk text of the Unworthy Initiate has the player as source. This leads to a wrong text when the race variable "$r" is included ("Sate your hunger on cold steel, $r!").
Fix:
Both text lines should have the Unworthy Initiate as source (see also https://youtu.be/U6ZWyvp9b3Q?t=199). Replaced the function call "sCreatureTextMgr->SendChat" with "Talk".

**Target branch(es):** master

**Issues addressed:** Closes #

**Tests performed:** Tested build, tested in-game

**Known issues and TODO list:** none

**NOTE** You no longer need to squash your commits, on merge we will squash it for you. (GitHub added a new feature)

PR #886, opened 2018-05-23
Fix quest 12733 (Death's Challenge)

Stoabrogga:
Prevent random movement of Death Knights; clean up internal player GUID list

**Changes proposed:**
- Follow the player at the beginning of the duel (prevent random movement)
- Clear the internal player GUID list if this NPC is not duelled for 10 minutes; this list is used to track the players which already duelled this NPC (you could theoretically flee from each duel and then have no Death Knights left to solve the quest; also the GUIDs did sum up over time)

**Target branch(es):** master

**Issues addressed:** none

**Tests performed:**
Tested build and in-game

**How to test the changes:**
As GM:
1. .quest remove 12733
2. .quest add 12733
3. .go creature 129520
4. Duel

### DIFF
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -217,6 +217,7 @@ public:
         uint64 _duelGUID;
         EventMap events;
         std::set<uint32> playerGUIDs;
+        uint32 timer = -1;
 
         uint32 GetData(uint32 data) const
         {
@@ -243,6 +244,9 @@ public:
                 playerGUIDs.insert(caster->GetGUIDLow());
                 _duelGUID = caster->GetGUID();
                 _duelInProgress = true;
+
+                timer = 600000; // clear playerGUIDs after 10 minutes if no one initiates a duel
+                me->GetMotionMaster()->MoveFollow(caster, 2.0f, 0.0f);
 
                 events.ScheduleEvent(EVENT_SPEAK, 3000);
                 events.ScheduleEvent(EVENT_SPEAK+1, 7000);
@@ -279,6 +283,16 @@ public:
 
         void UpdateAI(uint32 diff)
         {
+            if (timer != -1 && timer <= diff)
+            {
+                timer = -1;
+                playerGUIDs.clear();
+            }
+            else
+            {
+                timer -= diff;
+            }
+
             events.Update(diff);
             switch (events.ExecuteEvent())
             {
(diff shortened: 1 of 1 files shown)


src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
```diff
@@ -279,6 +283,16 @@ class npc_death_knight_initiate : public CreatureScript
 
         void UpdateAI(uint32 diff)
         {
+            if (timer != -1 && timer <= diff)
```
Nefertumm: I would use a boolean condition instead of this.

Stoabrogga:
@Nefertumm Thanks for pointing this out. "timer" would have been reduced until the integer value overflows. I changed the if statement accordingly.

wetbrownsauce:
Tested. Results:

-Challenged Death Knights indeed follow the player at the start of the duel.
-Upon being defeated, challenged Death Knights start independent 10min timers preventing that player from re-dueling them for 10min.
-After a Death Knight's timer expires, the player can re-duel them.

Possible issue:
-When a Death Knight is defeated by the player, the 10min timer starts for that player only. However, if another player duels that same Death Knight before player one's timer is up, the 10min timer seems to restart for both that new player and the original player. 
To test this potential issue in this PR:
1) Duel a Death Knight NPC with one character on one account and win. Confirm that you cannot re-duel that NPC. 
2) Wait 9minutes
3) Duel that same NPC with a different character on a different account and win. Confirm that you cannot re-duel that NPC on either character.
4) Wait the remaining 1minute in the original players timer.
5) The original character will now have to wait an additional 9minutes to challenge that NPC

Stoabrogga:
@wetbrownsauce thanks for testing. Could you please also take a look at #867? It is in the same area, almost at the beginning when starting with a Death Knight. I added a comment describing how to test it exactly.

> Possible issue:
> -When a Death Knight is defeated by the player, the 10min timer starts for that player only. However, if another player duels that same Death Knight before player one's timer is up, the 10min timer seems to restart for both that new player and the original player.

This works as intended. Currently, there is no reset timer, so when you cancel a duel by running away you could potentially get stuck on this quest, as each NPC tracks all players it has duelled without forgetting them. This is a very unlikely scenario, though, and it is even more unlikely that every NPC is duelled over days each 10 minutes. The method implemented should basically just accomplish a kind of cleanup for the tracked players per NPC.

BarbzYHOOL:
So can be merged?

wetbrownsauce:
Yes it should be fine to merge

FrancescoBorzi:
thanks a lot @Stoabrogga @wetbrownsauce 

merging this.

PR #887, opened 2018-05-24
Feature/devmessage

kraftmatic:
**Changes proposed:**

Adds a message that only displays to users with admin privileges level 2 or higher (as to not nag someone that was just given level one for gm privileges.  Message reminds them to join us on discord and contribute to the project

**Tests performed:**

Compiles, installs and performs as expected on linux.  There are no files loaded or outside resources queried so this should be multifunctional across linux/mac/windows with no modifications needed.

**How to test the changes:**

Make sure you have an account that has gm level set to 2-4 and login.  You should get a yellow onscreen message reminding you about Discord and asking you to contribute to the project

N/A


Deku:
Please develop your feature first and then make the PR, so we avoid having to review all these small commits fixing typos.

kraftmatic:
@Deku Normally I would do that but apparently Travis runs a much more... thorough version of cmake than I do on my system.  Most of these changes are to correct flags for fatal errors and I need Travis to parse it to make sure it will be happy.  The only way to have AC's Travis instance check my code is to have a PR open to trigger it.

Deku:
The Travis' build process is described [here](https://github.com/azerothcore/azerothcore-wotlk/blob/master/.travis.yml) in case you want to check under the same circumstances

PR #889, opened 2018-05-26
Change Trinity Core to AzerothCore

BarbzYHOOL:
**Changes proposed:**

-  Change text "Trinity Core" to "AzerothCore" in conf files

**Target branch(es):**

1.x/2.x etc.

- [ ] I let the PR open a bit in case there are more files to edit. If so comment here, or edit the PR directly


BarbzYHOOL:
or let's add a new label for this?

PR #892, opened 2018-05-29
GroupHandler.cpp - Clean useless group invites

BarbzYHOOL:
**Changes proposed:**

-  Remove pending group invites when we can't invite someone to a group (if group has just been disbanded)
-  Remove pending group invites when we can't invite someone to a group (full group)
- Improve comments

**Target branch(es):**

1.x/2.x etc.

**Tests performed:**

I tested nothing and not even sure about the second change

**How to test the changes:**

This is speculation on how to test:

- Compile
- Invite a player in your group
- With a formed group, invite a new player but disband the group before the player accepts the invite. It should disappear automatically from the player's screen
- With a full group of 5, invite someone. He should not see any invite prompt. You can also try this WITHOUT this pull request and see if he still gets the invite prompt (because I didn't test either)

### DIFF
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -171,15 +171,17 @@ void WorldSession::HandleGroupInviteOpcode(WorldPacket& recvData)
         }
         if (!group->AddInvite(player))
         {
+            group->RemoveAllInvites();
             delete group;
             return;
         }
     }
     else
     {
-        // already existed group: if can't add then just leave
+        // Group already existing, if can't add any more player (full group?), then just leave
         if (!group->AddInvite(player))
         {
+            group->RemoveAllInvites();
             return;
         }
     }
--- a/src/server/game/Handlers/GroupHandler.cpp
+++ b/src/server/game/Handlers/GroupHandler.cpp
@@ -169,6 +169,7 @@ void WorldSession::HandleGroupInviteOpcode(WorldPacket& recvData)
             delete group;
             return;
         }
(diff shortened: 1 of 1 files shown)


wetbrownsauce:
Tested. Does not work. 

> With a formed group, invite a new player but disband the group before the player accepts the invite. It should disappear automatically from the player's screen  

This does not happen, tested with and without PR, invite stays on screen in both cases, only goes away when player clicks accept or decline. 

> With a full group of 5, invite someone. He should not see any invite prompt. You can also try this WITHOUT this pull request and see if he still gets the invite prompt (because I didn't test either)  

The player does not see an invite request with or without this PR, when in a full group and attempting to invite a player, the group leader receives the message "Your party is full".  

Also, when in a group of 4, inviting two more people without either of them accepting or declining will pop up an invite for both those 2 players (as expected), when one accepts and gets into the group (filling it to 5 people), the invite for the other player who was invited stays on screen until they either decline or accept (if they accept it tells that player "your party is full")

BarbzYHOOL:
lol, ok someone else told me similar results (but he was chinese so I wasn't sure)

Not sure what to do with this though

FrancescoBorzi:
@BarbzYHOOL if there is a real issue linked with this, I suggest to open it as normal issue first. This PR can then linked to the issue (just to have a reference to your work - even if apparently it doesn't work) and closed. What do you say ?

BarbzYHOOL:
No issue opened or maybe it was in the discord, I don't remember, I think someone mentioned a fix in TC, I looked at the code, added it and added another part and then I guessed what it did but maybe I was wrong

If someone wants to take a look lol

PR #893, opened 2018-05-30
Gruul: Add a minimum cooldown for Cave In

Trystanosaurus:
Currently if the Gruul fight goes on for long enough he will cast a new cave in every frame. Added a minimum of 2.5 seconds (taken from Trinity).

**Changes proposed:**

-  Add a cooldown to Gruul's cave in ability, else he spams it at the end of the fight.
-  

**Target branch(es):**

1.x/2.x etc. 

**Issues addressed:**

Closes #

**Tests performed:**
Did you test in-game? Yes
What did you test? Fought Gruul, encounter worked correctly
Did you do all these tests on Linux, Mac or Windows? Windows

 

**How to test the changes:**
Try to kill Gruul before and after the changes. You'll need to kill him slowly to see a difference.
Current has Cave in starting at 29 seconds and removing 1.5 per cast. So it takes 21 casts for the Cave in Timer to become negative (around 300 seconds / 5 minutes) so he casts every frame at around 10 growths.

After the changes he will have a minimum time between casts of 2.5 seconds.

- [ ] 
- [ ]

### DIFF
--- a/src/server/scripts/Outland/GruulsLair/boss_gruul.cpp
+++ b/src/server/scripts/Outland/GruulsLair/boss_gruul.cpp
@@ -103,7 +103,8 @@ class boss_gruul : public CreatureScript
                     case EVENT_CAVE_IN:
                         if (Unit* target = SelectTarget(SELECT_TARGET_RANDOM, 0))
                             me->CastSpell(target, SPELL_CAVE_IN, false);
-                        _caveInTimer -= 1500;
+                        if (m_uiCaveIn_StaticTimer >= 4000)
+                            _caveInTimer -= 1500;
                         events.ScheduleEvent(EVENT_CAVE_IN, _caveInTimer);
                         break;
                     case EVENT_REVERBERATION:
--- a/src/server/scripts/Outland/GruulsLair/boss_gruul.cpp
+++ b/src/server/scripts/Outland/GruulsLair/boss_gruul.cpp
@@ -103,7 +103,7 @@ class boss_gruul : public CreatureScript
                     case EVENT_CAVE_IN:
                         if (Unit* target = SelectTarget(SELECT_TARGET_RANDOM, 0))
                             me->CastSpell(target, SPELL_CAVE_IN, false);
-                        if (m_uiCaveIn_StaticTimer >= 4000)
+                        if (_caveInTimer >= 4000)
                             _caveInTimer -= 1500;
                         events.ScheduleEvent(EVENT_CAVE_IN, _caveInTimer);
                         break;


src/server/scripts/Outland/GruulsLair/boss_gruul.cpp
```diff
@@ -103,7 +103,8 @@ class boss_gruul : public CreatureScript
                     case EVENT_CAVE_IN:
                         if (Unit* target = SelectTarget(SELECT_TARGET_RANDOM, 0))
                             me->CastSpell(target, SPELL_CAVE_IN, false);
-                        _caveInTimer -= 1500;
+                        if (_caveInTimer >= 4000)
```
pondaveia: When you said you added a minimum of 2.5 Seconds cooldown on cave in, did you mean 4 seconds?

Trystanosaurus: It only takes off time if the current cooldown is 4 seconds or more, so the final one is a bit less than 4. Working through it, with my code the minimum is 3.5 seconds.

Cast Number | Cooldown before Cast (s) | Total time in Fight (s)
------------ | ------------- | -------------
1 | 29 | 29
2 | 27.5 | 56.5
3 | 26 | 82.5
4 | 24.5 | 107
5 | 23 | 130
6 | 21.5 | 151.5
7 | 20 | 171.5
8 | 18.5 | 190
9 | 17 | 207
10 | 15.5 | 222.5
11 | 14 | 236.5
12 | 12.5 | 249
13 | 11 | 260
14 | 9.5 | 269.5
15 | 8 | 277.5
16 | 6.5 | 284
17 | 5 | 289
18 | 3.5 | 292.5



I had a very quick look for videos or anything on the exact timing but since he casts it without an emote and the cave in can be anywhere in his arena it's hard to find anything concrete.

pondaveia: 👍

wetbrownsauce:
Tested, but sort of unconventionally. I manually added the updated `boss_gruul.cpp` to the most recent version of Azerothcore (fd696c647518facbea1e157618cba1b33893fa83) and tested.   

The reason for that was due to two things:
1) I needed the old vmaps files from early 2018 in order to start the server using this PR.
2) Once started, I could not get past the "Retrieving character list" message when logging in.
However, after adding the `boss_gruul.cpp` from this PR to the newest azerothcore (as stated above), everything ran smoothly with no issue.   

Results: 
Prior to the PR being applied, Gruul would indeed use cave-in on every frame. AFTER applying the new .cpp, cave-in no longer spams on every frame during late fight. After ~5min into the fight, cave-in casts every ~3sec (I was using the in-game timer to track cave-ins, so this is an approximation limited by the lack of milliseconds on the gui). Cave-in no longer spams on every frame as it previously did.  

From what I can see, the change to `boss_gruul.cpp` works correctly.

BarbzYHOOL:
Your issues are not related (discussed in pm), but thanks for testing!

So it can be merged right?

BarbzYHOOL:
@Trystanosaurus thank you for that PR!!

PR #894, opened 2018-05-30
Prince Malchezaar: Add cooldown to Amplify Damage in phase 3

Trystanosaurus:
Added a cooldown to amplify damage in phase 3, else it's cast every frame. Used random between 20 and 30 seconds as that's what's set at the start of phase 3.

**Changes proposed:**

-  Add a cooldown to Prince Malchezaar's Amplify Damage (currently there is none)
-  

**Target branch(es):**

1.x/2.x etc.

**Issues addressed:**

Closes #

**Tests performed:**
Did you test in-game? Yes
What did you test? Fought and defeated the Prince normally
Did you do all these tests on Linux, Mac or Windows? Windows

**How to test the changes:**
Go to the prince in Karazhan and fight him, in phase 3 he currently spams Amplify Damage every frame after the first cast. I've added in a cooldown of 20-30 seconds (same as the inital cooldown is set to when he enters phase 3).

Which commands to use? Which NPC to teleport to? Teleport to Prince Malchezaar (template 15690) and engage, get him to 30% to enter phase 3.

- [ ] 
- [ ]

### DIFF
--- a/src/server/scripts/EasternKingdoms/Karazhan/boss_prince_malchezaar.cpp
+++ b/src/server/scripts/EasternKingdoms/Karazhan/boss_prince_malchezaar.cpp
@@ -421,6 +421,7 @@ public:
 
                     if (target)
                         DoCast(target, SPELL_AMPLIFY_DAMAGE);
+                        AmplifyDamageTimer = urand(20000, 30000);
                 }
                 else
                     AmplifyDamageTimer -= diff;


talamortis:
I must of missed this, thanks will merge now :)

BarbzYHOOL:
You tested it Tala??

src/server/scripts/EasternKingdoms/Karazhan/boss_prince_malchezaar.cpp
```diff
@@ -421,6 +421,7 @@ class boss_malchezaar : public CreatureScript
 
                     if (target)
                         DoCast(target, SPELL_AMPLIFY_DAMAGE);
+                        AmplifyDamageTimer = urand(20000, 30000);
```
ShinDarth: either the code indentation is wrong or you have an issue in your code: the way it is indented, it looks like the `AmplifyDamageTimer` should be part of the `if (target`) block, but it isn't because there are no { brackets }.

So it should be either:

```
if (target)
    DoCast(...);
AmplifyDamageTimer....;
```
or:

```
if (target)
{
    DoCast(...);
    AmplifyDamageTimer....;
}

```
In the first case you DO NOT WANT to include it in the if block, in the second case you DO WANT to include it in the if.

Trystanosaurus:
You're right, it should be the second one, he sets the timer only if he casts the spell, I'd missed the brackets when I copied from my local repository to the git repository, my bad. Does it need a new pull request?

```
if (target)
{
    DoCast(...);
    AmplifyDamageTimer....;
}
```

BarbzYHOOL:
No?? what are you saying?? just edit the file and push a new commit or force push to override old commit

Even you Nefertumm can edit it directly since you're member of AC

BarbzYHOOL:
Ah damn it's the merged PR lol, then it requires a new PR

Nefertumm:
Don't worry, i'll do it, it's only brackets.

PR #924, opened 2018-06-12
Make the continent transport and preloading optional.

FALL1N1:
**Changes proposed:**
Title explains it all. Make the continent transport and preloading optional.
There are two new options
(1) IsContinentTransport.Enabled => This will disable completely the continent transport like ships, zeppelins etc.. useful for PvP servers where players won't have to go from Orgrimmar to Northrend (for example) and greatly reduces memory usage.

(2) IsPreloadedContinentTransport.Enabled => This should be kept disabled(false) on small servers without enough memory as it is right now, on server load it preloads all continent transports and keep them in the memory so you will have ~700mb of unnecessary garbage (for small servers) in your server.

**Target branch(es):**
all i guess? 

**How to test the changes:**
compile and disable then *voila*.. your server consumes 600mb instead of 1300mb+ memory.


FALL1N1:
unknown

src/server/game/Maps/TransportMgr.cpp
```diff
@@ -404,51 +404,57 @@ void TransportMgr::SpawnContinentTransports()
         return;
 
     uint32 oldMSTime = getMSTime();
-    QueryResult result = WorldDatabase.Query("SELECT guid, entry FROM transports");
     uint32 count = 0;
-    if (result)
+    if (sWorld->getBoolConfig(CONFIG_ENABLE_CONTINENT_TRANSPORT))
     {
-        do
+        QueryResult result = WorldDatabase.Query("SELECT guid, entry FROM transports");
```
Nefertumm: This should be outside of the if, it's used on line 434. 👍

Nefertumm: This should be outside of the if, it's used on line 434.

src/server/game/Maps/TransportMgr.cpp
```diff
@@ -404,51 +404,57 @@ void TransportMgr::SpawnContinentTransports()
...
-    if (result)
+        sLog->outString(">> Spawned %u continent motion transports in %u ms", count, GetMSTimeDiffToNow(oldMSTime));
+    }
+    
+    if (sWorld->getBoolConfig(CONFIG_ENABLE_CONTINENT_TRANSPORT_PRELOADING))
     {
-        do
+        // pussywizard: preload grids for continent static transports
+        oldMSTime = getMSTime();
+        result = WorldDatabase.Query("SELECT map, position_x, position_y FROM gameobject g JOIN gameobject_template t ON g.id = t.entry WHERE t.type = 11");
```
Nefertumm: Here. :)

FALL1N1: stupid sunwell and their overwriting.. *sigh*

BarbzYHOOL:
Ok I tested and for me it still costs 900mb of ram (instead of 1500)

Stoabrogga:
Two remarks:
- Shouldn't this if statement
 `if (sWorld->getBoolConfig(CONFIG_ENABLE_CONTINENT_TRANSPORT_PRELOADING))`
 be included in this statement?
 `if (sWorld->getBoolConfig(CONFIG_ENABLE_CONTINENT_TRANSPORT))`
 Doesn't make sense to enable preloading if the transports are disabled.
- I think the default for `IsPreloadedContinentTransport.Enabled` should be 1 to reflect the current behaviour. Otherwise the preloading will be deactivated after merging the PR and people could be confused about it.

BarbzYHOOL:
" Doesn't make sense to enable preloading if the transports are disabled."

Yes, that's what I thought when i disabled in the conf

I also thought about your second issue but I don't know, most people here use AC for small or solo servers so both can be argued for.

Stoabrogga:
Tested the PR and disabled preloading (not the transports). The server needed ca. 300 MB less memory.

BarbzYHOOL:
Thank you @FALL1N1 !! and thank you for test and suggestion @Stoabrogga

PR #931, opened 2018-06-18
Change the read method in the module config file.

AyaseCore:
**Changes proposed:**

-  For module developers. Now they can use:
e.g.
`AC_ADD_CONFIG_FILE("${CMAKE_CURRENT_LIST_DIR}/conf/my_first_module.conf.dist")`
It is used to register module conf files. It can reduce a lot of unnecessary work.

**Target branch(es):** master

1.x/2.x etc.

**Issues addressed:**

https://github.com/azerothcore/azerothcore-wotlk/issues/981

**Tests performed:**

**How to test the changes:**
use this temporary module:https://github.com/AyaseCore/skeleton-module/tree/test

- [x] move LoadModuleConfigSettings() ,Now it looks like a mess.


AyaseCore:
```/home/travis/build/azerothcore/azerothcore-wotlk/src/server/worldserver/Master.cpp:114:9:` fatal error: 
      expected '(' after 'for'
    for each (std::string configFile in configFileList)
        ^
1 error `generated.```
Can't use for each in travis/pr?

File: src/server/worldserver/Master.cpp [hunk 1]
```diff
@@ -108,6 +108,43 @@ class FreezeDetectorRunnable : public ACORE::Runnable
     }
 };
 
+void Master::LodaModuleConfigSettings()
+{
+    Tokenizer configFileList(CONFIG_FILE_LIST, ',');
+    for each (std::string configFile in configFileList)
```
BarbzYHOOL: I think you must remove "each" and that's it (I'm not really into C++ but I looked at other loops in the code and they don't use "each")

AyaseCore: Ok. This is a c++ feature. The compiler used by Travis/pr may not be compatible.

Rochet2: https://docs.microsoft.com/fi-fi/cpp/dotnet/for-each-in

> for each, in
Iterates through an array or collection. This *non-standard* keyword is available in both C++/CLI and native C++ projects. However, its use is *not recommended*. Consider using a *standard* [Range-based for Statement (C++)](https://docs.microsoft.com/fi-fi/cpp/cpp/range-based-for-statement-cpp) instead.

Its a microsoft feature

AyaseCore: @Rochet2 Thank you for your answer! :smiley:

src/server/worldserver/Master.cpp
```diff
@@ -111,9 +111,9 @@ class FreezeDetectorRunnable : public ACORE::Runnable
 void Master::LodaModuleConfigSettings()
```
Winfidonarleyan: Maybe LoadModuleConfigSettings ?

AyaseCore: @Winfidonarleyan  yes. I didn't notice it.:expressionless:

BarbzYHOOL:
Ok, it has been tested by @MajorDOT and @Winfidonarleyan 

it works fine! We just need to update skeleton-module

Todo on skeleton-module:

Remove \src\cmake
Add the line to cmakelists or whatever file

BarbzYHOOL:
The new commit will automatically reload all module conf?? that would be nice

if so, need to add https://github.com/azerothcore/azerothcore-wotlk/issues/981 in the description

PS: would you mind having a review of the skeleton-module @AyaseCore  https://github.com/azerothcore/skeleton-module/pulls

AyaseCore:
> The new commit will automatically reload all module conf??
>>yes,but,modules need to use this approach.

BarbzYHOOL:
Thank you ayase! Now we must tell people to use this method and if possible update modules lol

PR #936, opened 2018-06-23
Core/Scripts: Some Loatheb mods.

FALL1N1:
Changes:
- Move from BossAI to ScriptedAI
- All left spores will now despawn upon boss death.
- Fixed Necrotic Aura's timer.
- New method of spawning spores, we don't have to share them with deathbloom as it has different timer.
- Boss now evades if pulled outside the combat area.

### DIFF
--- a/src/server/scripts/Northrend/Naxxramas/boss_loatheb.cpp
+++ b/src/server/scripts/Northrend/Naxxramas/boss_loatheb.cpp
@@ -23,6 +23,7 @@ enum Events
     EVENT_SPELL_DEATHBLOOM                      = 2,
     EVENT_SPELL_INEVITABLE_DOOM                 = 3,
     EVENT_SPELL_BERSERK                         = 4,
+    EVENT_SUMMON_SPORE                          = 5,
 };
 
 enum Texts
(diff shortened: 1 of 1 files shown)


FALL1N1:
bad travis, need the full compile error so i can fix it, (i don't have local build and i don't have the time to do it)

FALL1N1:
will fix build later, ignore this pr for now

a4501150:
I have a noob question. BOSSAI is inheriting from ScriptedAI, and suppose to be used for boss fight.. why we switch it to ScriptedAI?..

FALL1N1:
@a4501150  because of 'CheckEvadeIfOutOfCombatArea', it is missing from BossAI

FrancescoBorzi:
Looking for @FALL1N1  or someone else to completed the work that has been previously started here.

Selectah:
Well you can also include something like this

```
        bool IsInRoom()
        {
            if (me->GetExactDist(2909.0f, -3997.4f, 274.187f) > 50.0f)
            {
                EnterEvadeMode();
                return false;
            }

            return true;
        }
```

and on UpdateAI

```
            if (!IsInRoom())
                return;
```

So you still can leave Loatheb's BossAI which is anyay better.

FALL1N1:
#1712

PR #939, opened 2018-06-24
Core/Spells: Fixed Blade FURry not sharing damage/ignoring armor to a…

FALL1N1:
title explains it all, blade flurry didn't share damage to additional mobs


BarbzYHOOL:
Thank you @FALL1N1 !!

And thank you @Viste  and @Dreaxxx too

PR #943, opened 2018-06-24
Ahn'Kahet: Amanitar should cast the fungus only on players without "M…

FALL1N1:
Amanitar should cast the fungus only on players without "Mini" aura.


Dreaxxx:
Worked perfectly , but when the player haven't got the "mini" , Fungus appear in the totality of the room (i know you  didn't touch that , it's for a suggestion to reduce perimeter of the pop a small area around the player without the debuff).

BarbzYHOOL:
@Dreaxxx open a new issue for this if it's needed

BarbzYHOOL:
Thank you @FALL1N1

PR #959, opened 2018-07-01
Blade flurry shouldn't target totems.

lineagedr:
Tested in-game and it's working.


Dreaxxx:
Are you sure blade flurry shouldn't target totems ?

lineagedr:
@Dreaxxx Retail PTR 3.3 https://youtu.be/LYe_1qmCeRM?t=64
http://wowwiki.wikia.com/wiki/Shaman_totem 
"Totems are immune to AoEs and DoTs, and must be specifically attacked to be destroyed."
And it was fixed on sunwellpl

talamortis:
Tested and working, Thanks!

PR #970, opened 2018-07-09
Core - Fix grave not found causing crash

BarbzYHOOL:
This fix was provided by @FALL1N1 (can't add author commit directly on github)

**Changes proposed:**

-  Prevent crash when a grave is not found  

**Issues addressed:**

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/963

**Tests performed:**

None

**How to test the changes:**

Not sure how because it's not my issue and not my fix.

- [ ] Test the original issue
- [ ] Test the fix

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -489,7 +489,7 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 {
                     WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
                     
-                    if(grave)
+                    if (grave)
                     {
                         plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
                         plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -490,7 +490,7 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                     if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
                     {
                         plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
-                    plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
+                        plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
                     }
                 }
 
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -487,11 +487,11 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 }
                 else if (!plrMover->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IS_OUT_OF_BOUNDS))
                 {
-                    WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
-
-                    if ( grave)
+                    if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
+                    {
                         plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
                     plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
+                    }
                 }
 
                 plrMover->StopMovingOnCurrentPos(); // pussywizard: moving corpse can't release spirit
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -487,7 +487,9 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 }
                 else if (!plrMover->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IS_OUT_OF_BOUNDS))
                 {
-                    if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
+                    WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
+                    
+                    if(grave)
                     {
                         plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
                         plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());


src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -487,11 +487,11 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 }
                 else if (!plrMover->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IS_OUT_OF_BOUNDS))
                 {
-                    WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
-
-                    if ( grave)
+                    if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
+                    {
                         plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
                     plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
```
FALL1N1: Space

ShinDarth:
I don't really understand what it's actually fixing, it looks more a codestyle change to me.

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -487,11 +487,11 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 }
                 else if (!plrMover->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IS_OUT_OF_BOUNDS))
                 {
-                    WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
-
-                    if ( grave)
+                    if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
```
ShinDarth: missing space after `if`

FALL1N1:
@ShinDarth it will crash on the 2nd line of "grave->" if it's not found from the first declaration WorldSafeLocsEntry const* grave = ...

BarbzYHOOL:
I thought it was because of the `WorldSafeLocsEntry const* grave`

talamortis:
could anyone please give the cords of a GPS? oh which this happens so I could test??

talamortis:
Ok, Now looking at the code again I understand it :P I should of not had a quick glance but why did you remove the `if (grave)`

FALL1N1:
Because now we define the 'grave' inside the if check, saves space and it looks better

FALL1N1:
how to reproduce: purge the graveyard table

MajorDOT:
@FALL1N1 do you mean purge the Game_Graveyard_Zone?
I purged it, but when I die I just get teleported to the Barrens (did not apply patch) I don't crash.
I'll test in a few more areas, but I am not getting any crash.

FALL1N1:
http://prntscr.com/k70ill check the conditions

FrancescoBorzi:
what is the status of this?

BarbzYHOOL:
Waiting for some good input

wetbrownsauce:
Is there a reliable way to reproduce the crash? I've tried dropping all data from `game_graveyard_zone` and dying in various locations, both in world, instance, and bg, but I just get sent to the default barrens gy... no crash.

wetbrownsauce:
From what I understand (I don't know much c++), the code will theoretically execute like this without the PR:


```
//get grave
WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());

//if grave found, move player to it
if ( grave)
    plrMover->TeleportTo(grave->map_id, grave->x, grave->y, grave->z, plrMover->GetOrientation());
plrMover->Relocate(grave->x, grave->y, grave->z, plrMover->GetOrientation());
```

However, the last line in that block is outside the indentation block of the `if ( grave)` meaning it will execute even if the `if ( grave)` check fails. This results in the last line executing on a null value of grave.

The PR appears to change the code style around a bit, unintentionally masking the true effect of indenting that last line and putting {} around the two

BarbzYHOOL:
hmmm then need to unindent the last line?

FrancescoBorzi:
guys the indentation does not affect the behaviour in C++, I don't see how an indentation change can resolve a crash.

I would close this.

BarbzYHOOL:
because

```
if blabla
    do this
do something else
```
means
```
if blabla
{
    do this
}
do something else
```

but the PR does this
```
if blabla
{
    do this
    do something else
}
```

FrancescoBorzi:
@BarbzYHOOL oh yeah, my bad. I was only focussing on the first part and I assumed that the second part was indentation-change only (but it isn't).

Then I would revert the first part and keep the second one.

Nefertumm:
It's not an indentation change. If grave was found or not, relocate will be executed anyway, causing a segmentation fault because grave will be nullptr and you're trying to dereference it.

FALL1N1:
Exactly as @wetbrownsauce said, the grave pointer was outside the if() scope so if the grave was not found the pointer would be null hence it'd crash / cause the segmentation fault.

src/server/game/Handlers/MovementHandler.cpp
FrancescoBorzi: this can be reverted

Nefertumm:
It's not something related to graveyards. It's something like player out of bounds and falling.
https://github.com/azerothcore/azerothcore-wotlk/blob/eb2708bc79f62787d05dff6cd3e4ffae04213290/src/server/game/Handlers/MovementHandler.cpp#L477 Maybe @FALL1N1 can provides more information about how this crash happened to him the first time.

FALL1N1:
@Nefertumm I have written it from the crash logs provided somewhere here in the issue tracker, I don't use ac so I can't reproduce it, sorry.

Nefertumm:
According to the if conditions in the code, this might be tested with the following instructions:
- Drop the game_graveyard_zone table data.
- DIE and release the spirit.
- Go to any map.
- Fall off of that map (Outland might be a good option for this) or go with go xyz command with a very low Z coordinate (remember you can still use commands while death with /g).

Remember not to use gm fly while testing or gm on.

wetbrownsauce:
Tried, no crash. gm off and gm fly off

- jumped off edge of nagrand as ghost: barrens gy
- .go xyz'd to -1000 z in borean tundra both as ghost and while living: barrens gy
- .go xyz'd to -10000 z in borean tundra both as ghost and while living: barrens gy
- .go xyz'd to -10000 z at the edge of the veiled sea fatigue zone (by the blue barriers, literal edge of the map): barrens gy

I'm at a loss of how to reproduce this, let alone how a player would ever run into this

Nefertumm:
Well, we can't reproduce it, but the way that was programmed it's bad. If sometimes that piece of code is executed, it will be a segmentation fault crash. I think this is not intended, they missed the brackets.
In my opinion, I would merge this.

wetbrownsauce:
Fair enough, I agree.

src/server/game/Handlers/MovementHandler.cpp
FrancescoBorzi: You can revert the part where you put the declaration of the variable inside the if

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -487,7 +487,9 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
                 }
                 else if (!plrMover->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_IS_OUT_OF_BOUNDS))
                 {
-                    if(WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId()))
+                    WorldSafeLocsEntry const* grave = sObjectMgr->GetClosestGraveyard(plrMover->GetPositionX(), plrMover->GetPositionY(), plrMover->GetPositionZ(), plrMover->GetMapId(), plrMover->GetTeamId());
+                    
+                    if(grave)
```
FrancescoBorzi: missing space after if. It should be: `if (grave)`

Nefertumm: Done

FrancescoBorzi:
Feel free to merge

Nefertumm:
I'm merging this, it can fix a potential crash. Thank you all ppl!

BarbzYHOOL:
good job all

Issue #739, opened 2017-12-20
[Core]Glyph of the second talent can not be used


pklloveyou:
Description: Glyph of the second talent can not be used

Current behavior: You can not use the glyphs in the first game unless you learn the second talent unless you return to the game again

Expected behavior: Glyphs must be learned simultaneously with two talents and two talents online

Steps to reproduce the problem:

1. Learn double Kai Fu use glyphs,
2. Switch the second talent using the same glyph
3. Glyphs can not be used

**AC hash/commit**:  AzerothCore rev [4bc2124](https://github.com/azerothcore/azerothcore-wotlk/commit/4bc2124632fd6f7120a6b5d6539ec0ad4795c142)

Operating system: Windows Se7en / 64bit

Modules: not used

1.Learn the second talent

2.Use Glyphs for Talents

3.Switch the second talent to use glyphs  >>>>>>>>>>Wrong Tip: You have learned some glyphs

4.Back to the role of choice to enter the game again ....Glyph normal use

pklloveyou:
May be the problem here can not remove the old Glyph of aura
Player.cpp
    for (uint8 slot = 0; slot < MAX_GLYPH_SLOT_INDEX; ++slot)
        if (uint32 glyphId = m_Glyphs[GetActiveSpec()][slot])
            if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
                RemoveAurasDueToSpell(glyphEntry->SpellId);

ShinDarth:
please add your  `AC hash/commit:` value

pklloveyou:
@ShinDarth 
Thanks for reminding

pklloveyou:
Old problems?

BarbzYHOOL:
https://github.com/azerothcore/azerothcore-wotlk/issues/411#issuecomment-405388696

Issue #411, opened 2017-03-03
[CORE] Glyph(s) learnage


Franklampardst:
**Description**: Glyph Learn is Bug

**Current behaviour**: When You use a Glyph from first spec and learn Dual spec you can not use that glyph each used in first spec you must remove first spec glyph.

**Expected behaviour**: you must learn a glyph in 2 spec . first spec & dual spec too

**Steps to reproduce the problem**:

1. use a glyph
2. learn dual spec
3. use that glyph each used from first spec in second spec . you can't

**Branch(es)**: 0.x / 1.x / master (Specify the branch(es) affected by this issue) Master

**AC hash/commit**:  latest AC

**Operating system**:  Se7en 64bit

**Modules**:  Are you using modules? yes . anticheat - transmog

**Other Customizations**: Are you using any core/db patches or other customizations? If yes please specify them here.

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

lif3st3al3r:
Confirm

Franklampardst:
@kvipka @ShinDarth  not have any idea?

ShinDarth:
unfortunately Idk, never investigated on this issue

pklloveyou:
Confirm

pklloveyou:
Quit a game into the game can learn Glyph

pklloveyou:
@Yehonal 
Is it a mechanism that should be added?
Save all glyphs and reload glyph status while switching second genres

lineagedr:
https://github.com/azerothcore/azerothcore-wotlk/blob/dd300003e27ea6b5e8060d9f2adb8bf4cd7e913d/src/server/game/Entities/Player/Player.cpp#L26639
```C++
for (uint8 slot = 0; slot < MAX_GLYPH_SLOT_INDEX; ++slot)
        if (uint32 glyphId = m_Glyphs[GetActiveSpec()][slot])
            if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
                CastSpell(this, glyphEntry->SpellId, TriggerCastFlags(TRIGGERED_FULL_MASK&~(TRIGGERED_IGNORE_SHAPESHIFT|TRIGGERED_IGNORE_CASTER_AURASTATE)));
```

Repalce with:

```C++
[...]

PR #978, opened 2018-07-17
Update Player.cpp

lineagedr:
Fixed: Not being able to apply the same glyph from your previous spec after changing specs.
Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/739 https://github.com/azerothcore/azerothcore-wotlk/issues/411

##### CHANGES PROPOSED:

-  
-  

###### ISSUES ADDRESSED:

Closes 
https://github.com/azerothcore/azerothcore-wotlk/issues/739 
https://github.com/azerothcore/azerothcore-wotlk/issues/411

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26709,9 +26709,15 @@ void Player::ActivateSpec(uint8 spec)
 
     // xinef: apply glyphs from second spec
     for (uint8 slot = 0; slot < MAX_GLYPH_SLOT_INDEX; ++slot)
-        if (uint32 glyphId = m_Glyphs[GetActiveSpec()][slot])
-            if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
-                CastSpell(this, glyphEntry->SpellId, TriggerCastFlags(TRIGGERED_FULL_MASK&~(TRIGGERED_IGNORE_SHAPESHIFT|TRIGGERED_IGNORE_CASTER_AURASTATE)));
+    {
+	    uint32 glyphId = m_Glyphs[GetActiveSpec()][slot];
+
+	    if (glyphId)
+		    if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -26709,9 +26709,15 @@ void Player::ActivateSpec(uint8 spec)
...
-            if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
-                CastSpell(this, glyphEntry->SpellId, TriggerCastFlags(TRIGGERED_FULL_MASK&~(TRIGGERED_IGNORE_SHAPESHIFT|TRIGGERED_IGNORE_CASTER_AURASTATE)));
+    {
+	    uint32 glyphId = m_Glyphs[GetActiveSpec()][slot];
+
+	    if (glyphId)
+		    if (GlyphPropertiesEntry const* glyphEntry = sGlyphPropertiesStore.LookupEntry(glyphId))
+			    CastSpell(this, glyphEntry->SpellId, TriggerCastFlags(TRIGGERED_FULL_MASK&~(TRIGGERED_IGNORE_SHAPESHIFT | TRIGGERED_IGNORE_CASTER_AURASTATE)));
+
+		SetGlyph(slot, glyphId, true);
```
Sdyess: This sets glyph even if glyphId doesn't exist?

lineagedr: @Sdyess Is there an instance when the glyph doesn't exist?

lineagedr: It has to be outside the if statement or else the issue won't be resolved. Besides the code is from TC.

Sdyess: That doesn't even make sense. If the glyphid always exists, it would still always execute that function to set glyph.

BarbzYHOOL:
Thank you for the PR @lineagedr

talamortis:
i agree with @Sdyess with this but thank you for your contribution @lineagedr, Will close this PR

talamortis:
https://github.com/azerothcore/azerothcore-wotlk/commit/562833c9d6959a3f31578c1a9434a06ea3e1bf0f

PR #983, opened 2018-07-19
Core/Script: Warlock lifetap's damage

BarbzYHOOL:
This is again taken from the deceased "crypt-wow" server (AC based, and I knew the developer), I had to analyse the code to find what was changed or not since it was not properly done gitwise, so it's a blind pull request

##### CHANGES PROPOSED:

- Fix a warlock spell's damage? lol

##### TESTS PERFORMED:

None, maybe this PR is useless/invalid. I just took the code from that server

Read the code, then see what spell it affects and test the damage output.

- [ ] Test everything

##### Target branch(es):

Master

### DIFF
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -943,7 +943,8 @@ class spell_warl_life_tap : public SpellScriptLoader
                 Player* caster = GetCaster()->ToPlayer();
                 if (Unit* target = GetHitUnit())
                 {
-                    int32 damage = GetEffectValue();
+                    //int32 damage = GetEffectValue();
+					int32 damage = int32(GetEffectValue() + (6.3875 * GetSpellInfo()->BaseLevel));
                     int32 mana = int32(damage + (caster->GetInt32Value(PLAYER_FIELD_MOD_DAMAGE_DONE_POS+SPELL_SCHOOL_SHADOW) * 0.5f));
 
                     // Shouldn't Appear in Combat Log


Sdyess:
Is there proof this coefficient should be used? This only affects warlock's life tap, so misleading PR title.

Motorheadx:
It is not required you can leave it as it was

d3tr0l:
I examined the code further, prove me if i am wrong. Should be working as intended. But to make it 100% sure, i am gonna test it tomorrow.

d3tr0l:
Very interesting. The mana gain is allright. The life damage is not. The detailed tooltip for this spell is to be found here: https://wotlk.evowow.com/?spell=57946 . Funny part right...from this link https://wotlk-twinhead.twinstar.cz/?spell=57946 it is described without details and the SunwellCore devs propably used this as template. Since, they did not bothered with calculating the HP damage, only leaving the mana as it was intially from TC :) ... So, the tooltip ingame is showing data correctly, but the HP damage to self(player) is not changing. So the Life Tap is definetly bugged.

In your code is this:

int32 damage = int32(GetEffectValue() + (6.3875 * GetSpellInfo()->BaseLevel));

but it might be wrong, since the 
> int32(GetEffectValue()
is okay but the rest is not...see here:
https://wotlk.evowow.com/?spell=57946
it is int32(GetEffectValue() + Spirit*1,5

from your code i understand you are taking the base value as it is right now and you are adding 6.3875*spelldamage? I am not good with core scripting, i am more DB oriented...

It should be basevalue+spirit*1,5 to get it working.

BarbzYHOOL:
@d3tr0l I'm wondering if that fix was taken from a vanilla core instead, as you can see in 2012-2014 it was changed https://github.com/TrinityCore/TrinityCore/issues/5536

So I guess I can just close that PR, right?

d3tr0l:
yes, but keep in mind, the life tap HP damage to player is bugged, the tooltip is working fine, but in combatlog and the real damage is BASE only, so no calculation formula there...kinda in a good way for the warlock using that spell, idk should be put into issues.

BarbzYHOOL:
But you tried with and without the PR ?

d3tr0l:
BarbzYHOOL yes i tested, it is different, but still bugged. It at least "does something" but the formula is wrong :( i am goin to play a little bit with this and i will provide correct formula for your script later this week if i can.

BarbzYHOOL:
play with the normal formula then (not this PR), since it's more recent

PR #989, opened 2018-07-28
fix: config.cmake not work.

AyaseCore:
fix: config.cmake not work.
`conf/config.cmake.dist` copy and rename to `conf/config.cmake`
 `conf/config.cmake` is used to add custom cmake option.(look in https://github.com/azerothcore/azerothcore-wotlk/blob/4d7c06c304e54b52455cd748e4994227b9348482/CMakeLists.txt#L45)
in fact, It should work.
But actually, it doesn't work.
The problem is that "EXISTS" cannot use relative paths.

##### CHANGES PROPOSED:

fix: config.cmake not work.

-  

###### ISSUES ADDRESSED:

Closes 
none

##### TESTS PERFORMED:

Don't need to test.

Don't need to test.

none
- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ include(src/cmake/ac_macros.cmake)
 include(conf/config.cmake.dist)
 
 # load custom configurations for cmake if exists
-if(EXISTS "conf/config.cmake")
+if(EXISTS "${CMAKE_SOURCE_DIR}/conf/config.cmake")
   include(conf/config.cmake)
 endif()
 


BarbzYHOOL:
What was the issue?

AyaseCore:
`conf/config.cmake.dist` copy and rename-> `conf/config.cmake`.
It is used to add custom options.(but it not work.)

Problem discovered by accident. 
Fixed by the way.
It shouldn't need testing. It's just a minor problem.

@BarbzYHOOL

BarbzYHOOL:
ty @AyaseCore ;)

Issue #991, opened 2018-07-29
Vmap & Descriptor Issue


Knindzagxg:
After server release everything was pretty fine.
But, after almost 13 hours, those errors started and auth was locked.
Restarting Auth didn't help - Had to restart world as well -> No errors since  (Almost 7 hours)

I'm not sure what actually triggered, since there was 30+ players online.
Some of the objects are in Hellfire Peninsula, around Dark Portal or etc, and there is no players at all.
Also, i had few DDoS attacks (I'm checking with Host company about details as i'm typing)

https://pastebin.com/QdMmDqtb - Log link

How to reproduce: No idea sadly, i was away from PC when it happened.

System: Debian 9
Modules: Premium Account
Modifications: Nothing custom
AzerothCore rev. 07a451e1402c+ 2018-07-13 21:07:47 +0200 (master branch)

BarbzYHOOL:
We have got a leak. If you run worldserver and type:

    ls -al /proc/*/fd/* 2>/dev/null | egrep ".*log$"

You will see a huge list of worldserver logs opened for no reason. I had 700 DBERRORS.log opened


Temporary fix is to increase your file descriptor limit (look up google for that)

Winfidonarleyan:
I found a mistake, and I think I fixed it.

PR #908, opened 2018-06-04
Show default options, missing in config files

Winfidonarleyan:
#### Made a mistake, created new pull request. Old (https://github.com/azerothcore/azerothcore-wotlk/pull/906)

**Changes proposed:**

* Update code in Log.cpp (To avoid errors when opening authserver.exe)
* If the required settings are not found when the configuration file is opened, a notification is displayed and the value that was set by default

**Target branch(es):** master.

**Issues addressed:** No issue

**Tests performed:** Tested build, tested in-game (Windows 10 / VS 2017)

**How to test the changes:**
* Comment out the option not related to category Log (.conf and .conf.dist)
**Example** MaxPingTime

image

**Known issues and TODO list:**

- Notifications are working, if everything is in order with the options category Log

PR #992, opened 2018-07-30
Core/Logging: Fix out of file descriptors

Winfidonarleyan:
##### CHANGES PROPOSED:
-  Add check before opening a file

##### ISSUES ADDRESSED:
Closes https://github.com/azerothcore/azerothcore-wotlk/issues/991

##### TESTS PERFORMED:
I didn't use the tests on a clean core and got the following;
* Before 900+
* After 27
image

Tested in `Ubuntu 14.04.5 LTS (GNU/Linux 4.9.103-xxxx-std-ipv6-64 x86_64)`

- Before apply test commit
``` shell
Find the process ID of your worldserver :
pgrep worldserver

List the file descriptors (replace XXXXX with the number above) :
ls -al /proc/XXXXX/fd

Count the number of file descriptors :
ls -al /proc/XXXXX/fd | wc -l
```
- Apply test commit
- Repeat step 1

- [x] Need test Unix OS
- [x] Need test Windows OS
- [x] Need test in Build
- [ ] Need test in Game

### DIFF
--- a/src/common/Logging/Log.cpp
+++ b/src/common/Logging/Log.cpp
@@ -446,7 +446,8 @@ void Log::outSQLDriver(const char* str, ...)
     if (!str)
         return;
 
-    sqlLogFile = openLogFile("SQLDriverLogFile", NULL, "a");
+    if (!sqlLogFile)
+        sqlLogFile = openLogFile("SQLDriverLogFile", NULL, "a");
 
     va_list ap;
     va_start(ap, str);
@@ -513,7 +514,8 @@ void Log::outErrorDb(const char * err, ...)
         fflush(logfile);
     }
 
-    dberLogfile = openLogFile("DBErrorLogFile", NULL, "a");
+    if (!dberLogfile)
+        dberLogfile = openLogFile("DBErrorLogFile", NULL, "a");
 
     if (dberLogfile)
     {
@@ -630,7 +632,8 @@ void Log::outSQLDev(const char* str, ...)
 
     printf("\n");
 
-    sqlDevLogFile = openLogFile("SQLDeveloperLogFile", NULL, "a");
+    if (!sqlDevLogFile)
+        sqlDevLogFile = openLogFile("SQLDeveloperLogFile", NULL, "a");
 
     if (sqlDevLogFile)
     {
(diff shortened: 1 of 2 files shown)


BarbzYHOOL:
Discussed with @Stoabrogga (since we found the issue yesterday)

And here is a transcript:

- so it means the logs wont work with this pr?
- I think so, e.g. in function "openLogFile" NULL will be returned if the file is already open. So e.g. in function "outErrorDb", which calls "openLogFile", it will write nothing because NULL is returned.
By the way, in the original code all log files were initialized within the "Initialize" function, don't know why this has been changed. There the log files were initialized when the Log object is created and closed when the Log object is destroyed. I think this was a better way how to handle the log files.


Now I tested the PR and I can confirm this, there is no more hundred of logs opened, there is literally 0 while it should have 1 log of each opened (verified on an older worldserver)

Winfidonarleyan:
#### A little explanation:
* I removed some files `*.log` from initialization to avoid the crash (if open authserver)

Stoabrogga:
But this won't work this way. You have to ensure that each log file is opened only once, which was the case when they were opened during the instantiation of the Log object.

Could you please elaborate on the crash of the auth server? I don't understand this part.

Winfidonarleyan:
In config file [authserver.conf.dist](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/authserver/authserver.conf.dist) there are not all options associated with logging:
* LogDB.Char
* LogDB.RA
* LogDB.GM
* LogDB.Chat
* GmLogPerAccount
* GMLogFile
* CharLogFile
* DBErrorLogFile
* RaLogFile
* ChatLogFile
* SQLDeveloperLogFile
* CharLogDump
* CharLogDump.SeparateDir
* CharLogDump.Separate

Stoabrogga:
Hmm, I never saw such a crash of the auth server. If the specified log option is not set openLogFile will return NULL:
```
std::string logfn=sConfigMgr->GetStringDefault(configFileName, "");
if (logfn.empty())
    return NULL;
```
Misc.log could be a problem, as it is opened by both the auth and world server in the same way:
```
miscLogFile = fopen((m_logsDir + "Misc.log").c_str(), "a");
```
I think this would be a better approach to handle the log files:
- open all necessary log files in the Initialize-function (as it was the case originally); this is an easy method to ensure that all logs are opened only once and thread safe (Log is a singleton)
- make sure that log files, which are not configured, do not crash the server (need to investigate what was the exact cause of it)
- ensure that auth and world server do not write to the same log files

Winfidonarleyan:
Crash happens because the kernel will try to read the options from the config and wants to write that they are not there but cannot because logging cannot be initialized.
https://github.com/azerothcore/azerothcore-wotlk/pull/908

Stoabrogga:
I don't think your PR will work. Something like this
```
std::ofstream OutFile(m_logsDir + logfn);
if (OutFile.is_open())
    return NULL;
```
will not check if the file has been opened earlier.
This implicitly opens the file: `std::ofstream OutFile(m_logsDir + logfn);`
You then check if it is open, which is always true, then you return NULL. In this way no log file will ever be opened. See the ofstream specification for further info:
https://en.cppreference.com/w/cpp/io/basic_ofstream/basic_ofstream

You'll have to rework the whole Log class according to the approach I mentioned in the comment above in order to make it working again without generating hundreds of file descriptors. I can try to rework the class myself but this will take some time.

Winfidonarleyan:
You're right.

Stoabrogga:
Lazy initialization is not thread safe:
```
if (!sqlLogFile)
    sqlLogFile = openLogFile("SQLDriverLogFile", NULL, "a");
```
You should include some kind of locking mechanism. I think this should do the trick (please correct me if I'm wrong):
- Add
 `ACE_Thread_Mutex m_mutex;`
 to the private members in Log.h
 (don't forget `#include <ace/Thread_Mutex.h>`)
- Apply locking to all places where log files are opened, e.g.:
 ```
 {
     TRINITY_GUARD(ACE_Thread_Mutex, m_mutex);
     if (!sqlLogFile)
         sqlLogFile = openLogFile("SQLDriverLogFile", NULL, "a");
 }
```
 (the braces are important to limit the lock)
- The locking has to be applied to all log files:
 sqlLogFile
 dberLogfile
 sqlDevLogFile
 gmLogfile
 charLogfile
 charLogfile
 chatLogfile
 raLogfile
 miscLogFile

Rochet2:
> You should include some kind of locking mechanism

I havent looked into this much, but seems all access to Log singleton is protected by a thread mutex:
`#define sLog ACE_Singleton<Log, ACE_Thread_Mutex>::instance()`
https://github.com/Winfidonarleyan/azerothcore-wotlk/blob/d8c77cfc8160a4b514b010beca76a9b6168c81f6/src/common/Logging/Log.h#L198

Stoabrogga:
@Rochet2 Yes, I saw that, but thought that singletons are only thread safe concerning the constructor (only one instance of the class allowed), not the other functions. Are you sure that the locking mechanism of the singleton covers the whole object?

Rochet2:
> singletons are only thread safe concerning the constructor

Seems this is the case. Checked the sLog return type and it is just the Log instance. Probably should have been a wrapped object instead to provide an actual locking.

Stoabrogga:
I tested the locking I proposed on my own test server for 2 days. Until now without problems (no logging problems, only 1 file descriptor per log file). But I still think it would be a better solution to initialize the log files during the construction of the Log object because then you don't need any further locking. Locking on each log access could be a performance issue (don't know how much overhead it takes to acquire and release the lock).

FALL1N1:
can't we just make a false bool and flag it as true when we initialize them the first time, i think it'd be the most pain-free solution

Stoabrogga:
Same problem. This can lead to a racing condition while writing the new bool value into the memory. This could be fixed by using an atomic bool, though.

Stoabrogga:
On second thought an atomic bool would also be no solution to the thread-safety. It does not prevent two threads opening the same log, you'll have to work with locks here.

So, how to proceed from here?

BarbzYHOOL:
I suppose we can close this

PR #996, opened 2018-08-03
fix logger creating too much file descriptors

### LINKED ISSUE #991: Vmap & Descriptor Issue

Stoabrogga:
##### CHANGES PROPOSED:
This is PR #992 combined with the following modification:
- synchronize the check if a log file is initialized and the initialization itself using "ACE_Thread_Mutex"

##### ISSUES ADDRESSED:
Closes #991 

##### TESTS PERFORMED:
Built and tested in-game

- before applying the changes check the number of opened log files (could be several hundreds):
 `ls -al /proc/$(pgrep worldserver)/fd | egrep ".*\.log$" | wc -l`
- after applying the changes, build and start the core and repeat the command above; the number of opened log files should now be <10

none

##### Target branch(es):
Master

### DIFF
--- a/src/common/Logging/Log.cpp
+++ b/src/common/Logging/Log.cpp
@@ -99,7 +99,7 @@ void Log::Initialize()
     m_logsTimestamp = "_" + GetTimestampStr();
 
     /// Open specific log files
-    logfile = openLogFile("LogFile", "LogTimestamp", "w");
+    openLogFile(& logfile, "LogFile", "LogTimestamp", "w");
     InitColors(sConfigMgr->GetStringDefault("LogColors", ""));
 
     // Main log file settings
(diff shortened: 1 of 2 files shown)


Rochet2:
Assume pointer is 64 bit and half of it is written, then `somelogfile == NULL` would return false, but somelogfile is not usable.
So shouldnt all `FILE*` be wrapped in `std::atomic` to ensure the file is not read as non null and used before the file pointer is written completely?

Are there some problems with moving the initialization to the `Log::Initialize` function completely?

Stoabrogga:
@Rochet2 goot point, had the same thought after creating the PR. I now changed the sync according to the modifications I proposed in this comment: https://github.com/azerothcore/azerothcore-wotlk/pull/992#issuecomment-409304119

As for why not initialize the logs in `Log::Initialize`: This has to be answered by @Winfidonarleyan, as #908 was his PR. I think the problem here is that the config manager in some situations tries to log about missing configuration options, but the log has not initialized because it is dependent on the config manager.

Rochet2:
sConfigMgr should not be used in actively called parts of the code. It is very slow as it most likely actually reads and parses the config files each time. This is why World class has its own storage (cache) for the configs. It seems that currently it is being called inside many log functions which are called very often. (such as chat log)

All the performance, thread safety, initialization and such problems originate from the PR https://github.com/azerothcore/azerothcore-wotlk/pull/908
Maybe it should be reverted (partially?) and reimplemented differently. The PR moved all the config reading to the log functions which is what we dont want. We should revert that part maybe and just slap some default values for them before they are read from configs.
Also if there are some log configs that are not cached in the log object we should cache them there.

Stoabrogga:
Yes, I also think we should revert PR #908. Showing unused config options in the logs stands in no relation to the problems this PR causes. Perhaps this functionality could be implemented as server command.

Stoabrogga:
Close this as #908 has been reverted via #998

PR #1034, opened 2018-09-23
DB: Grimtotem Naturalist fix

### LINKED ISSUE #540: DB/Misc: Grimtotem Naturalist
**Description**: This mob in feralas is scripted in SAI (smart_scripts table)
when he's in tauren form he acts like he's a bear in melee but when he's in bear form he acts like he's tauren and casts from distance. That's why they don't move when you go a bit far from them.

**Current behaviour**: 
He doesn't shift in bear form while having the bear form buff. He uses bear attacks
If you die or flee or put gm on, he goes into bear form but casts spells...

**Expected behaviour**: The opposite

1. .npc add 7726
2. he has bear form and doesn't shift into bear
3. flee or die and see him transforming into bear but then he casts tauren form spells -__-

**Branch(es)**: master (Specify the branch(es) affected by this issue) 

**AC hash/commit**:  from july 2017

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47969127-db-misc-grimtotem-naturalist?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Pondaveia:
##### CHANGES PROPOSED:

- Added SAI to Grimtotem Naturalists

###### ISSUES ADDRESSED:

Closes #540 

##### TESTS PERFORMED:

- Tested in-game

- [x] Fixed morphing back to his model on reset

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1537701976962508800.sql
+++ b/data/sql/updates/pending_db_world/rev_1537701976962508800.sql
@@ -1,0 +1,15 @@
+INSERT INTO version_db_world (`sql_rev`) VALUES ('1537701976962508800');
+SET @ENTRY := 7726;
+DELETE FROM smart_scripts WHERE entryOrGuid = 7726 AND source_type = 0;
+UPDATE creature_template SET AIName="SmartAI" WHERE entry= @ENTRY;
+INSERT INTO smart_scripts (entryorguid, source_type, id, link, event_type, event_phase_mask, event_chance, event_flags, event_param1, event_param2, event_param3, event_param4, action_type, action_param1, action_param2, action_param3, action_param4, action_param5, action_param6, target_type, target_param1, target_param2, target_param3, target_x, target_y, target_z, target_o, comment) VALUES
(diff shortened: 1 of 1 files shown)


d3tr0l:
I would love if the fix of this was so easy. It is not working. TC's SAI fix cannot be used, since the core is not handling it properly. I tried a blizzlike SAI fix for this, and it is simply not working. First of all, the core is not handling the mana checks so that thing cannot be used to enter another phase. Seconds, The melee range is fucked...and sorry for being so rude but it is something that pisses me off a lot. Such a small thing, just one NPC, and it is hard af to fix this. The most I was able to do about this was this: https://gist.github.com/d3tr0l/364749964997a9a95a54c54f5afa2739 i had the solution with mana checks too but i dropped that, because it was just not working even with proper TC SAI parameters. The TC's fix in gist i have here: https://gist.github.com/d3tr0l/d12dfb6e1229236028c25780f781ce3c but if you use the TC one you have to do this too: DELETE FROM `world`.`creature_template_addon` WHERE  `entry`=7726; ...so you do not have the aura on that NPC before it gets into combat.

BarbzYHOOL:
This is working very well, no issue in bear form as D3tr0l said.

This is either perfect or 99% perfect.

PR #1037, opened 2018-09-24
GM: Unify GM commands

### LINKED ISSUE #1019: Unify GM commands
##### SMALL DESCRIPTION:

Unify GM commands because some names are confusing

##### CURRENT BEHAVIOUR:

Most game objects commands are:

`.gobject`

But some are like that:

```
go object
list object
lookup object
```

I think the commands should be all using "gobject" instead of "object".
Maybe, there could be an alias for "object" in order to not break addons like TrinityGM. Or we could fix these addons :P

**There are probably several other commands in that case.**

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

f66e0dc99e33ba3d4bcc1b2b2a689509bb50c816

Pondaveia:
##### CHANGES PROPOSED:

-  Added gobject to the list, look up and go commands in order to maintain a pattern.
-  Did not remove object commands to preserve Trinity add-ons compatibility.

###### ISSUES ADDRESSED:

Closes 
#1019 

##### TESTS PERFORMED:

- Compiles without errors and tested with success

.go gobject 34
.list gobject
.lookup gobject x

##### Target branch(es):

Master

### DIFF
--- a/src/server/scripts/Commands/cs_go.cpp
+++ b/src/server/scripts/Commands/cs_go.cpp
@@ -32,6 +32,7 @@ public:
             { "graveyard",      SEC_MODERATOR,      false, &HandleGoGraveyardCommand,         "" },
             { "grid",           SEC_MODERATOR,      false, &HandleGoGridCommand,              "" },
             { "object",         SEC_MODERATOR,      false, &HandleGoObjectCommand,            "" },
+            { "gobject",        SEC_MODERATOR,      false, &HandleGoObjectCommand,            "" },
             { "taxinode",       SEC_MODERATOR,      false, &HandleGoTaxinodeCommand,          "" },
             { "trigger",        SEC_MODERATOR,      false, &HandleGoTriggerCommand,           "" },
             { "zonexy",         SEC_MODERATOR,      false, &HandleGoZoneXYCommand,            "" },
--- a/src/server/scripts/Commands/cs_list.cpp
+++ b/src/server/scripts/Commands/cs_list.cpp
@@ -31,6 +31,7 @@ public:
             { "creature",       SEC_MODERATOR,  true,  &HandleListCreatureCommand,          "" },
             { "item",           SEC_MODERATOR,  true,  &HandleListItemCommand,              "" },
             { "object",         SEC_MODERATOR,  true,  &HandleListObjectCommand,            "" },
+            { "gobject",        SEC_MODERATOR,  true,  &HandleListObjectCommand,            "" },
             { "auras",          SEC_MODERATOR,  false, &HandleListAurasCommand,             "" }
         };
         static std::vector<ChatCommand> commandTable =
--- a/src/server/scripts/Commands/cs_lookup.cpp
+++ b/src/server/scripts/Commands/cs_lookup.cpp
@@ -43,21 +43,22 @@ public:
 
         static std::vector<ChatCommand> lookupCommandTable =
         {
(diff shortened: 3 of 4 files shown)


BarbzYHOOL:
Hmm cool! Maybe you could also add to the help column of the deprecated commands `[DEPRECATED]: use go gobject` for example

BarbzYHOOL:
so? no update?

src/server/scripts/Commands/cs_lookup.cpp
```diff
@@ -43,21 +43,22 @@ class lookup_commandscript : public CommandScript
...
-            { "object",         SEC_MODERATOR,      true,  &HandleLookupObjectCommand,          "" },
-            { "quest",          SEC_MODERATOR,      true,  &HandleLookupQuestCommand,           "" },
-            { "player",         SEC_GAMEMASTER,     true,  nullptr,                                "", lookupPlayerCommandTable },
-            { "skill",          SEC_MODERATOR,      true,  &HandleLookupSkillCommand,           "" },
-            { "spell",          SEC_MODERATOR,      true,  nullptr,                                "", lookupSpellCommandTable },
-            { "taxinode",       SEC_MODERATOR,      true,  &HandleLookupTaxiNodeCommand,        "" },
-            { "tele",           SEC_MODERATOR,      true,  &HandleLookupTeleCommand,            "" },
-            { "title",          SEC_MODERATOR,      true,  &HandleLookupTitleCommand,           "" },
-            { "map",            SEC_MODERATOR,      true,  &HandleLookupMapCommand,             "" }
+            { "area",           SEC_MODERATOR,      true,  &HandleLookupAreaCommand,        "" },
```
BarbzYHOOL: what happened here??

BarbzYHOOL:
I updated the sql myself

src/server/scripts/Commands/cs_lookup.cpp
Pondaveia: I adjusted the tabs

BarbzYHOOL:
Thank you @Pondaveia and @BarbzYHOOL :P

src/server/scripts/Commands/cs_go.cpp
```diff
@@ -32,6 +32,7 @@ class go_commandscript : public CommandScript
             { "graveyard",      SEC_MODERATOR,      false, &HandleGoGraveyardCommand,         "" },
             { "grid",           SEC_MODERATOR,      false, &HandleGoGridCommand,              "" },
             { "object",         SEC_MODERATOR,      false, &HandleGoObjectCommand,            "" },
```
BarbzYHOOL: Just wondering if we should've added a "deprecated" comment here
```suggestion
            { "object",         SEC_MODERATOR,      false, &HandleGoObjectCommand,            "" }, // DEPRECATED: use gobject instead
```

src/server/scripts/Commands/cs_list.cpp
```diff
@@ -31,6 +31,7 @@ class list_commandscript : public CommandScript
             { "creature",       SEC_MODERATOR,  true,  &HandleListCreatureCommand,          "" },
             { "item",           SEC_MODERATOR,  true,  &HandleListItemCommand,              "" },
             { "object",         SEC_MODERATOR,  true,  &HandleListObjectCommand,            "" },
```
BarbzYHOOL: Same here

Issue #1042, opened 2018-10-01
Crash on command while server running and players online: reload all


d3tr0l:
EXPECTED BLIZZLIKE BEHAVIOUR:

the command should perform without crash. command done inside console, not ingame.

STEPS TO REPRODUCE THE PROBLEM:
launch auth+world ... login into game...into server console type: reload all
and then just wait for tables reloading and crashing

BRANCH(ES):

master
AC HASH/COMMIT:

commit 205e8eb
-> have not updated since
OPERATING SYSTEM:

Win 10 x64, latest update
MODULES:

no
OTHER CUSTOMIZATIONS:

none

If you can not reproduce the crash, i will post logs.

Meltie2013:
I don't think MySQL version has anything to do with the crash during a reload from in-game. Sounds like more of a internal thing. Pretty sure a table in the reload chain is not up to date within the table structures or your command functions are faulty.

d3tr0l:
Weird. I got the database installed and updated properly. Although i get few warnings when importing the db files to mysql. But that is all the time. So is the fault only on my side or is this a issue?

BarbzYHOOL:
I have a recent AC and it does not crash

d3tr0l:
Barbz do you get some warnings when importing the databases to fresh install? Maybe my db import got somehow screwed up idk. Do you have the db on InnoDB?

Meltie2013:
@d3tr0l Appears to be a computer problem on your end mate. I can clearly see the freeze detection kicked in.

d3tr0l:
> 
> 
> @d3tr0l Appears to be a computer problem on your end mate. I can clearly see the freeze detection kicked in.

So maybe poorly configured mysql server? For example, low resources settings.

Meltie2013:
No... I am pretty sure its your computer.
a) not enough memory installed on your computer
b) you have bad CPU usage.

To run most cores, you need 5GB+ to run them decent enough without issues.

d3tr0l:
My cpu never goes over 25% of usage and memory usage is below 80% all the time. I have 8gb ram. I will try this with different mysql server. I am using atm one included in wamp package which sucks. And i am using the my-innodb-heavy-4G configuration file.

d3tr0l:
Can be closed. My bad. The max core stuck time set to default was the cause of this issue. Increased the time by 10 seconds and now it is wirking fine.

BarbzYHOOL:
[...]

PR #1046, opened 2018-10-04
CORE/conf: Note about MaxCoreStuckTime danger

BarbzYHOOL:
##### CHANGES PROPOSED:

-  Warning note to avoid stupid crashes.

###### ISSUES ADDRESSED:

https://github.com/azerothcore/azerothcore-wotlk/issues/1042

Crashes also happen when using the anticheat module with a low MaxCoreStuckTime

Note: maybe there is an issue with .reload all which makes the server considered "stuck". That's why I make this PR, if someone thinks there is a more serious issue, please comment below :)

### DIFF
--- a/src/server/worldserver/worldserver.conf.dist
+++ b/src/server/worldserver/worldserver.conf.dist
@@ -349,7 +349,8 @@ LogDB.Opt.ClearTime = 1209600
 #    MaxCoreStuckTime
 #        Description: Time (in seconds) before the server is forced to crash if it is frozen.
 #        Default:     0   - (Disabled)
-#                     10+ - (Enabled, Recommended 10+)
+#                     10+ - (Enabled, Recommended 30+)
+# Note: If enabled and the setting is too low, it can cause unexpected crash.
 
 MaxCoreStuckTime = 0
 


d3tr0l:
No serious issue at least for now. But i am posting the following suggestion: the antifreeze improvement to make "checkpoints" before the server executes such queries or larger operations for exaple reloading or when there is db dump in progress so it does allow some sort of detection tolerance range in seconds?For example lookup the size or lenght of query then precalculate possible tolerance and then run that execution/query with adjusted antifreeze, repeat process for every other action. I know most of advanced hardware solutions might not encounter these issues. And i do not know if the antifreeze is a part of the compiler solution or coded by mangos/tc/swc/ac contributors. Also, if the server crashes there must propably be some sort of OS app or script checking for such case. I once used auto restarter script in case of server freeze or crash but it never worked. I was away for longer than 12hrs and the server was down because of this since no other staff member had root acces to the server. Is there a possibility to get such a thing for acore? Or maybe even code that into the core, so the core itself has some sort of failsafe checking for crash/freeze and automatically attempts to restart. I noticed that having the core as os service simply is not enough as a solution. Because having the server frozen might never notify the os about and thus it does not restart. So it just keeps hanging for eternity. There might be external solutions, but having something like this as a part of the core would be very professional. Like core inside a core. One with world+auth, and another as security solution and monitoring the status of these two. But it is only a suggestion and even if this is accepted it sounds like an overkill at least to me.

Sdyess:
In general this shouldn't be a problem for normal people. On top of that .reload all is a lot and i cant imagine there are a ton of cases in which you should be using reload all vs reloading 1-2 tables at a time. The servers are not meant to be reloading like that, there is a reason it all gets cached at startup in containers.

As for the anticheat module, that seems to end up being a problem for the module and not for AC in general.

The changes are fine to merge but this is afaik, specifically because of one person having issues.

BarbzYHOOL:
Thank you @Sdyess I'll merge because it won't hurt anybody :p

PR #911, opened 2018-06-06
Change encoding in banner

Winfidonarleyan:
**Changes proposed:**
-  Correct encoding in banner (UTF-8 no BOM)
-  Correct line in banner

**Target branch(es):** master.

**Issues addressed:** Close (https://github.com/azerothcore/azerothcore-wotlk/issues/910)

**Tests performed:** Test build and test run.

**Tests platform:**  Windows 10. Visual Studio 2017 (15.7.3) x64

**How to test the changes:**
1. Apply my commit
2. Rebuild project
3. Run authserver and worldserver

**Known issues and TODO list:** No

PR #1055, opened 2018-10-18
save Main.cpp&Master.cpp with bom to avoid Compiler Error C2001 in vs…

iamchenxin:
…2017

##### CHANGES PROPOSED:

-  saved Main.cpp&Master.cpp with utf8bom to avoid complier error in vs2017
-  not sure if there are some Complier in Mac&Linux can not deal with bom, need tests by reviewers who use Mac and Linux. (by google result, seems all modern Compliers can deal with bom)

###### ISSUES ADDRESSED:
Compiler Error C2001 in vs2017

##### TESTS PERFORMED:
just changed the encoding of file,  do not need program tests.  but need review by different  compiler user.

need some compile report  by different  compiler users.

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/authserver/Main.cpp
+++ b/src/server/authserver/Main.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license: https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-GPL2
  * Copyright (C) 2008-2016 TrinityCore <http://www.trinitycore.org/>
  * Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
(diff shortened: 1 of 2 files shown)


BarbzYHOOL:
What about this? https://github.com/azerothcore/azerothcore-wotlk/pull/911

a4501150:
with bom, some Linux compilers may complain.

Viste:
OMG

iamchenxin:
> with bom, some Linux compilers may complain. Also bom is widely abused used by microsoft, for your problem just change your .editorsetting file to
> `charset = utf-8-bom`
> re-run cmake and everything is working as intended and with this option you need to add u8 before string's ""

Microsoft's option is here: https://msdn.microsoft.com/en-us/library/mt708821.aspx.
Seems it is a global environment option for project, is there some way to restrict it to a specific file ? or it will need to modifiy all the files in project to unicode.

iamchenxin:
Or just document this issues in installation guide , then some one who use vs2017 can save those files manually.
Yesterday I found something texts from Microsoft, [new options for managing character sets in the microsoft cc compiler](https://blogs.msdn.microsoft.com/vcblog/2016/02/22/new-options-for-managing-character-sets-in-the-microsoft-cc-compiler/).
They said:

> In a future major release of the compiler, we would like to change default handling of BOM-less files to assume UTF-8, but changing that in an update has the potential to cause too many silent breaking changes. Validation of UTF-8 files should catch almost all cases where that is an incorrect assumption, so my hope is that it will happen.

Maybe Ms will solve this problem in vs2019?

BarbzYHOOL:
Anybody can edit the wiki or propose changes in the discord

I also feel like this PR is not wanted but the issue is useful, we probably need to write the solution in the wiki. Someone can do it or explain clearly the solution ?

PR #1061, opened 2018-10-23
Fix voice samples for Headless Horseman during quest "Let the Fires Come!"

Stoabrogga:
##### CHANGES PROPOSED:
Add voice samples for the following texts:
- "The sky is dark. The fire burns. You strive in vain as Fate's wheel turns."
- "The town still burns. A cleansing fire! Time is short, I'll soon retire!"

##### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
Tested build on Ubuntu 16.04; tested in-game

Start daily quest "Let the Fires Come!", do not put out the fires.

none

##### Target branch(es):
Master

### DIFF
--- a/src/server/scripts/Events/hallows_end.cpp
+++ b/src/server/scripts/Events/hallows_end.cpp
@@ -662,9 +662,15 @@ class npc_hallows_end_soh : public CreatureScript
                             return;
                         }
                         if (counter == 5)
+                        {
                             me->MonsterYell("The sky is dark. The fire burns. You strive in vain as Fate's wheel turns.", LANG_UNIVERSAL, 0);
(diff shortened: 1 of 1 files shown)


Stoabrogga:
Sorry, but I don't have much time to spend, just wanted to fix this on my own server and create some kind of feedback for AC.

BarbzYHOOL:
Thank you @Stoabrogga !

@Sdyess I think it would require to improve the structure of every similar files etc, not doing it on a case by case basis :(

Sdyess:
@BarbzYHOOL Allowing a bad practice on a case by case basic is still a bad practice. Improving one file at a time is better than adding more code with the same poor practice.

PR #1091, opened 2018-11-28
Apps: fix db_assembler typo

BarbzYHOOL:
Just look at the file :P

@Yehonal nothing to say against it ?? it won't cause issues ?
We should instead get the name of the directory with basename command and make it a variable

Note: I noticed this when automating installation with an external tool and comparing with the db_assembler.sh


FrancescoBorzi:
@BarbzYHOOL what is the meaning of this? If I understood correctly, that is a suffix given to generated update files, in order to distinguish between base (`_base`), update (`_update`) and custom (`_custom`).

Why should we should change the suffix from `_update` to `_updates`?

BarbzYHOOL:
Hmm I thought I had written it

Well it's because it's the name of the directories, so it corresponds to the directories EXACTLY

BarbzYHOOL:
@FrancescoBorzi see the dir name https://github.com/azerothcore/azerothcore-wotlk/tree/master/data/sql/updates

FrancescoBorzi:
I would wait for approval from @Yehonal for this, just to be safe (I don't know bash very well)

BarbzYHOOL:
I tried it and it worked but we gonna wait forever for such a small thing lol

Issue #905, opened 2018-06-02
Escape Through Force - Quest


Knindzagxg:
Description: While escorting and selecting "Force option, since there is Stealth as well" attacking Owls should appear and you will have to fight your way out.

Current behaviour: There are no attacking Owls at all, which should come in waves once u trigger event.

Expected behaviour: Owl (Attacking) NPCs should appear. Waypoints and movement are good tho.

Steps to reproduce the problem: Will leave to someone with more exp - SAI problem.

Quest ID - 994
AC hash/commit: 778cd9a

Operating system: Linux, Debian 9

Modules: Congrats on level, Premium

Other Customizations: No.

Knindzagxg:
Quest work, but there should be attacking NPC's that come during escort.
At the moment, you just follow him and complete quest.

BarbzYHOOL:
tried the fix?

Issue #904, opened 2018-06-02
Escape Through Stealth - Quest


Knindzagxg:
**Description**: While escorting and selecting "Stealth option, since there is force as well" NPC should be stealthed while moving. Also, attacking Owls should be appear as well.

**Current behaviour**: There's no stealth on NPC and attacking Owls doesen't appear. Waypoints and movement is good tho.

**Expected behaviour**: Owl (Attacking) NPCs should appear and Volcor should be in stealth.

**Steps to reproduce the problem**: Will leave to someone with more exp - SAI problem.

Quest ID - 995
**AC hash/commit**:  778cd9a

**Operating system**:  Linux, Debian 9

**Modules**:  Congrats on level, Premium

**Other Customizations**: No.

Nefertumm:
#190

Issue #190, opened 2016-10-22
[Quest] Escape through stealth id995


Thurin93:
**Description**: The problem is the same as issue #189 
During this quest, you have to save Volcor from a cave. You have to defend him from any mobs and when near his bear Grimclaw, the quest is complete.

**Current behaviour**: Since Grimclaw is far away from the right spot, the quest won't be completed and Volcor disappears sitting.
![grimclaw](https://cloud.githubusercontent.com/assets/16744874/19622032/4e1bb57e-98a0-11e6-97f6-90fb7fcbb542.PNG)
 I inserted a screenshot. The arrow points where grimclaw should be, and my character where Volcor sits before disappearing.

**Expected behaviour**: Volcor says something and then the quest is complete.
http://www.wowhead.com/quest=994/escape-through-force Here you can see the right run of the quest.

**Steps to reproduce the problem**:
1. add quest id 995
2. Go in Darkshore and found Volcor
3. Follow him until near Grimclaw

**Branch(es)**: master

**AC hash/commit**:  https://github.com/azerothcore/azerothcore-wotlk/commit/8eb135e5456acebf44108f01bbee67fb07a4a317

**Operating system**:  

**Modules**:  duel reset

**Other Customizations**: crossfaction
##

Issue #189, opened 2016-10-22
[Quest] Escape Through Force Id 994


Thurin93:
**Description**: During this quest, you have to save Volcor from a cave. You have to defend him from any mobs and when near his bear Grimclaw, the quest is complete.

**Current behaviour**: Since Grimclaw is far away from the right spot, the quest won't be completed and Volcor disappears sitting
![grimclaw](https://cloud.githubusercontent.com/assets/16744874/19622032/4e1bb57e-98a0-11e6-97f6-90fb7fcbb542.PNG)
. I inserted a screenshot. The arrow points where grimclaw should be, and my character where Volcor sits before disappearing.

**Expected behaviour**: Volcor says something and then the quest is complete.
http://www.wowhead.com/quest=994/escape-through-force Here you can see the right run of the quest.

**Steps to reproduce the problem**:
1. add quest id 994
2. Go in Darkshore and found Volcor
3. Follow him until near Grimclaw

**Branch(es)**: master

**AC hash/commit**:  https://github.com/azerothcore/azerothcore-wotlk/commit/8eb135e5456acebf44108f01bbee67fb07a4a317

**Operating system**:  

**Modules**:  duel reset

**Other Customizations**: crossfaction
##

HeartWell:
```sql

-- 
-- DB/Quest: Escape Through Force
DELETE FROM `smart_scripts` WHERE `entryorguid`=3692 AND `source_type`=0;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(3692, 0, 0, 1, 19, 0, 100, 0, 994, 0, 0, 0, 64, 1, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 'Volcor - On Quest \'Escape Through Force\' Taken - Store Targetlist'),
[...]

PR #1100, opened 2018-12-07
DB/Quest Escape Through Force&Stealth

poszer:
Quest Escape Through Force & Escape Through Stealth

**CHANGES PROPOSED:**

-  All changes has done in DB ( Updates, SAI Edit) 
-  Sai was set to cast spell 6928 on NPC ( And spell with that id doesn't exist). 
- It's changed to use AURA 6928 (It's Form of the Moonstalker **Invisibility**) 
- Also, the Bear creature has update to correct position

**Closes** 
#189 
#905 
#904
#190 

**TESTS PERFORMED:**
Builded without errors. Tested in game, looks like it's working fine now.
[Test Video](https://youtu.be/KtVx5tB-0zg)
Also have tested this second quest ( Escape Through Stealth ) how it should working.
Since the Npc will use invisibility aura, there is no chance that he can be attacked by mobs.
The quest will be done after NPC leave that cave
I'll include video from other Emu's how the quest is done.
[Video show: How it should be](https://www.youtube.com/watch?v=y5AMXINUv3Q)

**HOW TO TEST THE CHANGES:**
.go xyz 4601.410645 -4.852083 69.678398
.quest add 993
Complete it by handing quest to NPC near you '' Volcor ''
After that try to do these two quests from Volcor :

1. Escape Through Force
2. Escape Through Stealth

**Target branch(es):**

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1544169970525311500.sql
+++ b/data/sql/updates/pending_db_world/rev_1544169970525311500.sql
@@ -1,0 +1,40 @@
+INSERT INTO version_db_world (`sql_rev`) VALUES ('1544169970525311500');
+DELETE FROM `smart_scripts` WHERE `entryorguid` IN (3692, 369200, 369201);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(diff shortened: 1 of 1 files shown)


wetbrownsauce:
Tested.  

-Both Escape through Force and Stealth complete successfully.
-In Escape Through Stealth, Volcor correctly stealths and is not attacked by mobs.
-In Escape Through Force, Volcor paths all the way to the proper road, fighting things along the way.

I cannot confirm if the mobs were correct during Escape Through Force, I fought the already spawned enemies at the entrance of the cave during the escort, whereas videos such as [this](https://youtu.be/wDJ4Un0EuJQ?t=1616) show 2 extra groups of enemies which spawn later during the escort.

Nefertumm:
Should we merge this? @wetbrownsauce

wetbrownsauce:
I would merge it since it does correct the issues that currently prevent the quest from being completed. I'm not sure if the mob spawns during the escort are correct but I suppose that is a separate issue from this.

poszer:
> 
> 
> Should we merge this? @wetbrownsauce

Since the quest was fcked totally before, i don't see any reason why not? It's fixed almost at all, just some npcs aren't spawned outside, but all other things works perfectly

Nefertumm:
I'll merge then :) Thanks @poszer and @wetbrownsauce.

PR #1158, opened 2018-12-29
Core/Quest Mr.Floppy's Perilous Adventure Quest

poszer:
Mr. Floppy's Perilous Adventure Quest - Fixes for visual effects, and NPC not attacking while walking

##### CHANGES PROPOSED:

-  Updated
-  

###### ISSUES ADDRESSED:

The issue which can be reproduced are that when doing quest, some npc's doesn't attack the Quest Creatures which is incorrect. They should be attacked while quest is in progress. Also Visual effects doesn't work
Closes 

##### TESTS PERFORMED:

Builded without errors on Win, works fine now.

1.  .go xyz 4326.597168 -3691.362549 263.840363 571
2.  Take the quest from NPC in front of you (Emily)
3. Or add it by command .quest add 12027  (Mr.Flppy's Perilous Adventure)
4. Continue to do quest until end

- [ ]
- [ ] 

##### Target branch(es):

Master

@Stoabrogga Thanks!!

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1546128548989877900.sql
+++ b/data/sql/updates/pending_db_world/rev_1546128548989877900.sql
@@ -1,0 +1,3 @@
+INSERT INTO version_db_world (`sql_rev`) VALUES ('1546128548989877900');
+UPDATE creature_template SET AIName="CombatAI" WHERE entry IN (26586, 26590);
+
--- a/data/sql/updates/pending_db_world/rev_1546128548989877900.sql
+++ b/data/sql/updates/pending_db_world/rev_1546128548989877900.sql
@@ -1,3 +1,4 @@
 INSERT INTO version_db_world (`sql_rev`) VALUES ('1546128548989877900');
-UPDATE creature_template SET AIName="CombatAI" WHERE entry IN (26586, 26590);
 
+UPDATE `creature_template` SET `VehicleId` = 0 WHERE `entry` = 26586;
+UPDATE `creature_template` SET `ScriptName` = 'npc_ravenous_worg' WHERE `entry` = 26590;
--- a/src/server/scripts/Northrend/zone_grizzly_hills.cpp
+++ b/src/server/scripts/Northrend/zone_grizzly_hills.cpp
@@ -140,7 +140,12 @@ public:
                     if (ObjectAccessor::GetCreature(*me, _mrfloppyGUID))
                     {
                         Talk(SAY_WORGHAGGRO1);
-                        me->SummonCreature(NPC_HUNGRY_WORG, me->GetPositionX()+5, me->GetPositionY()+2, me->GetPositionZ()+1, 3.229f, TEMPSUMMON_TIMED_OR_DEAD_DESPAWN, 120000);
+                        if (Creature* worg = me->SummonCreature(NPC_HUNGRY_WORG, me->GetPositionX() + 5, me->GetPositionY() + 2, me->GetPositionZ() + 1, 3.229f, TEMPSUMMON_TIMED_OR_DEAD_DESPAWN, 120000))
+                            if (Creature* Mrfloppy = ObjectAccessor::GetCreature(*me, _mrfloppyGUID))
+                            {
+                                worg->SetReactState(REACT_AGGRESSIVE);
+                                worg->GetAI()->AttackStart(Mrfloppy);
+                            }
                     }
                     break;
                 case 11:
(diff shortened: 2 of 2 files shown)


Stoabrogga:
I applied the changes and tested the quest, but it is not working correctly for me:
- The first worg, "Hungry Worg", just spawns, doing nothing. You can attack and kill it, though, but it does neither attack Mr. Floppy nor the player.
- The second worg, "Ravenous Worg", spawns and catches Mr. Floppy (animation correctly played), but otherwise also just stands there, doing nothing. You can attack and kill it, though, but it does not attack the player.

Tested on Ubuntu 16.04 with commit fffb925201894029748d5e53a6c78f2511858003.

A note for other testers: You have to remove the quest (```.quest remove 12027```) and then take it from Emily or otherwise the event is not starting.

Stoabrogga:
@poszer 
I applied the SQL statement. The worgs are now attacking, looks good :+1:

Stoabrogga:
I tested the quest a few more times and there seem to be a other issues:
- Although after applying "CombatAI" the worgs attack, the emote where the ravenous worg chews on Mr. Floppy does not seem to be played anymore: https://github.com/azerothcore/azerothcore-wotlk/blob/d03a1583c4a65218c0d2c85f3d284d068507098d/src/server/scripts/Northrend/zone_grizzly_hills.cpp#L187
- Grizzly bears seeing Emily suddenly stop moving and just watch her passing by.
- If Mr. Floppy is attacked by a grizzly bear Emily just continues running, which can lead to a broken chain of events.
- If you take too long to kill the ravenous worg Mr. Floppy just awakes and runs away together with Emily.

I really don't know what the CombatAI does or why it is necessary here (it seems to just not work without it). It also seems to me that this quest has many other problems. Sadly, I have no idea how to fix them.

BarbzYHOOL:
Hmm but it's better than before

Stoabrogga:
Yes, so I think this can be merged. The quest can always be improved through subsequent PRs.

Stoabrogga:
Ok, I've taken a closer look at this: The reason why the worgs are not attacking in the first place is that they have both their "VehicleId" set. The Hungry Worg does not need it, though, so to fix him, just set "VehicleId" to 0. The Ravenous Worg is a bit more complicated. As it is technically treated as a vehicle, it uses the VehicleAI, which is not really made for this purpose (it handles player vehicles). So switching to CombatAI is not wrong, but this AI prevents emote commands because the worg seems to enter combat when it is spawned, but is then pacified. The solution in this case would be a custom AI for the Ravenous Wolf.

@poszer 
Could you please update your SQL script as follows?
```
INSERT INTO version_db_world (`sql_rev`) VALUES ('1546128548989877900');

UPDATE `creature_template` SET `VehicleId` = 0 WHERE `entry` = 26586;
UPDATE `creature_template` SET `ScriptName` = 'npc_ravenous_worg' WHERE `entry` = 26590;
```

And please add this to "zone_grizzly_hills.cpp" after Mr. Floppy's AI in line 321:
https://github.com/azerothcore/azerothcore-wotlk/blob/d03a1583c4a65218c0d2c85f3d284d068507098d/src/server/scripts/Northrend/zone_grizzly_hills.cpp#L321
```
// Ravenous Worg
class npc_ravenous_worg : public CreatureScript
{
public:
    npc_ravenous_worg() : CreatureScript("npc_ravenous_worg") { }

    struct npc_ravenous_worgAI : public CombatAI
    {
        npc_ravenous_worgAI(Creature* creature) : CombatAI(creature)
        {
            _pacified = false;
            _attack = false;
        }

        void AttackStart(Unit* who)
        {
            if (me->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PACIFIED))
                _pacified = true;

            if (_pacified && !me->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PACIFIED))
                _attack = true;

            if (_attack)
                CombatAI::AttackStart(who);
        }

        void UpdateAI(uint32 diff)
        {
            if (me->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PACIFIED))
                _pacified = true;

            if (_pacified && !me->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_PACIFIED))
                _attack = true;

            CombatAI::UpdateAI(diff);
        }

        private:
            bool   _pacified;
            bool   _attack;
    };

    CreatureAI* GetAI(Creature* creature) const
    {
        return new npc_ravenous_worgAI(creature);
    }
};
```
It is based on the CombatAI but prevents starting an attack until the pacify-flag is cleared again, so the worg now correctly catches Mr. Floppy, chews on him, throws him away and attacks the player. I tested this on my test-server and it works fine.

poszer:
@Stoabrogga Is it okay now?

Stoabrogga:
Yes, but please add an empty line after the worg class definition for a better distinction from the next class:
https://github.com/azerothcore/azerothcore-wotlk/blob/9abd11339fc82f79796b3a211b221cebc93ae5dd/src/server/scripts/Northrend/zone_grizzly_hills.cpp#L368
It's easier to read this way.

BarbzYHOOL:
@poszer did you test the latest change though?

BarbzYHOOL:
cool

PR #1203, opened 2018-12-30
Core: Security against DOS

BarbzYHOOL:
##### CHANGES PROPOSED:

-  Prevents spamming auth server?

###### ISSUES ADDRESSED:

People reported that they got DOSed on their authserver and I found this thing on a forum and showed it to Rochet2.

##### TESTS PERFORMED:

None

As Rochet2 said
> I think all of the packets should be limited. Not sure by how much, that could depend on if the client could possibly send multiple of them (lets say someone spams a button 3 times, will he get disconnected?)
> I didnt check the other packets, but the 16 one has a DB query so it should at least be limited to some extent.

##### Target branch(es):

Master

### DIFF
--- a/src/server/authserver/Server/AuthSocket.cpp
+++ b/src/server/authserver/Server/AuthSocket.cpp
@@ -213,8 +213,8 @@ void AuthSocket::OnRead()
     {
         if (!socket().recv_soft((char *)&_cmd, 1))
             return;
-
-        if (_cmd == AUTH_LOGON_CHALLENGE)
+        
+        if (_cmd == AUTH_LOGON_CHALLENGE || _cmd == REALM_LIST) // realmlist packet
         {
             ++challengesInARow;
             if (challengesInARow == MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW)
--- a/src/server/authserver/Server/AuthSocket.cpp
+++ b/src/server/authserver/Server/AuthSocket.cpp
@@ -213,7 +213,7 @@ void AuthSocket::OnRead()
     {
         if (!socket().recv_soft((char *)&_cmd, 1))
             return;
-        
+
         if (_cmd == AUTH_LOGON_CHALLENGE || _cmd == REALM_LIST) // realmlist packet
         {
             ++challengesInARow;


Xgap:
I've found these things. Maybe the benefit to our work.

1.https://github.com/mmlTools/Old-Trinitycore-Revisions-Authentication-bypass-fix/commit/399252a064b5083bceb09071f0ea2ab5b47a22f4

2.written in any forum
```
// Read the packet from the client
void AuthSocket::OnRead()
{
    #define MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW 3
    uint32 challengesInARow = 0;
    uint8 _cmd;
    while (1)
    {
        if (!socket().recv_soft((char *)&_cmd, 1))
            return;
        
        if (_cmd == 16) // realmlist packet
        {
            ++challengesInARow;
            if (challengesInARow == MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW)
            {
                sLog->outString("Got %u AUTH_LOGON_CHALLENGE in a row from '%s', possible ongoing DoS", challengesInARow, socket().getRemoteAddress().c_str());
                socket().shutdown();
                return;
	    }
        }
	
        if (_cmd == AUTH_LOGON_CHALLENGE)
        {
            ++challengesInARow;
            if (challengesInARow == MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW)
            {
                sLog->outString("Got %u AUTH_LOGON_CHALLENGE in a row from '%s', possible ongoing DoS", challengesInARow, socket().getRemoteAddress().c_str());
                socket().shutdown();
                return;
            }
        }
```

sucofog:
The solution by @Xgap looks cleaner and good

FrancescoBorzi:
Please then open a separate PR so it can be tested easily

BarbzYHOOL:
Thank you @Xgap 

Your 2) is the same as mine

The 1) is a bit different as it differentiate each packet with a specific setting but I'm not able to judge if it's better or if it's useless.

Also it could be a switch instead of a duplicate "if"

Viste:
https://github.com/azerothcore/azerothcore-wotlk/pull/1207

vhiperdev:
@FrancescoBorzi @BarbzYHOOL @Nefertumm 

I think this PR can already be closed because of this https://github.com/azerothcore/azerothcore-wotlk/pull/2789

Helias:
This was a good PR, I don't know why we didn't test and merge, this block AuthFlooder, indeed I already implemented a check for REALM_LIST opcode here https://github.com/azerothcore/azerothcore-wotlk/commit/ce30c08fbcf3cbe527a1811890ce1341ca47bddb#diff-554241adfef7060889afc651707b31fb

So we can close this, but maybe, next time for this PR about security, we should pay more attention.

BarbzYHOOL:
i didn't even remember i pushed this :(

PR #1206, opened 2018-12-31
Core/Security:check for fake packets&bad addons that cause client crash

Viste:
##### CHANGES PROPOSED:

- add check for fake packets and disaster addons what make client crash

Closes: no issue for this 

to test  try crash client with some addon or fake packet 

##### Target branch(es):Master


src/server/game/Handlers/ChannelHandler.cpp
```diff
@@ -56,6 +57,10 @@ void WorldSession::HandleLeaveChannel(WorldPacket& recvPacket)
     std::string channelName;
     recvPacket >> unk >> channelName;
 
+    //check for fake packets and bad addons that cause client to crash
+    if (!ChatHandler(this).isValidChatMessage(channelName.c_str()))
+        return;
```
BarbzYHOOL: Why not make a small function for this? to avoid duplicate code

Viste: lazzy and its old good code )

Viste: oh ok :)

Viste:
ChatFake and etc

Viste:
https://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-exploits/376771-exploit-make-other-players-crash.html

Viste:
ant other

BarbzYHOOL:
Awesome PR though. Do you have a commit reference (TC, Mangos) ?

Viste:
> Awesome PR though. Do you have a commit reference (TC, Mangos) ?
no its old patch from trinity-core.ru :)

BarbzYHOOL:
Ok, even better

wetbrownsauce:
I can't seem to replicate the crash with any of those ownedcore links so I'm not sure how to test :S

BarbzYHOOL:
Isn't it related to this worldserver conf `ChatFakeMessagePreventing = 1`

Viste:
@wetbrownsauce i don't remember name of all addons(only ChatFake) what have all client crash funcs >_< just google first what can find )

src/server/game/Handlers/ChannelHandler.cpp
FrancescoBorzi: @Viste can you please update this PR with this suggestion?

Viste: yes after on next week

PR #1207, opened 2018-12-31
Core/WorldSession: Implement AntiDOS system.

Viste:
##### CHANGES PROPOSED:
Checking packets and suspicious amounts of opcodes received server side, kicking player when flooding server with opcodes.
Closes: No issue just some reports


src/server/game/Server/WorldSession.h
```diff
@@ -933,6 +939,36 @@ class WorldSession
...
+
+            private:
+                Policy _policy;
+                typedef std::unordered_map<uint16, PacketCounter> PacketThrottlingMap;
+                // mark this member as "mutable" so it can be modified even in const functions
+                mutable PacketThrottlingMap _PacketThrottlingMap;
+
+                DosProtection(DosProtection const& right) = delete;
+                DosProtection& operator=(DosProtection const& right) = delete;
+} AntiDOS;
```
Nefertumm: Tabs, also you need any WorldSession in order to declare AntiDOS,

Viste: yeah i know all done

src/server/game/Server/WorldSession.cpp
```diff
@@ -1335,3 +1336,296 @@ void WorldSession::InitWarden(BigNumber* k, std::string const& os)
...
+        }
+        case POLICY_BAN:
+        {
+            BanMode bm = (BanMode)sWorld->getIntConfig(CONFIG_PACKET_SPOOF_BANMODE);
+            uint32 duration = sWorld->getIntConfig(CONFIG_PACKET_SPOOF_BANDURATION); // in seconds
+            std::string nameOrIp = "";
+            switch (bm)
+            {
+                case BAN_CHARACTER: // not supported, ban account
+                case BAN_ACCOUNT: (void)sAccountMgr->GetName(Session->GetAccountId(), nameOrIp); break;
```
sucofog: ``case BAN_ACCOUNT: (void)sAccountMgr->GetName(Session->GetAccountId(), nameOrIp); break;``
change to
``case BAN_ACCOUNT: (void)AccountMgr::GetName(Session->GetAccountId(), nameOrIp); break;``

FrancescoBorzi:
@Viste I would like to merge this into `staging` and start testing it. But before that one last question:

can this have a negative impact on performance? I wonder since it's adding extra operations for every received packet. If that's the case, wouldn't it better to have this configurable (and disabled by default) ?

Just wondering

Viste:
@FrancescoBorzi Yes it can affect to performance because all actions log(in basic) or log+kick or log +kick+ban 
and i agree need to add a conf option for it

pklloveyou:
In the trial

BarbzYHOOL:
@pklloveyou how do you test this?

And yes, if it's fine, then need to add a config option and disable it by default. Could as well been disabled with cmake EXTRAS flag

pklloveyou:
tim 20190302154424 
@BarbzYHOOL  : (  Because I'm under attack！

Viste:
@pklloveyou how you configure your conf file

boom8866:
This is a reject packet prompt, the port received a false data connection, rejected the request,But I feel it has nothing to do with this configuration,1. Because player data is sent via fake packets. 2. Packet port passes without the player,Does little

pklloveyou:
@Viste @boom8866 This should be a CC attack

Issue #1187, opened 2018-12-30
DB restructuring - world - game_event


FrancescoBorzi:
This is part of: https://github.com/azerothcore/azerothcore-wotlk/issues/1118

**General note**: not all structure differences can or should be imported. Every change must be first verified. If a structure change is not suitable, we just close the issue.

---

About the field `announce`, if this is a missing feature we could just create it and mark it as UNUSED in our documentation until we implement it. Investigation is needed here.

The other 2 fields can be adapted. This also helps with MySQL 5.7 support because of removing `DEFAULT '0000-00-00 00:00:00'`.

```diff
--- Dumping structure for table acore_world.game_event
+-- Dumping structure for table tc_world.game_event
 DROP TABLE IF EXISTS `game_event`;
 CREATE TABLE IF NOT EXISTS `game_event` (
   `eventEntry` tinyint(3) unsigned NOT NULL COMMENT 'Entry of the game event',
-  `start_time` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00' COMMENT 'Absolute start date, the event will never start before',
-  `end_time` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00' COMMENT 'Absolute end date, the event will never start afler',
+  `start_time` timestamp NULL DEFAULT NULL COMMENT 'Absolute start date, the event will never start before',
+  `end_time` timestamp NULL DEFAULT NULL COMMENT 'Absolute end date, the event will never start after',
   `occurence` bigint(20) unsigned NOT NULL DEFAULT '5184000' COMMENT 'Delay in minutes between occurences of the event',
   `length` bigint(20) unsigned NOT NULL DEFAULT '2592000' COMMENT 'Length in minutes of the event',
   `holiday` mediumint(8) unsigned NOT NULL DEFAULT '0' COMMENT 'Client side holiday id',
   `holidayStage` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `description` varchar(255) DEFAULT NULL COMMENT 'Description of the event displayed in console',
   `world_event` tinyint(3) unsigned NOT NULL DEFAULT '0' COMMENT '0 if normal event, 1 if world event',
+  `announce` tinyint(3) unsigned DEFAULT '2' COMMENT '0 dont announce, 1 announce, 2 value from config',
   PRIMARY KEY (`eventEntry`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```

[...]

Issue #1118, opened 2018-12-17
[EPIC] Align DB structure to TrinityCore


FrancescoBorzi:
_Note: I'm opening this in order to finish the work previously started here_: https://github.com/azerothcore/azerothcore-wotlk/issues/236 

This is an **epic** task that groups all the tasks needed to upgrade the **structure of the AzerothCore Databases** in order to make it  as close as possible to the one of [TrinityCore 3.3.5](https://github.com/TrinityCore/TrinityCore/tree/3.3.5).

## Benefits

- **Documentation**. Currently we have no official DB documentation. We could just copy and reuse the same [DB documentation from tc](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/74663833/World+3.3.5) once this task is completed (this is already the case for many tables, but for some of them the structure is different and it's undocumented).

- **Cross-project collaboration**. It will be easier to import any DB-content fixes that are pushed into the 3.3.5 branch of TrinityCore. Having a look at the [commit history](https://github.com/TrinityCore/TrinityCore/commits/3.3.5) of that branch, you can notice that there are still several contributors that push DB fixes periodically.

- **Tooling**. AzerothCore would benefit of all existing tooling currently available for TC 3.3.5. For example the [JSON API](https://github.com/FrancescoBorzi/TC-JSON-API) and all the tools that rely on it (such as [Keira2 DB editor](https://github.com/Helias/Keira2) and many other tools like [those](https://github.com/FrancescoBorzi?tab=repositories&q=trinitycore)).

- **Migration**. It would be much easier for server administrators to migrate from TrinityCore to AzerothCore.

- **Simplicity**. The new structure simplifies many things, for example the huge `quest_template` table has being split into several smaller ones.

_**Personal note**: I believe a big part of this job has been already done, and it's worth to finish it. Also it looks like that there is no intention in TrinityCore to change the 3.3.5 DB structure anymore. So once we finish this, we can only benefit_

### World DB

- [x] https://github.com/azerothcore/azerothcore-wotlk/issues/1119 Simplify the `quest_template` table
- [x] https://github.com/azerothcore/azerothcore-wotlk/issues/1120 Update structure of the `*_loot_template` tables
- [x] https://github.com/azerothcore/azerothcore-wotlk/issues/1178 Misc minor changes
- [x] https://github.com/azerothcore/azerothcore-wotlk/issues/1179 Locale changes
[...]

PR #1228, opened 2019-01-03
Core/Events: announce event system improvement.

Viste:
##### CHANGES PROPOSED:
-  Remove ugly hack with switch and cases. make selection through DB
###### ISSUES ADDRESSED:
Closes : part of #1187 and #1118

How to test:
if in DB in announce colum we have 0 dont announce the announce about event shouldnt start
if we have 1 announce should be even if in config he off
and if we have 2 value be gets from config


src/server/game/Events/GameEventMgr.cpp
```diff
@@ -1189,14 +1190,9 @@ void GameEventMgr::UnApplyEvent(uint16 event_id)
...
-    switch (sWorld->getIntConfig(CONFIG_EVENT_ANNOUNCE))
-    {
-        case 0:                                             // disable
-            break;
-        case 1:                                             // announce events
-            sWorld->SendWorldText(LANG_EVENTMESSAGE, mGameEvent[event_id].description.c_str());
-            break;
-    }
+    uint8 announce = mGameEvent[event_id].announce;
+    if (announce == 1 || (announce == 2 && sWorld->getBoolConfig(CONFIG_EVENT_ANNOUNCE)))
```
Nefertumm:
```suggestion
     if (announce == 1 || (announce == 2 && sWorld->getIntConfig(CONFIG_EVENT_ANNOUNCE)))
```

Viste: :(

BarbzYHOOL:
That looks very useful. Any example with an existing script though?

Viste:
@BarbzYHOOL this is not related to game, only make struct of DB same as trinity and this is more clean when it do throug DB then as now by switch

data/sql/updates/pending_db_world/rev_1546538959760274673.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO version_db_world (`sql_rev`) VALUES ('1546538959760274673');
+
+ALTER TABLE `game_event`
+ADD COLUMN `announce` tinyint(3) unsigned NULL DEFAULT 2 COMMENT '0 dont announce, 1 announce, 2 value from config' AFTER `world_event`;
```
FrancescoBorzi: what does the `NULL` after `unsiged` do?

Viste: i miss not null >_<

BarbzYHOOL:
> @BarbzYHOOL this is not related to game, only make struct of DB same as trinity and this is more clean when it do throug DB then as now by switch

I know... but an example of how it is used in the CODE would be kind of nice but if it's ported from TC then ok fine...

Viste:
> > @BarbzYHOOL this is not related to game, only make struct of DB same as trinity and this is more clean when it do throug DB then as now by switch
> 
> I know... but an example of how it is used in the CODE would be kind of nice but if it's ported from TC then ok fine...

it used only as — 0 dont announce, 1 announce, 2 value from config, and only here. Its need for core can understand need send announce when event start or not.

Viste:
How to test:
if in DB in announce colum we have 0 dont announce the announce about event shouldnt start 
if we have 1  announce should be even if in config he off
and if we have  2 value be gets from config

wetbrownsauce:
Tested. 

- Set some game_events to occur every minute for easy testing.
- When announce = 2, events only announce in-game if config option is on.
- When announce = 1, events announce in-game even if config option is off.
- When announce = 0, events never announce in-game.

BarbzYHOOL:
thank you guys

PR #1230, opened 2019-01-03
MySQL: account table restructure

Viste:


data/sql/updates/pending_db_auth/rev_1546541118941824046.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO version_db_auth (`sql_rev`) VALUES ('1546541118941824046');
+
+ALTER TABLE `account` CHANGE `email` `email` VARCHAR(255) NOT NULL DEFAULT '';
```
FrancescoBorzi: missing new line(s) at the end of file(s)

FrancescoBorzi: also you can put those changes in one single sql file since it's the same database

Viste: i separate them to be more clean

BarbzYHOOL: put them in the same file please, it's more logic for our case (and with the "insert into revision" )

data/sql/updates/pending_db_auth/rev_1546540718864817294.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO version_db_auth (`sql_rev`) VALUES ('1546540718864817294');
+
+ALTER TABLE `account` CHANGE `last_login` `last_login` TIMESTAMP NULL;
```
Meltie2013: this should be like this
```alter table account change last_login last_login timestamp null default null```
it's a more clean way of doing it.

wetbrownsauce:
Not seeing any issues from this in game so no complaints from me. I agree with the proposed change by meltie though.

data/sql/updates/pending_db_auth/rev_1546540718864817294.sql
Viste: lol you make colum null by default i'm not. no clean or dirty

Meltie2013: Doing it the way you have it in this PR, will just repeat the issue you're trying to close. So your PR is just useless without a proper method of fixing it.

BarbzYHOOL: Yes, default should be Null I guess

FrancescoBorzi: `last_login` timestamp NULL DEFAULT NULL,`

Issue #1222, opened 2019-01-02
[QUEST] Ulag the cleaver


Maurowin:
##### SMALL DESCRIPTION:

After clicking on the rune, we must wait until the seal is lifted. However, you do not see a seal and you have to click on the door manually

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Like video
https://www.youtube.com/watch?v=TssjgNMJuPk&t=10s
![2019-01-02 1](https://user-images.githubusercontent.com/31663279/50609814-1d0f3380-0ed1-11e9-9a23-622cab9e2dfa.png)

##### CURRENT BEHAVIOUR:

The quest is currently as in the video below
https://www.youtube.com/watch?v=xrUCOwy3S8s

##### STEPS TO REPRODUCE THE PROBLEM:

1. learn quest id 1819
2. clic on the rune
3. see

##### BRANCH(ES):

master

grimgravy:
[TrinityCore Fixation](http://dev.qyh.name:800/snet-dev-group/TrinityCore/commit/029c4232a7f53b32cd233502eb7078e64f37e87e)
[SQL AC](https://pastebin.com/Fhx2PUFk)

FrancescoBorzi:
@grimgravy thanks! It would be nice if you could open a PR :)

poszer:
Tested in game with fix included. It's okay ( Players can now click on that ''Mausoleum Trigger'' and the doors are opening. The NPC Ulag the cleaver coming outside on Horse and attacking player)  - Just like on the Video. Will create PR for it. Thanks @grimgravy

pklloveyou:
Why not C++scripts

pklloveyou:
`/*######
## go_mausoleum_trigger
######*/

enum Mausoleum
{
    QUEST_ULAG      = 1819,
    NPC_ULAG        = 6390,
    GO_DOOR         = 176594
};

class npc_ulag_the_cleaver : public CreatureScript
{
public:
	npc_ulag_the_cleaver() : CreatureScript("npc_ulag_the_cleaver") { }

	CreatureAI* GetAI(Creature* creature) const
	{
		return new npc_ulag_the_cleaverAI(creature);
	}

	struct npc_ulag_the_cleaverAI : public ScriptedAI
	{
		npc_ulag_the_cleaverAI(Creature* creature) : ScriptedAI(creature) { }

		void Reset()
		{
			me->CastSpell(me, 23246, true);
[...]

PR #845, opened 2018-05-04
Fix overlapping mailbox in undercity (Trade Quarter, south)

Stoabrogga:
**Changes proposed:**
There are two overlapping mailboxes in Undercity, south Trade Quarter. Delete one of them.

**Target branch(es):** master

**Tests performed:** tested build incl. DB update, tested in-game

**Known issues and TODO list:** none

**NOTE** You no longer need to squash your commits, on merge we will squash it for you. (GitHub added a new feature)

PR #1245, opened 2019-01-08
DB/Quest Ulag the Cleaver

poszer:
[Quest Ulag the Cleaver](https://vanilla-twinhead.twinstar.cz/?quest=1819)

##### CHANGES PROPOSED:

-  Move Ulag the Cleaver to SAI
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1222

##### TESTS PERFORMED:

Tested in game with update. It's okay ( Players can now click on that ''Mausoleum Trigger'' and the doors are opening. The NPC "Ulag the cleaver" coming outside on Horse and attacking player) which is correct.

1. .quest add 1819
2. .go xyz 2372.087646 363.687805 38.098766 0
3. Click on item in front of you '' Mausoleum Trigger ''
4. Take a look on the doors. 
5. They should opening and the '' Ulag the cleaver '' will come on horse attacking you.

- [x] Waypoint of the NPC is correct now
- [x] SAI is correct now

##### Target branch(es):

Master

Thanks @grimgravy


grimgravy:
Do not forget to remove the [script ](https://github.com/azerothcore/azerothcore-wotlk/blob/2f5653c24fe0ca031c61cd1d3db6b7b400d6518d/src/server/scripts/EasternKingdoms/zone_tirisfal_glades.cpp#L147)

Stoabrogga:
@poszer 
I think you can change the movement yourself. This should be the correct line:
https://github.com/azerothcore/azerothcore-wotlk/blob/cfef52d19049e61c4a6c39471b8c356dac097bf2/data/sql/updates/pending_db_world/rev_1546933586341240000.sql#L25

```2371.5554, 358.37, 38.092,0```

These values describe position x,y,z and orientation, just use another position and orientation here.

Stoabrogga:
Tested it and the quest looks good, but was the seal always visible? I played a lot WotLK, and Forsaken are my favorites, but I don't recognize the seal on the crypt. Also found two old videos were the seal is not visible:
https://www.youtube.com/watch?v=U_5Et5tHdSg&t=395
https://www.youtube.com/watch?v=7gFnNU3wZdc&t=9

Sadly, I never played a Forsaken warrior, so cannot tell how the quest originally did work.

poszer:
> pt. Also found two old videos were th

Hmm, i saw that videos too. There are more versions, on one effect is always visible, on other one it's visible just when used item ( So, i don't know how it is correct, for example on this video is just like on our update, check it please [Ulag the Cleaver : Vanilla Quest](https://www.youtube.com/watch?v=TssjgNMJuPk)

Stoabrogga:
Oh, and another important note: You are deleting GUID 1532 from the gameobject table. This object was linked with template ID 2066 'Bonfire Damage'. I am not comfortable with deleting old entries, in my opinion it would be better to use a new GUID here.

Stoabrogga:
@poszer 
Could you please use GUID 195629 instead of 1532? 195629 is definitely not used as it was removed in PR #845 (was a duplicate mail box in Undercity). This would avoid deleting existing entries.

As for the seal: I think it is ok if the seal is always visible, even if it's not blizzlike. Actually I find it rather interesting, because it shows that something is caged there and the crypt is special.

poszer:
> @poszer
> Could you please use GUID 195629 instead of 1532? 195629 is definitely not used as it was removed in PR #845 (was a duplicate mail box in Undercity). This would avoid deleting existing entries.
> 
> As for the seal: I think it is ok if the seal is always visible, even if it's not blizzlike. Actually I find it rather interesting, because it shows that something is caged there and the crypt is special.

Sure, i'll do it. Thanks

FrancescoBorzi:
good job @poszer and @Stoabrogga

FrancescoBorzi:
@poszer I've updated your branch with master and restarted the build. Now it should pass

grimgravy:
> @poszer
> Could you please use GUID 195629 instead of 1532? 195629 is definitely not used as it was removed in PR #845 (was a duplicate mail box in Undercity). This would avoid deleting existing entries.
> 
> As for the seal: I think it is ok if the seal is always visible, even if it's not blizzlike. Actually I find it rather interesting, because it shows that something is caged there and the crypt is special.

Doubt  [WoWHead](https://www.wowhead.com/quest=1819/ulag-the-cleaver#screenshots)

Issue #1250, opened 2019-01-09
Incorrect encoding in DB after use db_assembler


Winfidonarleyan:
##### SMALL DESCRIPTION: After using db_assembler in DB incorrect data.
- `locales_*`
- `creature_template_locale`

I use clear db and choose `import-all: Assemble & Import all`
```sh
$ bash ./db_assembler.sh
1) all: Assemble all
2) bases: Assemble only bases
3) updates: Assemble only updates
4) customs: Assemble only customs
5) import-all: Assemble & Import all
6) import-bases: Assemble & Import only bases
7) import-updates: Assemble & Import only updates
8) import-customs:  Assemble & Import only customs
9) quit: Exit from this menu
[Please enter your choice]: 5
=====     DB ASSEMBLER MENU     =====
===== STARTING ASSEMBLY PROCESS =====
Generating /d/Git/azerothcore-wotlk/env/dist/sql/auth_base ...
Searching on /d/Git/azerothcore-wotlk/data/sql/base/db_auth/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/auth_update ...
Searching on /d/Git/azerothcore-wotlk/data/sql/updates/db_auth/ ...
Searching on /d/Git/azerothcore-wotlk/data/sql/updates/pending_db_auth/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/auth_custom ...
Searching on /d/Git/azerothcore-wotlk/data/sql/custom/db_auth/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/characters_base ...
Searching on /d/Git/azerothcore-wotlk/data/sql/base/db_characters ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/characters_update ...
Searching on /d/Git/azerothcore-wotlk/data/sql/updates/db_characters/ ...
Searching on /d/Git/azerothcore-wotlk/data/sql/updates/pending_db_characters/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/characters_custom ...
Searching on /d/Git/azerothcore-wotlk/data/sql/custom/db_characters/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/world_base ...
Searching on /d/Git/azerothcore-wotlk/data/sql/base/db_world/ ...
Generating /d/Git/azerothcore-wotlk/env/dist/sql/world_update ...
[...]

PR #1252, opened 2019-01-09
Bash: Fix incorrect apply sql files at import

Winfidonarleyan:
##### CHANGES PROPOSED:
-  Added option `--default-character-set=utf8`
-  Fix incorrect apply not English text

##### ISSUES ADDRESSED:
Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1250

##### TESTS PERFORMED: 
- `import-all: Assemble & Import all` in Windows

**Need full clean DB!**
1. Apply PR
2. Use `db_assembler`
3. Choose `5 - import-all: Assemble & Import all`
4. Check DB table e.g. `locales_broadcast_text`
5. First data should be as follows
image

- [x] in Windows
- [ ] in Linux

### DIFF
--- a/apps/db_assembler/includes/functions.sh
+++ b/apps/db_assembler/includes/functions.sh
@@ -299,7 +299,7 @@ function dbasm_db_import() {
 
     export MYSQL_PWD=$MYSQL_PASS
     
-	"$DB_MYSQL_EXEC" -h "$MYSQL_HOST" -u "$MYSQL_USER" "$dbname" < "${OUTPUT_FOLDER}${database}_${type}.sql"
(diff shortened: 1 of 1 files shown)


Winfidonarleyan:
I think this is for the correct reading of the file

BarbzYHOOL:
but stoabrogga is right, there seems to be an issue that can bite us back in the future

BarbzYHOOL:
setting specified in file is not applied
maybe other settings won't be applied

FrancescoBorzi:
Not sure. We need to do some search about it

BarbzYHOOL:
then this can be merged meanwhile so the other PR can be merged too (they are related)

but if we change something, let's not forget this "hackfix" :D

PR #1272, opened 2019-01-12
Fixed permission in docker-database

Helias:
Fixed permission in docker-database, added command chmod 755 before run the file db_assembler.sh in the Dockerfile for the database.


FrancescoBorzi:
thanks @Helias

PR #1279, opened 2019-01-13
DB/Locales: increase max_allowed_packet before running heavy queries

FrancescoBorzi:


FrancescoBorzi:
@BarbzYHOOL I don't really understand what is going there, but this PR should work already. It just would be nice to hear from someone that actually HAS the issue (because on my machine I don't have it so it will work anyway).

Travis CI had this issue, maybe I can revert it there to reproduce and see if it still works.

wetbrownsauce:
I'm currently testing some PR's, so might be a bit on this one, but I can reliably replicate the MySQL has gone away error, so that'll be a start

FrancescoBorzi:
Done. If the build passes now, then this is safe to be merged.

FrancescoBorzi:
@wetbrownsauce I guess checking the travis build would be ok since we had exacly the same issue there. You can focus on other PRs meanwhile :)

BarbzYHOOL:
It's not a good fix imo, because it changes the global setting of the user, without reverting it.

Btw the file I linked is for `max_connection_packet` so it's not exactly the same issue I guess? (also I compared the file and found nothing really useful for us)

However it's 40mb heavy, here it's 19mb

FrancescoBorzi:
@BarbzYHOOL I will check this later

Winfidonarleyan:
i use `--max-allowed-packet=1GB` from TC DBUpdater
https://github.com/TrinityCore/TrinityCore/blob/d609e515d7f5ab2f250fec409d9fb690f301b906/src/server/database/Updater/DBUpdater.cpp#L370

Issue #1105, opened 2018-12-09
Items quests loot party


pangolp:
##### SMALL DESCRIPTION:
The quest items when we are in party with a companion, remain in the npc and can only be looteados by 1 of the members of the group.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
It is hoped that they can be looteados, apparently there is a flag in the items that is the 2048. The problem is that they have to be filtered and studied item by item because they are not all loot party.

##### CURRENT BEHAVIOUR:
Example: https://www.wowhead.com/item=1129/ghoul-fang

##### STEPS TO REPRODUCE THE PROBLEM:

1. Kill an npc with a partner while partying.
2. Try to lootear both the item.

##### BRANCH(ES):

master

poszer:
Just tested changes with 2 characters. It should be like this, right @pangolp  ?

![ovak](https://user-images.githubusercontent.com/41213210/49702708-acd7ff80-fbfb-11e8-9108-ca51db8dfbdb.png)

pangolp:
That's right @poszer . There are items that only let one of the group grab, the other does not let him strip.

poszer:
> You can check it, with the missions of the starting levels, like the one of the dranei. There we find the first error and then investigate and is problem of the flag.

Did you try to change flag? Doesn't work?

pangolp:
Yes, if you change it works, but it's 5000 mission items and not all of them are to be grabbed in a group.

example:

```sql
SET @ITEM:=22889;
UPDATE `item_template` SET `Flags`=`Flags`|2048 WHERE `entry`=@ITEM;
```

poszer:
> Yes, if you change it works, but it's 5000 mission items and not all of them are to be grabbed in a group.
> 
> example:
> 
> ```sql
> SET @ITEM:=22889;
> UPDATE `item_template` SET `Flags`=`Flags`|2048 WHERE `entry`=@ITEM;
> ```

Yeah, i understand. Do this, and tell me is your issue fixed with it 
[da845a4078f7a208db8611049a2a73855914482b](https://github.com/poszer/azerothcore-wotlk/commit/da845a4078f7a208db8611049a2a73855914482b)

pangolp:
No, the mistake goes on. Putting the flag works, but the problem is that you have to check item by item.
[...]

PR #1299, opened 2019-01-16
DB/Loot: Fix items quests loot party

pangolp:
###### ISSUES ADDRESSED:

Closes 
https://github.com/azerothcore/azerothcore-wotlk/issues/1105

1. Get in group with friend or login with 2 accounts
2. Kill an npc with a partner while in group
3. Try to loot with both characters item.

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
+++ b/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
@@ -1,2 +1,2 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1547660829459489600');
-UPDATE `item_template` SET `Flags` = `Flags` | 2048 WHERE `class` = 12 AND `subclass` = 0 AND `bonding` = 4;
+UPDATE `item_template` SET `Flags`=`Flags`|2048 WHERE `class`=12 AND `subclass`=0 AND `bonding`=4;
--- a/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
+++ b/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
@@ -1,0 +1,2 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1547660829459489600');
+UPDATE `item_template` SET `Flags` = `Flags` | 2048 WHERE `class` = 12 AND `subclass` = 0 AND `bonding` = 4;
--- a/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
+++ b/data/sql/updates/pending_db_world/rev_1547660829459489600.sql
@@ -1,2 +1,3 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1547660829459489600');
+
 UPDATE `item_template` SET `Flags`=`Flags`|2048 WHERE `class`=12 AND `subclass`=0 AND `bonding`=4;


poszer:
> Do not fill in the data because they are all in the issue, it would be to repeat everything again. But if I have to fill it again, I will fill it.



> Do not fill in the data because they are all in the issue, it would be to repeat everything again. But if I have to fill it again, I will fill it.

Nvm, i edited this time. If you didn't notice, Issue template and PR template are different. For example, while opening issue you re writing **STEPS TO REPRODUCE THE PROBLEM**. But while opening PR, you re writing **HOW TO TEST _THE CHANGES_** - That's not same.

Maurowin:
there are quest items that can be picked up by all players, such as, for example, the head of a monster etc... however, all quest items to be picked up like skins or others cannot be picked up by all players on the same beast.

Winfidonarleyan:
I added a new line after first line

pangolp:
Would the arrangement be all right then?

Stoabrogga:
I'd advise against merging this PR. It updates over 3000 item templates without the possibility to revert the changes.

pangolp:
Ok. We'll have to keep looking for another solution then why at least we were bugged with items.  Maybe you didn't.

Stoabrogga:
@pangolp 
We'll find a solution. But this one here is too risky for my taste. If it turns out later that the change has unforeseen side effects you cannot easily revert it.

BarbzYHOOL:
Yes, me too. That seems too straightforward to be the right solution :D. Looks like a custom feature

wetbrownsauce:
I use this as a custom hack-fix on my server, and it has worked perfectly for the past month. However, I do not think it should be merged into the base branch. It is likely NOT the blizzlike way of doing this, as TC does not have the 2048 flag set on hardly anything in DB. I think a core-fix would be more appropriate.  

I have yet to see it affect something that it shouldn't, but just because it works, doesn't mean it is right.

FrancescoBorzi:
I agree with @wetbrownsauce , we could actually keep all those "hackfixes" somewhere. For example a repository?

Anyway, thank you very much @pangolp for contributing, we appreciate.

BarbzYHOOL:
Good idea, I made a repository for this: https://github.com/azerothcore/azerothcore-hackfixes

wetbrownsauce:
Hmmm, yeah maybe a separate repo of work-arounds and hackfixes with ways to reverse them included. Maybe even a wiki page explaining how to use/remove them.  

My bitwise math is a bit shaky so correct me if I'm wrong, but the following query would return every item that _would_ be affected by this PR before it is applied, no?
```
SELECT `entry`, `name` FROM `item_template`
WHERE (`Flags` & 2048 = 2048) AND (`class` = 12) AND (`bonding` = 4);
```

wetbrownsauce:
Also, as a side note for this specific PR, why is `subclass = 0` being checked? All items in the world DB with `class = 12` and `bonding = 4` have `subclass = 0` so it seems useless to include it.

Stoabrogga:
> My bitwise math is a bit shaky so correct me if I'm wrong, but the following query would return every item that _would_ be affected by this PR before it is applied, no?
> 
> ```
> SELECT `entry`, `name` FROM `item_template`
> WHERE (`Flags` & 2048 != 2048) AND (`class` = 12) AND (`bonding` = 4);
> ```
> 
> edit: left out an !

Yes, but what are you trying to accomplish? Dump all entries before applying the update statement?

wetbrownsauce:
Partly. 

I would like to dump all entries of items that would be affected so that I can:
1) easily revert.
2) go through the list and see if anything shouldn't be there.
3) get more practice with bitwise logic

:P

BarbzYHOOL:
post it on the new repo wetbrown or pangolp 

Even a simple text ".md" file could do the job

pangolp:
You're welcome. I'm just trying to collaborate, I'll find some other way to keep doing it. Thank you for taking the time to analyze it.

pangolp:
After many years, I was finally able to create a hook and with it, add a module that does this job.
https://github.com/pangolp/mod-quest-loot-party

Issue #1178, opened 2018-12-30
DB restructuring - world - minor changes


FrancescoBorzi:
This is part of: https://github.com/azerothcore/azerothcore-wotlk/issues/1118

**General note**: not all structure differences can or should be imported. Every change must be first verified. If a structure change is not suitable, we just close the issue.

---

A collection of minor changes which should be easy to do. They can be split in different issues if needed.

```diff
--- Dumping structure for table acore_world.creature_equip_template
+-- Dumping structure for table tc_world.creature_equip_template
 DROP TABLE IF EXISTS `creature_equip_template`;
 CREATE TABLE IF NOT EXISTS `creature_equip_template` (
   `CreatureID` mediumint(8) unsigned NOT NULL DEFAULT '0',
@@ -283,12 +294,12 @@ CREATE TABLE IF NOT EXISTS `creature_equip_template` (
   `ItemID1` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ItemID2` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ItemID3` mediumint(8) unsigned NOT NULL DEFAULT '0',
-  `VerifiedBuild` smallint(6) NOT NULL DEFAULT '1',
+  `VerifiedBuild` smallint(5) DEFAULT '0',
   PRIMARY KEY (`CreatureID`,`ID`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;

--- Dumping structure for table acore_world.creature_questitem
+-- Dumping structure for table tc_world.creature_questitem
 DROP TABLE IF EXISTS `creature_questitem`;
 CREATE TABLE IF NOT EXISTS `creature_questitem` (
   `CreatureEntry` int(10) unsigned NOT NULL DEFAULT '0',
   `Idx` int(10) unsigned NOT NULL DEFAULT '0',
   `ItemId` int(10) unsigned NOT NULL DEFAULT '0',
+  `VerifiedBuild` smallint(5) NOT NULL DEFAULT '0',
   PRIMARY KEY (`CreatureEntry`,`Idx`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;

--- Dumping structure for table acore_world.gameobject_questitem
+-- Dumping structure for table tc_world.gameobject_questitem
 DROP TABLE IF EXISTS `gameobject_questitem`;
 CREATE TABLE IF NOT EXISTS `gameobject_questitem` (
[...]

PR #1323, opened 2019-01-19
refactor(DB-world): minor changes

FrancescoBorzi:
##### CHANGES PROPOSED:

- Remove minor differences to sync with TC

###### ISSUES ADDRESSED:
Closes  https://github.com/azerothcore/azerothcore-wotlk/issues/1178

##### TESTS PERFORMED:

Compile, runs

The only affected fields that are actually used by the core are:
```
- `quest_template_addon`.`NextQuestID`
- `spell_bonus_data`.`entry`
- `waypoint_data`.`wpguid`
```

the test should just make sure that the data stored in those fields is not altered after this PR.

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1547906915621116396.sql
+++ b/data/sql/updates/pending_db_world/rev_1547906915621116396.sql
@@ -1,0 +1,52 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1547906915621116396');
+
+
+-- ALTER TABLE CHANGE COLUMN
+
+ALTER TABLE `creature_equip_template`
+	CHANGE COLUMN `VerifiedBuild` `VerifiedBuild` smallint(5) DEFAULT '0';
+
+ALTER TABLE `item_set_names`
+  CHANGE COLUMN `VerifiedBuild` `VerifiedBuild` smallint(5) DEFAULT '0';
+
+ALTER TABLE `item_template`
+  CHANGE COLUMN `VerifiedBuild` `VerifiedBuild` smallint(5) DEFAULT '0';
+
+ALTER TABLE `quest_template`
+  CHANGE COLUMN `VerifiedBuild` `VerifiedBuild` smallint(5) DEFAULT '0';
+
+ALTER TABLE `quest_template_addon`
+  CHANGE COLUMN `NextQuestID` `NextQuestID` mediumint(8) unsigned NOT NULL DEFAULT '0';
+
+ALTER TABLE `spell_bonus_data`
+  CHANGE COLUMN `entry` `entry` mediumint(8) unsigned NOT NULL DEFAULT '0';
+
+ALTER TABLE `version`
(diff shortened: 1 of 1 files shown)


wetbrownsauce:
Merged the PR into my local branch and updated my db. Compared the old db to the new for the columns specified, results are as follows:   

First off... Basic counts:

Test Preformed | Pre-PR | Post-PR
------------ | ------------- | -------------
COUNT(NextQuestID) with value > 0 | 586 | 586
COUNT(NextQuestID) with value = 0 | 8874 | 8878
COUNT(entry) from spell_bonus_data | 1268 | 1268
COUNT(wpguid) with value > 0 | 8 | 8
COUNT(wpguid) with value = 0 | 115097 | 115097

### **\*\*IMPORTANT\*\***
From the above data, we can see that 4 entries in `quest_template_addon` had their `NextQuestID` set to 0. These entries and their original data are as follows:

ID | NextQuestID
------------ | -------------
618 | -619
8554 | -619
12222 | -12222
12223 | -12223

Next I checked if the new records matched the old ones by exporting the pre-pr records to their own tables and comparing the old and new records with the following results:

`quest_template_addon` (non-zero values only due to the four changed records above)
```
SELECT COUNT(*) FROM `quest_template_addon`
LEFT JOIN `quest_template_addon_old`
ON quest_template_addon.NextQuestID = quest_template_addon_old.NextQuestID
WHERE quest_template_addon.ID = quest_template_addon_old.ID AND quest_template_addon.NextQuestID = quest_template_addon_old.NextQuestID;

Result: 586
```

`spell_bonus_data`
```
SELECT COUNT(*) FROM `spell_bonus_data`
LEFT JOIN `spell_bonus_data_old`
ON spell_bonus_data.entry = spell_bonus_data_old.entry
WHERE spell_bonus_data.entry = spell_bonus_data_old.entry;

Result: 1268
```

`waypoint_data` (non-zero values only)
```
SELECT COUNT(*) FROM `waypoint_data`
LEFT JOIN `waypoint_data_old`
ON waypoint_data.wpguid = waypoint_data_old.wpguid
WHERE waypoint_data.id = waypoint_data_old.id AND waypoint_data.point = waypoint_data_old.point AND waypoint_data.wpguid = waypoint_data_old.wpguid;

Result: 8
```

So if my checks are correct, all the data pre-pr matches the data post-pr with the in-question fields, with the exception of the four records mentioned above.

FrancescoBorzi:
@wetbrownsauce excellent job. Now I need to understand what those 4 negative quests are for.

wetbrownsauce:
Luckily there are only 4 records with negative `NextQuestID` and not a thousand, lol

FrancescoBorzi:
@wetbrownsauce it's not about the numbers, luckily you spotted this. We actually need to import 2 commits to fix that diff, and for that I opened a separate issue: https://github.com/azerothcore/azerothcore-wotlk/issues/1328

This can now be merged.

Issue #1326, opened 2019-01-19
[EPIC] Align SmartAI to TrinityCore


FrancescoBorzi:
This will make easy to exchange SmartAI scripts between AzerothCore and TrinityCore

### Global
- [x] Add missing fields in the table

### Event
- There are some minor difference in Events, shouldn't be an issue. But feel free to investigate.

### Action
- [x] Move `SMART_ACTION_SET_SIGHT_DIST` to `121`
- [x] Move `SMART_ACTION_FLEE` to `122`
- [x] Move `SMART_ACTION_ADD_THREAT` to `123`
- [x] Move `SMART_ACTION_LOAD_EQUIPMENT` to `124`
- [x] Move `SMART_ACTION_TRIGGER_RANDOM_TIMED_EVENT` to `125`
- [x] Move `SMART_ACTION_REMOVE_ALL_GAMEOBJECTS` to `126`

- [x] Remove the remaining extra overlap in enum (after the moves above)

- [x] Reserve missing enums, display proper error if someone tries to use them

- [x] `SMART_ACTION_ADD_QUEST` -> `SMART_ACTION_OFFER_QUEST` https://github.com/TrinityCore/TrinityCore/commit/ba4d6f63936fe3269e6c8b75358bc00a78fd49df
- [x] Add param3 to SMART_ACTION_CAST and SMART_ACTION_INVOKER_CAST (but be careful since we already have a param3 there, so maybe add that as param4 ?) https://github.com/TrinityCore/TrinityCore/commit/9072e0221b39c19fb6be154c623f45d6a9433db0
- [x] Get rid of unused parameter 'storageId' (third) in SMART_ACTION_SUMMON_CREATURE https://github.com/TrinityCore/TrinityCore/commit/a5c248b60ffd4981f6f62bc0640b4fcf1321a8c3
- [x] Reserve smart action 16 https://github.com/TrinityCore/TrinityCore/commit/4e55e797532ca404ce0e1e603762c17ead760f45
- [x] Core/Scripts: Extended SMART_ACTION_SET_INST_DATA to also set BossStates. https://github.com/TrinityCore/TrinityCore/commit/d1d0c64ad11cb3faf986a5332fec6b704fb8026c
- [ ] (Optional) cleaning https://github.com/TrinityCore/TrinityCore/commit/84bff76ba054683ca497b8b758cb86728d5f52a2
[...]

PR #1333, opened 2019-01-20
feat(SmartAI): improve add quest action

FrancescoBorzi:
##### CHANGES PROPOSED:

-  Rename ` SMART_ACTION_ADD_QUEST` to `SMART_ACTION_OFFER_QUEST`
-  This action can now open a dialog to offer the quest instead of adding it straight to the quest log
- With `action_param2` = `1` it will be like the old behaviour, with `0` it's the new behaviour
- All existing scripts will keep the same behaviour, except for creature `11216`

Feature by **@Treeston** imported from TC

###### ISSUES ADDRESSED:

- Updates https://github.com/azerothcore/azerothcore-wotlk/issues/1326

##### TESTS PERFORMED:

Builds

- Test that the current scripts that use this action are not broken and use the same behaviour (with the exception of creature `11216` that will have the new behaviour). So check some existing scripts that add quests.
- Test the script of creature `11216` and make sure it asks for quest with a quest dialog

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -393,7 +393,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                                 menu.SendQuestGiverQuestDetails(q, me->GetGUID(), true);
 #if defined(ENABLE_EXTRAS) && defined(ENABLE_EXTRA_LOGS)
                                 sLog->outDebug(
-                                        "LOG_FILTER_DATABASE_AI",
+                                        LOG_FILTER_DATABASE_AI,
                                         "SmartScript::ProcessAction:: SMART_ACTION_OFFER_QUEST: Player guidLow %u - offering quest %u",
                                         (*itr)->GetGUIDLow(),
                                         e.action.questOffer.questID);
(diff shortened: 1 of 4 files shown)


FrancescoBorzi:
sorry Treeston, I added the "@" before your name without actually meaning it. My only intention was to write your name to give credits (the notification was a mistake).

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -393,7 +393,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                                 menu.SendQuestGiverQuestDetails(q, me->GetGUID(), true);
 #if defined(ENABLE_EXTRAS) && defined(ENABLE_EXTRA_LOGS)
                                 sLog->outDebug(
-                                        "LOG_FILTER_DATABASE_AI",
+                                        LOG_FILTER_DATABASE_AI,
                                         "SmartScript::ProcessAction:: SMART_ACTION_OFFER_QUEST: Player guidLow %u - offering quest %u",
                                         (*itr)->GetGUIDLow(),
                                         e.action.questOffer.questID);
```
BarbzYHOOL: that looks horrible on github

FrancescoBorzi: what?

BarbzYHOOL: look

wetbrownsauce:
It would seem that the only `smart_script` that uses `action_type` 7 is 11216, aka the one using the new behavior, so all testing I did had to involve that.  

To test, I first checked that this new behavior functions, then after that I changed it back to the old behavior and re-tested to make sure the old behavior still functions. Results as follows:

- New Behavior: When selecting the gossip option: "How can I help?" from Eva Sarkhoff (the option after going through her whole story), the gossip window closes and pops-up the quest dialog offering a quest. This all seems correct. 👍 
- Old Behavior: After changing `action_param2` back to 1, doing the above results in the quest automatically being added to the players quest log, with not pop-up quest offer dialog. So it seems the old behavior still functions as it did if `action_param2` is set to 1. 👍

BarbzYHOOL:
@gtao725 would you like to help us test pull requests? If so, talk to me in the discord

Issue #1344, opened 2019-01-22
Table `command` have non-existing command 'spy', skip.


FrancescoBorzi:
##### SMALL DESCRIPTION:
Spy command added a new DB startup error that you can also reproduce each time you use `server info` in your console:

> AC>server info
AC>Table `command` have non-existing command 'spy', skip.
Table `command` have non-existing command 'spy', skip.
Table `command` have non-existing command 'spy', skip.
Table `command` have non-existing command 'spy', skip.
Table `command` have non-existing command 'spy', skip.
Table `command` have non-existing command 'spy', skip.
AzerothCore rev. 71e0ad005b01 2019-01-22 20:43:40 +0100 (master branch) (Unix, Debug)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
Server uptime: 18 minute(s) 31 second(s).
Update time diff: 10ms, average: 10ms.
AC> 

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/commit/71e0ad005b0191b561616c0cfecc212782e065f4

##### OPERATING SYSTEM:

GNU/Linux

##### MODULES:

No modules, clean AzerothCore

Deku:
@FrancescoBorzi is 6c233c4d8c49cd8f014f2df497d7c8a646b2bf74 applied in the database?

sucofog:
AddSC not declared on ScriptLoader.cpp for commandscripts

Deku:
It's declared but not invoked. Thanks for pointing that out.

PR #1346, opened 2019-01-23
fix(Core/Command): Invoke scripts for spy command

Deku:
Related to #1344

### DIFF
--- a/src/server/scripts/ScriptLoader.cpp
+++ b/src/server/scripts/ScriptLoader.cpp
@@ -629,6 +629,7 @@ void AddCommandScripts()
     AddSC_reload_commandscript();
     AddSC_reset_commandscript();
     AddSC_spectator_commandscript();
+    AddSC_spy_scripts();
     AddSC_tele_commandscript();
     AddSC_ticket_commandscript();
     AddSC_titles_commandscript();


src/server/scripts/ScriptLoader.cpp
```diff
@@ -629,6 +629,7 @@ void AddCommandScripts()
     AddSC_reload_commandscript();
     AddSC_reset_commandscript();
     AddSC_spectator_commandscript();
+    AddSC_spy_scripts();
```
BarbzYHOOL:
```suggestion
    AddSC_spy_commandscript();
```

to be on par with other command script names

btw need also to edit the name in the other cpp file

Deku: @BarbzYHOOL then I'll be including an article on the wiki on how to create a command and what rules to follow

Deku: And btw, it's not just a command script, it's a player script as well

BarbzYHOOL: ??

I mean that other scripts related to a command (which is what .spy is about, it's a command, if you remove the command it serves no purpose, am I mistaken?) are written like this apparently: `AddSC_commandname_commandscript();`

```
AddSC_reload_commandscript();
AddSC_reset_commandscript();
AddSC_spectator_commandscript();
```

No?

Deku: Yes, I understand your point. What I meant to explain is why it was written as "*_script*" and not "*_commandscript*"

BarbzYHOOL: but if it's in two categories, it should be in the "primary" category, which (to me) seems to be "commandscript"


btw when I use the suggest changes button, it's just a suggestion (a quick suggestion without a lot of brain time invested into :P)

FrancescoBorzi:
@Deku tested, does not work :(

The startup DB error is gone but the feature does not work. I've tried to spy some players but it doesn't work.

PR #1353, opened 2019-01-25
style: Reorder DBs in config.sh.dist

BarbzYHOOL:
##### CHANGES PROPOSED:

-  Now use the logical order of DBs.
- **Also disabled extractors compilation by default.** (CTOOLS if I'm not making a mistake here)
 

###### ISSUES ADDRESSED:

It was ordered like this:
- characters
- auth
- world

which is not the logical/usual order ^^

Also disabled tools compilation by default as it's something you only do rarely (or never).

### DIFF
--- a/conf/config.sh.dist
+++ b/conf/config.sh.dist
@@ -41,7 +41,7 @@ CSCRIPTS=ON
 # compile server
 CSERVERS=ON
 # compile tools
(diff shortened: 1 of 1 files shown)


BarbzYHOOL:
@Stoabrogga thank you, sir!

PR #1362, opened 2019-01-27
fix(travis): prevent failures in forks

FrancescoBorzi:
Fix from @Rochet2


Rochet2:
Test build https://travis-ci.com/Rochet2/azerothcore-wotlk/builds/98719137

Issue #1251, opened 2019-01-09
emote command allow chat between factions


IngMak:
##### SMALL DESCRIPTION:
write in chat: /e hello world
enemy factions can understand language.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
with default worldserver.conf enemy factions can`t understand messages.

##### CURRENT BEHAVIOUR:
using chat command "/e" allow chat between enemy factions with default .conf

##### STEPS TO REPRODUCE THE PROBLEM:
0. loggin with two characters from opposite factions (1 horde, 1 alliance)
1. write in chat: /e hello, can we be friends?

check chat is legible between factions.

##### EXTRA NOTES:
TC have this issue.
i think chathandler have to be reviewed.

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
 cfa2f86 

##### OPERATING SYSTEM:

windows/linux

##### MODULES:
anticheat

poszer:
Issue confirmed

poszer:
Client side problem :
[battle.net topic](https://us.battle.net/forums/en/wow/topic/2661227432)

PR #1382, opened 2019-01-28
[Fix] Emote chat between factions. Issue : #1251

masterking32:
According #1251 , Fix emote chat language between two factions.


FrancescoBorzi:
@masterking32 good job!

masterking32:
Need to close that xD,
I think wow doesn't support that.
[TOPIC : My Favorite Cross-Faction /emote](https://us.battle.net/forums/en/wow/topic/2661227432)

PR #1385, opened 2019-01-29
Docker: Naming errors for some sql files

Raschaad:
Docker: Naming errors for some sql files

##### CHANGES PROPOSED:

Fixes naming errors in `generate-databases.sh`  

###### ISSUES ADDRESSED:

When running `docker-compose up`, the `ac-database` exited, since it didn't find the following sql files:
- /sql/auth_update.sql
- /sql/characters_update.sql
- /sql/world_update.sql

By running `docker-compose run ac-database /bin/bash` i checked the files with `ls /sql` and could find the files, but they were named in plurally. So when I changed in the `generate-databases.sh` the names, the databases could be generated, and at last the authserver and worldserver could connect to the database.

##### TESTS PERFORMED:

Tests only done on Mac. But I guess, since the fix was made in docker, it should have the same effects on all platforms

1. `git clone https://github.com/Raschaad/azerothcore-wotlk.git`
2. Copy the data (client files) to `docker/worldserver/data` (as in the [Installation with Docker](https://github.com/azerothcore/azerothcore-wotlk/wiki/install-with-Docker) stated)
3. Generate your server configuration files: `./bin/acore-docker-generate-etc`
4. Compile AzerothCore: `./bin/acore-docker-build`
5. Run the containers: `docker-compose up`

With these changes, the server should be up after a few minutes. Before these changes, this didn't work for me.

##### Target branch(es):

Master

### DIFF
--- a/docker/database/generate-databases.sh
+++ b/docker/database/generate-databases.sh
@@ -21,13 +21,13 @@ mysql -u root -p$MYSQL_ROOT_PASSWORD acore_world < /sql/world_base.sql
 
 
 echo "Importing auth updates..."
-mysql -u root -p$MYSQL_ROOT_PASSWORD acore_auth < /sql/auth_update.sql
+mysql -u root -p$MYSQL_ROOT_PASSWORD acore_auth < /sql/auth_updates.sql
 
 echo "Importing characters updates..."
-mysql -u root -p$MYSQL_ROOT_PASSWORD acore_characters < /sql/characters_update.sql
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
@Raschaad thank you very much for your contributions.

Those files were renamed but we forgot to align the Docker script. So your fix is correct.

Good job!

BarbzYHOOL:
Yes seems correct, I didn't know docker used this, sorry :P

Issue #1269, opened 2019-01-12
[Talents] Mage permafrost apply the heal reduction even if dispelled


JorgeSally:
##### SMALL DESCRIPTION:

The permafrost (spell from mage) which is applied by frostbolt still persist even after the dispell of frostbolt (-20% until the end of of the frostbolt debuff duration)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

When we dispell the frostbolt, we don't get the -20% heal anymore (like the improved mind blast of the shadow priest)

##### CURRENT BEHAVIOUR:

When you apply the permafrost ( 11175, 12569, 12571 talents) to a target with frostbolt, he get -20% healing, the effect should last 12s, but if you dispell it before, you still have to wait 12s for the -20% to "disappear"

##### STEPS TO REPRODUCE THE PROBLEM:

1. Make a mage level at least level 17 to get the talents
2. Make a character that can heal himself and dispell (like a priest or a paladin)
3. Duel together and cast frostbolt with the mage
4. Dispell with the second character and heal yourself

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

a4c8e98d1f71e45daf26c827ec00ae2c2131daad

##### OPERATING SYSTEM:

Windows 10

##### MODULES:

Transmog NPC

##### OTHER CUSTOMIZATIONS:

No

PR #1389, opened 2019-01-30
Core/Spells: Mage permafrost talent

poszer:
- Will fix issue with permafrost (Talent ) apply the heal reduction even if dispelled.

Co-Authored-By: talamortis <talamortis@gmail.com>

##### CHANGES PROPOSED:

-  
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1269

##### TESTS PERFORMED:

Builded without errors, tested in game, Os: Win 

1. Make a mage level at least level 17 to get the talents
2. Make a character that can heal himself and dispell (like a priest or a paladin)
3. Duel together and cast frostbolt with the mage
4. Dispell with the second character and heal yourself

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -8241,13 +8241,6 @@ void Spell::PrepareTriggersExecutedOnHit()
 
             break;
         }
-        case SPELLFAMILY_MAGE:
-        {
-            // Permafrost
-            // if (m_spellInfo->SpellFamilyFlags[1] & 0x00001000 ||  m_spellInfo->SpellFamilyFlags[0] & 0x00100220)
-            // m_preCastSpell = 68391;
-            break;
-        }
     }
 
     // handle SPELL_AURA_ADD_TARGET_TRIGGER auras:
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -8244,8 +8244,8 @@ void Spell::PrepareTriggersExecutedOnHit()
         case SPELLFAMILY_MAGE:
         {
             // Permafrost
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -8244,8 +8244,8 @@ void Spell::PrepareTriggersExecutedOnHit()
         case SPELLFAMILY_MAGE:
         {
             // Permafrost
-            if (m_spellInfo->SpellFamilyFlags[1] & 0x00001000 ||  m_spellInfo->SpellFamilyFlags[0] & 0x00100220)
-            m_preCastSpell = 68391;
+            // if (m_spellInfo->SpellFamilyFlags[1] & 0x00001000 ||  m_spellInfo->SpellFamilyFlags[0] & 0x00100220)
+            // m_preCastSpell = 68391;
```
FrancescoBorzi: why commenting those lines instead of just removing them?

talamortis: We commented those lines just to be on the safe side as because its using flags we wouldn't know what else this effects (prob nothing) but we can always put them back in if needs be and then find a alternative fix for this

FrancescoBorzi: ok

Sdyess: Okay but we use version control, there's permanent history of what you remove. If you don't take them out now it'll likely never get removed.

FrancescoBorzi: That is also true. On one hand it's easier to find as commented rather than in history (but still possible to find). On the other hand, the code is much cleaner if we remove such dead code instead of commenting it. I personally prefer to remove them in such cases, but I wouldn't be strict about it.

talamortis:
Will merge this as @poszer already tested :)

PR #1391, opened 2019-01-31
fix(clang): fix build on version 7

Stoabrogga:
##### CHANGES PROPOSED:
Removed some unused private fields in preparation for using clang 7 (prevent compiler warnings).

###### ISSUES ADDRESSED:
None

##### TESTS PERFORMED:
Tested compilation on Ubuntu 16.04 using clang 7.0.1

Edit by Shin: see comment below

None

##### Target branch(es):
Master

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -143,7 +143,7 @@ void DamageInfo::BlockDamage(uint32 amount)
 
 ProcEventInfo::ProcEventInfo(Unit* actor, Unit* actionTarget, Unit* procTarget, uint32 typeMask, uint32 spellTypeMask, uint32 spellPhaseMask, uint32 hitMask, Spell* spell, DamageInfo* damageInfo, HealInfo* healInfo, SpellInfo const* triggeredByAuraSpell)
 :_actor(actor), _actionTarget(actionTarget), _procTarget(procTarget), _typeMask(typeMask), _spellTypeMask(spellTypeMask), _spellPhaseMask(spellPhaseMask),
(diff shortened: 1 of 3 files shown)


FrancescoBorzi:
@Stoabrogga good job.

- I tested the compilation with clang 6 and it works too
- travis already checked that clang 3 is also not broken

this change shouldn't have any impact in game, but I think we should double check this anyway before merging.

We could test for example:

- the boss Flame Leviathan
- try just a couple of spells
- engage fight with a couple of random NPCs and see if they still work fine

@wetbrownsauce when you have some time, can you please have a look a this?

wetbrownsauce:
I'm not seeing any issues in game, but I haven't been able to test flame leviathan because that fight is a bitch to test by yourself. If anyone has an easy way please let me know.

Stoabrogga:
Is it really necessary to test the Flame Leviathan? The game logic is not touched through this PR, only the unused constructor parameter "unit" is removed:
```FlameLeviathanPursuedTargetSelector(Unit* unit)```

This class is used in
```class spell_pursue : public SpellScriptLoader```
to check if valid targets exist for the pursuit, but the constructor parameter "unit" was never used in the first place (that's also what clang complained about).

I'm just asking because this PR holds back PR #1410 concerning the Travis configuration.

FrancescoBorzi:
No, I don't think it's strictly necessary.

Stoabrogga:
Ok, so if no one has anything against it, I'm marking this as "to be merged".

Stoabrogga:
Merging this in order to proceed with the Travis updates. Thanks for your feedback @wetbrownsauce @FrancescoBorzi

Issue #1156, opened 2018-12-27
DB errors about `achievement_criteria_data`


FrancescoBorzi:
Revision: b1b48b9

The table `achievement_criteria_data` produces the following startup errors:

> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 4989 Type: 19), ignored.
> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 4988 Type: 19), ignored.
> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5620 Type: 73), ignored.
> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5632 Type: 73), ignored.
> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 5619 Type: 73), ignored.
> 2018-12-26 15:46:14 Table `achievement_criteria_data` has data for non-supported criteria type (Entry: 4987 Type: 19), ignored.

We tried to remove them from the database and this is the feedback from testing by @wetbrownsauce:

> As for the achievement_criteria changes, most of the affected achievements are actually statistics for number of 5 player, 10 player, and 25 player instances entered. These don't appear to work before or after the PR, so the changes in their db table make no difference aside from removing start up warnings/errors.

wetbrownsauce:
Some additional info about the specific entries referenced in case it can help with testing:

- Entry 4987 in achievement_criteria.dbc is linked to Entry 932 in achievement.dbc which is "Total 5-player dungeons entered".
- Entry 4988 in achievement_criteria.dbc is linked to Entry 933 in achievement.dbc which is "Total 10-player raids entered".
- Entry 4989 in achievement_criteria.dbc is linked to Entry 934 in achievement.dbc which is "Total 25-player raids entered".
- Entry 5619 in achievement_criteria.dbc is linked to Entry 1241 in achievement.dbc which is "Mal'Ganis defeated (Caverns of Time: Stratholme)".
- Entry 5620 in achievement_criteria.dbc is linked to Entry 1515 in achievement.dbc which is "Mal'Ganis defeated (Heroic CoT: Stratholme)".
- Entry 5632 in achievement_criteria.dbc is linked to Entry 1103 in achievement.dbc which is "Lich King 5-player dungeons completed (final boss killed)".

[...]

PR #1405, opened 2019-02-02
fix: remove data for unsupported criteria type

FrancescoBorzi:
This change was already tested by @wetbrownsauce in another PR (which was closed because containing other unwanted changes), the result was:

> As for the achievement_criteria changes, most of the affected achievements are actually statistics for number of 5 player, 10 player, and 25 player instances entered. These don't appear to work before or after the PR, so the changes in their db table make no difference aside from removing start up warnings/errors.

##### CHANGES PROPOSED:

-  Fix DB startup errors about unsupported achievement criteria types

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1156

##### TESTS PERFORMED:

- Run and those DB errors are gone

- Just make sure we're not breaking anything that is currently working, apparently this was already tested by @wetbrownsauce 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1549149082825412953.sql
+++ b/data/sql/updates/pending_db_world/rev_1549149082825412953.sql
@@ -1,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1549149082825412953');
+
+DELETE FROM `achievement_criteria_data` WHERE `criteria_id` IN (4989, 4988, 5620, 5632, 5619, 4987);


FrancescoBorzi:
@wetbrownsauce since you already tested this before, feel free to mark this as ready to be merged

FrancescoBorzi:
@wetbrownsauce that is the intended behaviour, we can add those values back once those types are supported

FrancescoBorzi:
@wetbrownsauce I forgot to mention: we indeed need to fix it properly by implementing those types and adding their data back to the DB.

PR #1415, opened 2019-02-05
DB/Quest: Quest 4866 and 4224 moved from core to DB

pangolp:
##### CHANGES PROPOSED:

*  Add smart_scripts
*  Remove c++ script

###### ISSUES ADDRESSED: No

Closes 

##### TESTS PERFORMED:

* Test on Windows.

To test the changes in the first quest, you must:
* .q to 4866
* .aura 16468

To test the changes in the second mission:
* .q to 4224

* .list c 9563

None in particular, just change the c++ script to an SQL script

The ids of the gossip, I took them from the same script. I didn't really know which ones to use, but since those were free and there is no official information about them, take those. I don't have the translations, so at the moment they would only be in English and neither would the broadcast_text.

##### Target branch(es):

Master


src/server/scripts/ScriptLoader.cpp
```diff
@@ -202,7 +202,7 @@ void AddSC_instance_zulgurub();
 //void AddSC_alterac_mountains();
 void AddSC_arathi_highlands();
 void AddSC_blasted_lands();
-void AddSC_burning_steppes();
+//void AddSC_burning_steppes();
```
FrancescoBorzi: better remove this line

src/server/scripts/ScriptLoader.cpp
```diff
@@ -789,7 +789,7 @@ void AddEasternKingdomsScripts()
     //AddSC_alterac_mountains();
     AddSC_arathi_highlands();
     AddSC_blasted_lands();
-    AddSC_burning_steppes();
+    //AddSC_burning_steppes();
```
FrancescoBorzi: better remove this line

BarbzYHOOL:
good

pangolp:
Did you remove the lines or should I? The truth is that I was thinking about removing them, but then I thought that maybe in the future you could add something, but in that case, you would create such lines. Wouldn't the c++ file directly remove it also if you remove those lines? @FrancescoBorzi

pangolp:
All this I'm doing, that I don't know if it's too much or too little I owe you. And to use your core and to take advantage of all the arrangements that you make without returning anything to you, does not seem to me to be correct. They couldn't be great scripts, but at least I want to help you with something.

I had a huge patience with this github, because I'm new and that's good because the new contributors, you can trust.

@BarbzYHOOL @FrancescoBorzi

BarbzYHOOL:
Good work, and for the gossip IDs, I don't know, if they are unused then why not.

Would like the opinion of someone more experienced.

FrancescoBorzi:
about the SQL IDs, if possible I would keep the same as TC, if not possible then just the next free IDs should be fine

Stoabrogga:
Checked the gossip IDs and they are ok in my opinion. ID 2061 and 2062 already exist, but they contain the same rows as will be inserted by the SQL script.

pangolp:
@poszer Don't eliminate those lines in one of the commits?

@FrancescoBorzi Regarding the IDs, the truth is that I put those because I didn't have any information, I checked the TC tables and it seemed to me that they didn't have them either. That's why I relied on the numbers that appeared in the C++ script. It just so happened that those values were free. Too bad you don't have the broadcast_text or the translations.

@Stoabrogga Yes, the gossip coincide because in reality some were and others I added them, it happens that for the script to be complete, I inserted them again just in case.

pangolp:
Anyway, I realized that the npc to which the mission should be delivered, I think it lacks a flag to accept it. Because on the map he shows you the position, but when you arrive, he doesn't let you hand it over. Unless it's because I didn't do the chain and I directly gave myself the mission with the command.

comix1988:
I have tested both quests:
- the 4866 quest works perfectly;
- the quest 4224 works but you don't see the dialogs in the text window; (dialogues at this [link](https://wowwiki.fandom.com/wiki/Quest:The_True_Masters_(6));

dialogo

moreover the quest 4224 is part of a chain-line removed in 3.0.2

chain-line

@pangolp

pangolp:
@comix1988 Good afternoon, thank you for testing the missions.  Unfortunately, I did not have the texts.  As for whether it was deleted, I simply found it in the core and moved it to the database, but if it was deleted, then maybe we should delete it.  The mission texts, do you have them?

comix1988:
you can find [here](https://wowwiki.fandom.com/wiki/Quest:The_True_Masters_(6))

FrancescoBorzi:
good job @pangolp and @comix1988 !

BarbzYHOOL:
might need to delete the useless text from the db then

PR #1417, opened 2019-02-05
feat(DB): test accounts using various GM levels

BarbzYHOOL:
##### CHANGES PROPOSED:

All types of GM accounts for testing purpose.

##### TESTS PERFORMED:

I use these on my server.

##### NOTES

I'd like to remove the useless comments like:
```
/*!40000 ALTER TABLE `account_access` DISABLE KEYS */;
```

Should I ?

### DIFF
--- a/data/sql/base/db_auth/account_access.sql
+++ b/data/sql/base/db_auth/account_access.sql
@@ -19,16 +19,16 @@ CREATE TABLE `account_access`
 LOCK TABLES `account_access` WRITE;
 /*!40000 ALTER TABLE `account_access` DISABLE KEYS */;
 INSERT INTO `account_access` (`id`, `gmlevel`, `RealmID`) VALUES
-(1,  4,  -1),
-(2,  4,  -1),
-(3,  3,  -1),
-(4,  3,  -1),
-(5,  2,  -1),
-(6,  2,  -1),
-(7,  1,  -1),
-(8,  1,  -1),
-(9,  0,  -1),
-(10, 0,  -1);
+(1,  4, -1),
+(2,  4, -1),
+(3,  3, -1),
+(4,  3, -1),
+(5,  2, -1),
+(6,  2, -1),
+(7,  1, -1),
+(8,  1, -1),
+(9,  0, -1),
+(10, 0, -1);
 /*!40000 ALTER TABLE `account_access` ENABLE KEYS */;
 UNLOCK TABLES;
 /*!40103 SET TIME_ZONE=@OLD_TIME_ZONE */;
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
so the idea is changing the default accounts to give them different GM levels ?

I like the idea, actually in the old MaNGOS we had something like accounts named `administrator`,` gamemaster`, `moderator`,`player`, so the name of the account was already suggesting which level it had.

We could do something similar, to have the account names reflecting their security level.

Deku:
Agree with @FrancescoBorzi on this. @BarbzYHOOL can you please rename the default accounts and make them 4-5 and not 10?

Deku:
@BarbzYHOOL about your note, those are MySQL-specific instructions. Check:

- https://dev.mysql.com/doc/refman/8.0/en/comments.html
- https://stackoverflow.com/questions/2949177/what-does-xxxxxx-statement-generated-by-mysqldump-mean

BarbzYHOOL:
If I change name, then need to change passwords with the hash etc...

What about a comment column instead? Free to be written to

And 10 accs is better than 5  (2 of each)

BarbzYHOOL:
So guys, I really think it'd be better to have a comment column and then we can explain what is what for

That can be handy too when you have GMs or VIP testers or guests on your dev server, then you can let a comment for each user access.


The only small problem is that I need to edit the create table, I guess we don't need to add an update sql for that (alter table), it would be like a duplicate and we'd forget about it, what do you think?

FrancescoBorzi:
> So guys, I really think it'd be better to have a comment column and then we can explain what is what for

:+1: 

> The only small problem is that I need to edit the create table, I guess we don't need to add an update sql for that (alter table), it would be like a duplicate and we'd forget about it, what do you think?

not sure what you mean but if you want to edit the table structure (in this case by adding a column) you should add this change into an sql update file (do not change the base file)

BarbzYHOOL:
> not sure what you mean but if you want to edit the table structure (in this case by adding a column) you should add this change into an sql update file (do not change the base file)

Yes, I guess I confused stuff. Ok I'll do this soon (remind me if i forget lol)

Endeffect-AT:
Just looked around and i like the idea! And i agree that a comment column will be better then changing the name, myself would find a Use for that :D

BarbzYHOOL:
omg i'm busy, if you want to edit it endeffect it's okay

else i'll do it if I don't forget about it xD

Endeffect-AT:
Challenge accepted, will look at it soon :grinning:

BarbzYHOOL:
Gonna make a new PR for this

PR #1423, opened 2019-02-06
chore(PCH): remove "ScriptPCH.h" from cpp scripts

Stoabrogga:
##### CHANGES PROPOSED:
Remove all inclusions of "ScriptPCH.h" from the cpp scripts and replace them by the appropriate dependencies. "ScriptPCH.h" is intended to be used for precompiling headers, it should not be included directly because it hides the actual dependencies. It's sole purpose is to be used as a tool to reduce compilation time, controlled by the compiler options mentioned below.

###### ISSUES ADDRESSED:
None

##### TESTS PERFORMED:
Compiled using clang 3.8 and 7 on Ubuntu 16.04

Just compile the source. No functionality has been touched, only the header includes. Be sure to turn precompiled headers off, either through setting these options in config.sh:
```
CSCRIPTPCH=OFF
CCOREPCH=OFF
```

or through these cmake options:
```-DUSE_COREPCH=0 -DUSE_SCRIPTPCH=0```

This ensures that no precompiled headers exist and all necessary dependencies are correctly included.

None

##### Target branch(es):
Master


BarbzYHOOL:
So now, Travis will fail if a header is missing, right?

Stoabrogga:
@BarbzYHOOL 
Yes, this is correct, as Travis is configured to build without pre-compiled headers.

Deku:
It's still working with PCH on?

Stoabrogga:
@Deku 
Yes, the pre-compiled headers are actually configured here:
https://github.com/azerothcore/azerothcore-wotlk/blob/832bccd9fa8ae5f3dd20eaeed244e74cd700f2de/src/cmake/macros/FindPCHSupport.cmake#L91-L104

Stoabrogga:
I think this is good to go. Travis run was also ok, compiled without errors.

Stoabrogga:
Merged. Thanks for your reviews and feedback :+1:

FrancescoBorzi:
good job @Stoabrogga

BarbzYHOOL:
forgot to say, but thank you doing that chore (and uncovering the mystery of PCH xD)

PR #1427, opened 2019-02-07
On player complete quest hook

talamortis:
Added a new hook for when player completes a quest.

Tested and working in game


talamortis:
Thanks @FrancescoBorzi will merge :)

PR #1433, opened 2019-02-09
Core/Movement: Implement move time skipped handler 

Viste:
Implement CMSG_MOVE_TIME_SKIPPED handler and move it to MovementHandler.cpp

**Changes proposed:**

The way the MOVE_TIME_SKIPPED packet works is as follow:

-  The client sends it when the player [1] is moving but the client froze for a certain period of time.

- The packet only contains two information: the GUID of the moving unit. And the time elasped since the last movement packet has been sent.

- The meaning of the packet from the point of the view of the client that sends it is basically: I didn't run the physical simulation of the unit that I control for a certain period of time. Please, just play at a later time the last movement that was sent (new timestamp = previous timestamp + "time skipped").

- the server upon receiving that packet should
  - - retransmit the packet to the other clients that have a unit nearby.
  - - update the movement timestamp of that unit (new timestamp = previous timestamp + "time skipped").
- the clients upon receiving also update the movement timestamp of that unit (new timestamp = previous timestamp + "time skipped") so that the previous move sent is "played" at a later time.

[1] or more generaly the unit currently moved by that client: it can be another player (in case of mind control for example), a vehicle, ...

More information on the packet exchange can be found in [this sniff extract](https://gist.github.com/chaodhib/e66010a8ecb77b406ff86f6bbe93a5e0#file-move-time-skipped-sniff). In it, we can see 3 packets sent by the client. The unit is moving between the first and third packet (and also before the first and after the third but that's not important here). The distance traveled between the first and third packet is only 500ms * run speed. Yet, the time difference between the 1st & 3rd packet is 500ms + 86ms. 86ms is the "time skipped". This confirms that the client didn't run the physical simulation during that duration.

**Target branch:** master
**Issues addressed:** This is retail like. The implementation of that packet handler was missing. The movement should feel more smooth.

Original Author — [chaodhib](/chaodhib)


FrancescoBorzi:
what needs to be tested? what can this affect?

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -725,3 +725,35 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
...
+        return;
+    }
+
+    mover->m_movementInfo.time += timeSkipped;
+
+    WorldPacket data(MSG_MOVE_TIME_SKIPPED, recvData.size());
+    data.appendPackGUID(guid);
+    data << timeSkipped;
+    GetPlayer()->SendMessageToSet(&data, false);
+}
```
Winfidonarleyan: Need added new line

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -725,3 +725,35 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
     _player->SetSummonAsSpectator(false);
     _player->SummonIfPossible(agree, summoner_guid);
 }
+
+void WorldSession::HandleMoveTimeSkippedOpcode(WorldPacket& recvData)
+{
+    sLog->outString("WORLD: Received CMSG_MOVE_TIME_SKIPPED");
```
Winfidonarleyan: Need log debug

Viste:
@FrancescoBorzi  The movement should feel more smooth. and its all its affect to time skipped packets 
in fact its a diff between  movement packets sended to client

RaistlinAstragon:
I have tested this PR and i have made two printscreens of .pinfo command (for timing) and .gps (for coordinates) but i think is not the right manner to test this PR. What can i do for testing your modifications @Viste ?
printscreen_end
printscreen_initial

Viste:
better open WoW in 2windows and look at movement

comix1988:
I tested this PR as required, here is the [video](https://youtu.be/-ts571Bub04)

everything works normally

FrancescoBorzi:
wow, that's testing!

good job @Viste @comix1988 @RaistlinAstragon

BarbzYHOOL:
to me it looks like it will improve movement when there are packets lost, so no idea how to reproduce this scenario

RaistlinAstragon:
Great job @comix1988 ;)

src/server/game/Handlers/MovementHandler.cpp
Nefertumm: Winfi it's right, this needs to be only for debug purposes.

BarbzYHOOL: Example here https://github.com/search?q=org%3Aazerothcore+sLog-%3Edebug&type=Code (need to find the right category/filter)

BarbzYHOOL:
Still one thing left to do (see winfi's review)

src/server/game/Handlers/MovementHandler.cpp
Stoabrogga: I think this should be changed as follows. At least it is handled this way in other debug messages:
```suggestion
#if defined(ENABLE_EXTRAS) && defined(ENABLE_EXTRA_LOGS)
    sLog->outDebug(LOG_FILTER_NETWORKIO, "WORLD: Recvd CMSG_MOVE_TIME_SKIPPED");
#endif
```

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -725,3 +725,35 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
...
+    uint64 guid;
+    uint32 timeSkipped;
+    recvData.readPackGUID(guid);
+    recvData >> timeSkipped;
+
+    Unit* mover = GetPlayer()->m_mover;
+
+    if (!mover)
+    {
+        sLog->outString("WorldSession::HandleMoveTimeSkippedOpcode wrong mover state from the unit moved by the player [" UI64FMTD "]", GetPlayer()->GetGUID());
```
Stoabrogga: If this is an error message I think `sLog->outError` should be used:
```suggestion
        sLog->outError("WorldSession::HandleMoveTimeSkippedOpcode wrong mover state from the unit moved by the player [" UI64FMTD "]", GetPlayer()->GetGUID());
```

src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -725,3 +725,35 @@ void WorldSession::HandleSummonResponseOpcode(WorldPacket& recvData)
...
+        return;
+    }
+
+    // prevent tampered movement data
+    if (guid != mover->GetGUID())
+    {
+        sLog->outString("WorldSession::HandleMoveTimeSkippedOpcode wrong guid from the unit moved by the player [" UI64FMTD "]", GetPlayer()->GetGUID());
```
Stoabrogga: If this is an error message I think `sLog->outError` should be used:
```suggestion
        sLog->outError("WorldSession::HandleMoveTimeSkippedOpcode wrong guid from the unit moved by the player [" UI64FMTD "]", GetPlayer()->GetGUID());
```

FrancescoBorzi:
@Stoabrogga what about merging this (since it was tested already) and open a new PR to address your changes?

Stoabrogga:
@FrancescoBorzi 
Ok, I'll open a new PR after this one is merged.

src/server/game/Handlers/MovementHandler.cpp
BarbzYHOOL: yes thank you for example stoabrogga!

BarbzYHOOL:
or you can edit it directly in github too

PR #1446, opened 2019-02-10
chore(BG): change default MinPlayersPerTeam for new servers

FrancescoBorzi:
##### CHANGES PROPOSED:

Changed default `MinPlayersPerTeam` to more realistic values for private servers.

According to my experience as a player in small/medium servers, there is no real reason not to lower those values.

Notes:
-  For 99% of the cases, server administrator do want to lower those limits so let's just save an extra step
-  This affects only new AC setups, not existing servers
- Those values can be always changed according to the user's preference

##### TESTS PERFORMED:

The query runs

Trivial change, no tests needed.

##### Target branch(es):

Master


wlasser:
I think that is no good idea to make it at default database settings. It's can be like mod or patch.

Viste:
how to disapprove

FrancescoBorzi:
The whole point of this change was saving an extra step for 99% of the cases.

I don't really see the value of keeping the current defaults, but that's just my opinion and I'm going to close this if you guys disagree.

PR #1447, opened 2019-02-10
chore(conf): improve defaults

FrancescoBorzi:


wlasser:
I agree with update time interval and announcer of BG's. But other settings is under question.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2561,7 +2561,7 @@ BattleGround.PremadeGroupWaitForMatch = 1800000
 #        Default:     0 - (Disabled)
 #                     1 - (Enabled)
 
-Battleground.GiveXPForKills = 0
+Battleground.GiveXPForKills = 1
```
Deku: This shouldn't be modified, since it changes the original behaviour of the game.

src/server/authserver/authserver.conf.dist
```diff
@@ -166,7 +166,7 @@ LogColors = ""
 #        Default:     0 - (Disabled)
 #                     1 - (Enabled)
 
-EnableLogDB = 0
+EnableLogDB = 1
```
Deku: This should be used for debug purposes only.

src/server/worldserver/worldserver.conf.dist
```diff
@@ -656,7 +656,7 @@ LogColors = ""
 #        Default:     0 - (Disabled)
 #                     1 - (Enabled)
 
-EnableLogDB = 0
+EnableLogDB = 1
```
Deku: This should be used for debug purposes only.

FrancescoBorzi:
Thanks for the feedback guys. Now I'm at work and cannot edit, I will address your comments once I'm at home.

PR #1480, opened 2019-02-17
DB squash & begin AC 3.0.0 development

FrancescoBorzi:
## DO NOT MERGE


FrancescoBorzi:
- Tested with a traditional setup, compiling `worldserver` both with `gcc` and `clang` (ubuntu default packages). Both worked.
- Tested with Docker, everything is fine.

The DB installation procedure is much faster because of the DB squash.

As soon as the travis pipeline passes, we can proceed.

BarbzYHOOL:
for versions we should use tags (never tried before myself though)

PR #1486, opened 2019-02-17
feat(CI): improve Travis configuration

Stoabrogga:
##### CHANGES PROPOSED:
- use random DB names in order to detect statements which address a specific DB
- split compilation and DB check (DB check and dry run only have to run once)
- use separate shell scripts for easier module Travis integration

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
needs to be tested through this PR

see above

none

##### Target branch(es):
Master


.travis.yml
```diff
@@ -71,18 +69,28 @@ install:
...
-  - git clone --depth=1 --branch=master --single-branch https://github.com/ac-data/ac-data.git /home/travis/build/azerothcore/azerothcore-wotlk/env/dist/data
-  - cp ./data/travis/worldserver.conf ./env/dist/etc/worldserver.conf
-  - ./env/dist/bin/worldserver --dry-run
-  - ./apps/ci-error-check.sh
+
+jobs:
+  include:
+    - stage: db_test
+      env:
+        - TRAVIS_BUILD_ID="1"
```
FrancescoBorzi: Why do we need this again?

Stoabrogga: I wanted the job to be executed just for one build, but this doesn't work (it overrides the matrix and executes only this single job). I'm just playing around with Travis, still new to this.

apps/ci/ci-script.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echo "compile core"
+export CCACHE_CPP2=true
+ccache -s
+timeout 2580 bash ./acore.sh "compiler" "all"
+ccache -s
+
+echo "start worldserver dry-run"
```
FrancescoBorzi: this `echo` should be inside the `if` block

Stoabrogga: Fixed this, thanks

Stoabrogga:
I removed the clang versions from the Travis configuration and included them to the scripts. Should now be good to go. Thanks @FrancescoBorzi for taking a look at this.

FrancescoBorzi:
Excellent job @Stoabrogga

PR #1492, opened 2019-02-18
DB/Item: Hateful Gladiator's Longbow No Range

CookieMonsterDev:
##### CHANGES PROPOSED:

-  Fix Range and ammo type of weapon

###### ISSUES ADDRESSED:

Create a hunter and equip the weapon ".additem 42489"
Try to use a spell like "Volley"
Observe how the range is only 6 yards instead of 35 yards.

##### TESTS PERFORMED:

Tested on local windows. Works perfectly when applying my sql file

Create a hunter and equip the weapon ".additem 42489"
Try to use a spell like "Volley"
Observe how the range is now proper "Might have to delete your chache"

##### Target branch(es):

Master


comix1988:
I tested this PR and I did not notice any distance problems before implementing the SQL file.
The only problem was that the green circle was not marked on the ground but with this sql it was solved.

FrancescoBorzi:
good job @CookieMonsterDev , thanks @Nefertumm and @comix1988 for reviewing/testing

PR #1511, opened 2019-02-21
Core/Skills: Add FishingBaseSkillLevel dynamic change method

brussens:
##### CHANGES PROPOSED:
-  Added new method `ObgectMgr::ChangeFishingBaseSkillLevel(uint32 entry, int32 skill)` to change the fishing base skill level for a particular zone.

##### TESTS PERFORMED:
- [x] Tested in build Win10 x64
- [x] Tested in game

- [x] Test build in Linux (Travis)

### DIFF
--- a/src/server/game/Globals/ObjectMgr.h
+++ b/src/server/game/Globals/ObjectMgr.h
@@ -1013,6 +1013,7 @@ class ObjectMgr
         void LoadPetNumber();
         void LoadCorpses();
         void LoadFishingBaseSkillLevel();
+        void ChangeFishingBaseSkillLevel(uint32 entry, int32 skill);
 
         void LoadReputationRewardRate();
         void LoadReputationOnKill();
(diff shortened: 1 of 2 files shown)


src/server/game/Globals/ObjectMgr.h
```diff
@@ -1013,6 +1013,7 @@ class ObjectMgr
         void LoadPetNumber();
         void LoadCorpses();
         void LoadFishingBaseSkillLevel();
+        void ChangeFishingBaseSkillLevel(uint32 entry, int32 skill);
```
FrancescoBorzi: where is this method used?

brussens: I use this in module for some scripts

brussens: It's just that when I needed to change the value in the map, I realized that I couldn't change it, because the map is private.

BarbzYHOOL:
Good

PR #1516, opened 2019-02-22
feat(CI): add commit hashes to the automatic SQL import

### LINKED ISSUE #1449: Improve auto import messages for SQL scripts
Currently, the auto import script puts as title "Import pending SQL update file" which makes difficult to find an specific update using the History view in the data/sql folders.

Suggestion: Add the original commit title or hash so it is easier to explore. Example: "Import pending SQL update file from 694b406b8bf339255aeaf1481da027339a9b93e0"

captura

Stoabrogga:
##### CHANGES PROPOSED:
Add commit hashes to the automatic SQL import.

###### ISSUES ADDRESSED:
Closes #1449 

##### TESTS PERFORMED:
Tested on Ubuntu 16.04
(sadly, I don't know how to test the Travis run itself)

- clone azerothcore-wotlk
- reset to commit 38830f29c18159327707ff9d40cc97306b588c6a (this commit contains 3 pending SQL files)
 ```git reset --hard 38830f29c18159327707ff9d40cc97306b588c6a```
- manually apply the PR changes to ```apps/db_pendings/import.sh```
- execute ```source bin/acore-db-pendings```
- execute ```echo "$COMMIT_HASH"```; result (please also take note of the leading blank!):
 ```  275dbec76d023da6824e44c725813f10802287f8 6a53a73c1b235b111194e44b0dff5e6e6fda8927 38830f29c18159327707ff9d40cc97306b588c6a```

The Travis configuration is changed to also call ```source acore-db-pendings```, so the variable should be set for the commit comment when importing the SQL files.

none

##### Target branch(es):
Master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
   # push changes to git if any
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then git fetch --unshallow; fi
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then git checkout $TRAVIS_BRANCH; fi
-  - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ -n "$GITHUB_API_KEY" ]]; then git add -A . && git diff --cached --quiet || git commit -am "Import pending SQL update file (commit$COMMIT_HASH)" && git push https://$GITHUB_API_KEY@github.com/$TRAVIS_REPO_SLUG.git $TRAVIS_BRANCH; fi
+  - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ -n "$GITHUB_API_KEY" ]]; then git add -A . && git diff --cached --quiet || git commit -am "Import pending SQL update file..." -m "Referenced commit(s):$COMMIT_HASH" && git push https://$GITHUB_API_KEY@github.com/$TRAVIS_REPO_SLUG.git $TRAVIS_BRANCH; fi
   # sync staging with master
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ "$TRAVIS_BRANCH" = "master" ] && [[ -n "$GITHUB_API_KEY" ]]; then git fetch origin staging:staging && git checkout staging && git merge --no-edit master && git push https://$GITHUB_API_KEY@github.com/$TRAVIS_REPO_SLUG.git staging; git checkout master; fi
   - source ./apps/ci/ci-before_install.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -24,12 +24,12 @@ before_install:
...
+  - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ "$TRAVIS_BRANCH" = "master" ]; then source acore-db-pendings; fi
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then cd ..; fi
   # push changes to git if any
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then git fetch --unshallow; fi
   - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then git checkout $TRAVIS_BRANCH; fi
-  - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ -n "$GITHUB_API_KEY" ]]; then git add -A . && git diff --cached --quiet || git commit -am "Import pending SQL update file" && git push https://$GITHUB_API_KEY@github.com/$TRAVIS_REPO_SLUG.git $TRAVIS_BRANCH; fi
+  - if [ "$TRAVIS_BUILD_ID" = "1" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ -n "$GITHUB_API_KEY" ]]; then git add -A . && git diff --cached --quiet || git commit -am "Import pending SQL update file (commit$COMMIT_HASH)" && git push https://$GITHUB_API_KEY@github.com/$TRAVIS_REPO_SLUG.git $TRAVIS_BRANCH; fi
```
Deku: This could lead to an extra large title when there are several commits being imported, maybe commit references should go in the description section of the commit adding a second "-m" parameter, like:

`git commit -am "Import pending SQL update file..." -m "Referenced commit(s): $COMMIT_HASH"`

apps/db_pendings/import.sh
```diff
@@ -97,6 +99,7 @@ function import() {
                 echo "DROP PROCEDURE IF EXISTS \`updateDb\`;" >> "$newFile";
             fi;
 
+            COMMIT_HASH="$COMMIT_HASH $(git log --diff-filter=A "$entry" | grep "^commit " | sed -e 's/commit //')"
```
Deku: What if a commit included more than 1 revision? It will duplicate the hash in the list?

Stoabrogga:
@Deku 
Thanks for your comments. I pushed a new commit to address them:
- use ```sort -u``` to prevent duplicate commit hashes
- add the hashes to the commit message

Deku:
@Stoabrogga thanks for your quick response! Sadly, maybe I expressed myself incorrectly. When I refered to "revision" I meant "SQL script revision", as in a commit including many pending SQL files, so with the addition of `sort -u` that problem will not be addresed.

To solve this, you could add at the end of the for-loop in `import.sh` the following which will clean the COMMIT_HASH variable:

`COMMIT_HASH=$(echo $COMMIT_HASH | xargs -n1 | sort -u | xargs)`

An example of this:
imagen

Stoabrogga:
@Deku 
Thanks, you expressed yourself well, the fault was on my side, did not think this through completely. I pushed another commit to correct this, but used pattern matching instead of xargs. Tested it using commit 4158d2a0d8a38286eb3c4ff38e147a99252f6f07.

BarbzYHOOL:
i thought it would be in the commit's title

Deku:
@BarbzYHOOL from my reviews

> This could lead to an extra large title when there are several commits being imported...

PR #1547, opened 2019-03-02
Alert message on new account when password is too long

### LINKED ISSUE #1545: Password must be limited and we should use only latin
https://github.com/azerothcore/azerothcore-wotlk/blob/9643ccdb9d8f0321f54482af3de0aa35e1b14dbe/src/server/game/Accounts/AccountMgr.cpp#L21

our account creation is quite "buggy", world of warcraft client doesn't accept too long passwords

Currently we don't have any check on password and we accept UTF-8 characters directly

On trinitycore it has been already fixed: https://github.com/TrinityCore/TrinityCore/blob/951f130f20ef6e023c5d5c220eefb0ed9544ddef/src/server/game/Accounts/AccountMgr.cpp#L50

We can use such code

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70817133-password-must-be-limited-and-we-should-use-only-latin?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Ercules76:
##### CHANGES PROPOSED:

-  added error message in the database
-  entered password length control in the core

###### ISSUES ADDRESSED:

Closes 

#1545 

##### TESTS PERFORMED:

Compiled and tested, if password is too long the console show the message and no account are created

go on worldserver cli and create new account with this command:
account create account_name password_more_than_16_char

- [ ]
- [ ] 

##### Target branch(es):

Master


comix1988:
I have tested this PR and it works correctly:
before PR
acc2

after PR
acc1

good job @Ercules76

FrancescoBorzi:
good job @Ercules76 and @comix1988 !

BarbzYHOOL:
o_0 the error message was already there before though, I tested this 2 weeks ago.

I don't really understand what is being fixed here.

Note: the issue has not been closed

comix1988:
@BarbzYHOOL look at my first screen, no error is reported.

BarbzYHOOL:
yes, well I think it works fine with the command ".account modify password" though

FrancescoBorzi:
Good catch @BarbzYHOOL 

Confirmed. The command `account set password newpass newpass` still allows you to introduce a long string, which will not be allowed by the client:

image

This should also fixed. @Ercules76 would you like to take care of this?

BarbzYHOOL:
but shin, can you tell me why I already have seen this message on my SUPER OUTDATED core? the message "can not be longer than 16 characters (client limit)". Is it a new bug or is it also somewhere else?


I even talked about this with you recently (you know when I set a long password for the test accounts lol)

PR #1549, opened 2019-03-02
feat(core): allow to debug KickPlayer()

FrancescoBorzi:
##### CHANGES PROPOSED:

There are many places in the core where the `KickPlayer()` is called. This PR introduces a way to easily debug and understand where the kicking is coming from, even in environment where traditional debuggers are not available (e.g. Docker).

##### TESTS PERFORMED:

Build, runs, the debug log works.

The impact of this change is minimal and does not require further testing.

##### Target branch(es):

Master


src/server/game/Server/WorldSocket.cpp
```diff
@@ -119,8 +119,9 @@ bool WorldSocket::IsClosed(void) const
     return closing_;
 }
 
-void WorldSocket::CloseSocket(void)
+void WorldSocket::CloseSocket(std::string const& reason)
 {
+    sLog->outDebug(LOG_FILTER_CLOSE_SOCKET, "Socket closed because of: %s", reason.c_str());
```
Deku: Check if a reason was given first

Viste: +

FrancescoBorzi: @Deku @Viste thanks, please check it now

PR #1553, opened 2019-03-03
feat(cpp): enable C++17 support

FrancescoBorzi:
##### CHANGES PROPOSED:

-  Enable C++17 support

##### TESTS PERFORMED:

Runs, builds, tested in game (entering the world, try a few spells, everything looks normal).

- Tested on Ubuntu 18.04 with gcc
- Tested on Ubuntu 18.04 with clang

- Clean build cache first
- Just compile and have a quick look around the world and see if everything works as intended
-  Also, we need someone to test this on Windows

- [ ] Test on Windows (on traditional setup)

##### Target branch(es):

Master

### DIFF
--- a/src/cmake/compiler/clang/settings.cmake
+++ b/src/cmake/compiler/clang/settings.cmake
@@ -16,5 +16,5 @@ endif()
 
 # -Wno-narrowing needed to suppress a warning in g3d
 # -Wno-deprecated-register is needed to suppress gsoap warnings on Unix systems.
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11  -Wno-narrowing -Wno-deprecated-register")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17  -Wno-narrowing -Wno-deprecated-register")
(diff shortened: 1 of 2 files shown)


FrancescoBorzi:
@Naios any hint here?

Viste:
try to test it on CentOS today or tomorrow and then add review

Naios:
@FrancescoBorzi Probably the preferrable way nowadays is to use platform independent cmake feature flags instead as we [did on TC](https://github.com/TrinityCore/TrinityCore/blob/master/cmake/macros/ConfigureBaseTargets.cmake#L20-L34). Also `-std=c++1z` refers to the in progress version and `-std=c++17` should be used in order to avoid toolchain specific differences.

FrancescoBorzi:
@Naios yes, I've initially set `c++17` but that fails the build on clang 3.8. Apparently `-std=c++1z` works for both versions:

> You can use Clang in C++17 mode with the -std=c++17 option (use -std=c++1z in Clang 4 and earlier).

https://clang.llvm.org/cxx_status.html

It should still be better than using `c++11` I gues

Nefertumm:
Tested on windows, compiled normally and everything seems like always.

Deku:
@Nefertumm I think the merge was hasty. We should have taken into consideration what @Naios said.

FrancescoBorzi:
@Deku he needed it for another PR, we can always change it later

PR #1556, opened 2019-03-05
feat(Core/DB): Automatic DB updater.

Nefertumm:
##### CHANGES PROPOSED:

-  Add an automatic updater so we don't have to execute manually the updates each time a commit introduces a new sql file.
-  Cherrypicked from Oregon-Core. Adapted by me. (All the credits go for them though)

###### ISSUES ADDRESSED:

None

##### TESTS PERFORMED:

It does compiles, it doesn't work yet.

Not ready to be tested yet.

- [ ] Fix content load from the file.

##### Target branch(es):

Master

### DIFF
--- a/src/common/Database/DatabaseWorkerPool.h
+++ b/src/common/Database/DatabaseWorkerPool.h
@@ -20,6 +20,13 @@
 #include "QueryHolder.h"
 #include "AdhocStatement.h"
 
+#include "CompilerDefs.h"
+#include <iostream>
+#include <fstream>
+#if PLATFORM == PLATFORM_UNIX
+#include <sys/file.h>
+#endif
+
 #define MIN_MYSQL_SERVER_VERSION 50100u
 #define MIN_MYSQL_CLIENT_VERSION 50100u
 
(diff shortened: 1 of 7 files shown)


src/common/Database/DatabaseWorkerPool.h
```diff
@@ -469,6 +476,102 @@ class DatabaseWorkerPool
...
+#if PLATFORM == PLATFORM_UNIX
+                flock(fileno(fp), LOCK_SH);
+#endif
+                //------
+
+                struct stat info;
+                fstat(fileno(fp), &info);
+
+                // if less than 1MB allocate on stack, else on heap
+                char* contents = (info.st_size > 1024 * 1024) ? new char[info.st_size] : (char*)alloca(info.st_size);
```
Nefertumm: Maybe the problem is here. It fills all the contents's size with thrash, and then fread replace it, but not all of it. [You can see it here](https://media.discordapp.net/attachments/262884535694393344/552318868542980126/unknown.png?width=1154&height=702)

Rochet2: maybe use `sizeof(char)` as size and `info.st_size/sizeof(char)` as count for fread. Then the return value will tell you how many characters were read and you can use that as the parameter to mysql_real_query instead of `info.st_size`

src/common/Database/DatabaseWorkerPool.h
```diff
@@ -469,6 +476,102 @@ class DatabaseWorkerPool
...
+            if (mysql_real_query(handle, "START TRANSACTION", sizeof("START TRANSACTION") - 1))
+            {
+                sLog->outErrorDb("Couldn't start transaction for db update file %s", file);
+                return false;
+            }
+
+            bool in_trasaction = true;
+            bool success = false;
+
+            if (FILE* fp = ACE_OS::fopen(file, "rb"))
```
FrancescoBorzi: I think we should avoid adding more code that depends on ACE (ideally we should one day get rid of ACE since it can just be replaced with modern C++). I'm sure there are plenty of ways to read files with modern C++ without using ACE or any library. The same apply to all other parts where ACE is used in this PR.

Yehonal: indeed

Nefertumm:
Now it works (the main function at least...). When it tries to execute the query, it throws a db error (maybe because of this DELIMITER // )

FrancescoBorzi:
Maybe @Naios can give a hint about the delimiter issue? (he's the author of the DB updater on TC)

Naios:
The reason why TC doesn't apply the sql updates directly and uses the `mysql` binary for this directly is that the mysql connector behaves in a strange way when applying large batches of multiple queries.
We didn't look for the exact reason back in that time because also other features of the binary made us to use it.
Thus I would not recommend to do it that way because it could lead to inconsistencies in databases.

Nefertumm:
Thanks Naios. Is there a way to make it that way without using boost so i can work on it?

Naios:
In theory you can replace the boost calls by any other library but boost process is very flexible and I think there is no reasons not to use it, also the TC system in theory can be ported over too when sticking to the GPL license.

FrancescoBorzi:
About the license, AzerothCore inherits the GPLv2 license from MaNGOS/TrinityCore and also adds the AGPLv3 for new sources, to prevent privates modifying the core source code in any way without contributing back. To add private customizations they can always use [modules](http://www.azerothcore.org/wiki/Create-a-Module). So yes, we stick to GPL and expect our users to do the same.

Viste:
When i try oregoncore last time i fuckedup with they dbupdater so i disagree with porting it

Nefertumm:
I will rework it to make it like trinity, but i dont know if we should use boost.

Kitzunu:
As I understand it, this type of automatic DB updater is scrapped so I will close this as it has also been stale for over a year :)

Yehonal:
To give my 2 cents on it:

I think that for now it's better to include the db_assembler update process inside your updating process. It's safe to run it everytime after built the core for example.

PR #1558, opened 2019-03-06
fix(Docker): allow to use modules

FrancescoBorzi:
##### CHANGES PROPOSED:

-  Include modules in Docker build

##### TESTS PERFORMED:

Not yet tested.

Include a module in your sources as you would normally do in a traditional setup. Then run the docker setup and see if the module is correctly included.

##### Target branch(es):

Master


sirbao:
Just finished docker build, still no modules activ/working, the only difference so far, was during the docker build process the modules folder seems to be included see screenshot
finished docker build

FrancescoBorzi:
@sirbao thanks. Did you clean the build cache before compiling?

sirbao:
@FrancescoBorzi Yes i did clean the cache before compiling.

koffeine:
Adding `ADD modules /azerothcore/modules` line to `docker/build/Dockerfile` was not enough for me, I also had to remove `modules/*` line from `.dockerignore` file.

After building with these two changes modules were working in Docker.

FrancescoBorzi:
many thanks @koffeine !

PR #1567, opened 2019-03-11
Core/Ticket System: Ticket comment visual output bug 

Viste:
##### CHANGES PROPOSED:
-  Commenting on a ticket no longer sends the GM's name to the ticket handler for assignment

Original author: Tomatodev


src/server/scripts/Commands/cs_ticket.cpp
```diff
@@ -192,7 +192,10 @@ class ticket_commandscript : public CommandScript
         ticket->SaveToDB(trans);
         sTicketMgr->UpdateLastChange();
 
-        std::string msg = ticket->FormatMessageString(*handler, nullptr, ticket->GetAssignedToName().c_str(), nullptr, nullptr);
+        std::string msg = [&] {
```
jameyboor: There's no reason to use a lambda here.

Viste: why? Explain your opinion plz with args

jameyboor: Why would you in the first place? If you want to have an enclosed scope then just make an enclosed scope. By doing this you're creating unnecessary clutter and a lambda that has to catch every local by ref.

Viste: ok its good arg i agree, but as you can see in main post i just port it from TC and not me original author.
Proof: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Commands/cs_ticket.cpp#L207

jameyboor: https://github.com/TrinityCore/TrinityCore/commit/29826a92ef4cf7896e3be69757b015d47f3db006#diff-98d8f29d565bf98320917c18fbd1a9b8

This seems to be the commit. That commit is bad as well. Just because TC has it doesn't mean it should be ported like that. That original commit is bad code.

Viste: rewrite it as soon as i can

Helias:
**Before PR:**
when run .ticket comment ID_TICKET  COMMENT, in the chat appears "ticket assigned to Helias comment:"

**After PR**
when run .ticket comment ID_TICKET  COMMENT, in the chat appears "ticket Helias comment:" (without the assigned to, indeed because the ticket is not assigned when you comment a ticket).

So it works!

PR #1586, opened 2019-03-18
 feat(Core/Chat): Add config option to allow use LFG anywhere not only in city

Viste:
##### CHANGES PROPOSED:
-  Add config option to allow use LFG anywhere not only in city


FrancescoBorzi:
nice, good job @Viste

Helias:
Tested, it works

Issue #1496, opened 2019-02-19
Alterac DB errors


Mitradis:
##### SMALL DESCRIPTION:
Alterac Data Base errors

##### STEPS TO REPRODUCE THE PROBLEM:
1. Enter to Alterac
2. See DBErrors.log and Server.log
3. 

##### BRANCH(ES):
AzerothCore rev. 23d78ce6bc75 2019-02-18 18:22:36 +0700 (master branch) (Win64, Release)

##### EXTRA NOTES:
```
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 302, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 303, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 304, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 306, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 307, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 308, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 309, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 310, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 311, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 312, GUID: 0) not found for BG (map: 30, instance id: 1)!
2019-02-19 13:49:34 ERROR: Battleground::GetBGCreature: creature (type: 313, GUID: 0) not found for BG (map: 30, instance id: 1)!
[...]

PR #1587, opened 2019-03-18
Core/Battlegrounds: Fix log errors about creatures not found

grimgravy:
##### CHANGES PROPOSED:

-  Move "Battleground::GetBGCreature: creature not found for BG" error log to "GetBgMap" level when the caller expects no creatures to be found, like on Battleground startup.
-  remove wrong DB errors.
It is normal to have events applicable only to certain instance difficulties.

###### ISSUES ADDRESSED:
I also saw the problem not only in the alterac valley, also on the island of conquest, but able to fix the repair based on trinitycore

Closes [#1496](https://github.com/azerothcore/azerothcore-wotlk/issues/1496)

##### TESTS PERFORMED:
windows 7 operating system


FrancescoBorzi:
@grimgravy can you please provide instructions for testers? I guess we need to make sure that BGs still work fine. Maybe you can provide a list of what can be affected by this change i.e. what testers should double check

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -4168,13 +4168,11 @@ void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTrigger c
 {
     if (e.empty())
     {
-#if defined(ENABLE_EXTRAS) && defined(ENABLE_EXTRA_LOGS)
```
Stoabrogga: I think this should not be deleted as it controls debug log entries.

grimgravy: I need to do this so travis can release the checks

Stoabrogga: I don't understand, what do you mean by that?

Stoabrogga: Ok, I understand. Parameter "at" is unused if extra logging is disabled and the compiler complains about this. But this has to be resolved in another way, not by deleting the preprocessor directives, as you will get debug messages even when disabling extra logs.

grimgravy: So, I will revert to original changes.

Stoabrogga: You could add
```(void)at;```
at the beginning of the function. This way the compiler sees that the parameter is referenced and won't complain any more.

Comment (grimgravy) [reply src/server/game/AI/SmartScripts/SmartScript.cpp]: Thanks for the help.

**Comment by FrancescoBorzi** [comment]
How to test this?

**Comment by grimgravy** [comment]
@FrancescoBorzi, you can rest easy, there is no problem in the operation of bg. For if you hover problems post here

**Comment by FrancescoBorzi** [comment]
@grimgravy sorry I don't understand

**Comment by grimgravy** [comment]
@FrancescoBorzi, bgs are working normally, do not worry if there are problems post here.
Sorry, I use to translate the English the google translator, sometimes give problem to translate

File: src/server/game/AI/SmartScripts/SmartScript.cpp [hunk 2]
```diff
@@ -4197,10 +4197,6 @@ void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTrigger c
         }
         mEvents.push_back((*i));//NOTE: 'world(0)' events still get processed in ANY instance mode
     }
-    if (mEvents.empty() && obj)
-        sLog->outErrorDb("SmartScript: Entry %u has events but no events added to list because of instance flags.", obj->GetEntry());
-    if (mEvents.empty() && at)
-        sLog->outErrorDb("SmartScript: AreaTrigger %u has events but no events added to list because of instance flags. NOTE: triggers can not handle any instance flags.", at->entry);
```
FrancescoBorzi: why removing those?

grimgravy: This was removed by trinitycore. The SAI is correct, the errro information was not to be shown on the console [1496](https://github.com/azerothcore/azerothcore-wotlk/issues/1496)

FrancescoBorzi: @grimgravy thanks for the answer

FrancescoBorzi:
I guess a test for this would be a quick check that BGs still work properly

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -4175,7 +4175,7 @@ void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTrigger c
         if (at)
             sLog->outDebug(LOG_FILTER_DATABASE_AI, "SmartScript: EventMap for AreaTrigger %u is empty but is using SmartScript.", at->entry);
 #endif
-        return;
+        return (void)at;
```
Nefertumm: Why? Just leave it as `return;`, don't cast it

Viste:
@grimgravy use deepl.com he translates better then google

src/server/game/AI/SmartScripts/SmartScript.cpp
grimgravy: have a better idea. Post here

Stoabrogga: Hmm, would be better to remove it here and add it at the beginning of the function with an appropriate comment, e.g.:
```cpp
void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTrigger const* at)
{
    (void)at; // ensure that the variable is referenced even if extra logs are disabled in order to pass compiler checks

    if (e.empty())
    {
```

poszer:
**Tested without this commit :**
There are errors, which spam WC all the time while in the Battleground ( Alterac Valley )
Image below :
1. 
Before
2.
Before1

**Test with this commit :**
1. The modifications are applied correctly, without issue.
2. ========== Build: 5 succeeded, 0 failed, 10 up-to-date, 1 skipped ==========
3. Tested on OS: Win
4. Tested in game, everything is fine, without issues/error in World/ DDerrors.log.

GitVerge:
This should be fixed at Battleground::GetBGCreature level to decide whether or not you want to log this error. Just avoiding this function to do exactly the same without the log doesn't feel correct.

An optional param for this function to log or not would be better (true as default) ^^

FrancescoBorzi:
good job @grimgravy and @poszer

Issue #1497, opened 2019-02-19
Еndless enchantments


Mitradis:
##### SMALL DESCRIPTION:
Еndless enchantments.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
After using must be disappear.

##### CURRENT BEHAVIOUR:
Can be used infinity.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Using items:
44874
44872
44873
44871
44879
44878
44877
44876

##### BRANCH(ES):
AzerothCore rev. 23d78ce6bc75 2019-02-18 18:22:36 +0700 (master branch) (Win64, Release)

poszer:
Closing this.  I can't understand the issue reported. 
Next time provide more details, how we can reproduce this problem. 
Anyway, thanks for report @Mitradis

Helias:
Confirmed, they don't disappear

poszer:
Ok, know what's the problem. Will post fix soon

poszer:
@Mitradis  Could you please test the referenced PR.

Mitradis:
yes #1590 is same, this thread can be close

poszer:
@Mitradis Did it fix your opened issue?

poszer:
You provide wrong ID's. All of them are marked as '' This item is not available to players.'' . 
Use other ID's for correct Enchantment.
Closing this

PR #1590, opened 2019-03-20
DB: Item Enhancements (Permanent)

poszer:
Item Enhancements (Permanent) issue Fix

##### CHANGES PROPOSED:

-  Added spell charges
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1497

##### TESTS PERFORMED:

Tested, the query is updated without errors. Tested in game, all of items are available now for use just once, after used they re removed from the bags which is correct.

1. Create any character ( Level 80 ) 
2. Equip Head & Shoulder ( For level 80 )
3. Use .add command with next Id's and try to enchant your head & shoulder :

- 44874
- 44872
- 44873
- 44871
- 44879
- 44878
- 44877
- 44876

The item's should be removed from bag while enchancement applied.

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1553042372568459200.sql
+++ b/data/sql/updates/pending_db_world/rev_1553042372568459200.sql
@@ -1,5 +1,6 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1553042372568459200');
 
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44875;
 UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44874;
 UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44872;
 UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44873;
--- a/data/sql/updates/pending_db_world/rev_1553042372568459200.sql
+++ b/data/sql/updates/pending_db_world/rev_1553042372568459200.sql
@@ -1,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1553042372568459200');
+
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44874;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44872;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44873;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44871;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44879;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44878;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44877;
+UPDATE `item_template` SET `spellcharges_1`='-1' WHERE `entry`=44876;


Stoabrogga:
Were can the items be acquired as a player? According to wowhead they are not available to players, e.g. "Greater Inscription of the Storm":
https://www.wowhead.com/item=44874
Players can obtain this from Lillehoff if they are exalted:
https://www.wowhead.com/item=44135

poszer:
> I tested it, it works, but add the charge -1 property also to 44874, it should be the same kind of item.

Hmm, it already have charge -1. 
Also, nevermind where are these items obtainable, they should be used just once and destroyed from the bag. That's the point of this, without this you can enchant for example Head x20 times with 1x item of these listed enchantmens which is totally wrong. Btw for this with Exalted reputation, the id is 50338.
Every of these enchancements have 3 types of items :
1. Can be obtainable For the reputation 
2. Can be obtainable from the Vendors who's selling items which are Bind to Account ( Eg. Wintergrasp Vendors for the Shards... Top-level helm and shoulder faction-related enchants are now available as Bind-on-Account items that do not require any faction to use once purchased (they still require the appropriate faction level to purchase).)
3. Can be obtainable Normally from the Vendors in City ( Selling from the vendor )
The changed id's are these who are obtainable from the Vendors and which was wrong, didn't destroyed on use. Other are Correct and that's why they re not changed ( Other 2 types of these items are destroyed while used '' Bind to Account and Reputation items ).

Stoabrogga:
@poszer 
I don't think you are right about this. Item 50338 is only shown if your reputation is below exalted (you cannot buy it), after reaching exalted item 44135 is shown instead, which you can buy then. I think this was originally done because with patch 3.3 the reputation restrictions of these items were removed to only need one character grind to exalted (they are account bound after all). The items listed in this PR were originally test items I presume, so it doesn't harm if they are changed.

poszer:
> @poszer
> I don't think you are right about this. Item 50338 is only shown if your reputation is below exalted (you cannot buy it), after reaching exalted item 44135 is shown instead, which you can buy then. I think this was originally done because with patch 3.3 the reputation restrictions of these items were removed to only need one character grind to exalted (they are account bound after all). The items listed in this PR were originally test items I presume, so it doesn't harm if they are changed.

@Stoabrogga 
Maybe, but anyway they should have 1 charge just, that's correct. So if it looks how you said that 50338 is later changed to 44135 it will have 1 charge, right? That is also okay.
But, just tested that with Faction Exalted, it doesn't change nothing 50338 staying 50338 : 
hodir

Stoabrogga:
@poszer 
No, that was not what I meant. I meant Lillehoff's inventory. If your reputation is below exalted you will see item 50338 for sale, after reaching exalted you don't see 50338 anymore but 44135 instead. I think this was introduced to give players a clue what items are available and how to achieve them and afterwards provide them with an item which has no reputation restrictions anymore, so they can send it to another one of their characters.

poszer:
> @poszer
> No, that was not what I meant. I meant Lillehoff's inventory. If your reputation is below exalted you will see item 50338 for sale, after reaching exalted you don't see 50338 anymore but 44135 instead. I think this was introduced to give players a clue what items are available and how to achieve them and afterwards provide them with an item which has no reputation restrictions anymore, so they can send it to another one of their characters.

@Stoabrogga 
I understand, the topic has gone to other side. I want to say you, without this commit, when you get any of that enchant items, you re able to apply them xxx times. That's not correct. All of them should have just 1 charge, and to be applied on 1 item just. You can try it, just get exalted repo for example, use this item which is in our conversation, buy it from that vendor, apply it. It's still in your bag.
Than create new character , include this commit, buy item and apply it, it will be removed from your bag.

Stoabrogga:
@poszer 
Item 44135 has only 1 charge, even without this PR (it is not affected by this PR at all). The same for 50338, which cannot be acquired by players, only seen in the vendor inventory of Lillehoff while not being exalted with the Sons of Hodir.

The items which are affected by this PR (44875,44874,44872,44873,44871,44879,44878,44877,44876) cannot be obtained by players. Just look for yourself, there is no reference in the DB, only in table "item_template". Those items are not included in-game. So it doesn't matter if the charges for these items are modified, as only game masters can get them, e.g. via command "additem".

poszer:
> ,44873

@Stoabrogga 
Oh, i see now on wowhead. They re removed :laughing:  This item is not available to players.
Thank you, i'll close than this PR, and also the issue opened about these enchantments, since it's not valid.

PR #1602, opened 2019-03-21
 Fix(Core/Player): fix periodic eating and drinking emotes

Viste:
##### CHANGES PROPOSED:
-  Fixes drinking and eating periodic emotes. Blizzlike timings, blizzlike first timer. So ye. This is fixed. Mangos did it wrong, we did it right. Sniffs confirm it.
-  you may now feel like on retail when pointlessly drinking or eating stuff
thx to Ovahlord


comix1988:
I don't understand what this fix changes @Viste

Viste:
@comix1988 this make emotes of eat and drink blizzlike, Now the interval between the character bringing food or water to himself is the right one.

comix1988:
so it's just a visual change?

thanks for the explanation

Viste:
yes

sucofog:
It seems right, and tested, but the thing is on retail, the character brings food and drink back to himself after each timing, is it happening only to me?

Viste:
according to @poszer comment "After ( With this commit ) ; Visual is shown, The player sit drinking / eating which is correct totally."
yes its only you @sucofog

NixAJ:
Just adding a comment here.

I tried to start a discussion on the discord about this and promised Shin to start an issue on here, which I will in time.

https://github.com/TrinityCore/TrinityCore/pull/22380
This pull requests has some interesting information commented by xvwyh detailing information sniffed from the alpha. While alpha data shouldn't be taken as confirmed, there are a lot of evidence that point towards blizzard having 5 second internal unit timer acting as a heartbeat for trigger a multitude of different processes including but not limited to eating / drinking emotes, Spline syncing and certain Aura procs.

It might be incorrect to simply implement a timer specific for eating / drinking emotes, and potentially we should look into implementing a timer that handles the appropriate things?

pklloveyou:
Not working
Did I miss the update?

pklloveyou:
TIM截图20190323232825
There you go！！！！！！！

Viste:
@pklloveyou i retested now and all works fine

PR #1624, opened 2019-03-25
Eluna RegisterPlayerEvent 5 (PLAYER_EVENT_ON_SPELL_CAST)

wizzymore:
##### CHANGES PROPOSED:

Thanks to : @NixAJ  ( Now i know it ;) )

-  PLAYER_EVENT_ON_SPELL_CAST is now called when spell is casted

###### ISSUES ADDRESSED:

Closes 

##### TESTS PERFORMED:

- [x] Linux
- [x] Windows

You need to have Eluna Module
Create a test lua file
Insert this ( as an example )
```lua
local function CastSpell(event, player, spell, skipCheck)
	print(player:GetName() .. " casted spell id " .. spell:GetEntry());
end

RegisterPlayerEvent(5, CastSpell)
```

- [x] Test this on windows
- [x] Check if it slows down the server if there are more players online

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -4,6 +4,11 @@
  * Copyright (C) 2005-2009 MaNGOS <http://getmangos.com/>
  */
 
+#ifdef ELUNA
+#include "LuaEngine.h"
+#include "ElunaUtility.h"
+#endif
+
 #include "Common.h"
 #include "DatabaseEnv.h"
 #include "WorldPacket.h"
@@ -3647,7 +3652,9 @@ void Spell::_cast(bool skipCheck)
     {
         // now that we've done the basic check, now run the scripts
         // should be done before the spell is actually executed
-        //sScriptMgr->OnPlayerSpellCast(playerCaster, this, skipCheck); // pussywizard: optimization
+#ifdef ELUNA
+        sEluna->OnSpellCast(playerCaster, this, skipCheck);
+#endif
 
         // As of 3.0.2 pets begin attacking their owner's target immediately
         // Let any pets know we've attacked something. Check DmgClass for harmful spells only


poszer:
Will do test tomorrow for Win :+1:

Viste:
> Check if it slows down the server if there are more players online
it can may affect only when online rise more then 3000, but I don't have to at all, because it's not used.

wizzymore:
Didn't mean to close this, my bad

BarbzYHOOL:
it's NixAJ on github

FrancescoBorzi:
Good job @wizzymore @nixAJ

NixAJ:
Perhaps it would be a smart idea to create a hook for OnPlayerCastSpellById. It should reduce any lag that may have been an issue before.

wizzymore:
So, i checked the core and i don't seem to find a way to make this easyer on the server side and then i realised something, the only thing that will be executed will be the eluna function wich is not consuming memory ( i mean is consuming way less memory ) as the server already reports those spells being casted and they are already registered there.
I don't know how to explain myself but adding this function there will not slow anything at all because it will not add consuming memory thing like checking for spells in a big way, only reporting to lua that the spell has been casted.

Tryed it with 100 Player Bots and in the memory informations there is an increase so small that it will not be feeled by the server ( almost 1 mb of memory all the time ).
The 1mb of memory will be consumed only to the ones who have the eluna module activated.

If you didn't understand what i said, no problem, neither i partially so you can ask me anything.

BarbzYHOOL:
@Rochet2 @Kaev opinion, pros? :D

Kaev:
> 
> 
> @Rochet2 @Kaev opinion, pros? :D

The change itself shouldn't be a problem. But adding a badly written script to the core with this hook could be a huge performance impact. Handle with care.

wizzymore:
@Kaev how should i do this than? When should i insert the check for the eluna script? When to catch the spell being casted?

Kaev:
> 
> 
> @Kaev how should i do this than? When should i insert the check for the eluna script? When to catch the spell being casted?

The hook itself is fine. I just mentioned that people who use it should be careful with it.

BarbzYHOOL:
No need to test the hook much I guess it can be merged

Veraciel:
I have a same issue with the core I downloaded 2 weeks ago. Is this fixed?

BarbzYHOOL:
So do we merge @FrancescoBorzi ?

FrancescoBorzi:
I'd say yes

BarbzYHOOL:
thank you @wizzymore

PR #1626, opened 2019-03-25
Limiting password size

wizzymore:
##### CHANGES PROPOSED:

-  Limit password size when changing username and when changing the password

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1545

##### TESTS PERFORMED:

- [x] Linux
- [x] Windows

`account create testaccount 1234567890123456789`
It should return the error
`account create testaccount 123`
`account set password testaccount 1234567890123456789 1234567890123456789`
it should return the error again

Error that will be shown:
```
AC> account set password admin 1234567890123456789 1234567890123456789
AC>Your password can't be longer than 16 characters (client limit), password not changed!
```

##### Target branch(es):

Master


Helias:
OS: Ubuntu 18.04

I don't get the error message when I use set account, here the screen:
immagine

(The password didn't change, but the error message is different)

wizzymore:
> OS: Ubuntu 18.04
> 
> I don't get the error message when I use set account, here the screen:
> immagine
> 
> (The password didn't change, but the error message is different)

*face palm*
The command i used as an example was wrong, test it again @Helias

Helias:
Sorry, I did it very quickly, just copy and paste :laughing: 

I tested again, it works very well

BarbzYHOOL:
Thank you @wizzymore !

PR #1638, opened 2019-03-26
chore(Docker): add comments to explain the building steps

FrancescoBorzi:
Just adding comments, nothing to test here.


wizzymore:
https://github.com/azerothcore/azerothcore-wotlk/blob/master/pull_request_template.md cryes for not being used :-1:

FrancescoBorzi:
@wizzymore you're right, but for such tiny and obvious changes we can make an exception... can't we? :)

wizzymore:
Just joking, thanks for the comments as i now understood what docker is doing

PR #1664, opened 2019-04-01
[Worldserver] Increased limit of realms

### LINKED ISSUE #1403: Can't set realmID > 8 (in worldserver.conf)
##### SMALL DESCRIPTION:

We can't put a realmid above 8 in the worldserver.conf, else it returns "Realm ID not defined in configuration file"

We should be able to put realmid above 8, imagine if we have tons of realms or simply to sort them easily (dev realms on 10-20 IDs, testing realms on 20-30 etc...)

1. Change realmid to 9 in worldserver.conf
2. Can't launch worldserver
3. Cry

##### EXTRA NOTES:

This part is responsible (thanks @Rochet2 for finding it in 10 sec)
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/worldserver/Master.cpp#L472

This is related to the auth.account.online column. In that column, the realmid is put when you are online on that realmid. However it uses a flag, which doesn't seem to make much sense (since you can't be online on several realms at the same time with the same account).

Rochet suggests to either change this to a flat number. :

> simply scrapping the mask would already allow 255 realms or something

Or to change to uint 32 :

> And by searching realmID in the source code you can find where the mask is used
> For example there is this line of code LoginDatabase.DirectPExecute("UPDATE account SET online = online & ~(1<<(%u-1)) WHERE online & (1<<(%u-1))", realmID, realmID);

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

264dd0c65afc33e9613e42dee656697d088d8a74

Pondaveia:
### More than 8 realms available
Thanks @Rochet2 for all the help
##### CHANGES PROPOSED:

-  Altered the DB and the core to allow more than 8 realms.
With these changes the core will support until 255 available realms.
Due to the client only reading a package with a size of 8 bits, the core can't allow more realms to be added since any higher values won't be updated in the database. Otherwise, the server could behave abnormally because we won't be storing the current realm onto which a player is logged on.

###### ISSUES ADDRESSED:

Closes #1403 

##### TESTS PERFORMED:

I tested this on windows and it works.

1. Add a new realm on Auth Database -> Realmlist with ID 9 or higher (Also need diferent name)
2. Change the realm id on worldserver.conf to the new value that you added on the database
3. Launch worldserver and log onto the realm
4. Check worldserver for no query errors
5. On the auth database at accounts.online you will see the number of the realmID to which the player is connected to, in this case it should display 9.

##### Target branch(es):

Master


Rochet2:
If it wasnt done already,, you should change the usage of the online too.
Search for example for: `LoginDatabase.DirectPExecute("UPDATE account SET online = online & ~(1<<(%u-1)) WHERE online & (1<<(%u-1))", realmID, realmID);`

Pondaveia:
Alright but, why change it and what for?

Rochet2:
> Alright but, why change it and what for?

If you use realmid higher than 32 the online status will be wrong unless we change it -  I guess.
So either limit realmid to 32 or change the online to be the realmid that is online.

Example
```cpp
UPDATE ACCOUNT SET online = (1<<(33-1));
/* SQL Error (1264): Out of range value for column 'online' at row 1 */```

**Comment by Pondaveia** [comment]
I'm actually unsure on how to proceed with this one... I was looking at Trinity core and they have a strange approach, even their query is different. Should I limit this for now to 32 realms? How do I know if Trinity core has a X limit of realms? I mean the database can only go so far there's always a limit...

**Comment by Rochet2** [comment]
Seems to me like TC is using online as a boolean value (1/0). So the realmid is not limited.
Unsure if its possible to see which realm the player is logged into on TC though as it is on AC atm.

I would probably just set that online = realmid you are logged into. That way we can have as many realms as we want and we still know through DB which realm the player is online on.

**Comment by Pondaveia** [comment]
> Seems to me like TC is using online as a boolean value (1/0). So the realmid is not limited.
> Unsure if its possible to see which realm the player is logged into on TC though as it is on AC atm.
> 
> I would probably just set that online = realmid you are logged into. That way we can have as many realms as we want and we still know through DB which realm the player is online on.

Alright, then I'll just try to unmask the value and change the Update query. Cheers man!

File: src/server/worldserver/Master.cpp [hunk 1]
```diff
@@ -471,11 +471,21 @@ bool Master::_StartDB()
 
     ///- Get the realm Id from the configuration file
     realmID = sConfigMgr->GetIntDefault("RealmID", 0);
-    if (!realmID || realmID > 8) // pussywizard: above 8 spoils 8-bit online mask for the accounts
+    if (!realmID)
     {
         sLog->outError("Realm ID not defined in configuration file");
         return false;
     }
+    else if (realmID > 256)
```
BarbzYHOOL: see @Winfidonarleyan 's comment

BarbzYHOOL:
Ok so 0 and 256 don't work

IDs 1, 55 and 255 work fine

BarbzYHOOL:
If nobody has anything against it, it can be merged imo! ty @Pondaveia

PR #1673, opened 2019-04-03
Core/Commands: Quest Add

poszer:
Command Quest Add

##### CHANGES PROPOSED:

-  Do not allow to take same quest XX times
-  

###### ISSUES ADDRESSED:

As GM, you are able to add XX quests to character with command .quest add
And as example, if you add 10 quests, and you close one, the 9 remaining on your characters, Which isn't correct. The quest should be taken once, even with the command.

##### TESTS PERFORMED:

Builded without errorrs on WIn. Tested in game, while using command .quest add XX , character will receive quest. If you try again to add same quest with command, you can't and will receive msg that player is already on that quest.

First try without this commit : 
1. quest add 2930 
2. quest add 2930
3. quest add 2930 
4. Now, abandon that quest. 
5. There is still 2 remaining on character ( So you have to abandon it 2 more time )

With this commit : 
1. quest add 2930
2. quest add 2930 ( You can't add 2 same quests - Player will get msg '' You re already on that quest "
3. That is correct.

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/scripts/Commands/cs_quest.cpp
+++ b/src/server/scripts/Commands/cs_quest.cpp
@@ -67,7 +67,7 @@ public:
 		
 		if (player->IsActiveQuest(entry))
         {
-            handler->PSendSysMessage("You are already on that quest!");
+            handler->PSendSysMessage("This quest is already active!");
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


wizzymore:
Change the message to something like "This quest is already active" or " You already have this quest", makes more sense. But only if you want, tested the change and it works so it is okay. Nice work

poszer:
> Change the message to something like "This quest is already active" or " You already have this quest", makes more sense. But only if you want, tested the change and it works so it is okay. Nice work

@wizzymore Thanks, changed :+1:

BarbzYHOOL:
good job

Issue #1637, opened 2019-03-26
[Quest] Let Them Eat Crow  [Fix included]


Knindzagxg:
Let Them Eat Crow quest in Howling Fjord.

How to test:

- .quest  add 11227
- .tele HowlingFjord
- .additem 33238 5
-  Move to quest mark location
- Use provided quest item to spawn Hound

Once spawned, hound will remain there for, at least 20 minutes  (Prolly even more, but i stoped after 20th)

Let's  force  despawn after 5-10  seconds.

```sql
-- Creating a new SmartAI script for [Creature] ENTRY 23943 (name: Hungry Plaguehound)

-- Table creature_template
UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 23943;

-- Table smart_scripts
DELETE FROM `smart_scripts` WHERE (source_type = 0 AND entryorguid = 23943);
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(23943, 0, 0, 0, 1, 0, 100, 1, 5000, 10000, 0, 0, 41, 10000, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Hungry Plaguehound - Out of Combat - Despawn In 10000 ms (No Repeat)');
```

.reload  creature_template 23943
.reload smart_scripts

Try using quest item now.

wizzymore:
instead of this way of fixing, why don't you try to make it blizzlike?
- [ ] Make Plaguehound Cage create a guardian not a npc that doesn't move
- [ ] Make the guardian dissapear on quest complete

Knindzagxg:
Hmm, ill try to make it.
Had to make this fix quick, cause  someone could spawn tons of them and prolly crash server at the end.

BarbzYHOOL:
Some shit I had tried to fix long ago, it's INCOMPLETE:

```SQL
-- quest let them eat crow
-- https://github.com/TrinityCore/TrinityCore/pull/17948 c++
-- c++ to do

[...]

PR #1680, opened 2019-04-04
fix(Core/DB/Quest): Let Them Eat Crow

wizzymore:
##### CHANGES PROPOSED:

-  Fixing the quest

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1637

##### TESTS PERFORMED:

- [x] Windows
- [x] Linux (Docker)

.go c 107741
.quest add 11227
.additem 33238 5
Get to the quest marked location and use the item to summon the pet
Use the 5 items added to feed him
Complete the quest to see if he despawns

##### Target branch(es):

Master


Pondaveia:
You should change the queries from " to ' in order to keep consistency and to make sure the queries are recognized by most IDEs or Editors, but good job 👌I'll try to test later on

data/sql/updates/pending_db_world/rev_1554387633805928700.sql
```diff
@@ -0,0 +1,19 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1554387633805928700');
+
+DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId` IN (13, 17) AND `SourceEntry` IN (42767, 42788);
+INSERT INTO `conditions` (`SourceTypeOrReferenceId`, `SourceGroup`, `SourceEntry`, `SourceId`, `ElseGroup`, `ConditionTypeOrReference`, `ConditionTarget`, `ConditionValue1`, `ConditionValue2`, `ConditionValue3`, `NegativeCondition`, `ErrorType`, `ErrorTextId`, `ScriptName`, `Comment`) VALUES
+(13, 2, 42788, 0, 0, 31, 0, 3, 23943, 0, 0, 0, 0, "", "Spell 'Feed Plaguehound' targets NPC 'Hungry Plaguehound'"),
+(17, 0, 42788, 0, 0, 29, 0, 23943, 10, 0, 0, 0, 0, "", "Spell 'Feed Plaguehound' requires NPC 'Hungry Plaguehound' within 10 yards"),
+(13, 1, 42767, 0, 0, 31, 0, 3, 23945, 0, 0, 0, 0, "", "Spell 'Sic'em' targets NPC 'Fjord Crow'");
```
FrancescoBorzi: like @Pondaveia said, please change the double quotes to single quotes

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -6295,7 +6298,8 @@ void SpellMgr::LoadDbcDataCorrections()
     properties->Type = SUMMON_TYPE_TOTEM;
     properties = const_cast<SummonPropertiesEntry*>(sSummonPropertiesStore.LookupEntry(647)); // 52893
     properties->Type = SUMMON_TYPE_TOTEM;
-
+    if (properties = const_cast<SummonPropertiesEntry*>(sSummonPropertiesStore.LookupEntry(628))) // Hungry Plaguehound
```
Pondaveia: This is causing an error. To compare properties with the ones you're looking up too, you need '==' instead of '=' else you are attributing a new value to properties instead. I'm sure you didn't need the explanation but just a reminder because you probably forgot or miss-typed ;)

Pondaveia: @FrancescoBorzi He actually means it by assigning the value to properties to check if it's true, he's just missing parentheses you can close this

comix1988:
I tested this PR but the pet does not disappear even after 2 min since I completed the quest.
can someone else check?

comix1988:
I tried to test this fix again and now the quest works perfectly but the pet does not despawn even after 2 min.
@wizzymore

wizzymore:
@comix1988 check the changes now

comix1988:
it does not work yet. I think the problem is not the distance because when the quest is complete the pet is about 5 yards from my pg.
@wizzymore

poszer:
Hmm, Why do not use spell cast on summon this pet so it belong to player as minion? I'll try to fix this

poszer:
> Hmm, Why do not use spell cast on summon this pet so it belong to player as minion? I'll try to fix this
> P.S Also, i have tested this, the pet isn't despawned.

( Fix despawn and we can merge this ) - For any improvement it's easy to edit (DB changes)

wizzymore:
As now i have time, i will take a look on this in the next days, and @poszer with the ScriptMgr edit, the npc is spawned as a pet, the only thing i have to figure out is how to despawn the pet when the quest is completed, to find a better way cause the current commit works for me xD

Stoabrogga:
Fixed a few bugs:
- The variable @ENTRY was not initialized in the SQL script, that's why the despawn did not work
- Fixed distance for the despawn to reflect the comment: 10 yards instead of 20
- Added ```properties->Type = SUMMON_TYPE_PET;``` to the summon properties of the Plaguehound in order to get the pet action bar

I tested the whole thing already and it should be ready to be merged in my opinion.

PR #1700, opened 2019-04-08
Fix db_assembler on macOS

koffeine:
##### CHANGES PROPOSED:

- Change the shebang in `deps/jsonpath/JSONPath.sh`
	- from: `#!/bin/bash`
	- to: `#!/usr/bin/env bash`

###### ISSUES ADDRESSED:

I'd like to run db\_assembler on macOS to update my database.  
db\_assembler depends on JSONPath which uses the readarray command introduced in bash v4.  
On macOS the default bash is version 3.2, so I had to install a newer version from Homebrew which does support it.

The problem is, unless I change the shebang in JSONPath.sh it wants to use the older bash (because it points to `/bin/bash`).

I don't think this change will cause any problems because db\_assembler.sh already uses this kind of shebang.

- Setup the DB on Windows
- Setup the DB on Linux (or any other system using Docker)

##### Target branch(es):

Master


FrancescoBorzi:
Thanks for this contribution. This needs to be tested on Windows and Linux, to avoid regression.

btw @koffeine have you managed to run AC on macOS ? Because last time I've tried, it didn't work for me.

koffeine:
@FrancescoBorzi
No, the installation page on the wiki discouraged me from trying to build and run it natively in the first place.

I've run it successfully on a linux vm until I discovered the Docker build which I'm running since.

But the only easy way I found to keep my database up to date is to run the db_assembler script locally.

BarbzYHOOL:
GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2013 Free Software Foundation, Inc.

MacOS uses a bash from before 2013??

Kitzunu:
@FrancescoBorzi You have a Mac machine right?
Maybe you could test this quickly? :P

FrancescoBorzi:
@Kitzunu yeah but as I said:

> This needs to be tested on Windows and Linux, to avoid regression.

most likely the author tested it on Mac, so it's important to make sure that this is not broken on the other OSs too.

Kitzunu:
Ah

FrancescoBorzi:
Tested on macOS Catalina using command `bin/acore-db-asm 5`

### Before the PR

I see the error

```
/Users/shin/sources/azerothcore-wotlk/deps/jsonpath/JSONPath.sh: line 250: readarray: command not found
```

### After the PR

I don't see any error

So the PR is good for macOS 👍  however, as I said, we should test it also on windows and linux

vhiperdev:
Tested on **Linux Ubuntu** using command `bin/acore-db-asm 5
`
image



Tested on **Windows 10** using command `bin/acore-db-asm 5
`
image

Helias:
Good job @vhiperdev! :tada:

PR #1721, opened 2019-04-14
fix(Core): implement "SCRIPT_COMMAND_MOVEMENT" (35) to fix Defias Thug waypoint errors

Stoabrogga:
##### CHANGES PROPOSED:
The following errors are caused by a non existing waypoint script command:
```
ERROR: Unknown script command Unknown command: 35 ('waypoint_scripts' script id: 8025100).
ERROR: Unknown script command Unknown command: 35 ('waypoint_scripts' script id: 8014900).
```
Implement "SCRIPT_COMMAND_MOVEMENT" (35) based on this TC commit:
https://github.com/TrinityCore/TrinityCore/commit/85357c75c4a202e2b7b72e14fe0d4083f74c87bb
Thanks @Meltie2013

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
tested in-game

- ```.go creature 80149```
- stay in the area for a while
- the errors mentioned above should not occur after this PR is applied

none

##### Target branch(es):
Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1555246241741494428.sql
+++ b/data/sql/updates/pending_db_world/rev_1555246241741494428.sql
@@ -1,4 +1,0 @@
-INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555246241741494428');
-
-DELETE FROM `waypoint_scripts` WHERE `id` IN (8014900,8025100);
-UPDATE `waypoint_data` SET `action` = 0 WHERE `action` IN (8014900,8025100);
--- a/data/sql/updates/pending_db_world/rev_1555246241741494428.sql
+++ b/data/sql/updates/pending_db_world/rev_1555246241741494428.sql
@@ -1,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555246241741494428');
+
+DELETE FROM `waypoint_scripts` WHERE `id` IN (8014900,8025100);
+UPDATE `waypoint_data` SET `action` = 0 WHERE `action` IN (8014900,8025100);
--- a/src/server/game/Globals/ObjectMgr.h
+++ b/src/server/game/Globals/ObjectMgr.h
@@ -110,7 +110,8 @@ enum ScriptCommands
     SCRIPT_COMMAND_EQUIP                 = 31,               // soucre = Creature, datalong = equipment id
     SCRIPT_COMMAND_MODEL                 = 32,               // source = Creature, datalong = model id
     SCRIPT_COMMAND_CLOSE_GOSSIP          = 33,               // source = Player
-    SCRIPT_COMMAND_PLAYMOVIE             = 34                // source = Player, datalong = movie id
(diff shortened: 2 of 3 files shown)


Meltie2013:
@Stoabrogga Instead of removing the script from the database, why not implement the command? Would make more sense.

```
case SCRIPT_COMMAND_MOVEMENT:
	// Source must be Creature.
	if (Creature* cSource = _GetScriptCreature(source, true, step.script))
	{
		if (!cSource->IsAlive())
			return;

		cSource->GetMotionMaster()->MoveIdle();

		switch (step.script->Movement.MovementType)
		{
			case RANDOM_MOTION_TYPE:
				cSource->GetMotionMaster()->MoveRandom((float)step.script->Movement.MovementDistance);
				break;
			case WAYPOINT_MOTION_TYPE:
				cSource->GetMotionMaster()->MovePath(step.script->Movement.Path, false);
				break;
		}
	}
	break;
```

Stoabrogga:
@Meltie2013 
I would have done this if it were documented somewhere, but I did not find anything. Thanks for the hint, will take a look at it. You copied this from here I presume?
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Maps/MapScripts.cpp#L910-L929

Meltie2013:
Yeah, just using it as a reference for this mainly.

Stoabrogga:
@Meltie2013 
Thanks, I removed the SQL script and added the code for "SCRIPT_COMMAND_MOVEMENT" (35).

Pondaveia:
Maybe describe what this PR is supposed to fix/do? afaik You make the creature have a random walking at the end of the waypath before resuming?

Meltie2013:
@Stoabrogga 
No problem. It's honestly better to fix the issue then removing it and causing more issues from it.

Stoabrogga:
@Pondaveia 
This PR is just about fixing the missing command (and the errors caused by it) which had to be taken over from TC. As for the description of the command itself, I will add it later to our wiki, but it is not important for this PR (I also don't have much more information than the comments in the code).

I checked the DB and command 35 is only used for the two "Defias Thug" NPCs. As you correctly stated it is used here to force random movement for 20 seconds at the end of the path before restarting.

Stoabrogga:
PR for the AC wiki concerning the documentation for command 35:
https://github.com/azerothcore/wiki/pull/36

poszer:
Tested in game, this commit has fixed the issue with mentioned DB errors : 

> ERROR: Unknown script command Unknown command: 35 ('waypoint_scripts' script id: 8025100).
> ERROR: Unknown script command Unknown command: 35 ('waypoint_scripts' script id: 8014900).

Nice fix @Stoabrogga 

Before : 
defiasBefore

After:
defiasAfter

Meltie2013:
Glade you guys have it working now. Lucky I pointed @Stoabrogga in the right direction for the fix. 💯

poszer:
> Glade you guys have it working now. Lucky I pointed @Stoabrogga in the right direction for the fix.

Thank you @Meltie2013

Issue #1741, opened 2019-04-22
[NPC-QUEST] Escape from the staging grounds


comix1988:
DESCRIPTION:

the "escape from the staging grounds" quest (ID 10425) cannot be completed.

BLIZZLIKE BEHAVIOUR:

![questnether](https://user-images.githubusercontent.com/22867608/56489079-d0338500-64e0-11e9-9157-66a9f7e3c036.png)

CURRENT BEHAVIOUR:

we have more than one bug:
- the npc "captured protectorate vanguard" (ID 86833) is always visible;
- the npc "ethereum guardian" (ID 86834) is not attackable;(spawn time 26min)
- the quest can be obtained but cannot be completed because being unable to kill the npc "ethereum guardian", the npc "captured protectorate vanguard" dies.

STEPS TO REPRODUCE THE PROBLEM:

.go creature 86833 
wait 26min to spawn Ethereum guardian or .go creature 86834 (in gm on), target he e use .resp command

BRANCH(ES):
master

AC HASH/COMMIT:
commit 6021de887558a9fb4d93278a903d30ae6112e073

OPERATING SYSTEM:
linux

MODULES:
OTHER CUSTOMIZATIONS:

poszer:
Hmm interesting. As i can see from this creature in db, it has wrong flags : 
So posible issue is that, you can't kill/enter in combat with this NPC to escort Quest Creature.
> Ethereum guardian
> 256 | 0x0000 0100 | UNIT_FLAG_IMMUNE_TO_PC | disables combat/assistance with PlayerCharacters . 
Posting fix in some secs

poszer:
Also @comix1988 while doing this quest, please take the quest from Captured protectorate vanguard, than escort him. 

![QuestCaptuired](https://user-images.githubusercontent.com/41213210/56495186-3b875200-64f5-11e9-9704-64b25c1ca619.png)

Knindzagxg:
Quest it self working and can be completed which u can see bellow in video.

Problem remain with Arena Event:
* Vanguard and Gladiator should spawn at the same time and fight each other until player get close to them and kill Gladiator. At that point, Vanguard should offer quest and proceed with escort.

https://youtu.be/XKwj0yFPoZ0

PR #1742, opened 2019-04-22
fix(DB/Creature): NPC Ethereum guardian

poszer:
NPC Ethereum guardian shouldn't be immune

##### CHANGES PROPOSED:

-  Remove UNIT_FLAG_IMMUNE_TO_PC from NPC 'Ethereum guardian'.
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1741

##### TESTS PERFORMED:

Yes, the quest can be completed now without problem.

Look in the https://github.com/azerothcore/azerothcore-wotlk/issues/1741

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1555926066020418100.sql
+++ b/data/sql/updates/pending_db_world/rev_1555926066020418100.sql
@@ -1,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555926066020418100');
+
+-- Remove UNIT_FLAG_IMMUNE_TO_PC from NPC 'Ethereum guardian'
+UPDATE `creature_template` SET `unit_flags`='0' WHERE `entry`=20854;


comix1988:
@poszer @Pondaveia 
ok, now the NPC Ethereum guardian is attackable.
Reading the comments on wowhead it seems that the quest is made available only after helping the NPC by killing the Guardian.
The same mechanism as this quest https://wowwiki.fandom.com/wiki/Quest:A_Hearty_Thanks! (read the notes).
If it is possible to fix it perfectly, otherwise we will merge it like this

Pondaveia:
Nice catch @comix1988  👍

poszer:
Checking old videos I think this is correct, why WIP?

comix1988:
because I didn't know what label to put on.

for the video, if you watch the first few seconds the exclamation point is not there until it kills the guardian

poszer:
> because I didn't know what label to put on.
> 
> for the video, if you watch the first few seconds the exclamation point is not there until it kills the guardian

Or its because creature is already in combat? I will take a look on that when I come home

poszer:
This is ready for merge. Any improvement can be easily added ( Related to DB ).

data/sql/updates/pending_db_world/rev_1555926066020418100.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555926066020418100');
+
+-- Remove UNIT_FLAG_IMMUNE_TO_PC from NPC 'Ethereum guardian'
+UPDATE `creature_template` SET `unit_flags`='0' WHERE `entry`=20854;
```
FrancescoBorzi: just a note: this is not the best way to deal with flags, but it's OK to merge this since it has already been tested and approved.

However I've created a TODO in the wiki so we don't forget to do this: https://github.com/azerothcore/wiki/issues/39

Issue #1740, opened 2019-04-22
[NPC] Coilskar defender


comix1988:
DESCRIPTION:
the npc coilskar defender doesn't attack and it isn't possible to attack it

CURRENT BEHAVIOUR:

the npc coilskar defender doesn't attack and it isn't possible to attack it

STEPS TO REPRODUCE THE PROBLEM:

.go creature 70696 and try to attack it

BRANCH(ES):
master

AC HASH/COMMIT:
commit 6021de887558a9fb4d93278a903d30ae6112e073

OPERATING SYSTEM:
linux

MODULES:
OTHER CUSTOMIZATIONS:

poszer:
Same problem, flags :D

poszer:
@comix1988 when you can, try the referenced fixes :+1:

PR #1743, opened 2019-04-22
fix(DB/Creature): Coilskar defender

poszer:
Coilskar defender flags fixes

##### CHANGES PROPOSED:

- Removed unit_flags UNIT_FLAG_NOT_SELECTABLE ( Can't be selected by mouse or with /target {name} command.)
- Updated unit_flags to 'Only Swim'
- Removed extra flags CREATURE_FLAG_EXTRA_CIVILIAN ( creature does not aggro - ignore faction/reputation hostility)

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1740

##### TESTS PERFORMED:

Yes, everything is fine now.

There was issue with targeting this creature and creature doesn't enter in combat if the player is near, also doesn't attack player. 
1. .go creature 70696 try to target it/attack or just stay in range, and take a look will the creature attack you

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1555929516205604900.sql
+++ b/data/sql/updates/pending_db_world/rev_1555929516205604900.sql
@@ -1,0 +1,6 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555929516205604900');
+
+-- Coilskar defender updated flag to Only Swim
+UPDATE `creature_template` SET `unit_flags`='32768' WHERE `entry`=19762;
+-- Coilskar defender remove extra flag
+UPDATE `creature_template` SET `flags_extra`='0' WHERE `entry`=19762;
--- a/data/sql/updates/pending_db_world/rev_1555929516205604900.sql
+++ b/data/sql/updates/pending_db_world/rev_1555929516205604900.sql
@@ -1,6 +1,4 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555929516205604900');
 
--- Coilskar defender updated flag to Only Swim
-UPDATE `creature_template` SET `unit_flags`='32768' WHERE `entry`=19762;
--- Coilskar defender remove extra flag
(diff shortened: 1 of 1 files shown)


comix1988:
now it works correctly.

good job @poszer

data/sql/updates/pending_db_world/rev_1555929516205604900.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1555929516205604900');
+
+-- Coilskar defender updated flag to Only Swim
+UPDATE `creature_template` SET `unit_flags`='32768' WHERE `entry`=19762;
+-- Coilskar defender remove extra flag
+UPDATE `creature_template` SET `flags_extra`='0' WHERE `entry`=19762;
```
FrancescoBorzi: ideally we should:

1) use a single `UPDATE` statement 
2) just use numbers (and not strings) for numeric values

```sql
UPDATE `creature_template` SET `unit_flags`=32768, `flags_extra`=0 WHERE `entry`=19762;
```

3) ...and also, it is a good practice to properly remove/add flags using bitwise operators (but I don't know exactly which flags you are adding/removing here)

poszer: Hmm, okay. I'll update it :D @FrancescoBorzi 
For flags, i just removed UNIT_FLAG_NOT_SELECTABLE ( Can't be selected by mouse or with /target {name} command.) - That was issue if player wanted to attack that creature ( Can't target/select him ). Dunno, but for me it looks like wrong flag

FrancescoBorzi: it's fine, I will write a tutorial about how to better deal with flags. But this can be merged already.

poszer: Okay :+1:

PR #1753, opened 2019-04-23
Support .morph uint32 ids

brussens:
With display_id > 65535, the morph command just doesn't work

##### CHANGES PROPOSED:

-  change uint16 to uint32

Create display_id > 65535 in your creaturedisplayinfo.dbc and apply .morph it before and after the proposed changes.

##### Target branch(es):

Master

### DIFF
--- a/src/server/scripts/Commands/cs_modify.cpp
+++ b/src/server/scripts/Commands/cs_modify.cpp
@@ -1264,7 +1264,7 @@ public:
         if (!*args)
             return false;
 
-        uint16 display_id = (uint16)atoi((char*)args);
+        uint32 display_id = (uint32)atoi((char*)args);
 
         Unit* target = handler->getSelectedUnit();
         if (!target)


brussens:
@poszer Wait to be tested? You're serious, aren't you?

poszer:
> @poszer Wait to be tested? You're serious, aren't you?

Maybe there's a queue waiting. Do not except insta merge ;)

poszer:
> HOW TO TEST THE CHANGES:
> Create display_id > 65535 in your creaturedisplayinfo.dbc and apply .morph it before and after the proposed changes.

I tested and my client crash, maybe I did something wrong?  :thinking: 
crash

brussens:
Right, the problem is directly with your client. It's quite understandable, the problem is in reading the texture of the NPC.

BarbzYHOOL:
So Poszer's client is corrupted?

poszer:
> So Poszer's client is corrupted?

if the Dalaran use corrupted. Than i'm using too :smiley:  But i don't think so

brussens:
I've tested it in the Dalaran. I don't know how it can cause a mistake.

BarbzYHOOL:
Poszer:
> if the Dalaran use corrupted. Than i'm using too smiley But i don't think so

Poszer wrote like an idiot so it means nothing, he meant to say he uses the DalaranWOW client that is blizzlike and unmodified.

poszer:
We are talking about clients, and my reply is in that way

> So Poszer's client is corrupted?

>if the Dalaran use corrupted. Than i'm using too :smiley:  But i don't think so 

I didn't wrote : I tested this in Dalaran lol

comix1988:
sorry if I'm a noob .. how do I "create display_id> 65535 in your creaturedisplayinfo.dbc"?

Helias:
Use [mydbceditor](https://github.com/azerothcore/dbc-dev-kit) or [WDBX Editor](https://github.com/WowDevTools/WDBXEditor), modify the dbc file and add it on server-side and client-side.

For client-side, just make a folder in *WoW/Data/patch-X.MPQ/DBFilesClient/* and put here the creaturedisplayinfo.dbc

For server side, just overwrite the creaturedisplayinfo.dbc in server/data/dbc/

Sdyess:
What is the actual benefit of this? Just if you use custom data?

BarbzYHOOL:
I guess so. Btw @brussens you could have provided the DBC already edited to avoid errors like the screenshot above, no?

Viste:
type change testing It's not even funny

Viste:
you're so fucked up with these tests, you've come to this, seriously?

BarbzYHOOL:
send a creaturedisplayinfo.dbc with the entry plz, it's a pain for me to even try to do it

Winfidonarleyan:
I think no need test this, just merge
In TC
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Commands/cs_modify.cpp#L808

And in function `SetDisplayId(uint32 modelId)`, `modelId` type - uint32
```cpp
void Creature::SetDisplayId(uint32 modelId)
{ 
    Unit::SetDisplayId(modelId);

    if (CreatureModelInfo const* minfo = sObjectMgr->GetCreatureModelInfo(modelId))
    {
        SetFloatValue(UNIT_FIELD_BOUNDINGRADIUS, (IsPet() ? 1.0f : minfo->bounding_radius) * GetFloatValue(OBJECT_FIELD_SCALE_X));
        SetFloatValue(UNIT_FIELD_COMBATREACH, (IsPet() ? DEFAULT_COMBAT_REACH : minfo->combat_reach) * GetFloatValue(OBJECT_FIELD_SCALE_X));
    }
}
```

poszer:
Why?

brussens:
> Why?

Because the size here must be uint32 instead of uint16 because SetDisplayId(uint32 modelId) seems to accept uint32, so there can be no errors.

poszer:
> > Why?
> 
> Because the size here must be `uint32` instead of `uint16` because `SetDisplayId(uint32 modelId)` seems to accept `uint32`, so there can be no errors.

What's benefit of that?

brussens:
> What's benefit of that?

At least from what was planned and this variant is also used in TrinityCore

Vaex:
I guess this is used on TC because of later Expansions and it doesnt hurt to add for 335 branch aswell. When i exctract my CreatureDisplayInfo.dbc it goes max up to entry "32754" so it's not really necessary for my opinion, except you put custom things in your client.

FrancescoBorzi:
Just for more clarity: `display_id` looks to be only used in `target->SetDisplayId(display_id);` which was **already** taking in input a `uint32` (see `void Unit::SetDisplayId(uint32 modelId)` in `Unit.cpp`).

What @brussens did is basically avoiding that it looses precision by casting it directly to `uint32` instead of going to `uint16` first and then `uint32`.

In conclusion, I don't really see how this change could have caused a crash because the final value was `uint32` even before this PR.

PR #1754, opened 2019-04-24
C'Thun Dark Glare spell timer

fabianjana:
C'Thun Dark Glare spell should be casted about 4 seconds after C'Thun turns red, but is currently casting after just 1 second.

https://youtu.be/vZcR-OcrJGg?t=70

### DIFF
--- a/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
+++ b/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_cthun.cpp
@@ -296,7 +296,7 @@ public:
                         {
                             //Face our target
                             DarkGlareAngle = me->GetAngle(target);
-                            DarkGlareTickTimer = 1000;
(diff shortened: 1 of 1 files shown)


Pondaveia:
Correction: 3 seconds [link](https://wowwiki.fandom.com/wiki/C%27Thun_(tactics))

fabianjana:
Actually there is no info about Dark Glare (Red Beam) there, Green Beam is the one that cast every 3 seconds. I recorded while testing to compare:
- [Original value (1000)](https://www.youtube.com/watch?v=D9DXpZ-f2fE)
- [Modified value (4000)](https://www.youtube.com/watch?v=W8Aw-yAm2IQ)

I think that is better to set it to 4 seconds because the area of effect of the beam is really huge.

BarbzYHOOL:
Thank you @fabianjana

BarbzYHOOL:
yes, thanks merged

PR #1765, opened 2019-04-25
fix(Core/Spell): Greater Heal Refund

xDevICCI:
link 1: http://wotlk.cavernoftime.com/spell=37594

Description:
If your Greater Heal brings the target to full health, you gain 100 mana.

Link 2; http://wotlk.cavernoftime.com/spell=37595

Description: Restores 100 mana.

##### CHANGES PROPOSED:

-  Restore 100 mana for the priest when cast Greater Heal
-  

###### ISSUES ADDRESSED:

Closes 

- Cast Greater Heal in a Target with full HP
- Priest no restore 100 Mana

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8272,6 +8272,11 @@ bool Unit::HandleProcTriggerSpell(Unit* victim, uint32 damage, AuraEffect* trigg
                     CastDelayedSpellWithPeriodicAmount(this, trigger_spell_id, SPELL_AURA_PERIODIC_HEAL, basepoints0);
                     return true;
                 }
+                // Greater Heal Refund - Gain 100 mana , when target is full HP
+                if (auraSpellInfo->Id == 37594)
+                {
+                  trigger_spell_id = 37595;
+                }
                 break;
             }
             case SPELLFAMILY_DRUID:


Pondaveia:
Thanks for these fixes

talamortis:
hmm if i remeber rightly with this ( as i played priest on retail) it was only if the heal brought the player too 100% life that it granted 100 mana to the priest so for example

Unit has 300/500 health the priest heals the player for 202 this will grant the mana to the priest but if  unit was at 500/500 the mana proc would not happen

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8272,6 +8272,11 @@ bool Unit::HandleProcTriggerSpell(Unit* victim, uint32 damage, AuraEffect* trigg
                     CastDelayedSpellWithPeriodicAmount(this, trigger_spell_id, SPELL_AURA_PERIODIC_HEAL, basepoints0);
                     return true;
                 }
+                // Greater Heal Refund - Gain 100 mana , when target is full HP
+                if (auraSpellInfo->Id == 37594)
```
Pondaveia: As Talarmortis mentioned, can you add a check if victim isn't full HP?

GMKyle:
According to my test priest always gains 100 of mana when cast greater heal so this PR isn't needed. Trinity has a validation in 37594 spell_script maybe we can use it.

src/server/game/Entities/Unit/Unit.cpp
wizzymore:
```diff
-if (auraSpellInfo->Id == 37594)
+if (auraSpellInfo->Id == 37594 && victim->IsFullHealth())
```

Try this, i think this should work ( i am at work so i can't test it in game )

xDevICCI: ok

wizzymore: Does it work now?

wizzymore:
I checked now and this PR does nothing, the set already was giving 100 mana and this is not fixing anything.

What we need to modify is this: spell_pri_item_greater_heal_refund
I think i found the solution and i am compiling now, if it works i will post it here

src/server/game/Entities/Unit/Unit.cpp
P-Kito: I did not check the cicumstances, but why is this not in a seperate aura/spell script?

lineagedr:
- You need to have atleast 2x of the Avatar Raiment set equiped in order for it to give you 100 mana.
- The spell 37594: belongs to the Avatar Raiment set.
- Greater Heal never did refund 100 mana without the set.

The main problem is that by the time it goes to "spell_pri_item_greater_heal_refund" the player is already at full health making IsFullHealth() useless.

poszer:
Closing this, PR opened here https://github.com/azerothcore/azerothcore-wotlk/pull/1985

PR #1775, opened 2019-04-26
fix(CORE/Commands): .Pinfo

IntelligentQuantum:
##### TESTS PERFORMED:
Yes, imported without errors. Checked and tested in game, everything looks fine.

1: .guild create "Guild Name"
2: .Pinfo
3: Now you can see Guild Name Guild Rank Note Office Note And Xp - Total Xp

##### Target branch(es):
Master


poszer:
Tested, everything looks fine :+1: 
guildPINFO

BarbzYHOOL:
it looks off on your screen

btw we didn't have an issue with the characters shown?

poszer:
> it looks off on your screen
> 
> btw we didn't have an issue with the characters shown?

Yes. That's how is should be:
Here is my old image:
tcbefore

poszer:
Yes. That's how is looks @BarbzYHOOL 

Old image:
Screenshot_2

BarbzYHOOL:
@poszer the old image is the same as the above, same output shown

Sdyess:
result5, stmt2, etc...

Please use names that are more meaningful. result and stmt can make sense for one result and statement because they are unique to the scope generally, but now you're making it difficult to keep up with the names.

IntelligentQuantum:
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/Commands/cs_misc.cpp#L1899

BarbzYHOOL:
No, but you can change them in another PR

BarbzYHOOL:
Thank you

PR #1818, opened 2019-05-09
fix(Core): Reputation modifier now affects negative rep reward

BarbzYHOOL:
##### CHANGES PROPOSED:

-  Can create custom spell that modifies reputation gain that will work correctly
-  Can also stack more than 10 times the spell "Diplomacy" and it will work as it should lol (yeah, never gonna happen)

This is mostly for custom usage but it wasn't working logically.

###### ISSUES ADDRESSED:

Before that commit, when you killed a mob that earned reputation but also made you lose some
reputation, the reputation loss didn't occur. Same for quest reputation reward.

It worked like that:
If main reputation = positive
And secondary reputation = positive
--> Both reputation increased by X%

If main reputation = positive
And secondary reputation = negative
--> Main reputation increased by X%
--> Secondary reputation set to 0

##### TESTS PERFORMED:

I did booty bay quests and it worked fine, I killed pirates and it worked fine.

Before fix:
-  `.go creature 2596` Kill a pirate near booty bay and look at the reputation gain and loss
- `.aura 20599` 9 times and kill a pirate and look at reputation
- `.aura 20599` 10 times and kill a pirate (0 reputation loss)
- `.aura 20599` 11 times and kill a pirate (0 reputation loss)

After fix:
- `.aura 20599` 11 times and kill a pirate -> you lose more reputation thanks to the 110% reputation modifier

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -982,7 +982,7 @@ Player::~Player()
 }
 
 void Player::CleanupsBeforeDelete(bool finalCleanup)
-{ 
+{
     TradeCancel(false);
     DuelComplete(DUEL_INTERRUPTED);
 
@@ -990,7 +990,7 @@ void Player::CleanupsBeforeDelete(bool finalCleanup)
 }
 
 bool Player::Create(uint32 guidlow, CharacterCreateInfo* createInfo)
-{ 
+{
     //FIXME: outfitId not used in player creating
     // TODO: need more checks against packet modifications
     // should check that skin, face, hair* are valid via DBC per race/class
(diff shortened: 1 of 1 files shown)


BarbzYHOOL:
Note: the diff is only 2 lines long starting at https://github.com/azerothcore/azerothcore-wotlk/pull/1818/files#diff-5589ee47b3a7ad220cff105a8fe7eac3L7047 up to 7073. The rest is just automatically cleaned useless spaces

BarbzYHOOL:
Note to self: write the diff line number in the squash commit

BarbzYHOOL:
Ok, so after discussing it with Shard-MW, I looked at wowhead comments and they say that the reputation modifier should only affect positive gains, not reputation losses.

So this PR is useless unless we add a config option for this.

In top of that, there is now another issue uncovered: the modifier should not reduce the reputation loss, it should stay unchanged.

Kitzunu:
Any update?

PR #1804, opened 2019-05-06
fix(DB/Creature): Delete creature, wrong spawn

poszer:
Creatures is wrongly spawned there

##### CHANGES PROPOSED:

-  Delete one creature from Borean T. Creature is wrongly spawned there causing issue entering in combat with other creature (CreatureTextMgr: Could not find TextGroup 1 for Creature(Bone Warrior) GuidLow 97670 Entry 26126. Ignoring.
-  

###### ISSUES ADDRESSED:
 There is issue about it, but i won't close it, since there is more creatures with some db problems.
Related to this : https://github.com/azerothcore/azerothcore-wotlk/issues/1062

Closes 

##### TESTS PERFORMED:

Yes, now is fine, also no more db errors about that text.

##### HOW TO TEST THE CHANGES:

Just go creature 97670 , and take a look around. If you find another creature like that (fighting others) congratz :rofl: 
Including one image : 
https://image.prntscr.com/image/L3TIt80USk2j7R8pc69Miw.png
Red is where is he spawned atm.
Green is, where are others creatures like him. 

##### KNOWN ISSUES AND TODO LIST:

- [ ]
- [ ] 

##### Target branch(es):

Master

PR #1819, opened 2019-05-10
fix(DB): Defendo-tank 66D - fix target of "say" action

Stoabrogga:
##### CHANGES PROPOSED:
The Defendo-tank 66D talks on aggro, but uses "SMART_TARGET_ACTION_INVOKER" instead of "SMART_TARGET_SELF". This is corrected through this PR.

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
tested in-game

- ```.go creature id 25758```
- attack the creature

none

##### Target branch(es):
Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1557489332825322207.sql
+++ b/data/sql/updates/pending_db_world/rev_1557489332825322207.sql
@@ -1,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1557489332825322207');
+
+UPDATE `smart_scripts` SET `target_type` = 1 WHERE `entryorguid` = 25758 AND `source_type` = 0 AND `id` = 0;


comix1988:
the npc works correctly even after this change. for me it's OK.

poszer:
> the npc works correctly even after this change. for me it's OK.

It's working ofc. Problem is referenced PR, read that, it's not referenced without reason. Than (Let that creature to attack this creature, than check your Log). -> You will get errors 

After that, apply @Stoabrogga fix and check again when it aggain attack creature from referenced PR --> The DB errors should disappear?

Stoabrogga:
@comix1988 
The NPC is not working correctly. The SAI makes the target say text line 0 on aggro because it uses the wrong target type (invoker instead of self). This leads to the Bone Warrior saying "Die mortal!". The Bone Warrior is in the wrong area (see #1804), but without him this error would have never shown, because the target is most of the time the player. After applying this PR the tank correctly says it's creature text "Incoming $r flier! Shoot $g him : her; down!" (the tank is crazed as are all mechanical units in the area).

poszer:
> @comix1988
> The NPC is not working correctly. The SAI makes the target say text line 0 on aggro because it uses the wrong target type (invoker instead of self). This leads to the Bone Warrior saying "Die mortal!". The Bone Warrior is in the wrong area (see #1804), but without him this error would have never shown, because the target is most of the time the player. After applying this PR the tank correctly says it's creature text "Incoming $r flier! Shoot $g him : her; down!" (the tank is crazed as are all mechanical units in the area).

Tested, now is fine, doesn't get any db errors in log. Aynway we are deleting this creature due to wrong spawn. Thanks @Stoabrogga

PR #1836, opened 2019-05-15
fix(DB/Quest) Solved bug: A Spirit Ally? id 9847

Seebaby:
Quest: "A Spirit Ally?" id 9847
Questitem: "Feralfen Totem" id 24498

Error: The mission item has no cd and infinite snakes can be invoked.

Simple solved:

Thanks to @solidmaxtor for the fix

Closes 1822

##### TESTS PERFORMED:
I tested it myself and it worked.

Give yourself the quest quest: "A Spirit Ally?" id 9847

go to the place of the quest 

try to use the quest item more than once

##### Target branch(es):

Master

### DIFF
--- a/Quest: "A Spirit Ally?" id 9847
+++ b/Quest: "A Spirit Ally?" id 9847
@@ -1,0 +1,1 @@
+UPDATE `item_template` SET `spellcharges_1` = 1 WHERE `entry` = 24498;


poszer:
Thanks for contribution @Seebaby . But this isn't the correct way, please read here infos, than re-create it :
http://www.azerothcore.org/wiki/Dealing-with-SQL-files

BarbzYHOOL:
Ask us in chat if needed, don't waste the fix, we want it :P

Selyria:
i'll be happy to do as i said i'm a complete newcomer to all this but i'm getting ahead so slowly

Issue #1862, opened 2019-05-21
Low level repeatable quest exclamation marks appearing as yellow instead of blue


sluggor:
Low level repeatable quests appear as yellow exclamation marks instead of blue

##### SMALL DESCRIPTION:

When using the "Low Level Quests" tracker on the mini map, repeatable quests will appear as yellow exclamation marks over the heads of questgivers and on the minimap. They should appear as blue. If the quest is not low-level, it will appear properly as blue.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Repeatable low-level quests should appear as blue exclamation marks.

##### CURRENT BEHAVIOUR:

Repeatable low-level quests are appearing as yellow exclamation marks.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Find a repeatable, low-level quest where the exclamation mark is hidden.
2. Turn on "Low Level Quests" tracking on the mini map.
3. Exclamation mark will be yellow.

##### EXTRA NOTES:

The code handling this is below:

Link: [Player.cpp](https://github.com/azerothcore/azerothcore-wotlk/blob/0b9a0d8ea63cec42df9e19d013ca37cd08cc9613/src/server/game/Entities/Player/Player.cpp#L16805)

```
QuestStatus status = GetQuestStatus(questId);

if (status == QUEST_STATUS_NONE)
{
	if (CanSeeStartQuest(quest))
	{
		if (SatisfyQuestLevel(quest, false))
		{
			if (quest->IsAutoComplete())
				result2 = DIALOG_STATUS_REWARD_REP;
			else if (getLevel() <= (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)))
			{
				if (quest->IsDaily())
					result2 = DIALOG_STATUS_AVAILABLE_REP;
				else
					result2 = DIALOG_STATUS_AVAILABLE;
			}
			else
				result2 = DIALOG_STATUS_LOW_LEVEL_AVAILABLE;
		}
		else
			result2 = DIALOG_STATUS_UNAVAILABLE;
	}
}
```

The issue is that the code checks to see if it's a daily *after* the low-level check, so low-level quests never get recognized as repeatable. One possible fix (that works for me) would be:

```
        QuestStatus status = GetQuestStatus(questId);
        if (status == QUEST_STATUS_NONE)
        {
            if (CanSeeStartQuest(quest))
            {
[...]

PR #1867, opened 2019-05-22
Add code to ensure low-level quests have blue exclamation marks when …

sluggor:
…using tracker

Ensure low-level quest markers are blue when using map tracker

##### CHANGES PROPOSED:

-  Break apart logic checking for low level quest and if quest is repeatable

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1862

##### TESTS PERFORMED:

No errors on build. Ran through debugger to make sure variables were being set correctly.

Bring 80 level character to Isle of Quel'Danas and use "Low Level Quests" tracker on mini map to see all the dailies. Without the fix, they will be yellow. With the fix, they will be blue - the correct color for dailies.

None

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -16808,23 +16808,23 @@ QuestGiverStatus Player::GetQuestDialogStatus(Object* questgiver)
             if (CanSeeStartQuest(quest))
             {
                 if (SatisfyQuestLevel(quest, false)) {
-                    bool bLowLevel = (getLevel() > (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)));
+                    bool isLowLevel = (getLevel() > (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)));
 
                     if (quest->IsAutoComplete()) {
-                        if (bLowLevel)
+                        if (isLowLevel)
                             result2 = DIALOG_STATUS_LOW_LEVEL_REWARD_REP;
                         else
                             result2 = DIALOG_STATUS_REWARD_REP;
                     }
                     else {
                         if (quest->IsDaily()) {
-                            if (bLowLevel)
+                            if (isLowLevel)
                                 result2 = DIALOG_STATUS_LOW_LEVEL_AVAILABLE_REP;
                             else
                                 result2 = DIALOG_STATUS_AVAILABLE_REP;
                         }
                         else {
-                            if (bLowLevel)
+                            if (isLowLevel)
                                 result2 = DIALOG_STATUS_LOW_LEVEL_AVAILABLE;
                             else
                                 result2 = DIALOG_STATUS_AVAILABLE;
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -16807,19 +16807,29 @@ QuestGiverStatus Player::GetQuestDialogStatus(Object* questgiver)
         {
             if (CanSeeStartQuest(quest))
             {
-                if (SatisfyQuestLevel(quest, false))
-                {
-                    if (quest->IsAutoComplete())
-                        result2 = DIALOG_STATUS_REWARD_REP;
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -16807,19 +16807,29 @@ QuestGiverStatus Player::GetQuestDialogStatus(Object* questgiver)
...
-                if (SatisfyQuestLevel(quest, false))
-                {
-                    if (quest->IsAutoComplete())
-                        result2 = DIALOG_STATUS_REWARD_REP;
-                    else if (getLevel() <= (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)))
-                    {
-                        if (quest->IsDaily())
-                            result2 = DIALOG_STATUS_AVAILABLE_REP;
+                if (SatisfyQuestLevel(quest, false)) {
+                    bool bLowLevel = (getLevel() > (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)));
```
Sdyess: I'm not sure if this is intended to be hungarian notation or supposed to sound like "below level". Either way, I would recommend changing it.
```suggestion
                    bool isBelowLevel = (getLevel() > (GetQuestLevel(quest) + sWorld->getIntConfig(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF)));
```

sluggor: It's hungarian notation (boolean), but it's just a habit, we can make it whatever conforms most to the project's coding standards.

Although what do you mean by "changing its uses as well"? It's just a flag to indicate if it's a low level quest or not.

Sdyess: Sorry for not being more clear, by changing it's uses I just meant to also rename where it's being used. My suggested change had only been where it was declared.

sluggor: I see. I'm fine either way, but I think "isBelowLevel" is a little misleading. A low level quest, by definition, is below your level, it's just a question of how much.

Once you pass the threshold defined in the worldserver.conf file, the quest is considered "low level" and will no longer show up in game or on the mini map unless explicitly tracked. The relevant configuration options are as follows...

```
#
#    Quests.LowLevelHideDiff
#        Description: Level difference between player and quest level at which quests are
#                     considered low-level and are not shown via exclamation mark (!) at quest
#                     givers.
#        Default:     4  - (Enabled, Hide quests that have 4 levels less than the character)
#                     -1 - (Disabled, Show all available quest marks)

Quests.LowLevelHideDiff = 4

#
#    Quests.HighLevelHideDiff
#        Description: Level difference between player and quest level at which quests are
#                     considered high-level and are not shown via exclamation mark (!) at quest
#                     givers.
#        Default:     7  - (Enabled, Hide quests that have 7 levels more than the character)
#                     -1 - (Disabled, Show all available quest marks)

Quests.HighLevelHideDiff = 7
```

So, once the character's level is higher than the level of the quest plus the value defined by `Quests.LowLevelHideDiff `, it is considered "low level".

As far as naming convention goes, I'm fine with `isLowLevel`, but `isBelowLevel` seems to miss the mark.

Sdyess: The name you suggest is fine with me as well. Thanks for the pull request 😃

sluggor: OK, the change is in place. Still pretty new to GitHub, though, so please let me know if anything is off, thanks.

wizzymore:
Tested and it works! Good job

BarbzYHOOL:
Can you run the bash script in your repository root? it's called something like .git_commit...bash

Otherwise, awesome and thank you for contrib! @sluggor

src/server/game/Entities/Player/Player.cpp
BarbzYHOOL: ty sdyess

sluggor:
I don't see any bash scripts like that... What does it do, maybe I can just use git to do it manually?

BarbzYHOOL:
@sluggor it's here in fact, https://github.com/azerothcore/azerothcore-wotlk/tree/master/apps/git_tools

I put it there in case we have other git_tools in the future but I don't know if that was a good idea

sluggor:
Thanks, found it. The commit template has been updated.

PR #1900, opened 2019-05-28
[Spell/Core] Glyph of Shadow Word: Death

xDevICCI:
Glyph of Shadow Word: Death: Targets below 35% health take an additional 10% damage from your Shadow Word: Death spell.

https://wotlk-twinhead.twinstar.cz/?spell=55682

##### CHANGES PROPOSED:

-  Objectives below 35% of life receive an additional 10% damage using the spell of the Word of Shadows: death.

###### ISSUES ADDRESSED:

The Report does not exist, the first thing you should do is the following:

- Get the glyph: Glyph of Shadow Word: Death

- Go to a Boss or Raid and cast the spell when the target is or is below 35%

- You are not respecting what the glyph indicates

##### TESTS PERFORMED:

- 

- Get the glyph: Glyph of Shadow Word: Death

- Go to a Boss or Raid and cast the spell when the target is or is below 35%

- You are not respecting what the glyph indicates

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10564,6 +10564,14 @@ float Unit::SpellPctDamageModsDone(Unit* victim, SpellInfo const* spellProto, Da
                     if (victim->GetAuraEffect(SPELL_AURA_PERIODIC_DAMAGE, SPELLFAMILY_PRIEST, 0x100000, 0, 0, GetGUID()))
                         AddPct(DoneTotalMod, aurEff->GetAmount());
             }
+            // Shadow Word: Death
+            else if (spellProto->SpellFamilyFlags[1] & 0x2)
+            {
+                // Glyph of Shadow Word: Death
+                if (AuraEffect* aurEff = GetAuraEffect(55682, 1))
+                    if (victim->HasAuraState(AURA_STATE_HEALTHLESS_35_PERCENT))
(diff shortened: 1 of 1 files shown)


P-Kito:
Why not in proper Aura/Spell script?

GMKyle:
@P-Kito @poszer 
In this case i think it's better not use Aura/Spell script because if you look Unit::SpellPctDamageModsDone closely there are many glyphs(including Glyph of Smite ), spells..... I don't know why all this stuff ends there but it's seems more "generic", for example, the differences between  futures Aura/Spell scripts implementation (Glyph of Smite and Glyph of Shadow Word: Death)  would be minimal.
Anyways, i modifed Shadow Word Death spellscript to fix the damage increase and it's working but it's doesn't feel right.
note: I don't know why neither but trinity use the same code. :P

pklloveyou:
Pass the test and work as expected

FrancescoBorzi:
good job @xDevICCI and @pklloveyou 

waiting for a second opinion for the code review, because there were comments

Stoabrogga:
Looks ok for me. It's the same like on the TC 3.3.5 branch:
https://github.com/TrinityCore/TrinityCore/blob/c79571a09678a5aefe9e7f0d7a936932e0727b2e/src/server/game/Entities/Unit/Unit.cpp#L6722-L6729

xDevICCI:
thanks :)

PR #1909, opened 2019-05-29
fix(DB/Creature): Invisibility Creature Trigger

Selyria:
Fixes the visibility of the Creature Id 18374 because this is a trigger
Boss Shirrak the Dead Watcher Id 18371

Closes nothing

TESTS PERFORMED:
I tested it myself and it worked.
Fix is not on the server https://youtu.be/4mlLfyIoaN8
Fix on Server https://youtu.be/uFkdy-0-8ms

HOW TO TEST THE CHANGES:
Attack the Boss and wait for him to cast the spell. 
Focus fire on you sets a flame appears below you 

.go c 83388

Target branch(es):
Master


poszer:
Good, adding value of 128 (CREATURE_FLAG_EXTRA_TRIGGER - Creature is trigger-NPC / invisible to players only). Also nice catch :+1: Tested, everything is fine and this is correct. Thank you

Selyria:
I hope I did the right thing.

data/ sql/ updates/ pending_db_world/rev_1559170250353381400.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1559170250353381400');
+
+UPDATE `creature_template` SET `flags_extra` = 128 WHERE (entry = 18374);
```
BarbzYHOOL:
```suggestion
UPDATE `creature_template` SET `flags_extra` = `flags_extra` | 128 WHERE (entry = 18374);
```

I should have suggested this to other people before too but it's the best way for flags. In case someone have modified the flag, it will not erase it but it will make sure it has 128. Example, if it has 2, then it will have 130

FrancescoBorzi:
@Selyria you did, good job

BarbzYHOOL:
What is the name of the boss ? I'd like to merge but dunno the name

Selyria:
> What is the name of the boss ? I'd like to merge but dunno the name

Shirrak the Dead Watcher Id 18371

BarbzYHOOL:
thank you @Selyria

boom8866:
UPDATE `creature_template` SET `flags_extra` = 128 WHERE (entry = 20308);

diff 1 20308

Selyria:
> UPDATE `creature_template` SET `flags_extra` = 128 WHERE (entry = 20308);
> 
> diff 1 20308

I've got the fix in the new PR.

PR #1926, opened 2019-06-02
fix(DB/Core): Play "Lament of the Highborne" as music instead of sound

Stoabrogga:
##### CHANGES PROPOSED:
- Play "Lament of the Highborne" as music for all players in the area if "Sylvanas's Music Box" (item ID 52253) is used
- Play "Lament of the Highborne" as music for all players in the Undercity zone if the quest "Journey to Undercity" (ID 9180) is finished
- Small adjustment for the Thrall event ("Hero of the Mag'har"), which also uses "SMART_ACTION_MUSIC": Set "onlySelf" to 1 as the music should be played directly for all players in the zone. Does not need to be tested, as there is no difference, just should make it clear in the SAI script that the music is played directly.

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested in-game

"Sylvanas's Music Box":
- Either use the item
```.additem 52253```
or cast one of the summon spells directly:
```.cast 73332```
```.cast 73333```
- the music should be played for each player in the same area instead of the normal soundtrack

"Journey to Undercity"
- preparation:
```
.quest remove 9180
.quest add 9180
.tele RoyalQuarter
```
- finish quest "Journey to Undercity"
- the music should be played for each player in the Undercity zone instead of the normal soundtrack

none

##### Target branch(es):
Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1559456042362460279.sql
+++ b/data/sql/updates/pending_db_world/rev_1559456042362460279.sql
@@ -1,7 +1,7 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1559456042362460279');
 
--- use "SMART_ACTION_MUSIC" instead of "SMART_ACTION_SOUND" to play "Lament of the Highborne" as music for the players in the area
-UPDATE `smart_scripts` SET `action_type` = 216, `action_param2` = 1, `action_param3` = 2, `target_type` = 1, `target_param1` = 0, `comment` = 'Sylvanas Lamenter - On Update - Play Area Music' WHERE `entryorguid` = 39048 AND `source_type` = 0 AND `id` = 1;
+-- use "SMART_ACTION_MUSIC" instead of "SMART_ACTION_SOUND" to play "Lament of the Highborne" as music for all players within 50 yards distance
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Knindzagxg** [comment]
Tested, looks really nice.

Tho, dunno why, 3 errors appeared from no-where after restart, and have nothing to do with this.

```
2019-06-12 16:45:17 addToGossipMenuItems: GossipMenuId 8539 Item 0 not found
2019-06-12 16:45:17 Not handled grouped condition, SourceGroup 8539
2019-06-12 16:45:17 Script named 'boss_mother_smolderweb' does not have a script name assigned in database.
```

I think i remember something about Smolderweb, but not sure, will check it a bit later.

Stoabrogga:
@Knindzagxg 
Thank you for testing this. I don't think the errors are caused by this PR.

I did change the music box: Instead of playing for the whole area (or sub-zone) it will now play only for all players within 50 yards.

Stoabrogga:
According to this comment the music is also played for all players who come into range of the lamenters as long as the lamenters exist:
https://www.wowhead.com/item=52253/sylvanas-music-box#comments:id=1080787:reply=177093

I updated the SQL script accordingly.

Issue #1948, opened 2019-06-06
[Item in Dungeon] If use Saronite Bomb in Malygos  the platform is destroy - Video


solidmaxtor:
If use Saronite Bomb in Malygos the platform is destroy

Item: Saronite Bomb
https://www.wowhead.com/item=41119
Where:
Eye of Eternity - Malygos Raid 10/25

Possible gameobject related entry 193958 and 193960 
"The Focusing Iris"
tipe 10 GAMEOBJECT_TYPE_GOOBER
data0: open (LockId from Lock.dbc)
data2: eventID (event_script id)
data3: Time in ms before the initial state is restored
ScriptName: go_the_focusing_iris

STEPS TO REPRODUCE THE PROBLEM:
CHeck video

Test in last master
with and without modules
Ubuntu 16.04

Video
https://www.youtube.com/watch?v=1n7MmaWs2O8

Selyria:
Okay, the bug is really mega.

Good thing you found him.

talamortis:
Will assign myself but will look into this when I get a little more time.

PR #1964, opened 2019-06-12
Fix (DB/gameobject): Band-aid fix for Malygos platform.

Andrei-UPB:
Nexus Raid Platform no longer collapses on the first Saronite Bomb use
##### CHANGES PROPOSED:

-  Modify Nexus Raid Platform's health from 100 to 1.500.000, thus making it way harder to break
-  Approach a more elegant solution where the platform doesn't take any damage from siege effects in the first place. // DONE by changing the gameobject faction to 35.

###### ISSUES ADDRESSED:
 

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1948

##### TESTS PERFORMED:

Tested, the platform now breaks upon taking 1.500.000 damage ( an unreachable ammount by normal standards) rather than on 100 damage.

A saronite bomb does 150 siege damage, 25 players using Saronite Bombs on cooldown and considering 11 saronite bombs can be used before Malygos enters berserk mode (10 min), the maximum damage that will be dealt to the platform will be 41.250, if we add in Global Thermal Sapper Charges -250 siege damage per use, it will still take an unimaginable amount of pulls to actually break the platform if the platform does not recover hp between pulls (not tested yet, i'll keep working on it).
The fix is not perfect and is a band-aid solution , I will probably return with the proper one once I figure a way to make the platform immune to siege damage caused by bombs, Global Thermal Sapper Charge and the likes.

Enter the Eye of Eternity on 10 or 25 man modes.
.tele nexus or .go creature id 28859 with .gm fly on enabled (you might fall to your death otherwise)
.learn 61766 - Saronite Bomb or .learn 56488  Global Thermal Sapper Charge
Use the learned spell on the platform, notice  that it does take damage but no longer breaks due to the way higher HP threshold. 

- [x] Make the platform unable to be broken by players accidentally in the first place.
- [x] Changed the platform's faction to 35 from 0.

##### Target branch(es):

Master


Naxxramas-Server:
Have you try to reboot your server after thix fix ? Because i'v already try that and data can't saved after a rebooting

Andrei-UPB:
Yes, I did, restarting the server does not revert the change.

GMKyle:
@Andrei-UPB 
To prevent damage we can change gameobject faction to 35(friendly to all) can you or anyone test if boss encounter runs normally? I will run my own test but i have to wait until tomorrow :(

Andrei-UPB:
@GMKyle 
Hey, thanks for that , I'm new to this (got a CS background however) and had no idea gameobjects could have factions, but I figured it out and it's in the gameobject_template_addon. I changed the faction from 0 to 35 and the platform no longer takes damage. I left the data0 (health) change in as well just to make sure, the variable being 100 or 1500000 still takes 24bytes afaik, and just in case that someone does manage to somehow inflict siege damage to the platform it would still remain intact.

I have tested the fight post-application of the attached fix and Malygos does indeed break the platform (3/3 tests) when starting phase 3.

I would like more people to test it out and then it can be merged if everything's alright.

zabustak:
if you set 1.5 m HP also you have to change
``
go->ModifyHealth(-1500000, c);
``

data/sql/updates/pending_db_world/rev_1560396171418694700.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1560396171418694700');
+
+UPDATE `gameobject_template_addon` SET `faction`='35' WHERE  `entry`=193070; -- Nexus Raid Platform now has faction 35, instead of 0 and is friendly to players.
+UPDATE `gameobject_template` SET `Data0`='1500000' WHERE  `entry`=193070; -- Nexus Raid Platform has a lot more hp in order not to break in case any siege damage is taken, initial value was 100. (double safety)
```
FrancescoBorzi: Thanks for your contribution. Just a few notes:

- No need to wrap numbers in single quotes
- No need to use 2 statements, you can use one single `UPDATE` to change multiple attributes
- Please add an empty line at the end of the file

Example:

```sql
UPDATE `gameobject_template` SET  `faction`=35`, `Data0`=1500000 WHERE  `entry`=193070;
```

Vaex:
@FrancescoBorzi @Andrei-UPB This fix looks not bad but you see still the damage on that platform. A better way is to disable the Spell for EoE (visuals still working but there is no damage anymore)

```
DELETE FROM `disables` WHERE `sourceType` = 0 AND `entry` = 61766;
INSERT INTO `disables` (`sourceType`, `entry`, `flags`, `params_0`, `params_1`, `comment`) VALUES
(0, 61766, 17, '616', '', 'Disable Spell Saronite Bomb for Eye of Eternity Raid');
```

FrancescoBorzi:
@Vaex can you please open a PR with that?

Vaex:
@FrancescoBorzi sure

Stoabrogga:
Closed as the fix will be implemented via #2085.

Issue #1968, opened 2019-06-13
Implementing CC Delay


Shard-MW:
CC Delay

##### SMALL DESCRIPTION:
Crowd Control Delay improves the player's PvP experience, and isn't implemented yet in AzerothCore. 
CC Delay allows players to get cced at the same time if they cast their spell at the same time.
This feature is implemented on official servers, and even if the delay was reduced since WoD, it still exists.

Here is some retail videos where you can see CC Delay : 
1. [https://www.youtube.com/watch?v=spH5R2pTBFU](url)
-> 0:15 - Vanish Blind
-> 0:25 - Vanish Blind
-> 0:32 - Double Sap
-> 0:39 - Vanish Blind
-> 1:08 - Vanish Cheapshot
-> 1:59 - Double Blind

2. [https://www.youtube.com/watch?v=31WsqoVmVn0](url)
-> 0:44 - Vanish Death Coil
-> And more...

Of course, this works with all mechanic spells, like double sheep, cyclones etc etc.
Here is a good example of what CC Delay is, even it's not a video from retail : 
[https://www.youtube.com/watch?v=bvwvz4B9VjY](url)

##### EXPECTED BLIZZLIKE BEHAVIOUR:
As you can see on the videos above, when two players cast a cc on each other at the same time, they should be both cced.

##### CURRENT BEHAVIOUR:
Currently (master, commit : 613bb9d), only one of the two players is cced.

##### STEPS TO REPRODUCE THE PROBLEM:
You'll need two players to test it. There is a simple way to test it as rogue.

Before commit : 
1. Make a macro
`/cast [exists, harm] Vanish(rank 3)` or `/cast [exists, harm] Blind`
2. Start a duel with the second rogue and ask him to use only his Blind on you
3. Make sure that you don't have any target, and spam the macro like you're firing on a FPS game
4. Observe that you're Blind cced.

After commit : 
4. Observe that you've Vanished his Blind, or you're both Blind cced.

##### EXTRA NOTES:
This commit will enable CCDelay.
Note that it's my current configuration based on my PvP experience (I'm a 2800+ retail player), and it was tested on thousands of players on a TrinityCore live server, and there was no bugs.
This code haven't required some extra modifications when I implemented it on my AC server.
Obviously, I tested it too on AC, and the result was the same :)

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
Commit : 613bb9d00266681582e53cfe9437cd53eca0b0b8

[...]

PR #1969, opened 2019-06-13
fix(Core/Spells): Adding CCDelay

Shard-MW:
Implementing CCDelay

###### ISSUES ADDRESSED:
Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1968

##### TESTS PERFORMED:
Tests performed are explained in the issue

##### Target branch(es):
Master


src/server/game/Spells/Spell.cpp
```diff
@@ -6392,6 +6398,303 @@ SpellCastResult Spell::CheckPetCast(Unit* target)
...
+        if (_spell->Id == 2637 || _spell->Id == 18657 || _spell->Id == 18658)
+            return delayForDisorients;
+        // Feral charge - Bear
+        if (_spell->Id == 45334)
+            return delayForInstantSpells;
+        // Feral charge - Cat
+        if (_spell->Id == 49376)
+            return delayForInstantSpells;
+        // Maim
+        if (_spell->Id == 22570 || 49802)
```
lineagedr: ``if (_spell->Id == 22570 || _spell->Id == 49802)``

src/server/game/Spells/Spell.cpp
```diff
@@ -6392,6 +6398,303 @@ SpellCastResult Spell::CheckPetCast(Unit* target)
...
+            return delayForStuns;
+        // Pounce
+        if (_spell->Id == 9005 || _spell->Id == 9823 || _spell->Id == 9827 || _spell->Id == 27006 || _spell->Id == 49803)
+            return delayForOpenerStuns;
+        // Entangling roots
+        if (_spell->Id == 339 || _spell->Id == 1062 || _spell->Id == 5195 || _spell->Id == 5196 || _spell->Id == 9852 || _spell->Id == 9853 || _spell->Id == 26989 || _spell->Id == 53308)
```
lineagedr: Entire code could be improved there's no need to check for all of these spells where you could do:

``if (_spell->Mechanic == MECHANIC_ROOT)``

And for stuns

``if (_spell->Mechanic == MECHANIC_STUN)`` And so on and so fourth. Take a look at: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Miscellaneous/SharedDefines.h#L1203

src/server/game/Spells/Spell.cpp
Winfidonarleyan: Yes, yes

Shard-MW: Thanks @lineagedr fixed ! 
And it passed Travis build... Strange

src/server/game/Spells/Spell.cpp
Shard-MW: That's true but some spells like `Dragon's Breath`, `Frost Nova`, `Mage - Pet - Freeze`, `Felguard's Intercept` and more doesn't have any mechanic assigned, so I prefered to keep the basic functioning by doing case by case.
Plus, there is some spells like hunter traps, or priest's Mind Control that don't require a delay.
I know if there is a delay on `Mind Control` and the targeted player use Repentance at the same time, priest will get 1min cc instead of 10sec.

lineagedr: For spells such as Dragon's Breath you can use 
```if (_spell->SpellIconID == 1548)```
Frost nova: 
```MECHANIC_ROOT```
Water Elemental Freeze: 
```if (_spell->Id == 33395)```

It's fine to use Ids for spells that have no mechanics, but using this:
```
// Entangling roots
if (_spell->Id == 339 || _spell->Id == 1062 || _spell->Id == 5195 || _spell->Id == 5196 || _spell->Id == 9852 || _spell->Id == 9853 || _spell->Id == 26989 || _spell->Id == 53308)
```

Instead of this:
```
// Entangling roots 
if (_spell->Mechanic == MECHANIC_ROOT)
```

Is madness.

**Comment by BarbzYHOOL** [comment]
Can you run the bash script here https://github.com/azerothcore/azerothcore-wotlk/tree/master/apps/git_tools ? it's to help with writing commits

And thank you for the PR (you said you would open it and you did, a man of his word!)

File: src/server/game/Spells/Spell.cpp (hunk 2 above)
Comment (Shard-MW) [reply src/server/game/Spells/Spell.cpp]: @lineagedr Should be better now

**Comment by BarbzYHOOL** [comment]
@lineagedr another review ? please :P

**Comment by lineagedr** [comment]
> @lineagedr another review ? please :P

The code is the same. Maybe worse than before because now it's unorganized. It's still using dozens of spell ids instead of their mechanic.

**Comment by Shard-MW** [comment]
> @BarbzYHOOL
> The code is the same. Maybe worse than before because now it's unorganized. It's still using dozens of spell ids instead of their mechanic.

@lineagedr 
I agreed with your previous comments, but this one is absurd. 
It seems you don't know how switch statement and spells works.

As I said, _spell->Mechanic is the 4th column of Spell.dbc (https://wowdev.wiki/DB/Spell) and in a lot of cases it's not filled up with a mechanic even if the spell applies one.
So, the best way to check if the spell has a cc mechanic is to use `HasAura` like I'm doing in the default case.
```cpp
// src/server/game/Spells/SpellInfo.cpp     L. 867
bool SpellInfo::HasAura(AuraType aura) const
{
    for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
        if (Effects[i].IsAura(aura))
            return true;
    return false;
}
```
Switch's have the same access time on all elements, so it's better to use it rather than a loop + if / else chaining like `HasAura` does.
We can also remove the default case if we don't want the CCDelay applied on creature's spells as it's mainly configured for the class spells (for PvP).

Moreover, saying "this code is worse" without giving details or solution is completely useless. So please, next time you want to correct someone, make yourself more useful and show him how to do :)

GitVerge:
Checking by ids is really the way to go here in most cases. You don't want some spells to have a delay as it can break them. If you add delay to dual wield by mistake, it will send your offhand weapon to your mail each time you log into the game.

Just a small question, is this affecting spells casted by creatures? Can this break boss encounters?

Shard-MW:
> Just a small question, is this affecting spells casted by creatures? Can this break boss encounters?

@GitVerge 
It does affect spell casted by creatures only if they applies a cc aura like stuns, fear etc. And nope, it'll not break boss encounter, it's just going to add an extra time between when the effect is cast and when it reaches its target.
If you don't want to apply the delay on creature's spells, you can comment/remove the `default` case

GitVerge:
> @GitVerge
> It does affect spell casted by creatures only if they applies a cc aura like stuns, fear etc. And nope, it'll not break boss encounter, it's just going to add an extra time between when the effect is cast and when it reaches its target.
> If you don't want to apply the delay on creature's spells, you can comment/remove the `default` case

Yup I'll do that, thanks.

lineagedr:
@Shard-MW I've already showed you how to do it. And I've said that for the spells that have no mechanic you can use their Ids:

Look at your code:
And you're telling me that this:
```
case 339: // Druid - Entangling roots
case 1062:
case 5195:
case 5196:
case 9852:
case 9853:
case 26989:
case 53308:
```
Is better than this lol:
```
if (_spell->Mechanic == MECHANIC_ROOT)
```
I've rewritten your code: https://pastebin.com/9p5MkDZt

You can see the difference.

Shard-MW:
@lineagedr 
Since a `switch` statement doesn't check for each value, it's better than having a lot of `if`.

And nope, I'm saying that, this : 
```cpp
case 339: // Druid - Entangling roots
case 1062:
case 5195:
case 5196:
case 9852:
case 9853:
case 26989:
case 53308:
```
is better than this : 
```cpp
case SPELLFAMILY_DRUID:
        if (_spell->Mechanic == MECHANIC_BANISH)        // Cyclone
            return delayForBanishes;
        if (_spell->Mechanic == MECHANIC_ROOT)          // Entangling Root
            return delayForRoots;
        if (_spell->Mechanic == MECHANIC_STUN)          // Bash, Maim, Feral charge - Bear
            return delayForStuns;
        if (_spell->Mechanic == MECHANIC_SLEEP)         // Hibernate
            return delayForDisorients;
        if (_spell->SpellIconID == 495)                 // Pounce
            return delayForStuns;
        break;
```
because you'll have to check the familyName + check 5 `if` for Pounce for example.
Read pls [https://www.geeksforgeeks.org/switch-vs-else/](url)

And if you truly want to use CCDelay based on mechanics only, 
```cpp
for (uint8 i = 0; i < CCDArraySize; ++i)
     if (_spell->HasAura(auraWithCCD[i]))
          return auraCCDelay[i];
```
is enough to produce the same result as your entire (re)code.

lineagedr:
> @lineagedr
> Since a `switch` statement doesn't check for each value, it's better than having a lot of `if`.
> 
> And nope, I'm saying that, this :
> 
> ```c++
> case 339: // Druid - Entangling roots
> case 1062:
> case 5195:
> case 5196:
> case 9852:
> case 9853:
> case 26989:
> case 53308:
> ```
> 
> is better than this :
> 
> ```c++
> case SPELLFAMILY_DRUID:
>         if (_spell->Mechanic == MECHANIC_BANISH)        // Cyclone
>             return delayForBanishes;
>         if (_spell->Mechanic == MECHANIC_ROOT)          // Entangling Root
>             return delayForRoots;
>         if (_spell->Mechanic == MECHANIC_STUN)          // Bash, Maim, Feral charge - Bear
>             return delayForStuns;
>         if (_spell->Mechanic == MECHANIC_SLEEP)         // Hibernate
>             return delayForDisorients;
>         if (_spell->SpellIconID == 495)                 // Pounce
>             return delayForStuns;
>         break;
> ```
> 
> because you'll have to check the familyName + check 5 `if` for Pounce for example.
> Read pls [https://www.geeksforgeeks.org/switch-vs-else/](url)
> 
> And if you truly want to use CCDelay based on mechanics only,
> 
> ```c++
> for (uint8 i = 0; i < CCDArraySize; ++i)
>      if (_spell->HasAura(auraWithCCD[i]))
>           return auraCCDelay[i];
> ```
> 
> is enough to produce the same result as your entire (re)code.

No, this:
```
for (uint8 i = 0; i < CCDArraySize; ++i)
     if (_spell->HasAura(auraWithCCD[i]))
          return auraCCDelay[i];
```
Is for ALL spells with certain mods: Including creature spells. 
Where the recode is for player spells and then for other spells

And it's only this:

```
 if (_spell->Mechanic == MECHANIC_ROOT)          // Entangling Root
            return delayForRoots;
```
Not 
```
case 339: // Druid - Entangling roots
case 1062:
case 5195:
case 5196:
case 9852:
case 9853:
case 26989:
case 53308:
```

Shard-MW:
@lineagedr 
Other spells are including creature spells, and you're doing it here...
```cpp
// Other
for (uint8 i = 0; i < (sizeof(auraWithCCD) / sizeof(*auraWithCCD)); ++i)
     if (_spell->HasAura(auraWithCCD[i]))
          return delayForInstantSpells;   
```
And what's the point of using switch + conditions if you're doing the same treatment on player's and other's spells? T_T

lineagedr:
@Shard-MW How is it doing the same treatment? Above the loop the player spells have custom delay. If every player spell was returning the delayForInstantSpells then it would be the same treatment.

Anyway here's a version with switches: https://pastebin.com/TYGLj9eu:

talamortis:
I do understand the way that @Shard-MW has done this, as some spell don't contain mechanics, but i have implented my own version of this into Oregoncore 

https://github.com/talamortis/OregonCore/commit/0b074fccf75cf27085be2547f3c08470ac5b6e95

This way i have done it, reads the mechanics of the spell by using a loop for each spell to check the max effects and then returns the value.

I will go through this code when i have more time for a better review (most likely Friday) and can leave some feedback

Viste:
make a case for each spell - good choice(no). for now best option at @talamortis oregon
btw you can also use spell ATTRS for this ;)

lineagedr:
@Viste If I may ask, which spells does it break?

talamortis:
> 
> 
> make a case for each spell - good choice(no). for now best option at @talamortis oregon
> btw you can also use spell ATTRS for this ;)
> @lineagedr https://github.com/Expecto/Patchs/blob/master/Wotlk(TrinityCore)/CCDelay.diff this is old fucking hack and that hack break some spells, @Shard-MW do the shit more dirty but more right

also are you saying my way is incorrect?

Viste:
> > make a case for each spell - good choice(no). for now best option at @talamortis oregon
> > btw you can also use spell ATTRS for this ;)
> > @lineagedr https://github.com/Expecto/Patchs/blob/master/Wotlk(TrinityCore)/CCDelay.diff this is old fucking hack and that hack break some spells, @Shard-MW do the shit more dirty but more right
> 
> also are you saying my way is incorrect?

no you way is best for now

Shard-MW:
@Viste 
Actually, @talamortis 's code is my default case : 
https://github.com/azerothcore/azerothcore-wotlk/blob/f7c936bcd87eaba9fdb825dd41589d19daaf9520/src/server/game/Spells/Spell.cpp#L6591-L6593

I'm using the first switch to : 
- Don't break non-wanted spells like Priest's [Mind Control] or Hunter traps...
- List all player's spells and allow people to easily config their CCDelay
- Maybe add a config later to apply CCDelay or not, on creature's spells  (default case)
- Because a switch is faster than multiple loops + if/else statements (even if it's micro optimization) since it have the same access time to all his elements

BarbzYHOOL:
The switch here is just a micro optimized layer for specific spell IDs (the most used ones), while the default case is for all other spells. So, if I understand right, Shard-MW method has the best of both worlds

spinzaku:
Can't wait for this! great work @Shard-MW

FrancescoBorzi:
what is the status of this?

Yehonal:
Isn't it quite hacky? have you read the issues on trinitycore?

https://github.com/TrinityCore/TrinityCore/issues/1195

In particular this comment: https://github.com/TrinityCore/TrinityCore/issues/1195#issuecomment-7269783

is there any evidence about the CC Delay for wotlk around? i would investigate more about it.

However as said in TC, the delay comes from the way the packets are handled on retail, but no one think that they are "hard-coded" inside the server.  So you're faking it with this patch

Shard-MW:
@Yehonal 
Yep I've read issues on TC, and yes it's quite hacky since it doesn't reproduce the blizzard's behaviour at identical, but it looks like it.

The CCDelay did existed at the time of WOTLK. We can observe it on many videos etc (see #1968).
Plus, I think this can confirm it :
https://blue.mmo-champion.com/topic/318876-warlords-of-draenor-theorycraft-discussion/#post114

Yehonal:
Imho in this case would be better to create a module for it

BarbzYHOOL:
Why wouldn't blizzard add the delay on purpose ? that's a "feature" too

Yehonal:
> Why wouldn't blizzard add the delay on purpose ? that's a "feature" too

In which way it's a "feature"? 

However the "workaround" would be a module or a script that can be optionally enabled. @Shard-MW you can create an hook in few minutes implementing it as a module and we could also officially include in core (disabled by default) or just reference it in the issues.

The official fix should be a global queue on packets instead. Something that collects all network packets about spell cast on a different unit sending them every 400ms...so you can have a random delay of 0-400ms on every cast.

@Rochet2 @mik1893 anything to say? ;)

Shard-MW:
This is a feature because it activates the cross-cc, which actually existed at the time of Wotlk and should also exist on AC. 

Yes I can try to do a module, but I want to say that even if the CCDelay isn't very important for average players, it becomes primordial for advanced players because it can be used to break setups if it's well used. That's why I think it should be enabled by default on AC. 

Personal opinion : I don't think that fixing the CCDelay by setting up a "0-400ms global queue on packets" to get closer to the operation of Wotlk, to the detriment of the fluidity of the game, is attractive for players ... Even though PR isn't the "official" way to do it, I think it's still a good compromise between activating cross-cc and maintaining actual fluidity.

Yehonal:
If you create a module we could directly add it inside the AC project as
"officially supported module" and it can be enabled by default (but
disabled on demand too)
You could adapt this PR eventually.

Module would be better instead of directly inject the code in core

We should do same thing also for features like ahbot and such.

So we could start from it.

Il Ven 26 Lug 2019, 00:18 Shard <notifications@github.com> ha scritto:

mik1893:
my 2 cents: i did a lot of research and work on CC Delay back in the past when i was "running my own operation".
We eventually ended up implementing the CC Delay.
Reality is, it was just part of the global spell batching at the time wotlk was out, it wasn't an "intended" mechanic. 

The absolute dream would be to implement spell batching as it is also now in modern wow. with a configurable batch time window.
The mechanic is very well explained here: https://www.wowisclassic.com/en/news/spell-batching-classic-wow-confirmed/

Spell batching still exists in wow, as of now. In trinity core we basically have a "diff" which is after how many milliseconds is the status of things updated - the spell batching would be kind of a "sub-diff" that only applies to spell execution.

Yehonal:
Ok so the "queue" i've had in mind is called "spell batching" and was an official mechanism. Thanks for the research @mik1894!
It should not be difficult to implement imho. And a configurable option would be great

BarbzYHOOL:
From non technical eyes, it looks like the same thing to me but with a config option for the 400ms = even better :D

mik1893:
@Yehonal yes correct, it is a queue of actions. The weird thing is that because it impacts spells but not gameobject movement, that will generate some behaviours the classic players are familiar with but might confuse people. E.g. a spell evaluation happening before cast will happen at time X, but actually spell is cast at X + 400 where the player conditions changed - but still the spell will land.
Checks that are done AFTER target hit will instead be correctly based on when the action happens.

In WOTLK anyway, this "diff" time is not 400 MS. after a lot of research on videos by a group of tester in my old project, we came out with 150 to 200 ms as the correct number.

Yehonal:
> If you create a module we could directly add it inside the AC project as "officially supported module" and it can be enabled by default (but disabled on demand too) You could adapt this PR eventually. Module would be better instead of directly inject the code in core. We should do same thing also for features like ahbot and such things. So we could start from it.

after i've said it, why you closed it?

i don't understand why it's so difficult to move code in modules for you...we've such amazing possibility to keep the core clean and fixed (with module workarounds) at the same time...meh

BarbzYHOOL:
i feel like it's a git accident but not sure yet

Shard-MW:
@Yehonal Keep calm, it's just an accident. I'll reopen it and all my others PR. I've just deleted some of my branch by mistake...

Yehonal:
Any news about the module? :)

pklloveyou:
I used to have about a month and it feels pretty good now what state it is

Yehonal:
It must be done with a module, it's easy, safer and the best way to do it.
We should stop to add more and more custom options directly into the core.

BarbzYHOOL:
what mikk said looks to me like it was an official feature as I expected, and not really "custom"

however if it's considered custom, @Shard-MW could make it into a module, but please answer something

Yehonal:
The feature is official and it's not "a real feature". It's just a matter
of how the retail handled the spell packets, with a window queue of 400 ms
that allowed players to play with certain "strategies". We don't have such
queue at the moment and the spells packets are instant.
The code of this PR is just an hack to emulate such queue. That's why it's
better to provide a module

BarbzYHOOL:
Ok I understand

Shard-MW:
I'll do it, I just don't have time yet
But, about :
> This must be moved under a module that can be used as a workaround until the queue will be implemented

I don't think it's a good idea to implement the 400ms batch cuz players will feel the game more slower

Kitzunu:
Any update?

Goatform:
Here are some interesting sources:

https://github.com/magey/classic-warrior/wiki/Spell-batching

https://github.com/Evolvee/Spell-Delay-Research/issues/1

Blizzard forum:
Screenshot_10

Voxstrasza:
Any updates on this?

PR #1971, opened 2019-06-14
fix(Core/Spells): Correct ID of Warlock spellscript

poszer:
Use correct spell ID for Shadow Ward (Warlock spell)

##### CHANGES PROPOSED:

-  Set correct ID of warlock spell Shadow Ward.  It was set to -7235(Shadow Protection) which is wrong
https://wotlk.evowow.com/?spell=7235
-  Correct is to use -6229 (Shadow Ward)
https://wotlk.evowow.com/?spell=6229
- Also updated DB `world` -> `spell_script_names` to match these changes

###### ISSUES ADDRESSED:

Closes 

##### TESTS PERFORMED:

Yes, everything fine.

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1560493947655900000.sql
+++ b/data/sql/updates/pending_db_world/rev_1560493947655900000.sql
@@ -1,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1560493947655900000');
+
+UPDATE `spell_script_names` SET `spell_id`=-6229 WHERE `ScriptName`='spell_warl_shadow_ward';
--- a/src/server/scripts/Spells/spell_warlock.cpp
+++ b/src/server/scripts/Spells/spell_warlock.cpp
@@ -1310,7 +1310,7 @@ class spell_warl_health_funnel : public SpellScriptLoader
         }
 };
 
-// -7235 - Shadow Ward
+// -6229 - Shadow Ward
 class spell_warl_shadow_ward : public SpellScriptLoader
 {
     public:


data/sql/updates/pending_db_world/rev_1560493947655900000.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1560493947655900000');
+
+UPDATE `spell_script_names` SET `spell_id`=-6229 WHERE `ScriptName`='spell_warl_shadow_ward';
```
BarbzYHOOL:
```suggestion
UPDATE `spell_script_names` SET `spell_id`=6229 WHERE `ScriptName`='spell_warl_shadow_ward';
```

lineagedr: @BarbzYHOOL We need that minus to check for all ranks of the spell. Without it the spell script will only work for rank 1

poszer: @BarbzYHOOL no, it must be with - to check ranks. Just like @lineagedr explained

BarbzYHOOL: ok thank you, i suspected it was not a mistake but i wasn't sure :P

BarbzYHOOL:
I'm wondering why such a simple spell requires its own script. I'm sure there are values for everything in the DBC (and so the server knows how to interpret them, normally)

lineagedr:
> I'm wondering why such a simple spell requires its own script. I'm sure there are values for everything in the DBC for that spell, even the shadow resist (and so the server knows how to interpret them, normally)

I believe it needs it's own spell script because of:

sw

Issue #1096, opened 2018-12-01
Venture Co. Strip Miners Don't Melee Players Half The Time


wetbrownsauce:
##### SMALL DESCRIPTION:
Venture Co. Strip Miners in Stranglethorn often don't run to the player when aggro'd. 

##### EXPECTED BLIZZLIKE BEHAVIOUR:
When a Venture Co. Strip Miner is aggro'd, it throws dynamite and then runs to the player to melee them.

##### CURRENT BEHAVIOUR:
When a Venture Co. Strip Miner ( https://wowgaming.altervista.org/aowow/?npc=674 ) is aggro'd from a distance, they will do one of 3 things: 
1) Throw dynamite, and then resume their mining animation far from the player until the dynamite cooldown is finished.
2) Throw dynamite, and then move toward the player in a fragmented way, alternating stopping to mine and running until the player is reached. 
3) Run to the player and melee them. 

Maybe similar issue to https://github.com/azerothcore/azerothcore-wotlk/issues/540 ?

##### STEPS TO REPRODUCE THE PROBLEM:
1.  .go xyz -12981 -343 9 0
2.  Aggro a few Venture Co. Strip Miners.
3.  Try running away from them.

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
50aace0e52218ec7aa97a2b592140b30b192c996

##### OPERATING SYSTEM:
Windows 10

poszer:
Confirmed

BarbzYHOOL:
look if TC has fixed it :P

wetbrownsauce:
Just tested on a fresh compile of Trinity and a similar problem exists. Over there, the Strip Miners will run to the player to melee them as expected, however, as soon as they use their throw dynamite ability, they stop in place and never attempt to pursue the player just like here.

BarbzYHOOL:
Ok, then it requires to modify the smart_ai in the database, probably a "good  first issue" to fix

PR #1973, opened 2019-06-14
fix(DB/Creature): Venture Co. Strip Miners combat movement

poszer:
Venture Co. Strip Miners combat movement

##### CHANGES PROPOSED:

-  Fix combat issue with Venture Co. Strip Miners adding react state. They should move, and attack players
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1096

##### TESTS PERFORMED:

Yes, they move, they attack, they cast.

Read https://github.com/azerothcore/azerothcore-wotlk/issues/1096

- [ ]
- [ ] 

##### Target branch(es):

Master


Knindzagxg:
Seems good, had no errors in DB so far.

FrancescoBorzi:
good job @poszer and @Knindzagxg

Stoabrogga:
Why is the goblin attacking after it's react state is set to "passive"? (action_type "8", action_param1 "0")
https://github.com/azerothcore/azerothcore-wotlk/blob/d739289f6883759af1acf584414f2c2ef1d71976/data/sql/updates/pending_db_world/rev_1560521436127285100.sql#L12
Is this a bug? Looks very wrong to me, although it works.

poszer:
> Why is the goblin attacking after it's react state is set to "passive"? (action_type "8", action_param1 "0")
> https://github.com/azerothcore/azerothcore-wotlk/blob/d739289f6883759af1acf584414f2c2ef1d71976/data/sql/updates/pending_db_world/rev_1560521436127285100.sql#L12
> 
> 
> Is this a bug? Looks very wrong to me, although it works.

Screenshot_1

Stoabrogga:
@poszer 
Sadly your screenshot does not help clarifying this strange behaviour. It seems to me that your SQL script is based on a bug, because if you apply react state "passive" on respawn the goblins act as expected (ignoring the player), if applied in combat they attack. Could be a problem in the future if this bug is solved, because this will most likely break this SAI.

BarbzYHOOL:
Ok, apparently the flag was removed by Poszer (;-)) in the first commit but as Stoabrogga said, the rest is superfluous

poszer:
> Please replace your SAI script with this single update statement:
> 
> ```sql
> UPDATE `smart_scripts` SET `event_param1` = 0, `event_param2` = 0, `action_param2` = 1, `comment` = 'Venture Co. Strip Miner - In Combat - Cast Throw Dynamite' WHERE `entryorguid` = 674 AND `source_type` = 0 and `id` = 0;
> ```
> 
> Explanation:
> 
> * The cause of the original issue was combat movement on cast, see cast flag 64 in the wiki: http://www.azerothcore.org/wiki/smart_scripts
> * It has nothing to do with react state at all, so makes no sense to set it to passive (which is obviously bugged when used "in combat")
> * I also set the initial timer to 0, so the goblins will throw dynamite at the beginning of the fight

Are they moving ?

poszer:
> Ok, apparently the flag was removed by Poszer (;-)) in the first commit but as Stoabrogga said, the rest is superfluous

Nah, this issue is too old. I didn't touch default script. I just added react state since the Goblins didn't move when entering in combat

Stoabrogga:
> Nah, this issue is too old. I didn't touch default script. I just added react state since the Goblins didn't move when entering in combat

That is not correct. The original script had 64 as cast flag, here's the dump (see "11,7978,**64**"):
```sql
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`)
VALUES
(674,0,0,0,0,0,100,0,2000,6000,12000,15000,0,11,7978,64,0,0,0,0,2,0,0,0,0,0,0,0,0,'Venture Co. Strip Miner - In Combat CMC - Cast \'Throw Dynamite\'');
```
Your script did set it to "0":
https://github.com/azerothcore/azerothcore-wotlk/blob/d739289f6883759af1acf584414f2c2ef1d71976/data/sql/updates/pending_db_world/rev_1560521436127285100.sql#L13
So the react state just did nothing, which is on second thought not surprising when changing it during combat. So changing the cast flags was the actual solution here.

poszer:
> > Nah, this issue is too old. I didn't touch default script. I just added react state since the Goblins didn't move when entering in combat
> 
> That is not correct. The original script had 64 as cast flag, here's the dump (see "11,7978,**64**"):
> 
> ```sql
> INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`)
> VALUES
> (674,0,0,0,0,0,100,0,2000,6000,12000,15000,0,11,7978,64,0,0,0,0,2,0,0,0,0,0,0,0,0,'Venture Co. Strip Miner - In Combat CMC - Cast \'Throw Dynamite\'');
> ```
> 
> Your script did set it to "0":
> https://github.com/azerothcore/azerothcore-wotlk/blob/d739289f6883759af1acf584414f2c2ef1d71976/data/sql/updates/pending_db_world/rev_1560521436127285100.sql#L13
> 
> 
> So the react state just did nothing, which is on second thought not surprising when changing it during combat. So changing the cast flags was the actual solution here.

Okay. Anyway interesting and i think you re wrong, since when i added that react state, actually it fixed problem with these creatures (Movement in combat / Attacking). And you saw it already : 
https://github.com/azerothcore/azerothcore-wotlk/pull/1973#issuecomment-502454191
Without that, they re stucked in one place, not attacking players. So what is wrong there dunno, but setting it to 0 make creatures to attack.
PS it's fixed now, and thanks for this

Stoabrogga:
Nope, sorry to prove you wrong. I tried the following script, which adds the react state "passive" in combat (like you did in your script) to the already existing spell casting and leave the cast flags to "64":
```sql
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`)
VALUES
(674,0,0,0,0,0,100,0,0,0,0,0,0,8,0,0,0,0,0,0,1,0,0,0,0,0,0,0,0,'Venture Co. Strip Miner - In Combat - Set React State \'Passive\''),
(674,0,1,0,0,0,100,0,2000,6000,12000,15000,0,11,7978,64,0,0,0,0,2,0,0,0,0,0,0,0,0,'Venture Co. Strip Miner - In Combat CMC - Cast \'Throw Dynamite\'');
```
As expected the react state changes nothing (goblins do not melee), the issue was only about cast flag "64" which prevents melee combat (this flag is meant for pure casters). As for my comment https://github.com/azerothcore/azerothcore-wotlk/pull/1973#issuecomment-502454191 I was wrong there because I did miss the cast flag in the first place.

poszer:
> Nope, sorry to prove you wrong. I tried the following script, which adds the react state "passive" in combat (like you did in your script) to the already existing spell casting and leave the cast flags to "64":

I know that the script you shared is correct.
But i'm asking how it's working when we add this to the script, since action param is set to 0 : 

https://github.com/azerothcore/azerothcore-wotlk/blob/d739289f6883759af1acf584414f2c2ef1d71976/data/sql/updates/pending_db_world/rev_1560521436127285100.sql#L12

Also i have tested it, and it change the way how creatures attack ( WIthout this they won't attack players ) 
- Adding it to the script, it fixes issue.
As you can see, i have tested it, and that's why i open PR with this, bcoz i didn't find another way.
Also, Knindza tested it, and he said also it's fine, which means it fixed issue with Goblins not attacking.
Is something wrong in the core about this? Since i already did before one PR almost like this, which fixes issue with Sons of Hodir quest, posible it will need edit too ( Even if it work without problem)

Stoabrogga:
It wasn't the react state which fixed this, but setting the cast flags to "0" in your first commit:
https://github.com/azerothcore/azerothcore-wotlk/pull/1973/commits/d739289f6883759af1acf584414f2c2ef1d71976
Perhaps you did this by accident, I don't know, but this was the actual fix. I learned now that you cannot set the react state in combat, so the first entry in your script was without effect the whole time. I was mistaken here also because I did not see that you changed the cast flags of the spell. If you really would like to change the react state you have to do this out of combat, e.g. on respawn or on reset. When setting it to "passive" the effect would be that the creature won't get into combat anymore, even if you run into aggro range or attack it.

So for me everything is working correctly, there is also no bug in my opinion:
- changing react state in combat has no effect at all
- cast flag 64 ensures that the creatures only cast spells and prevents combat movement (the goblins still execute their emote state while waiting for their spell cooldown, which looks a bit silly)

Here are all relevant values concerning the cast flags (you can combine multiple flags):

Name | Flag | Hex | Comment
-- | -- | -- | --
SMARTCAST_INTERRUPT_PREVIOUS | 1 | 0x01 | Interrupt any spell casting
SMARTCAST_TRIGGERED | 2 | 0x02 | Triggered (this makes spell cost zero mana and have no cast time)
CAST_AURA_NOT_PRESENT | 32 | 0x20 | Only casts the spell if the target does not have an aura from the spell
SMARTCAST_COMBAT_MOVE | 64 | 0x40 | Prevent combat movement on cast, allow on fail range, mana, LOS

If you set the cast flags to "0" for a spell it means that it does not interrupt other spells. For the goblins it doesn't matter if the cast flags are set to "0" or "1" as they only have one spell to cast.

BarbzYHOOL:
@poszer @Stoabrogga thank you guys!

PR #1992, opened 2019-06-21
fix(Core/Spell): Haunted Memento

Stoabrogga:
##### CHANGES PROPOSED:
As it is now the Haunted Memento just spawns a ghost who stands there, does nothing, and also never despawns. Through this PR the Spell "Haunted" (53768) is changed so it will spawn a ghost who will follow the player around slowly for a few seconds and then despawns. It will respawn every few minutes as long as the player has the aura "Haunted".

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested successfully in-game

Just add the item:
```.additem 40110```

none

##### Target branch(es):
Master


Knindzagxg:
Tested, everything seems to be good.

Sometimes it will go little underground like other NPC's but that's nothing special.

Stoabrogga:
Great, thx @Knindzagxg :+1:

PR #2002, opened 2019-06-23
fix(DB/Creature/Event): Arena Vendors & Creatures

poszer:
Adding missing Arena Vendors , linking creatures with Game Event ID's:
* ID: 55 - Arena Season 3
* ID: 56 - Arena Season 4

##### CHANGES PROPOSED:

- Adding two missing creatures Drolig (ID:27722) and Drelik Blastpipe (ID: 27721) in Shattrah, They re located in the Scryer's Tier and Aldor Rise. Using free Guid's from DB (Guids: 75475, 79018).
- Linking Arena vendors from the season 3 & 4 to game event.

###### ISSUES ADDRESSED:
 We are missing two creatures Drolig (ID:27722 - Scryer's Tier) and Drelik Blastpipe (ID: 27721 - Aldor Rise) spawns in Shattrah. Also, others Vendors are always visible to players which is wrong. They should be linked with the Game_Event - Seasons. If the season(event) is active, the creatures should be visible too. If the season(event) is inactive they shouldn't be visible to players.

Closes 

##### TESTS PERFORMED:

Yes, query imported without errors. Creatures spawned at correct positions, visible while the event is active-Invisible while the event is inactive.

1. Log with any character, and start the event using command 
* .event start 55
* .event start 56
2. Check these creatures guids  .go creature 70996,75914,79018,75475,88156
3. They should be visible to players.

4. Now repeat same steps, but with commands
* .event stop 55
* .event stop 56
5. Check same creatures guids  .go creature 70996,75914,79018,75475,88156 
6. They shouldn't be visible to players.

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1561274158622782000.sql
+++ b/data/sql/updates/pending_db_world/rev_1561274158622782000.sql
@@ -1,0 +1,16 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1561274158622782000');
+
+-- Adding two missing creatures Drolig and Drelig Blastpipe in Shattrah, They re located in the Scryer's Tier and Aldor Rise
+DELETE FROM `creature` WHERE `guid` IN (79018,75475);
+INSERT INTO `creature` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `modelid`, `equipment_id`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecs`, `spawndist`, `currentwaypoint`, `curhealth`, `curmana`, `MovementType`, `npcflag`, `unit_flags`, `dynamicflags`, `ScriptName`, `VerifiedBuild`) VALUES
+(79018, 27722, 530, 0, 0, 1, 1, 0, 0, -2116.46, 5386.03, 53.8074, 0, 300, 0, 0, 6986, 0, 0, 0, 0, 0, '', 0),
+(75475, 27721, 530, 0, 0, 1, 1, 0, 0, -1877.61, 5650.35, 127.457, 1.45691, 300, 0, 0, 6986, 0, 0, 0, 0, 0, '', 0);
+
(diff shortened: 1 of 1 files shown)


Knindzagxg:
All good mate.

Tho, on Blizz, in later expansions, those vendors are always visible due Transmog that have been add later.
Maybe you can add something like module for those to be available always, without event.

poszer:
> All good mate.
> 
> Tho, on Blizz, in later expansions, those vendors are always visible due Transmog that have been add later.
> Maybe you can add something like module for those to be available always, without event.

Thanks mate for quick test and suggestion. As you can see, the linked creatures are using (guids) so there won't be problem with other creatures (To show them).

solidmaxtor:
¡ALERT!
these npc vendor should not be available in all expansion wotlk
In blizzlike mode is not avaliable  3.0.0 to 3.3.5.
they should only be added by custom servers, not in the main branch of azerothcore

poszer:
> ¡ALERT!
> these npc vendor should not be available in all expansion wotlk
> In blizzlike mode is not avaliable 3.0.0 to 3.3.5.
> they should only be added by custom servers, not in the main branch of azerothcore

Alert for what? xD And how you mean? As i said before, this is guid of creature linked to Arena Event, this creatures isn't spawned on one place, it's spawned on 2 or 3.
For example Vixton Pinchwhistle <Arena Vendor> This NPC can be found in Blade's Edge Mountains,Shadowmoon Valley,Champions' Hall,,Tanaris. And we have linked just one ''GUID'' to Event. Same is done with others creatures from this PR.
They exist, they re visible, just one creature from that entry isn't.

solidmaxtor:
Sorry if you express me wrong XD
The season 3 set is not available in wotlk, the sellers of your season should not be available.
Other sellers yes but without selling season 3 set.
Correct me if I'm wrong please.
I know since I played wotlk official, many dinosaurs years ago

PR #2027, opened 2019-06-29
feat(Core/Pool): improve pool manager

Stoabrogga:
##### CHANGES PROPOSED:
This is basically the implementation of this TC commit: https://github.com/TrinityCore/TrinityCore/commit/022575329ff6766485ccddb2053820326a6664bf :
- use "std::unordered_map" instead of "std::vector" in order to reduce memory usage (not much, perhaps 5-10%)
- identify missing pool entries in pool_template via error messages in the server log

Concerning the pool entries I applied the following fixes as they would now cause errors:
- Remove pool entry 5217 from pool_gameobject (only 1 game object "Saronite", not needed here)
- Remove pool entry 1047 from pool_creature (only 1 creature "Hematos", not needed here)
- Add a few missing Saronite nodes to pool_template

###### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested successfully in-game

- without applying the SQL script: check the startup errors
- after applying the SQL script the startup errors should disappear
- loot game objects and wait until respawn
- check RAM consumption before / after applying this PR

none

##### Target branch(es):
Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1561816769087112416.sql
+++ b/data/sql/updates/pending_db_world/rev_1561816769087112416.sql
@@ -1,0 +1,16 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1561816769087112416');
+
+-- Only one game object "Saronite", pool entry is not needed
+DELETE FROM `pool_gameobject` WHERE `pool_entry` = 5217;
+
+-- Only one creature "Hematos" (8976), pool entry is not needed
+DELETE FROM `pool_creature` WHERE `pool_entry` = 1047;
+
+-- Add missing Saronite nodes to pool template
+DELETE FROM `pool_template` WHERE `entry` IN (5450,5506,5608,5517);
+INSERT INTO `pool_template` (`entry`,`max_limit`,`description`)
+VALUES
+(5450,1,'Icecrown 189980, node 3'),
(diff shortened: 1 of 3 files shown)


Knindzagxg:
Tested.

No errors in DBlog -> All good.
Love the part for RAM part, working great as well -> All good. (For me was over 10% tho Debian 9)

Nice one Sto :D

wowmane:
Does it reduce memory usage?

wowmane:
@Stoabrogga, You're Great, Thank You.

I tested something:
When the character enters a new area, the amount of RAM increases, If core usage ram 20.2  change to 21.0 / Just for one area!

If this can be removed , very good for reduce memory.

test on linux; debian.

I'm ready to test, If you add commit.
I think, this is fixed in the trinitycore.

Stoabrogga:
@Knindzagxg 
Thanks for testing this. I also applied the changes of this PR several days ago without any issues. Reduces RAM a bit.

@wowmane 
If you enter a new area new grids will be loaded and this takes RAM. Grids will only be unloaded if no player is there anymore for a specified time (see "GridCleanUpDelay" in worldserver.conf, default 5 minutes). But this has nothing to do with this PR. This PR is only about pools.

wowmane:
@Stoabrogga 

Very good, please Also apply other issues to reduce RAM !

PR #2037, opened 2019-07-02
fix(Core): Online status of accounts

Deetribs:
The status of the accounts in the database is always "online" after disconnection.

##### TESTS PERFORMED:

Yes, everything fine.

Log out of your account, the online column will change to 0 after one minute :)

##### Target branch(es):

Master

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -126,7 +126,7 @@ WorldSession::WorldSession(uint32 id, WorldSocket* sock, AccountTypes sec, uint8
         m_Address = sock->GetRemoteAddress();
         sock->AddReference();
         ResetTimeOutTime(false);
-        LoginDatabase.PExecute("UPDATE account SET online = %u WHERE id = %u;", realmID, GetAccountId());
+        LoginDatabase.PExecute("UPDATE account SET online = 1 WHERE id = %u;", GetAccountId());
     }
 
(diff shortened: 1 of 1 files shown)


src/server/game/Server/WorldSession.cpp
```diff
@@ -126,7 +126,7 @@ WorldSession::WorldSession(uint32 id, WorldSocket* sock, AccountTypes sec, uint8
         m_Address = sock->GetRemoteAddress();
         sock->AddReference();
         ResetTimeOutTime(false);
-        LoginDatabase.PExecute("UPDATE account SET online = %u WHERE id = %u;", realmID, GetAccountId());
+        LoginDatabase.PExecute("UPDATE account SET online = 1 WHERE id = %u;", GetAccountId());
```
FrancescoBorzi: was this suppose to track in which realm the account is online?

FrancescoBorzi:
thanks @Deetribs for your contribution

src/server/game/Server/WorldSession.cpp
Deetribs: I don't think it serves to know in which realm the account is connected. But it's true we can keep that.
https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Server/WorldSession.cpp#L146

Stoabrogga:
tested successfully :+1:

PR #2072, opened 2019-07-07
fix(DB/Disables): Disable the Spell Saronite Bomb in Eye of Eternity …

### LINKED ISSUE #1948: [Item in Dungeon] If use Saronite Bomb in Malygos  the platform is destroy - Video

Vaex:
…& Trial of the Crusader Maps. This prevents destroying them

##### CHANGES PROPOSED:

-  Disable the Trigger Missile Spell from Saronite Bomb for Eye of Eternity and Trial of the Crusader

-  Both Platforms are safe now from the Damage 

###### ISSUES ADDRESSED:

Closes #1948

##### TESTS PERFORMED:

works as it should

Go to Eye of Eternity and Trial of the Crusader and use: https://www.wowhead.com/item=41119/saronite-bomb

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1562509195181388404.sql
+++ b/data/sql/updates/pending_db_world/rev_1562509195181388404.sql
@@ -1,0 +1,6 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1562509195181388404');
+
+-- Disable the Saronite Bomb Spell for Eye of Eternity and Trial of the Crusader Maps https://www.wowhead.com/item=41119/saronite-bomb
+DELETE FROM `disables` WHERE `sourceType` = 0 AND `entry` = 61766;
+INSERT INTO `disables` (`sourceType`, `entry`, `flags`, `params_0`, `params_1`, `comment`) VALUES
+(0, 61766, 17, '616,649', '', 'Disable Spell Saronite Bomb for Eye of Eternity & Trial of the Crusader');


GMKyle:
like i say in pr https://github.com/azerothcore/azerothcore-wotlk/pull/1964 To prevent damage we can change gameobject faction to 35(friendly to all), the life increased  is not needed.
Disabling spell is also valid of course, but we most testing others spells that do siege damage.

Vaex:
there are around 57 Spells which use SPELL_EFFECT_87 aka Siege Damage. Most of them are not obtainable trought an item. Most of them are for vehicles which you mount up or for Quest Items. Changing the Faction does the job aswell but those database values are most of the time sniffed, they should be as they be for my opinion :)

GMKyle:
you are right about siege damage spells but Global Thermal Sapper will need the same "validation".
Sniffed data it's important but.....193070: -- Nexus Raid Platform in azerothcore has no flags and no faction, in trinitycore has 32 | 1375, witch one is correct? answer It's not big deal, witch one works better?
it's the real deal in my opinion :P
Also notes that players cannot use saronite bombs(and other items if you include more) in those places without any reason "apparently"

solidmaxtor:
Work fine, Saronite bomb not cause damage.

Vaex:
@GMKyle yes correct :) TC seems to be more or less good

well just two maps be affected and they be small, this should be not a problem. Wintergrasp doesnt matter, Ulduar the same but Trial of the Crusader and Nexus shoud be disabled due to encounter problems.

Thanks @solidmaxtor for testing

GMKyle:
@Vaex  I simply don't like this approach :P

Andrei-UPB:
Given this issue is one I worked to fix as well on my local server, I can say that this approach is worse than both the suggested alternatives on every level:

1. Not only you restrict a spell usage because "the damage still shows" while using one of the two proposed fixes.
2. You also leave room for other similar spells to still work and break the platforms (Kyle mentioned Global Thermal Sapper Charge.
3. This also restricts the usage of the Saronite Bombs making them unable to fulfill their intended purpose (which is AoE damage on hostile enemies) for no reason (seemingly, to players) at all.

The best fix for this issue in my opinion is either changing the gobject's faction to 35 , or increase the object's HP up to a point where it is theoretically impossible for the object to drop below 0 HP.

I felt like my PR #1964 , and on top of it Kyle's comments would've been enough to develop a proper fix and close the issue.

In ToC's case, the Argent Coliseum Floor (195527) already has it's Data0 (Health Points) in gameobject_template set to 10000000 , implying someone else worked on it in the past and decided on going with the high health pool numbers (either because gameobject_template_addon wasn't a thing back then or the dev working on it did not consider faction from gameobject_template_addon)

The ideal fix for both EoE and ToC platforms should therefore be:

```
UPDATE `gameobject_template_addon` SET `faction`='35' WHERE  `entry`=193070; -- Nexus Raid Platform set faction 35 from 0

UPDATE `gameobject_template_addon` SET `faction`='35' WHERE  `entry`=195527; --  Argent Coliseum Floor set faction 35 from 0
```

Vaex:
@Andrei-UPB then in any case the option with the fraction change. Increase HP just to avoid something is not a solution and looks not really good at all.

solidmaxtor:
You're wrong Andrei, the damage from the saronite bomb is no longer displayed.
Increase the life of the platform causes that it does not break when it must, I test it and to put 1.5mll slow to break until the meeting ended and the looting boss

zabustak:
This is 9999 Level Hack Fix.. don't hate me..

**You can do better than that**

zabustak:
@solidmaxtor
You can put how much life you want at this platform even tens of millions, but you have to change the instance_eye_of_eternity.cpp

>         void ProcessEvent(WorldObject* /*unit*/, uint32 eventId)
        {
            switch(eventId)
            {
                case 20158:
                    if (GameObject* go = instance->GetGameObject(GO_PlatformGUID))
                        if (Creature* c = instance->GetCreature(NPC_MalygosGUID))
                        {
                            go->ModifyHealth(-Put here how much damage you want to give to the platform, c);
                            go->EnableCollision(false);
                        }
                    break;
            }
        }

Andrei-UPB:
Fine, have it your way :)

If I ever find a mob that can be slowed but shouldn't, i'll make sure to disable Frostbolt as it's the "elegant" fix and looks good.

GMKyle:
Ok guys this are my conclusions:
I checked both instances code(no changes needed) and used my own recommendation (change faction), all working good and players can use saronite bombs and ....etc on platforms(they not take damage). It's an easy testing so please try yourselve and stop non sense discussions

Vaex:
alright guys i fully agree. Disable a spell on a entire Map is used for other purposes. Well tbh this usage of bombs is not daily but we still saw someone using them on Malygos by accident or purpose.

i agree fully to change the Faction to 35 but please do not increase life like this, it's totally a nonsense

PR #2078, opened 2019-07-09
fix(DB/SAI): Technician Zhanaa's script on quest turn-in

Selyria:
**CHANGES PROPOSED:**

The fix should cause the event to start after the quest The Emitter ID 9312 has ended. 

The old smart_script is deleted because it does not belong to the NPC.
It belongs to the NPC with the ID 16514 which also has the script.

Before Fix https://youtu.be/US-QM6TTKiE
After Fix https://youtu.be/tZ1vvufOaac

**ISSUES ADDRESSED:**
none

**TESTS PERFORMED:**
tested successfully in-game

**HOW TO TEST THE CHANGES:**

.go c 7978
.quest add 9312
Pass the quest and look at the holographic emitter.
The script should now start

KNOWN ISSUES AND TODO LIST:
none

**Target branch(es):**
Master


Stoabrogga:
Tested successfully 👍

poszer:
Thank you for PR :+1:

PR #1730, opened 2019-04-19
[Hunter] Exotic pets, talents and Stables

Pondaveia:
##### CHANGES PROPOSED:

- Dismissed Exotic pet can't swap talents unless swapping to a tree with Beast Mastery
- Exotic non dismissed pet can't swap talents unless swapping to a tree with Beast Mastery

###### ISSUES ADDRESSED:

Closes #1621

##### TESTS PERFORMED:

Compiled and tested on Windows

##### HOW TO TEST THE CHANGES:

Create a hunter, get Beast Mastery from talents and:

1. Swap to a talent tree with no Beast Mastery
2. Do the same with Dismissed pet
3. Swap talents to a tree with Beast Master with and without pet

##### KNOWN ISSUES AND TODO LIST:

- [ ] Stables not displaying the dismissed active pet
- [ ] Taming a new beast with the current dismissed

##### Target branch(es):

Master

PR #2100, opened 2019-07-15
fix (Core/Hunter) Stable Master, exotics pets management, tame validations.

### LINKED ISSUE #1621: [Hunter] Pet Disappear [$20]
[Hunter] Pet Disappear (Priority Hight)

##### SMALL DESCRIPTION:
Hunter pets disappear when the try to call stable pet or tame new beast with active dismissed pet.

When hunter have a dismissed pet and try to tame a beast or call stable pet with spell or stable master , message must appear and can't be possible to do this action

##### CURRENT BEHAVIOUR:
Actually , Hunter with dismissed pet can Tame a beast and call other pet in stable , and the dismissed pet equiped by the hunter disappear.
You can see more information in this post : 
https://github.com/WoWManiaUK/Redemption/issues/2853 (related to trinitycore)

1. Make a Hunter with a pet equipped and dismiss the pet
2. Try to open stable or Tame a new beast
3. Your pet dismissed disappear after this actions.

##### OPERATING SYSTEM: Debian 9

<bountysource-plugin>

---
There is a **[$20 open bounty](https://www.bountysource.com/issues/71835162-hunter-pet-disappear?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1914: Hunter: Pets disappearing when interacting with stable or on BM -> MM/SV specialization change.
Hunter: Pets disappearing when interacting with stable or on BM -> MM/SV specialization change.

##### SMALL DESCRIPTION:

When changing the current pet with another from the stable via Call Stabled Pet or by changing specializations with a dismissed pet, you permanently lose the pet that's in your "Current Pet" stable slot.

You should under no circumstance lose any stabled pets whenever performing permutations with the Call Stabled Pet spell ( https://wotlk.evowow.com/?spell=62757 ) or interacting with a Stable Master NPC.

The "Current Pet" slot should still be filled by the  "current pet" you were using even after dismissing it, as there is a difference between dismissing your pet (un-summoning it) and abandoning it (PERMANENTLY deleting it).

##### CURRENT BEHAVIOUR:

Dismissing the pet (1), being mounted while using the Stable Master NPC (2) or switching specializations (3) removes said pet from the current pet stable slot (the slot appears empty), you can drag another pet from your stable roster into your current pet slot and lose the first one in the process instead of the pet swap going through, replacing the initial pet with one of your choosing from your four stable slots.

To be able to reproduce I recommend a fully decked pet roster, let's consider C as the current pet, S1 , S2, S3 and S4 as our stabled pets.

1. Have your current pet active (IE: not dismissed, not mounted) and accessing the stables in one of the two ways described above yields the following result:
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      C                            S1              S2                 S3            S4

``` 
2. Attempt to swap some pets around, let's say we want to take S4 out. We drag S4 into the Current Pet slot, thus Stable Slot 4 will be taken by C.
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      S4                            S1              S2                 S3            C

``` 
This is the INTENDED BEHAVIOUR and how it should ALWAYS work.

The screenshots below will answer most of the questions, but i'll describe it with a little graphic as well.

In the current behaviour section i've typed (1), (2), and (3) in parentheses , I'll post a how-to-reproduce for each of them.

(1). Dismissing the pet and swapping with another from the stable via Call Stabled Pet ( https://wotlk.evowow.com/?spell=62757 ) or interacting with a Stable Master NPC ( I used NPC entry 10050 , but you can use any stable master).

1. Dismiss Pet ( https://wotlk.evowow.com/?spell=2641 ) your current pet and accessing the stables in one of the two ways described above yields the following result:
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
   <empty>                         S1              S2                 S3            S4

``` 
2. Attempt to swap S4 to the Current Pet position. We drag S4 into the Current Pet slot, thus Stable Slot 4 SHOULD be taken by C.
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      S4                            S1              S2                 S3        <empty>

``` 
3. Stable Slot 4 is not occupied by the current (dismissed) pet C because the game interprets that as no pet existing, therefore stable slot 4 will be empty, and you will lose pet C forever.
4. You can repeat these steps until you've lost all but one of your five pets.

(2). The same way to reproduce as (1), except you don't have to dismiss your pet but be mounted (tested with a  regular ground mount) and interact with a Stable Master NPC (tested with https://wotlk.evowow.com/?npc=10050 ) , mounting has no effect over swapping pets via Call Stabled Pet because using the spell dismounts you, therefore i considered this a second way to replicate the bug.

1. Mount up and access the stables by talking to the Stable Master and choosing the "I'd like to stable my pet here." gossip, you will be greeted by this menu:
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
   <empty>                         S1              S2                 S3           <empty>

``` 
2. Attempt to swap S3 to the Current Pet position. We drag S3 into the Current Pet slot, thus Stable Slot 3 SHOULD be taken by S4, who became our current pet in the last reproduction cycle for (1). The result can be seen below:

```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      S3                            S1              S2                <empty>        <empty>

``` 
3. Stable Slot 3 is not occupied by the current pet S4 because you are mounted and the game interprets that as you not having a pet active, therefore stable slot 3 will be empty, and you will lose pet S4 forever.

(3). This issue is related to swapping specializations from Beast Mastery to Marksmanship or Survival and losing your exotic pet in the process.

1. Have an exotic beast as your current pet ( I tamed Skoll https://wotlk.evowow.com/?npc=35189 for this purpose ).
2. Swap specializations from Beast Mastery to Marksmanship or Survival.
3. Open your stables by using Call Stabled Pet or talking with a Stable Master NPC.
4. Your current pet slot is now empty and upon taking another pet from the stable you lose your exotic beast forever akin to (1) and (2). How this should work is you should be able to see your exotic pet in your current slot, but redded-out (unusable) and be able to move it in an empty stable slot or swap with an existing non-exotic pet usable in Marksmanship or Survival specializations
##### SCREENSHOTS:
I've added three imgur albums describing each of the issues mentioned above (1), (2), (3).
(1). https://imgur.com/a/rmh5CN0
(2). https://imgur.com/a/w7SxgdM
(3). https://imgur.com/a/1XCAvRF
##### EXTRA NOTES:

There's an earlier issue pertaining this but I found out after I already wrote and documented all of this :( , The earlier issue  can be found here : https://github.com/azerothcore/azerothcore-wotlk/issues/1621 

**If there is a way to merge both issues that would be fantastic as I would not want this to go to waste.**

I believe the fix might be related to a way to store the dismissed pet in the current pet slot  even post dismissal or when you are mounted instead of it becoming empty upon dismissal / mounting up.

I have no means of testing this on any other server application at the current time. I'll try to download a clean TC and test there.
##### BRANCH(ES):

master

##### AC HASH/COMMIT:

b67ea15
##### OPERATING SYSTEM:

Ubuntu 18

##### MODULES:

Transmogrification module, but I doubt there's anything related to that.

##### OTHER CUSTOMIZATIONS:

Yes, I did modify some database entries to fix other various bugs, however they should be unrelated.

Thank you in advance for looking over this, the issue seems daunting at first but I am sure that if there's a possibility to somehow store the current pet in it's slot even when dismissed or the player character is mounted, this bug will be resolved and no pets will ever be lost again.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74859839-hunter-pets-disappearing-when-interacting-with-stable-or-on-bm-mm-sv-specialization-change?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GMKyle:
thx to https://github.com/cmangos/mangos-wotlk , I re-used some logics and a couple of functions 

##### CHANGES PROPOSED:
- Improve https://github.com/azerothcore/azerothcore-wotlk/pull/1730
- Proper pet states management in stable master (dead, tempunsummon(player mounted), dismissed)  
-  If a hunter change to a talent tree with no Beast Mastery he can store his exotic pet in stable master
- Improve tame beast spell validation(proper manage pet states)
- Implement and Using SMSG_PET_TAME_FAILURE in some cases

Closes #1914
Closes #1621

##### TESTS PERFORMED:
Build without errors and tested in-game. (Win10 server)

Stable master tests:
Please read https://github.com/azerothcore/azerothcore-wotlk/issues/1914 explain all very well and include images 

Tame beast tests:
- Make a Hunter and tame some beast
- You will not be able to tame other beast, if your pet is dead, tempunsummon(player mounted), or dismissed; as intended.

**Also notes error messages in all tests** 

##### Target branch(es):

Master


Andrei-UPB:
Will look into it tomorrow and test it out, thank you in advance for the work Kyle.

src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -97,6 +97,56 @@ void Pet::RemoveFromWorld()
...
+    
+    if (!result)
+        return SPELL_FAILED_NO_PET;
+
+    Field* fields = result->Fetch();
+
+    uint32 petentry = fields[1].GetUInt32();
+    uint32 savedHealth = fields[10].GetUInt32();
+    uint32 summon_spell_id = fields[15].GetUInt32();
+    PetType petType = PetType(fields[16].GetUInt8());
```
FrancescoBorzi: not important, but for simplicity you can just use `auto petType = PetType(fields[16].GetUInt8());`

it's a good practice to use `auto` instead of repeating the type name twice, especially when they get quite long

src/server/game/Entities/Player/Player.cpp
```diff
@@ -27553,3 +27556,26 @@ void Player::SummonPet(uint32 entry, float x, float y, float z, float ang, PetTy
...
+
+    return false;
+}
+
+void Player::SendTameFailure(uint8 result)
+{
+    WorldPacket data(SMSG_PET_TAME_FAILURE, 1);
+    data << uint8(result);
+    SendDirectMessage(&data);
+}
```
FrancescoBorzi: since `Player.cpp` is already HUGE and editors have already troubles editing it, would it be possible to move this logic somewhere else?

src/server/game/Entities/Pet/Pet.cpp
GMKyle: done!!

src/server/game/Entities/Player/Player.cpp
GMKyle: it's possible but i can't figure out where, ``` Player.h``` perhaps? Remember this is used also in ``` spell_hunter.cpp```

Vaex: it could be also added in PetHandler or Unit.cpp

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -18484,6 +18484,14 @@ void Unit::NearTeleportTo(float x, float y, float z, float orientation, bool cas
     }
 }
 
+//GMKyle: placed here because Player.cpp is overcrowded
```
FrancescoBorzi: we can remove this comment

GMKyle: it's a joke obviously :P

src/server/game/Entities/Player/Player.cpp
GMKyle: done!!

src/server/game/Entities/Unit/Unit.cpp
GMKyle: done!!

FrancescoBorzi: thank you :)

src/server/game/Entities/Player/Player.cpp
Vaex: great :)

Andrei-UPB:
@GMKyle As promised, I tested https://github.com/azerothcore/azerothcore-wotlk/pull/2100 , and i bring both good and bad news.

This fix solves the issue but only in part. You can move your current dismissed pet into the stable and it works as intended. The problem (I believe it's an oversight and easily fixable) is if instead of moving the dismissed pet into the stables, you move one of the stabled pets into the current pet slot, and it gets overwritten.

To clarify:
The pet C is your current pet, but is dismissed for case (1) and (3) or considered dismissed because the player is mounted for (2).
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      C                                S1           S2               S3             S4
```
If you swap over C with  S1, it works as it should and you get the following result:
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      S1                             C              S2           S3               S4
```
But if you swap over S1 with C, you get the following result:
```
|Current Pet|                |Stable Slot 1| |Stable Slot 2| |Stable Slot 3| |Stable Slot 4|
      S1                              S2            S3           S4           <empty>

```

C -> S1 is working fine,
S1 -> C is not and still leads to the current pet being lost after the transfer.
The current pet is also listed as level 0 if dismissed (minor, shouldn't matter much).
Cross-posted from #1914  for easier readability.

valissa:
@Andrei-UPB it's kyle 
thx for testing. I'm out for vacations for 3-4 weaks, I must have missed something because in my local server all works fine. I will look over this. ... eventually :P

GMKyle:
I'm back @Andrei-UPB after run a few tests i can't reproduce the behavior that you described :(  like i already said, in my local server all works fine, also note that pet isn't dismissed when players is mounted. We need someone else test this.

Stoabrogga:
@GMKyle 
I think you have to sync your PR with the master. For example, the flag "CREATURE_TYPEFLAGS_EXOTIC" is now called "CREATURE_TYPE_FLAG_EXOTIC_PET".

Stoabrogga:
@GMKyle 
Thanks. I'm currently occupied with other tasks, but when I get the time I'll take a deeper look at this.

One thing, though: The main idea here is to save dismissed pets with slot 100 (PET_SAVE_NOT_IN_SLOT) in the DB so that they don't get lost, right?

GMKyle:
@Stoabrogga 
Trying to explain....xD
The stable masters malfunction occurs when player' pet is dismissed  (PET_SAVE_NOT_IN_SLOT = 100) and  when player is mounted (normally PET_SAVE_AS_CURRENT = 0 but can also be PET_SAVE_NOT_IN_SLOT = 100), because pet isn't accessible/returned  by player->GetPet function; finally also note that pet can be dead and being despawned by "un-natural" causes (player gets mounted insted go far away  from pet corp for e.g.) 
Hunter BM spec is also involve because change from BM spec to MM cause current exotic pet (if any) being dismissed

Stoabrogga:
Ok, I played around a little bit and it works quite well. I also could not reproduce the issue @Andrei-UPB described. But there's one thing which is strange: If the pet is dismissed and you access your stable multiple times you get different values for the pet level, e.g.:
pet1
pet2
pet3

Seems like sometimes the value is not initialized, could you please look into this?

pklloveyou:
Tested, perfect.

Stoabrogga:
Ok, everything seems well so far, so can be merged in my opinion.

BarbzYHOOL:
thank you all, especially gm-kyle

Issue #2094, opened 2019-07-15
Quest Ending Their World an event should start [$30]


Selyria:
##### SMALL DESCRIPTION:
After you have completed the quest Ending Their World an event should start

##### EXPECTED BLIZZLIKE BEHAVIOUR:
After you have completed the quest Ending Their World an event should start

https://youtu.be/EJWnyEpUTms?t=590

After submitting Ending Their World, the quest should be 	
Start The Unwritten Prophecy and you only need it when Velender almost stands behind you

##### CURRENT BEHAVIOUR:

When you give Ending Their World away, nothing happens after that.

##### STEPS TO REPRODUCE THE PROBLEM:
Give yourself the Quest 9759
.Quest Complet 9759
.go c 63379
Submit quest

##### BRANCH(ES):
Master

##### AC HASH/COMMIT:
AzerothCore rev. c47391bc94a1+ 2019-07-13 06:08:10 +0000 (master branch) (Win64, Release)

##### OPERATING SYSTEM:
Windows 10

##### MODULES:
mod-transmog
mod-world-chat

w5860363:
https://github.com/TrinityCore/TrinityCore/issues/18642 
https://github.com/TrinityCore/TrinityCore/pull/14149/files

PR #2102, opened 2019-07-16
fix(DB/Quest): Ending their world

poszer:
Ending their World, Quest NPC's. (Port from TC)

##### CHANGES PROPOSED:

-  On quest complete serveral npcs should spawn near the questender
-  

###### ISSUES ADDRESSED:

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2094

##### TESTS PERFORMED:

Yep. Once quest is completed at quest ender, the creatures will be spawned outside.

1. Give yourself the Quest 9759
2. .Quest Complet 9759
3. .go c 63379
4. Submit quest

- [ ]
- [ ] 

##### Target branch(es):

Master


Knindzagxg:
Tested in-game, seems ok.

Tho, sometimes they drop under texture, but i guess that's old movemap issue.

poszer:
> Tested in-game, seems ok.
> 
> Tho, sometimes they drop under texture, but i guess that's old movemap issue.

Yea, posible, Thanks for testing

Stoabrogga:
@poszer 
You are overwriting existing SAI for the following creatures:
17101,17116,17117,17214,17215,17240,17241,17242,17246,17311

This SAI could be of use for other quests or events but is not contained in your PR anymore and could break already working functionality. Please check the existing SAI before applying your PR. It is also important to check all quests or events those creatures are involved after applying this PR in order to avoid regression.

I don't recommend merging this PR as long as this is not sorted out.

poszer:
> @poszer
> You are overwriting existing SAI for the following creatures:
> 17101,17116,17117,17214,17215,17240,17241,17242,17246,17311
> 
> This SAI could be of use for other quests or events but is not contained in your PR anymore and could break already working functionality. Please check the existing SAI before applying your PR. It is also important to check all quests or events those creatures are involved after applying this PR in order to avoid regression.
> 
> I don't recommend merging this PR as long as this is not sorted out.

Well, just checked all of these SAI from AC database (Almost all of them are for emotes and things like that, just one for Action List ). But Nevermind, will leave than just the Exarch Admetius SAI which doesn't have SmartAI on AC. And will leave someone to sort all of those quests related to this event.

Stoabrogga:
I propose to set this PR in WIP and implement the event the right way without damaging the original behaviour of the NPCs. I will take a look at this when I have the time.

poszer:
> I propose to set this PR in WIP and implement the event the right way without damaging the original behaviour of the NPCs. I will take a look at this when I have the time.

This is fine. Only one SAI is in commit, When the quest is completed, these creatures will be spawned outside, which is correct. For others creatures (spawned) you can open new PR, and solve them

Stoabrogga:
@poszer 
I restored your original script and fixed several bugs:
- the original SAI of the NPCs is not touched anymore, instead the new SAI events are added (this would have caused much trouble if your original SQL script would have been merged)
- reset the script of Exarch Admetius after 5 minutes (otherwise you cannot trigger a new event until the server is restarted)
- ensure that "Cookie" McWeaksauce is only following his waypoint path if he is spawned, not when he is summoned (otherwise he would have just walked away from the party)
- add an additional SAI to set Prophet Velen active so that the summoned creatures near him will despawn even if no player is near anymore
- fixed wrong NPC names in the comments

Issue #2106, opened 2019-07-18
Scourgelord Tyrannus died mid-air. Body got stuck mid-air and can't be looted


andoys:
Scourgelord Tyrannus died mid-air. Body got stuck mid-air and can't be looted

##### SMALL DESCRIPTION:

Scourgelord Tyrannus died mid-air. Body got stuck mid-air and can't be looted

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Scourgelord Tyrannus body should drop down to the floor, so he can be looted

##### CURRENT BEHAVIOUR:

Sometimes you can kill Scourgelord Tyrannus mid-air. Making his body stuck mid-air and can't loot

![midair](https://user-images.githubusercontent.com/18629476/61439873-ec767f00-a974-11e9-9433-7bde11b1704f.png)

##### STEPS TO REPRODUCE THE PROBLEM:

1.  Kill Scourgelord Tyrannus mid-air
2. Stuck body mid-air
3. Can't loot

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

69c7fefad60a75cb2047d336287c5c55bec4d7fa

##### OPERATING SYSTEM:

 Ubuntu 16/18

solidmaxtor:
@ArashGorge  Maybe this need you script =)

ArashGorge:
> @ArashGorge Maybe this need you script =)

https://github.com/ArashGorge/azerothcore-wotlk/pull/7/files

just i forget to make rp at AC :X

PR #2122, opened 2019-07-22
fix(Core/Dungeon): Scourgelord Tyrannus died Body got stuck and can't be looted

ArashGorge:
Description: https://github.com/azerothcore/azerothcore-wotlk/issues/2106

##### CHANGES PROPOSED:

-  Teleport corpse to the ground
-  

###### ISSUES ADDRESSED:

Closes 
description https://github.com/azerothcore/azerothcore-wotlk/issues/2106

##### TESTS PERFORMED:

Tested in game
wind 10 vs 2017

- [ ]
- [ ] 

##### Target branch(es):

Master

### DIFF
--- a/src/server/scripts/Northrend/FrozenHalls/PitOfSaron/boss_scourgelord_tyrannus.cpp
+++ b/src/server/scripts/Northrend/FrozenHalls/PitOfSaron/boss_scourgelord_tyrannus.cpp
@@ -191,6 +191,7 @@ public:
                 pInstance->SetData(DATA_TYRANNUS, DONE);
             if (me->IsSummon())
                 me->ToTempSummon()->SetTempSummonType(TEMPSUMMON_MANUAL_DESPAWN);
+			me->NearTeleportTo(me->GetPositionX(), me->GetPositionY(), float(628.1), me->GetOrientation()); 
         }
 
         void KilledUnit(Unit* who)


solidmaxtor:
Tested many times for sucess in air  jaja
Great gob ArashGorge

Stoabrogga:
@talamortis 
Should we also revert the changes to "Noth" and the "Assembly of Iron" which were merged here?
bad8390dee2fba50a711a26e6537fbdbb0d8863d

They use a similar technique.

talamortis:
Yes.

Once the proper fix is applied then you will no longer get this error with creatures.

Stoabrogga:
@talamortis 
Is there someone working on a correct fix for these issues?

ArashGorge:
This is not possible! Because if you do this there may be a problem elsewhere and this problem exists only a few places
Some bosses fight in the air and it can't be done like the last boss of Dungeon Oculus, the (Ley-Guardian Eregos),its about 10 years we are playing 3.3.5, and we know all problems

Stoabrogga:
@ArashGorge 
Ok, so to understand this correctly: If this is fixed on a general level other bosses who should not fall would then fall through the map, which is also a problem, correct?

ArashGorge:
@Stoabrogga 
We have been playing around with these problems for a long time, the number of bosses that need to get down to the ground, I have fix a lot of them, and I have a few left over,
Bosses that should not fall and should be on the list may not be remembered, and we may have problems later with them

Stoabrogga:
@talamortis 
I'd say we merge this. I don't think this can be solved on a general level as some bosses need to fall and others don't. Also, I don't think it is a hackfix if you provide a specific action for a creature in it's appropriate CreatureScript.

talamortis:
No you add a check on inhabit type for the creature so if it can fly then return else fall to ground.

FrancescoBorzi:
if this doesn't break anything, I would still merge it (it's only affecting `boss_scourgelord_tyrannus.cpp`)

Yehonal:
Have someone tested it on TC? Is the issue present even there? (If not maybe they fixed it and we can check there) otherwise of course would be nice a general fix but i would also check some retail video to understand how that bosses used to fall down to the ground officially.

pklloveyou:
if (me->GetPositionZ() > 632.0f)
	me->NearTeleportTo(me->GetPositionX(), me->GetPositionY(), 628.157f, me->GetOrientation());

Stoabrogga:
I read a few things about this encounter on retail in 2009 / 2010 and I'm not sure about merging this anymore: On retail if you damage the boss during the roleplay event he just mounts Rimefang and you can no longer defeat him:
https://wow.gamepedia.com/Scourgelord_Tyrannus#Bugs
https://www.wowhead.com/npc=36658/scourgelord-tyrannus#comments:id=915952

To me this looks like an exploit if you kill him before he even gets the chance to fight back. So if you can't wait for him to finish his talk and jump down from his dragon you deserve getting no loot in my opinion.

ArashGorge:
> I read a few things about this encoun .. . .  in my opinion.

That's right, but it's not bad to fix !

Stoabrogga:
> 
> That's right, but it's not bad to fix !

Sure, but it should be fixed without applying an exploit. I think the correct way to handle this would be to render Tyrannus invulnerable until he leaves his dragon.

ArashGorge:
> > That's right, but it's not bad to fix !
> 
> Sure, but it should be fixed without applying an exploit. I think the correct way to handle this would be to render Tyrannus invulnerable until he leaves his dragon.

It's good idea, i'll check it tomorrow

ArashGorge:
New Fix :
Close as https://github.com/azerothcore/azerothcore-wotlk/pull/2209

PR #2162, opened 2019-07-31
chore(README.md): add Discord badge

FrancescoBorzi:
This gives more visibility to our chat, since some users struggled to find it.


poszer:
Could you please add this too.

[Average time to resolve an issue](http://isitmaintained.com/project/azerothcore/azerothcore-wotlk "Average time to resolve an issue")


or this / and this ? Just edit my comment, and copy/paste


[Percentage of issues still open](http://isitmaintained.com/project/azerothcore/azerothcore-wotlk "Percentage of issues still open")

FrancescoBorzi:
As discussed in chat, I'm not sure if this is a good idea since we use issues also as TODOs and random ideas, so for things that are not necessarily meant to be closed ASAP.  But that's just my opinion.

Anyway, I think this should be discussed in a separate PR so we can merge this meanwhile.

PR #2179, opened 2019-08-05
Core/AI: Add DoCastSelf helper. 

Viste:
##### CHANGES PROPOSED:
- add helper function for scripts that allows creatures and bosses easy cast spell on self
if you an idiot and cant read simple code do this:
create simple script with npc who cast some spell on self through  DoCastSelf functions
then look how its work 
##### Target branch(es):master


FrancescoBorzi:
This is a one-line addition that is not going to affect anything else and cannot break anything of the current functionality.

So, in this case, there is no need to test.

Kaev:
I don't really see a need for this but the code itself is fine.

Viste:
@Kaev for example https://github.com/azerothcore/azerothcore-wotlk/pull/2180/files#diff-911e1a4c9af665f500f91e471d9719d8R65
this is more easy to understand for new people and and more convenient in general.

PR #2194, opened 2019-08-12
fix(Core): Fix build - C3848 compile error

mik1893:
Fix compile error C3848 - expression having type 'type' would lose some const-volatile qualifiers in order to call 'function'
Some compilers will throw this errors, some won't (e.g. VS 2013 doesn't, VS 2019 does.)


Stoabrogga:
Merged, as it is not necessary to test anything except for the build. Thanks @mik1893 !

PR #2205, opened 2019-08-14
fix(Core/Worldsession): add option to prevent logout when AFK in a sanctuary zone

### LINKED ISSUE #2139: Feature Request: Allow people to AFK/Away in City without getting disconnected to character selection screen.  [$10]
Feature Request: Allow people to AFK/Away in City without getting disconnected to character selection screen.

Can be Module 
or
Core Edit (maybe configurable by worldserver.conf to enable/disable)
if this is clientside then no problem if someone can do it also

Currently there's no option for this.

This config is for people who are away/afk in character selection screen:
```
#    SocketTimeOutTime
#        Description: Time (in milliseconds) after which a connection being idle on the character
#                     selection screen is disconnected.
#        Default:     900000 - (15 minutes)
SocketTimeOutTime = 900000
```

<bountysource-plugin>

---
There is a **[$10 open bounty](https://www.bountysource.com/issues/77756323-feature-request-allow-people-to-afk-away-in-city-without-getting-disconnected-to-character-selection-screen?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mik1893:
Closes #2139 

This will prevent logout by sending a reason for denying logout - preventing CMSG_LOGOUT_REQUEST  to be spammed again immediatly.


andoys:
yessssssss!!!!! great work @mik1893

pklloveyou:
TIM截图20190814184435
Success. great job.

pklloveyou:
Is there a way to remove the”message"?

mik1893:
Yes - will complete the fix tonight

andoys:
This is the packet issue @lineagedr  was talking about the yellow spam message

poszer:
Not needed on master

FrancescoBorzi:
@poszer 

> Not needed on master

it has been requested here: https://github.com/azerothcore/azerothcore-wotlk/issues/2139

So yes, it is needed.

Also it is configurable and disabled by default, so it doesn't harm at all to have it.

poszer:
> @poszer
> 
> > Not needed on master
> 
> it has been requested here: #2139
> 
> So yes, it is needed.
> 
> Also it is configurable and disabled by default, so it doesn't harm at all to have it.

Good Luck than

pklloveyou:
The message is most likely generated by the client

mik1893:
yes. it's client generated so nothing to do about it :|
I thought it was some AFK text but apparently it's just coming out from the client

Stoabrogga:
Code looks ok, so is this still WIP or is it ready to be merged?

racknarox:
One question is there any way to modify it so that it not only works in the sanctuary, but throughout the world?

racknarox:
One question is there any way to modify it so that it not only works in the sanctuary, but throughout the world?

pklloveyou:
Does the server send a move or jump instruction to the client?

Stoabrogga:
> 
> 
> One question is there any way to modify it so that it not only works in the sanctuary, but throughout the world?

Should be easy to implement, perhaps we could adjust this PR for this:
- use another parameter name, e.g. "PreventAFKLogout", and make it an int instead of a bool
- values:
  0: normal behaviour
  1: prevent AFK logout in sanctuaries
  2: prevent AFK logout everywhere
- in the code use something like this:
  ```cpp
  bool preventAfkSanctuaryLogout = sWorld->getIntConfig(CONFIG_AFK_PREVENT_LOGOUT_SANCTUARY) == 1
      && GetPlayer()->isAFK() && sAreaTableStore.LookupEntry(GetPlayer()->GetAreaId())->IsSanctuary();

  bool preventAfkLogout = sWorld->getIntConfig(CONFIG_AFK_PREVENT_LOGOUT_SANCTUARY) == 2
      && GetPlayer()->isAFK();
  ```
  .
  .
  .
  ```cpp
  else if (preventAfkSanctuaryLogout || preventAfkLogout || GetPlayer()->duel || GetPlayer()->HasAura(9454)) // is dueling or frozen by GM via freeze command
      reason = 2;
  ```

Stoabrogga:
> 
> 
> Does the server send a move or jump instruction to the client?

As far as I understand this the client sends a "logout" request to the server (e.g. if too long AFK or triggered by the player) and the server can deny or accept the logout. You can also do nothing about the yellow message because this is written by the client, not the server.

pklloveyou:
@Stoabrogga Can the server send"move"or"jump"instructions to the client?

Stoabrogga:
> 
> Can the server send"move"or"jump"instructions to the client?
> Server regularly send "move" or "jump" to the client, the client release AFK state.

I don't think this is possible. Also to my understanding the client sets the AFK state, so it would not matter if the server forces actions on the player. But I don't have deep knowledge concerning the client / server communication, so could be totally wrong here.

Stoabrogga:
@mik1893 
Are you still working on this? What do you think about extending the parameter like I mentioned above?

Stoabrogga:
I added the 3rd option and tested this successfully:
- option 0: AFK logout is not prevented at all (default)
- option 1: AFK logout is only prevented in Sanctuaries (like Dalaran or Shattrath)
- option 2: AFK logout is prevented in all zones

As this PR does not cause any harm (option 0 is default) and it is a nice addition to the core I think this is ready to be merged.

abracadaniel222:
I love this feature. One small issue with this feature (which I can totally live with) is that after the client attempts to log out due to AFK and the server denies it, the user can't exit the game manually unless they clear the AFK status first by moving the character. 

Like I said, this is not a problem to me but can potentially make people confused as to why they can't log out, if they aren't aware of what is going on.

PR #2227, opened 2019-08-22
chore(authserver): improve startup console logs

FrancescoBorzi:
Changes:

- improved the error shown when the authserver port is invalid (showing the name of the parameter and the actual invalid value)
- showing the bind ip address and the port that the authserver is listening to

There is no need to test since these are cosmetic changes that just add some logs, there is no way they break anything (and I have tested them locally).


BarbzYHOOL:
yes ofc it's good!

PR #2247, opened 2019-08-30
Core/Vehicle: delay position update in Vehicles, this may be deleted

Winfidonarleyan:
##### CHANGES PROPOSED:
-  Port from TC https://github.com/TrinityCore/TrinityCore/commit/3532173ee340573be937b4c5799c155fa2cc906c

##### TESTS PERFORMED:
- [x] Test in game
- [x] Test in build

- See https://github.com/TrinityCore/TrinityCore/issues/17634

### DIFF
--- a/src/server/game/Entities/Vehicle/Vehicle.cpp
+++ b/src/server/game/Entities/Vehicle/Vehicle.cpp
@@ -498,19 +498,25 @@ void Vehicle::RelocatePassengers()
 {
     ASSERT(_me->GetMap());
 
+    std::vector<std::pair<Unit*, Position>> seatRelocation;
+    seatRelocation.reserve(Seats.size());
+
     // not sure that absolute position calculation is correct, it must depend on vehicle pitch angle
-    for (SeatMap::const_iterator itr = Seats.begin(); itr != Seats.end(); ++itr)
+    for (auto const& itr : Seats)
     {
-        if (Unit* passenger = ObjectAccessor::GetUnit(*GetBase(), itr->second.Passenger.Guid))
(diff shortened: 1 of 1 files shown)


pklloveyou:
Before the test, and after the test, are not found problems

FrancescoBorzi:
good job @Winfidonarleyan 

thanks a lot @pklloveyou !!! you are our testing star!

pklloveyou:
> Before the test, and after the test, are not found problems
Translation error

No issues were found before and after the merge

FrancescoBorzi:
yes, no worries, we got it! :)

pklloveyou:
Okay.

Yehonal:
This PR can be merged

Stoabrogga:
I just tested **without** this PR by going through the steps mentioned here:
https://github.com/TrinityCore/TrinityCore/issues/17634
There was no crash, so why do we need this fix if it does not fix anything?

Winfidonarleyan:
That's a need to ask @Yehonal.

Viste:
@Stoabrogga sunwell have a bit different vehicles system(its only based on TC)

Stoabrogga:
Hmm, ok, so if I understand this correctly: Although we don't suffer the crash it's better to update the passenger's position outside of the the loop which calculates the position?

Yehonal:
RelocatePassengers is a "redundant" method because when you call it the UpdatePosition of Unit is called and it calls again the RelocatePassengers.

So if you don't store the calculation of passenger position before Update it could be some wrong behaviour and in some rare case it can cause a crash.

Stoabrogga:
Ok, now I understand. Thanks for explaining this.

PR #2281, opened 2019-09-11
Core/Opcode: simple opcode filter to prevent send non-existing opcode

### LINKED ISSUE #2170: Exploit Crash spell ( Malformed Packet ) [ $20 bounty ]
hello
i have a crash on my source , that seems to be an exploit 
this is my crashlogs 
https://gist.github.com/blackdev01/827699f9682ff12155cf770530e09623
https://gist.github.com/blackdev01/88d91dfedfbf789f2e847f7966c7a2a6
https://gist.github.com/blackdev01/fafda762ee8e2fc1a15bb269740fe906
https://gist.github.com/blackdev01/5f7b2cd3d4191227b06027ed257f1ea7

this my rev :  https://github.com/azerothcore/azerothcore-wotlk/commit/a9b981d619c220d7459a963cac4e989215b638b1
os : debian 8
i saw an report about this problem , i think my problem is look like this one 

this post is a bounty post i'll pay 20$ for this
my Discord : ProGrammer#8649

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/78174280-exploit-crash-spell-malformed-packet-20-bounty?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Viste:
#### CHANGES PROPOSED:
-  add simple(hacky) opcode filter 
##### ISSUES ADDRESSED:
- ~Closes 2170(partially)~ and all other or possible issues with sending random packets to world server
##### Target branch(es):
- [x] Master


Helias:
Hey, good job man.
How can I test the PR?

Viste:
Send non existing opcode to server

BarbzYHOOL:
to send opcode, I don't know if @Stoabrogga made his tutorial yet but he sent me this, so you can try this (with an erroneous opcode):

```
You create a text file "opcode.txt" in the working directory of the server and then can execute it with:
.debug send opcode


Tried it with this content for opcode.txt:
631 uint32 6146


And it correctly played another background music on the client.
but be careful.... if executing .debug send opcode without "opcode.txt" you'll crash the server (it will take more and more RAM until the OS kills the process)


opcodes are here:
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Server/Protocol/Opcodes.h
```

Stoabrogga:
@BarbzYHOOL 
This won't work, as the debug command is just about sending an opcode to the client. We need the other way round here.

BarbzYHOOL:
ok I didn't read viste's comment correctly :D

then how do you proceed?

Stoabrogga:
I tried to find something which could send custom opcodes to the server but sadly was not successful.

BarbzYHOOL:
Maybe with some lua functions

@Kaev @Rochet2 @Yehonal guys, any idea on how to send opcode from client to server?

Ali-Khazaee:
@BarbzYHOOL https://github.com/Lillecarl/WowAddin

Kaev:
Sending a custom OpCode is only possible if you inject custom code, e.g. via the posted WowAddin or if you use a tool that works as a proxy and will let you modify the packet content before sending it to the server.

Helias:
Tested with AccLeito addon, it doesn't work.

FrancescoBorzi:
> Closes #2170(partially) and all other or possible issues with sending random packets to world server

> I tested this PR with AccLeito Addon, but the fix doesn't stop it. Still crashing.

@Viste what kind of attacks have you tested this PR against from?

Also... are we sure all possible valid opcodes are known to the core? If not, I wouldn't risk kicking people around randomly

Helias:
I'm thinking the following scenario:
- if a player is playing WoW and his client sent an opcode that we do not have in our enum list where the last value is **NUM_MSG_TYPES**, we could  kick a player that is just playing WoW

this is the worst case where this PR affect a bad behaviour, it doesn't need, because the core already ignore the opcode that are not in the enum list

Helias:
I updated the PR, removing the kick, but leaving the console log errors about receiving an unknown opcode.

BarbzYHOOL:
what about verifying that the core knows all opcodes ? then kick if not valid ?

why not ? seemed good, now it just logs to console

Viste:
@BarbzYHOOL core knows all needed for 335a opcodes :)

BarbzYHOOL:
then it should kick if not valid, no?

Viste:
@BarbzYHOOL yes its not my rework for just log i make kick & log

Helias:
I removed it because I don't know if the core knows really all the client command, for this reason I removed the kick.
Anyway the core delete the packet if it's invalid, so no-one can make any damage to the core.

PR #2283, opened 2019-09-12
feat(Core/Config): Add parameter to load all grids of all non-instanced maps on server start

Stoabrogga:
##### CHANGES PROPOSED:
- Add a new parameter "PreloadAllNonInstancedMapGrids" to the worldserver configuration which can be used to load all grids of all non-instanced maps on server start.
- Remove the parameters "GridUnload" and "GridCleanUpDelay" from the configuration file as they were never implemented or used by AC and only confuse people (including me :rofl:).

I took the "LoadAllCells" function from TC:
https://github.com/TrinityCore/TrinityCore/blob/537c4d1208384f924a144180abd3f7c3cf305a6d/src/server/game/Maps/Map.cpp#L244-L249

##### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested successfully in-game

Set the parameter "PreloadAllNonInstancedMapGrids" to 1 in the configuration and start the server. You need at least 12 GB RAM to test this (perhaps more on Windows), as preloading the maps will need an additional amount of ca. 9 GB RAM. On my test server the RAM usage is normally about 2 GB, with preloading the maps it increases to 11 GB.

none

##### Target branch(es):
- [x] Master
 

Stoabrogga:
Ok, I think I know what happens here: If the grids are loaded the SAI scripts are also loaded and executed, which means that creatures which are set active via SAI (like the Fel Reavers in Hellfire Peninsula) start their movement at once. This is actually a small additional benefit of loading all grids.

LichKing255:
> Ok, I think I know what happens here: If the grids are loaded the SAI scripts are also loaded and executed, which means that creatures which are set active via SAI (like the Fel Reavers in Hellfire Peninsula) start their movement at once. This is actually a small additional benefit of loading all grids.

Yes that's exactly what I meant

Stoabrogga:
I also added this information to the parameter description in the configuration file.

pklloveyou:
TIM截图20190914143138
Terrible.

Stoabrogga:
> 
> Terrible.

It's a lot of RAM, yes. But that's what this is all about: Caching data in RAM.

FrancescoBorzi:
> Terrible

It's not, the data will be just loaded faster when you need it since it's already in RAM :)

pklloveyou:
Got it.

Stoabrogga:
@pklloveyou 
As you tested this: Did you also play in-game with preloading enabled?

pklloveyou:
Enabled, no error was found in the process

PR #2286, opened 2019-09-13
Delete azerothcore motd

tails402:
Delete this server runs on azerothcore.


Viste:
@FrancescoBorzi @BarbzYHOOL

Issue #2155, opened 2019-07-30
[Exploit] allowing to know a gm was online


wowmane:
Description:
You can known a gm was online and invisible if he enters a public channel allowing ownership, and someone else enters that channel and gm leaves, for example test

Current behaviour: Tell us what happens
ownership changes from gm to player at gm leave/disconnect

Expected behaviour: Tell us what should happen instead
ownership must change from gm to player on player enter if gm was 1st on that channel and invisible.

Steps to reproduce the problem:

1. gm enters channel test invisible.
2. players enters channel test.
3. gm leaves channel or disconnects.
4. players get ownership, and now he knows a gm was invisible inside that channel.

REV: https://github.com/azerothcore/azerothcore-wotlk/commit/c7156c787fe8156f094090cb87a575151b31561d
SYSTEM: Windows 2017
Moudels: Anticheat

PR #2298, opened 2019-09-21
fix(Core/Chat): Channel exploit

wizzymore:
##### CHANGES PROPOSED:
-  Checking if the owner of the channel is a GM and is invisible

##### ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2155

##### TESTS PERFORMED:
- Tested by using 2 accounts, gm one and player one

Using the gm account join a channel:
/join CHANNEL_NAME
Using the player account join the same channel:
/join CHANNEL_NAME
Leave the channel with the GM account, it shouldn't say nothing.
Do the same order of joining a channel but leave with the player first ( who should now be the owner even tho he joined second ) and you will see a text that the GM ( the only player in the channel ) is the owner.

##### Target branch(es):
- [x] Master

##### Operating System
 - [x] Windows
 - [x] Linux

 

 
### DIFF
--- a/src/server/game/Chat/Channels/Channel.h
+++ b/src/server/game/Chat/Channels/Channel.h
@@ -201,7 +201,6 @@ class Channel
         void UnBan(uint64 guid);
         void Password(Player const* player, std::string const& pass);
         void SetMode(Player const* player, std::string const& p2n, bool mod, bool set);
-        void SetOwnerGM(Player const* player, bool on);
         void SetOwner(uint64 guid, bool exclaim = true);
         void SetOwner(Player const* player, std::string const& name);
         void SendWhoOwner(uint64 guid);
--- a/src/server/game/Chat/Channels/Channel.h
+++ b/src/server/game/Chat/Channels/Channel.h
@@ -150,8 +150,8 @@ class Channel
             if (state) flags |= MEMBER_FLAG_OWNER;
             else flags &= ~MEMBER_FLAG_OWNER;
         }
-        bool IsInvisible() const { return _invisible; }
-        void SetInvisible(bool on) { _invisible = on; }
+        bool IsOwnerGM() const { return _gmStatus; }
+        void SetOwnerGM(bool on) { _gmStatus = on; }
         bool IsModerator() const { return flags & MEMBER_FLAG_MODERATOR; }
         void SetModerator(bool state)
         {
@@ -175,7 +175,7 @@ class Channel
                 return false;
         }
     private:
-        bool _invisible = false;
+        bool _gmStatus = false;
     };
 
     public:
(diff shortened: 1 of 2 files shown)


pklloveyou:
Tested and working perfectly

wizzymore:
Yes @Stoabrogga you are right, we need to check based on that config.

I will make the change tomorrow or on monday as i am out of town.

Stoabrogga:
Great, thanks. No need to hurry, take your time.

wizzymore:
I think this should do it

wizzymore:
I will try your method and follow up with feedback if any

wizzymore:
Your method is just announcing when the owner is changes, we don't want that, we wan't to change the owner when the owner is a gm and the config GM Silent Join is enabled to the player that entered the channel, and we want to do this silently as we don't want to announce the player that he is the owner, that meaning either someone just left the channel or someone else is on the channel

Stoabrogga:
Ok, thanks for the explanation, I'll look into this again later.

src/server/game/Chat/Channels/Channel.h
```diff
@@ -197,6 +201,7 @@ class Channel
         void UnBan(uint64 guid);
         void Password(Player const* player, std::string const& pass);
         void SetMode(Player const* player, std::string const& p2n, bool mod, bool set);
+        void SetOwnerGM(Player const* player, bool on);
```
Stoabrogga: This function is not used, please remove.

wizzymore: Why not leaving there as an extension of the function for future edits?
I don't think is does anything wrong

If you really want to delete it i will

wizzymore:
Now this is working as inteded, thank your for all the help @Stoabrogga !

src/server/game/Chat/Channels/Channel.h
FrancescoBorzi: I would avoid adding unused code

Stoabrogga: @wizzymore 
Please remove the function. It is generally bad practice to leave dead code in software projects. It just increases the complexity of the code and makes it harder to maintain the software.

wizzymore:
Please label this as Needs to be tested so Testers can look into the PR

Stoabrogga:
Done. I'll also test this later once more.

wizzymore:
Ty for testing @Stoabrogga 

Idk why i get so many white spaces problems in Sublime Text.. I think i really need to change to VS Code

FrancescoBorzi:
@wizzymore just use CLion since you have the license already ;)

Issue #2316, opened 2019-10-01
Item random enchantments (stats) not generated on player buy/.send items


Shard-MW:
##### SMALL DESCRIPTION:
Item random enchantments (stats) not generated when player buys the item, but they're when we use .add/drop the item
Same for .send items 

##### EXPECTED BLIZZLIKE BEHAVIOUR:
When you buy an item with random enchantments (stats) from a vendor, stats should be generated
Items sent by mail (that were not looted) are supposed to get their random properties generated

##### CURRENT BEHAVIOUR:
Currently, the stats generation works only when a player loot the item from a mob or from .additem
When we try to buy an item with random stats from a vendor or using .send items, players get the item without stats.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Add an item with random stats to a vendor (.npc add it 727)
2. Buy it
3. Observe that the item has no random stats.

1. Use .send items
2. Look at your mailbox and observe that sent items don't have random properties

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
commit : 72faaf643d461ca5cf2798e019c62682b57f155d

##### OPERATING SYSTEM:
Windows 10

##### MODULES:
None

Stoabrogga:
Is there a "blizzlike" vendor who sells items with random stats? I thought those items only drop from mobs.

Shard-MW:
I don't know if there was some on retail, but currently there is no vendor that sells those items on AC.
(According to :
`SELECT DISTINCT entry FROM npc_vendor WHERE item IN (SELECT entry FROM item_template WHERE (RandomProperty != 0 OR RandomSuffix != 0));`)

Shard-MW:
Just noticed that sending items by using .send items don't generate Random Properties too.

PR #2317, opened 2019-10-02
Fix RandomProperties on player item buy (WIP)

Shard-MW:
**I don't think it's the best way to fix it. I think that RandomProperties should be generated at the item creation, but I need feedbacks before working on it because it may imply a "lot" of core changes**

Item's Random Properties are generated when the player buys an item

##### ISSUES ADDRESSED:
https://github.com/azerothcore/azerothcore-wotlk/issues/2316

#### TO DO LIST :
- [x] Fix random stats generation on player buy
- [ ] Fix rando mstats generation on .send items

##### TESTS PERFORMED:
- Bought item with Random properties (itemID : 727 for example)
- Stats were successfully generated

Built without errors on Windows.

##### Target branch(es):
- Master

#### EXTRA INFO : 
Found the fix there : 
http://www.ac-web.org/forums/showthread.php?165672-Items-with-random-enchants

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -22037,7 +22037,7 @@ inline bool Player::_StoreOrEquipNewItem(uint32 vendorslot, uint32 item, uint8 c
     }
 
     Item* it = bStore ?
-        StoreNewItem(vDest, item, true) :
+        StoreNewItem(vDest, item, true, Item::GenerateItemRandomPropertyId(item)) :
         EquipNewItem(uiDest, item, true);
     if (it)
     {


Stoabrogga:
Don't know if the effort is worth it just to fix a minor issue that's not even an issue concerning "blizzlike" behaviour (I don't think there was a vendor selling items with random stats on retail at that time). But as long as it is thoroughly tested, does not break anything and has no other side effects, ok.

BarbzYHOOL:
And what do you think of the actual change he made then?

(note: could also be made into a module but that seems overkill to me)

Shard-MW:
I don't think that item creation was handled this way on retail
Even if there is no vendor selling items with random properties, there was a "security" system that send you ur items by mail if something wrong happen.

With the current issue, items (with random properties) that are sent by mail don't get their random stats generated

Stoabrogga:
@BarbzYHOOL 

> 
> And what do you think of the actual change he made then?

Sorry, I have no idea what this could cause because I know not enough about the item system. This has to be checked by someone who has better knowledge on this.

PR #2320, opened 2019-10-03
feat(Core/SpellMgr): Worldserver option for ICC buff

Stoabrogga:
##### CHANGES PROPOSED:
This was an idea @Knindzagxg had, and I think it is quite neat. This PR adds 2 new parameters to the worldserver configuration:
- `ICC.Buff.Horde`: Buff spell for the Horde (0 for disable, default: 73822 (30% buff))
- `ICC.Buff.Alliance`: Buff spell for the Alliance (0 for disable, default: 73828 (30% buff))

After changing the values the server has to be restarted as the auras are initialized during the loading of spell_area.

I think this functionality is a good addition for admins who want to create a kind of progress concerning ICC.

The ICC buff entries in spell_area have been removed and this is now handled solely via the worldserver configuration.

Not related to this PR: The parameter "CONFIG_PRELOAD_ALL_NON_INSTANCED_MAP_GRIDS" had been treated as integer although it should be a bool. This is also corrected in this PR.

##### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested build successfully on Ubuntu 16.04 / clang 7
- tested successfully in-game

Just use different values for the parameters mentioned above and enter ICC.

none

##### Target branch(es):
- [x] Master
 

pklloveyou:
No effect.。

pklloveyou:
image
image

Stoabrogga:
@pklloveyou 
Did you restart the server after changing the configuration?

pklloveyou:
After modifying the configuration, you must restart the server to take effect

pklloveyou:
Tested and working perfectly

Yehonal:
For this kind of works it's better to use a module strategy. If a module is particulary useful we could also integrate it in azerothcore directly and develop it internally.
But we should avoid to customize the core itself

Stoabrogga:
Ok, sorry, but this is too much work and I even don't need this, just wanted to help @Knindzagxg. I'll close this PR.

FrancescoBorzi:
As I said in discord: If it takes too long to put it as a module, just merge it into the core.

The only suggestion I have it's to move that logic to another file because that one is already too big.

Stoabrogga:
Ok, let's give this another try...

Stoabrogga:
Pushed a new commit. The buffs are now solely configured in the worldserver config, new default is now no buff at all.

src/server/game/World/World.cpp
```diff
@@ -1328,7 +1328,11 @@ void World::LoadConfigSettings(bool reload)
     m_int_configs[CONFIG_AFK_PREVENT_LOGOUT] = sConfigMgr->GetIntDefault("PreventAFKLogout", 0);
 
     // Preload all grids of all non-instanced maps
-    m_bool_configs[CONFIG_PRELOAD_ALL_NON_INSTANCED_MAP_GRIDS] = sConfigMgr->GetIntDefault("PreloadAllNonInstancedMapGrids", false);
+    m_bool_configs[CONFIG_PRELOAD_ALL_NON_INSTANCED_MAP_GRIDS] = sConfigMgr->GetBoolDefault("PreloadAllNonInstancedMapGrids", false);
```
Winfidonarleyan: fix for https://github.com/azerothcore/azerothcore-wotlk/commit/ab55aa4e5a3223ada9a0593dd4cbcb7a37fc5bc4?

Stoabrogga: Not exactly a fix, as the parameter works, but it should be handled as boolean. Just wanted to include this here, I can also create a separate PR to fix this, but this is unnecessary in my opinion.

Winfidonarleyan: I don't think we need a new one either.

Stoabrogga: But I forgot to mention it in the PR description, will add this now.

Yehonal:
i have 2 more questions about it:

1) it should be the same of UPDATING spell_aura record with different spellId, so...since it's an edge case (i mean i never heard before about someone that would mimic the real buff progression)...however who was interested in this i think already changed the value in spell_aura without the need of a configuration

2) changing this value in conf could cause a bug with players that previously had that aura (?) it should be tested: enter ICC with 30% -> logout -> change the config to 20% -> login with the same character-> what happens? does the aura change?

However, about the blizzlike progression, we could think about some "extended" conditions for each kind of data (even creatures from one patch/expansion to another) but it's another topic and i'll talk elsewhere about it.

Stoabrogga:
@Yehonal 
As to answer your questions:
1) Yes, you can also accomplish this via SQL statements, but doing this via config is much easier for most admins (perhaps @Knindzagxg can tell more about the motivation of a config option).

2) The config change needs a server restart. It's basically the same like changing the data in spell_area and then restarting the server. So no bugs will occur.

Knindzagxg:
It's mostly to give admins more options within config so players can have true exp.

Knindzagxg:
All good, works like a charm.
Nice one mate.

Stoabrogga:
Thank you all for the feedback and for testing!

solidmaxtor:
Nice work @Stoabrogga, I have personally used it by modifying by SQL ever. Add this option will be easier to use. <3

BarbzYHOOL:
> For this kind of works it's better to use a module strategy. If a module is particulary useful we could also integrate it in azerothcore directly and develop it internally.
> But we should avoid to customize the core itself

that's more like a blizzlike option, else we could as well remove EVERY custom option from the confs

a module just for that would be overkill, no?

Yehonal:
> > For this kind of works it's better to use a module strategy. If a module is particulary useful we could also integrate it in azerothcore directly and develop it internally.
> > But we should avoid to customize the core itself
> 
> that's more like a blizzlike option, else we could as well remove EVERY custom option from the confs
> 
> a module just for that would be overkill, no?

We already discussed it before :) this conf just allow easier configuration of it for "users", and it's ok, but from a developer perspective a bunch of UPDATE queries were enough to do the same.

About custom conf to move on modules....something like rates they should not be moved under modules, but auction house for example (from trinitycore), it should. But it's another topic.

Stoabrogga:
I pushed another commit to update the comments in the config file. It should be made clear that changing the values needs a server restart.

PR #1811, opened 2019-05-07
feat(MMAPS/mapbuilder):  mmaps_generator now working with MapID >= 1000

Shard-MW:
Get the mmaps_generator working with MapID >= 1000.

mmaps_generator doesn't work with maps that have MapID >= 1000, theses changes will fix it.

##### CHANGES PROPOSED:
-  See pull request

###### ISSUES ADDRESSED:
The issue was that the mmaps_generator doesn't generate the mmaps properly when MapID is above 999. (length of MapID > 3)
Noticed that I've 3 custom maps : 996, 1514 and 1600.
Everything is fine with 996 (generating 996.mmap and 996YYXX.mmtile)
For map 1514 and 1600, it generate : 151.mmap, 151YYXX.mmtile and 160.mmap and 160YYXX.mmtile

The "problem" is that the tool is overriding files, so imagine if you've a custom map with ID 5710, it'll create 571.mmap and 571YYXX.mmtile instead of 5710.mmap and 5710YYXX.mmtile. In consequence, it'll override Northrend (MapID 571) mmaps.

##### TESTS PERFORMED:
Build without errors.
Mmaps working fine in-game : creature follows the player when aggroed and doesn't walk though walls/trees...
Theses changes don't affect non-custom maps
Tested on Windows 10.

##### HOW TO TEST THE CHANGES:
- [ ] Make a custom map with ID > 999, like 1246.
Extract maps, vmaps and mmaps for this new map.
See that the name of created files for MapID 1246 are 1246.mmap and 1246YYXX.mmtile
- [ ] Test with hashes of the old and new generated mmap folders to make sure the files did not change in any way. This way we know it did not break or change anything existing.

##### Target branch(es):
Master

PR #2321, opened 2019-10-03
feat(MMAPS/mapbuilder): mmaps_generator now working with MapID >= 1000

Shard-MW:
Get the mmaps_generator working with MapID >= 1000.

mmaps_generator doesn't work with maps that have MapID >= 1000, theses changes will fix it.

##### CHANGES PROPOSED:
-  See pull request

###### ISSUES ADDRESSED:
The issue was that the mmaps_generator doesn't generate the mmaps properly when MapID is above 999. (length of MapID > 3)
Noticed that I've 3 custom maps : 996, 1514 and 1600.
Everything is fine with 996 (generating 996.mmap and 996YYXX.mmtile)
For map 1514 and 1600, it generate : 151.mmap, 151YYXX.mmtile and 160.mmap and 160YYXX.mmtile

The "problem" is that the tool is overriding files, so imagine if you've a custom map with ID 5710, it'll create 571.mmap and 571YYXX.mmtile instead of 5710.mmap and 5710YYXX.mmtile. In consequence, it'll override Northrend (MapID 571) mmaps.

##### TESTS PERFORMED:
Build without errors.
Mmaps working fine in-game : creature follows the player when aggroed and doesn't walk though walls/trees...
Theses changes don't affect non-custom maps
Tested on Windows 10.

- [x] Make a custom map with ID > 999, like 1246.
Extract maps, vmaps and mmaps for this new map.
See that the name of created files for MapID 1246 are 1246.mmap and 1246YYXX.mmtile
- [x] Test with hashes of the old and new generated mmap folders to make sure the files did not change in any way. This way we know it did not break or change anything existing.

##### Target branch(es):
Master

**Reopened PR : https://github.com/azerothcore/azerothcore-wotlk/pull/1811** (because was closed by mistake)

### DIFF
--- a/src/tools/mmaps_generator/MapBuilder.cpp
+++ b/src/tools/mmaps_generator/MapBuilder.cpp
@@ -88,14 +88,14 @@ namespace MMAP
     void MapBuilder::discoverTiles()
     {
         std::vector<std::string> files;
-        uint32 mapID, tileX, tileY, tileID, count = 0;
+        uint32 mapID, tileX, tileY, tileID, count = 0, fsize = 0;
         char filter[12];
 
         printf("Discovering maps... ");
         getDirContents(files, "maps");
         for (uint32 i = 0; i < files.size(); ++i)
         {
-            mapID = uint32(atoi(files[i].substr(0,3).c_str()));
+            mapID = uint32(atoi(files[i].substr(0, files[i].size() - 8).c_str()));
             if (std::find(m_tiles.begin(), m_tiles.end(), mapID) == m_tiles.end())
             {
                 m_tiles.emplace_back(MapTiles(mapID, new std::set<uint32>));
@@ -107,7 +107,7 @@ namespace MMAP
         getDirContents(files, "vmaps", "*.vmtree");
         for (uint32 i = 0; i < files.size(); ++i)
         {
-            mapID = uint32(atoi(files[i].substr(0,3).c_str()));
+            mapID = uint32(atoi(files[i].substr(0, files[i].size() - 7).c_str()));
             if (std::find(m_tiles.begin(), m_tiles.end(), mapID) == m_tiles.end())
             {
                 m_tiles.emplace_back(MapTiles(mapID, new std::set<uint32>));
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
I guess we only need to double check that the standard map extraction/generation works fine.

No need to re-test with the custom map.

qyh214:
and you also should remove MPQ read restriction.
always we custom patch will use ~ patch-9.MPQ and patch-A.MPQ to patch-Z.MPQ.
in tools you should change,like but not this only:
https://github.com/azerothcore/azerothcore-wotlk/blob/b3876461e8e4333c5887d761288da4beea223b06/src/tools/map_extractor/System.cpp#L92

Helias:
Tested with this custom maps: https://github.com/Nadromar/Custom-Maps
the extractor works.

BarbzYHOOL:
@Shard-MW canyou do what @qyh214 suggested?? I think i had opened an issue about it, it's just that the extractors only look for MPQs up to patch-4.MPQ but not further

qyh214:
> 
> 
> > @Shard-MW canyou do what @qyh214 suggested (in another PR maybe?)?? I think i had opened an issue about it, it's just that the extractors only look for MPQs up to patch-4.MPQ but not further
> > Btw nice, thank you guys!!
> 
> Yes ofc I can do it, but idk if it's really useful since everyone can easily edit sources for extraction. I don't really wanted to change the current developper's habits so that's why I haven't inculded it in the PR
> I did it for me, I can release it if needed :)

I think that is part of the PR. Generally speaking custom maps are in custom patch like patch-4 to patch-9 and patch-a to patch-z also locale mpq like patch-enus-z.

so,I think you should add them in this PR.

Because this PR is about the implementation of a custom content, it should also support the custom mpq, which can be added some notes in the source code.

Shard-MW:
@qyh214 PR wasn't about implementation of custom content, but getting code working with every files.
But I understood you, and I'm ok to add this, I just need approval from contributors :)

BarbzYHOOL:
imo it can be in this PR but the best way to go, to always avoid mistakes, is to make a separate PR so we can merge this one meanwhile :)

PR #2347, opened 2019-10-12
feat(CI): add docker to travis build

FrancescoBorzi:


.travis.yml
```diff
@@ -73,3 +75,11 @@ jobs:
...
       script:
         - source ./apps/ci/ci-compile.sh
+
+    - stage: docker
+      env: TRAVIS_BUILD_ID="1"
+      script:
+        - ./apps/ci/docker/ci-docker-config.sh
+        - ./bin/acore-docker-generate-etc
+        - ./bin/acore-docker-build-no-scripts
+        - docker-compose up
```
Winfidonarleyan: miss new line

apps/ci/docker/.env.dist
```diff
@@ -0,0 +1,12 @@
...
+WORLDSERVER_LOGS=./docker/worldserver/logs
+
+AUTHSERVER_ETC=./docker/authserver/etc
+AUTHSERVER_LOGS=./docker/authserver/logs
+
+WORLD_EXTERNAL_PORT=8085
+AUTH_EXTERNAL_PORT=3724
+DB_EXTERNAL_PORT=9000
+DB_ROOT_PASSWORD=password
+
```
Winfidonarleyan: extra line

Stoabrogga:
I tried to somehow run the docker job in parallel to the other jobs including both stages, sadly without success. It was also not possible to take over the ccache from the other jobs, so I just added the docker job to the "run" stage. Normally this won't make any difference, only if the ccache is filled and the other "run" jobs succeed within a few minutes it will take 10 minutes more to finish the docker job, but that's ok in my opinion.

Merging this as it is an important addition to the CI.

PR #2368, opened 2019-10-21
feat(Core/GameEvent): Allow unspecified end_time for game events

Stoabrogga:
##### CHANGES PROPOSED:
As it is now many game events have their "end_time" set to 2020-12-31 which means that they won't start anymore after that date unless the date is changed in the "game_event" table. This PR adjusts the logic while loading the data from the table:
- If "end_time" is NULL it will be implicitly set to 2 years in the future on each server start.

Additionally the events have to be changed in the DB:
- The game events which used NULL before will be set to 1970-01-01 for their "start_time" and "end_time", which actually is the same as it was before (NULL initializes the uint64 variable with 0 and "time_t(0)" results in 1970-01-01).
- The game events which had an "end_time" in 2020 will be set to NULL, so they will always work in the future.

##### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested successfully in-game

- Test a game event which had NULL as "end_time", e.g. event 55 "Arena Season 3":
  `.event 55`
  It should show date 1970-01-01 as both "start" / "end"
- Test a game event which had an "end_time" in 2020, e.g. event 12 "Hallow's End" or 67 "AT Event Trigger (Tirion Speech)"
  `.event 12`
  `.event 55`
  It should show a date 2 years in the future for "end"

none

##### Target branch(es):
- [x] Master
 

Knindzagxg:
Values are correct now.

Tested, and no issues at all.

LichKing255:
Do we need end time? It seems to be of no use.

Stoabrogga:
@LichKing255 
Yes we need it in order to provide different options for administrators (e.g. they want a specific event to only take place once in a specific time interval). I don't want to remove options if not absolutely necessary.

PR #2377, opened 2019-10-24
Delete azerothcore motd

progressor42:
How to delete azerothcore motd


Stoabrogga:
Perhaps take a look at this:
https://github.com/azerothcore/azerothcore-wotlk/blob/9f593eb64e541eca05a857bec292a53e85cecfa8/src/server/worldserver/worldserver.conf.dist#L1443-L1455

PR #2304, opened 2019-09-23
feat(Core/CreatureAddon): increased visibility for large creatures

Stoabrogga:
##### CHANGES PROPOSED:
- Add a new column "isLarge" to `creature_addon` and `creature_template_addon`, holding a boolean value controlling if the creature is large or not (default: 0).
- Large creatures have their visibility increased to MAX_VISIBILITY_DISTANCE (250.0f).
- Large creatures are updated even if they are out of normal range of a player. I had to increase the cell visits for the players during map updates, but only trigger updates for large creatures, so it should not cost much performance. The large creatures are not active, they will stop getting updates after the player moves too far away.
- I tried to keep the code changes to a minimum in order to not break existing behaviour.
- I ported all large creatures from the TC database to AC. Examples:
  - Doomwalker (also added his missing waypoint movement)
  - Fel Reaver Sentry: Located in Abyssal Shelf, they are only used for the bombing runs
  - Dun Niffelem Spear Chain Bunny (Phase 2): They are used to cast the chains holding Hodir's Spear in Dun Niffelem (they were barely visible before)

Example: Doomwalker before / after:
WoWScrnShot_092319_170850
WoWScrnShot_092319_170958

##### ISSUES ADDRESSED:
none

##### TESTS PERFORMED:
- tested build on Ubuntu 16.04 / clang 7
- tested successfully in-game

##### HOW TO TEST THE CHANGES:
- Doomwalker: `.go -3465.99 312.405 93.4952 530`
- Fel Reaver Sentry: `.go 714.31 1885.59 176.008 530`
- Hodir's Spear: `.go 7235.8 -2844.83 822.885 571`

##### KNOWN ISSUES AND TODO LIST:
none

##### Target branch(es):
- [x] Master
 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #2378, opened 2019-10-25
fix(Core/GridNotifiers): increased visibility for large creatures, part 2

### LINKED ISSUE #2323: Strange visibility problem concerning active creatures
##### SMALL DESCRIPTION:
Active creatures have increased visibility (max visibility: 250 yards), but are still visible after walking out of this range if the player just stands still doing nothing.

The creatures should vanish if they move out of maximum visibility.

##### CURRENT BEHAVIOUR:
@LichKing255 posted an example here concerning a Fel Reaver:
https://github.com/azerothcore/azerothcore-wotlk/pull/2304#issuecomment-537847630

This can be tested easily by teleporting to a Fel Reaver (`.go creature fel_reaver`) and then just wait while it moves away (`.dist` command can be used to check the distance). If the Fel Reaver is more than 250 yards away it is still visible if the player does not move. After the player moves the Fel Reaver vanishes.

This affects also PR #2304 as it increases the visibility of large creatures and provides updates for them, similar to creatures which are set active.

see above

##### EXTRA NOTES:
none

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
b3876461e8e4333c5887d761288da4beea223b06

##### OPERATING SYSTEM:
Ubuntu 16.04

##### MODULES:
mod-autobalance
mod-npc-beastmaster
mod-transmog
mod-weapon-visual

##### OTHER CUSTOMIZATIONS:
none

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/81371623-strange-visibility-problem-concerning-active-creatures?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Stoabrogga:
##### CHANGES PROPOSED:
This is a follow-up to PR #2304. Currently there exist two problems with large objects:
- They don't initially appear at their full range.
- As for creatures, they don't disappear if the player stands still and just watches them moving (see #2323)

I tried to fix this by adjusting the VisibleNotifier to either handle only non-large or only large objects. The visibility check is then split into two parts: Default distance for non-large, maximum distance for large objects. So far my tests where successful and it looks much better now as large creatures and game objects are now appearing at full distance.

##### ISSUES ADDRESSED:
Closes #2323

##### TESTS PERFORMED:
- tested successfully in-game

- Check one of the Fel Reavers (`.go creature fel_reaver`) and then just wait doing nothing while it moves away. It should disappear after reaching maximum visibility distance.
- Teleport to Icecrown (`.go 6904.96 2323.82 474.47 571`) and fly around a bit, there are many large creatures there.

none

##### Target branch(es):
- [x] Master
 
### DIFF
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -974,7 +974,7 @@ class WorldObject : public Object, public WorldLocation
 
         bool isActiveObject() const { return m_isActive; }
         void setActive(bool isActiveObject);
-        bool IsVisibilityOverridden() const { return m_isVisibilityDistanceOverride; }
+        bool IsVisibilityOverridden() const { return m_isVisibilityDistanceOverride || m_isActive; }
         void SetVisibilityDistanceOverride(bool isVisibilityDistanceOverride);
         void SetWorldObject(bool apply);
         bool IsPermanentWorldObject() const { return m_isWorldObject; }
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -23042,9 +23042,12 @@ void Player::UpdateObjectVisibility(bool forced, bool fromUpdate)
 
 void Player::UpdateVisibilityForPlayer(bool mapChange)
 { 
-    Trinity::VisibleNotifier notifier(*this, mapChange);
-    m_seer->VisitNearbyObject(GetSightRange()+VISIBILITY_INC_FOR_GOBJECTS, notifier);
-    notifier.SendToSelf();
+    Trinity::VisibleNotifier notifierNoLarge(*this, mapChange, false); // visit only objects which are not large; default distance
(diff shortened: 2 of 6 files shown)


Knindzagxg:
Had no serious issues while testing.

Only small one -> Since we have them in max visibility, sometimes you will be able to see them going kinda trough ground, which is cause maps i think.

Then kinda issue everywhere, even sometimes on Blizz, but maybe in near future we will get new maps as well.

Stoabrogga:
Great, thanks for testing @Knindzagxg.

LichKing255:
if PreloadAllNonInstancedMapGrids enable , don't have problem ? Because some npcs are active

Stoabrogga:
@LichKing255 
No, because this PR only affects visibility of the creatures to the players. It doesn't matter if the active creatures start their movement when you preload all grids.

Issue #2383, opened 2019-10-25
Core/Worldstates: daily calendar deletion old events time


Helias:
We could add the worldstate WS_DAILY_CALENDAR_DELETION_OLD_EVENTS_TIME and a related config to set the deletion of old events.

I get this missing worldstate comparing our worldstates to TC.

Related code: https://github.com/TrinityCore/TrinityCore/commit/fb059722fdb18a94d47b3f44a5b2d7f183bf5a29#diff-69f0ea4bf362615d6993f2a38027d31a

Kitzunu:
Fixed by https://github.com/azerothcore/azerothcore-wotlk/pull/2799 , right?

Helias:
Yes, thanks @Kitzunu

PR #2387, opened 2019-10-26
feat(Core/AuthServer) Added check for AuthFlooder, now the server stop it!

Helias:
There is a public tool, AuthFlooder that DDOS the authserver with multiple requests, this little fix can easily stop it.

##### CHANGES PROPOSED:
-  Added a MAX request REALMLIST quantity in a row to 10, this will STOP the AuthFlooder

##### ISSUES ADDRESSED:
- This is a critical issue, a member can lock the client log in if open multiple AuthFlooder and target a victim server

##### TESTS PERFORMED:
- I used AuthFlooder with and without the fix, the fix stop it

Before the PR, I opened 3 AuthFlooder and in any I got this output in console:
immagine

The client wait in the "Authenticating" phase:
immagine

After the PR, here the AuthFlooder output text:
immagine

In the AuthServer console you will get `Got 10 REALM_LIST in a row from '127.0.0.1', possible ongoing DoS`.

I tried also to log in with my client 10 times, it works perfectly, the problem is make it "in a row", so multiple times in less than 1-5 seconds, hence, the normal using of the client is not affected from this check.

- [Download AuthFlooder](https://mega.nz/#!ExpSwSJJ!fc8wauzmiI_RB-UlZ7nYWzMuYT0uq_gbCx2XijSWxX8)
- Open AuthFlooder.exe, put the realmlist, example `localhost` and then username and password (in my example above they where username: gm, password: gm).

On **Linux** use `mono` instead of Wine.

##### Target branch(es):
- [x] Master

### DIFF
--- a/src/server/authserver/Server/AuthSocket.cpp
+++ b/src/server/authserver/Server/AuthSocket.cpp
@@ -208,6 +208,10 @@ void AuthSocket::OnRead()
 {
     #define MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW 3
     uint32 challengesInARow = 0;
+
+    #define MAX_AUTH_GET_REALM_LIST 10
+    uint32 challengesInARowRealmList = 0;
+
     uint8 _cmd;
     while (1)
     {
@@ -223,6 +227,15 @@ void AuthSocket::OnRead()
                 socket().shutdown();
                 return;
             }
+        }
+        else if (_cmd == REALM_LIST) {
+          challengesInARowRealmList++;
+          if (challengesInARowRealmList == MAX_AUTH_GET_REALM_LIST)
+          {
+              sLog->outString("Got %u REALM_LIST in a row from '%s', possible ongoing DoS", challengesInARowRealmList, socket().getRemoteAddress().c_str());
+              socket().shutdown();
+              return;
+          }
         }
 
         size_t i;


pklloveyou:
image
```c++
		//if (_cmd == AUTH_LOGON_CHALLENGE)
		if (_cmd == AUTH_LOGON_CHALLENGE || _cmd == REALM_LIST) // realmlist packet
		{
			++challengesInARow;
			if (challengesInARow == MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW)
			{
				sLog->outString("Got %u AUTH_LOGON_CHALLENGE in a row from '%s', possible ongoing DoS", challengesInARow, socket().getRemoteAddress().c_str());
				socket().shutdown();
				return;
			}
		}
```

src/server/authserver/Server/AuthSocket.cpp
```diff
@@ -208,6 +208,10 @@ void AuthSocket::OnRead()
 {
     #define MAX_AUTH_LOGON_CHALLENGES_IN_A_ROW 3
     uint32 challengesInARow = 0;
+
+    #define MAX_AUTH_GET_REALM_LIST 10
```
FrancescoBorzi: we could make it configurable via `authserver.conf`

Winfidonarleyan: Good idea, i think

Helias: This is a critical bug, I think that the priority is test and merge this PR asap.

I will add the config when meanwhile I'll make the PR for this issue https://github.com/azerothcore/azerothcore-wotlk/issues/2383

solidmaxtor:
Test and work perfect, its a very great work @Helias  <3

Winfidonarleyan:
Perfect work

BarbzYHOOL:
great jon

PR #2407, opened 2019-11-05
fix(Core/Crash): Improved Utf8toWStr() function to prevent crashes

Helias:
Me and @Stoabrogga figure out that `Utf8toWStr()` is responsible for server crashes when a malformed opcode and special characters are sent to the server.

##### CHANGES PROPOSED:
-  improved Utf8toWStr() like TC-335a do

##### ISSUES ADDRESSED:
- #2077
- #2043
- #2170
- #2198
- #2345
- other issues about exploiting crashes which functions turn on `Utf8toWStr()`

related to this PR https://github.com/azerothcore/azerothcore-wotlk/pull/2374

##### TESTS PERFORMED:
- I used AccLeito addon to send malformed CSMG_ADD_FRIEND opcode and special character to the server, with this PR the server does not crash.

##### HOW TO TEST THE CHANGES:
- use AccLeito Addon and try to get down the server with the Crash button

##### Target branch(es):
- [x] Master

PR #2411, opened 2019-11-07
Core/Misc: fix utf8 buffer overflow when characters are 4 octets wide

Viste:
##### CHANGES PROPOSED:
- prevent utf8 buffer overflow when characters are 4 octets wide
##### ISSUES ADDRESSED:
- Closes must slove all utf8 crashes
be a bad guy
##### Target branch(es):
- [x] Master


Viste:
codacy wrong here :C need ignore

Viste:
I'll reopen it when I have a full picture of what's going on in my head.

BarbzYHOOL:
it was for a similar result as https://github.com/azerothcore/azerothcore-wotlk/pull/2407#issuecomment-550080729 ??

Viste:
@BarbzYHOOL yes

BarbzYHOOL:
ok, it's complementary or a different method?

Viste:
trying to find a balance between TC and mangos variants

Helias:
I agree with this improving.
TC is "modern" for different code than AC, AC has old code like `utf8var = ""` instead of `utfvar.clear()` that are more professional and better.

PR #2412, opened 2019-11-07
Core/PacketIO: Restrict CMSG_EMOTE/CMSG_STANDSTATECHANGE to only allow emotes/stand states that client can send by itself

Viste:
##### CHANGES PROPOSED:
-  hardcode CMSG_EMOTE/CMSG_STANDSTATECHANGE to only allow emotes/stand states that client can send by itself
try to use emotes/stand states what you don't have in client

##### Target branch(es):
- [x] Master


src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -543,13 +543,20 @@ void WorldSession::HandleSetSelectionOpcode(WorldPacket & recv_data)
...
+        case UNIT_STAND_STATE_STAND:
+        case UNIT_STAND_STATE_SIT:
+        case UNIT_STAND_STATE_SLEEP:
+        case UNIT_STAND_STATE_KNEEL:
+            break;
+        default:
+            return;
+    }
+
     _player->SetStandState(animstate);
```
FrancescoBorzi: it would look more logical to move `_player->SetStandState(animstate);` inside the `switch`

Viste: No its must be after switch

FrancescoBorzi: ok for me but, just for my understanding, is it only about codestyle or I'm missing something?

Viste: set the state must handle after we get list of allowed stand states

Viste:
and?

Issue #2280, opened 2019-09-10
Memory Leak


LichKing255:
##### SMALL DESCRIPTION:
i find problem why vmaps not load #2215 .
worldserver start with 600 mb of memory, and after 2 hour when have 10x players in server , this used 90% of memory . when use 90% , vmaps not load.
In fact, the main problem is overuse of memory.

10 min uptime with 3x player :
https://mega.nz/#!W9UTEAZI!EWFDAaeK4qWOMPLHxTy8a9M7Yrmt5sS6yJBHKgvMMPg

15 min uptime with 5x player : 
https://mega.nz/#!qsNVhCwY!OnZ2iBjEikihbgB59oTXlOki6pfqIh16Dh7AiUo7O6k

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/commit/67d180ea776a171c6d660d964d6454ea504ce55b

##### OPERATING SYSTEM:

Windows Server 2016 Ram 4

##### MODULES:

nothing

##### OTHER CUSTOMIZATIONS:

nothing

Stoabrogga:
This is normal behaviour. 4 GB RAM is not enough, in my opinion you need at least 6 GB for your case in order to have enough resources left for the OS. My own server installation uses ca. 1.8-2 GB with only 1 player (myself) on Ubuntu 16.04.

FrancescoBorzi:
And why exactly you think it's a memory leak?

zabustak:
Operating system | Debian Linux 9 @ 4.20GHz, 8 cores ~60GB ram
4 days uptime server pop low ~6 gb ram 
14 days uptime same pop ~10 gb ram
https://ibb.co/5TZS1M9

I recommend you switch to Linux with larger resources, the server uses about 4GB RAM even if no one is playing

FrancescoBorzi:
closing as there is no evidence of a real memory leak 

using much RAM != memory leak
improperly using RAM == memory leak

LichKing255:
As you say, after a few days the memory must to be filled.
So every time the memory is full Vmaps will not load.

wowmane:
Ican, Confrim.

For test

1. set to GridUnload = 1 or 0
2. gm, mod speed + fly on and move to maps
3. memory will to leak for enter character to the maps, "Grids"

I tested:
If Memory is Usage 20%
I will to move and join maps, server/core unload grids "Cache"
[...]

PR #2413, opened 2019-11-07
Dep/Jemalloc: update to Jemalloc 5.2.1

Viste:
##### CHANGES PROPOSED:
-  update to Jemalloc 5.2.1 we have to old >_< (maybe partially reduce memory consumption)
build 
run
watch
write
##### Target branch(es):
- [x] Master


Viste:
@LichKing255 thanks

FrancescoBorzi:
@Viste is there anything specific that should be checked or we can just merge it?

Viste:
@FrancescoBorzi we can merge i think.
But It would be cool to see how to use memory with and without PR. Especially on windows.
As there was a lot of things fixed concerning leaks in windows

LichKing255:
> @FrancescoBorzi we can merge i think.
> But It would be cool to see how to use memory with and without PR. Especially on windows.
> As there was a lot of things fixed concerning leaks in windows

I didn't see much of an impact on the Memory Leak I reported earlier.
And my problem continues.
https://github.com/azerothcore/azerothcore-wotlk/issues/2280
issue closed but there is a problem with some windows.
I don't know this is memory leak or vmaps problem, but i know there is a problem.
There have been many reports but the reason is not clear.
Even after my report someone else reported it and it has not closed yet
https://github.com/azerothcore/azerothcore-wotlk/issues/2233
I've reviewed Trinity Commit and there is no solution to test it. You probably just have to merge it

Viste:
and?

Mitradis:
after this "update" (which gives nothing!) i stop build AC. Ty Viste.

Viste:
it's hard to live with mental illness

PR #2421, opened 2019-11-10
feat(License): delete address to license file

Winfidonarleyan:
##### CHANGES PROPOSED:
- Delete address to the license file (like TC)

I just don't want to get confused, because we have two licenses

### DIFF
--- a/src/cmake/compiler/clang/settings.cmake
+++ b/src/cmake/compiler/clang/settings.cmake
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU AGPL3 v3 license: https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-AGPL3
+# Copyright (C) 2016+     AzerothCore <www.azerothcore.org>
 # Copyright (C) 2008-2019 TrinityCore <https://www.trinitycore.org/>
 #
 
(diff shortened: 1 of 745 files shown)


Helias:
I'm not sure about this PR, I think that we could leave the license mention in the source file.

BarbzYHOOL:
Yehonal and Shin know more about licensing, let's see what they say

But why would it be better to remove them ?

Yehonal:
We use GPL2 only for old TC files, but all our new features are under AGPL2 that "fixes" various GPL2 lacks (such license has been created before the internet era).

PS Modules can use any kind of license they want, even not open-source.

FrancescoBorzi:
that's the point, we release new files under AGPL v3 but we can't change the license to the existing files (GPL v2). That's why we need to specify in each file which of the two licenses is used

BarbzYHOOL:
> that's the point, we release new files under AGPL v3 but we can't change the license to the existing files (GPL v2). That's why we need to specify in each file which of the two licenses is used

Yes, old licenses can never be removed unless we rewrite everything from scratch

Helias:
I think we can update and merge this PR.
Probably we don't need to add the URL of the license in the source code.

BarbzYHOOL:
@Yehonal

FrancescoBorzi:
there is no need for this change, as I said: we have different files having different licenses (old files GPL v2, new files AGPLv3) that's why IMHO we should keep it

BarbzYHOOL:
ok (sorry winfi :D)

PR #2432, opened 2019-11-13
fix(Core/GameObjectScript): Environmental damage of burning game objects

### LINKED ISSUE #2396: The fire will not produce fire damage
SMALL DESCRIPTION:
image
Bonfires should produce damage
Change "gameboject entry 94183" displayID to 197 to generate damage correctly
Bonfire material blocking the damage of the game player
```c++
.go o 23
```

EXPECTED BLIZZLIKE BEHAVIOUR:
Bonfires should produce damage

CURRENT BEHAVIOUR:
The fire did not hurt, 
but moved to the fire internal damage “.go o 23”, 
remove the fire GUID23, GUID24 damage properly，
Object 24 impenetrable object 23 deals damage to the player

```c++
.go o 23
```
As shown in
image

BRANCH(ES): master
AC HASH/COMMIT: d9986c0
OPERATING SYSTEM: Win 10 x64
MODULES: NO

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/82931881-the-fire-will-not-produce-fire-damage?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2397: Braziers fire damage radius too large [$25]
##### SMALL DESCRIPTION:
The fire damage radius of various braziers and other similar game objects is too large.

You should only get fire damage if standing very close to the braziers.

##### CURRENT BEHAVIOUR:
You suffer fire damage even when standing 1-2 yards away from the game object.

- near Thrall in Orgrimmar (`.go 1923.89 -4198.55 42.1597 1`)
- in the cave in Valley of Trials (`.go -605.034 -4172.43 41.0001 1`)

##### EXTRA NOTES:
none

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
2cb28e5f5342e92934b5d7d3acd771a7785be82d

##### OPERATING SYSTEM:
Ubuntu 16.04

##### MODULES:
mod-autobalance
mod-npc-beastmaster
mod-transmog
mod-weapon-visual

##### OTHER CUSTOMIZATIONS:
none

<bountysource-plugin>

---
There is a **[$25 open bounty](https://www.bountysource.com/issues/83028540-braziers-fire-damage-radius-too-large?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Stoabrogga:
##### CHANGES PROPOSED:
- Add 2 new game object scripts which handle the environmental damage spells only when the player is very near the boundary of the game object. This fixes small game objects, like braziers, burning the player from far away and big game objects, like bonfires, now finally can cause damage if standing near them. If more than one player is affected by the game object the environmental damage is placed on a random player to prevent multiple fire damage (this could easily kill low level players otherwise).
- Remove all the trap game objects which were used to deal fire damage (template IDs 2061 and 2066) as they are not needed anymore (this frees up 2858 entries which can be used in the future for other game objects).
- Change all game object templates of type "GAMEOBJECT_TYPE_SPELL_FOCUS" which used either ID 2061 or 2066 as traps to use the new game object scripts.

##### ISSUES ADDRESSED:
- Closes #2396
- Closes #2397

##### TESTS PERFORMED:
- tested successfully in-game

Just run near burning objects, like bonfires, braziers, campfires, fireplaces etc. Examples:
- `.tele ValleyOfWisdom`: On your way to Thrall the Mighty Blazes should not burn the player from far away.
- `.go o 23`: The bonfire should now cause fire damage

none

##### Target branch(es):
- [x] Master
 

sluggor:
Are there other traps dealing environmental damage that would need to be handled as well, or is it only fire?

It looks like the root of the problem is here: https://github.com/azerothcore/azerothcore-wotlk/blob/23b77c1303244ccc40e7998c63eab1f50be3a776/src/server/game/Entities/GameObject/GameObject.cpp#L616

The searcher() routine is converting the map coordinates to grid cell coordinates to perform the search, and finds players in the same grid cell. Unfortunately, the grid cell is larger than the small diameter of the intended fire damage, so the player triggers the trap from a larger radius than is specified.

If we let the searcher() find players in the grid cell, but then added a check to make sure that the map coordinates were within the intended radius, it might avoid a lot of work and changes.

For example, between lines 619 and 621, add

` 
if(target) 
{
	float trapX = GetPositionX();
	float trapY = GetPositionY();

	float playerX = target->GetPositionX();
	float playerY = target->GetPositionY();

	if((playerX < (trapX - radius) || playerX > (trapX + radius)) ||
	   (playerY < (trapY - radius) || playerY > (trapY + radius))) {
		target = NULL;
	}
}
`

That way, any players found within the grid cell but outside of the trap's radius would be nulled out and avoid taking damage.

This is a two dimensional check specifically for the fire damage problem, but could be enhanced if there are 3D traps out there with tiny radii.

Stoabrogga:
This won't fix the issues with those game objects. You have to trigger the environmental damage effect on the boundaries of the visible game objects, which vary heavily in size (braziers, bonfires etc.). The trap objects themselves are not needed at all. As far as I know it's only a problem concerning these specific objects so I kept the changes solely to them in order to not risk changing other existing behaviour.

sluggor:
That code does fix the fire damage radius problem - I'm using it in my code now, and I have to stand on the braziers in IF to take damage - but however you want to handle it is fine, it just seemed like a lot to change, so it may be difficult to test thoroughly.

Stoabrogga:
No, it's not. It's the least intrusive solution that I could think of. It only touches the game objects of type "GAMEOBJECT_TYPE_SPELL_FOCUS" which use either game object template ID 2061 "Campfire" or ID 2066 "Bonfire Damage" as traps (see "Data2" column). Those two templates are used throughout the world for this purpose. I did also check each of the deleted trap objects, and 2850 of them had the exact same position as their associated "fire" game object (8 of them were just misplaced). So in order to test this you just have to test a very small "fire" game object and a very large one, that's it (they are all the same, just different size).

pklloveyou:
I tested it and it worked great

pklloveyou:
image
image
image
image
image
image

data/sql/updates/pending_db_world/rev_1573682333945401103.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1573682333945401103');
+
+DELETE FROM `gameobject` WHERE `id` IN (2061,2066);
```
LichKing255: why delete all game objects flame ?

Stoabrogga: Please read the explanations in my comments above.

Mitradis:
Trinity => acore

PR #2410, opened 2019-11-06
fix(Core): prevent movement exploit

Helias:
I added some checks in `WorldSession::HandleMovementOpcodes()` function to prevent a possible exploit.

##### CHANGES PROPOSED:
-  added some checks for movement opcode inspired by TrinityCore and https://github.com/azerothcore/azerothcore-wotlk/issues/2327
- removed useless Opcodes attribute **isGrouppedMovementOpcode**, related condition which cause the server crash for malformed opcode and fixed indentation in Opcodes.cpp
- added some immutability features about bytebuffer exceptions

##### ISSUES ADDRESSED:
- Close #2409 
- Close #2422 
- Close #2327 

##### TESTS PERFORMED:
I tested the following things:
- MSG_MOVE_HEARTBEAT opcode malformed using WPS
- MSG_MOVE_SET_FACING opcode malformed using WPS
The server skip these malformed packet and prints some logs.
skip byte log

I tested also the normal playing that using (probably) "groupped" movements:
- gryphon mount
- horse mount
- boat movements in SOTA BG (2 players)
- tank movements in SOTA BG (2 players)
- horde movements and attack into SOTA tank
- blink spell

##### HOW TO TEST THE CHANGES:
- @lineagedr [comments](https://github.com/azerothcore/azerothcore-wotlk/pull/2410#issuecomment-552111752) explain which lua scripts use with WPS/Accleito tool to make crash the server.
- playing use all kind of movements that turn on client standard movement opcode

##### Target branch(es):
- [x] Master

PR #2434, opened 2019-11-14
Core/Movement: Fix warrior charge animation

LichKing255:
##### Current behaviour:

Warrior appears like teleporting, the charge animation starts, then warrior moves about 2-3 yards forward, then gets back, and immediately teleported to the target after that.

##### Expected behaviour:

Charge animation should be smooth.

1. Make Warrior
2. Make another character as target
3. Jump backwards and immediately press charge

Charge animation always looks smooth on the warrior's client, but very weird and broken on the target's client (this is not always happening, this macro to reproduce it easily:

> /cast charge
> /target player
> .cool
> /targetlasttarget

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -322,7 +322,7 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
     // pussywizard: typical check for incomming movement packets
     if (!mover || !mover->IsInWorld() || mover->IsDuringRemoveFromWorld() || guid != mover->GetGUID())
     {
-        recvData.rfinish();
+        recvData.rfinish();                     // prevent warnings spam
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/server/game/Handlers/MovementHandler.cpp
```diff
@@ -322,21 +322,27 @@ void WorldSession::HandleMovementOpcodes(WorldPacket & recvData)
...
         recvData.rfinish();                     // prevent warnings spam
         return;
     }
 
-    recvData.rfinish();                         // prevent warnings spam
+    if (!mover->movespline->Finalized())
+    {
+        recvData.rfinish();                     // prevent warnings spam
+        return;
+    }
```
Helias: This check is already contained in this PR https://github.com/azerothcore/azerothcore-wotlk/pull/2410

Winfidonarleyan:
https://github.com/azerothcore/azerothcore-wotlk/pull/2410/files#diff-649b0ea06820eb06283d1ccc8c83efdbR329

pklloveyou:
PR #2410　It should have fixed this bug　。I had another client that did not appear to intercept the Caton error after merging#2410

pklloveyou:
Is there a way to replicate this error after today's latest commit?I'll test it.

PR #2254, opened 2019-09-02
 Core/Spells: auras improvements 

Viste:
##### CHANGES PROPOSED:

-  Continue updating creatures that own auras applied to players. 
-  check for owner of aura in world
-  Increase number of auras sent to client except for group updates
-  Fix SPELL_AURA_MOD_TOTAL_STAT_PERCENTAGE unapply case for dead players

PR #2436, opened 2019-11-14
Core/Misc: utf8 handle and utilities improvments and other misc stuff

Viste:
##### CHANGES PROPOSED:
-  replace AccountMgr::normalizeString with Utf8ToUpperOnlyLatin to prevent use non latin where must be only latin
-  add some helper functions for future and etc
-  use vector instead of list to contain Spell and Aura scripts and hooks ( need for #2254 )
-  added additional argument to Tokenizer class to make it behave like strtok - not returning empty tokens in case of multiple consecutive separators in input string
-  some renames from trinity to acore
-  Moved rng functions to separate header and added utility functions to select a random element from a container where each element can have different chance of being selected
- Update SFMT to latest
build run be sure all is ok
##### Target branch(es):
- [x] Master


Viste:
WIP still

src/common/Define.h
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license: https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-GPL2
+ * Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU AGPL3 v3 license: https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-AGPL3
```
Winfidonarleyan: old files at `GPL v2`

Viste:
i think ready

Viste:
Codacy is stupid asshole and lie about  \0-terminated look at code all is handled>_<

Viste:
to many things what can fuckup many things open several small PRs

FrancescoBorzi:
@Viste yes, smaller PRs are always more manageable :P

PR #2465, opened 2019-11-24
fix(Core/Mailhandler):  fix in premium module

pangolp:
##### CHANGES PROPOSED:
-  It does not work the sending of mails by means of the module.

##### ISSUES ADDRESSED:
- Closes

##### TESTS PERFORMED:
- Tests carried out in Windows 10. Using the premium azeroth module.

##### Target branch(es):
- [x] Master

 

 

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -22,8 +22,11 @@ bool WorldSession::CanOpenMailBox(uint64 guid)
 {
     if (guid == _player->GetGUID())
     {
-        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
-        return false;
+        return true;
```
FrancescoBorzi: why changing this return?

Viste: to allow open mailbox in cheating way i guess

Bogir: it must be something like:

```suggestion
if (guid == _player->GetGUID())
{
  QueryResult result_prem = CharacterDatabase.PQuery("SELECT * FROM premium WHERE active = 1 AND AccountId = %u", _player->GetSession()->GetAccountId());
  if (!result_prem) {
    sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
    return false;
  } else { 
    return true;
}
```

Bogir: What is the premium-module you are talking about?

pangolp:
Yeah, it's the premium module. After I make the pull request, I realize that maybe it wasn't a theme of the core, nose of the module as you mention. It was a topic of conversation that you had inside the module, where you proposed that change. 
**https://github.com/azerothcore/mod-premium/issues/8**

PR #2354, opened 2019-10-14
fix(DB/SAI): Garm's Bane

Stoabrogga:
##### CHANGES PROPOSED:
- The Garm Invaders and Snowblind Followers will now run towards K3.
- The Garm Invaders are set active on respawn so that the NPCs also run when the player is out of reach. I ensured that the event stops if no player is near anymore by timing the kills / respawns appropriately.
- A few improvements concerning the "K3 Perimeter Turret" and the "Improved Land Mine".

##### ISSUES ADDRESSED:
Closes #2237

##### TESTS PERFORMED:
tested successfully in-game

##### HOW TO TEST THE CHANGES:
- preparation:
```
.q rem 12820
.q a 12820
.go 6299.75 -1156.48 424.188 571
```
- proceed with the quest

##### KNOWN ISSUES AND TODO LIST:
none

##### Target branch(es):
- [x] Master
 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #2468, opened 2019-11-26
fix(DB/SAI): Re-added Snowblind Follower interaction for quest 14090

Andrei-UPB:
Adds to : https://github.com/azerothcore/azerothcore-wotlk/pull/2354
References taken from https://wowgaming.altervista.org/aowow/?quest=14090 / https://wowgaming.altervista.org/aowow/?quest=14141

##### CHANGES PROPOSED:
-  Stoabrogga's PR #2354 added waypoints and event interaction to npc's in Garm's Rise but broke the functionality of quests 14090 and 14141
- This commit fixes that by re-adding the missing SAI lines

##### TESTS PERFORMED:
- Tested in-game, using the given quest item on a Snowblind Follower rewards credit to the quests as intended.

1. .go c id 29618
2. .q add 14090
3. and/or .q add 14141
3. Use Weighted Net (item 46885) on a Snowblind Follower
4. Verify if the quest objective updates or not.

##### Target branch(es):
- [x] Master

 

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1574778273250296900.sql
+++ b/data/sql/updates/pending_db_world/rev_1574778273250296900.sql
@@ -1,13 +1,6 @@
-DELETE FROM `smart_scripts` WHERE (`entryorguid`=29618 AND `source_type`=0);
+DELETE FROM `smart_scripts` WHERE `entryorguid` = 29618 AND `source_type` = 0 AND `id` IN (7,8,9);
 INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`)
 VALUES
-(29618, 0, 0, 1, 11, 0, 100, 0, 0, 0, 0, 0, 0, 20, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, "Snowblind Follower - On Respawn - Enable Auto Attack"),
-(29618, 0, 1, 0, 61, 0, 100, 0, 0, 0, 0, 0, 0, 21, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, "Snowblind Follower - Linked - Enable Combat Movement"),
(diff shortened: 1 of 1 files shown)


Stoabrogga:
Pushed a new commit with these changes:
- Only add the new entries to the script (if possible please do not overwrite complete scripts, as this is error-prone and ultimately led to my error in the first place).
- Use single quotes instead of double quotes
- Just make entry id 7 not repeatable instead of using timers in order to prevent multiple quest rewards using the same snobold

Already tested this successfully, so can be merged in my opinion. Thanks @Andrei-UPB

Andrei-UPB:
Thanks for the reply.

To be frank, this is my judgement for most if not all of your bullet points:

- If someone using AC updated their script to the one you worked on then noticed the quest being broken and fixed it on their own by adding your improvements on top of the already existing script (instead of replacing it entirely), deleting entries 7,8,9 will just re-write entries 1,2 and 3. That's why I believe it's better to delete all of the entries, and just write everything from scratch, in order to be certain that everyone is on the same page regarding this particular smart script.

I feel that when it comes to a script of this size, complexity wise whether you add 7 lines or 3 or deleting everything in the script or some arbitrary id's which lines might not always contain what we think they do it would be, and rewriting scripts in general, clearing everything is better than just adding on top of because said file might have been worked on previously, in an independent fashion.

- I use double quotes on comments is quite the standard because single quotes require an escape character (usually \) in cases the comment contains a single quote in case of particular words ( don't, won't, can't).

```
"If you're going to use apostrophes, 
       ^

you'll definitely want to use double quotes".
   ^
```
- I didn't know you could make the event non-repeatable, therefore the fix i found was to prevent credit being granted more times by placing an internal cooldown on the event. Blizzard uses to do the same thing with debuffs, that if you're patient enough to let run out, you can "tag" the same creature twice or more for quest completion.


Cheers for the help, today I learnt something, I hope everyone did :)

FrancescoBorzi:
@Andrei-UPB  thanks for your contribution, however, I disagree with this statement:

> If someone using AC updated their script to the one you worked on then noticed the quest being broken and fixed it on their own by adding your improvements on top of the already existing script

AC should be used in the following way:

1) any improvement/changes to the main core (which includes all the blizzlike contents) should be shared with the community via PRs (our AGPL license is quite clear about it)

2) any **custom** additions should be implemented using our [module system](http://www.azerothcore.org/wiki/The-Modular-Structure). Users can create their own custom modules  and they can decide either to share them with the community or keep them private

because of point 1), users should never fix things "on their own". Apart from the license issue, there are several other good reasons not to do it, but that goes beyond the scope of this thread.

Stoabrogga:
@Andrei-UPB 
- Concerning overwriting complete SAI scripts I don't share your opinion:
  - If people use customizations it's their problem, not AC's. As AC contributor you should always base your changes on a fresh DB and clean code base.
  - Overwriting SAI makes it harder to review PRs, because you have to compare the existing script with the new one.
  - As I already told it can cause much trouble by accidentally removing already working parts, like it has happened to me and on several other occasions in the past.

- Concerning single quotes vs double quotes: Please follow the guidlines described here: http://www.azerothcore.org/wiki/Dealing-with-SQL-files
  (single quotes for string values, backticks for field names, no double quotes)

- SAI documentation: http://www.azerothcore.org/wiki/smart_scripts (doesn't harm to take a look at this from time to time, I use it regularly)

PR #2549, opened 2020-01-06
fix(core): Improve output for 2 small errors

BarbzYHOOL:
## CHANGES PROPOSED:
- Give a possible error explanation when authserver can't bind to port, and when a PidFile can't be created.
- Change outString to outError for the pidfile (so it displays in red in the console AND now it's the same code in both worldserver and authserver)

## HOW TO TEST THE CHANGES:

- Launch two authservers on the same port and see the error
- Or have a permission error when creating the pidfile (like try to create it in /root)

But there is nothing to test really :D just check the code

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/worldserver/Master.cpp
+++ b/src/server/worldserver/Master.cpp
@@ -141,7 +141,7 @@ int Master::Run()
     if (!pidFile.empty())
     {
         if (uint32 pid = CreatePIDFile(pidFile))
-            sLog->outError("Daemon PID: %u\n", pid); // outError for red color
+            sLog->outError("Daemon PID: %u\n", pid); // outError for red color in console
         else
         {
             sLog->outError("Cannot create PID file %s (possible error: permission)\n", pidFile.c_str());
--- a/src/server/worldserver/Master.cpp
+++ b/src/server/worldserver/Master.cpp
@@ -141,10 +141,10 @@ int Master::Run()
     if (!pidFile.empty())
     {
         if (uint32 pid = CreatePIDFile(pidFile))
-            sLog->outString("Daemon PID: %u\n", pid);
+            sLog->outError("Daemon PID: %u\n", pid); // outError for red color
         else
         {
-            sLog->outString("Cannot create PID file %s.\n", pidFile.c_str());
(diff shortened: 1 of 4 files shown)


src/server/worldserver/RemoteAccess/RARunnable.cpp
```diff
@@ -56,7 +56,7 @@ void RARunnable::run()
 
     if (acceptor.open(listenAddress, m_Reactor) == -1)
     {
-        sLog->outError("Trinity RA can not bind to port %d on %s", raPort, stringIp.c_str());
+        sLog->outError("Trinity RA can not bind to port %d on %s (possible error: port already in use)", raPort, stringIp.c_str());
```
Viste: we can also change Trinity RA in this msg :)

BarbzYHOOL: i know but it must be done in another PR

src/server/worldserver/Master.cpp
```diff
@@ -141,10 +141,10 @@ int Master::Run()
     if (!pidFile.empty())
     {
         if (uint32 pid = CreatePIDFile(pidFile))
-            sLog->outString("Daemon PID: %u\n", pid);
+            sLog->outError("Daemon PID: %u\n", pid); // outError for red color
```
Stoabrogga: Could you please also add this comment here?
https://github.com/azerothcore/azerothcore-wotlk/blob/4a8f1de5381d1cf53cdcc4c5bb7d9ef7c99a8a8f/src/server/game/World/World.cpp#L2005

Viste: +1

BarbzYHOOL: we don't have anything like that already?? @Stoabrogga 

btw I couldn't find any error message about "can not bind to 0.0.0.0:port" for the worldserver (i'll try to see what happens to run two worldservers once i update my DB, today or tomorrow)

BarbzYHOOL: ah ok didn't understand, yes I will do it

BarbzYHOOL:
lol yesterday i found the error for launching two worldservers on same port, will make a tiny PR i guess

PR #2566, opened 2020-01-11
feat(Core/Config): rework configs load system

Winfidonarleyan:
PR based @BarbzYHOOL idea in PR (#2560)

## CHANGES PROPOSED:
* Add new function for load modules configs
* Add messages for the modules using config variant (see screenshots)
* Rework loading config method
* Replace type in functions `Get***Default` from `const char* name` to `std::string const& name` (for new logging system)
* Now the loading of configs modules is separated from the configs servers.  See `LoadAppConfigs()` and `LoadModulesConfigs();`

### Before
1. Load `conf..dist` conf
```cpp
if (!sConfigMgr->LoadInitial(cfg_def_file.c_str())) 
{
    printf("ERROR: Invalid or missing default configuration file : %s\n", cfg_def_file.c_str());
    return 1;
}
```
2. Load `.conf` conf
```cpp
if (!sConfigMgr->LoadMore(cfg_file))
{
    printf("WARNING: Invalid or missing configuration file : %s\n", cfg_file);
    printf("Verify that the file exists and has \'[worldserver]' written in the top of the file!\n");
}
```
3. Add list configs in `World` class in `game` library
```cpp
//set module config file list
sWorld->SetConfigFileList(CONFIG_FILE_LIST);
```
4. Load modules configs in fuction `SetInitialWorldSettings`
```cpp
///- Initialize the World
sWorld->SetInitialWorldSettings();
```
### After
1. Add list configs files (server and modules if app == worldserver)
```cpp
sConfigMgr->SetConfigList(std::string(configFile), std::string(CONFIG_FILE_LIST));
```
2. Load apps configs
```cpp
if (!sConfigMgr->LoadAppConfigs())
   return 1;
```
3. Load modules config (in Config class) after loaded db
```cpp
sConfigMgr->LoadModulesConfigs();
```
---

## TESTS PERFORMED:
- Build and run in Win at VS `16.7.0 Preview 3.1`
- Build and run in Linux `Debian10` at `Clang7`

## HOW TO TEST THE CHANGES:
- Build and run auth/worldserver with configs modules
- Check logs in console/file
- See screenshots for example

## Target branch(es):
- [x] Master

## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

### Media
- Screenshots
image
image

```
[...]

PR #2567, opened 2020-01-11
feat(Core/Logging): rework logging to POCO::Logging

Winfidonarleyan:
TODO
- [ ] #2566 
- [ ] auth
- [ ] common
- [ ] debug


BarbzYHOOL:
this PR must be done first before the config one, no ? so we can use the logging in it

Winfidonarleyan:
No, after config only

BarbzYHOOL:
> Replace type in functions Get**Default from const char name to std::string const& name (for new logging system)

this from https://github.com/azerothcore/azerothcore-wotlk/pull/2566 requires Pocco logging ??

Winfidonarleyan:
For this pr need pr #2566

PR #2576, opened 2020-01-13
refactor(CORE/scripts): Clearer error when script unassigned or alrea…

BarbzYHOOL:
Just make the error clearer with better english to immediately understand where the problem lies (as I was slightly confused when I saw the first error)

No need to test, just look at the files

Note: I don't really know if this whole block of code is useful at all as it's commented out https://github.com/BarbzYHOOL/azerothcore-wotlk/blob/d6ca699702286cdf54821302caaba3975ba427ba/src/server/game/Scripting/ScriptMgr.cpp#L68


Meltie2013:
@BarbzYHOOL That line of code you mention is very old and dates back to 2012 at the most, but was later rewritten later on. Here's the file from TrinityCore at that time not commented out.

https://github.com/TrinityCore/TrinityCore/blob/cc65aba7899cdd753dc77d3e1cf7e3f886173c1a/src/server/game/Scripting/ScriptMgr.cpp#L36-L131

BarbzYHOOL:
> @BarbzYHOOL That line of code you mention is very old and dates back to 2012 at the most, but was later rewritten later on. Here's the file from TrinityCore at that time not commented out.
> 
> https://github.com/TrinityCore/TrinityCore/blob/cc65aba7899cdd753dc77d3e1cf7e3f886173c1a/src/server/game/Scripting/ScriptMgr.cpp#L36-L131

alright but we can probably remove it no? also it's exactly the same code in the scriptMgr.h if i recall

Meltie2013:
If its commented out, you could remove it.

BarbzYHOOL:
I removed the commented code, if someone has something against this, just tell me and i'll revert that last commit

Winfidonarleyan:
Maybe move code from .h to .cpp?

BarbzYHOOL:
> Maybe move code from .h to .cpp?

yeah seems logic, should I do that?

Meltie2013:
@BarbzYHOOL Most important, you should find out if that line of code is duplicate or not; or find out a logical reason why it was commented out in the first place to be honest.

Stoabrogga:
Considering the latest discussion: Is this PR finished or still WIP?

BarbzYHOOL:
forgot to move the code but well, anyway i didn't introduce anything new, it was already like that in the old code ^^

PR #2577, opened 2020-01-14
fix(DB/SAI): Devilsaurs

Stoabrogga:
## CHANGES PROPOSED:
- Set the Devilsaurs in Un'Goro Crater active in order to be able to move on their own and pose an additional threat to the players.
- Delete the duplicate Tyrant Devilsaur

## ISSUES ADDRESSED:
none

## TESTS PERFORMED:
tested successfully in-game

## HOW TO TEST THE CHANGES:
just follow the Devilsaurs:
`.go cr id 6498` (Devilsaur)
`.go cr id 6499` (Ironhide Devilsaur)
`.go cr id 6500` (Tyrant Devilsaur)
`.go cr id 6584` (King Mosh)
They should move on their own even when no player is near.

none

## Target branch(es):
- [x] Master
 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1579041492812648634.sql
+++ b/data/sql/updates/pending_db_world/rev_1579041492812648634.sql
@@ -7,6 +7,9 @@ UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 6499;
 DELETE FROM `creature` WHERE `guid` = 23744;
 DELETE FROM `creature_addon` WHERE `guid` = 23744;
 
+-- All Devilsaurs: Set large in order to increase their visibility distance
+UPDATE `creature_addon` SET `isLarge` = 1 WHERE `guid` IN (23741,23743,23745,24437);
+
 -- All Devilsaurs: Set active in order to enable them walking around on their own
 DELETE FROM `smart_scripts` WHERE `entryorguid` = 6499 AND `source_type` = 0;
 DELETE FROM `smart_scripts` WHERE `entryorguid` IN (6498,6500) AND `source_type` = 0 AND `id` = 1;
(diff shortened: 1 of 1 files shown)


BarbzYHOOL:
Just being curious, before that, the devilsaurs didn't move at all ?

Stoabrogga:
They move, but only if a player is near and they stop if running out of view distance (because then they don't get any updates).

BarbzYHOOL:
well i always thought it was because of the grid unload setting but you told us it is not implemented

the thing is that I've seen TONS of patrols having that behaviour (example: the allies in the barrens, the patrol in stromgarde)

Stoabrogga:
Yes, but this is not necessary for those creatures. The "Active" mode should only be used sparsely for creatures like the Fel Reaver or the Storm Giant, creatures which should be able to "surprise" the player.

Creatures who are out of the range of players or active creatures don't get any updates anymore (this is due to performance).

BarbzYHOOL:
the patrols i'm mentioning should surprise the player, i mean it's just logic (and it's awful to see them stuck when flying over them on wyvern), they should always patrol, same for the Giant in arathi that starts a quest (myzrael if I recall)

i guess there are many other patrols that are wrong then and i'm wondering if it's the right way to proceed: to find them one by one and add the active option (but maybe it is)

Stoabrogga:
The Stone Guardians in Un'Goro Crater are neutral, they are no threat. I'd also recommend to not overuse active creatures as each one adds a little bit CPU load (the creature itself is updated and also each object within 50 yards radius if I remember correctly).

BarbzYHOOL:
> The Stone Guardians in Un'Goro Crater are neutral, they are no threat. I'd also recommend to not overuse active creatures as each one adds a little bit CPU load (the creature itself is updated and also each object within 50 yards radius if I remember correctly).

hmm ok for the stone guardians

Stoabrogga:
Added King Mosh (thanks @BarbzYHOOL for pointing this out).

Endeffect-AT:
Tested on win 10 pro 64 bit with ac commit bb42c7c+ without errors.

I ported my char to the given creatures and noticed the coords. A few seconds later i ported myself to Orgimmar or something and back to the creature. 

Every creature has moved on their own as expected! 👍

BarbzYHOOL:
thank you endeffect

PR #2583, opened 2020-01-16
fix(Core/Player): Only save players glyphs when needed

Shard-MW:
I posted it on Discord : 
I found out that for optimization reasons, AC use `NeedToSaveGlyphs()` to know if glyphs should be re-saved into db or not. So, according to `SetGlyph` function, `NeedToSaveGlyphs()` is becoming `true` (called by `SetNeedToSaveGlyph(true)`) when the player add/change a glyph. (That's pretty normal)
But, my question is, why is `SetNeedToSaveGlyph(false)` never called ? Even after save ? It means that it'll continue to re-save player's glyphs into db even if they were already saved and unchanged.

## CHANGES PROPOSED:
-  Add `SetNeedToSaveGlyph(false)` after Glyphs got saved 

## ISSUES ADDRESSED:
Commit : 70170f0bbd767c7d4656b9a3e33b7e21429f00ab
- `SetNeedToSaveGlyph(false)` was never used

## TESTS PERFORMED:
- Checked if glyphs are still saved
- Checked if the `Player::_SaveGlyphs()` function was NOT called when the player didn't rechange his glyphs (after a save).

Tested on Windows 10.

## HOW TO TEST THE CHANGES:
1. Add/Update glyph to your player
2. .saveall, wait for save or restart the server
3. Log into the game and check if the glyph is still there (and makes effect) or not.

## Target branch(es):
- [x] Master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -26648,6 +26648,8 @@ void Player::_SaveGlyphs(SQLTransaction& trans)
 
         trans->Append(stmt);
     }
+
+    SetNeedToSaveGlyphs(false);
 }
 
 void Player::_LoadTalents(PreparedQueryResult result)


FrancescoBorzi:
> Checked if the Player::_SaveGlyphs() function was NOT called when the player didn't rechange his glyphs (after a save).

Did you test the opposite? I mean: in `master` branch (without your PR), did you find useless calls to that method?

Shard-MW:
@FrancescoBorzi 
Yes, `_SaveGlyphs(...)` is called in `void Player::SaveToDB(bool create, bool logout)` only

Stoabrogga:
@FrancescoBorzi 
As the player data is saved regularly to the DB, but `SetNeedToSaveGlyph(false)` is never used and "m_NeedToSaveGlyphs" also never again set to false, the glyphs are saved every time which is not necessary.

Stoabrogga:
Tested successfully, Glyphs are still saved to the DB correctly. Can be merged in my opinion.

PR #2587, opened 2020-01-16
feat(DB): Test accounts using various GM levels

BarbzYHOOL:
## CHANGES PROPOSED:
- Differentiate 5 different type of test accounts for quick testing. People should rename/remove/change password from them though
- Add new column `comment` to "account_access"

Read some stuff from here for the default value https://dba.stackexchange.com/questions/28086/empty-longtext-or-null and other links, so I decided to make the comment empty instead of null 

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/1417 (old PR of mine)

## TESTS PERFORMED:
- None yet, if someone can test, it'd be cool or just remind me
-

## HOW TO TEST THE CHANGES:

Just run the SQL and see if any error (for the UPDATE statements)

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_auth/rev_1579213352894781043.sql
+++ b/data/sql/updates/pending_db_auth/rev_1579213352894781043.sql
@@ -20,7 +20,7 @@ WHERE (`account`.`username` = 'test1' AND `account`.`sha_pass_hash` = '047ce2264
 UPDATE `account`
 INNER JOIN `account_access`
 ON `account`.`id` = `account_access`.`id`
-SET `account_access`.`comment` = 3, `account_access`.`comment` = 'Test account - Ingame Admin'
+SET `account_access`.`gmlevel` = 3, `account_access`.`comment` = 'Test account - Ingame Admin'
 WHERE (`account`.`username` = 'test3' AND `account`.`sha_pass_hash` = 'e546bbf9ca93ae5291f0b441bb9ea2fa0c466176') OR
 (`account`.`username` = 'test4' AND `account`.`sha_pass_hash` = '61015d83b456a9c6a7defdff07f55265f24097af');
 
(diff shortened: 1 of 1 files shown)


BarbzYHOOL:
if wanted, i can remove the lock, I don't know I put it just in case

it should work now

Meltie2013:
@BarbzYHOOL Suggestion for you, instead of directly adding this to the auth database. You should have it as a option to install. Don't get me wrong, its nice to have; but some can be forgetful about the test accounts.

:+1:

BarbzYHOOL:
it will only update if the test accounts match the name and hash_pass, so if they have not been modified only (I asked francesoc borzi for that)

FrancescoBorzi:
@Meltie2013 this is not affecting existing servers where those accounts have been deleted/modified

FrancescoBorzi:
what's the status of this? has it been tested?

BarbzYHOOL:
you tested it francesco if i recall :P

FrancescoBorzi:
@BarbzYHOOL I only tested the 2 types of queries one by one, not the full thing

Kaev:
- You only change account_access.comment instead of account_access.gmlevel
- You shouldn't have the password hashes in the query since it's possible that people have changed the passwords for their test accounts (let's hope people did this for their public servers atlest...)

FrancescoBorzi:
> You shouldn't have the password hashes in the query since it's possible that people have changed the passwords for their test accounts (let's hope people did this for their public servers atleast...)

@Kaev that's on purpose: so we only change for those accounts which still have the default values

gengarshadowball:
Everything looks good:

image

BarbzYHOOL:
ty gengar

gengarshadowball:
I found an issue that I didn't find when testing:
```/* SQL Error (1314): LOCK is not allowed in stored procedures */```

Maybe revert merge, remove the lock and merge again?

PR #2600, opened 2020-01-20
feat(Core/Hook): New PlayerScript hooks

Stoabrogga:
## CHANGES PROPOSED:
Two new PlayerScript hooks to use by the [BeastMaster module](https://github.com/azerothcore/mod-npc-beastmaster):
CustomPetAllowedForDeathKnight
OverridePetType

This way it is finally possible to get rid of the BeastMaster's [patch file](https://github.com/azerothcore/mod-npc-beastmaster/blob/77df961c0bcc5e2cad71421956549be6043c6626/Pet.cpp.beastmaster.patch).

## ISSUES ADDRESSED:
none

## TESTS PERFORMED:
- tested with and without a modified version of the BeastMaster module

## HOW TO TEST THE CHANGES:
This can be tested using this PlayerScript which enables the Death Knight for custom Pets and uses another algorithm to determine the pet type:
```cpp
class PetInit_PlayerScript : public PlayerScript
{
public:
    PetInit_PlayerScript() : PlayerScript("PetInit_PlayerScript") { }

    bool CustomPetAllowedForDeathKnight(Player* /*player*/) override
    {
        return true;
    }

    uint8 OverridePetType(Player* /*player*/, CreatureTemplate const* cinfo) override
    {
        if (cinfo->IsTameable(true))
        {
            return HUNTER_PET;
        }
        else
        {
            return SUMMON_PET;
        }
    }
};
```

none

## Target branch(es):
- [x] Master
 
### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -154,7 +154,12 @@ bool Pet::LoadPetFromDB(Player* owner, uint8 asynchLoadType, uint32 petentry, ui
 
     // DK Pet exception
     if (owner->getClass() == CLASS_DEATH_KNIGHT && !owner->CanSeeDKPet())
-        return false;
+    {
+        if (!sScriptMgr->CustomPetAllowedForDeathKnight(owner))
+        {
+            return false;
+        }
+    }
 
     uint32 ownerid = owner->GetGUIDLow();
     PreparedStatement* stmt;
@@ -732,7 +737,19 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
             petType = HUNTER_PET;
             m_unitTypeMask |= UNIT_MASK_HUNTER_PET;
         }
-        else
+
+        switch (sScriptMgr->OverridePetType(m_owner->ToPlayer(), cinfo))
+        {
+            case SUMMON_PET:
+                petType = SUMMON_PET;
+                break;
+            case HUNTER_PET:
+                petType = HUNTER_PET;
+                m_unitTypeMask |= UNIT_MASK_HUNTER_PET;
+                break;
+        }
+
+        if (petType != SUMMON_PET && petType != HUNTER_PET)
             sLog->outError("Unknown type pet %u is summoned by player class %u", GetEntry(), m_owner->getClass());
     }
 
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -154,12 +154,8 @@ bool Pet::LoadPetFromDB(Player* owner, uint8 asynchLoadType, uint32 petentry, ui
 
     // DK Pet exception
     if (owner->getClass() == CLASS_DEATH_KNIGHT && !owner->CanSeeDKPet())
-    {
         if (!sScriptMgr->CustomPetAllowedForDeathKnight(owner))
-        {
             return false;
-        }
-    }
 
     uint32 ownerid = owner->GetGUIDLow();
     PreparedStatement* stmt;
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -732,7 +737,19 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
...
+                petType = SUMMON_PET;
+                break;
+            case HUNTER_PET:
+                petType = HUNTER_PET;
+                m_unitTypeMask |= UNIT_MASK_HUNTER_PET;
+                break;
+        }
+
+        if (petType != SUMMON_PET && petType != HUNTER_PET)
             sLog->outError("Unknown type pet %u is summoned by player class %u", GetEntry(), m_owner->getClass());
```
FrancescoBorzi: why not put this last line inside a `default` of the above `switch` ?

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -1755,6 +1755,32 @@ void ScriptMgr::OnPetInitStatsForLevel(Pet* pet)
     FOREACH_SCRIPT(PlayerScript)->OnPetInitStatsForLevel(pet);
 }
 
+bool ScriptMgr::CustomPetAllowedForDeathKnight(Player* player)
+{
+    bool ret = false;
+
+    FOR_SCRIPTS_RET(PlayerScript, itr, end, ret) // return false as default
+        if (itr->second->CustomPetAllowedForDeathKnight(player))
+            ret = true; // change ret value only if at least one script returns true
```
FrancescoBorzi: as soon as we find one script, we can already `return true;`

I guess there is no need to use `bool ret = false;`

just replace line 1764 with `return true` and line 1766 with `return false`

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -1755,6 +1755,32 @@ void ScriptMgr::OnPetInitStatsForLevel(Pet* pet)
...
+uint8 ScriptMgr::OverridePetType(Player* player, CreatureTemplate const* cinfo)
+{
+    uint8 ret = MAX_PET_TYPE;
+
+    FOR_SCRIPTS_RET(PlayerScript, itr, end, ret) // return MAX_PET_TYPE as default
+    {
+        uint8 petType = itr->second->OverridePetType(player, cinfo);
+
+        if (petType != MAX_PET_TYPE)
+            ret = petType;
```
FrancescoBorzi: same as above

src/server/game/Entities/Pet/Pet.cpp
Stoabrogga: Because the petType can be set before the switch statement and the switch statement just handles the hook.

src/server/game/Scripting/ScriptMgr.cpp
Stoabrogga: ok, I changed this part

src/server/game/Scripting/ScriptMgr.cpp
Stoabrogga: ok, I changed this part

Yehonal:
Those hooks are not well written, they are not for generic usage and strictly related to the beastmaster module.

The first should be changed to another name to be used more generically 

The second hook should implement the switch case directly inside and you can pass variables to change as a reference to be modified by the hook itself, so you can remove the switch feom the core

Stoabrogga:
Ok, sorry, but I don't have the time to constantly change things over and test them again, this is not worth the effort. Better stick to the patch file then.

PR #2615, opened 2020-01-23
feat(Core/Config): Parameter to set all creatures with WP movement active

Stoabrogga:
## CHANGES PROPOSED:
- Introduce a new parameter for the worldserver configuration:
  "SetAllCreaturesWithWaypointMovementActive"
  This will set all creatures active which use waypoint movement after they are loaded (on grid load). I used this for testing but it could be useful if you have enough CPU power and combine this with "PreloadAllNonInstancedMapGrids" which enables all creatures to start their movement on server startup.
- Remove max visibility distance for active creatures, as this is handled by [creature_addon.isLarge](http://www.azerothcore.org/wiki/creature_addon#islarge)

## ISSUES ADDRESSED:
none

## TESTS PERFORMED:
- tested successfully in-game

## HOW TO TEST THE CHANGES:
Enable the parameter "SetAllCreaturesWithWaypointMovementActive" and visit a few creatures with waypoint movement, e.g.:
- `.go cr id 3237` (Kodos)
- `.go cr id 1738` (Deathguard Terrence)
They should now keep moving around even if no player is near.

none

## Target branch(es):
- [x] Master
 

src/server/game/Entities/Object/Object.h
```diff
@@ -977,7 +977,7 @@ class WorldObject : public Object, public WorldLocation
 
         bool isActiveObject() const { return m_isActive; }
         void setActive(bool isActiveObject);
-        bool IsVisibilityOverridden() const { return m_isVisibilityDistanceOverride || m_isActive; }
```
Yehonal: why? you're using setActive() so you can keep it

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1520,9 +1520,7 @@ float WorldObject::GetGridActivationRange() const
 
 float WorldObject::GetVisibilityRange() const
 { 
-    if (isActiveObject() && !ToPlayer())
```
Yehonal: why you removed this check?

Stoabrogga: No need to increase visibility distance for active creatures. This should be controlled solely via "isLarge" flag in creature_addon or creature_template_addon.

src/server/game/Entities/Object/Object.h
Stoabrogga: No need to increase visibility distance for active creatures. This should be controlled solely via "isLarge" flag in creature_addon or creature_template_addon.

BarbzYHOOL:
And if `PreloadAllNonInstancedMapGrids` is not enabled, then it will still work but after a player has visited the grid, right?

Stoabrogga:
> 
> 
> And if `PreloadAllNonInstancedMapGrids` is not enabled, then it will still work but after a player has visited the grid, right?

Yes. Btw, I tested this using a system with 10 cores and these worldserver config options:
```
UseProcessors = 1023
MapUpdate.Threads = 10
```
CPU usage without preloading and without WP active creatures was around 1% per core and diff was 10ms. With preloading and WP active creatures (there are over 5500 creatures with WP movement) it was around 12% per core and diff was 15ms.

Stoabrogga:
I think this can be merged. The default value for the new option doesn't change anything and I tested the enabled option for several weeks now without any issues.

BarbzYHOOL:
alright good

Issue #2620, opened 2020-01-26
Quest items not being removed from inventory upon completion of quest


jwh1981:
##### SMALL DESCRIPTION:
Quest items are not removed after completing a quest (at least in the human starting zone)

##### EXPECTED BLIZZLIKE BEHAVIOUR:
When completing Millie's harvest (human), the grapes should be removed.

##### CURRENT BEHAVIOUR:
The current behavior is that any quest items are not removed from inventory upon completing a quest. For instance, diseased wolf hides are not removed when completing Wolves Across the Border (given by Eagan Peltskinner) nor are the grapes removed when completing Millie's Harvest quest.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Accept Millie's Harvest quest in human starting zone (Northshire)
2. Retrieve grape harvest
3. Turn in quest to Millie
4. Grapes remain inventory (clearing showing as being quest items and not normal items)

##### EXTRA NOTES:
This is running on a Windows 8.1 x64 Pro desktop with MySQL 5.7 and compiled with Visual Studio 2017 community (as x64) and release (not debug)

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
999d588c37ef0e56d9043f1aa33ad5e9796f74ca

##### OPERATING SYSTEM:
Windows 8.1 x64 Pro with all updates

##### MODULES:
None

##### OTHER CUSTOMIZATIONS:
None

Ashland:
Can confirm on latest commit , quest items not deleting

FrancescoBorzi:
@Kaev

Kaev:
My bad, had a brainfart in the following PR: https://github.com/azerothcore/azerothcore-wotlk/pull/2617
This will be fixed with: https://github.com/azerothcore/azerothcore-wotlk/pull/2626

FrancescoBorzi:
thanks @Kaev for quickly fixing this!

PR #2621, opened 2020-01-27
Update Player.cpp Fix Item not deleting after quest

Ashland:
fix item not deleting after quest completion

## CHANGES PROPOSED:
-  
-  

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2620

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -15838,7 +15838,7 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
     {
         if (ItemTemplate const* itemTemplate = sObjectMgr->GetItemTemplate(quest->RequiredItemId[i]))
         {
-               (quest->RequiredItemCount[i] > 0 && itemTemplate->Bonding == BIND_QUEST_ITEM && !quest->IsRepeatable() && !HasQuestForItem(quest->RequiredItemId[i], quest_id, true))
+               (quest->RequiredItemCount[i] > 0 && itemTemplate->Bonding == BIND_QUEST_ITEM && !quest->IsRepeatable() && !HasQuestForItem(quest->RequiredItemId[i], quest_id, true));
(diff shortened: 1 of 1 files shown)


Stoabrogga:
As a side note: This was changed in this commit:
https://github.com/azerothcore/azerothcore-wotlk/commit/999d588c37ef0e56d9043f1aa33ad5e9796f74ca

Perhaps we should just remove the if statement as exactly the same code is executed in the else block. Or we take over the TC implementation:
https://github.com/TrinityCore/TrinityCore/blob/ed939325374216d6a049e778d6ad864c65283ae6/src/server/game/Entities/Player/Player.cpp#L15077-L15080

https://github.com/TrinityCore/TrinityCore/blob/ed939325374216d6a049e778d6ad864c65283ae6/src/server/game/Entities/Player/Player.cpp#L15087-L15090

src/server/game/Entities/Player/Player.cpp
```diff
@@ -15840,6 +15840,8 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
         {
             if (quest->RequiredItemCount[i] > 0 && itemTemplate->Bonding == BIND_QUEST_ITEM && !quest->IsRepeatable() && !HasQuestForItem(quest->RequiredItemId[i], quest_id, true))
                 DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true);
+            else
+                DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true);
```
FrancescoBorzi: then why not removing the `if` condition?

Ashland: Ok I’ll try that and test it

Kaev:
Use this PR instead: https://github.com/azerothcore/azerothcore-wotlk/pull/2626

Ashland:
thank you @Kaev

PR #2622, opened 2020-01-27
potential crashes/bugs reported by static analysis #2

Kaev:
## CHANGES PROPOSED:
-  Fixed a few crashes/bugs that i found via static code analysis
- Removed duplicated code

## TESTS PERFORMED:
- Builds without errors

## Target branch(es):
- [x] Master

 

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -526,10 +526,6 @@ int32 AuraEffect::CalculateAmount(Unit* caster)
         switch (GetAuraType())
         {
             case SPELL_AURA_PERIODIC_DAMAGE:
-                // xinef: save caster depending auras, always pass 1 as stack amount, effect will be multiplicated at the end of the function by correct value!
```
Viste: why?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -15277,10 +15276,6 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
                         takeCharges = true;
                     break;
                 case SPELL_AURA_MECHANIC_IMMUNITY:
-                    // Compare mechanic
```
Viste: why?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Kaev: It is the same code as case SPELL_AURA_PERIODIC_LEECH. This case slips into case SPELL_AURA_PERIODIC_LEECH now. :)

src/server/game/Entities/Unit/Unit.cpp
Kaev: Same code as SPELL_AURA_MOD_MECHANIC_RESISTANCE, case slips into it now

src/server/game/Entities/DynamicObject/DynamicObject.cpp
```diff
@@ -135,7 +135,7 @@ void DynamicObject::Update(uint32 p_time)
             _aura->UpdateOwner(p_time, this);
 
         // _aura may be set to null in Aura::UpdateOwner call
-        if (_aura && (_aura->IsRemoved() || _aura->IsExpired()))
+        if ((_aura->IsRemoved() || _aura->IsExpired()))
```
Helias: are we sure about this? it doesn't complain if **_aura** is null while it checks its method?

Helias:
This PR seems pretty good, what tool do you use to make this?

src/server/game/Entities/DynamicObject/DynamicObject.cpp
Kaev: Yes, because this this block is inside of an if (_aura) statement: https://github.com/azerothcore/azerothcore-wotlk/blob/bb4dc311593802e7fdf73a7aa6470da7ab9360d4/src/server/game/Entities/DynamicObject/DynamicObject.cpp#L132

Kaev:
> 
> 
> This PR seems pretty good, what tool do you use to make this?

I use PVS Studio for the static code analysis. There are several hundreds more optimizations and even some bugs and other things that we can fix. PM me in Discord and we can talk further about this. :)

src/server/game/Entities/DynamicObject/DynamicObject.cpp
Kaev: Oh damn, i didn't saw the UpdateOwner call - You're right with this one. Good catch!

src/server/game/Entities/Unit/Unit.cpp
Viste: nope not the same here we also make takeCharges true

src/server/game/Spells/Auras/SpellAuraEffects.cpp
Viste: ok agree, just ask because of the second case. to make sure it's okay.

Viste: this can be resolved

src/server/game/Entities/Unit/Unit.cpp
Kaev: Please recheck, SPELL_AURA_MOD_MECHANIC_RESISTANCE also sets makes takeCharges true

Viste: ok nvm its github shit view i look in unit cpp through vim and i see all is ok can be resolved

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7794,7 +7794,7 @@ bool Player::IsOutdoorPvPActive()
 void Player::DuelComplete(DuelCompleteType type)
 { 
     // duel not requested
-    if (!duel)
+    if (!duel || duel->opponent || duel->initiator)
```
Stoabrogga: Shouldn't this be
```cpp
if (!duel || !duel->opponent || !duel->initiator)
```
or am I missing something?

src/server/scripts/Northrend/FrozenHalls/PitOfSaron/pit_of_saron.cpp
```diff
@@ -129,6 +129,7 @@ class npc_pos_leader : public CreatureScript
                                 n2->GetMotionMaster()->MovePoint(1, NecrolytePos2);
                                 n2->SetUInt32Value(UNIT_NPC_EMOTESTATE, EMOTE_STATE_READY1H);
                             }
+                            // TODO This spell check ins invalid
                             if (SPELL_NECROLYTE_CHANNELING)
```
Stoabrogga: Why not just remove all occurrences of "if (SPELL_NECROLYTE_CHANNELING)"? They are always true.

Kaev: I don't know if this statement is there for a reason. I just added a comment because at that place probably should be a check.

src/server/game/Entities/Player/Player.cpp
Kaev: Good catch!

src/server/scripts/Northrend/FrozenHalls/PitOfSaron/pit_of_saron.cpp
Stoabrogga: Agreed, we should fix this in another PR. I pushed a new commit fixing a typo in the comment, otherwise this PR looks good to me.

Stoabrogga:
Build is ok, I think this can be merged.

PR #2630, opened 2020-02-01
fix(Core/Misc): fix some exploits

Viste:
## CHANGES PROPOSED:
-  fix some exploits with guild create command, send wrong packet to stuned creature  and  crashfix on summon spawn/despawn

## TESTS PERFORMED:
- build 

## Target branch(es):
- [x] Master

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1012,7 +1012,7 @@ void WorldObject::setActive(bool on)
 
     m_isActive = on;
 
-    if (!IsInWorld())
+    if (on && !IsInWorld())
         return;
 
     Map* map = FindMap();
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -16765,6 +16765,9 @@ void Unit::SetStunned(bool apply)
 {
     if (apply)
     {
+        if (m_rootTimes > 0) // blizzard internal check?
+            m_rootTimes++;
+
         SetTarget(0);
         SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_STUNNED);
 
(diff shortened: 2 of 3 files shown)


BarbzYHOOL:
Can you also deal with this issue I found today:

```SQL
SQL(p): INSERT INTO guild (guildid, name, leaderguid, info, motd, createdate, EmblemStyle, EmblemColor, BorderStyle, BorderColor, BackgroundColor, BankMoney) VALUES(1, 'blablablabla SUPERWOW clan', 16, '', 'No message set.', 1580579935, 0, 0, 0, 0, 0, 0)
 [ERROR]: [1406] Data too long for column 'name' at row 1
Unhandled MySQL errno 1406. Unexpected behaviour possible.
```

How to reproduce:
Type `.guild create Mycharacter blablablabla blablablab blablabla`

If the name is too long it doesn't give an error in game (but it does in console) and it makes you think it worked fine but it DIDNT insert into the DB, so next reboot it will not work and there won't be any guild.


We need to add a check in the core command to prevent this from happening and to give user output ingame and not only on console.

Viste:
> Can you also deal with this issue I found today:
> 
> ```sql
> SQL(p): INSERT INTO guild (guildid, name, leaderguid, info, motd, createdate, EmblemStyle, EmblemColor, BorderStyle, BorderColor, BackgroundColor, BankMoney) VALUES(1, 'blablablabla SUPERWOW clan', 16, '', 'No message set.', 1580579935, 0, 0, 0, 0, 0, 0)
>  [ERROR]: [1406] Data too long for column 'name' at row 1
> Unhandled MySQL errno 1406. Unexpected behaviour possible.
> ```
> 
> How to reproduce:
> Type `.guild create Mycharacter blablablabla blablablab blablabla`
> 
> If the name is too long it doesn't give an error in game (but it does in console) and it makes you think it worked fine but it DIDNT insert into the DB, so next reboot it will not work and there won't be any guild.
> 
> We need to add a check in the core command to prevent this from happening and to give user output ingame and not only on console.

already in PR lol )))))))

BarbzYHOOL:
this `if !sObjectMgr->IsValidCharterName(guildName)` ?

by the way, funny you checked .guild create today, is it because of my change to mod-starter-guild ? xD

Viste:
partly

Tooldev94:
I checked these changes, everything works

src/server/game/Entities/Object/Object.cpp
```diff
@@ -1012,7 +1012,7 @@ void WorldObject::setActive(bool on)
 
     m_isActive = on;
 
-    if (!IsInWorld())
+    if (on && !IsInWorld())
```
Stoabrogga: Why only check if setting active on?

Viste: WorldObject::setActive(false) called after WorldObject::DespawnOrUnsummon() happening because the WorldObject would be set as not active but then Map::RemoveFromActiveHelper() would not be called because of not being in world anymore.
Whenever setActive() gets desync'd with what Map actually contains a rare crash would happen since Map only reads a few values of the active WorldObject, making it hard to reproduce. If this happens again some ASSERT calls should be added (or a check in Map::RemoveFromMap() could be added about if the WorldObject is in Map::m_activeNonPlayers container).

Stoabrogga: Ok, thanks for the explanation!

Yehonal:
Nice catch about the crashfix! Congrats!

Thank you for the rest ;)

BarbzYHOOL:
@Tooldev94 thank you for testing

PR #2648, opened 2020-02-05
[WIP] Implementing core build with github actions

Yehonal:
This implements the CI using github actions instead of travis

work in progress

### DIFF
--- a/.github/workflows/core_build.yml
+++ b/.github/workflows/core_build.yml
@@ -46,5 +46,5 @@ jobs:
           echo "LoginDatabaseInfo     = \"mysql;3306;root;root;acore_auth\"" >> ./env/dist/etc/worldserver.conf
           echo "WorldDatabaseInfo     = \"mysql;3306;root;root;acore_world\"" >> ./env/dist/etc/worldserver.conf
           echo "CharacterDatabaseInfo = \"mysql;3306;root;root;acore_characters\"" >> ./env/dist/etc/worldserver.conf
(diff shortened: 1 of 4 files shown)


Rochet2:
Are there some benefits expected from the change?

Yehonal:
> Are there some benefits expected from the change?

* no timeouts
* faster
* no 3rd party dependency
* more configuration options

that's it for now, however, I'm still experimenting

BarbzYHOOL:
will that mean updating all modules' .travis too? :'(

FrancescoBorzi:
@BarbzYHOOL it's not mandatory

Rochet2:
Other CI services get speed benefits from cache and prebuilt containers (preinstalled dependencies), but not sure you can use those on github actions. So the build would inevitably be slower on average, no? Atm latest build seems to take 17 min on travis and 32 min on github actions.

btw. a comment about clone depth.
On travis I have had some problems in situations where depth was 1 and multiple commits were pushed in short time. The CI jobs sit in queue and when started the clone would get only latest commit I assume. Then when the job is trying to checkout the actual commit to build it would fail.
Just something to think about.

Yehonal:
It's an edge case but it's possibile. However, the last commit should be
always the one to compile, isn't it?

About speed, I'm sure there's something in github action too. It's taking
30 minutes because I'm compiling without PCH

Helias:
Can we delete the branch **Yehonal-patch-2**, @Yehonal ?

BarbzYHOOL:
means yes

PR #2655, opened 2020-02-07
fix(DB): sql import

FrancescoBorzi:


BarbzYHOOL:
seems like it fixed it

maybe we should add somewhere that locking tables is prohibited and that people should just run their queries safely

data/sql/updates/db_auth/2020_02_07_00.sql
```diff
@@ -20,7 +20,6 @@ INSERT INTO `version_db_auth` (`sql_rev`) VALUES ('1579213352894781043');
 ALTER TABLE `account_access` ADD COLUMN `comment` VARCHAR(255) DEFAULT '';
 
 -- UPDATE ACCOUNTS
-LOCK TABLES `account_access` WRITE, `account` WRITE;
 /*!40000 ALTER TABLE `account_access` DISABLE KEYS */;
```
BarbzYHOOL: i hope these things wont interfere either

PR #2659, opened 2020-02-08
fix(DB/SAI): Missing Creature Twilight Corruptor

Endeffect-AT:
NPC id 15625 missing in creature table to complete quest 8735

## CHANGES PROPOSED:
-  Add creature 15625 with new guid 3110362 

Creature was added in patch 1.9.0 and should be there in 3.3.5a since it was never removed. 
Source: https://wowwiki.fandom.com/wiki/Twilight_Corrupter

Source of position found online: 
unnamed
Post from 2009: https://twilight-sucks.livejournal.com/486659.html

## ISSUES ADDRESSED:
- Closes 

## TESTS PERFORMED:
- Imported sql with HeidiSQL on win 10 pro x64
- Checked successfully ingame

## HOW TO TEST THE CHANGES:

1. Import sql file
2. Check creature ingame 

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1581196474128672400.sql
+++ b/data/sql/updates/pending_db_world/rev_1581196474128672400.sql
@@ -1,4 +1,5 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1581196474128672400');
 
+UPDATE `creature_template` SET `flags_extra` = 0 WHERE `entry` = 15625;
 DELETE FROM `creature` WHERE `guid`=3110362;
 INSERT INTO `creature` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `modelid`, `equipment_id`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecs`, `spawndist`, `currentwaypoint`, `curhealth`, `curmana`, `MovementType`, `npcflag`, `unit_flags`, `dynamicflags`, `ScriptName`, `VerifiedBuild`) VALUES (3110362, 15625, 0, 0, 0, 1, 1, 0, 0, -10325.9, -490.507, 50.3208, 5.67792, 300, 0, 0, 832750, 0, 0, 0, 0, 0, '', 0);
--- a/data/sql/updates/pending_db_world/rev_1581196474128672400.sql
+++ b/data/sql/updates/pending_db_world/rev_1581196474128672400.sql
@@ -1,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1581196474128672400');
+
+DELETE FROM `creature` WHERE `guid`=3110362;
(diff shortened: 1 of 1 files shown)


BarbzYHOOL:
Nice thank you, but does the creature have a script too?

Endeffect-AT:
@BarbzYHOOL 
Unfortunately I can not answer you, but I lack the knowledge about smart scripts.

BarbzYHOOL:
Check the related entry in creature_template, look up at the bottom columns "AIName" if it's SmartAI then check the related entry in smart_scripts

Then can test it ingame xD

Endeffect-AT:
If i understand correctly it's a C++ script
image

gengarshadowball:
Yea, that's a core script

Stoabrogga:
It's this CreatureScript:
https://github.com/azerothcore/azerothcore-wotlk/blob/832bccd9fa8ae5f3dd20eaeed244e74cd700f2de/src/server/scripts/EasternKingdoms/zone_duskwood.cpp#L38

Endeffect-AT:
Do I understand correctly that this code has to spawn the creature when the player is on this quest? So it didn't appear all the time?

https://github.com/azerothcore/azerothcore-wotlk/blob/832bccd9fa8ae5f3dd20eaeed244e74cd700f2de/src/server/scripts/EasternKingdoms/zone_duskwood.cpp#L142-L164

Stoabrogga:
Yes, seems to be the case. Probably adding it directly to the creature table is wrong in this case.

Endeffect-AT:
Okay so the script just doesn't work in that case :sweat_smile: 
I'm not a core dev so didn't looked on any script because I can only read the code a little as a layperson

Stoabrogga:
So he does not spawn if you're on the quest "The Nightmare's Corruption"?

Endeffect-AT:
Correct. As I have described above, the quest cannot be completet because the creature does not spawn.

Stoabrogga:
Hmm, the script is assigned to area 4017 in "[areatrigger_scripts](http://www.azerothcore.org/wiki/areatrigger_scripts)"
| entry | ScriptName        |
|-------|-------------------|
|  4017 | at_twilight_grove |

But there doesn't seem to be an area 4017 in 3.3.5:
http://www.azerothcore.org/wiki/AreaTable

Perhaps it is just using the wrong area ID as "Twilight Grove" has the ID 856.

Stoabrogga:
Sorry, I used the wrong area table... 4017 is a reference to this:
http://www.azerothcore.org/wiki/DBC-AreaTrigger

Endeffect-AT:
Is there any way to see creatures like that which are summoned by a script in creature table? If i look at the creature table i couldn't find the ID 15625 and that's why I assumed that the entry is completely missing. 

It would be very helpful for people like me which has not much knowledge about c++.

Stoabrogga:
No, because it is summoned. Summoned creatures are (obviously) not part of the creature table. What you could do is search the source code for the creature ID 15625, you'll then find this:
https://github.com/azerothcore/azerothcore-wotlk/blob/832bccd9fa8ae5f3dd20eaeed244e74cd700f2de/src/server/scripts/EasternKingdoms/zone_duskwood.cpp#L21

I do this regularly when looking for specific spell, creature, gameobject IDs etc. which I assume are used in core scripts.

Endeffect-AT:
Learned something again, thank you @Stoabrogga! 

To solve the problem now should I close the PR and open an isse?

Stoabrogga:
You're welcome. If you don't know why the areatrigger script is not running, then yes, close this PR and open an issue. What you could test if you have the time: Teleport to the area 4017 and check if it is the correct position for the Twilight Grove:
`.go -10636.200195 -391.307007 101.877998 0`

Sadly, I currently don't have the time to test this.

Stoabrogga:
The script works correctly. You need quest 8735 **without** the quest item and have to run down the way into the grove in order to trigger it because areatrigger 4017 is just a small part of that way. It won't work if you teleport into the grove, but players cannot do this anyway.

Endeffect-AT:
That was my mistake, I did not assume an area trigger. It works as expected!

PR #2240, opened 2019-08-26
Core/Misc: Move EventMap into its own source file.

Viste:
##### CHANGES PROPOSED:
-  Move EventMap into its own source file.
-  Move implementation details to .cpp.
- Add std::chrono::duration overloads to EventMap makes possible to write:
```c++
events.ScheduleEvent(1, Seconds(10));
// ... or ...
events.ScheduleEvent(2, Minutes(1) + Seconds(20));
// ... or with C++14 support:
events.ScheduleEvent(2, 45s);
```
##### HOW TO TEST THE CHANGES:
 fight with bosses scripted creatures on battlegrounds and etc all should be as before
also you can rewrite scripts from miliseconds to seconds
##### Target branch(es): master

Thanks to @Naios

PR #2238, opened 2019-08-26
 Common/Util: Introduce a task scheduler which schedules task with c++11 and 17 techniques

Viste:
#### CHANGES PROPOSED:
-  Use std::function as callback (methods, lamdas & std::bind).
-  Uses std::chrono::duration as duration type (Milliseconds, Seconds...) with c++11 user defined literals we could just write "11s" for example (MSVS 2015, gcc & clang).
- Provides a lot of overloads for easier usage:
  - Static durations or random durations between min and max.
  - Named (with group ids - group id 0 is possible) or anonymous tasks.
  - Asynchronous tasks.
- Is adapted to the current needs of scripting:
  - Repeat tasks through the TaskContext
  - Schedule new tasks from the context.
  - Provides access to the times repeat counter of a task,
    this will help to schedule multi step events (dialogs for example).

See https://gist.github.com/Naios/6d143538abdc3d32fd2b for usage examples.

Big thanks to @Naios

also move scheduled map scripts counter methods into MapManage

##### Target branch(es): master

PR #2660, opened 2020-02-09
Core/Misc: Introduce a new task scheduler and move EventMap into its own source file

Viste:
This is a combination of #2240  and #2238 
## CHANGES PROPOSED:
-  Use std::function as callback (methods, lamdas & std::bind).
-  Uses std::chrono::duration as duration type (Milliseconds, Seconds...) with c++11 user defined literals we could just write "11s" for example (MSVS 2015, gcc & clang).
- Provides a lot of overloads for easier usage:
  - Static durations or random durations between min and max.
  - Named (with group ids - group id 0 is possible) or anonymous tasks.
  - Asynchronous tasks.
- Is adapted to the current needs of scripting:
  - Repeat tasks through the TaskContext
  - Schedule new tasks from the context.
  - Provides access to the times repeat counter of a task,
    this will help to schedule multi step events (dialogs for example).

See https://gist.github.com/Naios/6d143538abdc3d32fd2b for usage examples.

Big thanks to @Naios

 - also move scheduled map scripts counter methods into MapManage
-  Move EventMap into its own source file.
-  Move implementation details to .cpp.
- Add std::chrono::duration overloads to EventMap makes possible to write:
```c++
events.ScheduleEvent(1, Seconds(10));
// ... or ...
events.ScheduleEvent(2, Minutes(1) + Seconds(20));
// ... or with C++14 support:
events.ScheduleEvent(2, 45s);
```
## TESTS PERFORMED:
- Build
## HOW TO TEST THE CHANGES: 
 - this useless tests not my problem :)) 
## Target branch(es):
- [x] Master


Viste:
scheduler part tested in #2238

Viste:
No one wants a useless innovation. Those who want to write scripts easier can find I think.

FrancescoBorzi:
what's the status of this? ready to be tested? if yes, what should be checked?

Viste:
> what's the status of this? ready to be tested? if yes, what should be checked?
its ready from feb 9 but only 1% of people can test this lol because need write own script
also
> scheduler part tested in #2238

FrancescoBorzi:
> You just need to check the existing scripts (all dungeons, part of the spells, npc) that everything is OK.

for this 99% of people can do it )

Viste:
sad story but as always all needs only quests

BarbzYHOOL:
no one knows what it's for (you included) so no one cares

also it requires some C++ devs to review this, not a simple quest tester

Viste:
> no one knows what it's for (you included) so no one cares

CHANGES PROPOSED read it and you understand

PR #2665, opened 2020-02-13
feat(Core/Hook): New GlobalScript hook

Stoabrogga:
## CHANGES PROPOSED:
Introduce a new GlobalScript hook which is called before the WorldObject phasemask check. This can be used to get rid of the [Guildhouse module](https://github.com/azerothcore/mod-guildhouse) patch file, but also for other kind of stuff concerning the phases as you are able to access the WorldObject and can also update the phasemask against which the check is running later.

## ISSUES ADDRESSED:
none

## TESTS PERFORMED:
- tested successfully in-game using a small GlobalScript

## HOW TO TEST THE CHANGES:
I used this small GlobalScript to test the hook:
```cpp
class GuildHouseGlobal : public GlobalScript
{
public:
    GuildHouseGlobal() : GlobalScript("GuildHouseGlobal") {}

    void OnBeforeWorldObjectSetPhaseMask(WorldObject const* worldObject, uint32& /*oldPhaseMask*/, uint32& /*newPhaseMask*/, bool& useCombinedPhases, bool& /*update*/) override
    {
        if (worldObject->GetZoneId() == 876)
            useCombinedPhases = false;
        else
            useCombinedPhases = true;
    }
};
```
It ensures that if in the zone "GM Island" (876) the phaseMask is not treated as combined phases, but as single individual phases. This means you only see objects using the exact same phasemask value as the player, so combining phases is no longer possible in this area.

Test 1:
- go to any zone except GM Island
- ensure that GM mode is off and you are in phase 1 (normal phase): `.mod p 1`
- add a chicken: `.npc a t 620`
- target the chicken and set phase 6: `mod p 6` (the chicken will vanish)
- try setting you own phase to phases 1-9
- the chicken will only be visible if the phase mask contains phases 2 or 4, so phasemask values 2,3,4,5,6,7; it will not be visible for phasemask value 1,8,9

Test 2:
- go to GM Island: `.tele GMIsland`
- proceed with the same steps as above
- the chicken will now only be visible for phasemask value 6

none

## Target branch(es):
- [x] Master
 
### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -1057,7 +1057,7 @@ void WorldObject::CleanupsBeforeDelete(bool /*finalCleanup*/)
 void WorldObject::_Create(uint32 guidlow, HighGuid guidhigh, uint32 phaseMask)
 { 
     Object::_Create(guidlow, 0, guidhigh);
-    m_phaseMask = phaseMask;
+    SetPhaseMask(phaseMask, false);
 }
 
 uint32 WorldObject::GetZoneId(bool /*forceRecalc*/) const
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -971,7 +971,7 @@ elunaEvents(NULL),
 #endif
 LastUsedScriptID(0), m_name(""), m_isActive(false), m_isVisibilityDistanceOverride(false), m_isWorldObject(isWorldObject), m_zoneScript(NULL),
 m_transport(NULL), m_currMap(NULL), m_InstanceId(0),
-m_phaseMask(PHASEMASK_NORMAL), m_notifyflags(0), m_executed_notifies(0)
+m_phaseMask(PHASEMASK_NORMAL), m_useCombinedPhases(true), m_notifyflags(0), m_executed_notifies(0)
 {
     m_serverSideVisibility.SetValue(SERVERSIDE_VISIBILITY_GHOST, GHOST_VISIBILITY_ALIVE | GHOST_VISIBILITY_GHOST);
     m_serverSideVisibilityDetect.SetValue(SERVERSIDE_VISIBILITY_GHOST, GHOST_VISIBILITY_ALIVE);
@@ -1058,12 +1058,6 @@ void WorldObject::_Create(uint32 guidlow, HighGuid guidhigh, uint32 phaseMask)
 { 
     Object::_Create(guidlow, 0, guidhigh);
     m_phaseMask = phaseMask;
-}
-
-bool WorldObject::InSamePhase(uint32 phasemask) const
-{
-    sScriptMgr->OnBeforeWorldObjectInSamePhase(this, phasemask);
-    return (GetPhaseMask() & phasemask);
 }
 
 uint32 WorldObject::GetZoneId(bool /*forceRecalc*/) const
@@ -2866,7 +2860,8 @@ void WorldObject::MovePositionToFirstCollisionForTotem(Position &pos, float dist
 }
 
 void WorldObject::SetPhaseMask(uint32 newPhaseMask, bool update)
-{ 
+{
+    sScriptMgr->OnBeforeWorldObjectSetPhaseMask(this, m_phaseMask, newPhaseMask, m_useCombinedPhases, update);
     m_phaseMask = newPhaseMask;
 
     if (update && IsInWorld())
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -40,6 +40,7 @@
 #include "Group.h"
 #include "Chat.h"
 #include "DynamicVisibility.h"
+#include "ScriptMgr.h"
 
 #ifdef ELUNA
 #include "LuaEngine.h"
@@ -1057,6 +1058,12 @@ void WorldObject::_Create(uint32 guidlow, HighGuid guidhigh, uint32 phaseMask)
 { 
     Object::_Create(guidlow, 0, guidhigh);
     m_phaseMask = phaseMask;
+}
+
+bool WorldObject::InSamePhase(uint32 phasemask) const
+{
+    sScriptMgr->OnBeforeWorldObjectInSamePhase(this, phasemask);
+    return (GetPhaseMask() & phasemask);
 }
 
 uint32 WorldObject::GetZoneId(bool /*forceRecalc*/) const
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -799,7 +799,7 @@ class WorldObject : public Object, public WorldLocation
         virtual void SetPhaseMask(uint32 newPhaseMask, bool update);
         uint32 GetPhaseMask() const { return m_phaseMask; }
         bool InSamePhase(WorldObject const* obj) const { return InSamePhase(obj->GetPhaseMask()); }
-        bool InSamePhase(uint32 phasemask) const;
+        bool InSamePhase(uint32 phasemask) const { return m_useCombinedPhases ? GetPhaseMask() & phasemask : GetPhaseMask() == phasemask; }
 
         virtual uint32 GetZoneId(bool forceRecalc = false) const;
         virtual uint32 GetAreaId(bool forceRecalc = false) const;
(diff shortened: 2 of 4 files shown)


Yehonal:
The idea is nice but, please don't do it in this way.

The **InSamePhase** function is a critical method that must reside inside the header file.
That method, in fact, is called by the "Visit" functions that are called hundreds of times per second by the grid notifier (depending on the server population). To improve performances this method must be "inlined": http://www.cplusplus.com/articles/2LywvCM9/

During the compilation, the compiler will copy the function inside the caller instead of referring it (executing an instruction that resides inside the same CPU registry is faster than looking for it inside the cache/ram)

Therefore, this hook can visually decrease the performances of the entire core so I suggest to do benchmarks and/or think another smart way to do it.

However, About this issues, years ago I created a module that used about 14 phases that are unused in-game and can be reused to create entire parallel dimensions. @Winfidonarleyan remember it :P 
I think it could be a good solution for this approach if I've understood what you're trying to do.

Otherwise, another smart way would be to create a static map of values that can be modified by an hook at server runtime, it's ofc still slower than "return (GetPhaseMask() & phasemask);", that's resolved in O(1) , but the compiler should be smart enough to optimize the check when the map is empty.

I'll check if I'll found other solutions, meanwhile I'm open to your observations

Stoabrogga:
@Yehonal 
The GuildHouse module provides a single phase for each guild, so you potentially need **many** individual phases, probably hundreds, that's why the patch is changing the behavior for zone 876:
https://github.com/azerothcore/mod-guildhouse/blob/e56170eaf9ba405c893d6dfa7f9e263f6121f97b/guildhouse.patch

It removes the ability to combine phases for this zone and provides one phase for each phasemask value.

So if you think this hook can impact performance it would be better to close this PR and stick to the patch in my opinion. Btw, I'm just trying to help Barbz a bit on this module, I don't use it myself.

Yehonal:
I've seen the credits about it, maybe @talamortis and @Rochet2 can say something more. I'm looking inside that module code now

Yehonal:
Well, after a fast read about the code, the funny thing I've discovered is that it's using the guild id to set the mask. That's mean that you can set max smallint(5) -> 32767, it's a big number but if you think that a guild can be disbanded, deleted, recreated as many time as you want and that there's nothing to fix the gap. This system could explode after few months on populated servers.

I don't think it's well-engineered. But we can wait for @talamortis and @Rochet2 if they worked on it

Stoabrogga:
@Yehonal 
Where did you see smallint(5)? The guild id is using int(10) in the DB.

Yehonal:
> @Yehonal
> Where did you see smallint(5)? The guild id is using int(10) in the DB.

Ok, there's an error in our wiki: http://www.azerothcore.org/wiki/creature , http://www.azerothcore.org/wiki/gameobject 

The phasemask there is a smallint(5) , we must change it to int(10). 
So, summarizing it up: that "patch" removes the  bitmask feature for that specific area. In this way guild id and phasemask become a bijective countable set offering 4 billions of different non-overlapped phases that could be used in those places where there are no used phases.

It's an interesting feature that could be engineered in the correct way preserving performance  too.
I'll think about it.

Stoabrogga:
I don't know how to create a stress test for this, but I compiled my core including PR #2615, preloading all grids and setting all creatures with WP movement to active (over 5500 active creatures, which all have to check their phases in order to interact with their surroundings). CPU usage was roughly the same with or without the hook.

Stoabrogga:
Executed a third test:
- preloading all grids
- set all creatures with WP movement active
- use this hook with the GlobalScript mentioned in the first comment

Still the same CPU usage.

Stoabrogga:
Btw, just as an additional clarification: What phases the player can see is controlled using spell auras, e.g.:
Spell ID 55773: Apply Aura: Phase (65)
Spell ID 55774: Apply Aura: Phase (129)
Spell ID 59062: Apply Aura: Phase (64)
Spell ID 60815: Apply Aura: Phase (128)

See also aura type 261 "SPELL_AURA_PHASE": https://github.com/azerothcore/azerothcore-wotlk/blob/e22d78ecd6997bc11f8620b759364f119dac431a/src/server/game/Spells/Auras/SpellAuraDefines.h#L311
handled here:
https://github.com/azerothcore/azerothcore-wotlk/blob/c16f21a2715db4182444713c7b1f09417a9a3924/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L1705

Multiple auras can be applied using a great variety of means, e.g.:
- table [spell_area](http://www.azerothcore.org/wiki/spell_area)
- SAI
- core scripts (CreatureScript, AreaTriggerScript, GameObjectScript, ...)
etc.

So this system is very flexible and also the reason why the Guildhouse module has to restore the phasemask of the player according to his auras if not on GM Island (handled via PlayerScript).

Stoabrogga:
@Yehonal 
I think I've found a better solution and pushed a new commit. I added a new boolean data member "m_useCombinedPhases" to WorldObject in order to control how the phaseMask is treated:
- true (default): use phaseMask as bit mask combining up to 32 phases
- false: use phaseMask to represent single phases only (up to 4294967295 phases); can only be set using the new GlobalScript hook "OnBeforeWorldObjectSetPhaseMask"

The function "InSamePhase" will stay in the header file and only add the check for this boolean in order to decide how to calculate the return value. The new hook can be tested using this GlobalScript:
```cpp
class GuildHouseGlobal : public GlobalScript
{
public:
    GuildHouseGlobal() : GlobalScript("GuildHouseGlobal") {}

    void OnBeforeWorldObjectSetPhaseMask(WorldObject const* worldObject, uint32& /*oldPhaseMask*/, uint32& /*newPhaseMask*/, bool& useCombinedPhases, bool& /*update*/) override
    {
        if (worldObject->GetZoneId() == 876)
            useCombinedPhases = false;
    }
};
```

Yehonal:
@Stoabrogga you're amazing ! that's exactly what I was thinking about but I never had the time to formalize it in this week (job & internet issues)

Yehonal:
PS I'm very glad to know that you're encouraged to improve your code instead of discouraging yourself when I suggest to review it (such happened before with other contributors)

Yehonal:
I've approved it. Anyway, a good test is needed before merging it

Stoabrogga:
Thanks. I already tested this multiple times using the instructions in my first comment, but perhaps @BarbzYHOOL can also test this together with this PR for the GuildHouse module:
https://github.com/azerothcore/mod-guildhouse/pull/14

Yehonal:
We must be sure that SetPhaseMask in the WorldObject class is the SSOT used to change the m_phaseMask value, otherwise that hook can be dangerous.

Doing a fast research m_phaseMask is private, and it's nice. But at the moment it's also changed by the _Create method too. So, would be nice to avoid issues, to put the SetPhaseMask inside the _Create method, in this way that variable is always changed by the setter and never directly.

this check then should avoid crashes when the object is not in world yet (at creation phase)
https://github.com/azerothcore/azerothcore-wotlk/blob/38e8503d21ba42f887c2d0c99029cfa4ed67ac93/src/server/game/Entities/Object/Object.cpp#L2865

Stoabrogga:
@Yehonal 
Thanks. I pushed a new commit using "SetPhaseMask" inside of "WorldObject::_Create" to initialize the phase mask.

Yehonal:
Nice! please @BarbzYHOOL can you test it with that module? and if you can do this simple steps:

1) go to the zone specified by the module and test some "custom" phases
2) teleport your player in a known area where there are some normal phases and see that everything is working as intended

BarbzYHOOL:
guys i removed this module days ago...

Yehonal:
Why?

Stoabrogga:
Sorry, but I'm giving up on this. Did an extensive test using "mod-guildhouse", but I did not manage to restore the phases and visibility of the objects correctly after visiting the guild house and teleporting to a new area. This is not worth the effort in my opinion.

Stoabrogga:
Ok, I'm an idiot... I forgot to set the "useCombinedPhases" to "true" if not in the right area:

```cpp
class GuildHouseGlobal : public GlobalScript
{
public:
    GuildHouseGlobal() : GlobalScript("GuildHouseGlobal") {}

    void OnBeforeWorldObjectSetPhaseMask(WorldObject const* worldObject, uint32& /*oldPhaseMask*/, uint32& /*newPhaseMask*/, bool& useCombinedPhases, bool& /*update*/) override
    {
        if (worldObject->GetZoneId() == 876)
            useCombinedPhases = false;
        else
            useCombinedPhases = true;
    }
};
```

The module is working quite well now. Tested teleporting the player from the guild house to Dun Niffelem where you get multiple phasing auras.

BarbzYHOOL:
the hook works fine, no need for the .patch anymore

FrancescoBorzi:
well done @Stoabrogga !

PR #2669, opened 2020-02-16
 Core/Packet: fix crash happening when someone sends small packets that is processed directly to WorldSocket

### LINKED ISSUE #2666: crash server
Server crashes

##### SMALL DESCRIPTION:
after 5-20min server just crashes
Previously, this did not happen and there was an uptime of 4 days. Today there are problems

##### EXTRA NOTES:
```
terminate called after throwing an instance of 'ByteBufferPositionException'
  what():  Attempted to get value with size: 4 in ByteBuffer (pos: 0 size: 8)
Aborted

```
##### BRANCH(ES):
master

##### OPERATING SYSTEM:
ubuntu

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/88521376-crash-server?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Viste:
## ISSUES ADDRESSED:
- Closes #2666 
## Target branch(es):
- [x] Master


src/server/game/Server/WorldSocket.cpp
```diff
@@ -676,7 +676,15 @@ int WorldSocket::ProcessIncoming(WorldPacket* new_pct)
...
         {
             case CMSG_PING:
-                return HandlePing (*new_pct);
+            {
+                try
+                {
+                    return HandlePing(*new_pct);
+                }
+                catch (ByteBufferPositionException const&) {}
+                sLog->outError("WorldSocket::ReadDataHandler(): client sent malformed CMSG_PING");
```
FrancescoBorzi: shouldn't we add the`outError` inside the `catch` block ?

Viste: ByteBufferPositionException already return exception ;)

Viste:
```
ByteBufferPositionException::ByteBufferPositionException(bool add, size_t pos,
                                                         size_t size, size_t valueSize)
{
    std::ostringstream ss;

    ss << "Attempted to " << (add ? "put" : "get") << " value with size: "
       << valueSize << " in ByteBuffer (pos: " << pos << " size: " << size
       << ")";

    message().assign(ss.str());
}
```

Dreaxxx:
Problem : 

When we try to link a Quest in chat (guild/whisp/world,etc...) player client crash. This problem only appear with quests , i'v try with item, spell , achiev , no problem. 
Tested with French client

Viste:
useless

Helias:
> useless

Partially disagree, I tested several times with and without PR sending a malformed packet with the opcode CMSG_PING.

Without PR:
the client crash after a few seconds

With PR:
the client crash instantly.

I think that is better with your PR, we should merge it.

I tested, it works.

FrancescoBorzi:
good job @Viste and thanks @Helias for testing this

PR #2678, opened 2020-02-18
Core/Misc: Remove dependency on undefined behaviour

Viste:
## CHANGES PROPOSED:
-  https://stackoverflow.com/a/4328396

## Target branch(es):
- [x] Master

### DIFF
--- a/src/common/Collision/BoundingIntervalHierarchy.h
+++ b/src/common/Collision/BoundingIntervalHierarchy.h
@@ -18,29 +18,26 @@
 #include <algorithm>
 #include <limits>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


Yehonal:
The code is OK. But, What does it improve exactly?

The union method is a common way to handle it and should be equivalent

Viste:
@Yehonal  according to the standard, the union method definitely undefined behavior. Use of memcpy free us form undefined behavior

BarbzYHOOL:
@Yehonal you could write a last word too, not just "approve" (plz)

Feel free to add "to be merged" label too

PR #2702, opened 2020-02-24
chore(Core): restore little comment

Helias:
I removed a comment in the previous PR :sweat_smile:


src/server/game/Handlers/AuctionHouseHandler.cpp
```diff
@@ -651,7 +651,7 @@ void WorldSession::HandleAuctionListOwnerItems(WorldPacket & recvData)
     uint32 listfrom;
 
     recvData >> guid;
-    recvData >> listfrom;
+    recvData >> listfrom;       // not used in fact (this list not have page control in client)
```
BarbzYHOOL:
```suggestion
    recvData >> listfrom;       // not used in fact (this list does not have page control in client)
```

PR #2704, opened 2020-02-24
Custom scripts for whisper must be run after player name validate and…

masterking32:
Custom scripts for whisper must be run after player name validate and the receiver was valid.


Yehonal:
Sorry what is this? why we need it?

I checked the code twice and you're creating a "redundant call", moreover moving that function from the **case CHAT_MSG_WHISPER:** to that line on bottom, doesn't help that case itself and completely changes the hook behaviour

masterking32:
I think still can make the server crash, If you try something like that (#2407) if server have any custom script for chats, Like chat censure or etc ...

PR #2684, opened 2020-02-20
refactor(HandleAuctionListOwnerItems): prevent crash

Helias:
## CHANGES PROPOSED:
-  Refactor the function HandleAuctionListOwnerItems to prevent crash from malformed packet (opcode **CMSG_AUCTION_LIST_OWNER_ITEMS**).
  

## ISSUES ADDRESSED:
- Closes #2683 
- probably related to #2666

## TESTS PERFORMED:
- I tested it as the issue show
- after the refactor I also tested several times the function during the "normal" play, I put on the auction some items and in "Auctions" I can see all my own items in the auction.

[chardump](https://paste.ubuntu.com/p/QdHgD8FgDt/)

## HOW TO TEST THE CHANGES:
- To reproduce the sending of the malformed packet just follow the instructions in issue #2683.
- To verify the normal play in wow just open an auction house in-game, put something in the Auction and see in the "Auction" frame if you can see your own items in the Auction.

Thx @SolidMaxtor who reported to me this issue.

## Target branch(es):
- [x] Master

## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #2710, opened 2020-02-25
[Core/Crash] - Similar like #2684 for HandleAuctionListItems

masterking32:
Similar #2684 and [9e307455a83e40045eeb562d8b3de30c408842b1](https://github.com/azerothcore/azerothcore-wotlk/commit/9e307455a83e40045eeb562d8b3de30c408842b1).

First, We need to check creature GUID.


Yehonal:
Wouldn't be better to do it after and not before the async?

I know that the async function is executed after a small-time, but It's just a common pattern if you need to do actions on the creature later

PR #2722, opened 2020-02-27
style(DB): battleground_template entries now commented by type (arena…

BarbzYHOOL:
… or bg)

## CHANGES PROPOSED:
- Increased comment length from 32 to 38
- Fixed typo
- Added full name for Nagrand Arena

Can be useful for people who don't know WOW by heart
Can be used to select only arenas or only battlegrounds

## TESTS PERFORMED:
- Applied to DB

## HOW TO TEST THE CHANGES:

 no test needed

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1582810463583691504.sql
+++ b/data/sql/updates/pending_db_world/rev_1582810463583691504.sql
@@ -1,0 +1,18 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1582810463583691504');
+
+ALTER TABLE `battleground_template`
+CHANGE `Comment` `Comment` char(38) COLLATE 'utf8_general_ci' NOT NULL AFTER `ScriptName`;
+
+UPDATE `battleground_template` SET `Comment` = 'Alterac Valley (battleground)' WHERE `ID` = '1';
+UPDATE `battleground_template` SET `Comment` = 'Warsong Gulch (battleground)' WHERE `ID` = '2';
+UPDATE `battleground_template` SET `Comment` = 'Arathi Basin (battleground)' WHERE `ID` = '3';
+UPDATE `battleground_template` SET `Comment` = 'Nagrand Arena / Ring of Trials (arena)' WHERE `ID` = '4';
+UPDATE `battleground_template` SET `Comment` = 'Blades\'s Edge Arena (arena)' WHERE `ID` = '5';
+UPDATE `battleground_template` SET `Comment` = 'All Arenas (arena)' WHERE `ID` = '6';
+UPDATE `battleground_template` SET `Comment` = 'Eye of The Storm (battleground)' WHERE `ID` = '7';
+UPDATE `battleground_template` SET `Comment` = 'Ruins of Lordaeron (arena)' WHERE `ID` = '8';
+UPDATE `battleground_template` SET `Comment` = 'Strand of the Ancients (battleground)' WHERE `ID` = '9';
+UPDATE `battleground_template` SET `Comment` = 'Dalaran Sewers (arena)' WHERE `ID` = '10';
+UPDATE `battleground_template` SET `Comment` = 'The Ring of Valor (arena)' WHERE `ID` = '11';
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/rev_1582810463583691504.sql
```diff
@@ -0,0 +1,18 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1582810463583691504');
+
+ALTER TABLE `battleground_template`
+CHANGE `Comment` `Comment` char(38) COLLATE 'utf8_general_ci' NOT NULL AFTER `ScriptName`;
+
+UPDATE `battleground_template` SET `Comment` = 'Alterac Valley (battleground)' WHERE `ID` = '1';
+UPDATE `battleground_template` SET `Comment` = 'Warsong Gulch (battleground)' WHERE `ID` = '2';
+UPDATE `battleground_template` SET `Comment` = 'Arathi Basin (battleground)' WHERE `ID` = '3';
+UPDATE `battleground_template` SET `Comment` = 'Nagrand Arena / Ring of Trials (arena)' WHERE `ID` = '4';
```
Helias: These data are taken from BattlemasterList.dbc, and ID 4 is "Nagrand Arena".

Helias:
I partially agree with this PR because I think that we should make compliant this table with BattlemasterList.dbc where there is a field `InstanceType` which shows you if a row is Battleground or Arena.

For now, we can just merge this but in the future, we should be closed to the DBC content files imho.

data/sql/updates/pending_db_world/rev_1582810463583691504.sql
BarbzYHOOL: https://wow.gamepedia.com/Ring_of_Trials it's just a comment meant to inform the player, when i researched it i couldn't find "nagrand arena"

BarbzYHOOL:
> I partially agree with this PR because I think that we should make compliant this table with BattlemasterList.dbc where there is a field `InstanceType` which shows you if a row is Battleground or Arena.
> 
> For now, we can just merge this but in the future, we should be closed to the DBC content files imho.

you mean there should be an extra row for InstanceType??

Helias:
Yep, because in the DBC there is,
https://wowdev.wiki/DB/BattlemasterList

BarbzYHOOL:
yes but values are 3 and 4, confusing

dunno, if someone wants to voice his opinion

data/sql/updates/pending_db_world/rev_1582810463583691504.sql
BarbzYHOOL: so i feel like my change is better xD

Helias: could be, but I think that in part we should follow the blizz standard

BarbzYHOOL: it's just a comment, and in the DBC it's the column "name" not "comment"

gengarshadowball:
In my opinion both of you are right!
1. We should add a comment that allows better understanding of what the battleground template is
2. We should implement hooks and a new column but instead of an integer a bit and update the wiki and add hooks to get this instance Type where 0 is battleground and 1 is arena.

3 and 4 feels like they had other ideas for numbers 0, 1 and 2. Since there is only 2 types of battlegrounds, a bit is small and easy to understand.

gengarshadowball:
All I cans say is if you decide to merge this, I'll add the instanceType to the core of you guys agree.

Helias:
@gengarshadowball I agree with you, you can take the data from DBC and make a parse of the value in this way (like you said):
- 3 => 0 (Battleground)
- 4 => 1 (Arena)

You can call the field "isArena", at the same time, if you have some more time, you can change the function "BattlegroundMgr::isArena()" to check the DBC data (imported to the database, so check the database) instead of using a custom algorithm.

This is the algorithm which annoys me.
image

BarbzYHOOL:
@gengarshadowball ;)

Issue #2690, opened 2020-02-21
[BRD] Coffers loot seem incorrect


BarbzYHOOL:
##### SMALL DESCRIPTION:

In the black vault, the secret safe and the dark coffer do not seem to hold the correct loot tables.

##### CURRENT BEHAVIOUR:

https://wow.gamepedia.com/Secret_Safe this one had 0 loot
https://classic.wowhead.com/object=160845/dark-coffer#comments this one had two loot but it was a backpack and a green pair of boots, seems like it's missing crafting items

##### STEPS TO REPRODUCE THE PROBLEM:

1. .tele blackrockdepths
2. .add 11000
3. .add 11078 12
1. Open all the coffers in the black vault, kill the boss and loot the secret safe (behind the dark coffer) (or just do it as GM mode maybe it works)
1. .add 11197
1. open the dark coffer

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

8760a9f

Kitzunu:
Suggested fix for Secret Safe Entry = 161495
```sql
DELETE FROM `gameobject_loot_template` WHERE `entry` IN (161495);
INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
(161495, 22205, 25, 0, 1, 1, 1), -- Secret Safe
(161495, 22255, 25, 0, 1, 1, 1),
(161495, 22256, 25, 0, 1, 1, 1),
(161495, 22254, 25, 0, 1, 1, 1),
(161495, 11309, 100, 1, 1, 1, 1);
```
For dark coffer I hope there is an existing reference entry for it, cuz I am way too lazy to add 170+ entries x))

FrancescoBorzi:
@Kitzunu thanks, can you please [open a PR](http://www.azerothcore.org/wiki/How-to-create-a-PR) with it?

Kitzunu:
ah yea, sure

BarbzYHOOL:
> For dark coffer I hope there is an existing reference entry for it, cuz I am way too lazy to add 170+ entries x))

after studying how the loot system works, I don't see a reason for a reference entry for the loot as it's certainly a unique loot template never found anywhere else lol

BarbzYHOOL:
no, in fact it can possibly exist a reference entry unused because it's used for bosses that drop multiple items from a referenced loot table (example: kel thuzad)


[...]

PR #2738, opened 2020-03-06
fix(DB/Gameobject): Secret Safe loot

Kitzunu:
Ref https://github.com/azerothcore/azerothcore-wotlk/issues/2690

Fixed loot for gameobject = 161495

(I don't know the stuff with rev and shiz so feel free to edit it <3)

Tests: None

### DIFF
--- a/data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
+++ b/data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
@@ -1,0 +1,9 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('XXXXXXXXXXXX');
+
+DELETE FROM `gameobject_loot_template` WHERE `entry` IN (161495); -- Secret Safe
+INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
+(161495, 22205, 25, 0, 1, 1, 1),
+(161495, 22255, 25, 0, 1, 1, 1),
+(161495, 22256, 25, 0, 1, 1, 1),
+(161495, 22254, 25, 0, 1, 1, 1),
+(161495, 11309, 100, 1, 1, 1, 1);
--- a/data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
+++ b/data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
@@ -1,9 +1,9 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('XXXXXXXXXXXX');
 
-DELETE FROM `gameobject_loot_template` WHERE `entry` IN (161495); -- Secret Safe
+DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104); -- Secret Safe
 INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
-(161495, 22205, 25, 0, 1, 1, 1),
-(161495, 22255, 25, 0, 1, 1, 1),
-(161495, 22256, 25, 0, 1, 1, 1),
-(161495, 22254, 25, 0, 1, 1, 1),
-(161495, 11309, 100, 1, 1, 1, 1);
+(11104, 22205, 25, 0, 1, 1, 1),
+(11104, 22255, 25, 0, 1, 1, 1),
+(11104, 22256, 25, 0, 1, 1, 1),
+(11104, 22254, 25, 0, 1, 1, 1),
+(11104, 11309, 100, 1, 1, 1, 1);
--- a/data/sql/updates/pending_db_world/rev_1583690478008671900.sql
+++ b/data/sql/updates/pending_db_world/rev_1583690478008671900.sql
@@ -1,9 +1,13 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1583690478008671900');
 
-DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104); -- Secret Safe
+-- Set logic ID number
+UPDATE `gameobject_template` SET `Data1` = '161495' WHERE `entry` = '161495';
+
+-- Delete old entry and add new one with correct ID
+DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104,161495); -- Secret Safe
 INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
-(11104, 22205, 25, 0, 1, 1, 1),
-(11104, 22255, 25, 0, 1, 1, 1),
-(11104, 22256, 25, 0, 1, 1, 1),
-(11104, 22254, 25, 0, 1, 1, 1),
-(11104, 11309, 100, 1, 0, 1, 1);
+(161495, 22205, 0, 0, 1, 1, 1),
+(161495, 22255, 0, 0, 1, 1, 1),
+(161495, 22256, 0, 0, 1, 1, 1),
+(161495, 22254, 0, 0, 1, 1, 1),
+(161495, 11309, 100, 1, 0, 1, 1);
--- a/data/sql/updates/pending_db_world/rev_1583690478008671900.sql
+++ b/data/sql/updates/pending_db_world/rev_1583690478008671900.sql
@@ -1,4 +1,4 @@
-INSERT INTO `version_db_world` (`sql_rev`) VALUES ('XXXXXXXXXXXX');
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1583690478008671900');
 
 DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104); -- Secret Safe
 INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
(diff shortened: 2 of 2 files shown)


data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('XXXXXXXXXXXX');
```
Kitzunu: I don't understand what/if I should edit the XXXXXXXXXXX part ^^ 

(Same with the name of the file)

Kitzunu:
I will fix the issue, it is me being an idiot and forgetting how gameobjects work lol

data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
FrancescoBorzi: @Kitzunu you are not supposed neither edit that line nor to have XXXXX as a value. I guess you skipped a step:

image

I think you forgot to execute the file `./data/sql/updates/pending_db_world/create_sql.sh` which will generate the sql file for you, containing the right value

Kitzunu: Ah yes

data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
```diff
@@ -0,0 +1,9 @@
...
+
+DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104); -- Secret Safe
+INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
+(11104, 22205, 25, 0, 1, 1, 1),
+(11104, 22255, 25, 0, 1, 1, 1),
+(11104, 22256, 25, 0, 1, 1, 1),
+(11104, 22254, 25, 0, 1, 1, 1),
+(11104, 11309, 100, 1, 1, 1, 1);
```
BarbzYHOOL: all the items are in groupid = 1, so it means only one item will be chosen

and since you put chance 100 to item 11309, it will always be this item, so there is soemthing wrong too here (otherwise thank you already)

Kitzunu: Well, only 1 should be selected, and the one that has 100 also has QuestRequired = 1 so it should not interfere with the other's

Kitzunu: The only case where there should be two drops from what I understand is if the player has the quest required for item 11309

BarbzYHOOL: my bad loool, sorry I had not verified the loot from wowpedia/wowhead before talking

Kitzunu: :P

data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
BarbzYHOOL: this is really easy, just run the create_sql.sh with bash (bash here or double click it) and copy the number and replace XXXXXXXXX by that number

BarbzYHOOL:
Interesting how the loot table is different on classic and "BFA" https://classic.wowhead.com/object=161495/secret-safe

data/sql/updates/pending_db_world/rev_XXXXXXXXXXXX.sql
BarbzYHOOL: Ok so I checked the loot from wowhead, the chances are a bit different :D 26, 24, 24, 23
But I guess it doesn't matter

HOWEVER i thought about this thing again, and the quest item must be in groupid=0 and not in groupid = 1 else it means when you have the quest, it will not drop any blue item, only the quest item

In top of that, I verified how it is for another boss and it's done this way (blue items in groupd id 1 and quest item in groupd id 0)

Kitzunu: WoW Head sucks when it comes to chances because their "addon" that collects data some users use on private servers which make the chances very off in some cases.

BarbzYHOOL: ok and we can see that it's probably intended to be 25 25 25 25 (and it makes sense)

BUT THEN, you could as well put chance = 0, which means equal between all (100/4 = 25 :P)

BarbzYHOOL:
Ok tested this, works fine, hard to retest multiple times as I need to reopen all doors and kill the boss each time (such a pain)

Funny note: I killed the golem guards and got 1 blue item, then I opened the chest and got the EXACT same item (it's an instance shared blue drop apparently)

FrancescoBorzi:
good job @Kitzunu  and thanks @BarbzYHOOL for testing!

BarbzYHOOL:
hmm there is a last trivial issue, we should UPDATE and not DELETE INSERT since the rows are already there, but i lost the original DB values, dunno how to get them back :'(

Kitzunu:
Personally I prefer DELETE / INSERT but it is up to how your guys' standards are (Haven't really dug through them since this is how I always use to do stuff)

BarbzYHOOL:
it's how we have to do it (written in the wiki), and it's the proper way to proceed, i just don't have the original values anymore :'(

FrancescoBorzi:
while UPDATE is usually preferred over DELETE/INSERT, sometimes it is easier to just use DELETE/INSERT and can be acceptable when dealing with tables that hold multiple-rows per entity (for example loots, smartAI, etc..) so I wouldn't be strict on this.

BarbzYHOOL:
yes it's acceptable but it's just extra code for no reason :D

FrancescoBorzi:
if it's not too many additional lines of code, we don't care since it will be squashed anyway each time we release a new version of the DB

if, like I said, you do:

> you should only try to DELETE and INSERT the rows that you are actually changing.

basically you have only 1 extra line for the DELETE, compared to UPDATE (unless you can combine more UPDATE(s) together, but that's too peaky :D)

BarbzYHOOL:
stop contradicting me

to be merged xD

BarbzYHOOL:
i added "ready to be reviewed" by accident lol just saying

BarbzYHOOL:
```sql
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1583690478008671900');

-- Set logic ID number
UPDATE `gameobject_template` SET `Data1` = '161495' WHERE `entry` = '161495';

-- Delete old entry and add new one with correct ID
DELETE FROM `gameobject_loot_template` WHERE `entry` IN (11104,161495); -- Secret Safe
INSERT INTO `gameobject_loot_template` (`entry`, `item`, `chance`, `questrequired`,  `groupid`, `mincount`, `maxcount`) VALUES
(161495, 22205, 0, 0, 1, 1, 1),
(161495, 22255, 0, 0, 1, 1, 1),
(161495, 22256, 0, 0, 1, 1, 1),
(161495, 22254, 0, 0, 1, 1, 1),
(161495, 11309, 100, 1, 0, 1, 1);
```

Ok I improved it slightly so the chances are calculated equally (according to wiki).
I also changed the ID to be the same as the gameobject_template

what do you think? seems better, cleaner

FrancescoBorzi:
@BarbzYHOOL yes, using the group is better as 4 items with 25% is probably meant to be "always one of them" (even though we can't be 100% sure unless we test in retail many times).

About using the same ID, I'm not sure: have you checked that the former ID wasn't used in other gamobjects? The purpose of allowing the loot-ID to be different than the ID itself was meant to reuse stuff

BarbzYHOOL:
> @BarbzYHOOL yes, using the group is better as 4 items with 25% is probably meant to be "always one of them" (even though we can't be 100% sure unless we test in retail many times).
> 
> About using the same ID, I'm not sure: have you checked that the former ID wasn't used in other gamobjects? The purpose of allowing the loot-ID to be different than the ID itself was meant to reuse stuff

25*4 =100 so it's the same thing as kitzunu did, i just use the "correct way" of doing it in my opinion, in the extremely RARE case someone wants to add a 5th item, it will automatically calculate the correct % xD

I have verified the ID is not used and this loot table is not reusable, it's 100% unique to this game object (especially the quest item Heart of the Mountain)

FrancescoBorzi:
> 25*4 =100 so it's the same thing as kitzunu did, i just use the "correct way" of doing it in my opinion, in the extremely RARE case someone wants to add a 5th item, it will automatically calculate the correct % xD

you didn't get my answer, sorry if I wasn't clear.

I meant to say that typically when we see things like 4 items with 25% or 3 items with 33.3%....  we actually need groups as only 1 item will be dropped from that group and they all have equal chance. So your improvement was indeed **correct**.

However, I'd like to underline that having 4 items in 1 group with equal chance, it is technically different than having 4 non-grouped items having 25%. As in the latter case, you can even have 0 or all the 4 of them dropped at once. While by using groups, you areALWAYS dropping ONE item from the group.

> I have verified the ID is not used and this loot table is not reusable, it's 100% unique to this game object (especially the quest item Heart of the Mountain)

then it's good to go, feel free to edit the PR

BarbzYHOOL:
>However, I'd like to underline that having 4 items in 1 group with equal chance, it is technically different than having 4 non-grouped items having 25%. As in the latter case, you can even have 0 or all the 4 of them dropped at once. While by using groups, you are ALWAYS dropping ONE item from the group.

we know it :P (you taught it to me in discord then i taught it to kitzunu)

> you didn't get my answer, sorry if I wasn't clear.
> I meant to say that typically when we see things like 4 items with 25% or 3 items with 33.3%.... we actually need groups as only 1 item will be dropped from that group and they all have equal chance. 

yeah yeah i had understood, just wanted to make sure you got it right because you said "didn't touch loot template for 10 years" lol

BarbzYHOOL:
https://github.com/azerothcore/azerothcore-wotlk/pull/2738/commits/386c17704e32fb0a3e46867305dc49a55f106d97

BarbzYHOOL:
@Kitzunu thank you!!

PR #2751, opened 2020-03-12
fix:(DB/Item) Mechanical Greench / Wand of Holiday Cheer

Kitzunu:
How it works now:
Currently they do not get their duration decreased if the player is offline.

How it should work: 
It should decrease even if the player is offline.
Proof:
https://classic.wowhead.com/item=21325/
https://classic.wowhead.com/item=21328/ 

Tests:
Ran SQL

How to test:
1. .add 21328
2. .add 21325
3. log out for a day


pklloveyou:
Tested. Time goes down

PR #2752, opened 2020-03-12
Fix:(DB/Quest) Shadow's Edge quest_offer_reward

Kitzunu:
Issue:
Currently the quest_offer_reward is like this
```
Listen well, $C. In accepting this weapon your fate is sealed.$B$BOvercome or succumb.$B$BI have placed your feet upon this path. You are therefore my personal responsibility. Should you falter, I am duty-bound to deliver you from this life.$B$BRemember my words, $C, and do not fail.
```

Which is wrong according to WoWHead where it should be 

```
Listen well, <class>. In accepting this weapon your fate is sealed.

Overcome or succumb.

I have placed your feet upon this path. You are therefore my personal responsibility. Should you failter, I am duty-bound to deliver you from this life.

Remember my words, <race>, and do not fail.
```

Tests:
Works

How to test:
1. .q add 24743
2. .go c 201042


Si1ker:
Hello @Kitzunu 

As always, I test before and after the changes (as it should be)

I haven't included the SQL changes and this is what I got:
2020-03-22 23_17_08-Window


- So what's wrong? It's in the same way as Wowhead describes.
Source: https://www.wowhead.com/quest=24743/shadows-edge

Kitzunu:
Basically it has $C on two places whoch is wrong.
A) it should not be capital so $c
B) Second one should refer to Race not class so $r

Si1ker:
Oh I didn't notice that. 😖 
I have included the SQL changes 🐱 👍 
Quest offer reward text showed as expected in description, good to go 🚀

PR #2757, opened 2020-03-14
fix(DB/Spell): Clam drops

Kitzunu:
Issue: @BarbzYHOOL Found out that the drops of all clams are wrong based on https://wowwiki.fandom.com/wiki/Clam?oldid=2076011 (2009 data)

Issues fixed:
Fixed drop rate and drops for 

- Small Barnacled Clam (Entry = 5523)

- Thick-shelled Clam (Entry = 5524)

- Soft-shelled Clam (Entry = 15874)

- Big-mouth Clam (Entry = 7973)

- Jaggal Clam (Entry = 24476)

Tests:
As far as I can see it works


FrancescoBorzi:
@Kitzunu thanks for your contribution. Unfortunately your SQL code is causing new errors:

```
Table 'spell_loot_template' entry 58168 group 1 has total chance > 100% (107.500000)
Table 'spell_loot_template' entry 58165 group 1 has total chance > 100% (108.500000)
Table 'spell_loot_template' entry 58160 group 1 has total chance > 100% (121.000000)
Table 'spell_loot_template' entry 58172 group 1 has total chance > 100% (105.000000)
```
that is why the Travis build is failing:

https://travis-ci.org/github/azerothcore/azerothcore-wotlk/jobs/662272663?utm_medium=notification&utm_source=github_status

Kitzunu:
Oh yea that is right. I will fix this ASAP

BarbzYHOOL:
I was thinking that all groupid should be 0 because you have a chance to get all items at once

then I realized that a clam should have only 1 pearl, else it's a bowling stadium

Si1ker:
SQL changes applied, no errors found. 👍
All good here, values are applied correctly in-game 🚀

BarbzYHOOL:
maybe we should create referenced template for shared "pearls" loots
example clam 1 and 2 loots the same pearl
so refer to the same line in reference_loot_template

Kitzunu:
It has no advantage in doing it that way.

If I guess I think reference template is if you have a lot of entries of items that a lot of creatures/gobs/spells should use so you don't "overpopulate" a normal loot_template.

imo it doesn't matter, only creates more job 🤷‍♂

BarbzYHOOL:
good development = no duplicate data but well i think in that case only 1 line would be deduplicated

PR #2762, opened 2020-03-15
fix(DB/Gossip): Corrupted Cat gossip

Kitzunu:
@BarbzYHOOL found this bug, where the Corrupted Cat did not have a gossip.

This fixes the gossip thing 🗡 

How to reproduce:
1. `.qu add 4506`
1. `.go cr 40765`
1. Summon the tiger in the green well


BarbzYHOOL:
Tested, now he says "grr..." and it looks professional!

Only "issue" is that the cat immediately transforms into tiger, but in vanilla video the cat goes to the well and then transforms into tiger

I didn't find how to delay the transformation either, but it's so minor that unless someone wants to research SAI or has an immediate solution, it's not worth it

BarbzYHOOL:
Here is the kitten for SAI https://db.frostgram.eu/?npc=9937

BarbzYHOOL:
ty @Kitzunu

PR #2771, opened 2020-03-16
fix(DB/Quest): Old Whitebark's Pendant

Kitzunu:
Currently, in the quest_offer_reward it is missing some spaces and is wrongly written compared to https://classic.wowhead.com/quest=8474/old-whitebarks-pendant

How to test:
1. .q add 8474
2. .q comp 8474
3. .go c id 15398
4. Hand in the quest

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1584389363968563900.sql
+++ b/data/sql/updates/pending_db_world/rev_1584389363968563900.sql
@@ -1,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1584389363968563900');
+
+UPDATE `quest_offer_reward` SET `RewardText`= "This... this pendant.  I gave it to Old Whitebark after his people helped us rebuild our village.$B$BI guess this means he's...$B$B<The blood elf clears her throat as she regains her composure.>$B$BI appreciate you bringing this to me, $N.  There is something I'd like to ask of you." WHERE `ID`=8474;
--- a/data/sql/updates/pending_db_world/rev_1584389363968563900.sql
+++ b/data/sql/updates/pending_db_world/rev_1584389363968563900.sql
@@ -1,3 +1,3 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1584389363968563900');
 
-UPDATE `quest_offer_reward` SET `RewardText`= "This... this pendant.  I gave it to Old Whitebark after his people helped us rebuild our village.$B$BI guess this means he's...$B$B<The blood elf clears her throat as she regains her composure.>$B$BI appreciate you bringing this to me, $N.  There is something I'd like to ask of you." WHERE `ID`=8474;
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/rev_1584389363968563900.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1584389363968563900');
+
+UPDATE `quest_offer_reward` SET `RewardText`= "This... this pendant.  I gave it to Old Whitebark after his people helped us rebuild our village.$B$BI guess this means he's...$B$B<The blood elf clears her throat as she regains her composure.>$B$BI appreciate you bringing this to me, $N.  There is something I'd like to ask of you." WHERE `ID`=8474;
```
FrancescoBorzi: please use single quotes to wrap strings (no double quotes). When you have to type a single quote inside a string, use two single quotes:

```
UPDATE example SET test = 'look, it''s the way to go';
```

data/sql/updates/pending_db_world/rev_1584389363968563900.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1584389363968563900');
+
+UPDATE `quest_offer_reward` SET `RewardText`= 'This... this pendant.  I gave it to Old Whitebark after his people helped us rebuild our village.$B$BI guess this means he\'s...$B$B<The blood elf clears her throat as she regains her composure.>$B$BI appreciate you bringing this to me, $N.  There is something I\'d like to ask of you.' WHERE `ID`=8474;
```
FrancescoBorzi: use two single quotes instead of  backslash

`it''s perfect`

`it\'s ok but not perfect`

pklloveyou:
Sorry, I'm in Chinese.

Si1ker:
Hello 🐱 , just some notes regarding how to test these kind of changes so we can keep it for the future.

1- if a quest comes from an item, you have to include in the "how to test" section to add the item and then accept the quest from the item as you cannot directly execute **.q add 8474**
2020-03-22 21_52_20-Window

2- If the only thing you have to do to complete the quest is going to talk with an NPC, it is not required **.q comp 8474** because it is already completed, the only task left is to complete the quest with the NPC.

(without **.q comp 8474**)
2020-03-22 22_00_03-Window



Having said that 😸 , the correct way "How to test:" for this PR would be:
= - = - = - = - = - = - = - = - = - = - = - = - =
.additem 23228
Start the quest from the item
.go c id 15398
Read the dialogue before completing the quest
= - = - = - = - = - = - = - = - = - = - = - = - =

All good here, ready to be merged ! 🚀 @FrancescoBorzi @Kitzunu 
2020-03-22 22_03_34-Window

Kitzunu:
Oh ok, missed ty! Will know for next time ❤️

PR #2781, opened 2020-03-20
refactor(Core): Remove player title defines

IntelligentQuantum:
https://paste.ubuntu.com/p/dNbDGvyHBW/

## Target branch(es):
- [x] Master


Viste:
yes
```
    for (int i = 0; i < KNOWN_TITLES_SIZE; ++i)
        SetUInt64Value(PLAYER__FIELD_KNOWN_TITLES + i, 0);  // 0=disabled
    SetUInt32Value(PLAYER_CHOSEN_TITLE, 0);

    SetUInt32Value(PLAYER_FIELD_KILLS, 0);
    SetUInt32Value(PLAYER_FIELD_LIFETIME_HONORABLE_KILLS, 0);
    SetUInt32Value(PLAYER_FIELD_TODAY_CONTRIBUTION, 0);
    SetUInt32Value(PLAYER_FIELD_YESTERDAY_CONTRIBUTION, 0);```
from player.cpp

BarbzYHOOL:
?? in which practical case ingame @Viste ?

Kitzunu:
any update?

Hacaw:
@IntelligentQuantum I agree with you that those constants aren't used anywhere, but what happens if those constants are used in some modules ? @Helias are you aware of any ?
@Yehonal is it safe to delete this, from my point of view, yes but I haven't check any modules so far, only the usage in the project.

Helias:
I think we could comment instead of just remove them

I prefer to comment them because maybe they can be helpful to port some scripts from TC or from other cores

Yehonal:
I used them on AzerothShard, but they were in a module. We don't need this unused dead code inside the core

Helias:
let's remove them

Si1ker (approved):

BarbzYHOOL:
what about this module, is it broken now? @si1ker @yehonal @helias @hacaw @intelligentquantum
https://github.com/azerothcore/mod-pvp-titles/blob/master/src/mod_pvp_titles.cpp

Yehonal:
Idk, but the author can easily move that list inside the module directly

Yehonal:
After a quick look from my phone, looks like the module already has the list in that file. So probably won't be affected

Issue #2806, opened 2020-03-26
gm rank 1 is enemy with players


V3nomAnc3r:
##### SMALL DESCRIPTION:
gm rank 1 is enemy with players.
create new character horde rank 1, create new character alliance rank 0

##### EXPECTED BLIZZLIKE BEHAVIOUR:
It should not be an enemy

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/commit/19144d8bff39b77f2618904c64bb75b56ae5e4c2

##### OPERATING SYSTEM:

windows 10

##### MODULES:

No

IntelligentQuantum:
Try this https://paste.ubuntu.com/p/syqWpX5r8b/

V3nomAnc3r:
tnx works perfect

FrancescoBorzi:
@IntelligentQuantum @V3nomAnc3r the fix should be on master before closing the issue.

Can you open PR with it?

PR #2807, opened 2020-03-26
fix(Core/Accounts): GM Rank 1 hostile to players

IntelligentQuantum:
Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2806

## Target branch(es):
- [x] Master


Si1ker:
V3nomAnc3r and me already tested in issue #2806 .
Approval and go 🚀  !

@FrancescoBorzi

Si1ker:
@Helias or @BarbzYHOOL can you link this PR to the issue so once the PR gets merged the issue is closed? Ty 🐱

Helias:
@Si1ker, done :rocket:

BarbzYHOOL:
@Helias don't put it in title else it will be in our git history (though it can be added in description)

Instead, write "Closes xxx" in the PR description and it should work fine (even with "Close" typo, should work)

`Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2806`

Helias:
@BarbzYHOOL oh, ok, for me it's the same :sweat_smile:

FrancescoBorzi:
I thought about this a bit, and I realised there could be cases when you want to have a player who is also some kind of moderator but not a GM.

This PR is still good as probably the majority of people expect a `SEC_MODERATOR` to be considered a GM.

However, I think it would be nice to make this configurable (that can be done in a separate PR).

@IntelligentQuantum you are the king of configurations, would you like to take this?

BarbzYHOOL:
It only works if you have gm mode on

IntelligentQuantum:
@FrancescoBorzi Yes sure :smile: .

Kitzunu:
Thanks for the PR!

PR #2809, opened 2020-03-26
fix(Core/Packets): Remove from default case some repetitive movement packets

Nefertumm:
## CHANGES PROPOSED:
-  Commit name :P Remove from default case some repetitive movement packets in order to not get kicked by the minimum limit (100, which is too low for those packets)

## TESTS PERFORMED:
- Compiled and tested. It's safe to merge.

## Target branch(es):
- [x] Master

 

Nefertumm:
@jackpoz Those packets also caused some wrong kicks (reported on a 1k players base server), I don't know in Trinity, but certain actions sent more than the default packet amount allowed here. Just letting you know :)

PR #2814, opened 2020-03-29
fix(DB/Creature): Naxxramas Cultist

Kitzunu:
image

Currently, some are not kneeling which they should be doing.

How to test:
1. .go c 128030


LichKing255:
this is not large model , why you change to large ?

Kitzunu:
It was never changed, it always had isLarge = 1 because it means they will have a maximum visibility range of 250 yards.
And that is intended since creatures in a raid, afaik, should be able to always be seen (within 250yards ofc).

Anyhow, before this change, it also had this. You can also check the other GUID's of other Cultists. They also have it

Kitzunu:
Thank you @Si1ker

PR #2837, opened 2020-04-03
fix(DB/creature): Add waypoint for Morin Cloudstalker

aleigood:
Adding a waypoint for the NPC Morin Cloudstalker

## CHANGES PROPOSED..:
- DATABASE

## ISSUES ADDRESSED:
Morin Cloudstalker is currently missing a waypoint like official data.

## TESTS PERFORMED:
Tested on a Linux server, NPC can walk with the waypoint.

## HOW TO TEST THE CHANGES:
.go creature id 2988 transmits to NPC location and observes its route of movement.

## KNOWN ISSUES AND TODO LIST:
None

## Target branch(es):
- [x] Master

## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #2841, opened 2020-04-03
chore(DB): fix syntax in sql update

FrancescoBorzi:


BarbzYHOOL:
I tried a fix here https://github.com/azerothcore/azerothcore-wotlk/pull/2837

FrancescoBorzi:
Let's just keep that part of the PR that fixed the syntax. In my IDE looks better and it's the official way to escape SQL strings

Kitzunu:
Ty for the PR!

PR #2851, opened 2020-04-05
fix(Core/Chat): Display better error if player not found

BarbzYHOOL:
The character can be offline too, so let's indicate it.

## CHANGES PROPOSED:
-  When sending message, if player is not found it will say "not found", but if player is offline, it should also say it. Here is a small "fix" that uses a better string instead

## TESTS PERFORMED:
- None yet, I can't test anything atm
-

## HOW TO TEST THE CHANGES:

- BEFORE PR: Try to send this command to an online character `.send message one_of_your_character looool`. Then try to send this command to an offline character.
- ATER PR: Same procedure, but also try with other chat command that require a character name (like .send mail, .send money and mayeb others?)

Concerns: 
- I don't know if it will affect many other commands and will return this string (maybe it's btter, maybe not, hence why needs feedback).
- I feel like a lot of the code in that file could be deduplicated -_-'

## Target branch(es):
- [x] Master

 

Si1ker:
@BarbzYHOOL
changes included, win 10 / last rev 44c3bef
2020-04-05 16_54_01-Window

So with this PR, it must be something like Player is offline or doesn't exist?

BarbzYHOOL:
yes, damn i can't test coz i have many pending changes and i'm too busy, but i couldn't let this disappear from my memory so I tried to fix it without testing and it seems like it didn't work -_-'

you sure you got the PR? if so then put WIP label, i'll have to fix it later, can't do much atm without testing

Zoidwaffle:
I just tested this - based on a very recent #7ec204 with this patch applied.

WoWScrnShot_022621_181732

Above is before the patch... So as expected.

After:
WoWScrnShot_022621_182135

Poor tauren tried to send a message to Barbz, asking if this could be merged.

tl;dr I added a warlock tauren as comic relief and the patch doesn't seem working as intended

FrancescoBorzi:
I don't see any difference in the before/after o_O

Zoidwaffle:
I just wanted to have a tauren warlock say "hi/moo" to @BarbzYHOOL since I tested his patch... I couldn't see any difference either - and I did say the patch doesn't seem working as intended!

PR #2858, opened 2020-04-09
fix(Core/Creature): Use proper name for wander distance

Kitzunu:
Co-Authored-By: ratkosrb <ratkosrb@users.noreply.github.com>

Cherry-picked from https://github.com/TrinityCore/TrinityCore/commit/6176ce92ea01fd33b11c2e7f8e9ebb509e80c944 and a bit modified for acore.

## CHANGES PROPOSED:
-  Change spawndist to wander_distance for less confusion since it has nothing to do with any spawn distance. It is for the random movement generator's max wandering distance.

## RELATED ISSUE
- https://github.com/azerothcore/azerothcore-wotlk/issues/1118

## TESTS PERFORMED:
- Builds

## HOW TO TEST THE CHANGES:

- Apply and build

## KNOWN ISSUES AND TODO LIST:

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #2866, opened 2020-04-10
Remove eluna from module dependencies

Rochet2:
## CHANGES PROPOSED:
- Remove Eluna as a dependency for travis build
- Changes in the core are hindered by having to update modules before updating core. Instead each part should be updated on its own after core has updated.

We can maybe have separate build in the travis script [or another branch](https://github.com/azerothcore/azerothcore-wotlk/tree/module_cron) to check for compatibility with modules daily or even on push. If it is in main travis script then should make it allow failure to allow PRs not to require updating modules beforehand.

## Target branch(es):
- [x] Master

 
### DIFF
--- a/apps/ci/ci-install-modules.sh
+++ b/apps/ci/ci-install-modules.sh
@@ -3,8 +3,6 @@
 set -e
 
 echo "install modules"
-git clone --depth=1 --branch=master --recursive https://github.com/azerothcore/mod-eluna-lua-engine.git modules/mod-eluna-lua-engine
-
 git clone --depth=1 --branch=master https://github.com/azerothcore/mod-autobalance.git modules/mod-autobalance
 git clone --depth=1 --branch=master https://github.com/azerothcore/mod-transmog.git modules/mod-transmog
 git clone --depth=1 --branch=master https://github.com/azerothcore/mod-npc-beastmaster.git modules/mod-npc-beastmaster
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
> We can maybe have separate build in the travis script or another branch to check for compatibility with modules daily or even on push.

That's a good idea, but before that is implemented how do we realise that we are breaking compatibility with the Eluna module?

If this PR is merged as it is (without extra checks added elsewhere), the problem would then be: if you break the Eluna module, the AC build will pass. And the Eluna build will not be triggered so you will not realise it immediately. You will realise it once a new Eluna build is triggered only

Rochet2:
@FrancescoBorzi 
If the main travis script does the module build then it can be set to allow fail. It will build on every push, but can fail without blocking merging etc.
We can modify this PR to do that instead. But not sure when I get around to that myself.
Not too familiar with the travis setup we have here nor bash..

If the script is elsewhere (see link in main post for example) then the tests are run daily. So you dont know immediately, but you know within a day at least. Updating the modules will likely not be instant anyways.

Kitzunu:
(The same answer as I gave on Discord, but just so more people can take part of it)

This issue that you describe already exists. Just the other way around.
Currently, the compatibility with AC + Eluna is broken since we had to update Eluna first to get https://github.com/azerothcore/azerothcore-wotlk/pull/2858 to pass.

All that I have been proposing is that AC should be updated first, and then Eluna. So we don't have to update Eluna -> AC.

Rochet2:
Closing since the linked PR above handles this.

PR #2887, opened 2020-04-13
feat(CI): migrate from Travis to GitHub Actions

Rochet2:
## CHANGES PROPOSED:
- Move from travis to github actions
- Allows a lot faster builds (4 minutes at best, 40 minutes at worst, docker takes 10 minutes always sadly)
- Allows more concurrent builds (22 instead of 4)
- Checking module compatibility can be configured not to prevent merge to master for PRs

While I was thinking of using docker, I do not have a docker setup at the moment and decided to build on the existing travis and github action setups.
I believe a similar setup to docker would be using the container like yehonal was using before my commits.
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer
Using container would possibly provide more consistent dependencies. Currently the builds run in the github preconfigured environment.

## ISSUES ADDRESSED:
- This PR uses the following PR as a base: Closes https://github.com/azerothcore/azerothcore-wotlk/pull/2648
- This PR achieves the same as the following PR: Closes https://github.com/azerothcore/azerothcore-wotlk/pull/2866

## TESTS PERFORMED:
- [x] It builds
- [x] It can import pending SQL files on master only
- [x] PR can show failed build of modules and can still be merged even if other builds block the merge

## HOW TO TEST THE CHANGES:
- Review the code
- Review the building of different checks in github actions

- [ ] Merge must be done by admin with settings access since travis check is missing.
- [ ] After merge must let the merge build once and configure in repo settings that the builds without modules are required to pass for PRs on the master branch.

## Target branch(es):
- [x] Master

 

Kitzunu:
Nice @Rochet2 !

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,59 @@
+name: core-build
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
```
FrancescoBorzi, line 9: what would change if we use `fail-fast`? I guess it would be good for a contributor to see if the PR fails asap, but maybe I'm missing something here

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,59 @@
...
+      matrix:
+        compiler: [clang6, clang7]
+        modules: [with, without]
+    runs-on: ubuntu-18.04
+    name: ${{ matrix.compiler }}-${{ matrix.modules }}-modules
+    env:
+      COMPILER: ${{ matrix.compiler }}
+    steps:
+      - uses: actions/checkout@v2
```
FrancescoBorzi, line 18: what does `actions/checkout@v2` mean?

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Checkout modules
+        run: ./apps/ci/ci-install-modules.sh
+        if: matrix.modules == 'with'
```
FrancescoBorzi, line 21: shouldn't the `if` come before `run`, or this order is mandatory?

.github/workflows/core_build.yml
Rochet2: fail-fast actually aborts all the jobs if one fails.
So all would basically show a failure if single one fails.
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

.github/workflows/core_build.yml
Rochet2: Its the standard way to do git clone in github actions.
Travis does that step automatically before your script starts to run.
If you create a new github action with default template it uses that command as the first step to download your repository code: https://i.imgur.com/zTbtGzl.png

.github/workflows/core_build.yml
Rochet2: The order does not matter.
Its basically an object with attributes (name, run, if) that we define here and all values will be available when it is used.
Can change the if to be before if that is clearer in some way.

.github/workflows/core_build.yml
FrancescoBorzi: ok, then let's keep it to `false`

.github/workflows/import_pending_sql.yml
```diff
@@ -0,0 +1,32 @@
...
+          git fetch --unshallow origin ${BRANCH}
+          git add -A .
+          git commit -am "Import pending SQL update file..." -m "Referenced commit(s):${GITHUB_SHA}" || true
+        env:
+          BRANCH: ${{ steps.extract_branch.outputs.branch }}
+      - name: Push changes
+        uses: ad-m/github-push-action@fe38f0a751bf9149f0270cc1fe20bf9156854365
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ steps.extract_branch.outputs.branch }}
```
FrancescoBorzi, line 32: do we have a way to test this? or, I guess, we'll find out once it's on master?

apps/db_assembler/includes/functions.sh
```diff
@@ -107,6 +107,7 @@ function dbasm_createDB() {
     else
 		echo "Creating DB ${dbname} ..."
         dbasm_mysqlExec "$confs" "CREATE DATABASE \`${dbname}\`" ""
+        dbasm_mysqlExec "$confs" "CREATE USER IF NOT EXISTS '${CONF_USER}'@'${MYSQL_HOST}' IDENTIFIED BY '${CONF_PASS}';"
```
FrancescoBorzi, line 110: nice one

.github/workflows/import_pending_sql.yml
Rochet2: I already tested that on my own fork by making a dummy update file that was then imported.
https://github.com/Rochet2/azerothcore-wotlk/runs/580733452?check_suite_focus=true
https://github.com/Rochet2/azerothcore-wotlk/commit/f9a6041b1a576659d542a17c43a7509864d2d839
also it seems github has you token automatically .. so I didnt need to set it anywhere for it to work in this workflow. Unsure if I have already set some token before, but it would look like adding some secret env was not necessary.

apps/db_assembler/includes/functions.sh
Rochet2: Part of yehonal's changes.

.github/workflows/import_pending_sql.yml
FrancescoBorzi: amazing!

Rochet2:
Added few todos in main post.

Yehonal:
@Rochet2 have you seen this PR before? https://github.com/azerothcore/azerothcore-wotlk/pull/2648/files

Issue #1883, opened 2019-05-25
Command .mailbox doesn't work & never worked


BarbzYHOOL:
##### SMALL DESCRIPTION:

Command .mailbox never worked

##### EXPECTED BLIZZLIKE BEHAVIOUR:

It should work when you're a GM on/off GM mode

##### CURRENT BEHAVIOUR:

In console, I get this `Barbz attempt open mailbox in cheating way.` *notice the perfect english*

##### STEPS TO REPRODUCE THE PROBLEM:

1. Get a GM character
2. `.send mail your_character "hello" "butch"`
1. Log out and log back in
3. Put GM mode on (or off, same thing)
1. Type `.mailbox` = doesn't work (and check the console error)
1. Open the mail box manually, like a normal player and you will see your mail.

##### EXTRA NOTES:

Bug always present since the beginning.
I'm on an OLD core though.

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

e8aff9feac0fdcce5591b9c6e5e1eb1f77d57bc7

##### OPERATING SYSTEM:

Ubuntu

Selyria:
Yes I can confirm that you are not going to be completely blocked even if you go to a mailbox afterwards you can no longer send anything.

only after a relogg everything goes again except for command .mailbox

lineagedr:
I guess a fix would be to check the gmlevel of the player:
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/MailHandler.cpp#L23


`if (_player->GetSession()->GetSecurity() < SEC_MODERATOR)
		{
			sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
			return false;
		}`

BarbzYHOOL:
Hmm, maybe @lineagedr however I think to check the GM level there are other methods if I recall.

Like this `if (m_player->IsGameMaster())`

lineagedr:
> Hmm, maybe @lineagedr however I think to check the GM level there are other methods if I recall.
> 
> Like this `if (m_player->IsGameMaster())`
> 
> ```c++
> if (guid == _player->GetGUID() && ! m_player->IsGameMaster())
> {
>   sLog->outError("%s attempts to open the mailbox by cheating.", _player->GetName().c_str());
>   return false;
> }
> ```
> 
> maybe this?

You've said that
> It should work when you're a GM on/off GM mode

IsGameMaster checks for .gm on

[...]

PR #2889, opened 2020-04-14
fix(Core/MailHandler): Security check

Kitzunu:
## CHANGES PROPOSED:
-  Add a security check for Moderator

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1883

## TESTS PERFORMED:
- Builds
- Tested ingame

## HOW TO TEST THE CHANGES:

1. do .mail before and after PR

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Handlers/MailHandler.cpp
+++ b/src/server/game/Handlers/MailHandler.cpp
@@ -24,7 +24,7 @@ bool WorldSession::CanOpenMailBox(uint64 guid)
 {
     if (guid == _player->GetGUID())
     {
-        if (!SEC_MODERATOR)
+        if (_player->GetSession()->GetSecurity() < SEC_MODERATOR)
         {
             sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
             return false;
--- a/src/server/game/Handlers/MailHandler.cpp
+++ b/src/server/game/Handlers/MailHandler.cpp
@@ -24,8 +24,11 @@ bool WorldSession::CanOpenMailBox(uint64 guid)
 {
     if (guid == _player->GetGUID())
     {
-        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
-        return false;
+        if (!SEC_MODERATOR)
+        {
+            sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
+            return false;
+        }
     }
     else if (IS_GAMEOBJECT_GUID(guid))
     {


Helias:
**SEC_MODERATOR** include admins/gm and console, right?

Kitzunu:
Yes i tested

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -22,8 +22,11 @@ bool WorldSession::CanOpenMailBox(uint64 guid)
 {
     if (guid == _player->GetGUID())
     {
-        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
-        return false;
+        if (!SEC_MODERATOR)
```
lineagedr, line 25: if (!SEC_MODERATOR) ??? wut

This should be:
```
if (_player->GetSession()->GetSecurity() < SEC_MODERATOR)
```

Kitzunu: Does it matter? O.o
Works with !SEC_MODERATOR

Kitzunu: image
With Account Security 1,2,3,4 (MOD,GM.ADMIN,CONSOLE)
With Account Security 0 regular mailbox works as intended and nothing is broken

lineagedr: It matters. Right now your if statement looks like this:
```
if (!1)
```
It will always return true; 
meaning anyone regardless of their gmlevel can open the mailbox.

Kitzunu: I think SEC_MODERATOR includes the above too, because without this change, .mailbox don't work for anything, and now it works for all

Viste: @lineagedr is right

Kitzunu: Pls explain why so I can learn. Because every test I make this works

Helias:
Wtf, I didn't think about that
I was thinking that SEC_MODERATOR was a variable with assigned the role of the player.

Actually you are doing if(!1) as viste and lineagedr said.

because SEC_MODERATOR is a constant value

Kitzunu:
yea Line explained in Discord to me how it works, so I understand it now

Helias:
really thanks @lineagedr  and @Viste

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -22,8 +22,11 @@ bool WorldSession::CanOpenMailBox(uint64 guid)
...
     if (guid == _player->GetGUID())
     {
-        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
-        return false;
+        if (_player->GetSession()->GetSecurity() < SEC_MODERATOR)
+        {
+            sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
+            return false;
+        }
     }
```
Helias, line 30: You can avoid the second if merging it with the first, like in this way:
```C++
    if (guid == _player->GetGUID() && _player->GetSession()->GetSecurity() < SEC_MODERATOR) {
        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
        return false;
    }
```

FrancescoBorzi: also the `return` makes the `else` not needed

Kitzunu: so..:
```cpp
    if (guid == _player->GetGUID() && _player->GetSession()->GetSecurity() < SEC_MODERATOR) {
        sLog->outError("%s attempt open mailbox in cheating way.", _player->GetName().c_str());
        return false;
    }
    
    if (IS_GAMEOBJECT_GUID(guid)
```
?

Viste: yes but { on next line after if :D

Kitzunu: Okay, I will see if I can do this tomorrow :)

Helias:
missclick, just waiting your last edits

src/server/game/Handlers/MailHandler.cpp
Kitzunu: I don't want to do this, I tested it locally and this change breaks not only the command but the entire mailbox system

PR #2910, opened 2020-04-19
fix(core/packet): prevent client crash from malformed packet/message

Helias:
## CHANGES PROPOSED:
-  added a check in the ChatHandler to prevent the crash of the client 

**Co-author: @Quel-Dalar**

## TESTS PERFORMED:
- I tested the attack and the related fix on Windows and LInux.
**BEFORE PR:** the client of the victim target crashes.
**AFTER PR**: the server filter the message so the client does not crash.

## HOW TO TEST THE CHANGES:
- download [this](http://usefulness.altervista.org/azerothcore/WoW-Crash-Player.zip) 
- add ADDON "AccLuaCrashPlayer" in you addons wow client
- Replace wow.exe with wow.exe in ImmuneWowExe/
- Move all files in DLL and injector/ to WoW root files
- Open your wow client in character select window
- Run extremeinjectorv3.7.2 and press inject (CrashApi.dll is charge in DLL list)
- login on world
- use the followings macro
```
/ct
/cn PLAYER_NAME // this to make crash only one player
/run TimerFind:SetChecked(1)  // this for start MASS kick players
/run TimerFind:SetChecked(0) // this for stop MASS kick players
```
replace PLAYER_NAME with the name of a victim

`/ct` does not work properly, ignore it and just use /cn PLAYER_NAME

## Target branch(es):
- [x] Master


Helias:
I am sorry for this all commits : /, but probably I missed something when I updated my fork xD
Anyway the file diff it's just 2-3 lines of code, so easy to review and that's it.

merdlin:
I am being disconnected from the server when I test.

Helias:
@merdlin but not when you import the PR, right?

merdlin:
> @merdlin but not when you import the PR, right?

Yes correct.

P-Kito:
grafik

Something like this already exists, maybe just modify?

Helias:
@P-Kito, it's the same, I could move my check inside that big if, but it's the same, anyway, I added it after that to make easy to see and compare

src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -329,6 +329,11 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket & recvData)
         }
     }
 
+    // prevent crash player
+    if (msg.find("| |Hquest") != std::string::npos) {
```
Nefertumm, line 333: Mm you don't even need the brackets :)

Quel-Dalar:
@merdlin for test you need Replace wow.exe with wow.exe in ImmuneWowExe/ for evade auto kick.

src/server/game/Handlers/ChatHandler.cpp
Helias: I know, but I really prefer this code-style :P

Helias:
I would have several tests for this little PR and at the same time I hope to merge it today.

disclosurez:
PR seems fine, but could not replicate ingame maybe i do soemthing wrong?

Nefertumm:
Tested, the fix is working.

PR #2933, opened 2020-04-23
fix(Core/Pet): Fix armor column

Chromie-WoW:
## CHANGES PROPOSED:
-  Changing armor from colum 9 to column 4
-  

## ISSUES ADDRESSED:
- Query has the column values out of order. Currently has armor as column 9 which is spirit, when armor is actually column 4

Line 3121 is getting armor at field[9] which is actually spirit, changed it to 4 for armor column

## TESTS PERFORMED:
-
-

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 

lineagedr:
Why should it be at 4? It won't make any difference. Also take a look at:https://github.com/azerothcore/azerothcore-wotlk/blob/3e1fa25ad1e2b5221f837eb395b25f96e2d7663e/src/server/game/Globals/ObjectMgr.cpp#L3126

After your changes it will set pet strength equal to the armor value and agility to strength value and so on and so fourth. Consider changing it to:

```pLevelInfo->stats[i] = fields[i+5].GetUInt16(); ```

**Comment by Chromie-WoW** [comment]
Can you please explain, i am new to this and trying to start helping. I noticed in the query to pull from per_levelstats, it has armor back at column 9 which it is actually column 4, so the way it is now, the query is actually pulling spirit. So yes, moving armor up to 4 is going to throw them all off, but it seems like that needs to be proper for pet stat pulling. Then with your comment change to fix the order even more, is something i missed?

**Comment by BarbzYHOOL** [comment]
could also directly write the column names instead??

**Comment by P-Kito** [comment]
This is just "cosmetic", right?

**Comment by lineagedr** [comment]
@Chromie-WoW The order in the table is irrelevant. What's imortant is the order in the result query. 
imimg

**Comment by Chromie-WoW** [comment]
> @Chromie-WoW The order in the table is irrelevant. What's imortant is the order in the result query.
> imimg

Ok that makes sense, then i guess this is just cosmetic. I am trying to find an error in lock pets having a ton of extra hp, lvl 1 imp pet having almost 300hp, this does not seem to be it.

**Comment by lineagedr** [comment]
@Chromie-WoW It is cosmetic but after you set the 4th column to armor that pushed the strength column position to 5 and the loop at https://github.com/azerothcore/azerothcore-wotlk/blob/3e1fa25ad1e2b5221f837eb395b25f96e2d7663e/src/server/game/Globals/ObjectMgr.cpp#L3126 
```

QueryResult result = WorldDatabase.Query("SELECT creature_entry, level, hp, mana, armor, str, agi, sta, inte, spi, min_dmg, max_dmg FROM pet_levelstats");

creature_entry = 0
level = 1
hp = 2
mana = 3
armor = 4
str = 5
agi = 6
sta = 7
inte = 8
spi = 9
min_dmg = 10
max_dmg = 11

for (int i = 0; i < MAX_STATS; i++) // Loop 0 to 5
{
   pLevelInfo->stats[i] = fields[i+4].GetUInt16(); // fields[i+4] i is 0. 0+4 = 4.
}
pLevelInfo->stats[i]  // "i" is 0 at start of the loop. 0 represents Strength see below

enum Stats
{
    STAT_STRENGTH                      = 0,
    STAT_AGILITY                       = 1,
    STAT_STAMINA                       = 2,
    STAT_INTELLECT                     = 3,
    STAT_SPIRIT                        = 4
};

So: pLevelInfo->stats[STAT_STRENGTH] = fields[i+4].GetUInt16(); // Field 4 is now armor
```
You can see how problems can arise.

**Comment by Chromie-WoW** [comment]
Def helps
thank you,
i was looking at 
ObjectMgr.h
line 512 
struct PetLevelInfo
{
    PetLevelInfo() : health(0), mana(0), armor(0), min_dmg(0), max_dmg(0) { for (uint8 i=0; i < MAX_STATS; ++i) stats[i] = 0; }

    uint16 stats[MAX_STATS];
    uint16 health;
    uint16 mana;
    uint32 armor;
    uint16 min_dmg;
    uint16 max_dmg;
};

File: src/server/game/Globals/ObjectMgr.cpp [hunk 1]
```diff
@@ -3118,7 +3118,7 @@ void ObjectMgr::LoadPetLevelInfo()
 
         pLevelInfo->health = fields[2].GetUInt16();
         pLevelInfo->mana   = fields[3].GetUInt16();
-        pLevelInfo->armor  = fields[9].GetUInt32();
+        pLevelInfo->armor  = fields[4].GetUInt32();
```
FrancescoBorzi, line 3121: looks like the final result is the same as before

Hacaw:
Can we close this @FrancescoBorzi ?

Helias:
I agree to close this abbandoned PR, it seems a tiny edit.

PR #2989, opened 2020-05-01
[Zone] Desolace

Knindzagxg:
Updated missing AI for some creatures, but haven't checked whole zone yet.

To test, apply SQL.
Reload creature_template ID
.npc add ID


Geekena:
damage and spells are ok
no problems

Kitzunu:
Ty for the PR!

PR #2990, opened 2020-05-01
[Zone] Swamp of Sorrows

Knindzagxg:
Updated missing AI for few NPC's, but haven't check whole zone yet.

To test, apply SQL.
Restart Server
.npc add ID


Geekena:
damage and spells are ok
no problems

Kitzunu:
Ty for the PR!

PR #2992, opened 2020-05-01
fix(DB/Creature): Dragonmaw Bonewarder

Knindzagxg:
2 missing spells are now up for Dragonmaw Bonewarders.

Apply SQL.
.reload creature_template ID
.npc add ID


merdlin:
Hello works, but things are missing in the description.

Knindzagxg:
Yeah sorry mate.
Had to made a lot of them, but people should be able to read SQL file and see what's happens xP

pangolp:
Thanks for the report. Try to test the changes. But I can't find any information about the spell: https://wowhead.com/spell=6205/curse-of-weakness. I can't get any information about spell 707 either. If you can point me to the information where it says this npc must have such spells, I'll thank you. Thank you very much.

Even though it's true that you made a lot of reports, and I really appreciate it, it would be interesting if you provide a little more information, because on top of that we don't have testing, if you don't give them information, no one is going to test it.

pangolp:
Well, it's already approved. Nothing to test then.

Knindzagxg:
Always try with classic DB mate.

Official Wowhead removed a lot of spells / NPC's so it's like they don't exist anymore.
Something like https://classicdb.ch/?npc=1057#abilities for example.

PR #3025, opened 2020-05-11
feat(Core/Loot): Increase the max loot possible by 2

BarbzYHOOL:
Before, it was 16, now let's use the max potential of the client.

## CHANGES PROPOSED:
-  Increase the max loot possible by 2 (from 16 to 18)

The idea came from a guy who researched it on TC https://github.com/TrinityCore/TrinityCore/issues/23948

## ISSUES ADDRESSED:

This is useful on certain bosses that drop a ton of items, some bosses drop 15 items by default, so if you increase the loot amount (with the configuration options) or add 2 new items, it will not show up and you're gonna debug this for days. At least, 2 more items can't hurt and it will only happen for custom servers

## TESTS PERFORMED:
- Tested ingame, works fine!
-

## HOW TO TEST THE CHANGES:

- Take a boss that drops a lot of items, like Azuregos, check his loot_creature_template and increase the amount of dropped item example mincount and maxcount to 10, then check ingame how many items it drops. Before PR, 16 max, after PR 18 max

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 

Kitzunu:
We need to check so this doesn't break anything.
AKA check all *_loot_templates to see if there are entries above 16 that are in different groups or referenced on the same ENTRY. if that is the case, then we cannot allow this :(

BarbzYHOOL:
> We need to check so this doesn't break anything.
> AKA check all *_loot_templates to see if there are entries above 16 that are in different groups or referenced on the same ENTRY. if that is the case, then we cannot allow this :(

Can you give an example of what issue can arise from this?

FrancescoBorzi:
nice one @BarbzYHOOL 

please can anyone do a quick in-game check that the loots still work fine? then it's good to go

BarbzYHOOL:
there is nothing to check, the loots work fine. it's just a number

I verified on azuregos and on a random mob just to please you Shin

PR #3039, opened 2020-05-15
[Zone] Darkshore

Knindzagxg:
Darkshore is now fully updated, since every creature in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele Darkshore and run / test around.


Kitzunu:
2020-05-15 17:01:24 SmartAIMgr: Entry 2205 SourceType 0 Event 1 Action 11 uses min/max params wrong (6500/1000), skipped.

Knindzagxg:
Changed -.-'
Supposed to be 10s not 1s as max, so creature can cast Battle Shout after gaining rage.

Geekena:
damage and spells are ok
no problems found

PR #3046, opened 2020-05-17
[Zone] Stonetalon Mountains

Knindzagxg:
Stonetalon is now fully updated, since every creature in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele StonetalonMountains and run / test around.


merdlin:
Tested, it works.

PR #3063, opened 2020-05-26
fix(Core/ChatHandler): Prevent messages > 255 characters

Helias:
I added a check of length 255 of the "msg" variable because the maximum number of characters which a player can send via the WoW client is "255".
The critical issue here consists to send to the server (using some tools and addons) a whisp of more than 255 characters (a lot more) which makes crash the client.

I am not sure if this can break something in the game or in some custom addons (I don't think so).

## CHANGES PROPOSED:
-  add a check to not send a big message to the client

## ISSUES ADDRESSED:
- different servers are affected by this problem and several clients crash everyday

## TESTS PERFORMED:
**before PR**: the client crash

**after PR**: the client does not crash because the packet is blocked

## HOW TO TEST THE CHANGES:
I can't share how and the related tools to make the crash here publicly, but I can share you privately (contact me on Discord).

In the meanwhile, you can review the code and check the normal game, try to send/receive all kind of messages with 2 WoW opened.

### Credits
I have to thanks different reporters and the users who gave me the tools and the related instructions for this that I can't mention actually.

## Target branch(es):
- [x] Master

 

Helias:
I want to be sure that usually the ADDONS do not use more than 255 characters, because apparently the client using "Say" (Whisp, yells etc.) can't send more than 255 characters, but I am not sure about addons, because the addon that I am using to make crash the client use more than 255 chracters and that's strange, I don't know if it is forcing something (I think so) or an addon "normally" can send more than 255, so this can break the functionalities of some addons.

So please, test a lot the normal game using several addons

Helias:
Thanks all for review, really.

This PR affects the communication **client -> server -> client** (the chat messages), and specifically the addons messages to the server (and to other clients).

I would merge this ASAP, but at the same time I would have 1-2 testers which test the "normal game" using several Addons in-game, hope soon I will have some feedback from testers or from a servers which can get this PR.

Kitzunu:
I have high doubts this would cause any errors since TC uses the same check since 2 years ago.
https://github.com/TrinityCore/TrinityCore/blob/5b7b21c362f8127e98e63e939a419956bc188e20/src/server/game/Handlers/ChatHandler.cpp#L221-L222

Helias:
I had the same idea :grin: , but I didn't find the time to inspect TrinityCore, this is a good proof for this code, thanks a lot @Kitzunu

Helias:
I received some feedback for this PR, also based on the TC code I think that we can merge this.

Kitzunu:
Ty for the PR!

PR #3064, opened 2020-05-26
fix(Core): const at the end of set function.

Baggins800:
Remove `const` at the end of SetGridId(), this `const` implies the member value cannot change, the compiler may drop the const but it is not guaranteed.
Edit: fix typo

### DIFF
--- a/src/server/game/Grids/NGrid.h
+++ b/src/server/game/Grids/NGrid.h
@@ -44,7 +44,7 @@ class NGrid
         }
 
         uint32 GetGridId(void) const { return i_gridId; }
-        void SetGridId(const uint32 id) const { i_gridId = id; }
+        void SetGridId(const uint32 & id) { i_gridId = id; }
         int32 getX() const { return i_x; }
         int32 getY() const { return i_y; }
 


Helias:
Thanks for the contribution.

Is there a way to test that this function still work correctly?

please, explain step by step how to test the functionality of this function (in-game)/

Baggins800:
The simple test I performed was with basic implementation in a `main.cpp`,
```
#include <iostream>
#include <cstdint>
typedef uint32_t uint32;
using namespace std;
class SetMemberTest {
 public:
  void SetGridId(const uint32 & id) { i_gridId = id; }
  //void SetGridId(const uint32 id) const { i_gridId = id; } // this does not compile for me
  void print_member() {
    cout << i_gridId << endl;
  }

 private:
  uint32 i_gridId;
};

int main () {
  SetMemberTest t;
  t.SetGridId(5);
  t.print_member();
  
  return 0;
}
```
followed by, `clang++ main.cpp && ./a.out`, which yielded the result `5`, as expected.
The previous function `void SetGridId(const uint32 id) const { i_gridId = id; }`  causes the program not to compile. You can reproduce the above code an see if you have a different result?

FrancescoBorzi:
@Baggins800 there are conflicts

Baggins800:
> @Baggins800 there are conflicts

it seems the function was removed, the merge is not longer required.

PR #3074, opened 2020-05-26
refactor(issue_template): Move it to Githubs "new" place

Kitzunu:
* Move the issue_template to github's "new" place

* Refactor the template to remove the "description" part to exclude repetition of a bug.


FrancescoBorzi:
@Yehonal

BarbzYHOOL:
seems good, we'll add "suggestion" option too when this goes live

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,3 +1,12 @@
+---
+name: Bug report
+about: Create a report to help us improve.
+title: ''
+labels: ''
+assignees: ''
+
```
BarbzYHOOL, line 7: but i do'nt know what this part is about though

Kitzunu: That is Githubs new stuff, to auto assign or lable or title... Can be left blank

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -45,12 +50,14 @@ Examples:
 - does this feature work in other server appplications (e.g. CMaNGOS, TrinityCore, etc...) ?
 -->
 
+
+
+
 ##### BRANCH(ES):
-<!-- Specify the branch(es) affected by this issue: master, 0.x, 1.x, or another branch. -->
+<!-- Specify the branch(es) affected by this issue: master. -->
```
FrancescoBorzi, line 57: why changing this comment? If we are going to assume that everyone is always using `master` branch, then we can remove completely this section

Kitzunu: Afaik master is the only branch we use, I only keep it in case we'd do something different in the future

.github/ISSUE_TEMPLATE/bug_report.md
BarbzYHOOL: ok cool

.github/ISSUE_TEMPLATE/bug_report.md
BarbzYHOOL: yes, better change it else it's confusing

PR #3082, opened 2020-05-28
fix(Core/BG): do not start new BGs if there are BGs with free slots

FrancescoBorzi:
## CHANGES PROPOSED:

- Prevent new Battlegrounds to start if there are already other Battlegrounds that have free slots  
- Happened mostly when a lot of players queued at the same time (for example when a BG just finished)

## ISSUES ADDRESSED:

- Closes https://github.com/azerothcore/mod-cfbg/issues/11

## TESTS PERFORMED:
- Build, run, etc... (OK)
- Tried to reproduce the issue before the fix (OK)
- Tried to reproduce the issue after the fix (OK)
- Also tried to make a BG start when the current BG running is full (OK)
- Also tried BG with bracket <80 (OK)

## HOW TO TEST THE CHANGES:

For testing purposes, you can:

```
UPDATE battleground_template SET MinPlayersPerTeam = 1
UPDATE battleground_template SET MaxPlayersPerTeam = 2
```

This will make a BG start with 1 player per team, and will consider a BG as "full" when there are 2 players per team inside the BG.

Then in order to test it:

1) Try to start a BG
2) Try to add players to it until it's full
3) Try to make another BG start

Also you can try to reproduce the original issue:

- Try to make multiple players join BG at the same time, only 1 BG should start until it's full

## Target branch(es):
- [x] Master

 

src/server/game/Battlegrounds/BattlegroundQueue.cpp
```diff
@@ -712,6 +712,11 @@ void BattlegroundQueue::BattlegroundQueueUpdate(BattlegroundBracketId bracket_id
                 for (auto itr : m_SelectionPools[TEAM_ALLIANCE + i].SelectedGroups)
                     BattlegroundMgr::InviteGroupToBG(itr, bg, itr->RealTeamID);
         }
+
+        if (!bg_template->isArena() && !bgsToCheck.empty()) {
```
Yehonal, line 716: nope

Yehonal:
Short answer:

it's good for me

Long answere:

it's one of the best 3-lines-of-code PRs ever seen before, super-well tested and documented xD

src/server/game/Battlegrounds/BattlegroundQueue.cpp
Kitzunu, line 716: just for me to understand... What does 
```
!bgsToCheck.empty()
```
do?

FrancescoBorzi: It checks if there are non-empty BGs

Kitzunu: I guess why next question would be, why would there be an empty BG? Or is that the "phase" when nobody as accepted the BG queue?

FrancescoBorzi: @Kitzunu sorry I wasn't clear enough:

I check if there are still BGs that are running AND that are NOT full. Under this condition, there is no reason to go further and start new BGs.

New BGs should be started when there are NO non-full BGs (so either when there are no BGs at all or when all the current BGs are full).

I hope it's clear now. If it's not clear yet, please let me know :)

Kitzunu: Yea, makes sense.

I have not tested the PR yet, but how would this work if for example:

Alliance Team: 4/5
Horde Team: 4/5

In Queue: Two groups of two (2 Ally 2horde both in a party)

Would it start a new game, or just sit endlessly in a queue until the 2 last spots in the first game is filled?

FrancescoBorzi: I haven't tested such a scenario, however this should not be affected by my PR

Yehonal: The bug is inside "CheckNormalMatch". I'll explain you it in PVT

FrancescoBorzi:
@Yehonal  I don't get your point, even if `FillPlayersToBG` should also be changed, we still don't want to go further down and start a new BG if the players have been invited inside it.

Yehonal:
> @Yehonal I don't get your point, even if `FillPlayersToBG` should also be changed, we still don't want to go further down and start a new BG if the players have been invited inside it.

When players are invited the IsInvitedToBGInstanceGUID is set on them to avoid using those player to fill up the m_SelectionPool for new BGs.

So, even though the process continues (as it should be) the invited players won't be used later if we set the check correctly

Yehonal:
After another research, I've found another thing that could cause the "issue" itself:

https://github.com/azerothcore/azerothcore-wotlk/blob/2e597c06650d16c20492a16419272e24e6b9ea3b/src/server/game/Battlegrounds/BattlegroundQueue.cpp#L416

At this line inside the "FillPlayersToBG" you've a check about the team balancing, if you've the crossfaction BG module is quite easy that this check will be always true with the result that only 1 player will be added to the pre-existent bg **selectionPool** while the others (eventhough they were selected before) will be kicked out from that list and will be put inside the new bg **selectionPool**.

This PR allows you to re-try the selection 1 by 1 so the algorithm is able to put them in a balanced way, but it's a workaround to fix the crossbg with the balanced configuration.

You should test instead with the balance configuration off letting the module  to balance the team instead (or fixing the module itself)

Yehonal:
And the CFBG forces the group "kick" too: https://github.com/azerothcore/mod-cfbg/blob/558e2c53532436a4af5470c6be009ea8d8ba86b8/src/CFBG.cpp#L374

Yehonal:
Please can someone test it? https://github.com/azerothcore/mod-cfbg/pull/17

diluh:
Tested along with that other PR #3091  worked very good!

PR #3085, opened 2020-05-28
wiki(Security): Add a security policy

Kitzunu:
Maybe this is a bit of a reach, or it is good to have?

You tell me :D


Winfidonarleyan:
About support `MariaDB`
https://github.com/azerothcore/azerothcore-wotlk/blob/d1e83f6e3d6255802e123845f3377685d9b78695/src/cmake/macros/FindMySQL.cmake#L31

SECURITY.md
```diff
@@ -0,0 +1,34 @@
...
+| ------------------ | ------------------ |
+| master             | :white_check_mark: |
+
+We support these versions of MySQL:
+
+| MySQL Version | Supported          |
+| ------------- | ------------------ |
+| 5.6           | :white_check_mark: |
+| 5.7           | :white_check_mark: |
+| 8.0           | :x:                |
```
FrancescoBorzi, line 17: soon will be supported

Kitzunu: Yep, but so far not :D

Kitzunu:
@Yehonal Thoughts? <3

Yehonal:
Thanks for this contribution :)

Yehonal:
Shouldn't be linked somewhere?

Yehonal:
However, basically we still support the version 2.x (the branch) even though no one is using it practically I guess, but if someone discover a security issue there or something like that, we should fix that security issue on both master and 2.x (cherry pick)

Moreover, is it related to the requirement page?

Kitzunu:
Just https://github.com/azerothcore/azerothcore-wotlk/security

BarbzYHOOL:
how is it related to security though?

it's totally unrelated (imo), the mariadb versions etc...

PR #3095, opened 2020-05-31
[Zone] Dustwallow Marsh

Knindzagxg:
Marsh is now fully updated, since every creature in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele DustwallowMarsh and run / test around.


Geekena:
damage and spells are ok
no problems found

PR #3099, opened 2020-05-31
fix(Core/Vendors): Load npc_vendor items in the right order

BarbzYHOOL:
## CHANGES PROPOSED:
- Before that commit, if you copied a npc_vendor, the order would not be replicated (if all items were on the same `slot`). Now it works fine and it's consistent.  

I found the bug, spent hours on this, then resorted to ask to @Rochet2 and he gave this idea which works fine! Also thanks to Malow for giving me insights

## TESTS PERFORMED:
- Copied a vendor ingame and it worked, before that it was just random bullshit (still no idea why the original vendor worked fine)

## HOW TO TEST THE CHANGES:

- Copy a vendor, check his items, same order

## Target branch(es):
- [x] Master

 

BarbzYHOOL:
No need to test really, but if someone can find why the original vendor worked fine, it would be interesting

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -8316,7 +8316,7 @@ void ObjectMgr::LoadVendors()
 
     std::set<uint32> skip_vendors;
 
-    QueryResult result = WorldDatabase.Query("SELECT entry, item, maxcount, incrtime, ExtendedCost FROM npc_vendor ORDER BY entry, slot ASC");
+    QueryResult result = WorldDatabase.Query("SELECT entry, item, maxcount, incrtime, ExtendedCost FROM npc_vendor ORDER BY entry, slot ASC, item, ExtendedCost");
```
FrancescoBorzi, line 8319: the SQL format is wrong, ` ASC` should be at the end

Helias: lol, fixed

Rochet2: > `ASC` should be at the end

The manual for MySQL 8.0 says https://dev.mysql.com/doc/refman/8.0/en/sorting-rows.html
> The DESC keyword applies only to the column name immediately preceding it
> The default sort order is ascending

So it should be immediately after the column the sorting was intended for and the default sorting is ASC.

The sorting was intended for the `slot` column, like we can see in the original code.
However, currently the ASC does nothing since the sorting is ASC by default anyways - at least on MySQL.

BarbzYHOOL: lol everybody knows it's ASC/DESC per column, i don't get why you said that the borzis and changed it... no sense

but as rochet2 said, the default is ASC, but we put ASC on item to make it clearer to the reader

revert back to the last commit imo

FrancescoBorzi: sorry, I thought that was a mistake and not specifically meant

Helias: restored, sorry xD

BarbzYHOOL: np np

Mitradis:
we need revert this PR because after this all items in all vendors have randomly sorting

Rochet2:
> we need revert this PR because after this all items in all vendors have randomly sorting

Reverting this still results in random sorting, but possibly less often.
Rather than reverting this, the items should have a slot defined or they should be ordered by some additional criteria, such as the item entry, to keep them consistent between restarts in case proper slots are not defined.

If the sorting criteria is changed, then it should probably also be changed here https://github.com/azerothcore/azerothcore-wotlk/blob/d82a8d59f7e852dc609045ba9bfa78b94ebc1737/src/common/Database/Implementation/WorldDatabase.cpp#L28 in addition to the place changed in this PR.

Here is an example SQL for numbering slots in DB for all of the rows with slot = 0. Though if renumbering is done as an update, maybe a bunch of UPDATE statements is a better option to keep all databases ordered in the same way. However, it could be generated from the data produced with this query.
```sql
SET @ROW_NUMBER := 0;
SET @prev_entry := -1;
UPDATE npc_vendor,
(
  SELECT 
    @ROW_NUMBER:=CASE
      WHEN @prev_entry = entry 
    THEN @ROW_NUMBER + 1
      ELSE 0
    END AS num,
    @prev_entry:=entry AS entry2,
    entry AS e, slot AS s, item AS i, extendedcost AS ec
  FROM
    npc_vendor
  WHERE slot = 0
  ORDER BY entry
) subq
SET slot = num
WHERE slot = 0 AND entry = e AND item = i AND extendedcost = ec;
```

PR #3101, opened 2020-05-31
feat(CI): refine cache usage in gh-actions

Rochet2:
## CHANGES PROPOSED:
Common keys block the efficient usage of cache in forks. When one build succeeds, other builds of the workflow or a later commit will match that cache with the common key. The stored cache may be for completely different compiler altogether, so cache is never hit during build.

Without the common key, caches for wrong compilers are not matched and caches are searched from the main repository, which contains recent caches for all of the compilers.

Details:
See https://github.com/azerothcore/azerothcore-wotlk/pull/3099 for an example of this scenario.
In this example the first commit triggers (non docker) builds and all of them finish within 5 minutes: https://github.com/azerothcore/azerothcore-wotlk/runs/725498934
A second commit is created after 15 minutes and some of its builds use over 40 minutes to build even if the changes are minimal.

When looking at the timestamps of each build we can see that the 4 minute build (https://github.com/azerothcore/azerothcore-wotlk/runs/725498934#step:17:2) succeeds slightly after the 40 minute build tries to fetch the cache (https://github.com/azerothcore/azerothcore-wotlk/runs/725503994#step:4:1). However, another build (https://github.com/azerothcore/azerothcore-wotlk/runs/725498938#step:17:3) has already succeeded, so the 40 minute build will use the cache from that one.

## Target branch(es):
- [x] Master

 
### DIFF
--- a/.github/workflows/core_build.yml
+++ b/.github/workflows/core_build.yml
@@ -30,8 +30,6 @@ jobs:
           restore-keys: |
             ccache:${{ matrix.compiler }}:${{ matrix.modules }}-modules:${{ github.ref }}
             ccache:${{ matrix.compiler }}:${{ matrix.modules }}-modules
-            ccache:${{ matrix.compiler }}
-            ccache
       - name: Configure OS
         run: source ./apps/ci/ci-install.sh
         env:


Kitzunu:
Ty for the PR!

PR #3103, opened 2020-06-01
fix(build): eluna build

Helias:
## CHANGES PROPOSED:
Change return type from `char*` into `std::string` of the function  `GetFieldName(uint32 index)` of `ResultSet`

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/mod-eluna-lua-engine/issues/28

## TESTS PERFORMED:
- I build the sources on Linux

## HOW TO TEST THE CHANGES:
- I think it's just necessary to build

## Target branch(es):
- [x] Master

 

SamuelGogarty:
tested this source tree on linux ubuntu 18 LTS.
Builds without issue, eluna module works correctly, am able to login and play normally.

ValeTheVioletMote:
I had a similar issue. I added eluna after the fact of the cmake, and I was using the docker build, so I believe it cached the cmake step that was causing the problem. Solution was to delete the docker/build/cache folder and rebuild.

Rastrian:
Reverse the commit https://github.com/azerothcore/azerothcore-wotlk/pull/3103/commits/bd4143f0a0c9f93af545c56d6d6fca5a7a68138d and the eluna module work fine

Issue #3121, opened 2020-06-12
Deprecated item on player creation needs to be removed


BarbzYHOOL:
##### SMALL DESCRIPTION:

We have an useless grey item for death knight on player creation. This item was only added during test server and readded in cataclysm (i verified on wowhead, wowpedia etc)

Here is the fix:
```sql
DELETE FROM `playercreateinfo_item` WHERE `itemid` = 40582 AND `class` = 6 AND `amount` = -1;
```
The fix is very specific in case someone edited the item itself so it is identified exactly, **do not change it, please.**

*Note: I would post the fix myself if my working environment wasn't so dirty, I have to clean my core/repo before being able to do that, that's why i post it in an issue, sorry*

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Nothing, it's blizzlike to have no hearthstone for DK until cataclysm. (it was temporarily added in beta realms for 3.0 but removed after). So we just need to remove that from our DB

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

19c3a70

lineagedr:
```sql
`amount` = -1;
```

The -1 means it will remove the item not add it. I downloaded a fresh AC today and I was wondering why my DK had the 
"[TDB PH] - unsused Scourgestone" item. This https://github.com/azerothcore/azerothcore-wotlk/pull/3122 commit broke removing items with the playercreateinfo_item table.

FrancescoBorzi:
@lineagedr could you please open a new report with this information? cc @Kitzunu

Issue #3119, opened 2020-06-12
playercreateinfo_item amount is limited to 127 wtf


BarbzYHOOL:
##### SMALL DESCRIPTION:

If we want to put 200 items on player creation (ex: 200 arrows for hunters) we can't.

##### CURRENT BEHAVIOUR:

The datatype is wrong and totally useless (like most tables, proper SQL is not so easy).
We can have -127 to 127 (or 128).
What we need is 0-255 (unsigned tinyint) or more (unsigned smallint).

##### STEPS TO REPRODUCE THE PROBLEM:

1. Put an item in that table with amount = 250. You can't

##### EXTRA NOTES:

**How to fix suggestion:**

SQL
```SQL
ALTER TABLE `playercreateinfo_item` CHANGE `amount` `amount` tinyint(3) unsigned NOT NULL DEFAULT 1 AFTER `itemid`; 
```

CPP, in ObjectMgr.cpp:
```CPP
// change this to the corresponding GetIntX() amount
uint8 amount   = fields[3].GetInt8();
```

*Note: I would post the fix myself if my working environment wasn't so dirty, I have to clean my core/repo before being able to do that, that's why i post it in an issue, sorry*

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

bf74a3a

Kitzunu:
Well, then we should make it smallint? Because some arrows (maybe all?) stack in 1000

BarbzYHOOL:
yes probably, i had the arrows example come to my mind while writing it ^^

PR #3122, opened 2020-06-12
fix(Core/playercreateinfo_item): Amount of allowed item

Kitzunu:
## CHANGES PROPOSED:
* Increase the amount allowed from tinyint signed to smallint unsigned

* Remove deprecated item from playercreateinfo_item

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3121
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3119

## TESTS PERFORMED:
- Builds
- Tested to run Worldserver

## HOW TO TEST THE CHANGES:

1. Build
2. Run Worldserver, see so that you can add items > 127 on creation

## Target branch(es):
- [x] Master

 

Kitzunu:
Does this really need to be tested? :shrug:

FrancescoBorzi:
@Kitzunu if you tested this, then I'd say it's small enough to be accepted.

data/sql/updates/pending_db_world/rev_1591992052509403400.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1591992052509403400');
+
+DELETE FROM `playercreateinfo_item` WHERE `itemid` = 40582 AND `amount` = -1;
+
+ALTER TABLE `playercreateinfo_item`
+	CHANGE `amount` `amount` SMALLINT UNSIGNED NOT NULL DEFAULT 1,
+	ADD `Note` VARCHAR(255) AFTER `amount`;
```
BarbzYHOOL, line 7: what is that `note` column about? @Kitzunu i feel like it's a mistake

Kitzunu: So you dont have to google what the race and item are. So you can add a comment like "Warrior - Shadowedge"

BarbzYHOOL: call it "comment" then? like in other tables

and yes i can see it being useful, it won't take much space :D

BarbzYHOOL:
if the CPP part looks fine to you @Helias @FrancescoBorzi then can be merged

data/sql/updates/pending_db_world/rev_1591992052509403400.sql
Kitzunu: I call it note because other playercreateinfo call it that

BarbzYHOOL: hmm it's true on playercreateinfo_spell and spell_custom even though it's shit

dunno, mixed feelings

Kitzunu: better do it this way and then refactor in the future imo

FrancescoBorzi:
looks good, just a tiny note: changes to the DB **structure** are considered "Core" changes. With "DB" changes, we refer to changes to the **contents** of the DB, without altering its structure.

Just a convention. It has been like this since MaNGOS times ;)

data/sql/updates/pending_db_world/rev_1591992052509403400.sql
BarbzYHOOL: in fact the note is not used by the core so we could just rename to comment in this pr and also for the 2 other tables xD

BarbzYHOOL:
> looks good, just a tiny note: changes to the DB **structure** are considered "Core" changes. With "DB" changes, we refer to changes to the **contents** of the DB, without altering its structure.
> 
> Just a convention. It has been like this since MaNGOS times ;)

good to know after 5 years doing the wrong thing then...

not sure if logic though since the db change is more consequent than the core change

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -3333,7 +3333,7 @@ void ObjectMgr::LoadPlayerInfo()
                     continue;
                 }
 
-                int32 amount   = fields[3].GetInt8();
+                int32 amount   = fields[3].GetUInt16();
```
FrancescoBorzi, line 3336: @Kitzunu  I didn't notice this but why are we storing an unsigned int inside a signed `int32` ?

PR #3123, opened 2020-06-12
Allow db_assembler to connect to non default mysql ports

peti446:
## CHANGES PROPOSED:
-  Add a new parameter in the conf/config.sh `MYSQL_PORT`
-  Add `-P "$MYSQL_PORT"` to all mysql commands in apps/db_assembler/includes/functions.sh
- To maintain backwards compatibility, if `MYSQL_PORT` has not been set it will use the default port

## ISSUES ADDRESSED:
Using the tool to construct or update the databse on a non default port is a bit dificult and might be hard for non experienced users

## TESTS PERFORMED:
 - Using a mysql server on a non default port constructed and updated the database
 - Using a mysql server on the default port constructed and updated the database

## HOW TO TEST THE CHANGES:
 - Setup a mysql with a non default port on Windows/Max/Linux
 - Setup a mysql with a default port on Windows/Max/Linux
 - Change the conf/config.sh to set the port to non default
 - Run db_assembler to construct or update the database
 - To test backwards compatibility, remove the MYSQL_PORT from conf/config.sh
 - Run db_assembler to construct or update the database

## Target branch(es):
- [x] Master

 

FrancescoBorzi:
@Yehonal please have a look at this, I'm not really confident with the AC bash system myself

apps/db_assembler/includes/functions.sh
```diff
@@ -252,19 +256,23 @@ function dbasm_db_backup() {
 		MYSQL_PASS=$PROMPT_PASS
 	fi
 
+	if [[ -z ${MYSQL_PORT+x} ]]; then
+		MYSQL_PORT=3306
+	fi
+
```
BarbzYHOOL, line 262: why twice the same code

apps/db_assembler/includes/functions.sh
```diff
@@ -301,22 +309,26 @@ function dbasm_db_import() {
...
```
BarbzYHOOL, line 315: thrice!

BarbzYHOOL: just make a small function and reuse it

peti446: You are right about the function, done so in the last commit, I also changed it instead of checking if the variable is set it will always set it up to the default port and after it it will eval the configs string and if the port is set overwrite it. This is needed as, if for some reason you char and/or world databse does not have the MYSQL_PORT in the string (they should use default port) and the Auth databse has a non default port it will use the non default port for char and world.

merdlin:
Tested, it works.

apps/db_assembler/includes/functions.sh
```diff
@@ -293,7 +301,7 @@ function dbasm_db_import() {
     fi
 
     echo "importing $1 - $2 ..."
-
+	dbasm_resetMYSQLPort
```
Yehonal, line 304: indentation please

apps/db_assembler/includes/functions.sh
```diff
@@ -7,42 +7,49 @@ function dbasm_resetExitCode() {
 	exit 0
 }
 
+function dbasm_resetMYSQLPort()
+{
+	MYSQL_PORT=3306
```
Yehonal, line 12: Can you explain me why do we need to reset the port everytime? the default value is already inside the config.sh.dist file that is loaded first

peti446: So, this was added for backwards compatibility, where users still using the old config files and only added the 'MYSQL_PORT' to one of the db config, thus resulting in the following one using the wrong port (the custom one instead of the default one).
For example:
```
DB_AUTH_CONF="MYSQL_USER='root'; \
                    MYSQL_PASS=''; \
                    MYSQL_HOST='127.0.0.1';\
					MYSQL_PORT='3336';\
                    "

DB_CHARACTERS_CONF="MYSQL_USER='root'; \
                    MYSQL_PASS=''; \
                    MYSQL_HOST='127.0.0.1';\
                    "

```
In this example without resseting the port, the db will use the port 3336, as it was the last one set, so reseting this would mitigate this sort of situation of user error.

Yehonal: Ok clear. This is an edge case since the DB_*_CONF are evaluated in those methods. However, I'm sorry to say something "against" @BarbzYHOOL but I think for DX (developer experience) is better a one-line native syntax for default values.

Check parameter-substitution documentation: https://tldp.org/LDP/abs/html/parameter-substitution.html

Here you can find a proper example: https://tldp.org/LDP/abs/html/parameter-substitution.html

Something like: ${MYSQL_PORT:=3306}

Yehonal: it's a non common syntax so a comment would be appreciated:

```
# if MYSQL_PORT not set, use default port
${MYSQL_PORT:=3306}
```

peti446: Thanks for the links to documentation!
I have been playing around with them for a while, but they cannot be used to resolve the current edge case, at least in my bash (wls ubuntu) , the value for MYSQL_PORT is carried over, so MYSQL_PORT is initialized to whatever value was used in the last DB operation in my test case to 3336. 
Example Config:
```
DB_AUTH_CONF="MYSQL_USER='root'; \
                    MYSQL_PASS=''; \
                    MYSQL_HOST='127.0.0.1';\
                    "

DB_CHARACTERS_CONF="MYSQL_USER='root'; \
                    MYSQL_PASS=''; \
                    MYSQL_HOST='127.0.0.1';\
					MYSQL_PORT='3336';\
                    "

DB_WORLD_CONF="MYSQL_USER='root'; \
                    MYSQL_PASS=''; \
                    MYSQL_HOST='127.0.0.1';\
                    "
DB_AUTH_NAME="acore_auth1"
DB_CHARACTERS_NAME="acore_characters"
DB_WORLD_NAME="acore_world"
```

Output (The output is done just before the eval):
```
3306
acore_auth1 is not empty, base importing skipped
3306
acore_characters is not empty, base importing skipped
3336
acore_world is not empty, base importing skipped`

```
As you can see acore_world is initialized to 3336 and the ${MYSQL_PORT:=3306} will not get the behavior we want; So the solution is either setting it it 3306 every time.
We could use parameter substitution but it would be 2 calls (${MYSQL_PORT:=3306} and ${MYSQL_PORT+3306}) but it is more clear to use the an asigment imho.

The solution would look like this just before each eval:
```
# MYSQL_PORT needs to be reseted as the next eval might not overwite the current value causing the commands to use wrong port
MYSQL_PORT=3306
```

Yehonal: Looks good for me :) I just don't see the need of a dedicated function for just a single value

peti446:
This is now ready for testing

Helias:
Any updates? Anyone tested? @merdlin did you test it? what you did exactly? what OS did you use to test..?

peti446:
@Helias It will require retest but is still working for me locally.

FrancescoBorzi (approved):

nicoaravena:
Tested, works good, even when the port is wrong it does not continue with the process.

Kitzunu:
Thanks =)

PR #3130, opened 2020-06-15
refactor(Core/Config): Maintain some configs from Database

Kitzunu:
WIP


Kitzunu:
@Yehonal I created a branch on the repo, so if you implement the basis to make it work, then I will just convert *all* of the configs

data/sql/updates/pending_db_world/rev_1592179496642830400.sql
```diff
@@ -0,0 +1,20 @@
...
+('ICC.Buff.Horde', 73822, 'Specify ICC buff\r\n
+It is necessary to restart the server after changing the values!\r\n
+Default: ICC.Buff.Horde = 73822\r\n
+Spell IDs for the auras:\r\n
+73816 -  5% buff Horde\r\n
+73818 - 10% buff Horde\r\n
+73819 - 15% buff Horde\r\n
+73820 - 20% buff Horde\r\n
+73821 - 25% buff Horde\r\n
+73822 - 30% buff Horde');
```
Kitzunu, line 20: Would look something like this I suppose :D

Yes, no? @Yehonal

data/sql/updates/pending_db_world/rev_1592179496642830400.sql
```diff
@@ -0,0 +1,20 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1592179496642830400');
+
+CREATE TABLE IF NOT EXISTS `world_config` (
+  `name` VARCHAR(255) NOT NULL,
+  `value` INT UNSIGNED NOT NULL,
+  `comment` TEXT,
+   PRIMARY KEY (`name`)
+) ENGINE=MyISAM DEFAULT CHARSET=UTF8MB3 COMMENT='Holds configs for the World Server';
```
Kitzunu, line 8:
```suggestion
CREATE TABLE IF NOT EXISTS `world_config` (
  `type` TINYINT UNSIGNED NOT NULL,
  `name` VARCHAR(255) NOT NULL,
  `value` INT UNSIGNED NOT NULL,
  `comment` TEXT,
   PRIMARY KEY (`type`, `name`)
) ENGINE=MyISAM DEFAULT CHARSET=UTF8MB3 COMMENT='Holds configs for the World Server';
```

Actually, what about adding a `type` column too?
That way we could sort and store the configs by type, for example
```
1 - Server Settings
2 - Warden Settings
3 - Player Interactions
100 - Ungrouped
```
So on so forth.
I think that could be a good idea, so that we don't just have a table full of random configs just laying around

data/sql/updates/pending_db_world/rev_1592179496642830400.sql
Yehonal: Yes :) the table is ok

Yehonal:
> @Yehonal I created a branch on the repo, so if you implement the basis to make it work, then I will just convert _all_ of the configs

Unfortunately, I don't have the time to do it right now, I'm busy with the CMS and several other projects.

If you want I could teach you how to work on that part of code

Kitzunu:
Works for me.
I think it could be a fun project

Winfidonarleyan:
Hmm, looks good, now we need to attach it to the core.

Kitzunu:
yeh, that's the part I don't know how to do x)

data/sql/updates/pending_db_world/rev_1592179496642830400.sql
```diff
@@ -0,0 +1,46 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1592179496642830400');
+
+CREATE TABLE IF NOT EXISTS `world_config` (
+  `type` TINYINT UNSIGNED NOT NULL DEFAULT 50,
+  `name` VARCHAR(255) NOT NULL,
+  `value` INT UNSIGNED NOT NULL,
```
Winfidonarleyan, line 6: If the value is float?

Winfidonarleyan:
Maybe I can help with this implementation. I've already started to rework the config system.

data/sql/updates/pending_db_world/rev_1592179496642830400.sql
Kitzunu: But there is no float in the config? Or maybe I am wrong?
When I did a quick look in the config I could only find values 0 - in the range of `INT UNSIGNED`

Winfidonarleyan: Configuration has 4 types
1. string
https://github.com/azerothcore/azerothcore-wotlk/blob/34077115a5de7d640517caf8136f8648f8f195e3/src/server/game/World/World.h#L373
2. int
https://github.com/azerothcore/azerothcore-wotlk/blob/34077115a5de7d640517caf8136f8648f8f195e3/src/server/game/World/World.h#L200
3. bool 
https://github.com/azerothcore/azerothcore-wotlk/blob/34077115a5de7d640517caf8136f8648f8f195e3/src/server/game/World/World.h#L74
4. float
https://github.com/azerothcore/azerothcore-wotlk/blob/34077115a5de7d640517caf8136f8648f8f195e3/src/server/game/World/World.h#L182

Kitzunu: true, totally missed that

Yehonal: I don't think that adding "type" column is the best way to do it, you can't store a "float" value inside an INT anyway.

I would go with value as a VARCHAR and then doing the conversion on core side based on type inside the core.

NOTE: If the core is not able to convert the DB value (maybe someone mispelled it) then we can throw an error inside the console and use the default value

Yehonal: We also have textual configurations (MOTD for example) and we could have more of them in future

Kitzunu: Type was just a suggestion so we can sort the configs, it would have no use in the core.
see https://github.com/azerothcore/azerothcore-wotlk/pull/3130#discussion_r439885565

Kitzunu: https://github.com/azerothcore/azerothcore-wotlk/blob/34077115a5de7d640517caf8136f8648f8f195e3/src/server/worldserver/worldserver.conf.dist#L6-L31
Just like this, but in the DB ^^

Winfidonarleyan:
Example https://github.com/TrinityCore/TrinityCore/pull/25321

src/server/game/Config/WorldConfig.h
```diff
@@ -0,0 +1,53 @@
...
+{
+	GAME_CONFIG_TYPE_BOOL,
+	GAME_CONFIG_TYPE_INT,
+	GAME_CONFIG_TYPE_FLOAT,
+	GAME_CONFIG_TYPE_STRING,
+
+	GAME_CONFIG_TYPE_UNKNOWN
+};
+
+class AC_GAME_API WorldConfig
```
Kitzunu, line 21: I think this is the main reason why build is failing, but cannot figure out what the AC equivalent is

Winfidonarleyan: Ahahah, ac don't support dynamic lib, need del this macros

PR #3134, opened 2020-06-15
fix(Core/BG): fix BG_QUEUE_INVITATION_TYPE_NO_BALANCE

FrancescoBorzi:
## CHANGES PROPOSED:

If you set `Battleground.InvitationType = 0` the expected behaviour would be:

> Normal, Invite as much players to battlegrounds as queued, don't bother with balance

as described in `worldserver.conf`. However, it looks like that before this PR this only affected the **initial** invitation (when the BG starts). After the BG started, it kept balancing the teams even if `Battleground.InvitationType` was set to `0`.

This caused problems when using modules like [mod-cfbg](https://github.com/azerothcore/mod-cfbg) that has its own balancing system. So when using such modules, it's preferrable to disable the core's balance using `Battleground.InvitationType = 0`.

This PR makes sure that ``Battleground.InvitationType = 0` actually disables the core's balancing system, **always**.

This PR does **NOT** affect the behaviour when setting `Battleground.InvitationType` to a value different than `0`.

## ISSUES ADDRESSED:

-  Prevent new Battlegrounds to start if there are already other Battlegrounds that have free slots (when using CFBG module)

- Related with issues similar to the one reported here azerothcore/mod-cfbg#11 (however this is core's fault and not the module's)

## TESTS PERFORMED:

- Build, run, etc...
- Tested in game
- As the mod-cfbg is responsible to balance, I've tested that it keeps balancing even when setting `Battleground.InvitationType = 0`

## HOW TO TEST THE CHANGES:

- Play with Battlegrounds
- Try to change `Battleground.InvitationType` in your `worldserver.conf` and make sure that it still works as expected

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Battlegrounds/Battleground.cpp
+++ b/src/server/game/Battlegrounds/Battleground.cpp
@@ -1310,8 +1310,11 @@ void Battleground::AddOrSetPlayerToCorrectBgGroup(Player* player, TeamId teamId)
 
 uint32 Battleground::GetFreeSlotsForTeam(TeamId teamId) const
 {
-    // if BG is starting and CONFIG_BATTLEGROUND_INVITATION_TYPE == BG_QUEUE_INVITATION_TYPE_NO_BALANCE, invite anyone
-    if (GetStatus() == STATUS_WAIT_JOIN && sWorld->getIntConfig(CONFIG_BATTLEGROUND_INVITATION_TYPE) == BG_QUEUE_INVITATION_TYPE_NO_BALANCE)
+    if (!(GetStatus() == STATUS_IN_PROGRESS || GetStatus() == STATUS_WAIT_JOIN))
+        return 0;
+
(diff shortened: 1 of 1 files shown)


pangolp:
Is there any way I can test it by myself inside a test server or does it require a certain number of accounts / people connected?

merdlin:
> Is there any way I can test it by myself inside a test server or does it require a certain number of accounts / people connected?

.debug bg 1

FrancescoBorzi:
> > Is there any way I can test it by myself inside a test server or does it require a certain number of accounts / people connected?
> 
> .debug bg 1

That's right. Alternatively, you could also test a more "real" scenario, for example using:

```sql
UPDATE battleground_template SET MinPlayersPerTeam = 2;
```

this will make your BGs start with 2 players required per faction (2 vs 2)

vhiperdev:
https://cdn.discordapp.com/attachments/520013597460398083/722279329991753758/unknown.png

Test performed using cfbg module

I've tested that it keeps balancing even when setting *Battleground.InvitationType = 0*

3 player alliance

PR #3136, opened 2020-06-19
chore(Config): Update GM.StartLevel config value

Kitzunu:
## CHANGES PROPOSED:
-  Change GM.StartLevel from 80 to 1 because it makes no sense to force a GM accounts players to be level 80 from a base install.  + it is defaulted to 1 in the core, so it makes 0 sense to override a default value like this.
```cpp
    m_int_configs[CONFIG_START_GM_LEVEL]        = sConfigMgr->GetIntDefault("GM.StartLevel", 1);
```

## HOW TO TEST THE CHANGES:

- create a character with an account that has GM Access

## Target branch(es):
- [x] Master

 

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2167,7 +2167,7 @@ GM.LogTrade = 1
 #        Description: GM character starting level.
 #        Default:     1
 
-GM.StartLevel = 80
+GM.StartLevel = 1
```
FrancescoBorzi, line 2170: I think it's handy to have it at `80` for GMs. Maybe we should just fix the `Default` ?

Kitzunu: Personally I hate it being default 80 for GMs, just because for me it makes no sense to default force 80 on a GM accounts :shrug:

FrancescoBorzi: ok, fair enough :)

IntelligentQuantum:
No need to test ^^

BarbzYHOOL:
yes, logic also it will stupidly takes all the first realm achievements if i recall
we should also prevent GMs from taking first realm achievs

PR #3144, opened 2020-06-20
fix(DB/SAI): improve Mulgore zone

Knindzagxg:
Mulgore is now fully updated, since every creature in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele Mulgore and run / test around.


merdlin:
Tested, it works.

data/sql/updates/pending_db_world/rev_1592668418033075700.sql
```diff
@@ -0,0 +1,145 @@
...
+(2964, 0, 0, 0, 0, 0, 100, 0, 2100, 3400, 12100, 13400, 11, 13322, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 'Windfury Sorceress - In Combat - Cast \'13322\'');
+
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 2965;
+DELETE FROM `smart_scripts` WHERE (source_type = 0 AND entryorguid = 2965);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+(2965, 0, 0, 0, 0, 0, 100, 0, 2800, 3200, 10800, 13200, 11, 9532, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 'Windfury Matriarch - In Combat - Cast \'9532\''),
+(2965, 0, 1, 0, 2, 0, 100, 1, 10, 40, 0, 0, 11, 332, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Windfury Matriarch - Between 10-40% Health - Cast \'332\' (No Repeat)');
```
Kitzunu, line 145: newline

ninjaskija:
@Knindzagxg 
I've noticed that the Palemane Poacher mob is attacking in range only at first and then is starting to follow you and attack in melee. I think the range attack should be prio over melee. The same happens with the Palemane Tanner mob. I know that the change should reside in the **smart_scripts** somewhere but looking at the values I'm not sure what flag is responsible for this behavior.
I thought as first that the non-repeatable flag was the culprit but I don't think is that because if I keep running and keep the distance it starts to shoot again but then tries again to come in melee range

FrancescoBorzi:
@ninjaskija as this PR has already been merged, if you think there is a problem please open a new issue

Knindzagxg:
Will see what i can do about mate.

A bit busy in RL, but should be done with job in next few days.

Issue #2147, opened 2019-07-28
SpellInfo - Possible errors with positive/negative effects


Shard-MW:
##### SMALL DESCRIPTION:
After working on fixing issue #2129, I noticed that there is **maybe** some wrong calculations about SpellInfo. 
I compared TC - SpellInfo.cpp (SpellInfo::_isPositiveEffectImpl)
https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3376
with AC - SpellInfo.cpp (SpellInfo::_IsPositiveEffect)
https://github.com/azerothcore/azerothcore-wotlk/blob/d8d0bed3508c69f735ae73bc199598068252560c/src/server/game/Spells/SpellInfo.cpp#L2545
I think that if Magmadar's Enrage spell (https://wowgaming.altervista.org/aowow/?spell=19451) was shown as negative instead of positive, there should be some other spells that have the same error.
My PR (#2146) doesn't fix the error for all spells, but only for Magmadar's Enrage. 

##### EXTRA NOTES:
I think porting the SpellInfo::_isPositiveEffectImpl from TC could fix it since they didn't had Magmadar's Enrage issue. 

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
Commit : 8d4a694e82b0a6f4e26eba44dfec9837b226b598

##### OPERATING SYSTEM:
Windows 10

##### MODULES:
None

Yehonal:
Have you noticed that this spell triggers the 19272 spell (https://wowgaming.altervista.org/aowow/?spell=19272)?

According to the spell_editor it's a SCHOOL_DAMAGE therefore it's negative

![image](https://user-images.githubusercontent.com/147092/85346404-8f286d80-b4f5-11ea-9abd-52e73b157f43.png)


Moreover here: https://github.com/azerothcore/azerothcore-wotlk/blob/d8d0bed3508c69f735ae73bc199598068252560c/src/server/game/Spells/SpellInfo.cpp#L2740 

we can clearly read that if a positive spell casts a non-positive spell, then the the spell aura of the "main cast" will be considered a "debuff".

I'm not saying that's correct but that's what happen. We can rubber-ducking the trinitycore code to understand if they are using a different approach and why

Yehonal:
[...]

PR #3149, opened 2020-06-21
refactor(Core/SpellInfo): Positive/Negative spells from SpellFamilyName into DB

Kitzunu:
## CHANGES PROPOSED:
* Implement SPELL_ATTR0_CU_POSITIVE_EFF0

* Implement SPELL_ATTR0_CU_POSITIVE_EFF1

* Implement SPELL_ATTR0_CU_POSITIVE_EFF2

* Move SpellFamilyName spells into DB to define if they are POSITIVE or NEGATIVE spells

* Rename column from entry to spell_id to make more sense

## TESTS PERFORMED:
- Builds
- Tested in-game (works)

## HOW TO TEST THE CHANGES:

1. .aura 61716
2. See that this is a Positive spell (Appears as a Buff)
3. .aura 34709
4. See that this is a Negative spell (Appears as a Debuff)

This means that the new Attribute works.

## Target branch(es):
- [x] Master

 

Shard-MW:
Hello,
I wonder what's the main goal of this PR ?
It's quite interresting because it allows us to "force" a spell to be positive through db, but honestly I don't think that's the way to go since positive/negative should be calculated from spells attributes...

I made an issue 1 year ago about something related to this https://github.com/azerothcore/azerothcore-wotlk/issues/2147 . 
Imo, going this way (hardcoding positives/negatives spells) will not solve the roots of the problem

Yehonal:
The more we can remove hardcoded IDs from the code the better it is!

However,  personally I don't like having a "spell_custom_attr" table. There's no "custom" thing in our database, they are only extra fields normalized from the spell_dbc because are not often use.

something like spell_extra_attr should be nice and ofc we need to change the code wherever the "custom" key is used. But maybe it's better to do it in another PR (?)

About the positive/negative debate instead, I'll make some research asap

Yehonal:
@Shard-MW thanks for your ISSUE about it, it's a very good start ;)

Kitzunu:
Well yes, this is custom, because these attributes do not exist in DBC therefore custom.

As for this PR, I am simply just moving it from hardcoded in CORE to DB.
It won't change any behavior as what I am removing in CORE is deciding for some spells if they should be NEGATIVE or POSITIVE.

Usually, this is handled in DBC (I think) but sometimes Blizzard was lazy I guess :)

Kitzunu:
Basically, the functionality is staying the same. Nothing will change for the user except that it is handled in the Database instead of being hardcoded in the core :)

Kitzunu:
@Shard-MW It does read attributes, but some spells are missing the attribute of being Positive or Negative (See your PR for example that you made a year ago that forces Magmadar's Enrage to be positive)

Before we had the Custom Attribute of NEGATIVE but never POSITIVE, so the main reason I add this is so that we don't need to hardcode the POSITIVE part when we already could force spells NEGATIVE

Yehonal:
@Kitzunu AzerothCore is a MMO framework first, that is compatible with wotlk opcodes. I would that everyone can keep this mindset. 

BTW "Custom" implies that it is something modified compared to something "official", but the DBC are just table extracted from the client that are a subset of what blizzard have on their server...and for sure they don't have a "custom" table xD

That's why for me "custom" makes no sense when it's something that's "our".

Kitzunu:
> AzerothCore is a MMO framework first, that is compatible with wotlk opcodes. I would like that everyone can keep this mindset.

I have no idea where I implied that AC is not compatible with WotLK opcodes?

> BTW "Custom" implies that there's something changed compared to something "official", but the DBC are just table extracted from the client that are a subset of what blizzard has on their server...and for sure they don't have a "custom" table xD

That's why it is custom x))))

Shard-MW:
> @Shard-MW It does read attributes, but some spells are missing the attribute of being Positive or Negative (See your PR for example that you made a year ago that forces Magmadar's Enrage to be positive)
> 
> Before we had the Custom Attribute of NEGATIVE but never POSITIVE, so the main reason I add this is so that we don't need to hardcode the POSITIVE part when we already could force spells NEGATIVE

Yes I know and I understand your opinion, but the question is, do we really need theses changes since we'll have to remove this anyway in the future ? haha

@Yehonal 
I think it's "custom" attributes because, as our friend @Kitzunu said, it handles some extra spell info that are not in dbc (like spell damage amount divided by the number of players that it hits). And since we don't really know how blizz handles it, we added "custom" xD. But yeah, "extra" seems more accurate for me

Yehonal:
@Kitzunu take a look at those:

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3406

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3413

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3409

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3410

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3411

https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellInfo.cpp#L3418

maybe we should check if they have issues and eventually add them

Yehonal:
@Shard-MW I've replied to your issue: https://github.com/azerothcore/azerothcore-wotlk/issues/2147#issuecomment-647829243

@Kitzunu can be interesting for you too

Kitzunu:
I get very mixed thoughts.

Do we want this or not? I dont want to waste more time if it is not wanted

Yehonal:
> I get very mixed thoughts.
> 
> Do we want this or not? I dont want to waste more time if it is not wanted

Moving it into the database is nice because you can always force a spell to be negative or positive via the database. That's a good idea imho. Go ahead!

However, we should also improve the algorithm ;)

Yehonal:
In fact, I was still suggesting you to test those spells that I've linked to you because TC is forcing them to be negative/positive and...since we don't have them in our switch/case...maybe we can check if they're bugged and eventually add inside your database PR

Kitzunu:
Yea ik, that was why I asked if I should bother :P

Will do it later today

Kitzunu:
@Yehonal Checked them and added them now

Kitzunu:
@Yehonal ready to merge?
Other stuff that can be moved can be done later

Yehonal:
Can anyone give us evidence of tests about this PR? @Kitzunu you can do it too if you have screenshots/video

Kitzunu:
I will provide video

Kitzunu:
https://youtu.be/R2REuYTorQo Here you see me aura a spell, it is negative, then apply SQL and the aura is positive

Yehonal:
Thank you! merged

FrancescoBorzi:
excellent work

Yehonal:
@Kitzunu I was checking id by id if you missed any spell in database, and I've some doubts:

1) 
                case 19451: // Enrage (Magmadar)
                    return true;

Enrage is not present in your SQL, that's good if we merge the other PR as soon as possible https://github.com/azerothcore/azerothcore-wotlk/pull/3153, but be aware that now it's bugged basically.

2) I've not found some of the following spells in your SQL (I manually got the ID from SpellWork since the check in code is done via SpellFamilyFlags/SpellIconID and looks like that Mind Flay is not in your list for example). Can you double check them?

            // mind flay
            if (SpellFamilyFlags[0] & 0x800000)
                return false;
            // Envenom
           if (SpellIconID == 2237)
                return true;
           // Slice and Dice
           else if (SpellFamilyFlags[0] & 0x40000)
               return true;

Kitzunu:
1. Well aware
2. Hmm weird, I used spellwork too but can double check

FrancescoBorzi:
I guess it's because when @Kitzunu did that work, @Yehonal 's PR wasn't on master yet

anyway I've merged the other PR so it should be fine now

Voxstrasza:
@Kitzunu this PR broke some boss mechanics that are meant to be debuffs but apply as a buff - phase punch on algalon is one of them

from your pr
`(64412, @SPELL_ATTR0_CU_POSITIVE_EFF0 | @SPELL_ATTR0_CU_POSITIVE_EFF1),`

phase punch: https://www.wowhead.com/spell=64412/phase-punch

will be sending you a DM but I just wanted to drop a comment here

Kitzunu:
DELETE FROM spell_custom_attr WHERE spell_id = 64412

as I explained in the dm it was probably when I used spellworks to find the spells for
if (SpellIconID == 242)
                return true;

PR #3161, opened 2020-06-25
[WIP] minor docker improvements

Yehonal:
This PR will be used to improve the docker layer for AzerothCore.

Currently, there are some processes that are not "DRY" and some minor wrong patterns

- [ ] github action step to produce the pre-compiled images for: world, auth, mysql and assets ( https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/installer/includes/functions.sh#L204)

- [ ] create the docker-compose.prod.yml that will use those pre-built images directly

- [ ]  put a volume (commented) option for etc bind


docker/assets/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine
+
+# install the required dependencies to get assets
+RUN apk add curl
+RUN mkdir /assets
+RUN curl -L https://github.com/wowgaming/client-data/releases/download/v7/data.zip > "/assets/data.zip" \
```
Yehonal, line 6: @Kapel would be better to use the `./acore.sh client-data` to download that data instead of this link directly

Kapel: @Yehonal this would require actually making a copy of acore.sh and main.sh (as acore.sh is just a wrapper) into the container. Also would have to copy apps dir inside. Just to call this single line, lol. If I add a variable for the v7 we can use it as a tag for tagging this image too.

Yehonal: What about bind instead of copy, then unbind when everything done?

We have tons of scripts and it's easy to have breaking things if we don't remeber to change them everywhere

However, if you don't sugfest ir then you could create a variable inside config.sh with that link so that we can use it for both. Then you just need to copy the config.sh eventually.

Let me kniw

trapped: Tuning in here, if I may: it might also be a good idea to join the various `RUN` steps into a single one (i.e. `RUN apk --no-cache add curl && mkdir -p /assets && ./acore.sh client-data`) in order to produce a single layer for the whole operation in the final Docker image (and adding `--no-cache` to `apk` to avoid bloating the Docker image with the package repo cache).

locus313:
Been doing some testing with github actions to publish to docker hub. Thought it might be useful for this pr:

```yml
name: Publish to Docker Hub

on:
  push:
    branches:
      - master
      - docker-hub

jobs:
  docker-worldserver:
    strategy:
      fail-fast: true
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - name: Docker login
        uses: docker/login-action@v1
        with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_TOKEN }}
      - name: Configure
        run: |
          docker --version
          docker-compose --version
      - name: Build
        run: |
          docker build -t ${{ secrets.WORLDSERVER_REPO }} -f docker/worldserver/Dockerfile docker/worldserver
      - name: Tags
        if: success()
        run: |
         docker tag ${{ secrets.WORLDSERVER_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.WORLDSERVER_REPO }}:${{ github.sha }}
         docker tag ${{ secrets.WORLDSERVER_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.WORLDSERVER_REPO }}:latest
      - name: Push
        if: success()
        run: |
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.WORLDSERVER_REPO }}:${{ github.sha }}
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.WORLDSERVER_REPO }}:latest
 
  docker-authserver:
    strategy:
      fail-fast: true
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - name: Docker login
        uses: docker/login-action@v1
        with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_TOKEN }}
      - name: Configure
        run: |
          docker --version
          docker-compose --version
      - name: Build
        run: |
          docker build -t ${{ secrets.AUTHSERVER_REPO }} -f docker/authserver/Dockerfile docker/authserver
      - name: Tags
        if: success()
        run: |
         docker tag ${{ secrets.AUTHSERVER_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.AUTHSERVER_REPO }}:${{ github.sha }}
         docker tag ${{ secrets.AUTHSERVER_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.AUTHSERVER_REPO }}:latest
      - name: Push
        if: success()
        run: |
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.AUTHSERVER_REPO }}:${{ github.sha }}
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.AUTHSERVER_REPO }}:latest

  docker-database:
    strategy:
      fail-fast: true
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - name: Docker login
        uses: docker/login-action@v1
        with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_TOKEN }}
      - name: Configure
        run: |
          docker --version
          docker-compose --version
      - name: Build
        run: |
          docker build -t ${{ secrets.DATABASE_REPO }} -f docker/database/Dockerfile .
      - name: Tags
        if: success()
        run: |
         docker tag ${{ secrets.DATABASE_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.DATABASE_REPO }}:${{ github.sha }}
         docker tag ${{ secrets.DATABASE_REPO }} ${{ secrets.DOCKER_USER }}/${{ secrets.DATABASE_REPO }}:latest
      - name: Push
        if: success()
        run: |
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.DATABASE_REPO }}:${{ github.sha }}
         docker push ${{ secrets.DOCKER_USER }}/${{ secrets.DATABASE_REPO }}:latest
```

docker/assets/Dockerfile
locus313: @Yehonal looking at the function used in `./acore.sh client-data` it looks like it assumes the bin path which ends up being `./env/dist/bin/` I see the url is hard coded in the function as well. 

I am wondering if it might be a little easier to add it to the .env.dist and pass the url to the Dockerfile this way. similar to how we currently handle the "DB_ROOT_PASSWORD" for the database container. I think we do it this way we would not have to use copy or a bind mount. 

Then just update the client-data function to load in the url from the env.dist file. What do you think?

locus313:
@Yehonal shouldn't we close this one out now?

PR #3175, opened 2020-07-08
feat(Core/Command): Reenable .teleport name on offline player

BarbzYHOOL:
- Reenable the possibility to `.teleport name xxx $home` even if the character is offline.

I tried it before and after PR with the command above on an online/offline player.

Merge it when you want, but the quicker the better (for my own usage xD)


BarbzYHOOL:
to be fair, i think it's really safe, i just uncommented code that existed, and i tried before and after (i mean I fixed this because I need it fixed)

PR #3176, opened 2020-07-08
fix(Core/Command): Can comment closed tickets

BarbzYHOOL:
Before this PR, we couldn't comment a closed ticket.
Now we can.

I'd like to know if some server owner find this inconvenient (maybe because they don't trust their GMs etc)? I personally need this option but maybe we could improve the feature to be configurable.

@icenet @disclosurez @Si1ker what do you guys think about this?


IceNet:
+Rep

Si1ker:
@BarbzYHOOL Totally useful, let's go for it! 🌟 🐱

BarbzYHOOL:
alright thanks guys

FrancescoBorzi:
@Si1ker why are you removing labels from PRs? o_O

Issue #1045, opened 2018-10-03
Debuff stacking Issue


santahashi:
##### SMALL DESCRIPTION:

Debuffs such as (Winter's chill ...) are not being stacked by 2 or more people

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Winter's chill
If the stacks is already 3 from Mage A, when Mage B attacks, the stacks should be increase to 4 

##### CURRENT BEHAVIOUR:

1) Mage A attacked a monster with FrostBolt 3 times --> Winter's chill stacked up to 3 times relatively
2) Mage B then attacked that monster again with FrostBolt, Winter's chill stacks resetted to 1

Talent ID: https://wowgaming.altervista.org/aowow/?spell=11180

##### STEPS TO REPRODUCE THE PROBLEM:

Winter's chill
1) Mage A attacked a monster with FrostBolt 3 times --> Winter's chill stacked up to 3 times relatively
2) Mage B then attacked that monster again with FrostBolt, Winter's chill stacks resetted to 1

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

Commit f2d16c3003ec17c31e57561e463ddd51097b4e12

##### OPERATING SYSTEM:

Windows Server 2012

##### MODULES:

None

##### OTHER CUSTOMIZATIONS:

None

BarbzYHOOL:
https://github.com/TrinityCore/TrinityCore/issues/21635

santahashi:
> [TrinityCore/TrinityCore#21635](https://github.com/TrinityCore/TrinityCore/issues/21635)

That is actually a different issue, that was frost bolt's effect not being renewed when another frost bolt cast.
This one the aura of Winter's chill from Mage A being completely replaced by Winter's chill aura of Mage B, in which it should collaborate as additional stacks.

So even the current trinity-core is broken, however, their latest release commit ed63d15dfe956360851018c55a61d6d27dd743c4 (Feb 19) got that fixed.
Worth a look, I just don't know where to look into.
https://github.com/TrinityCore/TrinityCore/releases/tag/TDB335.64

d3tr0l:
Also, is there a possibility there is same issue with rogue poisons etc or other classes debuffs? This needs further investigation.

a4501150:
inside method:
Aura* Unit::_TryStackingOrRefreshingExistingAura

pussywizard add check for find current aura from spell and change its stackamount, or refresh its duration
```
[...]

PR #2276, opened 2019-09-09
fix: debuff stacking issue for mage winter'schill

daromero2:
##### CHANGES PROPOSED:
-  
-  

##### ISSUES ADDRESSED:
- Closes

##### TESTS PERFORMED:
-
-

##### KNOWN ISSUES AND TODO LIST:

- [ ]
- [ ] 

##### Target branch(es):
- [x] Master

 

 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #3187, opened 2020-07-13
Winter's Chill now stacks from multiple sources in the same manner as…

ChrisCGalbraith:
… Sunder Armor.

Signed-off-by: Chris <chriscgalbraith@gmail.com>

## CHANGES PROPOSED:
-  Add Winter's Chill id 12579 to SpellMgr.cpp similar to Player Sunder Armor.

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1045

## TESTS PERFORMED:
- Builds without errors.
- Tested in-game.
- Used multiple clients to use two mages to check for correct stacking.
- Only tested on Windows.

## HOW TO TEST THE CHANGES:

- In-game, run two clients simultaneously.
- .learn Frostbolt, any rank will do. I used 27072 (rank 13).
- Mod level to 60, spend talent points down to Winter's Chill and take 3/3.
- Teleport to burning steppes, find nearest creature and cast a frostbolt from each client.
- Winter's Chill now stacks from multiple sources.

- [ ]

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3163,6 +3163,7 @@ void SpellMgr::LoadSpellCustomAttr()
             case 34655: // Snake Trap, Deadly Poison
             case 11971: // Sunder Armor
             case 58567: // Player Sunder Armor
+            case 12579: // Player Winter's Chill
             case 29306: // Naxxramas(Gluth's Zombies): Infected Wound
                 spellInfo->AttributesCu |= SPELL_ATTR0_CU_SINGLE_AURA_STACK;
                 break;


ChrisCGalbraith:
This should also close: https://github.com/azerothcore/azerothcore-wotlk/pull/2276

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/issues/1045#issuecomment-616419096

BarbzYHOOL:
I'm just curious, but why is the spell attribute flag in the DBC is:

   SPELL_ATTR3_STACK_FOR_DIFF_CASTERS           = 0x00000080, //  7 separate stack for every caster

and here it's 0x00400000

HMM while i'm writing this I realize it's a custom attribute in `enum SpellCustomAttributes`, so why not use the existing attribute that i mentioned above?

Kitzunu:
SPELL_ATTR3_STACK_FOR_DIFF_CASTERS makes sense yes.

Kitzunu:
But I don't understand why all mages should be able to stack on 1 aura?
It would basically mean you could have 7 mages in a group, all cast frost bolt, and max the stacks in 1 cast. (Which is what I have found no proof to either)

BarbzYHOOL:
offtopic: that's why naming matters `SPELL_ATTR3_STACK_FOR_DIFF_CASTERS` is long but it explains clearly what it does, no need to read anything else, we know what it does. that's why i cry at all the horrible names all the time xD

Kitzunu:
I haven't checked other spells, just done research on this one :P

Malow:
@Kitzunu While I don't have any proof I'm 99.9% sure that Winter's Chill is indeed supposed to stack as a single aura that different casters can increment, much like how Sunder Armor already works for example, I do know for a fact Winter's Chill worked like this in Vanilla and TBC (here is proof from Classic: https://youtu.be/bxjFjYX1N9w?t=42), and I see no reason why blizzard would've changed this. 

I think the possible options here are:
1. Have a single aura on the target shared by different casters, each incrementing the existing aura's application count. Definitely the best solution.
2. Have a single aura but different casters override the previously existing debuff. From what I can read in the original issue this is how it currently works. This is also a really bad solution as it would mean having 2 Winter's Chill specced mages would suck because you'd never get higher than like 1 or 2 stacks of it as they keep resetting the stack with their own. This is probably the second worst option in my opinion. I know some spells have worked like this in the past, I think for example Mind Flay from Shadow Priests worked like this in early Vanilla which is why having more than 1 Shadow Priest in a raid wasn't really possible. So I'm fairly certain that by WotLK they had fixed all spells that had stacking issues like this.
3. Have multiple auras on the target, each providing the crit % and stacking. This is obviously a bad solution and incorrect, because a raid with many mages would let you get to 100% crit, and also Winter's Chill is part of a debuff group together with Improved Scorch and Improved Shadowbolt and all of these are mutually exclusive effects, so Winter's Chill stacking with itself as multiple auras is probably the worst solution.
4. Have multiple auras on the target but only apply the effect of the strongest one. This approach is a bit messy as you would get a lot of debuffs that appear on the target but that actually doesn't have any effect, but at least it doesn't affect gameplay really in a negative way.


@BarbzYHOOL To me SPELL_ATTR3_STACK_FOR_DIFF_CASTERS sounds like the flag that if true would put you into option 3/4 of the ones I listed above and if false put you into option 1/2, while SPELL_ATTR0_CU_SINGLE_AURA_STACK if true would put you into option 1 if true and option 2 if false.

Regarding how Chris has chosen to implement it: It seems to me that the way he has done it is the correct way to do it as the state of the core is right now. All the spells in the game as far as I can see that works like this (single aura stacking from multiple casters) are listed here in this switch-case. Though the one thing I'm a bit unsure of is if all the spells listed in this switch-case here should be moved into the spell_custom_attr table in World-DB maybe. I'm not entirely sure how that table works but I assume it's meant to apply these custom attribute to spells just like this switch-case does, but in a non-hardcoded way.

ChrisCGalbraith:
> 
> 
> hmm how does it work for other spells with similar concept @Kitzunu ?
> 
> offtopic: that's why naming matters `SPELL_ATTR3_STACK_FOR_DIFF_CASTERS` is long but it explains clearly what it does, no need to read anything else, we know what it does. that's why i cry at all the horrible names all the time xD

We can see here at https://youtu.be/YFMWKyYmioY?t=54 , double frost mage setup, their Winter's Chill stacks from each mage, one does not over-write the other.

BarbzYHOOL:
sure this PR is acceptable if that's how it's done previously in the core (aka in a bad way, it's not the first time i see a custom flag created for something that already exists in the DBC, but it's probably because at the time of creation, all the flags had not been identified), but using the official flag might be better

I don't know what you said Malow, seems confused, would probably require testing

And then it's normal (blizzlike) that 5 mages can reach the full stack in 1 second if casting all at the same time?

Malow:
"And then it's normal (blizzlike) that 5 mages can reach the full stack in 1 second if casting all at the same time?"

Yes, just how a single Fire-mage can cast 1 Scorch on the target or how a single Warlock can cast 1 Shadowbolt and apply the same effect that all 5 stacks of Winter's Chill do. Even with Winter's Chill properly stacking from different casters it's still the weakest spell in this debuff-group as the others apply much faster and easier.

ChrisCGalbraith:
> 
> And then it's normal (blizzlike) that 5 mages can reach the full stack in 1 second if casting all at the same time?

I think the answer to this is yes. The winter's chill talent gives you % crit chance on your frostbolt spell in addition to the WC buff. It's quite a cookie-cutter talent if you're speccing frost. If you could only have one mage in a raid with WC, this seems antithetical to the direction blizzard took it's buff/debuff stacking in WotlK.

As frost is the go-to PvP spec, imagine playing in BG's with other mages, and your WC keeps overwriting and resetting stacks on a target, competing with your friendly mages. This is a win situation for the enemy, doesn't make any sense.

BarbzYHOOL:
note: next time name your branch to the PR, so you can easily make multiple PRs and not overlap work :)

all in all i'm for this PR to be merged (but I haven't tested it, I don't even know this spell lol) and ofc thank you Chris (and Malow)

ChrisCGalbraith:
> 
> 
> note: next time name your branch to the PR, so you can easily make multiple PRs and not overlap work :)

Thanks for pointing this out - I realise I hadn't branched my fork for the PR, much appreciated bud!

pklloveyou:
Tested.

Issue #3028, opened 2020-05-13
World Crash (SIGSEGV, Segmentation fault)


CosmeF:
Crash log:

https://gist.github.com/CosmeF/c8b4427be4eb5b105723ee8b2d01425a

##### SMALL DESCRIPTION:
Server crash, gdb generated the log above.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Not a clue, sorry

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
f97d494ffc90c3b3dde041564d6b2ca20ec22905

##### OPERATING SYSTEM:
Ubuntu 18.04

##### MODULES:
mod-anticheat-master
mod-boss-announcer
mod-eluna-lua-engine-master
mod-emblem-transfer
mod-premium
mod-reward-played-time-master
mod-transmog-master

##### OTHER CUSTOMIZATIONS:
I've lowered conflarate delay to 0 on https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Unit/Unit.cpp#L3928

Meltie2013:
@CosmeF Looks like its related to SAI but its a useless crash log without symbols being loaded, cause the server is compiled in release.

Need a log in Debug or RelWithDebInfo.

CosmeF:
@Meltie2013 Thanks a lot. I've checked the flag for aditional debug code, so hopefuly I'll catch some extra info next time, so I can share it on this issue.

hlainin:
This server error appears periodically. Is there a solution? How to output the error log, my logs are empty. in the logs this error "Segmentation fault"

Kitzunu:
@hlainin You need to build in Debug if you want a useful log

Helias:
@CosmeF, did you solve? is there any progress?

CosmeF:
Server did crash but haven't been able to register a more detailed crash log. Now i'm still waiting for it to crash again and check if the crash log works with some extra flags enabled.

Kitzunu:
Did you build in RelWithDebInfo or Debug this time?

CosmeF:
Yes, the build is now running whith coredebug flag. Still waiting for the crash though.

CosmeF:
Crashed again, segmentation fault. Log appears to be the same than last time. Am I doing something wrong here? I use "-WITH_COREDEBUG=1" option on cmake, using GDB. The modules are the same since last time.

[...]

PR #3195, opened 2020-07-17
fix(DB/Sai): Crash

Kitzunu:
* Fixes https://github.com/azerothcore/azerothcore-wotlk/issues/3028

## CHANGES PROPOSED:
-  Remove faulty SAI

Gameobjects should not be allowed to be in combat with zone

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3028

## HOW TO TEST THE CHANGES:

This was tested and confirmed in https://github.com/azerothcore/azerothcore-wotlk/issues/3028#issuecomment-654536304

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1594986175990195100.sql
+++ b/data/sql/updates/pending_db_world/rev_1594986175990195100.sql
@@ -1,4 +1,4 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1594986175990195100');
 
-DELETE * FROM `smart_scripts` WHERE `entryorguid` = 124371 AND `id` = 4;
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
@Kitzunu good that this solves the crash. However the core should be robust enough to not crash even when there are invalid SAI in the DB. The core should rather just give an error instead (and the SAI of course should be removed from the DB too, as you did).

Kitzunu:
Indeed, that is in my todo but this is the only case so it is more important to do this first imo

FrancescoBorzi:
@Kitzunu totally agree. Maybe it's worth opening a ticket for that too? so we don't forget

Kitzunu:
Got friends over atm but screenshotted so I will remember

Issue #1064, opened 2018-10-24
Installer.sh is only using remote "origin"


BarbzYHOOL:
##### SMALL DESCRIPTION:

When we choose option 3) pull, it will only choose the "origin" remote. What if we changed remotes and branches' names?

##### EXPECTED BLIZZLIKE BEHAVIOUR:

We should be able to edit this in the conf/config.sh(.dist) file.

##### CURRENT BEHAVIOUR:

Script is guessing what is the branch/remote to "git pull"...

##### STEPS TO REPRODUCE THE PROBLEM:

1. Change remote origin to something else
2. Choose option 3) of installer.sh
3. It won't pull from azerothcore's official repo

##### BRANCH(ES):

master

##### AC HASH/COMMIT:

879553ddf83085031641b19102e9cd97a3e334e8

PR #3196, opened 2020-07-17
chore(APPS/Installer): Pull from config branch

elmerbud:
# Script is guessing what is the branch/remote to pull.

## CHANGES PROPOSED:
-  Added INSTALLER_PULL_FROM in conf/config.sh.dist.
-  Default value of INSTALLER_PULL_FROM is "master".

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1064

## TESTS PERFORMED:
- Tested on ubuntu 20.04 

## HOW TO TEST THE CHANGES:
- Change remote origin to something else
- Choose option 3) of installer.sh
- It will pull from the branch defined in config

## Target branch(es):
- [x] Master

 
### DIFF
--- a/apps/installer/includes/functions.sh
+++ b/apps/installer/includes/functions.sh
@@ -30,16 +30,16 @@ function inst_updateRepo() {
     AC_REPO_CONFIG="$AC_PATH_CONF/config.sh"
     AC_REPO_CONFIG_DIST="$AC_PATH_CONF/config.sh.dist"
     if [[ -f "$AC_REPO_CONFIG" ]]; then
-	echo "Configuration $AC_REPO_CONFIG found."
-	source "$AC_REPO_CONFIG"
+        echo "Configuration $AC_REPO_CONFIG found."
+        source "$AC_REPO_CONFIG"
     else
-	echo "Configuration $AC_REPO_CONFIG not found. Looking for DIST."
-	if [[ -f "AC_REPO_CONFIG_DIST" ]]; then
-	    echo "Configuration $AC_REPO_CONFIG_DIST found."
+        echo "Configuration $AC_REPO_CONFIG not found. Looking for DIST."
+        if [[ -f "AC_REPO_CONFIG_DIST" ]]; then
+            echo "Configuration $AC_REPO_CONFIG_DIST found."
             source "$AC_REPO_CONFIG_DIST"
-	else
-	    INSTALLER_PULL_FROM="master"
-	fi
+        else
+            INSTALLER_PULL_FROM="master"
+        fi
     fi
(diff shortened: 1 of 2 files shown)


apps/installer/includes/functions.sh
```diff
@@ -27,7 +27,9 @@ function inst_configureOS() {
 
 function inst_updateRepo() {
     cd "$AC_PATH_ROOT"
-    git pull origin $(git rev-parse --abbrev-ref HEAD)
+    source "$AC_PATH_CONF/config.sh.dist" #hack to get the parameter for the branch name
+    echo "Pulling from $INSTALLER_PULL_FROM branch defined in config.sh.dist"
```
FrancescoBorzi, line 31: why from `config.sh.dist` ? shouldn't it pull from `config.sh` ?

@Yehonal

elmerbud: Should I add first to check if `config.sh` exist, if not read from `config.sh.dist` ?

FrancescoBorzi: yes of course ;)

elmerbud:
How is it now?

apps/installer/includes/functions.sh
Yehonal: what is this...

It's not the way, wait a second

Yehonal: As you can see here: https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/installer/includes/includes.sh

The installer includes the bash_shared/common.sh

https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/bash_shared/common.sh#L4

Here we already handle the conf.sh and conf.dist.sh for ALL our bash scripts, in fact, you can notice that in other function we directly use configuration variables without including those files.

Can you try without your "source" stuff please?

conf/config.sh.dist
```diff
@@ -170,3 +170,6 @@ DB_AUTH_NAME="acore_auth"
 DB_CHARACTERS_NAME="acore_characters"
 
 DB_WORLD_NAME="acore_world"
+
+# Branch configuration for installer to pull from
+INSTALLER_PULL_FROM="master"
```
Yehonal, line 175: Since we have several branches and someone could clone -b "2.x" for example.
For a future-proof configuration would be better to write something like:

```
# Branch configuration for the installer to pull from.
# By default git will select the current working branch
# You can change it to "master" if you want the latest updates
INSTALLER_PULL_FROM=$(git rev-parse --abbrev-ref HEAD)
```

FrancescoBorzi:
@elmerbud there are conflicts

PR #3205, opened 2020-07-21
fix(build): remove warning

Helias:
Fix this warning
```C++
azerothcore-wotlk/src/common/Configuration/Config.cpp: In member function ‘bool ConfigMgr::LoadAppConfigs(const string&)’:
azerothcore-wotlk/src/common/Configuration/Config.cpp:212:27: warning: zero-length gnu_printf format string [-Wformat-zero-length]
         sLog->outString("");
```

How to test: build the core before and after, the warning will be fixed with this PR.

I don't think that this requires any tests, only the build from travis, so after a quick review I think that we can merge it


src/common/Configuration/Config.cpp
```diff
@@ -209,7 +209,6 @@ bool ConfigMgr::LoadAppConfigs(std::string const& applicationName /*= "worldserv
     // #2 - Load .conf file
     if (!sConfigMgr->LoadMore(_initConfigFile))
     {
-        sLog->outString("");
```
FrancescoBorzi, line 212: maybe this was supposed to print a new empty line?

Kitzunu: There are a lot of them, esp when loading worldserver, the use I cant understand

Winfidonarleyan:
```cpp
sLog->outString();
```

Helias: but the reason of this empty sLog? it's just a newline or..?

Winfidonarleyan: yes, empty line, cosmetic only

Kitzunu:
Ty for the PR :)

PR #3207, opened 2020-07-21
docs(issue_template): Add a template for feature requests

Kitzunu:


FrancescoBorzi:
not sure if we want this here or in the forum

Kitzunu:
I simply overweight it here since there are so many feature/suggestions requests here anyway

PR #3209, opened 2020-07-21
docs(pull_request_template): Introduce the new github way

Kitzunu:


Kitzunu:
This will allow us to have several PR templates too if that is wanted in the future

Helias (approved):

PR #3223, opened 2020-07-25
fix(core/mail): prevent client crash while receiving a malformed mail

Helias:
This PR prevent the crash of the client in case of receiving a specific malformed mail.
```
If you want to test this you need to be an AC contributor with a high rank, if so contact me in private on discord.
```
This is really a critical bug.

## CHANGES PROPOSED:
-  add some filter sending/showing mail

## ISSUES ADDRESSED:
- no issues

## TESTS PERFORMED:
before PR: the client crash
after PR: the client does not crash and can see the malformed mail

## HOW TO TEST THE CHANGES:
```
If you want to test this you need to be an AC contributor with a high rank, if so contact me in private on discord.
```

## Target branch(es):
- [x] Master

 

Tyrelis:
I would like to help although I am not high enough rank.

Kitzunu:
I mean the fix is going to be public anyway so why resteict testing?

wowissues:
It's works!

Tyrelis:
> I mean the fix is going to be public anyway so why resteict testing?

Because users can abuse it.

Helias:
Added other checks, I am not sure if my solution is the best, but sure it solves the problem and prevents malicious mail

Tyrelis:
> Added other checks, I am not sure if my solution is the best, but sure it solves the problem and prevents malicious mail

Very few members can test this fix as they don't have high enough contributor rank. So minus will push the changes or have it tested by one of the staff. Since it's critical issue time is of the essence.

wowissues:
> > Added other checks, I am not sure if my solution is the best, but sure it solves the problem and prevents malicious mail
> 
> Very few members can test this fix as they don't have high enough contributor rank. So minus will push the changes or have it tested by one of the staff. Since it's critical issue time is of the essence.
> 
> --Update FrancescoBorzi approved it so minus will push it--


Since I provided this exploit, I checked out the fix from @Helias and it fixes the problem!

Helias:
Guys, for me we can merge this I tested it several time and wowissues too.
In this period I didn't find the time, but I would already merge it

Kitzunu:
Thank you for the PR! 👍

PR #3229, opened 2020-07-27
refactor(scripts/BlackrockMountain): Remove commented NPC scripts

pangolp:
## CHANGES PROPOSED:
- Remove unused lines of scripts.

## ISSUES ADDRESSED:
- Closes 

## Target branch(es):
- [x] Master

 

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockDepths/blackrock_depths.cpp
```diff
@@ -442,622 +442,17 @@ enum DughalQuests
...
-#define SAY_WINDSOR_AGGRO2          "I bet you're sorry now, aren't you !?!!"
-#define SAY_WINDSOR_AGGRO3          "You better hold me back $N or they are going to feel some prison house beatings."
-#define SAY_WINDSOR_1               "Let's get a move on. My gear should be in the storage area up this way..."
-#define SAY_WINDSOR_4_1             "Check that cell, $N. If someone is alive in there, we need to get them out."
-#define SAY_WINDSOR_4_2             "Get him out of there!"
-#define SAY_WINDSOR_4_3             "Good work! We're almost there, $N. This way."
-#define SAY_WINDSOR_6               "This is it, $N. My stuff should be in that room. Cover me, I'm going in!"
-#define SAY_WINDSOR_9               "Ah, there it is!"
-
 enum MarshalWindsor
```
Winfidonarleyan, line 445: Maybe need to delete? and same enums

pangolp: Okay, I'll check. Maybe they need to be eliminated.

pangolp: Confirmed, there were 3 enum that were not used. Thanks for the observation.

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockDepths/blackrock_depths.cpp
```diff
@@ -1208,10 +586,4 @@ void AddSC_blackrock_depths()
     new npc_phalanx();
     new npc_lokhtos_darkbargainer();
     new npc_rocknot();
-    // Fix us
```
Kitzunu, line 1211: 1. Are these NPC's scripted somewhere else?
2. If no, then how broken are these scripts? Unnecessary to remove scripts if they are only slightly broken and not implemented elsewhere. If that were the case it would be better to uncomment them so that the NPC's actually have a script and see over if it is correct to an extent

pangolp: The truth is that I do not know what function those lines fulfill, but I consider that if they are commented, they should not be in the emulator. It's a subjective argument, it depends on who's looking at it. I don't have enough information, so I made the pull request, to know what the rest of the community thinks about it.

pangolp: The idea was to check the gossip, and on the way, I found that script that had 611 commented lines and it called my attention. Honestly if they are part of a script, I think they should be somewhere else if they are not used. Out of 1200 lines that the whole file has, 600 are comments.

pangolp: If those scripts are unfinished, then we should finish them last, but leaving them in comment form, I think is not the best way. And if they were moved to the database, through SmartAI, they should not be commented inside the script. It's a subjective assessment.

Si1ker (approved):
I have checked this script in TC, they do not have any of this in their blackrock_depths.cpp:
https://github.com/TrinityCore/TrinityCore/blob/48eba61e90e3958613ceca05bb6a550fce8816ad/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockDepths/blackrock_depths.cpp


Moreover, I haven't see a DB implementation for those NPCs in the commit history.

PR #3236, opened 2020-07-28
chore(LICENSE): Move LICENSE-AGPL3.md to .github

Kitzunu:


FrancescoBorzi:
doesn't this break all existing links to the license? including the ones among the source files...

Kitzunu:
Fsir point

FrancescoBorzi:
so maybe it's better to leave this file where it is :)

PR #3216, opened 2020-07-22
fix(DB/GameTables): correct links to data from gt* DB tables

Winfidonarleyan:
## CHANGES PROPOSED:
- Support long values for description and etc.
- Support signed values in `spelldifficulty_dbc`
- Correct links to data in `gt*_dbc` DB tables

## ISSUES ADDRESSED:
- Closes #3242
- Closes #3237
- Closes #3198
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3268

## TESTS PERFORMED:
- Apply default dbc data from https://github.com/Winfidonarleyan/dbc_sql

## HOW TO TEST THE CHANGES:
- Apply default dbc data from https://github.com/Winfidonarleyan/dbc_sql
- #3198

## Target branch(es):
- [x] Master
 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #3242, opened 2020-08-01
fix(Core/Player): Fix incorrect mana regeneration at level 1

### LINKED ISSUE #3237: Mana does not regenerate
I am on commit: https://github.com/azerothcore/azerothcore-wotlk/commit/91bbbf08eb73837f50469ad5dac7a106cecadb8f and I found that mana is not regenerating. My Setup is Debian 10 x86_64 with MySQL 5.7.  I tested with a human mage and ran my mana down to near 0. I waited about 5 min and my mana is still only 5 (it never increased). I confirmed that Rate.Mana in worldserver.conf is 1.

I also tested with an Undead Mage also and the mana failed to regenerate.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92170709-mana-does-not-regenerate?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSkorpiox:
## CHANGES PROPOSED:
- Edit Player::OCTRegenMPPerSpirit to correct an incorrect value

## ISSUES ADDRESSED:
- Closes #3237 

## TESTS PERFORMED:
- Tested in-game
- Tested only on Windows

## HOW TO TEST THE CHANGES:
- Create a level 1 character
- Use mana
- Watch if mana regeneration is correct

## Target branch(es):
- [x] Master

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -6006,8 +6006,7 @@ float Player::OCTRegenMPPerSpirit()
     if (level > GT_MAX_LEVEL)
         level = GT_MAX_LEVEL;
 
-//    GtOCTRegenMPEntry     const* baseRatio = sGtOCTRegenMPStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level-1);
-    GtRegenMPPerSptEntry  const* moreRatio = sGtRegenMPPerSptStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level-1);
+    GtRegenMPPerSptEntry  const* moreRatio = sGtRegenMPPerSptStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level);
     if (moreRatio == NULL)
         return 0.0f;
 


Kitzunu:
How do you know this is will return the correct amount of regeneration?

TC uses the same thing with -1 and regeneration works fine there.

TheSkorpiox:
First, sGtRegenMPPerSptStore is refering to acore_world.gtregenmpperspt_dbc into the database
Then, we have pclass (which is 9 for a mage for exemple), GT_MAX_LEVEL which is 100, and the character level

LookupEntry gets the data stored at the id passed in parameter. Before, for a mage level 1 it was 800 `(9-1)*100+1-1`, now i fixed it to 801 `(9-1)*100+1`

As of why, this is related to the data in that table.
Each GT_MAX_LEVEL entries represent the mana regeneration for a given class. Id's start at 1 in a database context, so each class should take values from `1 * pclass` to `GT_MAX_LEVEL * pclass`. It is not right to have 800 for a level 1 character, because it refers to the last entry of the previous class.

If TrinityCore uses -1 and it works fine, that means something is wrong with the values stored in that particular table

TheSkorpiox:
Here is to illustate:

State of the database
image

Logs that i added, first 2 values are for a level 2 character, last 2 for a level 1 character
image

src/server/game/Entities/Player/Player.cpp
```diff
@@ -6006,8 +6006,7 @@ float Player::OCTRegenMPPerSpirit()
     if (level > GT_MAX_LEVEL)
         level = GT_MAX_LEVEL;
 
-//    GtOCTRegenMPEntry     const* baseRatio = sGtOCTRegenMPStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level-1);
-    GtRegenMPPerSptEntry  const* moreRatio = sGtRegenMPPerSptStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level-1);
+    GtRegenMPPerSptEntry  const* moreRatio = sGtRegenMPPerSptStore.LookupEntry((pclass-1)*GT_MAX_LEVEL + level);
     if (moreRatio == NULL)
```
pklloveyou, line 6010:
``` c++
-if (moreRatio == NULL)
if (moreRatio == nullptr)
```
Can this fix the error?

merdlin:
Is this PR correction of these issues? #3237

TheSkorpiox:
Yes, as stated in the description
image

Winfidonarleyan:
No need this changes

TheSkorpiox:
Please elaborate, i'd be happy to know why

BarbzYHOOL:
Check Winfy's fix first https://github.com/azerothcore/azerothcore-wotlk/pull/3216 and test it (without your PR) please because it seems that these issues come from deep changes made by Winfi

TheSkorpiox:
I did, and it works. I close my PR

HelloKitty:
@TheSkorpiox was right though. DBC tables like gtregenmpperspt are not handled in emulators properly. Correct range of regen for a class goes from, for example, 601 to 700. It's quite obvious when you consider the formula (pclass-1)*GT_MAX_LEVEL + level-1 is just invalid for Warriors as for level one it would attempt to retrieve key 0.

Imo it's an off by one error in both TC and Azerothcore and changing the other PR which seemed like it probably changed the DBC tables to fit this incorrect core handling seems silly.

PR #3250, opened 2020-08-04
[Zone] Ungoro Crater

Knindzagxg:
Ungoro Crater is now fully updated, since some creatures in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele UngoroCrater and run / test around.


pklloveyou:
Tested

Kitzunu:
ty

PR #3251, opened 2020-08-04
[Zone] Eversong Woods

Knindzagxg:
Eversong is now fully updated, since some creatures in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele EversongWoods and run / test around.


Knindzagxg:
> 
> 
> 2020-08-04 16:23:13 SmartAIMgr: Entry 15949 SourceType 0 Event 0 Action 11 uses min/max params wrong (9500/1000), skipped.

Quick typing -.-' Sorted mate.

pklloveyou:
Tested

Kitzunu (approved):

PR #3253, opened 2020-08-04
[Zone] Isle of Quel'Danas

Knindzagxg:
Few creatures updated in Deadscar since they were missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele Sunwell and then move / fly to Deadscar where you doing daily quests.


pklloveyou:
Tested

PR #3254, opened 2020-08-04
feat(DB/SAI): [Azeroth] Missed & Skipped Creatures

Knindzagxg:
During tests, missed few creatures in previous pulls.

To test it:
Apply SQL
Restart your server

.npc add ID


pklloveyou:
Tested

PR #3255, opened 2020-08-04
feat(DB/Creature): [Zone] Thousand Needles

Knindzagxg:
Thousand Needles is now fully updated.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele ThousandNeedles and run / test around.


pklloveyou:
Tested

PR #3256, opened 2020-08-04
feat(DB/SAI): [Azeroth] Classic Rares - Part I

Knindzagxg:
Still haven't check all of them, but fixed quite a lot of them so far.

To test it:
Apply SQL
Restart your server

.npc add ID


data/sql/updates/pending_db_world/rev_1596556919928240000.sql
```diff
@@ -0,0 +1,322 @@
...
+(3253, 0, 0, 0, 4, 0, 100, 0, 0, 0, 0, 0, 11, 7278, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 'Silithid Harvester - On Aggro - Cast \'7278\'');
+
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 5409;
+DELETE FROM `smart_scripts` WHERE (source_type = 0 AND entryorguid = 5409);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+(5409, 0, 0, 0, 11, 0, 100, 0, 0, 0, 0, 0, 41, 35000, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 'Harvester Swarm - On Respawn - Despawn In 35000 ms');
```
Kitzunu, line 242: I havent had time to check but is this true?

BarbzYHOOL: at least prior to that I bet the spawns never despawned, making the encounter much harder after a wipe

Kitzunu: That is what it does, when the creature respawns, it will despawn in x ms

Knindzagxg: Yeah, it's good like that.

It's bound to creature above (ID 3253) and rare spell ID 7278 -> It will be spawned on aggro.
Supposed to last 2 minutes, but we learned from Northrend issues that those made lots of trouble with memory, and i made it to be only 35 seconds (Which should be enough for 1 level lower character to kill it).

In other words, i didn't wanted to risk that someone can do quick multi pulls and somehow find a way to spawn lots of those for 2 minutes and crash server xP

BarbzYHOOL: yeah but they should despawn after 35s WHEN out of combat, not during combat

Kitzunu: Yea I agtee to use out of combat

Knindzagxg: Updated.

pklloveyou:
Tested

PR #3259, opened 2020-08-05
feat(bash): support for derivated distro + refactor conf

Yehonal:
## CHANGES PROPOSED:

Support for ubuntu/debian derivated distro with a simple switch case

Our bash system is able to recognize the current OS and eventually install dependencies automatically. For linux we currently support debian and ubuntu, however, there are some distributions based on ubuntu/debian that share the same package manager. In that case we can use the ubuntu/debian installer file.

In this case I've added the support for KDE Neon distrubution based on ubuntu

## TESTS PERFORMED:

Tested on KDE neon and ubuntu

## HOW TO TEST THE CHANGES:

Run ./acore.sh install-deps


FrancescoBorzi:
Affects only Neon and has been tested by the author, good to go

apps/bash_shared/common.sh
```diff
@@ -1,7 +1,7 @@
 function registerHooks() { hwc_event_register_hooks "$@"; }
 function runHooks() { hwc_event_run_hooks "$@"; }
 
-source "$AC_PATH_CONF/config.sh.dist" # "hack" to avoid missing conf variables
+source "$AC_PATH_CONF/dist/config.sh" # include dist to avoid missing conf variables
```
FrancescoBorzi, line 4: why renaming/moving the `config.sh.dist` ?

Yehonal: for 2 reasons:

1) copy/past is easier, you don't have to change the name of the file

2) .dist file are not recognized by editors, you need to specify the syntax type manually.

PR #3306, opened 2020-08-16
fix(DB/Dungeon): improved Scholomance

Knindzagxg:
Changes:

Min, Max & Modifier Damage for whole Dungeon.
Added flags for each boss, so they will appear with skull instead regular level elite.

All checks have been done trough Classic Realms and Youtube videos.

How to test:
Apply SQL
Restart Server and delete your Cache.
Make sure to add very very basic gear towards Dungeon ilevel.

Before you start checking damage:
.mod hp 50000
.learn 1908 (Uber heal over time)

If there is any issue with damage and if you get instantly killed, lemme know.

PS. Last 4 classic dungeons should be finished in next few weeks and then i will focus on Wotlk. TBC Dungeons are fully updated.


Geekena:
damage and spells are ok
i was unable to see last boss until other are killed
is that ok?

Knindzagxg:
Yea, thats fine.
To start last one, all previous have to be killed due teleportations.

Kitzunu:
You could use the command `.i setbossstate`

Kitzunu (approved):

PR #3311, opened 2020-08-17
feat(CI): add Ubuntu 18.04 to the CI

FrancescoBorzi:
- Make the OS configurable in the CI core build, this allows us to test different systems
- Adding one build with Ubuntu 18.04, so we can support the latest 2 LTS versions of Ubuntu.


Rochet2:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations
The build matrix seems to support adding specific build parameters to it, which can avoid the copy paste if the scripts are otherwise identical or somewhat similar.

PR #3314, opened 2020-08-18
feat(CI): add support for GCC

FrancescoBorzi:
Add support for GCC:

- ubuntu 20.04 with gcc-10
- ubuntu 20.04 with gcc-9 (default)
- ubuntu 18.04 with gcc-7 (default)

Also:
- remove clang9 with modules, since checking modules with clang 10 is enough
- remove clang6 with ubuntu 20.04, as we already test it in ubuntu 18.04
- rename `clang10` to just `clang` in 20.04, as it's the default one (consistency)

Co-authored-by Viste <viste02@gmail.com>

PR #3315, opened 2020-08-19
WIP: feat(CI): Enable GCC support on the CI

meerd:
The build system doesn't finalize successfully with up to date GCC versions because of the warnings that are treated as errors. This MR enables the GCC support back again.

EDIT by Shin:

there is a cyclic dependency that is causing a linking error when using GCC, the line is `GetCaster()->VisitNearbyObject(15.0f, searcher);` from `instance_zulfarrak.cpp`. According to @meerd, this is caused by:

> instance_zulfarrak.cpp L144 has a dependency to a module from game library, 
> and 
> MovementHandler.cpp L727 (ArenaSpectator::HandleSpectatorSpectateCommand(&chc, summoner->GetName().c_str())) also relies on the script module.

### DIFF
--- a/apps/ci/ci-install.sh
+++ b/apps/ci/ci-install.sh
@@ -2,7 +2,61 @@
 
 set -e
 
-cat >>conf/config.sh <<CONFIG_SH
+COMPILATION_OPTIONS='-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache'
+# These flags below are disabled for gcc, gcc 10 and clang 10.
+ADDITIONAL_OPTIONS='-DCMAKE_C_FLAGS="-Werror" -DCMAKE_CXX_FLAGS="-Werror"'
+C_COMP=Unknown
+CPP_COMP=Unknown
+
+case $COMPILER in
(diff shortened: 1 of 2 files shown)


apps/ci/ci-install.sh
```diff
@@ -2,59 +2,82 @@
...
+    C_COMP=clang-9
+    CPP_COMP=clang++-9
+    COMPILATION_OPTIONS="${COMPILATION_OPTIONS} ${ADDITIONAL_OPTIONS}"
     ;;
 
   "clang10" )
     time sudo apt-get install -y clang-10
-    echo "CCOMPILERC=\"clang-10\"" >> ./conf/config.sh
-    echo "CCOMPILERCXX=\"clang++-10\"" >> ./conf/config.sh
-    # disable -Werror for clang-10 TODO: remove when this is fixed https://github.com/azerothcore/azerothcore-wotlk/issues/3108
```
FrancescoBorzi, line 52: I'd preserve this TODO somewhere, and possibly open an issue for GCC as well and linking it here. So we don't forget :)

FrancescoBorzi:
closing in favour of:

https://github.com/azerothcore/azerothcore-wotlk/pull/3314

PR #3319, opened 2020-08-19
fix(CI): launch CI unit test with relative path

peti446:
## CHANGES PROPOSED:
-   Make the test CI run using relative path instead of hard-coded to avoid build problems in case of the path changing in any way, this includes the fork being named other than azerothcore-wotlk.

## Target branch(es):
- [x] Master


FrancescoBorzi:
nice catch @peti446

FrancescoBorzi (approved):

PR #3358, opened 2020-08-26
fix(Core/player): do not show zone at first login

### LINKED ISSUE #3354: Location shows when new char is created
##### CURRENT BEHAVIOUR:

When you create a new character it shows the location on the character select screen which isnt the normal behaviour. 

If shouldnt show any location at all

1.  Create a new character
2.  See the location

##### EXTRA NOTES:

The location doesnt show on TrinityCore. So it works there.

##### AC HASH/COMMIT:

d8ad0c520060280fe48fb9d73e2284cad241794d

##### OPERATING SYSTEM:

Windows 10 Enterprise (real activated version)

##### MODULES:

No

##### OTHER CUSTOMIZATIONS:

No

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92449366-location-shows-when-new-char-is-created?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Helias:
Do not show location at first login.

## CHANGES PROPOSED:
- if it is the first login send `0` in the world server packet as `zone` instead of the value of the zone

## ISSUES ADDRESSED:
- Closes #3354 

## TESTS PERFORMED:
- created new character and checked if the zone is showed or not at first login

## HOW TO TEST THE CHANGES:
- create new character
- check if the zone is showed at first login

## Target branch(es):
- [x] Master

 

Tyrelis:
I have tested the PR before and after the fix and created a new character. The issue seems to be resolved I did not see any issues as a result pop up.

FrancescoBorzi (approved):

FrancescoBorzi:
good job @Helias and thanks @Tyrelis for testing this

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2018,14 +2018,19 @@ bool Player::BuildEnumData(PreparedQueryResult result, WorldPacket* data)
...
+
     *data << uint8(skin);
     *data << uint8(face);
     *data << uint8(hairStyle);
     *data << uint8(hairColor);
     *data << uint8(facialStyle);
 
     *data << uint8(fields[10].GetUInt8());                   // level
-    *data << uint32(fields[11].GetUInt16());                 // zone
+    *data << zone;                                           // zone
```
Kitzunu, line 2033: Shouldn't this be `*data << uint32(zone);`?

Just referencing to the lines above that have `*data << uint8(skin);` for example

Kitzunu: @Helias

Helias: `zone` variable is already casted to `uint32()` at line 2024 at its initialization, so, I don't think it's needed to re-cast it.

probably you already understood but I want to be clear:

actually the zone variable contains `uint32((atLoginFlags & AT_LOGIN_FIRST) != 0 ? 0 : fields[11].GetUInt16())` so, it's like
```C++
*data << uint32((atLoginFlags & AT_LOGIN_FIRST) != 0 ? 0 : fields[11].GetUInt16());  // zone
```
if I do another cast uint32() it will be
```C++
*data << uint32(uint32((atLoginFlags & AT_LOGIN_FIRST) != 0 ? 0 : fields[11].GetUInt16()));
``
so like `uint32(uint32( ...something... )`;

Comment (Kitzunu) [reply src/server/game/Entities/Player/Player.cpp:2033]: Yea but so is `    uint8 skin = fields[5].GetUInt8();` etc...
Perhaps it could be nice to have a standard to go after?

**Comment by felwyn** [comment]
Works fine for me now too. Thank you for fixing it.

File: src/server/game/Entities/Player/Player.cpp (hunk 1 above)
Comment (Helias) [reply src/server/game/Entities/Player/Player.cpp:2033]: mhm, you mean to move the casting next-to *data instead of the initialization?
from the readability yes, but from the evaluation of the expressions it's better to do it in the initialization imho

because it's like:
```C++
int x = 'a';
....
*data << int(x);
```

because the 'a' can be converted into an integer, but it should do when you assign the value, not later. otherwise, the compiler will do it automatically or throw an error or just consider it differently (I think).

anyway, I am not an expert of C++, but in my mind the casting should be at the initialization.

It's nice to have a standard and I can change it, let me think about it, this evening maybe I will update it

Kitzunu: I just copied the code from the file :D

Helias: done

PR #3390, opened 2020-08-31
[Auth] Frozen Feature

Knindzagxg:
Transfering from SC by @FrancescoBorzi request.

Feature for Auth table and Frozen Account information.


Geekena:
tested 
no problems and frozen tab appear now

data/sql/updates/pending_db_auth/rev_1598881657136430300.sql
```diff
@@ -0,0 +1,13 @@
...
+/*
+ * General: Auth Update
+ * Update by Knindza | <www.azerothcore.org>
+ * Copyright (C) <www.shadowburn.net> & <www.lichbane.com>
+*/
+
+/* Content 3.3.5 */
+
+ALTER TABLE `account`
+	CHANGE COLUMN `active` `active` TINYINT(1) UNSIGNED NOT NULL DEFAULT '0' AFTER `recruiter`,
```
Kitzunu, line 11: This column doesn't exist so you'd need to add this too :)

Kitzunu:
Also shouldn't there be some Core changes to make this work?
Otherwise it is just a useless table

Knindzagxg:
I think its for Website Armory.

Wasnt able to find more info on those that are on Sunwell Core.

Kitzunu:
Better to not add this imo, they are blind changes that we have no idea what they would have been used for

Knindzagxg:
Im gonna check more into this as well.

Pretty sure its for website, but will know more soon.

FrancescoBorzi:
> Also shouldn't there be some Core changes to make this work?
> Otherwise it is just a useless table

I was thinking the same

Knindzagxg:
Closing this until i find more information for what is used.

Currently, it's only for website to check Disabled accounts.

Issue #3422, opened 2020-09-02
ICC Intro Event - Saurfang/Muradin


Voxstrasza:
##### CURRENT BEHAVIOUR:
Saurfang and Muradin show dialog to start Deathbringer Saurfang event.  Tirion doesn't initiate intro event after killing the two The Damned at the beginning of the instance.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
When you kill the two The Damned at the door, the intro event should start.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Enter ICC (any difficulty, on alliance or horde characters)
2. Kill the two The Damned
3.  No intro event.

##### EXTRA NOTES:
I believe this broke when #3294 was merged - original bug report was to de-spawn the NPCs after the event, but instead the event does not start at all.

#3261 was also merged to correct Saurfangs model.  Unsure if related.

##### AC HASH/COMMIT:
2632cd6fcd8b

##### OPERATING SYSTEM:
Win10

##### MODULES:
N/A

##### OTHER CUSTOMIZATIONS:
N/A

Winfidonarleyan:
need fix
```
NPC_HIGH_OVERLORD_SAURFANG_DUMMY            = 32315,
``

Shadofiend:
confirm

Akimboss:
I called this error, but no one listened :D, this error caused by the query you added here Core/DB fix Icecrown Citadel Raid - Saurfang Intro Roleplay (#3261) ( this query UPDATE `creature` SET `id` = 37187 WHERE `id` = 32315; ) how can you replace the ID with another ID ? with all my respect that's just dumb ... the ID should be ModelID not ID of NPC, in general without talking too much revert the query and the event will work fine and for the model, just update the model to 30583 which is the modelID of Saurfang ... you can either do this by database or ingame by going to npc and use ! .Npc set model 30583, and it's fixed forever for event and model, please test before future merge ...

Voxstrasza:
> 
> 
[...]

PR #3261, opened 2020-08-05
fix(DB/Creature): ICC Saurfang correct model

r0m1ntik94:
## CHANGES PROPOSED:
- PARTIALLY fix issue #3246 

## TESTS PERFORMED:
- Tested on Ubuntu 18.04

## HOW TO TEST THE CHANGES:
Apply SQL and restart server

PR #3432, opened 2020-09-05
fix(DB) ICC Intro Event - Saurfang/Muradin MODELID

Akimboss:
Hello this is the fixe for #3261 and #3422 
The #3261 caused the bug #3422 it wasn't the right way to do it, and i call it even when they merged it first time, but no one listen till someone else reported the bug, so here's the fixe and good luck everyone


data/sql/updates/pending_db_world/rev_1599321019527522400.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1599321019527522400');
+UPDATE creature_template SET modelid1='30583' WHERE (entry='32315');
```
Kitzunu, line 2: No.
Update the modelID in `creature` not `creature_template`.

Akimboss: I updated it

FrancescoBorzi: please don't use strings where you can just use numbers:

```sql
UPDATE creature_template SET modelid1 = 30583 WHERE entry = 32315;
```

FrancescoBorzi:
@Akimboss first of all many thanks for your contribution!

Just a suggestion for your future PRs, do not use `master` branch. Create 1 new branch per every PR.

You can follow this ;) https://www.azerothcore.org/wiki/How-to-create-a-PR

Remberance:
tested, worked.

FrancescoBorzi (approved):

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
FrancescoBorzi: @Akimboss please resolve @Kitzunu 's comment

locus313:
tested. looks good.

Goatform:
@FrancescoBorzi  @Akimboss 

So this is not fully correct but I think it won't matter in the long run.

NPC entry 32315 should keep the same model.
Wowhead source (picture): https://www.wowhead.com/npc=32315/high-overlord-saurfang

The NPC entry that was used in that change is correct, however it should not be used the way it was used (when you merged). You never remove SQL row just like that as you did in commit like this: 3a56cdbd4a

Source for entry 37187 - https://www.wowhead.com/npc=37187/high-overlord-saurfang
(You can even look at picture examples and you will see that it will show that NPC inside ICC)

So to sum this up, the problem with commit 3a56cdbd4a was:

1. SQL code itself... you are not supposed to remove 1 row ... changing manually entry ids like that) but instead, create a totally new row if the one does not exist in DB already.
2. When you are doing change like that you have to also update the C++ part (script) which he did not and somehow that commit was approved.

(The list of the problem is not referring to this PR but to the commit that caused the issue with all of this)

peti446:
@Goatform is correct, the script needs to be updated to reflect the new ID or just rever the old GUID to use the specific ID and model, also the creature_template for 32315 does not appear to be in my database ?

Remberance:
37187 maybe in the front of Saurfang(ICC 4TH boss)

32315 in the gate of icc? i guess.

Akimboss:
> 37187 maybe in the front of Saurfang(ICC 4TH boss)
> 
> 32315 in the gate of icc? i guess.

Yuup, this query has been tested successfully, waiting to be merged 👍

frytik:
If someone want to take look into full intro, you can here https://pastebin.com/raw/2rPZBrnu https://pastebin.com/raw/CaCvd3qj

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
Kitzunu: I still stand by my comment here.

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1599321019527522400');
+
+UPDATE creature_template SET modelid1 = 30583 WHERE entry = 32315;
```
Kitzunu, line 3: add \` around keywords

Kitzunu (changes requested):

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
Akimboss: done

FrancescoBorzi:
It's not clear to me whether @Goatform 's concerns have been addressed or there is still some disagreement towards this change (I haven't checked myself - currently I can't run the client).

Kitzunu:
I still stand by reverting the first commit and then updating modelid in `creature`

Remberance:
我还是说中文吧，我英文水平too  bad
原来那个32315是在大门前的萨鲁法尔大王，只是模型有问题。然而37187是在小萨鲁法尔前的大王，他会有开战的gossip，所以会影响非官方。

而且大门前出发弗丁喊话的npc entry不对自然会影响剧情触发。所以修改32175的模型(即本帖)是正解。

peti446:
@Remberance I Used a translater to translate it so not sure if that will work, I tested this and on my end it did not fix it as the entry was still 37187, I am here with @Kitzunu and simply change the model ID and the npc entry back to 32315 as this is how the script is done:
`UPDATE 'creature' SET 'id' = '32315', 'modelid' = '30583' WHERE ('guid' = '200984');` I have tested this and it works like a charm

Also, I would like to mention that the given creature mentioned in this PR does not exits, so I assume we would also need to add to this PR the creature in the correct position after we change the template.

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1599321019527522400');
+
+UPDATE `creature_template` SET `modelid1` = 30583 WHERE `entry` = 32315;
```
Kitzunu, line 3:
```suggestion
UPDATE `creature` SET `id` = 32315, `modelid` = 30583 WHERE `guid` = 200984;
```

Kitzunu (changes requested):

data/sql/updates/pending_db_world/rev_1599321019527522400.sql
Goatform: @Kitzunu 
``` sql
UPDATE `creature` SET `id` = 32315, `modelid` = 30583, `curhealth`='4183500' WHERE `guid` = 200984;
```
Do it like this in case some guys were going back and forth with fixing the issue temporarily.

peti446:
The fix works, he has no longer the ui interaction and killing the two The Dammed  it triggers the interaction

FrancescoBorzi:
so should this be merged?

peti446:
@FrancescoBorzi I would say yes it fixes the problem and reverts it to what is used to be.

FrancescoBorzi (approved):

PR #3437, opened 2020-09-06
fix(Scripts/Wintergrasp): fix disabled always false

FrancescoBorzi:
## CHANGES PROPOSED:
-  Fix enable/disable Wintergrasp

## ISSUES ADDRESSED:

- Required for: https://github.com/azerothcore/azerothcore-wotlk/pull/3314

## TESTS PERFORMED:
- None

## HOW TO TEST THE CHANGES:

- Try to enable/disable Wintergrasp and make sure it's really enabled/disabled accordingly

## Target branch(es):
- [x] Master

 

Helias (approved):

sanctum32:
In base C rules, 0 = false, 1 or higher value is considered as true. This check previously was 
```c
if (true == false)
```
which means it was always false ;)

its good catch and pretty much this shouldn't require PR at all and you could just make direct commit :P

FrancescoBorzi:
> this shouldn't require PR at all and you could just make a direct commit

we never push commits directly to master, even the simplest commit can be subject to mistakes and so we open PR first but we can indeed merge quickly in case of small changes

pklloveyou:
Tested

PR #3468, opened 2020-09-10
fix(Core/Spell): Implement SPELL_ATTR0_CU_NEEDS_AMMO_DATA

Kitzunu:
## CHANGES PROPOSED:
-  Implement SPELL_ATTR0_CU_NEEDS_AMMO_DATA
-  This will fix the animation of the spell 60988

## ISSUES ADDRESSED:
- Closes 

## TESTS PERFORMED:
- Builds
- Tested in-game

## HOW TO TEST THE CHANGES:

1. .cast 60988
See that without the PR the animation will show as a cube.

## Target branch(es):
- [x] Master

 

FrancescoBorzi (approved):

src/server/game/Spells/Spell.cpp
```diff
@@ -4291,60 +4291,93 @@ void Spell::WriteCastResultInfo(WorldPacket& data, Player* caster, SpellInfo con
...
+        {
             // hardcode areas limitation case
             switch (spellInfo->Id)
             {
                 case 41617:                                 // Cenarion Mana Salve
                 case 41619:                                 // Cenarion Healing Salve
+                {
                     data << uint32(3905);
                     break;
+                }
```
FrancescoBorzi, line 4308: I don't think we should require brackets in switch-case, but I'm not against it either :)

Kitzunu: I just saw someone that used it so I used it for all.
I also thought it was kinda weird

Nefertumm: Only needed if you define variables inside the case.

src/server/game/Spells/Spell.cpp
```diff
@@ -4340,12 +4345,19 @@ void Spell::WriteCastResultInfo(WorldPacket& data, Player* caster, SpellInfo con
...
-             break;
+                 }
+
+                 ItemTemplate const* proto = sObjectMgr->GetItemTemplate(item);
+                 if (proto && proto->ItemLimitCategory)
+                 {
+                     data << uint32(proto->ItemLimitCategory);
+                 }
+                 break;
+             }
```
Kitzunu, line 4360: I dont understand the gcc warning about this

FrancescoBorzi: @Kitzunu it means you have a `case` inside a `switch` where you don't specify neither `break` nor `[[fallthrough]]``:

```
Spell.cpp:4360:14: error: this statement may fall through
```

https://en.cppreference.com/w/cpp/language/attributes/fallthrough

it's always better to specify `[[fallthrough]]` whenever you did NOT put a `break` INTENTIONALLY. This is meant to avoid bugs where the programmer just forgets to add `break`.

Kitzunu:
For some reason, this does not work anymore.

PR #3480, opened 2020-09-14
fix(Scripts/Ulduar): cast electrical charge only if player dies

mslmdevu:
## CHANGES PROPOSED:
The current implementation casts Electrical Charge
(https://www.wowhead.com/spell=61902/electrical-charge)
even though the unit that died is not a player.

This change fixes it so the logic is inverted; the player type is checked first before performing the spell in phase 3.

## TESTS PERFORMED:
- We tested this with a full 25-man raid on a server using this core.


FrancescoBorzi (approved):

Helias (approved):

Nefertumm:
Nice catch!

Si1ker:
Tested, fix is working, I have been using minions, it won't generate new charges when they die in phase 3

The picture shows 0 charges.

image

Kitzunu:
Ty for the PR!

PR #3513, opened 2020-09-28
fix(DB/creature_template): Nerubian Burrower's immunity

### LINKED ISSUE #3486: [Raid-ToC] Nerubian Burrower can be cc-able from all sources
good
 **REPORT CRUSADER TEST:**

1. **Npc  https://es.wowhead.com/npc=34607/perforador-nerubiano**

 **Raid Boss anub'arak, either normal or heroic, the skills that I will mention can be used with the following skills, which remains to be fixed:**

**https://es.classic.wowhead.com/spell=9484/chain-not-dead (Chain Undead)**

https://en.wowhead.com/spell=3355/freezing-trap (Freezing Trap)

https://es.classic.wowhead.com/spell=20066/repentimiento (Repentance)

https://es.wowhead.com/spell=57546/ahuyentar-el-mal-superior (Drive away evil)

https://db.wowlatinoamerica.com/?spell=60192 (Freezing Arrow)

https://es.classic.wowhead.com/spell=339/enwrapping-roots (Entangling roots)

**TTest the video malfunction demonstrated in this RAID Facebook server azeroth project**
https://www.facebook.com/WoW-Azeroth-Project-335-Camino-a-la-Luz-del-Alba-108019180728305/videos/toc-25h-raid-1-y-2/534619004070804/    error report Video of the last minutes since they put CC since this server has an emulator  **azeroth core** server _  **Azeroht project_**

Test the video
[https://youtu.be/uaXSPY6oAKc?t=4553](url)
 
### **¿HOW IT SHOULD WORK?**

TankSpot's Guide to Heroic Anub'arak (25 Hard)
OFFICIAL  https://www.youtube.com/watch?v=zDmdb9RYecY

_In the tankspot video they explain that only stun are used to cut the cast of the adds, other cc's are never mentioned this video is from 2009_
https://www.youtube.com/watch?v=PlMtNG8D2x0

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/93010031-raid-toc-nerubian-burrower-can-be-cc-able-from-all-sources?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Si1ker:
This change will make the Nerubian Burrower immune to any cc effect, only interrupt or stun effects are allowed.

Reference is taken from retail videos and TankSpot guide at YouTube:

TankSpot's Guide to Heroic Anub'arak (25 Hard)
https://www.youtube.com/watch?v=zDmdb9RYecY

In the TankSpot video, they explain that only interrupt or stun effects are used to cut the cast of the adds, other cc's are never mentioned.
https://www.youtube.com/watch?v=PlMtNG8D2x0

= = = = = = = = = = = = = = = = = = = =

Original immune mask: 549520145
Fixed immune mask: 550187859

**Change makes immune to the following spells:**

-- 2094 | Blind
-- 339 | Entangling Roots
-- 60192 | Freezing Arrow
-- 3355 | Freezing Trap Effect
-- 49203 | Hungering Cold
-- 20066 | Repentance
-- 33786 | Cyclone
-- 9484 | Shackle Undead

**NPC per mode:**

-- 34607 | ToC 10-man Normal Raid
-- 34648 | ToC 25-man Normal Raid
-- 35655 | ToC 10-man Heroic Raid
-- 35656 | ToC 25-man Heroic Raid

**Mechanic_immune_mask added and what it blocks:**

MECHANIC_DISORIENTED = 2 | Prevents the spell 2094
MECHANIC_ROOT = 64 | Prevents the spell 339
MECHANIC_FREEZE = 4096 | Prevents the spell 60192, 3355 and 49203
MECHANIC_KNOCKOUT = 8192 | Prevents the spell 20066
MECHANIC_BANISH = 131072 | Prevents the spell 33786
MECHANIC_SHACKLE = 524288 | Prevents the spell 9484

## CHANGES PROPOSED:
-  Make the Nerubian Burrower immune to any cc effect, only interrupt or stun effects are usable

## ISSUES ADDRESSED:
- Closes #3486 
- Without this change, you can easily win this encounter by using different types of CC on these adds which are an important mechanic of the encounter, you can leave them out of combat by using very long spell effects and focus on Anub'arak

## TESTS PERFORMED:
- Applied SQL changes
- Tested in-game

## HOW TO TEST THE CHANGES:

**Before applying the SQL changes:**
- .npc add 34607
- Use any of the following spells against this add with .cast [Spell ID below]
9484 / 3355 / 20066 / 57546 / 60192 / 339 / 33786
- Any of these spells will take effect

**After applying the SQL changes:**
- Try to use any of the spells on the NPC while you combat with him, none of them will take effect
- Any stun or interrupt spell is still usable

- This NPC has a debuff called "Expose Weakness" with Spell ID "67721" , when in combat, and after some time, it will stack up to 9 stacks. The problem is when another NPC like this one, uses the same debuff against a player, it resets the stacks.

In combat with 1 Nerubian Burrower: It will apply "Exose Weakness" until 9 stacks
In combat with 2 Nerubian Burrower: One will apply "Expose Weakness", at the second time, the debuff will increase to 2 stacks but when the other Nerubian Burrower applies "Expose Weakness" it resets the stacks to 1

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
+++ b/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
@@ -1,5 +1,9 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601277180341917800');
 
+/*
+ * Update by Silker | <www.azerothcore.org> | Copyright (C)
+*/
+
 UPDATE `creature_template` SET `mechanic_immune_mask`=`mechanic_immune_mask`|2|64|4096|8192|131072|524288 WHERE `entry` IN 
 (34607, 34648, 35655, 35656);
 
--- a/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
+++ b/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
@@ -1,8 +1,8 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601277180341917800');
 
--- Immune to Spell IDs: 9484 / 3355 / 20066 / 57546 / 60192 / 339 / 33786
-UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34607; -- ToC 10-man Normal Raid
-UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34648; -- ToC 25-man Normal Raid 
-UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35655; -- ToC 10-man Heroic Raid
-UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35656; -- ToC 25-man Heroic Raid
+-- Immune to Spell IDs: 9484 / 3355 / 20066 / 10326 / 60192 / 339 / 33786
+UPDATE `creature_template` SET mechanic_immune_mask=550187859 WHERE  entry=34607; -- ToC 10-man Normal Raid
+UPDATE `creature_template` SET mechanic_immune_mask=550187859 WHERE  entry=34648; -- ToC 25-man Normal Raid 
+UPDATE `creature_template` SET mechanic_immune_mask=550187859 WHERE  entry=35655; -- ToC 10-man Heroic Raid
+UPDATE `creature_template` SET mechanic_immune_mask=550187859 WHERE  entry=35656; -- ToC 25-man Heroic Raid
 
--- a/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
+++ b/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
@@ -1,0 +1,8 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601277180341917800');
+
+-- Immune to Spell IDs: 9484 / 3355 / 20066 / 57546 / 60192 / 339 / 33786
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34607; -- ToC 10-man Normal Raid
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34648; -- ToC 25-man Normal Raid 
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35655; -- ToC 10-man Heroic Raid
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35656; -- ToC 25-man Heroic Raid
+
--- a/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
+++ b/data/sql/updates/pending_db_world/rev_1601277180341917800.sql
@@ -1,8 +1,8 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601277180341917800');
 
 -- Immune to Spell IDs: 9484 / 3355 / 20066 / 10326 / 60192 / 339 / 33786
-UPDATE `creature_template` SET mechanic_immune_mask=550187859 WHERE  entry=34607; -- ToC 10-man Normal Raid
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/rev_1601277180341917800.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601277180341917800');
+
+-- Immune to Spell IDs: 9484 / 3355 / 20066 / 57546 / 60192 / 339 / 33786
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34607; -- ToC 10-man Normal Raid
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=34648; -- ToC 25-man Normal Raid 
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35655; -- ToC 10-man Heroic Raid
+UPDATE `creature_template` SET mechanic_immune_mask='550187859' WHERE  entry=35656; -- ToC 25-man Heroic Raid
```
FrancescoBorzi, line 7: you should not pass strings where numbers are expected

Kitzunu: Also dont override flags, add them using mim=mim|flag

Mim = mechanic immune mask

HoustonCore:
@Si1ker copy this)
```sql 
-- Immune ToC 10-man Normal Raid
UPDATE `creature_template` SET `mechanic_immune_mask`=`mechanic_immune_mask`|0|4|8|9|16|22|23|29 WHERE `entry` IN 
(34607, 34648, 35655, 35656); -- ToC 10-man Normal Raid

HoustonCore:
I wrote how it would be correct.

Si1ker:
@HoustonCore Thanks for trying but it is wrong.
I have tested your query, it results in an immune mask of 549520159 which makes the NPC vulnerable to most of the cc's in my example.

Only my suggested immune mask of 550187859 will work towards the intention of this PR. The immune mask I suggested is the correct one, but the way you get to it is.. not the best according to the "AC way" which I understand.

I am updating now my PR description to be more clear about the issue addressed and the query based on @FrancescoBorzi  and @Kitzunu  format directions.

Waiting for their approval 🐱

HoustonCore:
I didn't understand you correctly, sorry)
Default "mechanic_immune_mask" in AC 549520145 if you want "mechanic_immune_mask" 550187859 you only need to add 6 masks.

We shouldn't use numbers like this "|2|64|4096|8192|131072|524288" this is not correct. I have corrected my query and now it is correct.

```sql
-- Immune ToC 10-man Normal Raid
UPDATE `creature_template` SET `mechanic_immune_mask`=`mechanic_immune_mask`|1|6|12|13|17|19 WHERE `entry` IN 
(34607, 34648, 35655, 35656); -- ToC 10-man Normal Raid

Kitzunu:
@HoustonCore You cannot do that because that would be adding flags that do not exist, you can find all flags here https://www.azerothcore.org/wiki/creature_template#mechanic_immune_mask

Kitzunu (approved):

HoustonCore:
@Kitzunu Yes I got it wrong again the Value (2^Bit) now @Si1ker do it right.

FrancescoBorzi (approved):

iyasenai:
I have tested the changes, please watch the video.

https://www.youtube.com/watch?v=E3f2pWtrkzM

FrancescoBorzi:
@iyasenai thanks! you can add "AzerothCore" to the title of your youtube videos, so it can be found when searching for "AzerothCore"

Winfidonarleyan:
Video, very very good! @iyasenai good job!

Issue #3032, opened 2020-05-14
(DB/Dungeon/Quest): Missing quest items in Shadow Labyrinth Heroic


TheWatchingI:
##### SMALL DESCRIPTION:
Quest items are missing in the heroic mode of the Shadow Labyrinth.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
The codex of blood (https://tbcdb.com/?quest=10094) and the soul devices (https://tbcdb.com/?quest=10091) should be available in the Shadow Labyrinth in heroic mode.

##### CURRENT BEHAVIOUR:
The quest items are missing, but only on heroic. In the normal version of the dungeon they are available.

##### STEPS TO REPRODUCE THE PROBLEM:
1.  Add the quests "The Codex of Blood" and "The Soul Devices"
2. Walk through the Shadow Labyrinth in heroic mode

##### EXTRA NOTES:
Like I said: The quests work on normal difficulty.

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
cdf518a

##### OPERATING SYSTEM:
Windows 10

##### MODULES:
ah-bot
morphsummon
solo-lfg

wizzymore:
Hi,

This can be fixed easily, but i can't find proof of the game object being present in the heroic dungeon

Racknar0:
Tested 

![image](https://user-images.githubusercontent.com/38267074/98484959-f4e56f80-21e0-11eb-993c-d32240a35f10.png)


I doubt that the Gobject was in the air but check and your position is correct


![image](https://user-images.githubusercontent.com/38267074/98484974-2827fe80-21e1-11eb-91cc-5e4c55f7c2c3.png)

PR #3546, opened 2020-10-05
fix(DB/Dungeon/Quest): The Codex of Blood

wizzymore:
## CHANGES PROPOSED:
- Make The Codex of Blood to be seen in heroic as well

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3032

## TESTS PERFORMED:
- Checked the GObject in normal and heroic and is working as intended

## HOW TO TEST THE CHANGES:

Change dificulty to normal first, and then test again on heroic, it should appear on both normally
.go c 69955
Take the quest The Codex of Blood ( https://tbcdb.com/?quest=10094 )
Make sure you are level 70 ( if not, set your level to 70 and relog to see the quest )
.go gob 23860

## Target branch(es):
- [x] Master

 

Helias (approved):

data/sql/updates/pending_db_world/rev_1601868704763174600.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1601868704763174600');
+
+DELETE FROM `gameobject` WHERE (`id` = 182947) AND (`guid` IN (23860));
```
FrancescoBorzi, line 3: this query may fail if there is already a row having `guid` = `23860` and `id` different than `182947`.

If you want to make sure the query doesn't fail, then:

```sql
DELETE FROM `gameobject` WHERE `guid` = 23860;
```

Also, if (and ONLY if) you want to make sure your new `INSERT` will be the ONLY spawn of the gameobject `182947`, you can add this too:

```sql
DELETE FROM `gameobject` WHERE `id` = 182947;
```

resulting in:

```sql
DELETE FROM `gameobject` WHERE `guid` = 23860;
DELETE FROM `gameobject` WHERE `id` = 182947; -- only if you want to delete all current spawns of 182947
```

FrancescoBorzi (changes requested):

data/sql/updates/pending_db_world/rev_1601868704763174600.sql
wizzymore: Would it be better to have something like this?
```sql
 UPDATE `gameobject` SET `spawnMask` = 3 WHERE (`guid` = 23860 and `map` = 555);
```
As we only need the spawnMask of 23860 to be updated in the map 555 (Auchindoun: Shadow Labyrinth).

FYI: The generated query i added in the `.sql` was generated by Keira3

FrancescoBorzi: ah yes, definitely. If you are updating an existing row, always prefer `UPDATE`

wizzymore: Perfect :D

FrancescoBorzi (approved):

Gierlappen:
Tested here https://github.com/azerothcore/azerothcore-wotlk/issues/3032

Gierlappen (approved):

Racknar0:
Tested! 

image

I doubt that the Gobject was in the air but check and your position is correct

image

PR #3568, opened 2020-10-11
fix(Config): Set automatic arena pts distrib + up player limit

BarbzYHOOL:
## CHANGES PROPOSED:
- Arena points must be distributed automatically like in official WOW. https://wow.gamepedia.com/Arena_team
- Player limit was arbitrarily set to 100. It should be unlimited by default.

Also not being enabled by default means people could forget to set these configuration correctly and ruin their server launch. haha not cool

- [ ]  If this is validated, any member of the staff can edit this PR to edit the default value in the CPP functions too (i did this quickly with the web tool)
- [ ] 

## Target branch(es):
- [x] Master

 

src/server/worldserver/worldserver.conf.dist
```diff
@@ -187,7 +187,7 @@ Compression = 1
 #                     1+  - (Enabled)
 #                     0   - (Disabled, No limit)
 
-PlayerLimit = 100
+PlayerLimit = 0
```
Kitzunu, line 190: Change line 186

Kitzunu: Too

src/server/worldserver/worldserver.conf.dist
```diff
@@ -2901,7 +2901,7 @@ Arena.RatingDiscardTimer = 600000
 #        Default:     0 - (Disabled)
 #                     1 - (Enabled)
 
-Arena.AutoDistributePoints = 0
+Arena.AutoDistributePoints = 1
```
Kitzunu, line 2904: Line 2901 too

BarbzYHOOL: you mean the default description? gonna do it

btw you can do it by clicking "edit this file" :P

asfo:
The problem I see with "no-limit" in world is that you can't control your resources, I know that most of the servers are low-population but, in the Wiki it states that for 100 users the hardware for RAM required is 16gb, so, with this change in the "no-limit" some new founders could start a wow server without having a reference on limits. I would say that based on the wiki, this should be set to the amount of users for 8gb of RAM since is the most-common VPS with best prices...

And I know this is "blizz-like" but even Blizzard have a limit :P

BarbzYHOOL:
> The problem I see with "no-limit" in the world is that you can't control your resources, I know that most of the servers are low-population but, in the Wiki, it states that for 100 users the hardware in RAM required is 16gb, so, with this change in the "no-limit" some new founders could start a wow server without having a reference on limits. I would say that based on the wiki, this should be set to the number of users for 8gb of RAM since is the most-common VPS with the best prices...
> 
> And I know this is "blizz-like" but even Blizzard has a limit :P

It's not our role to determine what their hardware is (everyone should run tests before tho), and not our role to limit the players count. In the end, a good owner will check in details every option in the config. Can be said for the previous config default value too but at least this one is not arbitrary and seems more logic to me ^^

PR #3571, opened 2020-10-14
feat(Actions): Automatically add labels on PRs

Kitzunu:
Automatically add Labels to PRs when they are opened depending on which files has been changed


Kitzunu:
Well, this honestly just needs approval seeing as it works :)

Kitzunu:
Love macOS...

.github/label-pr.yml
```diff
@@ -0,0 +1,12 @@
+- regExp: ".*\\.sql+$"
+  labels: ["DB"]
+- regExp: ".*\\.(cpp|h)+$"
+  labels: ["CORE"]
```
FrancescoBorzi, line 4: This is wrong and reminds me that we should write this somewhere on our wiki.

By "Core" we mean everything that is:

- C++ **except** the contents of `/src/server/scripts`, which are instead "Scripts" (for historical reasons).
- DB structure changes. So SQL changes can be "Core" too.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,16 @@
+# Workflow to associate labels automatically
+name: labeler
+# Trigger the workflow on pull request events
+on: [pull_request]
+jobs:
+  label:
+    runs-on: ubuntu-18.04
```
FrancescoBorzi, line 7: why not using latest Ubuntu? `ubuntu-20.04`

Kitzunu: Because those who made the script use that one, dno

.github/label-pr.yml
Kitzunu: well, if you do DB changes that count as core too, then it will probably have a core change as well.
This was more thought of helping people to more easily understand what type of PR it is and what way it would need to be tested

.github/workflows/labeler.yml
FrancescoBorzi: shall we try latest?

.github/label-pr.yml
FrancescoBorzi: sure @Kitzunu , but can we exclude `/src/server/scripts` from applying "CORE" ?

Kitzunu: Dont know

.github/workflows/labeler.yml
Kitzunu: why not ^^

.github/label-pr.yml
Kitzunu: I have no idea if this will work. regEx is a regular expression

.github/label-pr.yml
```diff
@@ -0,0 +1,12 @@
+- regExp: ".*\\.sql+$"
+  labels: ["DB"]
+- regExp: "^(src\\server\\scripts\\.(cpp|h))$"
```
Kitzunu, line 3: @FrancescoBorzi

FrancescoBorzi: @Kitzunu is this supposed to exclude the scripts? where is the part of "including all cpp/h files" ?

Kitzunu: I just thought so too. Will look mord into it when Im not trying to fix it on my phone xd

FrancescoBorzi (approved):
@Kitzunu fine for me as long as you'll take care that this works as intended. It looks like there is no way to make sure this works until we merge it, so you can merge it!

Kitzunu:
I think there is a way, like I did woth CORE now.

Gonna test it and draw a conclution after

Kitzunu:
I cant figure out hwo to exclude it...

I have tried every guide there is....

Kitzunu:
@FrancescoBorzi It isn't possible to exclude paths, but what if we add a new label called Scripts that only apply to them, so it will have CORE & Script?

Kitzunu:
Now it works decently :)

PR #3575, opened 2020-10-15
fix(DB/Quest): The Black Box

Kitzunu:
## CHANGES PROPOSED:
-  Make The Black Box not a repeatable quest per this comment https://classic.wowhead.com/quest=708/the-black-box#comments:id=2940175:reply=1245192

## TESTS PERFORMED:
- Builds

## HOW TO TEST THE CHANGES:

Well, can't be tested since the quest actually doesn't exist in-game... But right should be right :)

## Target branch(es):
- [x] Master

 

Helias (approved):

Helias:
If the quest does not exist in game I think that it is not so important to change it xD 
but ok :yum:  :+1:

PR #3579, opened 2020-10-16
fix(Action): Labeler

Kitzunu:
## CHANGES PROPOSED:
-  Changing the action to a better one to hopefully get a working labeler with good exceptions


.github/labeler.yml
```diff
@@ -0,0 +1,23 @@
+DB:
+  - src/**/*.sql
+
+CORE:
+  - src/**/*.cpp
+  - src/**/*.h
+  - '!src/server/scripts/*'
```
FrancescoBorzi, line 7: are you sure it shouldn't be `'!src/server/scripts/**` ? (with double `*`)

.github/labeler.yml
```diff
@@ -0,0 +1,23 @@
...
+
+Script:
+  - src/server/scripts/*.cpp
+  - src/server/scripts/*.h
```
FrancescoBorzi, line 11: same remark as above

.github/workflows/labeler.yml
```diff
@@ -1,16 +1,12 @@
...
-name: labeler
-# Trigger the workflow on pull request events
+name: "Pull Request Labeler"
 on: [pull_request]
+
 jobs:
-  label:
-    runs-on: ubuntu-20.04
+  triage:
+    runs-on: ubuntu-latest
```
FrancescoBorzi, line 6: I think it's better to pin the version, so `ubuntu-20.04` in this case. So we have control on it.

.github/labeler.yml
Kitzunu: Yes

.github/labeler.yml
```diff
@@ -0,0 +1,22 @@
+DB:
+  - src/**/*.sql
+
+CORE:
+  - src/**/*.cpp
+  - src/**/*.h
```
FrancescoBorzi, line 6: shouldn't we exclude `src/server/scripts/**` ? or the category `Script` will make this excluded?

FrancescoBorzi (approved):

.github/labeler.yml
Kitzunu: actually you are right, it should be /**/*.cpp to include the subfolders too. My bad

Issue #3550, opened 2020-10-06
[Crash] Issue with .player learn command when the player already know the spell in console+SOAP


asfo:
Original thought:
I was thinking that after running several commands through soap this crashes the worldserver.

First update:
I found that it was related to the `.player learn` command and thought that the crash was due to a lot of commands running at the same time. 

New update:
The worldserver crashes after running the `.player learn` command when the player already knows that spell. So it is not `.send mail` only when I run `.player learn` commands, I attached that this looks like is related to `GetPlayer()` method.

##### CURRENT BEHAVIOUR:
When I run the `.player learn` command and the player already knows a spell, this crashes the worldserver. I have a way to fix this using web code, but should not happen if running commands directly to the console.

Here is the crash log:
```
set logging on
set debug timestamp
run

433790.411171 [Thread debugging using libthread_db enabled]
433790.411205 Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
433790.737338 [New Thread 0x7ffff49ed700 (LWP 65230)]
433790.745126 [New Thread 0x7ffff397f700 (LWP 65231)]
433790.793680 [New Thread 0x7ffff257f700 (LWP 65232)]
433809.295454 [New Thread 0x7fffb71ff700 (LWP 65281)]
433815.545749 [New Thread 0x7fffa13ff700 (LWP 65284)]
433815.546124 [New Thread 0x7fffa09fe700 (LWP 65285)]
433815.546483 [New Thread 0x7fff9f9ff700 (LWP 65286)]
433815.547661 [New Thread 0x7fff9f1fe700 (LWP 65287)]
433815.547703 [Thread 0x7fff9f9ff700 (LWP 65286) exited]
433815.548126 [New Thread 0x7fff9deab700 (LWP 65288)]
433815.548579 [New Thread 0x7fff9d1ff700 (LWP 65289)]
433815.548813 [New Thread 0x7fff9c7fe700 (LWP 65290)]

Thread 6 "worldserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffa13ff700 (LWP 65284)]
0x00000000004f85bc in WorldSession::GetPlayer() const ()
gdb.txt:4: Error in sourced command file:
Undefined command: "bt433518".  Try "help".
```

##### EXPECTED BLIZZLIKE BEHAVIOUR:
This should return "player already know that spell" in the console and not crash.

[...]

PR #3595, opened 2020-10-20
fix(Core/CS): Fixed learn command using null session when send over SOAP

peti446:
## CHANGES PROPOSED:
-  Add a check to see if session is invalid as it will be for soap calls

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3550

## TESTS PERFORMED:
- Created simple SOAP script executing .player learn (spellid of an already known spell)

## HOW TO TEST THE CHANGES:

1. Create Soap script
2. Execute .learn player playername spelli_already_known_spell

## Target branch(es):
- [x] Master

 

src/server/scripts/Commands/PlayerCommand.cpp
```diff
@@ -44,7 +44,7 @@ bool PlayerCommand::Learn(ChatHandler* handler, Player* targetPlayer, uint32 spe
 
     if (!allRanks && targetPlayer->HasSpell(spell))
     {
-        if (targetPlayer == handler->GetSession()->GetPlayer())
+        if (handler->GetSession() && targetPlayer == handler->GetSession()->GetPlayer())
```
Kitzunu, line 47: Just for my understanding... Why do we check GetSession twice?

peti446: The first check is out of scope, and it will not return the function all the time. So a new check is necessary. We could move the block up but IMO it would make it a bit harder to read.

FrancescoBorzi (approved):

src/server/scripts/Commands/PlayerCommand.cpp
Kitzunu: as long as it works im fine with it

Helias (approved):

Si1ker:
CodeFactor Failing after 2s — 1 issue found.

src/server/scripts/Commands/PlayerCommand.cpp#L11-L66
Complex Method

@FrancescoBorzi @peti446

FrancescoBorzi:
@Si1ker it's ok, sometimes it fails because of existing errors

Issue #3599, opened 2020-10-22
[ICC] Light's Hope Muradin/Saurfang Wrong Models


Voxstrasza:
##### CURRENT BEHAVIOUR:
Currently, Muradin Bronzebeard has the wrong model at Light's Hope.  He uses the High Overlord Saurfang model.  If you try to manually change the model to the correct one (.npc set model 30508 in-game) it will set Muradin to the correct model, but it will also set Saurfang's model to Muradin.  It seems that because of the changes made to the scripted roleplay event, the NPC ID is shared.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
Muradin and Saurfang should have separate models and separate NPC IDs, but still retain their respective scripted events (it's different for alliance and horde)

##### STEPS TO REPRODUCE THE PROBLEM:
1.  Make alliance character
2.  Go into ICC, look next to Tirion
3.  Click Muradin Bronzebeard, .npc set model 30508 to change to correct model
4. Make horde character
5. Saurfang is now Muradin?

##### EXTRA NOTES:
N/A

##### AC HASH/COMMIT:
d5bb6ba1c461

##### OPERATING SYSTEM:

Win10

##### MODULES:
N/A

##### OTHER CUSTOMIZATIONS:

Custom boss health pools

Akimboss:
Yo, first of all just for information using npc set model ig won't fix the problem because what you actually did with this command is updating model of that NPC with DB GUID which means Muradin and saurfang will have same ID that you added ... in other words this will only work on world mobs where this mob isn't in raid or something, and for the probleme here's the fix for it you can test it #3600

Voxstrasza:
> 
> 
> Yo, first of all just for information using npc set model ig won't fix the problem because what you actually did with this command is updating model of that NPC with DB GUID which means Muradin and saurfang will have same ID that you added ... in other words this will only work on world mobs where this mob isn't in raid or something, and for the probleme here's the fix for it you can test it #3600

ah okay that makes sense

Voxstrasza:
Closing, fixed by #3600 - waiting for it to be merged

FrancescoBorzi:
> Closing, fixed by #3600 - waiting for it to be merged

please don't close issues even if there are already PRs for them. The issue will be closed automatically once the PR is merged to master. We should not close an issue as long as the bug is present on master.

PR #3600, opened 2020-10-22
[FIX]- [ICC] Light's Hope Muradin/Saurfang Wrong Models #3599

Akimboss:
Hello, i've added the fix for the issue #3599 , the problem will be fixed for both Muradin and Saurfang ( adding new modelID ) thank you.
PS: when you apply the SQL if you still see the wrong modelid, DON'T panic !! just use .reload creature_template 32315 and restart and everything is fine only use that if you still see old model (sometimes needs to reload creature_template ig using comand)


Voxstrasza:
tested - fixes muradin but changes saurfang's model to 27905 which isn't the ICC model

FrancescoBorzi (approved):

iThorgrim-Hub (approved):

PR #3623, opened 2020-10-25
fix(DB/Zone):improved Dun Morogh

Knindzagxg:
Zone now fully updated, since some creatures in zone was missing SmartAI.

To test it:
Apply SQL
Restart your server

.npc add ID
or
.tele to zone and run / test around.


aradep:
Tested, everything works but the ranges for Juvenile Snow Leopard(1199), Young Black Bear(1128), and Snow Leopard(1201) seem a bit low, most of the time SAI doesn't have a chance to trigger.

FrancescoBorzi:
@Knindzagxg

Knindzagxg:
Yeah, thats on purpose @aradep 
Most times they don't even cast on Retail, but you notice "daze" sometimes only when people pull multiple NPCs.
Otherwise, players would be dazed 24/7 in low level zones xP

FrancescoBorzi (approved):

FrancescoBorzi:
> Yeah, thats on purpose @aradep
> Most times they don't even cast on Retail, but you notice "daze" sometimes only when people pull multiple NPCs.
> Otherwise, players would be dazed 24/7 in low level zones xP

sounds reasonable to me. Unless @aradep has something against, for me it's good to be merged

Si1ker (approved):

aradep:
@Knindzagxg should these creatures actually be "casting" daze? there is already a chance-on-hit daze effect that all mobs have. https://wow.gamepedia.com/Daze

Knindzagxg:
Yeah for some reason they have daze spell that proc super super rarely.
Regular daze is hit from behind and its core based, but there is also this one xP

PR #3635, opened 2020-10-29
Core/Spells: castSpell and castcustomspell functions will return cast…

sanctum32:
… result

## CHANGES PROPOSED:
-  This change will change spell cast functions to return cast result. It can be used at scenarios like... Where spell cast follows with emote. For example Bronjahm (first Forge of Souls boss), he spams Corrupt Soul and when he ends his mana, he still says cast text but spell cast actually is failing. In this case we could check spell cast result and then let him to output his text.
This also can be used to test spell scripts if secondary effect casts are successfuly performed

<h2><i>This change will not effect any currently scripted continent by default!</i></h2>

## ISSUES ADDRESSED:
- None

## TESTS PERFORMED:
- Tested mainly with bronjahm. 
-

## HOW TO TEST THE CHANGES:

This can be only tested by scripters/programmers because none of current continent is effected by this

replace in Bronjahm script (https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/Northrend/FrozenHalls/ForgeOfSouls/boss_bronjahm.cpp#L152)
```c
    Talk(SAY_CORRUPT_SOUL);
    me->CastSpell(target, SPELL_CORRUPT_SOUL, false);
```
with
```c
if (me->CastSpell(target, SPELL_CORRUPT_SOUL, false) == SPELL_CAST_OK) 
{
    Talk(SAY_CORRUPT_SOUL);
}
```
and wait till he cast. Also make look if he uses text when he is out of mana.

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 

FrancescoBorzi (approved):

Si1ker:
@FrancescoBorzi Are you sure? without further testing or additional feedback?
Don't want to have more broken stuff for changes that "won't affect"

sanctum32:
@Si1ker at most this was changed to return value, previously it was void function, thats where i make that statement that this doesn;t effects current behavior. Rest things are pretty well described in PR description

src/server/game/Spells/Spell.cpp
```diff
@@ -3381,26 +3384,30 @@ void Spell::prepare(SpellCastTargets const* targets, AuraEffect const* triggered
...
-        SendCastResult(SPELL_FAILED_SPELL_IN_PROGRESS);
+        SendCastResult(SPELL_FAILED_SPELL_UNAVAILABLE);
         finish(false);
-        return;
+        return SPELL_FAILED_SPELL_UNAVAILABLE;
     }
 
-    if (DisableMgr::IsDisabledFor(DISABLE_TYPE_SPELL, m_spellInfo->Id, m_caster))
+    //Prevent casting at cast another spell (ServerSide check)
+    if (!(_triggeredCastFlags & TRIGGERED_IGNORE_CAST_IN_PROGRESS) && m_caster->IsNonMeleeSpellCast(false, true, true, m_spellInfo->Id == 75) && m_cast_count)
```
FrancescoBorzi, line 3403: just wondering why you changed the order of the if checks

FrancescoBorzi:
@Si1ker the point is that this PR is simply adding a return value for functions that were previously returning nothing (void).

This means that whatever code currently uses these functions, it's using them WITHOUT using their return values (because they don't return anything). This PR makes them return something, and this something will just be ignored whenever they are already used. This PR just gives the possibility to NEW code to use these return values.

So logically it can't break anything.

Of course, I understand your concern: you can NEVER be 100% sure that a code change cannot introduce bugs. But there is no specific testing that we can do for this PR, so we can't ask testers to test anything specific. I trust @sanctum32 to have tested this already:

> Tested mainly with bronjahm.

src/server/game/Spells/Spell.cpp
sanctum32: Base case for this change is to avoid necessary calls. Why you should perform check above if spell is disabled anyways and won't get executed anyways

FrancescoBorzi:
@Si1ker check discord please

Issue #3602, opened 2020-10-23
(Event) Hallow's End - Rachelle Gothena/mailbox location


locus313:
##### CURRENT BEHAVIOUR:

- Rachelle Gothena is spawning in the same location as a mailbox out side the auction house

![image](https://user-images.githubusercontent.com/4015312/96959049-0249ec80-14b4-11eb-80dc-a76e0a53b147.png)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

- Still trying to find the exact information on this, but dont think there should actually be a mailbox right outside of the auction house. From what I could find the spawn location for Rachelle Gothena looks about right:
https://wowgaming.altervista.org/aowow/?npc=15354

##### STEPS TO REPRODUCE THE PROBLEM:

1. Go to the auction house in Orgrimmar
2. See that Rachelle Gothena is spawned right in the mailbox that is right outside the auction house

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/commit/32da0eeb9e39b362eae50944a5b16069301403c6

##### OPERATING SYSTEM:

Centos7

##### MODULES:

mod-ah-bot
mod-autobalance
mod-cfbg
mod-custom-login
mod-individual-xp
mod-transmog

PR #3645, opened 2020-11-03
fix(DB/Object) Hide mailbox during hallow's end event

locus313:
## CHANGES PROPOSED:
-  Hide mailbox during hallow's end event

Before pr:

image

After pr:

image

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3602

## TESTS PERFORMED:
- Apply sql and restart server
- Start hollow's end event
- Check that mailbox disappears
- Stop Hollow's end event
- Check that mailbox reappears

## HOW TO TEST THE CHANGES:

- Apply sql and restart server
- .event start 12
- Check that mailbox disappears
- .event stop 12
- Check that mailbox reappears

## Target branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1604442802893059360.sql
+++ b/data/sql/updates/pending_db_world/rev_1604442802893059360.sql
@@ -1,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1604442802893059360');
+
+DELETE FROM `game_event_gameobject` WHERE `guid` IN (150747);
+INSERT INTO `game_event_gameobject` (`eventEntry`, `guid`) VALUES (-12,150747);


Meltie2013:
@locus313 Why would you hide the mailbox during the event? This was never done on retail servers.
It's just better to fix the position of the mailbox instead.

Helias:
agree with @Meltie2013, I suggest to move the NPC to a different location

Kitzunu:
Both solutions are hacky since we dont have the required sniffs or proof.

So the question lands in, what is the least hacky

Si1ker:
"So the question lands in, what is the least hacky"
Move the npc.

Kitzunu:
I'd say remove the mailbox because the only proof we have is that the npc is correct

Gierlappen:
I don't understand why we don't just move the npc as i suggested in my PR and if somebody find a better or blizzlike solution we can just change it...

To move the Mailbox or hide it during the event is just more incorrect. I am the first to want correctness but if there is no source what is the blizzlike behaviour we still can improve things what is good to get people for this project.

As a contributor i wan't to say that i would contribute much more to this project if we go off this "if it's not blizzlike we accept the crap we have before we improve it" line. This is just demotivating.

I love this project and will continue contributing but i hab to say this.

Meltie2013:
@Kitzunu Hiding the mailbox would be pointless and not blizzlike, cause the mailbox is in the right location. Plus you are removing the ability for players to visit the mailbox after auction house action as well by hiding it.
NPC is not in the right location and needs to be moved to the right of it. You guys making a bigger deal out it then it needs to be.

locus313:
I will just close this pr since it sounds like more people are in favor of moving the npc.

Kitzunu:
Improvements yes.
The only reason why I'd prefer this over moving the NPC is that the screenshot from the previous or. but then, it is not my preference that counts

FrancescoBorzi:
> I don't understand why we don't just move the npc as i suggested in my PR and if somebody find a better or blizzlike solution we can just change it...
> 
> To move the Mailbox or hide it during the event is just more incorrect. I am the first to want correctness but if there is no source what is the blizzlike behaviour we still can improve things what is good to get people for this project.
> 
> As a contributor i wan't to say that i would contribute much more to this project if we go off this "if it's not blizzlike we accept the crap we have before we improve it" line. This is just demotivating.
> 
> I love this project and will continue contributing but i had to say this.

I agree. We should be flexible. Also, whatever PR we merge, we can always make it better later.

PR #3714, opened 2020-11-16
refactor(Apps/Defines): fix codefactor warning

Kitzunu:
According to Codefactor 
(https://www.codefactor.io/repository/github/azerothcore/azerothcore-wotlk/file/master/apps/bash_shared/defines.sh)

using backticks is deprecated and should be replaced with $()


Helias (approved):

apps/compiler/includes/functions.sh
```diff
@@ -46,7 +46,7 @@ function comp_configure() {
 
 
 function comp_compile() {
-  [ $MTHREADS == 0 ] && MTHREADS=`grep -c ^processor /proc/cpuinfo` && MTHREADS=$(($MTHREADS + 2))
+  [ $MTHREADS == 0 ] && MTHREADS=$(grep -c ^processor /proc/cpuinfo) && MTHREADS=$(($MTHREADS + 2))
```
Helias, line 49: not sure if are needed the double quote outside $(grep...) or in $(git...)

Kitzunu: ?

FrancescoBorzi (approved):

apps/compiler/includes/functions.sh
Helias: example
```bash
_cur_branch="$(git --git-dir="$J_PATH/.git/" --work-tree="$J_PATH/" rev-parse --abbrev-ref HEAD)"
```

or something similar

Kitzunu: Not according to codefactor, but I never programmed in whatever language this is

PR #3716, opened 2020-11-16
fix(apps/scripts): Add quotes for directory paths containing empty spaces

locus313:
## CHANGES PROPOSED:
-  add quotes to variables used for the directory paths 

Based on orginal pr from BarbzYHOOL:
https://github.com/azerothcore/azerothcore-wotlk/pull/2755
## ISSUES ADDRESSED:
- Closes 

## TESTS PERFORMED:
- Used the db_assembler.sh inside a path containing empty spaces
- Tested options 6-8
- Verified that the sql files were assembled correctly

## HOW TO TEST THE CHANGES:

- Use the db_assembler.sh inside a path containing empty spaces
- Test options 6-8

## KNOWN ISSUES AND TODO LIST:

- [ ] include.sh for modules with sql files will need to be updated to put a quote round the path variable
- [ ] conf/conf.sh.dist for modules with sql files will need to be updated to put a quote round the path variable

## Target branch(es):
- [x] Master

 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #3721, opened 2020-11-17
feat(ci): Add apps directory to bash label

locus313:
## CHANGES PROPOSED:
-  add apps directory to bash label

## ISSUES ADDRESSED:

Noticed the labeler remove the bash label on this pr:
https://github.com/azerothcore/azerothcore-wotlk/pull/3716

## Target branch(es):
- [x] Master

 

FrancescoBorzi (approved):

.github/labeler.yml
```diff
@@ -13,6 +13,7 @@ Documentation:
 
 Bash:
   - ./*.sh
+  - apps/**/*.sh
```
Kitzunu, line 16:
```suggestion
  - apps/**/*.sh
  - conf/**/*.sh
  - deps/**/*.sh
  - modules/**/*.sh
```

Kitzunu (changes requested):
Looks like we need to specify them all sadly

.github/labeler.yml
locus313: Your right, looks that way. Thanks @Kitzunu

Kitzunu (approved):

.github/labeler.yml
locus313: @Kitzunu actually I think we might still need the 
`./*.sh`
otherwise it wont pickup the acore.sh and install.sh in the root. Do you agree?

Kitzunu: I tested this on a seperate repo and ./ seems to be as broken as the any: and all: sadly

locus313: Gotcha, sounds good. I have added it again.

Si1ker (approved):

PR #3737, opened 2020-11-21
refactor(Core/SpellMgr): Part 2

Kitzunu:
I promised I'd start on it as soon as the other PR was merged :)))))

## CHANGES PROPOSED:
-  
-  

## ISSUES ADDRESSED:
- Closes 

## TESTS PERFORMED:
-
-

- [ ]
- [ ] 

## Target branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Spells/SpellMgr.cpp
+++ b/src/server/game/Spells/SpellMgr.cpp
@@ -3310,7 +3310,7 @@ void SpellMgr::LoadSpellInfoCorrections()
         63137  // Force Cast (HACK: Target shouldn't be changed; summon position should be untied from spell destination)
         }, [](SpellEntry* spellInfo)
     {
-        spellInfo->EffectImplicitTargetA[EFFECT_0] = TARGET_DEST_DB;
+        spellInfo->Effects[EFFECT_0].TargetA = SpellImplicitTargetInfo(TARGET_DEST_DB);
     });
 
     ApplySpellFix({
@@ -3366,7 +3366,7 @@ void SpellMgr::LoadSpellInfoCorrections()
     // Howl of Azgalor
     ApplySpellFix({ 31344 }, [](SpellEntry* spellInfo)
     {
-        spellInfo->EffectRadiusIndex[0] = EFFECT_RADIUS_100_YARDS; // 100yards instead of 50000?!
+        spellInfo->Effects[0].RadiusEntry = sSpellRadiusStore.LookupEntry(EFFECT_RADIUS_100_YARDS); // 100yards instead of 50000?!
     });
 
     ApplySpellFix({
@@ -3514,13 +3514,13 @@ void SpellMgr::LoadSpellInfoCorrections()
     // Desecration Arm - 36 instead of 37 - typo?
     ApplySpellFix({ 29809 }, [](SpellEntry* spellInfo)
     {
-        spellInfo->EffectRadiusIndex[0] = 37;
+        spellInfo->Effects[0].RadiusEntry = sSpellRadiusStore.LookupEntry(37);
     });
 
     // Sic'em
     ApplySpellFix({ 42767 }, [](SpellEntry* spellInfo)
     {
-        spellInfo->EffectImplicitTargetA[0] = TARGET_UNIT_NEARBY_ENTRY;
+        spellInfo->Effects[0].TargetA = SpellImplicitTargetInfo(TARGET_UNIT_NEARBY_ENTRY);
     });
 
     // Master Shapeshifter: missing stance data for forms other than bear - bear version has correct data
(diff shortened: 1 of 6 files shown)


src/server/game/Spells/SpellMgr.cpp
```diff
@@ -2679,7 +2679,7 @@ void SpellMgr::UnloadSpellInfoStore()
 
 void SpellMgr::UnloadSpellInfoImplicitTargetConditionLists()
 {
-    for (uint32 i = 0; i < mSpellInfoMap.size(); ++i)
+    for (uint32 i = 0; i < GetSpellInfoStoreSize(); ++i)
```
FrancescoBorzi, line 2682: you can use just `auto` instead of `uint32` especially in cases like this

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -3261,26 +3261,26 @@ void SpellMgr::LoadSpellCustomAttr()
...
     sLog->outString();
 }
 
-inline void ApplySpellFix(std::initializer_list<uint32> spellIds, void(*fix)(SpellEntry*))
+inline void ApplySpellFix(std::initializer_list<uint32> spellIds, void(*fix)(SpellInfo*))
 {
     for (uint32 spellId : spellIds)
     {
-        SpellEntry const* spellInfo = (SpellEntry*)sSpellStore.LookupEntry(spellId);
+        SpellInfo const* spellInfo = (SpellInfo*)sSpellStore.LookupEntry(spellId);
```
FrancescoBorzi, line 3272: also here, it's pointless to repeat `SpellInfo` twice in a row, we can replace the first `SpellInfo` with `auto`

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -3261,26 +3261,26 @@ void SpellMgr::LoadSpellCustomAttr()
...
         if (!spellInfo)
         {
             sLog->outErrorDb("Spell info correction specified for non-existing spell %u", spellId);
             continue;
         }
 
-        fix(const_cast<SpellEntry*>(spellInfo));
+        fix(const_cast<SpellInfo*>(spellInfo));
```
FrancescoBorzi, line 3279: just curious: why using a `const_cast` instead of `static_cast` here?

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -7177,32 +7177,32 @@ void SpellMgr::LoadDbcDataCorrections()
         spellInfo->AttributesEx7 |= SPELL_ATTR7_INTERRUPT_ONLY_NONPLAYER;
     });
 
-    for (uint32 i = 0; i < sSpellStore.GetNumRows(); ++i)
+    for (uint32 i = 0; i < GetSpellInfoStoreSize(); ++i)
```
FrancescoBorzi, line 7180: `uint32` should just be `auto`

src/server/game/Spells/SpellMgr.cpp
```diff
@@ -7177,32 +7177,32 @@ void SpellMgr::LoadDbcDataCorrections()
...
     {
-        SpellEntry* spellInfo = (SpellEntry*)sSpellStore.LookupEntry(i);
+        SpellInfo* spellInfo = mSpellInfoMap[i];
```
FrancescoBorzi, line 7182: if (and only if) `mSpellInfoMap[i]` is already a `SpellInfo*`, then we can use `auto` instead of `SpellInfo`

src/server/game/Spells/SpellMgr.cpp
Kitzunu: ask trinity Q.Q

src/server/game/Spells/SpellMgr.cpp
Kitzunu: so "auto* spellinfo"?

src/server/game/Spells/SpellMgr.cpp
Viste:
````
`const_cast` can be used to remove or add `const` to a variable; no other C++ cast is capable of removing it (not even reinterpret_cast). It is important to note that modifying a formerly `const` value is only undefined if the original variable is const; if you use it to take the `const` off a reference to something that wasn't declared with `const`, it is safe. This can be useful when overloading member functions based on `const`, for instance. It can also be used to add const to an object, such as to call a member function overload.```

src/server/game/Spells/SpellMgr.cpp
Viste: auto& spellinfo

Kitzunu: Thanks Viste <3

src/server/game/Spells/SpellMgr.cpp
Kitzunu: Build error with that

Kitzunu:
Okay we got two issues:

1. This doesn't work at all.
2. the "if" statement in the bottom of SpellMgr.cpp gets error

Viste:
> Okay we got two issues:
> 
>     1. This doesn't work at all. (as you can tell from the DBErrors)
> 
>     2. the "if" statement in the bottom of SpellMgr.cpp gets error
there must be more work not only at spellmgr :))

PR #3746, opened 2020-11-22
fix(DB): failure due to missing semicolumn

FrancescoBorzi:


Si1ker (approved):
oops !

Kitzunu:
How does that even pass checks

FrancescoBorzi:
> How does that even pass checks

Good question. It does because one query without `;` is valid, multiple queries aren't.

FrancescoBorzi:
@locus313 maybe we can improve the `.github/workflows/check_pending_sql.yml` workflow and add this check

locus313:
@FrancescoBorzi just submitted a pr to add the check: 
https://github.com/azerothcore/azerothcore-wotlk/pull/3749

PR #3749, opened 2020-11-23
feat(CI): add check for semicolon at end of pending sql update file

locus313:
Co-Authored-By: Kitzunu <24550914+Kitzunu@users.noreply.github.com>

## CHANGES PROPOSED:
-  Add a check in the pending sql workflow to find if there is a semicolon at the end of the end file

## ISSUES ADDRESSED:
- Closes none

## TESTS PERFORMED:
test with a file without a semicolon:
image

test with a file with a semicolon:
image

## Target branch(es):
- [x] Master

 

Kitzunu:
Only question I have is how would this behave if the user for example added blankspaces at the end of the file while still ending with a semicolon on the last line of text.

Would it fail or continue?

locus313:
I actually tested that. it would continue. Its for this exact reason why I used `-c 2` instead of `-c 1`. But for example if someone add's two whitespaces at the end of the file, then it would fail.

FrancescoBorzi (approved):

Kitzunu:
Yea I tried to research if we could ignore blanks but found nothing

Issue #3712, opened 2020-11-16
Core: Rested bonus persists outside of inn/capitals [$15]


Kitzunu:
Reported by @Maurowin 
##### CURRENT BEHAVIOUR:

Currently if you exit the inn or a capital city you will still store up the rest.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

When you leave the inn or capital city you should no longer store up rest.
When you are resting it is displayed like this on your character image
![image](https://user-images.githubusercontent.com/24550914/99258671-35fd0580-2819-11eb-8737-5c49b335963d.png)
https://wowwiki.fandom.com/wiki/Rest?oldid=2331121

This can clearly be shown here https://www.youtube.com/watch?v=q9ufWJOlCrQ

##### STEPS TO REPRODUCE THE PROBLEM:

1. Walk into an inn / Capital
2. Walk out and see that the rest will continue to store up

##### OPERATING SYSTEM:

W10

Maurowin:
Confirm. More precision, you will continue to store up during few seconds like +-30sec (random). The "zzz" mark staying too few seconds.

FrancescoBorzi:
was this bug always there?

Kitzunu:
Yea it was reported over a year I go I think but I closed it back then https://github.com/azerothcore/azerothcore-wotlk/issues/671

Helias:
@Kitzunu claim the bounty if you didn't

BarbzYHOOL:
Alright, i'll show you how to debug an issue...

so... i thought maybe there is a deep bug (as I told you @Kitzunu in private message) and i did more tests in the Ratchet Inn

Basically, the areatrigger for the Inn is 30 radius. (prob the same on every inn?)
Actually, in the code, it doesn't remove the "zzzz" when you leave the Inn's areatrigger radius, it removes the "zzzz" you leave the "area" as in `UpdateArea()` and THEN it checks if you are inside the inn's radius. I don't have any idea what is considered an Area or not sadly.
So when I leave the Inn, the UpdateArea() function triggers, but I'm still INSIDE the inn's radius.

So basically I get this when it doesn't remove the "zzzz":
```
trigger radius: 30.000000
distance to inn: 27.280859
```

And when it correctly removes the "zzzz" (example, you teleport out of the zone or run very fast), of course "distance to inn" is > 30.

[...]

PR #3780, opened 2020-11-26
fix(Core/Game): Rest & Inn Improvements

Meltie2013:
## CHANGES PROPOSED:
-  Implemented missing Rest Flag Masks
-  Improved rest flag checks for Inns, Cities & Sanctuaries
-  Remove nasty hack fixes that just made no sense
-  Attempt to improve area trigger radius checks

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3712

## TESTS PERFORMED:
- Watch the video here by @Maurowin: https://youtu.be/BwmdouFTv9Y

## HOW TO TEST THE CHANGES:
- Login a newly created character or pre-existing character (race doesn't matter)
- Head to any Inn, City or Sanctuary zone
- Check if player gets rest flag on entering the area
- Leave the zone and see if flag is updated

- Radius around Inns are to large and require further updates are mentioned below in the feedback.

## Target branch(es):
- [x] Master

## Feedback
To fix radius issues around Inns, more updates are required on Position functions across the core. Math for it is incorrect and most of the time does nothing. Another PR will need to be opened for that separately. Changes made thus far are working as intended according to @Maurowin and the test they have done.

 

Maurowin:
https://youtu.be/BwmdouFTv9YSounds pretty good to me!
However, then remains to be confirmed, can the two factions take advantage of the rest in the hostels of the opposing faction?

If yes, it does not work.

The zone of rest is a little too wide and that it sees particularly in the inn of "Sentinel Hills" at the beginning and especially at the end of the video. I get the rest bonus while I'm 3 to 4 meters outside the hostel.

Translated with www.DeepL.com/Translator (free version)

Meltie2013:
@Maurowin The wide radius stand point, it seems like a different issue. I can add a check to make sure the rest flag is removed when leaving the Inn.

Maurowin:
@Meltie2013 ooh, so we should open a new issue ?

Meltie2013:
@Maurowin It's a possibility till it can be confirmed.

Maurowin:
@Meltie2013 concerning this point : can the two factions take advantage of the rest in the Inn of the opposing faction?
u know something ? 
To be confirmed but I think that both factions can benefit from it

Meltie2013:
@Maurowin Alliance and Horde should only benefit from their own Inn's. From what I can remember back in the day of retail, this should be a big no.

FALL1N1:
Inns are faction-based unless contested/neutral (e.g dalaran / shattrath / booty bay / gadgetzan...etc)

Meltie2013:
@FALL1N1 That is also true, that been fixed in this PR as well. No checks existed for that.

Maurowin:
The radius still to large and I have the feeling that the exit of the inns zone takes time to update. In the capitals there is no problem.
https://youtu.be/MK_7RnVZTjc

Maurowin:
@FrancescoBorzi u can see my videos of my tests.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -3023,20 +3032,22 @@ void Player::SetInWater(bool apply)
...
     if (!trigger || GetMapId() != trigger->map)
         return false;
 
-    if (trigger->radius > 0.f)
+    if (trigger->radius > 0)
     {
         // if we have radius check it
         float dist = GetDistance(trigger->x, trigger->y, trigger->z);
-        if (dist > trigger->radius)
+        if (dist > trigger->radius + delta)
```
BarbzYHOOL, line 3044: what the fuck is that???

so to fix it, "they" added artificial 5 meters???????? this is the "correct solution" ?? this is as dirty as before

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1776,6 +1777,14 @@ void Player::Update(uint32 p_time)
     {
         if (p_time >= m_zoneUpdateTimer)
         {
+            // On zone update tick check if we are still in an inn if we are supposed to be in one
```
BarbzYHOOL, line 1780: really no better way can be found??

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7687,18 +7678,11 @@ void Player::UpdateArea(uint32 newArea)
...
-            RemoveByteFlag(UNIT_FIELD_BYTES_2, 1, UNIT_BYTE2_FLAG_FFA_PVP);
-    }
-    else if (!(areaFlags & AREA_FLAG_CAPITAL))
-    {
-        AreaTrigger const* atEntry = sObjectMgr->GetAreaTrigger(GetInnTriggerId());
-        if (!atEntry || !IsInAreaTriggerRadius(atEntry))
-            RemoveRestState();
-    }
+    uint32 const areaRestFlag = (GetTeamId() == TEAM_ALLIANCE) ? AREA_FLAG_REST_ZONE_ALLIANCE : AREA_FLAG_REST_ZONE_HORDE;
+    if (area && area->flags & areaRestFlag)
```
BarbzYHOOL, line 7682: can't you make use of the existing `area->IsInn(GetTeamId(true))` instead?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7658,27 +7669,7 @@ void Player::UpdateArea(uint32 newArea)
...
-    bool isInn = area->IsInn(GetTeamId(true));
-    if (zone)
-    {
-        areaFlags |= zone->flags;
-        isSanctuary |= zone->IsSanctuary();
-        isInn |= zone->IsInn(GetTeamId(true));
-    }
-
-    // previously this was in UpdateZone (but after UpdateArea) so nothing will break
-    if (isSanctuary)    // in sanctuary
```
BarbzYHOOL, line 7681: can u eventually tell testers how to verify these things are still working somehow ingame after the rework?

BarbzYHOOL:
@Maurowin can you test to set the worldserver.conf server type to FFAPVP, then open two clients. Your characters should have the "gurubashi arena pvp icon". Then you go with both inside an inn and you see if the icon disappears and if you can't attack each other.

Please

BarbzYHOOL:
> @Meltie2013 ooh, so we should open a new issue ?

no it's not an issue, it's the inn radius in the DBC, it's blizzlike and it's just probably made to ease the calculation and the world building creation for the blizzard's devs.

src/server/game/Entities/Player/Player.cpp
BarbzYHOOL: ok rochet2 the boss enlightened me on why the client can't be trusted when exiting the inn, hence why there is prob no other efficient way to check when out of the inn. Client could never send the "not in inn anymore" and so players would cheat the resting system forever

so i guess no better way

BarbzYHOOL:
> The radius still to large and I have the feeling that the exit of the inns zone takes time to update. In the capitals there is no problem.
> https://youtu.be/MK_7RnVZTjc

that's because when you leave a capital you exactly leave the zone. When you leave an inn, you're not leaving a zone per se so there is not a perfect trigger (hence why the +5 yards and also the tiny delay)

Maurowin:
> 
> 
> @Maurowin can you test to set the worldserver.conf server type to FFAPVP, then open two clients. Your characters should have the "gurubashi arena pvp icon". Then you go with both inside an inn and you see if the icon disappears and if you can't attack each other.
> 
> Also can you test if logging out in capitals is instant?
> 
> Please

Will test this tomorrow

Meltie2013:
@Maurowin I added additional check to update zone when player's position is updated. Make sure to pull this before additional testing is made.

src/server/game/Entities/Player/Player.cpp
Meltie2013: These checks are handled else where in the core, so they were removed.

src/server/game/Entities/Player/Player.cpp
Meltie2013: This is needed, else you would be back at square one of the rest flags not updating again.

src/server/game/Entities/Player/Player.cpp
Meltie2013: When `GetTeamId` is called, its looking for the team id.  Adding additional true or false really doesn't do anything but swapping the positions of true and false;

Better way of handling it.
```cpp
TeamId GetTeamId() const { return m_team == ALLIANCE ? TEAM_ALLIANCE : TEAM_HORDE; }
```

BarbzYHOOL: i mean that "area->IsInn(GetTeamId(true))" is already pretty much:
```cpp
    uint32 const areaRestFlag = (GetTeamId() == TEAM_ALLIANCE) ? AREA_FLAG_REST_ZONE_ALLIANCE : AREA_FLAG_REST_ZONE_HORDE;
    if (area && area->flags & areaRestFlag)
```

Maurowin:
There is no evolution and I even have the impression that it is necessary to go further when we leave the inn. 

The update " going out" must have a big problem.
https://youtu.be/crYCMO9YL9o

BarbzYHOOL:
ok so we need to revert https://github.com/azerothcore/azerothcore-wotlk/pull/3780/commits/f3af4725ec970c06bdaaeb5942746a6b7f99a221 ??

Maurowin:
The previous resquest sweater was slightly better.

Can we get inspiration from the code that manages the ffa mode when we go in and out of a hostel? Because this one works perfectly and the resquest should work exactly like it.

WoWScrnShot_112920_215008
WoWScrnShot_112920_215012

Meltie2013:
@BarbzYHOOL This commit https://github.com/azerothcore/azerothcore-wotlk/commit/f3af4725ec970c06bdaaeb5942746a6b7f99a221 will be reverted.

Maurowin:
It's better !

So like we talking in Discord, updateStatus dont work as expected.

Great job !

Helias (approved):

BarbzYHOOL:
@Maurowin ok we gonna merge this, it's much better than before ingame (even though pixel perfect) and i talked with meltie and rochet, this seems like the best way to proceed and the most complete + a nice refactor of the ugly code we had

I put the "to be merged" label, if anyone doesn't agree, speak loudly!

Si1ker (approved):

PR #3798, opened 2020-11-28
fix(DB/spell_custom_attr):  Debuffs incorrectly applied as buffs

Voxstrasza:
## CHANGES PROPOSED:
-  Change 'Phase Punch' to correctly be applied as a debuff on Algalon 10 and 25 in Ulduar
-  Change 'Death Coil' to correctly be applied as a debuff on High King Maulgar in Gruul's Lair

## ISSUES ADDRESSED:
- #3149 placed spells to be modified as positive/negative within the database; but the original PR set spells that are supposed to be negative as positive. 

## TESTS PERFORMED:
- Deleting the row in the table set 'Phase Punch' as a debuff
- Adding row for set 'Death Coil' as debuff

## HOW TO TEST THE CHANGES:
- Apply SQL, restart server, create character
- Run Ulduar 10 or 25 and pull Algalon, check if Phase Punch is a debuff or not
- Run Gruul's Lair and wait for Death Coil to be cast on you

## Target branch(es):
- [x] Master


FrancescoBorzi:
@locus313 looks like we found a case where the `;` check fails.

The issue is the comment after the last `;` in this case, which is totally valid, but the check doesn't realise it.

Voxstrasza:
> 
> 
> @locus313 looks like we found a case where the `;` check fails.
> 
> The issue is the comment after the last `;` in this case, which is totally valid, but the check doesn't realise it.

In order to avoid the failed check, I can remove the comments when the PR is complete - but consider this a WIP as there are more debuffs listed as buffs that need to be checked out and removed.

Voxstrasza:
After reviewing the original PR, it seems that Phase Punch is the only one incorrectly listed - updated this PR and is ready for review

locus313:
@FrancescoBorzi Thanks for pointing this out. I got a solution for this and submitted a pr to update the check. https://github.com/azerothcore/azerothcore-wotlk/pull/3799

Si1ker:
@Voxstrasza  This one still appears as buff instead of debuff so people can right-click on it to remove it from themselves.
.tele gruul

image

data/sql/updates/pending_db_world/rev_1606594157029530500.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1606594157029530500');
+
+-- Correct spell Algalon the Observer, 10 & 25 player 'Phase Punch' to be applied as negative buff
+
+DELETE FROM `spell_custom_attr` WHERE `spell_id` = 64412;
+
+-- Correct spell High King Maulgar, 'Death Coil' to be applied as negative buff
+
+INSERT INTO `spell_custom_attr` (`spell_id`, `attributes`) VALUES ('33130', '12288');
```
FrancescoBorzi, line 9: always use a `DELETE` before an `INSERT` (this makes your query re-executable), also do not use quotes for numeric values

FrancescoBorzi (changes requested):

data/sql/updates/pending_db_world/rev_1606594157029530500.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1606594157029530500');
+
+-- Correct spell Algalon the Observer, 10 & 25 player 'Phase Punch' to be applied as negative buff
+
+DELETE FROM `spell_custom_attr` WHERE `spell_id` = 64412;
+
+-- Correct spell High King Maulgar, 'Death Coil' to be applied as negative buff
+
+INSERT INTO `spell_custom_attr` (`spell_id`, `attributes`) VALUES (33130,12288);
```
FrancescoBorzi, line 9: please add a `DELETE` of `33130` before its `INSERT`

FrancescoBorzi (changes requested):

FrancescoBorzi (approved):

Voxstrasza:
@FrancescoBorzi is this ok for merge?

FrancescoBorzi:
the code looks ok but I haven't tested myself

 @Si1ker so have you tested this? should it be merged?

Si1ker (approved):

Voxstrasza:
@Si1ker @FrancescoBorzi good for merge?  there might be more debuffs/buffs out there that are messed up but i can fix that in another pr when it's found

FrancescoBorzi:
I haven't checked this myself, tagging @Si1ker

Issue #3820, opened 2020-12-01
Apply clang-tidy modernize-* checks


FrancescoBorzi:
[clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/list.html) is

> a clang-based C++ “linter” tool. Its purpose is to provide an extensible framework for diagnosing and fixing typical programming errors, like style violations, interface misuse, or bugs that can be deduced via static analysis. 

Among all its [available checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html), there is a collection of `modernize-*` checks that offer automated fixes. It would be nice to apply them to our codebase in order to make it more "modern" and resolve potential issues and warnings.

[This way](https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/) it is possible to run `clang-tidy` and automatically fix the whole codebase.

Current progress:

- [ ] modernize-avoid-bind
- [ ] modernize-avoid-c-arrays
- [ ] modernize-concat-nested-namespaces
- [ ] modernize-deprecated-headers
- [ ] modernize-deprecated-ios-base-aliases
- [ ] modernize-loop-convert
- [ ] modernize-make-shared
- [ ] modernize-make-unique
- [ ] modernize-pass-by-value
- [ ] modernize-raw-string-literal
- [ ] modernize-redundant-void-arg
- [ ] modernize-replace-auto-ptr
- [ ] modernize-replace-disallow-copy-and-assign-macro
- [ ] modernize-replace-random-shuffle
- [ ] modernize-return-braced-init-list
- [ ] modernize-shrink-to-fit
- [ ] modernize-unary-static-assert
- [ ] modernize-use-auto
- [ ] modernize-use-bool-literals
- [ ] modernize-use-default-member-init
- [ ] modernize-use-emplace
- [ ] modernize-use-equals-default
- [ ] modernize-use-equals-delete
- [ ] modernize-use-nodiscard
- [ ] modernize-use-noexcept
- [ ] modernize-use-nullptr
- [ ] modernize-use-override
- [ ] modernize-use-trailing-return-type
- [ ] modernize-use-transparent-functors
- [ ] modernize-use-uncaught-exceptions
- [ ] modernize-use-using

FrancescoBorzi:
@azerothcore/developers @azerothcore/testers  I'm finally done with this. I need reviewers/testers for all the linked PRs, which you can find here:

[...]

PR #3835, opened 2020-12-02
refactor(Core): apply clang-tidy modernize-use-nodiscard

FrancescoBorzi:
- Part of https://github.com/azerothcore/azerothcore-wotlk/issues/3820
- Apply `clang-tidy` with `modernize-redundant-use-nodiscard`
https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nodiscard.html
- This is a really simple change, no testing needed.

It looks a massive change but what `[[nodiscard]]` does is simply:

> The [[nodiscard]] attribute can be used to indicate that the return value of a function shouldn't be ignored when you do a function call. If the return value is ignored, the compiler should give a warning on this.

so if the build CI passes (means no warnings), we are good to go


FrancescoBorzi:
the build with modules is failing because of a bug in Eluna, fix here: https://github.com/ElunaLuaEngine/Eluna/pull/340

Helias (approved):

PR #3843, opened 2020-12-04
refactor(Core/Social): Cleanup

FrancescoBorzi:
Just a small cleanup


src/server/game/Entities/Player/SocialMgr.cpp
```diff
@@ -315,16 +315,13 @@ PlayerSocial* SocialMgr::LoadFromDB(PreparedQueryResult result, uint64 guid)
     if (!result)
         return social;
 
-    uint8 flags = 0;
-    std::string note = "";
```
FrancescoBorzi, line 319: they are only used inside the `do` block

src/server/game/Handlers/Socialhandler.cpp
```diff
@@ -113,7 +113,7 @@ void WorldSession::HandleAddIgnoreOpcode(WorldPacket& recv_data)
         return;
 
     uint64 IgnoreGuid = MAKE_NEW_GUID(lowGuid, 0, HIGHGUID_PLAYER);
-    FriendsResult ignoreResult = FRIEND_IGNORE_NOT_FOUND;
+    FriendsResult ignoreResult;
```
FrancescoBorzi, line 116: this value was never used, it will be either `FRIEND_IGNORE_SELF`, `FRIEND_IGNORE_ALREADY` or `FRIEND_IGNORE_FULL` (see below)

Viste (approved):

src/server/game/Handlers/Socialhandler.cpp
Viste: Yes, but I decided to leave everything as is.

Si1ker (approved):

PR #4075, opened 2020-12-28
fix(DB/Quest): Shadow Vault Decree

wizzymore:
## Changes Proposed:
- Make Quest Item usable where it normally should be used.
Before:
image
After:
image
I've took in consideration exploiting this under the edge and it is not working ;D

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3284

## Tests Performed:
- Windows

## How to Test the Changes:

.go c 124058
.q a 12943
Try to use the quest item before and after

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #4081, opened 2020-12-28
chore(CI): Improve error message of check_pending_sql

Kitzunu:


locus313 (approved):

locus313 (approved):

locus313:
Just thought of something else. We probably should change the name of the workflow now too. 
https://github.com/azerothcore/azerothcore-wotlk/blob/master/.github/workflows/check_pending_sql.yml#L1
Maybe just change it to "check-pending-sql"

Kitzunu:
Done

locus313 (approved):

Kitzunu:
Weird stuff that https://github.com/azerothcore/azerothcore-wotlk/pull/4075 does not fail on multiple newlines @locus313 O.o

locus313:
I had forgotten that the `tr -d '\n'` part of the if filters out the new line characters. if we dont do this then we have to change the `tail -c 1` part of the if to include more characters (most likely just change it to 2), but this brings more possibilities of false negatives on the check.

Kitzunu:
F that

PR #4097, opened 2020-12-29
fix(Core/Spell): Killing Spree range

Kitzunu:
Co-Authored-By: Keader <keader.android@gmail.com>

## Changes Proposed:
Spell Killing Spree is missing some hits when the target is moving. The spell responsible for the main hand damage is: 57841, and off-hand damage: 57842.
The issue is: 57841 has 5 yards range, but 57842 has "self range". So if the target is moving, 57841 will hit, but 57842 will fail (out of range).

## Tests Performed:
- Builds
- Tested in-game

## How to Test the Changes:

1. Use .cast 57841 in a creature with a 3~4y distance. Only 1 damage will be applied
2. Use .cast 57841 very close to the creature. 2 Damages will be applied.

## Target Branch(es):
- [x] Master

 

Helias (approved):

pklloveyou:
Tested

Issue #4057, opened 2020-12-26
[Wintergrasp] Object visibility distance


Wowhungarian:
##### CURRENT BEHAVIOUR:
The problem is that, the destroyable game objects like Towers and Walls have low visibility distance at WG. Therefore [Wintergrasp Cannons](https://wowgaming.altervista.org/aowow/?npc=28366), which have greater visibility distance, than  the WG Towers, loads and falls down to the ground. Because of that, the Cannons on the Towers are usually unusable.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
Cannons stays at their position.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Go away from WG Tower Cannons until, they disappeare.
2. Approach the Towers.
3. Check Cannons.

##### AC HASH/COMMIT:
d3aff86c32334ad86d93aa37f09221c162049cf7

##### OPERATING SYSTEM:
Windows 10

##### MODULES:
mod-anticheat

##### OTHER CUSTOMIZATIONS:
nope

##### SCREENSHOT:
![cannon](https://user-images.githubusercontent.com/36567380/103166841-624b6f00-4826-11eb-98d0-66fd49979693.jpg)

It can also fall under the tower texture.

![cannon2](https://user-images.githubusercontent.com/36567380/103172999-187b7c80-4858-11eb-9d52-0bb53e077ff7.jpg)

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/commit/2779833768bbe9967006d6d16888dace05b4bcb6

PR #4107, opened 2020-12-30
fix(Core/Object): Increase wg object visibility

Wowhungarian:
## Changes Proposed:
-  Wintergrasp game objects like Towers and Walls have low visibility distance. Therefore the Cannons on the Towers appears before the Towers itselves.
-  This is not a permanent solution, but it has done the job.

## Issues Addressed:
- Closes  https://github.com/azerothcore/azerothcore-wotlk/issues/4057

## Tests Performed:
- tested on windows
- built without errors

## How to Test the Changes:
- .gm on
- .tele wintergrasp
- .gm fly on
- .mod speed 10
- fly to the Wintergrasp Fortress
- check towers and cannons (towers should render, before cannons)

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Entities/Object/Object.h
+++ b/src/server/game/Entities/Object/Object.h
@@ -28,11 +28,11 @@ class ElunaEventProcessor;
 #define INTERACTION_DISTANCE        5.5f
 #define ATTACK_DISTANCE             5.0f
 #define MAX_SEARCHER_DISTANCE       150.0f // pussywizard: replace the use of MAX_VISIBILITY_DISTANCE in searchers, because MAX_VISIBILITY_DISTANCE is quite too big for this purpose
-#define MAX_VISIBILITY_DISTANCE     300.0f // max distance for visible objects, experimental
+#define MAX_VISIBILITY_DISTANCE     250.0f // max distance for visible objects, experimental
 #define VISIBILITY_INC_FOR_GOBJECTS 300.0f // pussywizard
 #define VISIBILITY_COMPENSATION     15.0f // increase searchers
 #define SPELL_SEARCHER_COMPENSATION 30.0f // increase searchers size in case we have large npc near cell border
-#define VISIBILITY_DIST_WINTERGRASP 280.0f
+#define VISIBILITY_DIST_WINTERGRASP 175.0f
 #define SIGHT_RANGE_UNIT            50.0f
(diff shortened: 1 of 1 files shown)


lineagedr:
Why not just increase the VISIBILITY_DIST_WINTERGRASP instead of VISIBILITY_INC_FOR_GOBJECTS?

Wowhungarian:
@lineagedr Cause object visibility is a general issue. There are lot of world objects like doors etc, which needs visibility increase. But But yes, I also saw that this range is counted form 2 variables. You are right, that can be also a good solution.

Wowhungarian:
@lineagedr I forgot to mention, that VISIBILITY_DIST_WINTERGRASP will increase not just the gobject distance, but it also the creature visibility distance, which can cause huge fps drop at wg.

FrancescoBorzi:
@Wowhungarian are you sure such a generic change has no negative side-effects?

Wowhungarian:
@FrancescoBorzi Np, we need to find out how generic it excatly is. But even if it will affect a few gobjects, I'm sure we won't have a noticable performance issue, since most of the time, there just aren't much gobjects in 300 yard (if it's even 300 yard). We can't say the same with creatures. Increasing their visibility will always cause performance drop.

Kitzunu:
If you want to, yuou could backport TCs system

Wowhungarian:
As well as I know, TC has a worse state in terms of wg objects.

Wowhungarian:
Btw we need to edit  VISIBILITY_DIST_WINTERGRASP 's value too because of an other bug, which I will explain soon, but it should never be greater than what the VISIBILITY_INC_FOR_GOBJECTS has, cause the cannon problem will remain, since that one also affects cannon visibility.

pklloveyou:
Tested

pklloveyou:
Far is better than near

Mitradis:
need update @Wowhungarian

Si1ker:
VISIBILITY_INC_FOR_GOBJECTS and other defines have been moved from Object.h to [ObjectDefines.h](https://github.com/azerothcore/azerothcore-wotlk/blob/f6a17164be8700d4e457a51071ca4034ecbc5b73/src/server/game/Entities/Object/ObjectDefines.h#L17)
Therefore the PR is invalid, if still interested in this topic you can re-open the PR with the necessary changes.

PR #4231, opened 2021-01-07
fix(DB/Update): Fix missing delete

Kitzunu:
We really need a check for this @locus313 ... Can't allow things like this to pass


wizzymore:
Hmm, wouldn't it be easier to have an action that applies the query two times to check for this kind of errors?

FrancescoBorzi (approved):

Issue #4225, opened 2021-01-07
server crash


mpfans:
https://github.com/azerothcore/azerothcore-wotlk/commit/b2761626fec1d454acb58f08c0e0e3d8c9b83e47 will cause the server to crash for a period of time

![20210107091652](https://user-images.githubusercontent.com/18187177/103839027-76665a80-50c9-11eb-9e8f-32791fd2ad71.jpg)

OS:ubuntu18.04
Module:mod-anticheat
AzerothCore rev. 9d9b71eda1c0+ 2021-01-05 19:46:50 +0000 (master branch)

FrancescoBorzi:
@Yehonal

mpfans:
no any custom code
this is git log
![image](https://user-images.githubusercontent.com/18187177/103894885-7e0a1b80-512a-11eb-980f-927041919638.png)

Yehonal:
https://github.com/azerothcore/azerothcore-wotlk/pull/4220

Kitzunu:
if this will be a long wip then it is not good. Need a quick fix for a crash...

Yehonal:
https://github.com/azerothcore/azerothcore-wotlk/pull/4220/commits/a840bed9681c90df5fbc6573c434668c1d22d90d

This is the commit that includes the fix.

That PR won't be a very long WIP though. The main TODOs are already completed, just small stuff missing, I would like to merge it asap. But if you want we can cherry pick that commit directly on master

Yehonal:
https://github.com/azerothcore/azerothcore-wotlk/pull/4235

mpfans:
> #4235

Update the latest version and then accept this PR for testing?

PR #4235, opened 2021-01-08
fix(Core/Crash): safe calculation in GetMeleeAttackPoint

Yehonal:
## Changes Proposed:

complex float calculations can have weird behaviours. I've added a check to use a simpler formula in case of catastrophic failures

## Issues Addressed:
- Closes  https://github.com/azerothcore/azerothcore-wotlk/issues/4225

## How to test:

- Engage multiple NPCs at the same time (lot of them) to benchmark the stability of the repositioning system


Helias (approved):

mpfans:
Upgrade the latest AC, and then accept this PR for testing?

PR #4259, opened 2021-01-11
Adding information to the ip2nation table and ip2nationCountries

pangolp:
## Changes Proposed:
-  Through the information provided by the web [ip2nation.com](http://www.ip2nation.com/ip2nation/Download), the idea is to add the information within the table.

## Issues Addressed:
- Closes 

## Tests Performed:
- If you use the `.player info` command you should be able to get the player's country.

## Target Branch(es):
- [x] Master

 

Kitzunu:
The question is if we really should populate these tables. A lot of data for those who do not want or need it. and can be easily imported from the website by anyone

@FrancescoBorzi @Helias @Yehonal

pangolp:
> The question is if we really should populate these tables. A lot of data for those who do not want or need it. and can be easily imported from the website by anyone like https://www.azerothcore.org/wiki/ip2nation suggests
> 
> @FrancescoBorzi @Helias @Yehonal

It can be an alternative, I don't really know. I thought it might be a good idea to import the data and know where the people are from.

Yehonal:
Instead of adding it within the core, we could open/create a repo for it and link that repo to our catalogue. Would it be a good solution?

Kitzunu:
Well it is linked on the wiki already :shrug: But I suppose that works too

pangolp:
This could be an option. Anyway, in the wiki, the file you download is all together and it has an INSERT for each record. Here I left only 1 INSERT. I think it would be interesting, to leave it somewhere, for whoever wants to add it.

FrancescoBorzi:
@pangolp thanks a lot for this contribution, however I agree with @Kitzunu and @Yehonal . I don't think this should be part of the core as it's a lot of data and not everyone needs it, an optional feature is enough :)

pangolp:
Yes, I think it's a good idea.

pangolp:
I also think it's a good idea to have it as an optional resource.

den13501:
tested, it work

pangolp:
> tested, it work

Thanks for the test, but in the end, will not be incorporated into the emulator, will be available for those who want to apply it, but as something optional.

Kitzunu:
I have made a module, will port later

PR #4321, opened 2021-01-21
fix(DB/gameobject): The Damaged Chest respawn rate

Zoidwaffle:
## Changes Proposed:
- Lower the respawn rate from 5 minutes to 2s.

## Issues Addressed:
- The Damaged Chest has a 5 minute respawn rate - which makes questing as a warlock painful if there are other warlocks on the same quest.
- The best video I could find: [Youtube](https://youtu.be/5TXyVDWkaec?t=197) (from Classic). It appears as if the chest never really disappears and definitely not for long if it does.
- There are other videos but they seem to be from private servers with a 5 min spawn rate, so they are not added here
- [Others with a similar reported issue](https://github.com/FeenixServerProject/Archangel_2.4.3_Bugtracker/issues/3310)

PLEASE NOTICE: I'm not 100% sure if this is a Blizzlike fix or not. I have been unable to verify this - but the video from classic indicates the current respawn rate is way too long... and logically I think the chest should be there so you don't search the right place without it being there, just because some other warlock looted it 3 minutes earlier...

## Tests Performed:
- .quest add 1805
- .tele splithoof
- Run further into the cave, stay on the right, across the water on the left hand side, the chest appears
- Loot the chest - wait for 5 minutes and notice other frustrated warlocks unable to finish their quests...
- Apply the fix and enjoy a good time with your fellow warlocks next to you!

## How to Test the Changes:

Just see **Tests Performed**.

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

pklloveyou:
Tested

Issue #4304, opened 2021-01-17
Darkshore cemetaries not available for horde players


Gizmoh:
##### CURRENT BEHAVIOUR:
If as a horde player you die in Darkshore, currently tested at auberdine harbor, your spirit appears at ashenvale, and not even the nearer cementery in ashenvale.
![imagen](https://user-images.githubusercontent.com/4155081/104849971-6faac500-58cb-11eb-91d1-ba63a0f0d5a8.png)
![imagen](https://user-images.githubusercontent.com/4155081/104849985-7b968700-58cb-11eb-9a95-3f45a0eff821.png)
![imagen](https://user-images.githubusercontent.com/4155081/104850002-7fc2a480-58cb-11eb-8a01-34435d5d6b41.png)

##### EXPECTED BLIZZLIKE BEHAVIOUR:
You should appear in a cemetery in Darkshore instead of Ashenvale

##### STEPS TO REPRODUCE THE PROBLEM:

1.  Be Horde
2.  Die at Auberdine
3.  ???
4. Profit

##### AC HASH/COMMIT:
df600f9946c

##### OPERATING SYSTEM:

Ubuntu 20.04
##### MODULES:

    mod-cfbg
    mod-duel-reset

##### OTHER CUSTOMIZATIONS:

None.
##### SERVER:

ChromieCraft

nicoaravena:
This is not the only spirit healer with that problem, in Tirisfal Glades if you die near the scarlet monastery, you will be spawned as ghost in the Brill spirit healer, and there is a near one there.
Also it appears that the correct behaviour is to find the nearest respawn in the map you are located, except for the main cities for the opposite faction (as you can find here https://www.wowhead.com/npc=6491/spirit-healer#comments:id=621748)

Zoidwaffle:
@nicoaravena : maybe you should open an issue regarding this?

I tried to look into it, but I don't know how to solve it.

The issue is, as I see it, that everything outside of the monastery is zone 85. Once you enter the building, you are in zone 796.

This was tested as Alliance.

Die in zone 796 and you end up in the GY marked with blue on the map... but die in zone 85, and you end up near Brill instead.

There are also "Area" code changes besides the zone ones, but the graveyard_zone table in the database doesn't use that so I don't know if this can be even fixed unless some core code is rewritten to handle this (and I have no idea if this is Blizz-like or not).

Helias:
Are we sure that this is not blizzlike?

[...]

PR #4322, opened 2021-01-21
fix(DB/Graveyard): Add Horde graveyards in Darkshore

Zoidwaffle:
## Changes Proposed:
~~- Move the Horde GY in Darkshare into Darkshore (Twilight Vale)~~
~~- Notice that I remove the Horde specific GY reference and share it between Horde and Alliance (it's already used by both factions if you die in zone 141 as Horde or 148 as Alliance)~~
- Latest commit: Notice that I remove the Horde specific GY reference and share the 2 GY's there between Horde and Alliance

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4304

## Tests Performed:
- You need to be a Horde character
- .tele darkshore
- .die
- Release spirit, notice you are far away in Ashenvale
- Apply the fix, now you remain in the Darkshore zone when you release your spirit
- Depending on if you die in the Northern or Southern part of the map, you end up at different GY's.

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

nicoaravena:
They are 2 spirit healers in darkshore https://classic.wowhead.com/npc=6491/spirit-healer

If I die new the top one, I will respawn at the bottom as ghost. It should be able to respawn near both when apply.
imagen

Zoidwaffle:
I also found the 2 spirit healers but I have no idea if they are both supposed to be used by both factions are not.

If they are indeed shared between factions, I can make a fix for that as well... I should have time to make a commit to the PR addressing this later today I hope.

FrancescoBorzi:
> I also found the 2 spirit healers but I have no idea if they are both supposed to be used by both factions are not.
> 
> If they are indeed shared between factions, I can make a fix for that as well... I should have time to make a commit to the PR addressing this later today I hope.

I see, I'd say while we are not still sure about the other GY, we can still merge this one because it improves the game already

Zoidwaffle:
Hm, I'm not sure how the GY's are differentiated in the same zone.

All of Darkshore seems to be zone 148.

But there are 2 GY's for Alliance there (this is before the fix btw).
```
> select * from graveyard_zone WHERE `GhostZone`=148;
+-----+-----------+---------+-----------------------------------------+
| ID  | GhostZone | Faction | Comment                                 |
+-----+-----------+---------+-----------------------------------------+
|  35 |       148 |     469 | Darkshore, New Auberdine GY - Darkshore |
| 469 |       148 |     469 | Darkshore, Twilight Vale GY             |
| 512 |       148 |      67 | Ashenvale, Kargathia GY                 |
+-----+-----------+---------+-----------------------------------------+
```

So depending on where in the zone you die, you end at different GY's as Alliance. As Horde you are send far away (which obviously seems wrong).

I do think the current fix is an improvement - but I would like to understand where the logic is that sends a dead player in the same zone to different GY's.

Zoidwaffle:
OK, I managed to find some sort of proof...

[Horde GY is North of Auberdine](https://2wowdoctors.wordpress.com/2008/03/18/the-more-than-complete-guide-on-how-to-get-a-ravager-pet-as-a-horde-hunter/)

See step 15. in the guide above.

So it true, the Horde should indeed also be able to use the Northern GY.

But I don't know how to fix that (yet).

Zoidwaffle:
OK, so at the risk of talking to myself (sorry for the spam!)...

If both GY's are shared anyway, then the new fix should work.

I quickly tested it myself - where I did end up at the Northern or Southern GY depending on where I died.

FrancescoBorzi:
thanks @Zoidwaffle !

@nicoaravena could you please test this as well?

nicoaravena:
Works good now. Good job.

imagen
imagen

PR #4332, opened 2021-01-23
fix(gcc): type warning

FrancescoBorzi:
This will not fix the build, but will fix one of the 2 issues causing it to fail.


Kitzunu (approved):

FrancescoBorzi:
ok, one of the issues was fixed by this. Merging it and addressing the other one separately.

PR #4334, opened 2021-01-23
docs(SECURITY.md): rebrand supported dependencies

FrancescoBorzi:


.github/SECURITY.md
```diff
@@ -2,30 +2,69 @@
 
 ## Supported Versions
 
-We support these versions of AzerothCore:
+We support the following versions of dependencies.
+
+:white_check_mark: = supported
+:red_circle: = NOT supported
```
Kitzunu, line 8: We dont need to tell them what we don't support, anything that is not supported is just automatically not supported

.github/SECURITY.md
```diff
@@ -2,30 +2,69 @@
...
+Versions of MySQL:
 
 | MySQL Version | Supported          |
 | ------------- | ------------------ |
-| 5.6           | :white_check_mark: |
-| 5.7           | :white_check_mark: |
 | 8.0           | :white_check_mark: |
+| 5.7           | :white_check_mark: |
+| 5.6           | :white_check_mark: |
+| 5.5 and lower | :red_circle:       |
```
Kitzunu, line 24: Nope, we dont support 5.6

.github/SECURITY.md
FrancescoBorzi: @Kitzunu this is to be more explicit about things we know for sure that won't work at all (or are strongly discouraged to use). There are some versions that we don't check in our CI but they still probably work, we need to make a dinstinction

Kitzunu: Well as for DB Versions at least, we don't need to specify what we do not support because for example running the server without MySQL 5.7 or above is impossible

FrancescoBorzi: we do as we explicitly want to tell the user that a lower version should NOT be used, to distinguish cases were we just can't guarantee it works but there is nothing wrong about trying it, e.g. ubuntu 20.10

.github/SECURITY.md
FrancescoBorzi: right, updated thanks

Helias (approved):

.github/SECURITY.md
```diff
@@ -2,30 +2,68 @@
...
+| 18.04           | :white_check_mark: |
+
+Versions of macOS:
+
+| macOS Version   | Supported          |
+| --------------- | ------------------ |
+| 10.15           | :white_check_mark: |
 
-Note: We do not support any repacks that may or may not have been made based on AzerothCore.
+**Note**: We do NOT support any repacks that may or may not have been made based on AzerothCore.
```
Kitzunu, line 66:
```suggestion
**Note**: We do NOT support any repacks that may or may not have been made based on AzerothCore. Nor do we support any of the releases that are made under AzerothCore.
```

Kitzunu: Something like that, perhaps it could be worded better

PR #4368, opened 2021-01-29
fix(DOCKER): Reduce context build time

nethw:
This PR makes the docker initialization time less than 1 second, when it was between 1 - 3 minutes before, using a VPS with 4GZ 4 cores CPU.

## Changes Proposed:
-  removing all files copied (build cache, data, git files, sources)
-  creating new volumes mounted

## Issues Addressed:
- No issue. It was annoying to wait 2 minutes for the initialization time of docker everytime I made a change to the core.

## Tests Performed:
- I'm using this branch on my realm hosted on a machine
- It works.

## How to Test the Changes:
- just launch `./bin/acore-docker-build`

## Target Branch(es):
- [x] Rochet2-patch-1

### DIFF
--- a/.github/workflows/core_build.yml
+++ b/.github/workflows/core_build.yml
@@ -182,4 +182,4 @@ jobs:
           docker-compose --version
       - name: Build
         run: |
-          docker build -t azerothcore/database -f docker/database/Dockerfile .
+          docker-compose build ac-database
--- a/bin/acore-docker-build
+++ b/bin/acore-docker-build
@@ -7,7 +7,6 @@ docker run \
     -v /"$(pwd)"/conf/dist:/azerothcore/conf/dist \
     -v /"$(pwd)"/src:/azerothcore/src \
     -v /"$(pwd)"/modules:/azerothcore/modules \
-    -v /"$(pwd)"/CMakeLists.txt:/azerothcore/CMakeLists.txt \
     -v /"$(pwd)"/docker/build/cache:/azerothcore/build \
     -v /"$(pwd)"/docker/worldserver/bin:/binworldserver \
     -v /"$(pwd)"/docker/authserver/bin:/binauthserver \
--- a/bin/acore-docker-build
+++ b/bin/acore-docker-build
@@ -3,6 +3,11 @@
 docker build -t acbuild -f docker/build/Dockerfile .
 
 docker run \
+    -v /"$(pwd)"/deps:/azerothcore/deps \
+    -v /"$(pwd)"/conf/dist:/azerothcore/conf/dist \
+    -v /"$(pwd)"/src:/azerothcore/src \
+    -v /"$(pwd)"/modules:/azerothcore/modules \
+    -v /"$(pwd)"/CMakeLists.txt:/azerothcore/CMakeLists.txt \
     -v /"$(pwd)"/docker/build/cache:/azerothcore/build \
     -v /"$(pwd)"/docker/worldserver/bin:/binworldserver \
     -v /"$(pwd)"/docker/authserver/bin:/binauthserver \
--- a/bin/acore-docker-build-no-scripts
+++ b/bin/acore-docker-build-no-scripts
@@ -3,6 +3,10 @@
 docker build --build-arg ENABLE_SCRIPTS=0 -t acbuild -f docker/build/Dockerfile .
 
 docker run \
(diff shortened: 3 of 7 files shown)


nethw:
looks like the database requires the data folder, im about to mount the host folder for it

nethw:
`core-build / docker-database` test can't pass until the test uses `docker-compose build`. 
This PR added a few volumes which are required for launching the database container

nethw:
> `core-build / docker-database` test can't pass until the test uses `docker-compose build`.
> This PR added a few volumes which are required for launching the database container

edited the CI

nethw:
this PR is ready. I tested it on my own machine, it works.

docker-compose.yml
```diff
@@ -18,6 +18,9 @@ services:
       - type: volume
         source: ac-database
         target: /var/lib/mysql
+      - type: bind
+        source: ./data
+        target: /azerothcore/data
```
locus313, line 23: the database docker is a multistage container. This bind will only get added to the last stage of the docker, which will not work as the /data is only used in the stage and the first time the docker is build and will not generate the databases correctly(it will create the databases but they will be empty). We should leave the `COPY data /azerothcore/data` command in the dockerfile

docker/database/Dockerfile
```diff
@@ -7,7 +7,6 @@ RUN apk add --no-cache bash
 COPY apps /azerothcore/apps
 COPY bin /azerothcore/bin
 COPY conf /azerothcore/conf
-COPY data /azerothcore/data
```
locus313, line 10: as i mentioned in my other comment, we should leave this in place.

.github/workflows/core_build.yml
```diff
@@ -182,4 +182,4 @@ jobs:
           docker-compose --version
       - name: Build
         run: |
-          docker build -t azerothcore/database -f docker/database/Dockerfile .
+          docker-compose build ac-database
```
locus313, line 185: would change this back after adding the copy command back into the dockerfile.

bin/acore-docker-build
```diff
@@ -3,6 +3,10 @@
 docker build -t acbuild -f docker/build/Dockerfile .
 
 docker run \
+    -v /"$(pwd)"/deps:/azerothcore/deps \
+    -v /"$(pwd)"/conf/dist:/azerothcore/conf/dist \
+    -v /"$(pwd)"/src:/azerothcore/src \
+    -v /"$(pwd)"/modules:/azerothcore/modules \
```
locus313, line 9: please also add `-v /"$(pwd)"/.git:/azerothcore/.git \`. if you don't you will end up getting errors like this:
```
fatal: not a git repository (or any of the parent directories): .git
--
    Could not find a proper repository signature (hash) - you may need to pull tags with git fetch -t
    Continuing anyway - note that the versionstring will be set to "unknown 1970-01-01 00:00:00 (Archived)"
```

bin/acore-docker-build-no-scripts
```diff
@@ -3,6 +3,10 @@
 docker build --build-arg ENABLE_SCRIPTS=0 -t acbuild -f docker/build/Dockerfile .
 
 docker run \
+    -v /"$(pwd)"/deps:/azerothcore/deps \
+    -v /"$(pwd)"/conf/dist:/azerothcore/conf/dist \
+    -v /"$(pwd)"/src:/azerothcore/src \
+    -v /"$(pwd)"/modules:/azerothcore/modules \
```
locus313, line 9: same here, please add the .git as well.

docker-compose.yml
Romain-P: this is a problem because the core build does not require those data, and this is a 4GB+ folder. the problem comes from the fact the same .dockerignore file is used for the 3 images. the build of acore-worldserver does not require ./data

nethw: this is a problem because the core build does not require those data, and this is a 4GB+ folder. the problem comes from the fact the same .dockerignore file is used for the 3 images. the build of acore-worldserver does not require ./data

locus313: Are you storing your map data in `./data` directory (the data folder should not be 4GB+)? When using docker you should actually be putting the map data into `docker/worldserver/data/`.(see https://www.azerothcore.org/wiki/Install-with-Docker)  This avoids the issue have having the map data get sent during the docker build context. Without the map data the `./data` folder is about ~462MB.

PR #4379, opened 2021-01-30
fix (Core/Commands) remove character creation flag from .save command

55Honey:
## Changes Proposed:
The .save command has a `true` flag which claims it should create a character. It should be changed to `false` because the character obviously already exists at the time the command is used.

See https://github.com/azerothcore/azerothcore-wotlk/blob/0a8a7ef1494b1dcc43bdf5134f7c4836acc66f4e/src/server/game/Entities/Player/Player.cpp#L26487 regarding the syntax of the arguments.

## Issues Addressed:
Insert statement erroring out because character already exists. No saves to db happening.
image

## Tests Performed:
- Built on Win10 64 with MySQL 5.7.18
- Tested on default account test1 and test10. .save works

## How to Test the Changes:
- .save
- Watch worldserver console not erroring
- Possibly watch the db prior to and after the .save and see if e.g. level gains were saved to the `characters` table.

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -1093,11 +1093,11 @@ public:
         {
             if (Player* target = handler->getSelectedPlayer())
             {
-                target->SaveToDB(true, false);
+                target->SaveToDB(false, false);
             }
             else
             {
-                player->SaveToDB(true, false);
+                player->SaveToDB(false, false);
             }
             handler->SendSysMessage(LANG_PLAYER_SAVED);
             return true;
@@ -1107,7 +1107,7 @@ public:
         uint32 saveInterval = sWorld->getIntConfig(CONFIG_INTERVAL_SAVE);
         if (saveInterval == 0 || (saveInterval > 20 * IN_MILLISECONDS && player->GetSaveTimer() <= saveInterval - 20 * IN_MILLISECONDS))
         {
-            player->SaveToDB(true, false);
+            player->SaveToDB(false, false);
         }
 
         return true;


pklloveyou:
Tested

BarbzYHOOL:
need to analyze the code as to what it was like that and why nobody has the bug, i'm on an old core and no issue

others have no issue either

so analyze the code and explain why it must be changed

55Honey:
https://github.com/azerothcore/azerothcore-wotlk/blob/cea31fd0d30ca6cb7219b69206fc21d097221bea/src/common/Database/Implementation/CharacterDatabase.cpp#L318
This one is called from the `true` flagged part of the void.

And this one is used if the `create` is passed on as `false`:
https://github.com/azerothcore/azerothcore-wotlk/blob/cea31fd0d30ca6cb7219b69206fc21d097221bea/src/common/Database/Implementation/CharacterDatabase.cpp#L327

The difference regarding INSERT and UPGRADE is clearly there. Why does it work for others when INSERT is used but a primary key with that index already exists in the sql table?

Kitzunu:
if (create) is when you create a new character right?

55Honey:
> if (create) is when you create a new character right?

I guess so. But i could not track down other parts of the core which call this void with `create = true`.

55Honey:
That is all i found, but it seems not to refer to the above void at all.
https://github.com/azerothcore/azerothcore-wotlk/blob/c8f43d85849c40e08d0da38913489426a0c35388/src/server/game/Handlers/CharacterHandler.cpp#L635

BarbzYHOOL:
> Our character obviously already exists at the point where anyone can use .save on it, so it doesn't make sense in my opinion, to pass the bool create argument as true here. In fact, i wonder how it can work for anyone else?

yes, logic, but maybe it saves without using .save

Try to make a new character and see if it's saved in the db when logging out, then make another char and close the worlserver and see if it saves it.

Btw, the query could do all in one too instead of extra cpp lines, like INSERT INTO BUT IF EXISTS UPDATE (this is not the syntax but it is possible to do it)

55Honey:
> save



> yes, logic, but maybe it saves without using .save

My point is to fix a (why only to my DB?) broken .save command tho :)

Since im not a SQL pro, my question stands: How can an INSERT Statement with an existing primary key (the character guid) not error out? 
Since the core does not use DELETE here before the INSERT.

> Try to make a new character and see if it's saved in the db when logging out, then make another char and close the worlserver and see if it saves it.

I can tell that in both these scenarios characters are saved on my local core.

FrancescoBorzi:
> Since im not a SQL pro, my question stands: How can an INSERT Statement with an existing primary key (the character guid) not error out?
> Since the core does not use DELETE here before the INSERT.

to update an existing row without deleting it first, you can use `UPDATE`

55Honey:
> > Since im not a SQL pro, my question stands: How can an INSERT Statement with an existing primary key (the character guid) not error out?
> > Since the core does not use DELETE here before the INSERT.
> 
> to update an existing row without deleting it first, you can use `UPDATE`

Sorry for being so vague, but i know that it works that way. My question was more about, why it does Not produce an error in other peoples SQL servers. The core currently uses an INSERT statement currently without DELETE first instead of UPDATE on an already existing row. Which is basically the whole point of this PR.

FrancescoBorzi:
> > if (create) is when you create a new character right?
> 
> I guess so. But i could not track down other parts of the core which call this void with `create = true`.

@55Honey what IDE are you using? I can find them using CLion and, if you need it, we have open-source licenses for it.

image

image

FrancescoBorzi (approved):

FrancescoBorzi:
about the PR, it makes sense to me and it also has been tested. So marking it as [To Be Merged].

Thanks @55Honey and @pklloveyou !

55Honey:
> @55Honey what IDE are you using? 

Im using IntelliJ. Id rather blame my lack of C++ knowledge than the IDE tho :)

Issue #4399, opened 2021-01-31
Gathering Quest While in Party bug


pangolp:
### SMALL DESCRIPTION:
When in a party with a quest with drops shows it is lootable (see picture below) even if the corpse has nothing.

### EXPECTED BLIZZLIKE BEHAVIOUR:
Once the corpse is looted it should no longer be interactable and the cursor should be a gauntlet instead of bag.

### CURRENT BEHAVIOUR:
The body has no sparkles but you can "loot" it which does nothing and hovering over it shows a bag instead of a gauntlet. When you loot it in this state you bend over to pick something up but nothing happens and no loot window appears.

![105868962-5933f600-5fbc-11eb-8df6-c103aaefeb15](https://user-images.githubusercontent.com/2810187/106400220-e5e71580-63fb-11eb-804d-a244294e7af1.png)

### STEPS TO REPRODUCE THE PROBLEM:
I had my wife who was playing with me confirm the issue existed on her end. She is not using a UI mod. I did two quests in the BE starting area that had quest drops and they worked like above. I created a new character and the same thing happened. Seems to only happen while in a party.

### AC HASH/COMMIT:
- chromiecraft/azerothcore-wotlk@df600f9

### OPERATING SYSTEM:
- Ubuntu 20.04

### MODULES:
1. mod-cfbg
2. mod-duel-reset

### OTHER CUSTOMIZATIONS:
- None.

### SERVER:
- ChromieCraft

pangolp:
The original report can be found here:
https://github.com/chromiecraft/chromiecraft/issues/20

lineagedr:
https://github.com/TrinityCore/TrinityCore/commit/a49544cc187d3a156b4615907d6b82057364fcc4

pangolp:
Excellent contribution @lineagedr. I will see if I can replicate the arrangement.

pangolp:
I am currently compiling, testing the changes, and creating the pull request.

pangolp:
It works like a charm. Only 1 of the 2 can get the item, and the bag disappears.
![WoWScrnShot_020121_120224](https://user-images.githubusercontent.com/2810187/106476130-7b7ab780-6485-11eb-9d30-31b197c7ec59.jpg)

pangolp:
@lineagedr thanks again for the information.

PR #4409, opened 2021-02-01
Gathering Quest While in Party bug

pangolp:
## Changes Proposed:
Correct the problem that the bag when obtaining the item, by one of the group members, currently does not disappear, and remains so, until the body disappears. In this way, it is prevented the immediate stripping of the creature being in a group for example.

## Issues Addressed:
- Closes 

https://github.com/azerothcore/azerothcore-wotlk/issues/4399
https://github.com/chromiecraft/chromiecraft/issues/20

## Tests Performed:
- Windows

## How to Test the Changes:

Create a group, and perform a quest with an item that must be obtained by killing a creature. If the item does not have the flag that everyone can obtain it, only 1 should be able to obtain it.

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -18595,6 +18595,9 @@ bool Player::isAllowedToLoot(const Creature* creature)
     if (loot->isLooted()) // nothing to loot or everything looted.
         return false;
 
+    if (!loot->hasItemForAll() && !loot->hasItemFor(this)) // no loot in creature for this player
+        return false;
+
     Group* thisGroup = GetGroup();
     if (!thisGroup)
         return this == creature->GetLootRecipient();
--- a/src/server/game/Handlers/LootHandler.cpp
+++ b/src/server/game/Handlers/LootHandler.cpp
@@ -392,13 +392,10 @@ void WorldSession::DoLootRelease(uint64 lguid)
                 loot->roundRobinPlayer = 0;
 
                 if (Group* group = player->GetGroup())
-                {
                     group->SendLooter(creature, nullptr);
-
-                    // force update of dynamic flags, otherwise other group's players still not able to loot.
-                    creature->ForceValuesUpdateAtIndex(UNIT_DYNAMIC_FLAGS);
-                }
             }
+            // force dynflag update to update looter and lootable info
+            creature->ForceValuesUpdateAtIndex(UNIT_DYNAMIC_FLAGS);
         }
     }
 
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -781,6 +781,14 @@ uint32 Loot::GetMaxSlotInLootFor(Player* player) const
     return items.size() + (itr != PlayerQuestItems.end() ?  itr->second->size() : 0);
 }
 
+bool Loot::hasItemForAll() const
+{
+    for (LootItem const& item : items)
+        if (!item.is_looted && !item.freeforall && item.conditions.empty())
+            return true;
+    return false;
+}
+
 // return true if there is any FFA, quest or conditional item for the player.
 bool Loot::hasItemFor(Player* player) const
 {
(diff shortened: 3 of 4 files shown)


pangolp:
While the pull request solves part of the problem. We have to check another related problem. In the database, there are items that have to have a flag, the number 2048, which allows everyone to get the item. The problem is that it is not in all, then, we will have to detect them later and add the corresponding flag.

```sql
-- https://www.wowhead.com/item=22889/vial-of-moth-blood
UPDATE `item_template` SET `Flags`=`Flags`|2048 WHERE `entry`=22889;
```

loot

```
| Flag | Hex    | Comments                                    |
|------|--------|---------------------------------------------|
| 2048 | 0x0800 | Item is party loot and can be looted by all |
```
```sql
SELECT `entry`, `name`, `flags` FROM `item_template` WHERE `entry`=22889;
```
```
| entry | name               | flags |
|-------|--------------------|-------|
| 22889 | Vial of Moth Blood | 0     |
```

Kitzunu:
A lot of quests shouldnt have it.

For example the majority of quests where you have to gather more than one item should not have that flag

pangolp:
> A lot of quests shouldnt have it.
> 
> For example the majority of quests where you have to gather more than one item should not have that flag

It is that most of them do not have it, however, some do. That is why I do not propose to add them all. Rather, we will have to detect which are the ones that should be modified. But that is the subject of another issue or pull request. This fix helps to remove the bag, in case it does not have the flag, which is also necessary. Because if it does not have the flag, and on top of that, this modification is not done, the bag will always remain in the npc once it is dead. For which, the usual behavior is that once someone gets the item, if he does not have the flag, it is no longer available, for example, to be skinned, in the case of a beast.

pangolp:
The pull request does not assign the flag, it removes the bag.
The comment I made above should actually go in another issue.

pangolp:
This item is a case of one that should have the flag, but I agree that they are not all.

pangolp:
Another example of an item that should be obtained by everyone is this: https://www.wowhead.com/item=22934/lasher-sample and only let one person in the group actually obtain it. That is why I mentioned that although they are 2 different topics, they are related. That is why I may have generated confusion. But we have another pull request, that I also have to try, because maybe it will solve the problem.

https://www.wowhead.com/items?filter=166:98;3:1;0:0
That could be a possible list. But I must continue my research.

55Honey:
> A lot of quests shouldnt have it.
> 
> For example the majority of quests where you have to gather more than one item should not have that flag

From my leveling experience on a Master core from last years June, i can confirm that especcially in Northrend lots of questitems were only lootable by a single character. Especcially those obtained from objects.

pangolp:
> > A lot of quests shouldnt have it.
> > For example the majority of quests where you have to gather more than one item should not have that flag
> 
> From my leveling experience on a Master core from last years June, i can confirm that especcially in Northrend lots of questitems were only lootable by a single character. Especcially those obtained from objects.

Thanks @55Honey. Correct. The pull request partially solves the problem. Because what it does, is that at least, one gets it and then it is no longer available. The complement to this is to check the loot party items and add the flag to them. Not all items have this condition, but there are several.

pangolp:
Here is an example of an item that has the flag.
Maybe not many are missing, but evidently that is the arrangement.
```sql
SELECT `entry`, `name`, `flags` FROM `item_template` WHERE `entry`=36746;
```
```
| entry | name                      | flags |
|-------|---------------------------|-------|
| 36746 | Goramosh's Strange Device | 2048  |
```
I think this fix should be applied, and at the same time, add the flag to the items. Now I am going to make another pull request, with the items that I have detected and that I know are loot part items.

pangolp:
Here we have information about items from level 1 to 19.
https://wowgaming.altervista.org/aowow/?items&filter=cr=98:5;crs=1:1;crv=0:0;minle=1;maxle=19#0+2

pklloveyou:
It works, but I don't know if it will affect other actions

pklloveyou:
I'll put it on the server on Monday for players to test

pangolp:
I tried it, with several missions and that's what happened. The loot would appear, someone would get it, and the second one you couldn't get it anymore, but at least, the bag disappears and allows you to skin it for example, in case it's a beast. Now, however, if you were in a party, you could not get the hides. Thanks for your interest @pklloveyou

pklloveyou:
So far, no player has reported an error

FrancescoBorzi (approved):

pangolp:
Thank you @pklloveyou

Issue #4403, opened 2021-02-01
Quest 'Alas, Andorhal' is a group quest but not group friendly


Zoidwaffle:
##### CURRENT BEHAVIOUR:

Link to the quest: https://classic.wowhead.com/quest=211/alas-andorhal (Alliance)
And also https://classic.wowhead.com/quest=105/alas-andorhal (Horde)

This is a group/elite quest where players need to kill Araj the Summoner - he drops a chest-like "Araj's Phylactery" that contains a shard you need for the quest.

However, it will despawn once looted by a single player.

This makes it a huge pain to do as you need to wait quite some time for a respawn.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

The phylactery should be lootable by all on the quest.

##### STEPS TO REPRODUCE THE PROBLEM:

1. .go c id 1852
2. .quest add 105 (Horde - or 211 for Alliance)
3. /tar araj
4. .die
5. Loot the phylactery... it despawns right after looting and it shouldn't.

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/tree/1774a322c8f54ee5a59a8dfb2cf4fd8a00c3b55d

##### OPERATING SYSTEM:

Debian 10

##### MODULES:

mod-account-achievements
mod-ah-bot
mod-arac
mod-autobalance
mod-congrats-on-level
mod-morph-all-players
mod-morphsummon
mod-npc-buffer
mod-transmog
mod-weapon-visual

Zoidwaffle:
... and I see it's the same issue for https://classic.wowhead.com/quest=5803/arajs-scarab as well... But whatever can fix the first issue here, can probably trivially fix this one as well.

Helias:
I don't remember if Keira3 provides any field to specific that the quest item need to be shared between players, but I think so, check Keira3 and explore the quest and the item, may it's easy to figure out how to fix.
I suggest to compare a generic quest item and quest that works as you expect and compare it with this

Zoidwaffle:
What makes this special is the thing that you don't kill some NPC and then loot... You kill the NPC that spawns a chest that you then loot.

So the chest is lootable by all and you get the 1 or 2 items you need (depending if you have 1 or 2 quests), but then the chest despawns so no others can finish their quests. I have poked around in the DB with and without Keira3, but I haven't found a way to fix it yet.

Kitzunu:
Fixed

[...]

PR #2750, opened 2020-03-12
Fix:(Core/Gameobjects) Fix so non-consumable goobers don't despawn on use

Kitzunu:
Fixes #2658 

Test performed:

- [x] Builds

- [x] In-game

Branch: Master

How to test:
1. Use any consumable goober without the change and with the change.

**I want to state that this is NOT my doing, it was taken from these commits originally made by Wyrserth on TrinityCore https://github.com/TrinityCore/TrinityCore/commit/50c5d30d13fe12af2a2028baf845fb6eab329a57 & https://github.com/TrinityCore/TrinityCore/commit/2b1e8d135bd3f3a3ddf7f29471cf6116da223175**

PR #4413, opened 2021-02-02
fix(Core/Gameobject): Spell spawned chests should not despawn on loot

Zoidwaffle:
## Changes Proposed:
-  Don't despawn temporary quest-chests when looted once
- This was taken from [TC](https://github.com/TrinityCore/TrinityCore/pull/22124/files)

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4403
- There should be many more quests with the same issue... See other references from [TC](https://github.com/TrinityCore/TrinityCore/pull/22124)

## Tests Performed:
- Compiled the core without errors
- Looted the phylactery from Araj in different ways, it seems to work as intended

## How to Test the Changes:

This is for a single quest only!
- .go c id 1852
- .quest add 105 (Horde - or 211 for Alliance)
- /tar araj
- .die
- Loot the phylactery... it despawns right after looting unless the patch is applied

- [ ] I do not claim to fully understand if these changes have unintentional side effects... Please help testing or sanity-checking if you can!

## Target Branch(es):
- [x] Master

 

Helias (approved):

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -706,9 +706,18 @@ void GameObject::Update(uint32 diff)
...
+                        {
+                                UpdateObjectVisibility();
+                                SetLootState(GO_READY);
+                        }
+                        else
+                        {
+                                SetRespawnTime(0);
+                                Delete();
+                        }
+                        return;
```
FrancescoBorzi, line 720: isn't a bit too much space for indentation? we usually use 4 spaces

Zoidwaffle: Fair enough, I just lined up as with tabs... I changed it to 4 spaces now :smile:

pklloveyou:
```
.ad 32566
```
Can this be loot all the time?

Zoidwaffle:
Apparently, yes... The umbrella seems to despawn, but the picnic grill can be looted as many times as you want.... It stayed there for a few minutes or so before disappearing.

So... Is that good or bad? :eyes:

pklloveyou:
It should disappear immediately after the loot

pklloveyou:
I remember AZ restoring the code a long time ago

Zoidwaffle:
Anyway, the picnic grill should be lootable as far as I can see here https://www.youtube.com/watch?v=KA0_zE_-OZk - so that's actually an improvement I believe.

So far it seems like this patch does fix several items - but does it break stuff? Well, I did also leave a warning in the pull request, I have no idea if it breaks more than it fixes :grin:

Kitzunu:
From code logic this shouldn't break anything.

I think @pklloveyou is referring to this PR https://github.com/azerothcore/azerothcore-wotlk/pull/2750 which was reverted

pklloveyou:
YES

pklloveyou:
If only we knew all the affected spells, we could test them all

Footman:
This pull request is not enough. There needs to be done more thing that depend on restockTime from DBC. Because right now you dont refill loot into chest after it was hidden here. TC did good job there already

claudiodfc:
### Tested Sucessfully!

1. Killed Araj while having quest 211 as an Alliance character
2. Looted the Araj's Phylactery Shard item
3. Gameobject doesn't despawn after being looted

image

Issue #4416, opened 2021-02-02
The loot party flag is missing on some quest items.


pangolp:
##### CURRENT BEHAVIOUR:

Currently, quest items, which are obtained by killing a creature, some of them cannot be obtained by all players in the group. This is due to the fact that a flag, 2048, has not been added to these items, which allows everyone to obtain this item. But, not all the items can be benefited by this flag, they are only one group and I don't think they are all.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

When in a group, all members of the group should be able to obtain the quest item, as long as it is of the "loot party" category.
![bark_of_the_walkers](https://user-images.githubusercontent.com/2810187/106635798-029c5e00-6560-11eb-90cd-a8fb68dddc34.png)

##### STEPS TO REPRODUCE THE PROBLEM:

1. Create a party with another person or several persons (no band / raid)
2. Look for a mission where you must obtain an item by killing an npc.
3. You may notice that only 1 succeeds and then the bag is left, but nothing more can be done. Although, perhaps, that error can be fixed in a [current pull request](https://github.com/azerothcore/azerothcore-wotlk/pull/4409) and what you will observe is that only one can get it, when in fact there should be several.

##### AC HASH/COMMIT:

https://github.com/azerothcore/azerothcore-wotlk/commit/4409ce38bcf30dc6208e623e15c44f8556d068ad

##### OPERATING SYSTEM:

- Windows 10

pangolp:
I created the report, and now I will proceed to create the pull request, with the first 38 items. I want to do it in parts. So that the changes can be reflected little by little, and not all at once. I also leave below 2 examples. Of 2 items, which has the loot party, but only 1 of them, has the flag.
```sql
SELECT `entry`, `name`, `flags` FROM `item_template` WHERE `entry` IN (36746, 22889);
```
```
| entry | name                      | flags |
|-------|---------------------------|-------|
| 22889 | Vial of Moth Blood        | 0     |
| 36746 | Goramosh's Strange Device | 2048  |
```

Kitzunu:
Closing this as discussed on Github some days ago.

To sum it up, it is basically wrong for WotLK and even for the most part up to Retail Today.

The flag is for the most part only applicable for quests where you have to kill a specific person, i.e "Kill Hogger, Get Hogger's Head" and it is not applicable for quests like "Collect 15 boar legs".

pangolp:
[...]

PR #4417, opened 2021-02-02
[ WIP ] Adding the loot party flag to the missing items.

pangolp:
Although not all items have the flag, it is preferable to apply the SQL in this way, because it is also simpler to revert in the future.
If something were to go wrong, which I see as difficult, given that I am testing it beforehand, it would be easy to reverse.

## Changes Proposed:
-  Update the flag of the items below.

## Issues Addressed:
- https://github.com/azerothcore/azerothcore-wotlk/issues/4416

## Tests Performed:
- You can try, any item in the list, which does not have the flag 2048. And you will see that it is only obtained by 1 of the group.

## Target Branch(es):
- [x] Master

 

pangolp:
I am going to close it better, because I am going to rephrase it.

Issue #4415, opened 2021-02-02
Quest items lootable by multiple party members.


55Honey:
### CURRENT BEHAVIOUR:
When looted in perfect sync (Interact with target activated via keybroadcasting), quest items which should drop only for one toon, can be looted with 2-5 toons. Without perfect sync, only one toon gets the reward.
Tested on Pygmy Surf Crawler and Makrura Spellhide for the quest "A Solvent Spirit".

### EXPECTED BLIZZLIKE BEHAVIOUR:
Only one quest item per corpse should be obtaiable for these kind of collection quests. Not one per toon.

### STEPS TO REPRODUCE THE PROBLEM:
To reproduce it one would need an external program, e.g. Hotkeynet.
Then have it send a key to all windows in perfect sync to make them all loot an once.

### EXTRA NOTES:
Does not affect normal solo play or not even normal party play.

### AC HASH/COMMIT:
https://github.com/chromiecraft/azerothcore-wotlk/commit/b905ce8c354b7cfac899ac3ab67209d2995b4de2

### OPERATING SYSTEM:
OS:
Ubintu 20.04

### MODULES:
Chromiecraft server

### OTHER CUSTOMIZATIONS:
Chromiecraft server

Kitzunu:
I remember seeing the fix for this on TC yesterday... Now I can't find it xd

Kitzunu:
https://github.com/TrinityCore/TrinityCore/commit/586c00fe2e226da99968ebe763d3226c6134658c

FrancescoBorzi:
> [TrinityCore/TrinityCore@586c00f](https://github.com/TrinityCore/TrinityCore/commit/586c00fe2e226da99968ebe763d3226c6134658c)

thanks @Kitzunu . It seems simple, can you please provide a PR?

Kitzunu:
Already working on it :D

Zoidwaffle:
I believe I can confirm the issue - I have seen some loot behavior where items that would usually only drop once, appear to be looted by multiple gnomes (but not all) if interacting as a group. I cannot reproduce it reliably though... I multibox 5 characters at once.

@Kitzunu if you make a fix, I can apply it and try to see if I can trigger the issue.

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/pull/4420 @Zoidwaffle

PR #4420, opened 2021-02-02
fix(Core/Player): Prevent looting already looted items

Kitzunu:
* https://github.com/TrinityCore/TrinityCore/commit/586c00fe2e226da99968ebe763d3226c6134658c

Co-Authored-By: Gildor <521036+jildor@users.noreply.github.com>

## Changes Proposed:
-  Prevent looting already looted items

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4415

## Tests Performed:
- Builds

## How to Test the Changes:

Try steps in https://github.com/azerothcore/azerothcore-wotlk/issues/4415 or

1. get two players in a group
2. .quest add 47 (in two players)
3. .go cre id 475
4. kill npc
5. click with two players the npc to see loot
6. do .reload all for lag server (you can do 2-3 times in a row)
7. click with two players the quest item before reload all finish
8. two players get the item

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

pklloveyou:
#4409 ？？Does it work the same way?

pklloveyou:
``` if (!item || (item && item->is_looted)) ``` Is it necessary to write like this?

pangolp:
> `if (!item || (item && item->is_looted))` Is it necessary to write like this?

If I remember correctly, what the parentheses usually do and give priority. Actually, it is as if they were 2 conditionals, instead of 3. Because the 2 in the parenthesis, are counted as one, which are then compared with the other. At least in some languages, it works that way.

Kitzunu:
> `if (!item || (item && item->is_looted))` Is it necessary to write like this?

Nope, not needed. It is already a safe check
Because !item would not be null if item->is_looted pass

Issue #4528, opened 2021-02-10
DB/Gameobject: Tristanfall Pumpkin


Stifler82:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/63

##### CONTENT PHASE:
1-19

##### SMALL DESCRIPTION:
GameObject: [Tirisfal Pumpkin](https://wotlkdb.com/?object=375) at **Solliden Farmstead** in [Tirisfal Glades](https://wotlkdb.com/?zone=85) I think it has a wrong respawn time for 3.3.5a, because their respawn time is after 10+ minutes currently and I suppose it should be less than 10 minutes.

This GameObject in question is used to complete:
Quest: [Fields of Grief](https://wotlkdb.com/?quest=365)
The Quest requires you to loot 10 [Pumpkins ](https://wotlkdb.com/?item=2846) from [Tirisfal Pumpkin](https://wotlkdb.com/?object=375)
The Quest can be taken and delivered by:
**NPC_Start:** [Deathguard Simmer](https://wotlkdb.com/?npc=1519)
**NPC_End:** [Apothecary Johaan](https://wotlkdb.com/?npc=1518)

##### EXPECTED BLIZZLIKE BEHAVIOUR:
I assume (because I don't know for sure), that the Respawn Time of this Game Object should be around 5 minutes.

##### CURRENT BEHAVIOUR:
10+ minute spawn, too long.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Go to [Tirisfal Glades](https://wotlkdb.com/?zone=85) **Solliden Farmstead** Area
2. Take the Quest: [Fields of Grief](https://wotlkdb.com/?quest=365) From: **NPC_Start:** [Deathguard Simmer](https://wotlkdb.com/?npc=1519)
3. Loot some pumpkins 
4. And wait for its respawn.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/b905ce8c354b7cfac899ac3ab67209d2995b4de2

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

Stifler82:
[...]

PR #4529, opened 2021-02-10
fix(DB/gameobject): Improve quest Fields of Grief

Zoidwaffle:
## Changes Proposed:
- ~~Increase spawn locations from 10 -> 32 (let's make it look like a field and not a tiny garden with pumpkins)~~
-  ~~Reduce respawn time from 700s -> 120s~~
- ~~This has been taken from the current TC database (no other sources found)~~
- https://classicdb.ch/?item=2846#comments - says spawn around 5 mins, currently it's 700s
- https://classic.wowhead.com/quest=365/fields-of-grief#comments - 10 locations shown

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4528

Thanks to @Kitzunu and @nicoaravena for helping sorting things out.

## Tests Performed:
- ~~Applied SQL, noticed the many pumpkins~~
- Applied SQL, faster respawn of pumpkins

## How to Test the Changes:

- `.quest add 365`
- `.tele sollidenf`
- loot pumpkins, notice respawn is 5 minutes

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Zoidwaffle:
@Kitzunu - I need to mention, that on TC these are pooled. But there are 32 spawn-locations and the pool is 30... So... Well... I'm not sure that part is worth carrying over. It shouldn't give a visual difference that matters IMHO - we could just remove 2 locations instead and keep it at 30 fast growing pumpkins in fixed locations.

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
```diff
@@ -0,0 +1,39 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1612957477367642592');
+
+-- Lower respawn of and add more spawn locations for Tirisfal Pumpkin
+￼
```
Kitzunu, line 4: remove weird symbol

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
```diff
@@ -0,0 +1,39 @@
...
+DELETE FROM `gameobject` WHERE `id`=375;
+
+INSERT INTO `gameobject` (`id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `position_x`, `position_y`, `position_z`, `orientation`, `rotation0`, `rotation1`, `rotation2`, `rotation3`, `spawntimesecs`, `animprogress`, `state`, `ScriptName`, `VerifiedBuild`) VALUES
```
Kitzunu, line 7: missing GUID

Kitzunu (changes requested):

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
Zoidwaffle: Dammit, seems like some invisible stuff is added when I copy/paste stuff from the browser (I based this on a previous PR)

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
Zoidwaffle: I added those now.

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
```diff
@@ -0,0 +1,39 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1612957477367642592');
+
+-- Lower respawn of and add more spawn locations for Tirisfal Pumpkin
+
+DELETE FROM `gameobject` WHERE `id`=375 AND `guid` IN (45042, 45043, 45157, 45194, 45195, 45196, 45197, 45198, 45200, 45201);
+
+INSERT INTO `gameobject` (`id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `position_x`, `position_y`, `position_z`, `orientation`, `rotation0`, `rotation1`, `rotation2`, `rotation3`, `spawntimesecs`, `animprogress`, `state`, `ScriptName`, `VerifiedBuild`) VALUES
```
Kitzunu, line 7: You are still missing GUID.

You are inserting 31 rows (Which are missing GUID) and deleting 10 GUIDS that are not being inserted here?

nicoaravena:
Careful, the TC fix, is wrong, 120 seconds and 32 respawns are from Cata, the correct one is 10 and 5 minutes.

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
Zoidwaffle: OK, then I missed the point. I delete the old ones and let the database itself handle what numbers it wanted to assign since it's auto-incremented.

Should I instead reuse the first 10 GUID's and then add the later spawns with higher GUIDs?

Zoidwaffle:
Oh, do you have a link?

It will be trivial to modify it to just lower the respawn time from 700s -> 300s and keep the current spawning locations.

nicoaravena:
Let me search it, because I did the research a month ago because I have the same doubt while leveling on chromiecraft.

Zoidwaffle:
OK, so if @nicoaravena is right, then the fix should simply be this:

```
UPDATE `gameobject` SET `spawntimesecs`=300 WHERE `guid` IN (45042, 45043, 45157, 45194, 45195, 45196, 45197, 45198, 45200, 45201);
```

data/sql/updates/pending_db_world/rev_1612957477367642592.sql
Zoidwaffle: This has been scrapped, so closing!

Zoidwaffle:
Yep, I ended up rolling most of the stuff back - now it should, hopefully, be pretty close to Classic at least... Probably better than close to Cata.

Thanks for all the help, I updated the description with new links etc..

nicoaravena:
This is a comment from TBC: https://www.wowhead.com/item=2846/tirisfal-pumpkin#comments:id=265015

But I can't remember where I read the full detail. Just found that the quest was added in [1.11](https://www.wowhead.com/quest=407/fields-of-grief) then revamped in patch [4.0.3](https://www.wowhead.com/quest=24975/fields-of-grief), removing it and added again from scratch.
imagen

pklloveyou:
Waiting for test mark

pklloveyou:
Tested

FrancescoBorzi (approved):

PR #4577, opened 2021-02-15
fix(core/Scripts): 'The Attack!' Quest Improvements/NPC Spawn Adjust

Voxstrasza:
## Changes Proposed:
-  During the quest 'The Attack!', Tyrion addresses the player, but it shows as '$n'.  Script change to change that to the actual player's name
-  The Priestess isn't a real NPC - The robot during this quest turns into the NPC.  She was spawned in the library (essentially making two).  Deleting the spawn removed the NPC and is only used for the event.
- Imported script change from [TrinityCore](https://github.com/TrinityCore/TrinityCore/pull/25613)
- Not part of the quest, but the same area - NPC Alicia was spawned in the wall.  Adjusted the spawn point.

## Tests Performed:
- Build, applied SQL, tested the quest.  It works well.

## How to Test the Changes:
1. .go c 43667
2. .quest reward 2746
3. Pick up the quest from Tyrion, read the text, watch the event.

## Target Branch(es):
- [x] Master


pklloveyou:
Tested

FrancescoBorzi (approved):

PR #3815, opened 2020-11-30
[WIP] Cpp modernization

namreeb:
## CHANGES PROPOSED:
-  ACE dependency removed
-  New use of modern C++ threading and async features
-  Boost.Asio dependency introduced

## ISSUES ADDRESSED:
- Closes #3451 #99 

## TESTS PERFORMED:
-  Build works on Windows
-

## KNOWN ISSUES AND TODO LIST:

- [ ] ASIO support
- [ ] Remove ACE
- [ ] CI update

## Target branch(es):
- [x] Master

 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

Issue #3451, opened 2020-09-08
Remove ACE dependence [$300]


Winfidonarleyan:
# TASK
- [x] `ASSERT` - #4950
- [x] `Configuration` - #4608
- [x] `Method_Request` - #3459
- [x] `Stack_Trace` #6574
- [x] `Task` #5454
- [x] `Guard_T` - #4821
- [x] `Event_Handler` (signal) - #4877
- [x] `Based` #6574
- [x] `TSS_T` #5454
- [x] `INET_Addr` #6574
- [x] `Reactor` #5679
- [x] `Acceptor` #6574
- [x] `Future` #5611
- [x] `Queue` #5611
- [x] `Basic_Types`
- [x] `Semaphore`  - #4951
- [x] `AutoPtr` - #3454
- [x] `OS_NS_time` - #3455
- [x] `Atomic_Op` - 4a9476b
- `Thread` (there is a separate issue for this with a 150$ bounty: https://github.com/azerothcore/azerothcore-wotlk/issues/99)

**The list will be replenished**

# Steps
- [x] 1. Update `acelite`, set lib to `STATIC` and disable incorrect fix `std::auto_ptr` #3450
- [x] 2. Rework `Config` #3453 
- [x] 3. Delete `AutoPtr` #3454
- [x] 4. Replace `Ace Atomic` 4a9476b
- [x] 5. Replace `Ace Time` #3455
- [x] 6. Replace `ACE_Method_Request` #3459
- [x] 7. Replace `ACE_Configuration` #4608
- [x] 8. Replace `ACE_Thread_Mutex` | `ACE_GUARD` #4821
- [x] 9. Replace `ACE_Event_Handler` #4877
- [x] 10. Replace `ACE_ASSERT` #4950
- [x] 11. Delete `ACE_Semaphore` #4951
- [x] 12. Delete `ACE_TSS` #5454
- [x] 13. Delete `Future` and `Queue` #5611
- [ ] 14. `The end` #6980 and #5679

Mitradis:
@Winfidonarleyan plz fix last PR.

FrancescoBorzi:
the bounty here is EXCLUDING the Thread part, which has already 120$ in a separate issue: https://github.com/azerothcore/azerothcore-wotlk/issues/99

FrancescoBorzi:
[...]

PR #4608, opened 2021-02-19
feat(Core/Config): rework config and delete ACE inherited

Winfidonarleyan:
## Changes Proposed:
- Delete `ACE_Configuration` inherited
- Full switch to std lib
- Cherrypick `StringConvert` from TrinityCore
- Add CMake macro for copy default and modules configs
- Add new getters `GetOption<bool>` and etc
- Add info about all modules configs after generating sources
- Add a new folder for all configs `configs`
- Add a new folder for all module configs `configs/modules`
---
- New folder for default configs non Windows systems
image
- New folder for modules configs non Windows systems
image

## Issues Addressed:
- Part #3451
- Part #3815

## Tests Performed:
- Builds and in-game

## Target Branch(es):
- [x] Master

----
## Media
1. New folder `configs` in Windows
image
2. New folder `configs/modules` in Windows
image
3. Config list after CMake source generated
image
4. Testing in game
image


FrancescoBorzi:
@Winfidonarleyan the build is failing

src/common/Configuration/Config.h
```diff
@@ -17,40 +19,46 @@ class ConfigMgr
...
+    std::string const GetConfigPath();
+    std::vector<std::string> const& GetArguments() const;
+    std::vector<std::string> GetKeysByString(std::string const& name);
 
-    std::string GetStringDefault(std::string const& name, const std::string& def, bool logUnused = true);
-    bool GetBoolDefault(std::string const& name, bool def, bool logUnused = true);
-    int GetIntDefault(std::string const& name, int def, bool logUnused = true);
-    float GetFloatDefault(std::string const& name, float def, bool logUnused = true);
+    template<class T>
+    T GetOption(std::string const& name, T const& def, bool showLogs = true) const;
```
FrancescoBorzi, line 36: @Winfidonarleyan I like the approach using generic types (templates). However, I think we should also keep the old methods and mark them as deprecated so we won't break modules and can give the opportunity to migrate them without failures. EXAMPLE:

```cpp
// @deprecated DO NOT USE - use GetOption<float> instead.
float GetFloatDefault(std::string const& name, float def, bool logUnused = true)
{
    return GetOption<float>(name, def, logUnused);
}
```

Kitzunu: Or the owner of the modules update to the new standard.

Perhaps it could be kept for a transitional period, a couple of weeks. But after that imo we dont need duplicate code

FrancescoBorzi: yes it's definitely for a transitional period, but 2 weeks is not enough

Viste:
image

Mitradis:
jv5l3v9svse-1-kopiya

Mitradis:
Tested on windows, work.

Winfidonarleyan:
Done

FrancescoBorzi:
@azerothcore/developers  @azerothcore/testers @pklloveyou @mpfans please have a look at this

nicoaravena:
Using ubuntu 20.04 and g++, works fine, running without errors also no warnings while compiling.

FrancescoBorzi:
@nicoaravena @Mitradis have you tried using the config? e.g. changing a value, then doing a `.reload conf` and check if the change is reflected

nicoaravena:
imagen
Have this error with the command.

FrancescoBorzi:
> imagen
> Have this error with the command.

Thanks @nicoaravena for your feedback. Summoning @Winfidonarleyan

Winfidonarleyan:
@nicoaravena retry

Mitradis:
Work

nicoaravena:
@Winfidonarleyan now works good.

mpfans:
update to https://github.com/azerothcore/azerothcore-wotlk/commit/b9e132a9416197a78d72e08780c4cead2d039219
and then Merge this PR,Compilation error
image

Winfidonarleyan:
Re run cmake

mpfans:
ubuntu 18.04  test failed....

image

Winfidonarleyan:
What is your clang? Show full CMake clean log

FrancescoBorzi:
@mpfans you probably have a compiler that is too old. It compiles fine on Ubuntu 18.04 using `clang 6` and `gcc 8.4` and higher versions. Older versions are not supported

Issue #4514, opened 2021-02-09
Most vanilla dungeon bosses, and raid creatures are not correctly immune to Crowd Control


T1ti:
##### CURRENT BEHAVIOUR:
- I have not tested that extensively on every dungeon bosses. But I can say that from what I have tested, you can CC all bosses in stratholme, scholomance, dire maul north... pretty much fears, roots, stuns etc all work. It makes all of them stupidly easy
It's ridiculous, you can even cc raid mobs 
![image](https://user-images.githubusercontent.com/40864460/107321870-a2e40c80-6aa3-11eb-9789-6b749da12156.png) 

##### EXPECTED BLIZZLIKE BEHAVIOUR:
Most of those high level dungeons bosses should be immune to CC. I can't say precisely which ones aren't but anything medium/high level supposed to be strong is usualy immune, especially if they are bigger sized.

Making a list would take super long, I suggest just copying the data from a core that got it right, vmangos is the most accurate for vanilla content.

##### STEPS TO REPRODUCE THE PROBLEM:

1. test CC abilities against creatures that are supposed to be immune.

##### AC HASH/COMMIT:
c5400b301259e048cfb89d3e591bef565d51cec3

##### OPERATING SYSTEM:
Windows 10

BarbzYHOOL:
yeah, we noted this when reworking bosses, if u want to fix it, hit me up on discord i'll give u a procedure and then you can make a PR out of it, what do you think?

Zoidwaffle:
Hey, I basically did what @T1ti suggested - that is, I looked in the vmangos DB and compared creature_template entry+name with AC.

Then I took the changes to mechanic_immune_mask, if any, and assumed vmangos being right.

It's a fairly long and boring list to check...

PR #4630, opened 2021-02-22
fix(DB/creatures): Update CC immunities for multiple creatures

Zoidwaffle:
## Changes Proposed:
Align `mechanic_immune_mask` in `creature_template` with vmangos (assuming they have better data).

To do this, some conversions and assumptions are being made.

The bit masks explained, left side is AC and right is vmangos where there are diferences.

```
    MECHANIC_NONE                  = 0,
    MECHANIC_CHARM                 = 1,
    MECHANIC_DISORIENTED           = 2 = MECHANIC_CONFUSED
    MECHANIC_DISARM                = 3,
    MECHANIC_DISTRACT              = 4,
    MECHANIC_FEAR                  = 5,
    MECHANIC_GRIP                  = 6 = MECHANIC_FUMBLE
    MECHANIC_ROOT                  = 7,
    MECHANIC_SLOW_ATTACK           = 8 = MECHANIC_PACIFY
    MECHANIC_SILENCE               = 9,
    MECHANIC_SLEEP                 = 10,
    MECHANIC_SNARE                 = 11,
    MECHANIC_STUN                  = 12,
    MECHANIC_FREEZE                = 13,
    MECHANIC_KNOCKOUT              = 14,
    MECHANIC_BLEED                 = 15,
    MECHANIC_BANDAGE               = 16,
    MECHANIC_POLYMORPH             = 17,
    MECHANIC_BANISH                = 18,
    MECHANIC_SHIELD                = 19,
    MECHANIC_SHACKLE               = 20,
    MECHANIC_MOUNT                 = 21,
    MECHANIC_INFECTED              = 22 = MECHANIC_PERSUADE
    MECHANIC_TURN                  = 23,
    MECHANIC_HORROR                = 24,
    MECHANIC_INVULNERABILITY       = 25,
    MECHANIC_INTERRUPT             = 26,
    MECHANIC_DAZE                  = 27,
    MECHANIC_DISCOVERY             = 28,
    MECHANIC_IMMUNE_SHIELD         = 29,
    MECHANIC_SAPPED                = 30,
    MECHANIC_ENRAGED               = 31 = MECHANIC_SLOW_CAST_SPEED (not used)
    Not in use                     = 32 = Undocumented
```
Where the names are the same, they get the same value carried over from vmangos to AC.

Bit 2 while having a slightly different name, is treated the same and carried over.

Bit 6, 8, 22, 31 and 32 from vmangos are ignored.

Bit 6, 8, 22 and 31 from AC are kept.

Another important detail is that a match has to be made for both `entry` and `name` from  `creature_template` before it's considered a match.
## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4514

## Tests Performed:
- I checked a few bosses - and the changes seem "fair" to me, but I have not looked into all creatures at all.
- A lot of thanks to @T1ti for helping figuring out this mess

## How to Test the Changes:

- Try out some CC on creatures mentioned before and after applying the SQL
- Checking exactly what immunities are granted, can be found easily with keira3.

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Kitzunu (approved):
Going to blindly approve this as it will probably improve the overall experience, and in case something were to be wrong it is easily changed :)

Zoidwaffle:
@Kitzunu : yeah, I agree. Verifying all this manually is a pretty hardcore task... but I show changes done exactly because of this reason :grin:

T1ti:
Tldr : big yes for me that's already what I had suggested

So I had already did a deep comaprison of vmangos vs azerothcore data for the mechanic immune masks

- 866 vanilla creatures have a creature immune mask in AC, 1610 in vmangos
-  the main issue isn't even that the creatures that are missing a mask, it's even worse, for many of them masks are incomplete.
vmangos left side, AC right side.
image

image
 For example, jandice barrov only had fear and horror immunities checked, while vmangos had a lot more https://discord.com/channels/@me/809852809028960267/810025828783489034

- there are exactly 1335 differences between the two tables when counting mask value differences. So about 600 more do have a mask but different (probably missing) data

I see that your file is only 1264 lines, there are some missing ? I noticed 1335 entries were different between the two tables, or did you only check cases where vmangos mask was bigger ?

Zoidwaffle:
@T1ti - yeah, I totally tried to implement your idea as I mentioned in your issue :)

I made a quick-and-dirty little shell script to do the comparison.

I didn't care about performance - it's not running all the time (just to say I didn't try to optimize much).

```
#!/bin/bash

# mysql acore_world -B -e 'select entry, mechanic_immune_mask, name from creature_template;' | sed "s/\t/,/g" > acore_mechanic_immune_mask.csv
SOURCE="acore_mechanic_immune_mask.csv"
#mysql vmangos -B -e 'select entry, mechanic_immune_mask, name from creature_template;' | sed "s/\t/,/g" > vmangos_mechanic_immune_mask.csv
TARGET="vmangos_mechanic_immune_mask.csv"

cat "${SOURCE}" | while read i
do
  entry="$(echo "$i" | awk -F ',' {'print $1'} )"
  mechanic_immune_mask="$(echo "$i" | awk -F ',' {'print $2'} )"
  name="$(echo "$i" | awk -F ',' {'print $3'} )"
 
  jentry="$(egrep ^"${entry}," vmangos_mechanic_immune_mask.csv)"
  jmechanic_immune_mask="$(echo "$jentry" | awk -F ',' {'print $2'} )"
  jname="$(echo "$jentry" | awk -F ',' {'print $3'} )"

  if [[ -z ${j} ]]; then
    # match on entry, check name
    if [[ "${name}" == "${jname}" ]]; then
      if [[ "${mechanic_immune_mask}" == "${jmechanic_immune_mask}" ]]; then
        echo "-- ${entry}, ${name}, ${mechanic_immune_mask}, no changes"
      else
        echo "UPDATE \`creature_template\` SET \`mechanic_immune_mask\`=${jmechanic_immune_mask} WHERE \`entry\`=${entry}; -- '${name}' mechanic_immune_mask was ${mechanic_immune_mask}"
      fi
    else
      echo "-- ${entry}, '${name}' no match on name '${jname}'"
    fi
  else
    echo "-- ${entry}, '${name}', no match on entry"
  fi
done
```
So I took 2 dumps of the relevant tables... then I compared the entry and mask, only if it had a perfect match, would it compare the mechanic_immune_mask. Only if the mask was different, it would produce a line - taking the mask from vmangos.

It will output what happens for every single line btw, so finally I just took the output and selected the relevant stuff using `egrep "^UPDATE"`.

FrancescoBorzi:
ok for me too, @T1ti could you give this a try in-game and check a bunch of creatures? just to double-check, then it's good to go

T1ti:
I tested some monsters and it seems good msot of them are now correctly immune to cc

I decided to verif the mechanics differences between the two emulators to make sure it matches, here are some differences I found which might have some side effects : 
Here's the enum in vmangos : 
- mechanic 8 is `PACIFY` in vmangos and 0 use it
-> In AC, it is `Slow_attack`, and it's probably used more.
Need to check if some creatures used it in AC, if so restore

- mechanic 22 is `PERSUADE `in vmangos and 0 use it.
-> in AC it is used for `INFECTED` - e.g. Frost Fever, Blood Plague etc..
- Enrage (31) seems to be missing from the vmangos enum, creatures that used it can be found easily in AC by filtering _mask = or >1073741824_

Zoidwaffle:
Nice catch, I guess I can make some logic making a more nuanced comparison of immunities... Shell scripting ahoy! Not sure when I can get this done, within a couple of days I hope, as my next evenings are kinda busy

Zoidwaffle:
OK, I have some basics going for this. So we just need to properly map the immunity values to each other.

https://www.azerothcore.org/wiki/creature_template vs. https://github.com/cmangos/issues/wiki/creature_template

As @T1ti mentions, the mask MECHANIC_PACIFY has no equivalent (not judging by name at least). It has the same value as MECHANIC_SLOW_ATTACK in AC. I have lined up the differences here:

MECHANIC_PACIFY vs. MECHANIC_SLOW_ATTACK (well... it could be the same effect, not sure)
MECHANIC_FUMBLE vs. MECHANIC_GRIP (sounds like the same effect)
MECHANIC_CONFUSED vs MECHANIC_DISORIENTED (sounds like the same effect)
MECHANIC_PERSUADE vs MECHANIC_INFECTED (no idea, sounds like different things)

My current guess is:
MECHANIC_CONFUSED vs MECHANIC_DISORIENTED = treated as the same
MECHANIC_FUMBLE vs. MECHANIC_GRIP = treated as the same

MECHANIC_PACIFY ignore this one from vmangos. If MECHANIC_SLOW_ATTACK is set in AC, keep it.
MECHANIC_PERSUADE ignore this one from vmangos. If MECHANIC_GRIP is set in AC, keep it.

There's also some examples of masks not making sense being in vmangos... It doesn't match what I see in the documentation for cmangos at least.

Please see the latest commit - it has some info about what I found doing what I tried to describe above. Before a commit, a lot of warning/debug info can be removed (but now they are easier to find).

T1ti:
I don't see ID 32 in either enums
This is AC's 
image

Zoidwaffle:
Interestingly... My check for bitmask 32 in AC does nothing - that is, none are found. Which is good. I just made that check since it was needed in vmangos.

But for vmangos, it's found for 134 creatures. Which is... well... interesting. It seems to be wrong at least - or equivalent to "immune to everything and then some".

I'm pretty tired of these numbers by now tbh, but I think it looks correct. The only thing is if my assumptions about some of the mechanics being the same, based on almost similar names, is correct or not.

T1ti:
You're right, 161 creatures in vmangos have mask 32, none seem to use 31

Zoidwaffle:
I didn't consider the vmangos enum being wrong... But hey, then my code was at least correct in that regard!

Please see the updated information in the original description above - I believe it's ready to merge now.

Zoidwaffle:
@T1ti - did you have time to look at this again?

T1ti:
What do you need me to look at ?

Zoidwaffle:
Ah, just if it looks fine or not. If it seems fine, it should get a "ready to be tested/reviewed" label.

T1ti:
Fyi this is fumble, it's an old alpha spell
https://classicdb.ch/?spell=593
https://classicdb.ch/?spell=5917

So fumble = chance to hit reduction imo

pacify should be unable to attack :
 https://classicdb.ch/?spell=10730
It's kinda the same as attack speed can't be reduced

So fumble and grip should not treated as the same, otherwise the rest looks good

Zoidwaffle:
@T1ti thanks for figuring this stuff out. I have updated it so MECHANIC_GRIP  != MECHANIC_FUMBLE and the old AC bit is carried over.

Zoidwaffle:
I'm not sure how to proceed with this one.

Testing all mobs 1 by 1 is... while not impossible, almost the same as saying this will never get merged.

Everything changed is trivial to roll back - the old bitmasks are in the commit for that very reason.

I suggest we move forward - and if we broke a few mobs, we can easily fix those. The sooner the better - lot's of players testing these mobs in ChromieCraft.

FrancescoBorzi:
@Zoidwaffle agreed, as long as you promise to be around and be ready to fix in case something bad happens :)

Zoidwaffle:
I promise nothing! :grin:

Si1ker:
That's more related to what devs have to do day by day with AC, if something happens then somebody's gonna fix it, as it always has been. Revert is always an option. Next to merge.
@Zoidwaffle @FrancescoBorzi

Si1ker (approved):

T1ti:
Anyways we went from several thousands of errors to a possible few so

JamesGent:
Make sure to let Thunderclap work cc testing with @T1ti.

T1ti:
I can confirm there is an issue with the attack speed reduction immunity, all of the mobs were immune to thunderclap in the Deadmines

Zoidwaffle:
So we kept the flag from AC and it's wrong? From a quick glance, I could see 7 Deadmines bosses with that mask. If we are sure they shouldn't be immune, it's easy enough to fix...

https://github.com/azerothcore/azerothcore-wotlk/pull/4977 - so that should be better, right?

T1ti:
Very few creatues are immune to AS slows, not even sure which but I think you can even debuff msot bosses attack speed, the creatures that are immune to it are the exception

Issue #4518, opened 2021-02-09
Gathering proffession node spawns/respawns/locations are incorrect


The-GhostRider:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/64

##### SMALL DESCRIPTION:

**Mining and herbalism spawns are very strange/incorrect.** 

##### EXPECTED BLIZZLIKE BEHAVIOUR:
**Mining and herb nodes should have fairly regular respawn times and spawn locations. All areas should have at least some base level of mine/herb nodes**

##### CURRENT BEHAVIOUR:
**Some areas seem to have very little to no mining/herb node spawns at all. For example herb nodes in bloodmyst isle are very rare in general, and mining nodes in loch modan seem to be very rare as well. 

Locations of the spawns seem to be correct, frequency seems to be off, as are respawn times. 
I'm not aware of the blizzlike respawn times but from wow classic it seems to be about 5-30 minutes. I have had herbs not respawn in >30 min. I have also been told (circumstantially) of some herbs respawning in seconds in elwynn forest.**

##### STEPS TO REPRODUCE THE PROBLEM:

**1. The gathermate and gathermate data will show potential node spawn locations. In some zones you wont see a herb/mining node spawn in almost all (99%) of these locations despite the zone being empty
Gathermate: https://www.gnarlyguides.com/wotlk/addons/gathermate/
Gathermate data 3.3.5a : https://www.gnarlyguides.com/wotlk/addons/gathermate-data/
2. Some herbs in azuremyst isle take longer than 30 min to respawn
3. Most if not all node locations for herbalism in bloodmyst isle and mining in loch modan are unpopulated. (check against gathermate addon data)** 

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/b905ce8c354b7cfac899ac3ab67209d2995b4de2

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

Stifler82:
I confirm, there are not enough Herbs and Minerals on the map.
**This is all I have found.**

[...]

PR #4636, opened 2021-02-23
fix(DB/object): Pool copper veins on Bloodmyst Isle

Zoidwaffle:
## Changes Proposed:
-  Pool all copper veins (there are only 3 of them, so the direct value of this PR is limited... but it makes sense if more spawn points are added later)

## Issues Addressed:
- This is a nicer way to control mining nodes
- Relevant for AzerothCore issue #4518 (but it does not close it)

## Tests Performed:
- Teleport around

## How to Test the Changes:

```
.go xyz -2411.3 -11965 13.399 530 -- 120324 Bloodmyst Isle, Copper Vein, spawn 1
.go xyz -2614.69 -12192.1 28.0703 530 -- 120325 Bloodmyst Isle, Copper Vein, spawn 2
.go xyz -2623.13 -11772.2 9.2479 530 -- 120363 Bloodmyst Isle, Copper Vein, spawn 
```

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1614071599270991320.sql
+++ b/data/sql/updates/pending_db_world/rev_1614071599270991320.sql
@@ -1,0 +1,18 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1614071599270991320');
+
+-- Create pooling for Copper Veins on Bloodmyst Isle
+
+DELETE FROM `pool_template` WHERE `entry`=11646;
+INSERT INTO `pool_template` (`entry`,`max_limit`,`description`) VALUES
+(11646,1,"Copper Veins - Bloodmyst Isle");
+
+-- Add gameobjects to pools
+DELETE FROM `pool_gameobject` WHERE `guid` IN (120324,120325,120363);
+
+INSERT INTO `pool_gameobject` (`guid`,`pool_entry`,`chance`,`description`) VALUES
+(120324,11645,0,"Bloodmyst Isle, Copper Vein, spawn 1"),
+(120325,11645,0,"Bloodmyst Isle, Copper Vein, spawn 2"),
+(120363,11645,0,"Bloodmyst Isle, Copper Vein, spawn 3");
+
+-- Lower respawn of veins to 1 minute
+UPDATE `gameobject` SET `spawntimesecs`=60 WHERE `guid` IN (120324,120325,120363);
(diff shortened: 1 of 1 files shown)


Stifler82 (commented):
you have to put more spawn points here, they are too few ..
at least another 7 - 8 at least..

WoWScrnShot_022321_190107

Zoidwaffle:
@Stifler82 - I'm already working on a follow up... This is just making a pool.

I have 180 spawn locations mapped out, just need to sort out some logic before making a PR.

Helias (approved):

FrancescoBorzi:
> @Stifler82 - I'm already working on a follow up... This is just making a pool.
> 
> I have 180 spawn locations mapped out, just need to sort out some logic before making a PR.

are you planning to do it in this PR or will open a new one for that? just wondering, both options are fine.

Zoidwaffle:
I would actually prefer this one to get merged in first - and then make a "generic pool upgrade for a zone" PR afterwards, since it could be relevant to do something similar for... well... all zones. It's a lot of work, but focusing on the lower levels zones first, could be a nice improvement for ChromieCraft.

I'm cross-posting from issue #4518, but here's what I wrote (with a few edits):

---

I spend a bit of time on it - and it's kinda fun to try and make this stuff work automatically... But besides that, there are a few things to consider.

There are 180 spawn points for Bloodmyst divided into something like ~100 copper, 50 tin and 30 silver.

One way would be to simply make a giant pool with all of these... and if only 20 spawn, relatively speaking, it's most likely to be copper etc., making the "rare" veins rare.

Another solution, as I can see they went for in the Ghostland zone, would be making ~100 nodes (all the copper locations) and then make them spawn where there's 70% for copper, 20% for tin and 10% for silver...

The "make all 180 spawns" would bring the map into pretty much exactly matching the "Gathermate" addon (I literally took the coordinates from there). The "make spawns able to be 3 things" would perhaps be more elegant, at the cost of a slightly more advanced db-structure.

Personally I might prefer to just use all 180 spawns and call it a day. Then if weird spawns are found, they can just be removed or relocated a bit...

Either way, both options are superior as compared to now IMHO... And I assume something similar could/should be be done for more zones, so I'm trying to see if I can make this "the AC way of handling nodes for herbs/minerals".

Anyone with strong preferences regarding this?

Issue #2510, opened 2019-12-17
Mining/Herb Node Distance Fix [$150]


heyaapl:
##### SMALL DESCRIPTION:
I've been working for the better part of four months cleaning up all GOB nodes (Ore, Herbs, Chests, Quest Objects, etc) to reduce the over-spawning issues by remapping the entire GOB Pool tables. However, I'm now running into issues with how AzerothCore spawns in new nodes. It arbitrarily spawns nodes at the rate defined in the database, without Blizzlike considerations for reducing double-spawning 2 Copper Ore nodes next to each other.

##### EXTRA NOTES:
At a high-level, I'm thinking this would require either an additional table or single column to the existing Pool_Template table that would store a minimum distance that must exist between node spawning. A default value of 0 keeps the current spawning system the same, while a value of say 15 would require 15 points would require a distance of 15 XYZ coordinates is required before a node could spawn next to it. This will help (a) reduce the number of groupings needed in the database to better control spawn rates of nodes (b) allow for easier maintenance and control for server owners who would like to control the spawn rates of nodes.

This will require both core changes and database changes. I can assist with the SQL updates but will need assistance with the core edits and testing.

##### BRANCH(ES):
master

Ali-Khazaee:
As i understand you want to make object do not get spawned near each other as minimum distance 15 yard ?

heyaapl:
> As i understand you want to make object do not get spawned near each other as minimum distance 15 yard ?
This should be configurable by Pool_Template, as depending on the GOB in question, it may change. Also to clarify, I think for this we should avoid using in-game yards and instead a distance of XYZ from each spawn point. But for example, 15 points away makes sense for say Copper Ore nodes in Durotar, but will need to be different for the Christmas Present GOBs below the Christmas Tree or other Quest GOBs in small spaces.

Ali-Khazaee:
Those node spawn points are took from game so they shouldn't get changed also in retail i have seen object get spawns near each other but not in same point, so this shouldn't be configurable by Pool_Template, They should not spawn in same point twice right ?

heyaapl:
[...]

PR #4689, opened 2021-03-03
fix(DB/gameobjects): Mineral veins on Bloodmyst Isle

Zoidwaffle:
## Changes Proposed:
-  Increase amount of mineral veins from 3 -> 122
- Add them all to a single pool, size 30
- Nodes respawn, randomly, after 5 minutes if taken

## Issues Addressed:
- Based upon data from Gathermate - I tried not to spawn similar objects too close, not spawn things in wrong locations etc..

## Tests Performed:
- Applied SQL
- All locations visited...

## How to Test the Changes:

- Learn the proper tracking skill: `.learn 2580` (mining), `.learn 2383` (herbs), `.learn 2481` (treasure)
- `.tele bloodmyst`
- `.mod speed 10` (or faster)
- `.gm fly on`
- Fly around looking for minerals/herbs/treasure

- Even though I checked all locations, I cannot guarantee some NPC isn't spawning inside a node. Hopefully we are at worst talking minor adjustments.

## Target Branch(es):
- [x] Master

 

Zoidwaffle:
I also added a PR for treasures... there is work being done on a PR for herbs as well. So consider testing all 3 PR's at once when they are ready.

data/sql/updates/pending_db_world/rev_1614780210274517793.sql
```diff
@@ -0,0 +1,271 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1614780210274517793');
+
+-- Rewrite positions for minerals for zone: Bloodmyst Isle
+
+-- Before: 3 locations
+-- After: 122 locations
```
Si1ker, line 6: Please avoid including stuff into the SQL update that either are already on the PR or "can be" placed in the PR.
The other short comments below to let the lector know about the intention of the query is ok.

Zoidwaffle: OK, I removed 2 of those lines.

Si1ker (approved):

Zoidwaffle:
OK, so checked out all locations.

Found like 28 mineral veins (not the expected 30)... Video: https://youtu.be/IwwQLq61_FU (remember you can speed it up).

Mitradis:
if you increased amount of mineral veins from 3 -> 122 will this exacerbate this problem?
https://github.com/azerothcore/azerothcore-wotlk/issues/2510
or is it a separate location and therefore this problem will not affect it

Zoidwaffle:
@Mitradis : I can test it later. I made an "updated" version of my "visit everything" script... So it's busy visiting the 285 herb locations and trying to track the amount automatically without me having to do anything. I can try it out with different pool values to see if some weirdness is going on.

Zoidwaffle:
... and a bit more.

In theory.... If anyone went over all the zones, then everything could get "cleaned up" and this issue would hopefully solve itself.

I have made some simple shell scripts to help out in that regard, but it still takes some effort to fix a zone and there's a lot of those... Of course, not all zones are "broken" - from looking at the database investigating this stuff, I can see that some zones got a lot of love and attention already while others are clearly in need of a do-over.

Bloodmyst just happened to be really bad, so it was my first attempt of a major fix in this regard.

If there are other Linux users around, I could share my simple scripts and perhaps more could work on fixing stuff - because fixing a zone is easily ~2 hours of work, even if it's semi-automated...

Zoidwaffle:
bloodmyst_minerals

23 of 30 mineral veins found, this was with ~30 herbs and ~10 treasure activated in the zone.

Zoidwaffle:
This is super weird... I tried increasing it to 50 spawns, but only 29 appeared (total with herbs. etc. 90).

Then 10 of which 8 appeared (total 50 objects)

Back to 30 of which 25 appeared... (total of 40 objects)

I really don't understand exactly what happens - I can understand I see a few nodes less because of weird angles when teleporting around, but I would expect that to be 10% or less.

As my final test.

Spawn 30 herbs, 30 minerals and 10 chests.... then visit every spot and count...

9/10 treasures.
31/30 herbs (wut?)
25/30 minerals.
A total of 65 our of an expected 70 spawns. It's within my 10% expected error margin at least!

bloodmyst_total

So it seems like the numbers set for the pool are close but not accurate. The weird part is 31 herbs when only 30 should be there... If anyone with understanding of the spawning mechanics have some information regarding this, it would be awesome.

TL;DR: I think this PR is an overall improvement, even if the pool numbers are apparently approximated.

Zoidwaffle:
I redid my experiment... visited 435 nodes (yet, there's that many) once more on a restarted core. So the locations should be different when compared to the one above.

28/30 herbs.
9/10 treasure.
29/30 minerals.

A total of 66 of 70 objects found.
bloodmyst_all2

Issue #4701, opened 2021-03-04
Need more Pumpkins in Tirisfal Glades


JamesGent:
Need more Pumpkins in Tirisfal Glades

##### CURRENT BEHAVIOUR:

8 Pumpkins currently meaning 1 player has to wait ~10 mins for the respawn to do the quest.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Should be about 20+ pumpkins so that 2 players can do the quest together and nobody waits. ClassicWoW has the spawns at 21, and I don't see any patches saying they reduced the amount.

https://classic.wowhead.com/object=375/tirisfal-pumpkin

https://classic.wowhead.com/quest=365/fields-of-grief

In this video you can see the recorded pick up 10 pumpkins with ease, and you can see 3+ other players also picking up pumpkins in the same area at a slower rate.

https://www.youtube.com/watch?v=J3stuaLo25I

##### STEPS TO REPRODUCE THE PROBLEM:

1. Go to Tirisfal Glades
2. Pick up Fields of Grief quest (365)
3. Observe lack of pumpkins.

eTzmNcbkrng:
It's already Blizzlike.
https://www.youtube.com/watch?v=8dNMBWh2wxY&t=4m

Zoidwaffle:
Hey, I had this one recently...

https://github.com/azerothcore/azerothcore-wotlk/pull/4529

back then, it was decided that the current version is correct as pre-Cata.

eTzmNcbkrng:
@Zoidwaffle Yeah I was watching that one when you worked on it, wish I had found this video back then.

Si1ker:
So, has #4529 already resolved this issue?

eTzmNcbkrng:
I don’t think so because all that did was reduce the respond to 5 minutes.
Respawn should be 2 minutes and there should be a lot more than 10 nodes.

nicoaravena:
Actually they are 10 pumpkins, 8 in the field, 2 in the tower (as wotlk was), the respawn time was fixed in #4529.
![imagen](https://user-images.githubusercontent.com/4155059/110050296-65853e80-7d32-11eb-9927-3b9032c1fac8.png)


Also the proposal is from classic blizz server, which was changed from original classic to match with the respawn times and the nodes cuantity from cataclysm.

Zoidwaffle:
Well... Personally, I don't mind aligning the quest more with "it makes sense even if not totally as it was back in the days".

[...]

PR #4722, opened 2021-03-05
fix(DB/gameobject): Improve quest Fields of Grief (part2)

Zoidwaffle:
## Changes Proposed:
-  Increase spawn locations from 10->32.
-  Add them to a pool, 20 spawns at once (easy to change this)
- Respawn time lowered from 5->2 minutes

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4701

## Tests Performed:
- Applied SQL ~~, not tested in game~~
- Tested in game

## How to Test the Changes:

- `.quest add 365`
- `.tele sollidenf`
- loot pumpkins, notice respawn is 2 minutes and the spawns are more numerous and in more locations than before

- This was already discussed in: https://github.com/azerothcore/azerothcore-wotlk/pull/4529 where these suggested changes where found wrong.
- This is not a "purist" change, as it would be more "Classic" than "Vanilla". Opinions regarding this might differ - this is merely a suggestion.
- For what it's worth, this is closer to TC than what we have now.

## Target Branch(es):
- [x] Master

 

Si1ker (approved):

FrancescoBorzi:
updated label as the PR has then been tested by @Zoidwaffle

eTzmNcbkrng:
https://github.com/azerothcore/azerothcore-wotlk/issues/4701#issuecomment-791753842
>It's hard to say how valid the data is, being potentially poisoned by private servers etc. and so on.

In [part one](https://www.youtube.com/watch?v=Tyc3S-Bl8oU) of this video series, we can see the player's character selection screen and he is playing on a realm called Eonar, which is a valid EU (his accent puts him in EU) realm.
One of the characters is Terminator, a L49 Warrior.
https://worldofwarcraft.com/en-us/character/eu/eonar/terminator - L10 (stat squish) Mechagnome Warrior in full heirlooms,
I was unable to find/read any of the other characters on the Eonar realm.
Also, the video was uploaded in May 2010 while Wrath was still current.

pklloveyou:
Tested

Zoidwaffle:
@eTzmNcbkrng : OK, so the video is probably fine. But how to interpret what happens...

He picks up all the pumpkins in ~2½ minute. There are several pumpkins visible in the field when done, like 5ish and that's not counting the couple that might by the tower. So a 5 minutes respawn and 10 nodes is definitely wrong (what we have before this PR).

So what do you think? Is this PR fine as it is or should the amount of spawns / respawn rate be adjusted?

eTzmNcbkrng:
I think the number of possible nodes is the biggest unknown at this point.  
At 6:04 - 6:06 we see him loot two nodes next to a cart, at 6:29 we can see one of them has already respawned.

I think this means there is a minimum number of nodes always available, and this number is either different to the maximum number of available nodes (in which case a respawn timer is needed) or it is the same as the maximum number of nodes (in which case a respawn timer is not needed).

Zoidwaffle:
I think this is where we use our interpretive freedom - I have lowered the amount of spawns to 15 (down from 20). So there are fewer pumpkins around and they start respawning after 2 minutes. It should give something reminiscent of what could be seen in the video. The respawn timer and amount of spawns can be adjusted, but I think this feels about right.

Is it perfect? Probably not.

Is it better than before? Yes, I believe so.

necropola:
Just updated my local AC and checked. I found 12 Pumpkins (10 in the field), 1 at the tower and 1 at a stable/farm house (probably missed 3). They respawned after 2 minutes. Looks pretty good to me. Well done.

PR #4045, opened 2020-12-24
fix(Core): Activate creatures and objects during opening cinematics

Si1ker:
This change will result in the capability to see NPCs/GOs during character intro cinematic
### Warning:
* Requires re-running map extractor to extract camera m2 files from data

## Changes Proposed:
-  **Port:** TrinityCore/TrinityCore@87b4533
-  **Adapt:** Porting adapted to AC structure
- **Standardize:** Small cleanup from the original port and add brackets

## Issues Addressed:
- Closes #4039 

## Tests Performed:
- N/A

## How to Test the Changes:

- Create any new character
- Watch the whole cinematic
- The NPCs/GOs should be shown in the cinematic

## Known Issues and TODO List:

- N/A

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #4738, opened 2021-03-06
Fix building with gcc-8.x on debian 10

necropola:
## Changes Proposed:
Fix building on systems with older versions of gcc, e. g. debian 10 with gcc 8.3

## Tests Performed:
- Now builds on `debian 10` with `gcc-8.3` and `-DTOOLS=1`
- Reportedly builds on `Ubuntgu 20.04` with `gcc-9.3.0` and `DTOOLS=1`, too 

## Target Branch(es):
- [x] Master

### DIFF
--- a/src/cmake/platform/unix/settings.cmake
+++ b/src/cmake/platform/unix/settings.cmake
@@ -44,3 +44,10 @@ else()
     INTERFACE
       -D_BUILD_DIRECTIVE="${CMAKE_BUILD_TYPE}")
 endif()
+
+if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.1.0)
+  message(STATUS "UNIX: Older stdc++ library requires stdc++fs, too")
+  target_link_libraries(acore-default-interface
+    INTERFACE
+      stdc++fs)
+endif()


Winfidonarleyan:
Please use https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/cmake/compiler/gcc/settings.cmake

necropola:
> Please use https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/cmake/compiler/gcc/settings.cmake

Globally for all targets (if gcc < 9) or just for the one target that causes the error?

necropola:
I've now added a check for `gcc < 8.4.0` to both `src/cmake/compiler/clang/settings.cmake` and `src/cmake/compiler/gcc/settings.cmake` and add the library to `acore-default-interface`, if true. Still no idea, if this is (these are) the correct place(s). But at least Windows now builds successfully, too.

Helias:
This is a very specific PR, but the build seems ok, what's wrong? with a version of debian it does not work?

necropola:
It's not building on debian 10 with gcc 8.3 and clang-7.0 unless I add `stdc++fs` to the link libraries. Either globally or by modifying just the `CMakefile` for the `map_extractor`.
```
/usr/bin/ld: CMakeFiles/mapextractor.dir/System.cpp.o: in function `ExtractCameraFiles(int, bool)':
System.cpp:(.text+0x68b8): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: CMakeFiles/mapextractor.dir/System.cpp.o: in function `std::filesystem::__cxx11::path::path<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::filesystem::__cxx11::path>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::filesystem::__cxx11::path::format)':
System.cpp:(.text._ZNSt10filesystem7__cxx114pathC2INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE[_ZNSt10filesystem7__cxx114pathC2INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE]+0x8f): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/tools/map_extractor/CMakeFiles/mapextractor.dir/build.make:174: src/tools/map_extractor/mapextractor] Error 1
make[1]: *** [CMakeFiles/Makefile2:941: src/tools/map_extractor/CMakeFiles/mapextractor.dir/all] Error 2
```
It's a miracle to me why it builds/links on Ubuntu 18.04 with gcc-8-4 and clang-6. I have donwloaded the `libstdc++-8-dev` package from Ubuntu and checked the symbols. To me it looks like you would need to add `stdc++fs` on Ubuntou 18.04, too.

necropola:
I've moved the check to `src/cmake/platform/unix/settings.cmake` and changed the version check to `< 9.1.0` since this appears to be the first gcc (and therefore libstdc++) release which did not require an additional library for `std::filesystem`.

necropola:
I suggest to close this pull request since it's clearly beyond my (cmake) capabilities to properly fix the issue. A real fix would need to check whether the C++ standard librrary used by the build system (whichever that is) already contains the `filesystem` functions/symbols (maybe through a compile check) and if not add an additional system library, e. g. `stdc++fs`, either to all targets or just the one currently affected.

Helias:
Man, I really appreciate your try, but I don't know how to help you
I fought with the filesystem library that has been introduced with the Cinematic PR ( https://github.com/azerothcore/azerothcore-wotlk/pull/4045 ) but later we've deprecated the gcc version that didn't support the filesystem library (I tried with experimental/filesystem too but without no luck).

Hope next time to be more helpful, sorry.

necropola:
I just keep my original change in git stash and re-apply after each fetch/rebase. The underlying cause for issues like this are feature-hungry compiler/language designers and developers on one hand and extremely conservative distro maintainers on the other.
My main linux host is (still) running current stable debian 10 and I'm bumping into issues like this all the time. But I already have an ArchlLnux VM which always has the lastest and greatest versions of pretty much everything I need.

FrancescoBorzi:
@Knindzagxg not sure but maybe this PR can solve your issue

necropola:
Not my original PR , but the following fixes it now (since the `stdc++fs` dependency was moved fromn `tools` to `common`):
```diff
diff --git a/src/common/CMakeLists.txt b/src/common/CMakeLists.txt
index 7a2997089..2db021ce5 100644
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -62,6 +62,7 @@ target_link_libraries(common
     openssl
     threads
     jemalloc
+    stdc++fs
     fmt)

 set_target_properties(common
```

Issue #4670, opened 2021-03-01
DB/Creature: Kolkar Drudge and Kolkar Outrunner extreme Fire Resistance


illfated:
##### ISSUE: FACTION SIDE:
![Horde](https://user-images.githubusercontent.com/1884642/108204991-63a96180-711c-11eb-9ed9-138c233070a6.png)

Original report: https://github.com/chromiecraft/chromiecraft/issues/107

This ticket content is based on comments written by @necropola in the linked issue ticket.

##### CURRENT BEHAVIOUR:

- https://wotlkdb.com/?npc=3119/kolkar-drudge
- https://wotlkdb.com/?npc=3120/kolkar-outrunner

The NPC mobs Kolkar Drudge and Kolkar Outrunner resist more than 50% of any Fire Damage thrown at them. Frost Damage works as expected, though.

This is from the combat log of my level 11 Blood Elf Mage (Fireball and Frostbolt are both Rank 2):

![image](https://user-images.githubusercontent.com/38162891/109538972-010f7880-7ac1-11eb-8a80-8ef52b996a75.png)

Fireballs work fine on other NPCs in the area, e. g. Scorpids.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

They should not have any elemental resistance at all …
… but I cannot find any information/source about elemental resistance on NPCs in any database, although I know that there are certain NPCs in game (usually elementals) that are supposed to have high resistance against a particular school of magic.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Go to Kolkar Crag in Southern Durotar near Sen'jin Village.
2. Fight a few of the Kolkar as a Mage.
3. Watch the difference in damage between Frost and Fire spells.
4. Check your Combat Log.

##### EXTRA NOTES:

Based on the discussion in https://github.com/chromiecraft/chromiecraft/issues/107, we have discovered that the issue is currently present in TrinityCore 3.3.5 as well.

I have both a current AC-3.3.5 and a current TC-3.3.5 running locally on my Linux Host and both show the same **funny** fire resistance. The following screenshot is from hitting a Kolkar Drduge with a Fireball Rank 16 from my level 80 Fire Mage on a current TC-3.3.5 server.

![image](https://user-images.githubusercontent.com/7461137/109433417-82f39900-7a10-11eb-9ed7-cbdeb1ee4ed1.png)

Though I don't think just because it is also present in TC makes it correct.

See also the other comments in the linked issue ticket for related information.

[...]

PR #4753, opened 2021-03-07
fix(DB/creature_template): Resistances for level 1-20 creatures

Zoidwaffle:
## Changes Proposed:
-  Align creature resistances for levels 1-20 with vmangos
-  Done after some discussions here https://github.com/azerothcore/azerothcore-wotlk/issues/4670 and https://github.com/chromiecraft/chromiecraft/issues/107

## Issues Addressed:
- Some resistances seem a bit high for low level creatures - no idea of the origin of these numbers, as it's hard to confirm

## Tests Performed:
- Looked at numbers to be changed, it could make sense
- Applied SQL
- Not tested in game

## How to Test the Changes:

- `.go c id <any entry id from the SQL>`
- Check combat log, use appropriate spells

- The SQL looks like it does since I created it using code... Not super pretty, but the intention is showing in a fairly easy to read manner, exactly what will change.
- I prefer to segment these changes, if they are interesting, into ~10 level PR's going forward, not to swamp any reviewers with possibly hundreds of creatures at once

## Target Branch(es):
- [x] Master

 

T1ti:
Changes look good

illfated (approved):
Looks very good to me. I have no improvement suggestions aside from cosmetic ones.

Si1ker:
@Zoidwaffle  From where are taken the vmangos values? or you want just to copy/paste what you saw in another place without proof/evidence in another source that those are the correct ones? just curious..
image

T1ti:
I agree that in this case vmangos isn't a much better source, what it is good for tho is being able to see all the differences and then comparing or looking at the two logically 

So, some changes that look weird : 
-giving 30 shadow resist to Living flame
-giving ~20 to all resistances to water spirit. 
image
- 80 arcane resistance to a demon ? 
image

Si1ker:
Hope @Zoidwaffle has a good blizzlike source before changes.
@T1ti indeed some changes like the ones you mentioned are weird, but no TBC books for end-game woltk 3.3.5a content, please.

T1ti:
The book doesn't have anything that came out later than patch 2.2/2.3 anyways lol

Zoidwaffle:
@Si1ker : I don't have proof of anything. It was the initial assumption that vmangos had better data - perhaps that was wrong. But I tried to make it so all the differences are easy to spot and the list isn't so long it's impossible to check it out.

This was triggered by the Kolkar-creatures having 70 fire resistance which just seems wrong (to me at least) but I don't have any hard evidence for it.

Hacki95:
While the book may not be an "100%" accurate source its also a matter of, "How wrong are we right now?". If it was just blatantly wrong before its better to "guess" a more accurate value than to leave it at the wrong one. Although I wouldn't call using the book guessing.

On that note, can't someone with a retail classic account check certain cases like the Kolkar with 70 Fire Resistence? It wouldn't even need hours of testing but just a few fire spells on those mobs. It would be pretty obvious if Blizzard also uses a high resistence values on those mobs. In the end this wouldn't our new values to be 100% accurate, but it would atleast prove them being closer than before.

In the end its also pretty trivial to change back SQL values if they turn out to not be accurate when wotlk classic comes out. We surely don't have the right values at the moment.

Zoidwaffle:
@T1ti : Yeah, Xabraxxis is removed... and Living Flame - well, if it had been resistance to fire it would probably have been fine, so I removed it as well.

That leaves the water spirit - again, I would like it to have resistance to maybe fire/frost, but not fire/nature/shadow/arcane.... So I also removed that one.

What is left now is mostly some weak resistances being removed, some weak ones being added... The exceptions being some weird OP resistances by Gorefang and some Kolkar... The only other huge change is Dust Devil, where 40 NR might be fine or not... AC says yes, vmangos says no.

I definitely don't think any of the current changes are game-breaking in any way - and lowering the 50+ resistances is like an act of courtesy to new players.

necropola:
> On that note, can't someone with a retail classic account check certain cases like the Kolkar with 70 Fire Resistence? It wouldn't even need hours of testing but just a few fire spells on those mobs. It would be pretty obvious if Blizzard also uses a high resistence values on those mobs. In the end this wouldn't our new values to be 100% accurate, but it would atleast prove them being closer than before.

That is pretty much what I was originally suggesting, i. e. take a few samples on Classic to get a better impression of how much current resistance values are off. In my original report I have listed 23 creatures below level 20 that have/had significant (> 50) reistance and counted a bit over 90 (< level 20 bracket) that have/had any resistance in AC.  

The idea or rather my suggestion for a long-term plan is to clean this (NPC resistances and abilities) up bracket by bracket as ChromieCraft progresses. I further think that manual checks of the combat log in Classic (Classic TBC, ...) are good to get an estimate of the current data quality in AC, but are too much manual work to check all NPCs/creatures of any given bracket.

If we had an Addon (a crossbread of a wowhead client and a combat log analyzer) which would record creature resistances and abilities and had enough people running it on Classsic, Classsic, TBC, ... and uploading the data to a new AC DB. now that woudl be great, right?

Zoidwaffle:
I added "Waiting for feedback".

So do these changes look good or not?

If not, what do you suggest changing?

FrancescoBorzi (approved):

FrancescoBorzi:
codestyle looks good, marked as [Waiting to Be Tested]

T1ti:
I'm pretty sure dust devils are supposed to be immune to nature, so their nature resistance shouldn't matter

Zoidwaffle:
@T1ti - dang, you are sharp! The immunity in AC is 0... but as you said: `spell_school_immune_mask: 8` and 8=nature.

So the resistance should logically be 0 (which it will be if this is merged).

illfated (approved):

Zoidwaffle:
A week has passed - I believe this is good enough to get merged.

I don't think further testing or waiting for feedback is needed. So marking as "Ready to merge" - if you disagree, feel free to change it.

PR #4749, opened 2021-03-07
fix(DB/Core) Fix the CLS damage system and update creature_classlevelstats

Hacki95:
## Changes Proposed:
-  Reenable the min/max damage CLS system in Creature::SelectLevel in Creature.cpp that has been disabled
-  Fix the underlying calculation issue in Creature::CalculateMinMaxDamage in StatSystem.cpp. AttackMultiplier was wrongfully multiplied with AttackSpeed. I used the reverse engineered CLS calculations TC/vmangos/cmangos use.
- Correct several wrong values in creaure_classlevelstats. Multiple damage_base and attackpower values were wrong before. I imported those values from TC's database.
- Correct several wrong DamageModifier values in creature_template. Using TC values for those aswell. The values we had before were mostly a flat 1 or completly off because they were based on the old system. With the new values the calculations lead to the desired result.
- Drop "mindamage", "maxdamage", "minrangedamage",  "maxrangedamage", "attackpower" and "rangedattackpower" from creature_template as those values aren't needed anymore.
- Fix all code that references those columns.
- Implement BaseVariance and RangeVariance (mostly for custom mobs).

## Issues Addressed:
- Closes #4570 

- While no other issues are open at the moment, using the CLS calculation instead of using fixed min/max values could potentially fix a lot of NPCs which had wrong sniffed data before. This bug existed since Commit 1 of AzerothCore :)

## Tests Performed:
- Tested "reload creature_template".
- Used the Spell "Beast Lore (ID=1462) on around 20 different mobs to check their min/max damage values. Those values now correctly change if a mob is for example Level 35 vs Level 36. All values I tested so far have been ~99% to the pre-calculated min/max values so it seems like the new damage_base values are correct or atleast much better than the values we had before.
- Played for a few hours on my offline realm and everything I tried looks great.

## How to Test the Changes:
 - Download the PR, compile the code and apply the DB update.
 - Make a Hunter or simply learn spell 1462. 
 - Fly around the world and check the mobs min/max attack values.

## Known Issues and TODO List:
- I have no way to confirm the values I took from TC are 100% accurate. BUT they are way better than the precalculated values we used before and every mob I tested has been correct.

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #4637, opened 2021-02-23
fix(DB/creatures): Correct BaseAttackTime and RangeAttackTime

Zoidwaffle:
## Changes Proposed:
-  "normalize" attack speeds using a [TC commit](https://github.com/vmangos/core/blob/4a1083821eecdd77129e053e9c77b966c15f6924)
~~-  copy vmangos values as they are hopefully more accurate~~
- Import 2 [commits](https://github.com/vmangos/core/commit/4a1083821eecdd77129e053e9c77b966c15f6924) from [vmangos](https://github.com/vmangos/core/commit/f0a1fbfc87d97d8a876bf569ca9eb09729714cdd) 

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4524

## Tests Performed:
- Checked a few values, I really don't know if either are true.

## How to Test the Changes:

- Check out the before/after values for your favorite changed creature

## Known Issues and TODO List:

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #4767, opened 2021-03-09
fix(DB/creature_template): Set RangeAttackTime to default value

Zoidwaffle:
## Changes Proposed:
- Set  ranged attack times to default for all creatures
- Suggested here: https://github.com/azerothcore/azerothcore-wotlk/pull/4637

## Issues Addressed:
- This was done by TC here: https://github.com/TrinityCore/TrinityCore/commit/b7b1c2df790f1eb9203f084045e90ee5f9e1ebb5
- ~~I'm not sure if we should do the same, so marking this for in need of feedback~~
- This is just to take the discussion, you are very welcome to disagree
- It could be relevant for https://github.com/azerothcore/azerothcore-wotlk/pull/4749

## Tests Performed:
- None

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1615275106559909076.sql
+++ b/data/sql/updates/pending_db_world/rev_1615275106559909076.sql
@@ -1,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1615275106559909076');
+
+-- Reset ranged attack times
+UPDATE `creature_template` SET `RangeAttackTime` = 2000 WHERE `RangeAttackTime` > 0;
--- a/data/sql/updates/pending_db_world/rev_1615275106559909076.sql
+++ b/data/sql/updates/pending_db_world/rev_1615275106559909076.sql
@@ -1,4 +1,4 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1615275106559909076');
 
 -- Reset ranged attack times
-UPDATE `creature_template` SET `RangeAttackTime` = 2000 WHERE `RangeAttackTime` > 0;
+UPDATE `creature_template` SET `RangeAttackTime` = 2000;


Hacki95:
I don't think its a good idea to make a blank 2000 sec for every mob. Have you checked TC or vmangos by chance how many mobs they have with RangeAttacktimes which are different from 2000? Just to compare how many mobs could potentially have a different timer. I think its given that our own values are completly off.

Zoidwaffle:
Yep, made me wonder as well... The TC commit has this comment "Don't see the point of holding old funny values of TBC and classic creatures and at the same time let parser update it every time to default value.". Does that make sense to you (or anyone else)?

I made the commit as it was suggested by @T1ti in #4637 - and I'd rather take it into a new commit. So here we are.

Zoidwaffle:
@Hacki95 - and to answer your questions...

AC has 3065 creatures targeted by this PR.
TC had 3036.

So without investigating further, I think they are pretty comparable. I could find the exact ones with differences if that's interesting... But before that, is the TC comment about "parser update" also relevant for AC?

Hacki95:
> 
> 
> Yep, made me wonder as well... The TC commit has this comment "Don't see the point of holding old funny values of TBC and classic creatures and at the same time let parser update it every time to default value.". Does that make sense to you (or anyone else)?
> 
> I made the commit as it was suggested by @T1ti in #4637 - and I'd rather take it into a new commit. So here we are.

Just guessing but I think it could be that their Sniffed Data gives them those funny values but when they verify them by checking with a parser it always returns them to 2000?

Hacki95:
> 
> 
> @Hacki95 - and to answer your questions...
> 
> AC has 3065 creatures targeted by this PR.
> TC had 3036.
> 
> So without investigating further, I think they are pretty comparable. I could find the exact ones with differences if that's interesting... But before that, is the TC comment about "parser update" also relevant for AC?

Mhh. Yeah I was just a bit put off by the SQL statement changing every single value thats over 0 to 2000, that would mean in return that every single ranged mob has an attack of 2000.

But if thats the value, thats the value. I just thought it sounded weird.

T1ti:
the vmangos parser has 0 for all ranged speeds

data/sql/updates/pending_db_world/rev_1615275106559909076.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1615275106559909076');
+
+-- Reset ranged attack times
+UPDATE `creature_template` SET `RangeAttackTime` = 2000 WHERE `RangeAttackTime` > 0;
```
FrancescoBorzi, line 4: not sure about this, marking as request changes in order to prevent merging this by mistake

FrancescoBorzi (changes requested):

data/sql/updates/pending_db_world/rev_1615275106559909076.sql
Zoidwaffle: Very fair, I'm not sure about it either :smile:

offl:
Here you go https://github.com/TrinityCore/TrinityCore/commit/b7b1c2df790f1eb9203f084045e90ee5f9e1ebb5#commitcomment-48111172

FrancescoBorzi:
thanks @offl !

offl:
You're welcome, here it was changed to 2000 with the same reason https://github.com/TrinityCore/TrinityCore/commit/8ec98b4e137a3c8a38d20bd5c94c1089ede24959#commitcomment-7687020

Zoidwaffle:
We got a nice explanation from offl regarding similar changes in TC.

It's now 15 days ago and no others have commented. I think a decision should be made - should we implement this or not?

My suggestion is "yes", but if not, then this PR can be closed.

FrancescoBorzi:
@Hacki95 are you still against this?

Hacki95:
> @Hacki95 are you still against this?

Didn't see the notification sorry! No it makes complete sense I would approve the changes! Funny that MaNGOS did literally chose random values, 2000 sounds much more reasonable.

Zoidwaffle:
Nice, so I assume we are good to go! I'm not removing the WIP label as @FrancescoBorzi should get the honor along with the requested changes :rocket:

Kitzunu (approved):

PR #4773, opened 2021-03-09
fix(DB/creature_addon): Remove extra attacks from Naxxramas NPCs

Si1ker:
Co-authored by @T1ti & @Hacki95

## Changes Proposed:
- Remove [Double Attack](https://db.rising-gods.de/?spell=18943) aura from Anub'Rehkan and Venom Stalkers
- Remove [Double Attack](https://db.rising-gods.de/?spell=19818) aura from Gluth

## Issues Addressed:
- Closes #4751 

## Tests Performed:
- Tested in-game

## How to Test the Changes:

Execute -> restart -> test
.go c (Use any of the GUIDs below in here)

GUIDs = 127814, 127782, 127868, 127869, 127870, 127871

## Known Issues and TODO List:

- N/A

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #4778, opened 2021-03-09
Revert "fix(DB/creature_addon): Remove extra attacks from Naxxramas NPCs"

Si1ker:
Reverts azerothcore/azerothcore-wotlk#4773


Si1ker:
Will PR the re-addition of those values into the DB in some minutes

Issue #4791, opened 2021-03-11
Feral Charge - Cat travel speed should be faster


laasker:
##### CURRENT BEHAVIOUR:
Druid spell https://wowgaming.altervista.org/aowow/?spell=49376 (Feral Charge - Cat) have its animation/travel time slower than intended

Right now its like that:

https://www.youtube.com/watch?v=KrWjRh578vM

##### EXPECTED BLIZZLIKE BEHAVIOUR:
Should have a faster animation like that:   (retail videos btw)

https://www.youtube.com/watch?v=lLe3KJG2teQ&ab_channel=ZybakTV at 0:20 and at 0:37 

https://www.youtube.com/watch?v=XcQdFJjL4SQ&ab_channel=ZybakTV at 1:12

https://www.youtube.com/watch?v=-vt1gim7obQ&t=102s&ab_channel=wowmoviesandvideos at 4:08 and 9:20

##### STEPS TO REPRODUCE THE PROBLEM:
1. 
2. 
3. 

##### AC HASH/COMMIT:
https://github.com/azerothcore/azerothcore-wotlk/commit/51342ef9b500

##### OPERATING SYSTEM:
Windows 10

Helias:
I did a PR, test it if you can and read also the section test performed pls

PR #4795, opened 2021-03-11
fix(Core/SpellEffect): increase speed of cat charge

Helias:
I don't know if my changes are really useful or not, it's really hard to get if the speed is increased actually or not.
I don't think that I made some break changes, I did several tests on this.

## Changes Proposed:
-  increase the speed of Feral Charge - Cat

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4791

## Tests Performed:
- tested several times in-game changing a lot of the parameters but apparently, there is a "limit" of the VELOCITY.
- There is a max velocity value that you can't overcome, this maximum is assigned considering the client limit as the comments says in MoveSplineInit.cpp line 123 `// limit the speed in the same way the client does`.

## How to Test the Changes:
.cast 768
.cast 49376 on a target

## Target Branch(es):
- [x] Master

 

pklloveyou:
I feel that the speed has not improved

laasker:
Yea, the speed is the same : P

Not sure if it helps but, i found in spell.dbc that Feral Charge - Cat 49376 casts
Charge 61138 https://wotlk-twinhead.twinstar.cz/?spell=61138
and that charge casts another charge https://wotlk-twinhead.twinstar.cz/?spell=61132

I tried to change the id to both of these but the speed is the same : (

Helias:
Understood, the problem is the limit in MoveSplineInit.cpp line 123, should we overcome this limit? In TC is the same.
Are we sure that the youtube video with the increased speed is correct?

laasker:
> Understood, the problem is the limit in MoveSplineInit.cpp line 123, should we overcome this limit? In TC is the same.
> Are we sure that the youtube video with the increased speed is correct?

Yes, Zybak youtube channel videos are all from Retail, so it is how the spell correctly works

Mitradis:
why you replace "else if" on if?

laasker:
i decided to test this again and worked!
im really, really sorry. I dont know what i had done wrong, but this does increases the speed of Feral Charge - Cat.

Helias:
@laasker so the PR fix the issue? it's hard to test this, don't worry

@Mitradis because inside the if there is a return, it is unrelevant if you use else if or if in this case

laasker:
@Helias Yes the PR fix the issue, thank you

vhiperdev (approved):

FrancescoBorzi (approved):

PR #4826, opened 2021-03-13
fix(DB/creature_template): Update BaseAttackTime for various creatures

Zoidwaffle:
## Changes Proposed:
-  Modify BaseAttackTime for multiple creaturese
-  Suggested by @T1ti, I only helped out with some DB/comparison stuff and made the PR

## Issues Addressed:
- Game accuracy :grin: 

## Tests Performed:
- Applied SQL

## How to Test the Changes:

- Get a beating from anything in the list

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Zoidwaffle:
@T1ti : please check if this is what you had in mind, thanks!

pklloveyou:
Increasing attack time will increase single damage

pklloveyou:
example
```sql

UPDATE creature_template SET RangeAttackTime = 2000, BaseAttackTime = 2000 WHERE entry = 15550;
UPDATE creature_template SET RangeAttackTime = 2000, BaseAttackTime = 2000 WHERE entry = 16152;

```

Zoidwaffle:
To make it visible: I changed the respawn from 10m to 5m. This is because the default decay is 5m for elites, and in total it makes it a 10m respawn rate (which is what the goal was).

I tested the above and it seems correct on my own server - please correct me if I'm wrong.

T1ti:
> Increasing attack time will increase single damage

with the CLS system that's normal, the CLS provides a DPS, not a flat dmg value, modifying the attack time will adjust the damages so the creature's dps always stay the same

pklloveyou:
> > Increasing attack time will increase single damage
> 
> with the CLS system that's normal, the CLS provides a DPS, not a flat dmg value, modifying the attack time will adjust the damages so the creature's dps always stay the same

yes That's what I'm talking about, but if the attack interval is too long, it will cause a single big damage

T1ti:
> > > Increasing attack time will increase single damage
> > 
> > 
> > with the CLS system that's normal, the CLS provides a DPS, not a flat dmg value, modifying the attack time will adjust the damages so the creature's dps always stay the same
> 
> yes That's what I'm talking about, but if the attack interval is too long, it will cause a single big damage

that is blizzlike and intended

T1ti:
So about those values, I got them from the vmangos sniff database from Classic.
For patchwerk and Maexxna, I didn't want to include any of the Naxxramas changes as Wotlk probably updated them, but I verified both of those on retail and they still match so those are still right for 3.3.5. There are a bunch of other Naxxramas changes I didn't include that I verified on retail that don't seem to match 3.3.5

Si1ker (approved):

T1ti:
Mr. Smite attack time should be 2000, I made a mistake 3000 is his final form with the hammer

Zoidwaffle:
Check, he has been removed.

ChromieCraftServer:
has this been tested in game?

Eldarix:
Hello !! I did some testing after adding the mod, with some from the list this is what I find:

WoWScrnShot_042221_083035
WoWScrnShot_042221_083308
WoWScrnShot_042221_083645
WoWScrnShot_042221_084449
WoWScrnShot_042221_084605
WoWScrnShot_042221_084741
WoWScrnShot_042221_084114

necropola:
Does the list of creatuires include any low level creatures that have this "Thrash" ability? A trieel hit with increased weapon damage (due to slower speed) woudl be even more devastating.

T1ti:
Thrash is a passive aura, it's not related to attack speed

necropola:
I undesrstand that decreasing the Attack Speed autoamatically increases the damage of a single hit because DPS stays the same. Since "Thrash" grants 2 extra attacks, I was thinking that the damage of those attacks would then increase, too.

**Example :**
The creature **Cracked Skull Soldier (1523)** has a SmartAI ability **Thrash (589)** which grants 2 extra attacks which in itself is quite a powerful ability for a level 8-9 NPC, if I might say so. Now, if you were to decrease attack speed for this creature, woudn't it become even more bursty?

T1ti:
but it will proc less often as thrash is usualy a % chance, dps will be the same.

necropola:
Well, the example creature mentioned above has a 100% chance to cast "Thrash" on aggro via SmartAI. And I think, there are quite a few more creatures doing this.

PR #4884, opened 2021-03-17
fix(DB/gameobject): Align respawn time of fish schools / wreckage

Zoidwaffle:
## Changes Proposed:
- Change respawn time to 3600s if less than 3600s.
- Reference: https://legacy-wow.com/fishing-guide-pool-appearance/ (60 min might be a bit much, but sounds about right according to this). I don't think it should be less than 15m, but anywhere from 15-60 min would seem fine to me.

## Issues Addressed:
- Closes https://github.com/chromiecraft/chromiecraft/issues/214
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4890
- The issue is that the respawn timer can be found as low as 2 minutes. This makes it too trivial to fish up materials and valuables from the same spot continuously.

## Tests Performed:
- Applied SQL
- Not tested in game

## How to Test the Changes:

- Get your fishing hat!
- Visit some affected nodes - get the `guid` from the SQL changes and do `.go object <guid>`. Examples:
- `.go object 17289`
- `.go object 64783`
- Fish it up, wait for it to respawn while drinking something with alcohol in it

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Zoidwaffle:
I'm aware some of this can be solved by using pools - so if a zone has 10 fishing spots like this, then they should probably get pooled and a lower respawn rate could be fitting. The stuff here is indeed pooled, there's 624 entries `pool_gameobject` and 321 unique pool references. But none of those pools are then pooled again...

Did that make sense?

TL;DR: The respawn rate should be higher than 2 minutes. Maybe 60 is too much, but I don't suggest going lower than 15 minutes. The closest I could find as "evidence" said ~60-65 so I settled on 60.

FrancescoBorzi (approved):

Zoidwaffle:
I removed "School of Tastyfish" as a target of this change.

It will still target pools of type:
Floating Wreckage
Floating Debris
Oil Spill
Firefin Snapper School
Greater Sagefish School
Oily Blackmouth School
Sagefish School
School of Deviate Fish
Stonescale Eel Swarm
Highland Mixed School
Pure Water
Bluefish School
Brackish Mixed School
Mudfish School
School of Darter
Sporefish School
Steam Pump Flotsam

Delcidx:
Tested in game.

- The respawn time of the fish schools/wreckage is greater than 15 minutes.
Tested
tested2

PR #4887, opened 2021-03-17
fix(DB/pool_template): Treasures in Teldrassil

Zoidwaffle:
## Changes Proposed:
- Pool treasures in Teldrassil
- 10 spawn locations, 4 in the pool
- Lower respawn to 5 minutes

## Issues Addressed:
- This should make it more enjoyable to visit the starting zone for nelves if there are multiple players at once.

## Tests Performed:
- Applied SQL
- Visited all locations, twice
teldrassil_treasure1

teldrassil_treasure2

## How to Test the Changes:

- `.go object 49528,49529,49621,49622,49623,49624,49625,49626,49627,49628` (only select 1 guid at a time)

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1615990232375736422.sql
+++ b/data/sql/updates/pending_db_world/rev_1615990232375736422.sql
@@ -1,0 +1,28 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1615990232375736422');
+
+SET
+@POOL            = '11651',
+@POOLSIZE        = '4',
+@POOLDESC        = 'Treasures - Teldrassil',
+@GUID = '49528,49529,49621,49622,49623,49624,49625,49626,49627,49628';
+
+-- Create pool(s)
+DELETE FROM `pool_template` WHERE `entry`=@POOL;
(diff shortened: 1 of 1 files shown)


FrancescoBorzi (approved):

Hacki95:
While this makes it more enjoyable, is there any indication if this is blizzlike at all or are you just changing the values without knowing?

Zoidwaffle:
Well...

https://www.warcrafttavern.com/wow-classic/guides/chest-locations/ this indicates there should be way more chests than now.

Notice a lot of them are clustered - and it would seem kinda fair that they should be pooled, so you don't enter a single mine and pick up 4 chests if you are the only player online.

I only checked out the area to see if there where any chests at all - if it had only... say 5... then I would have cross-referenced Gathermate to see where to spawn some more. As mentioned already, there should probably be more than the 10 current spawn locations.

Adding a pool for the 10 chests now is a minor improvement, before adding the missing chests and perhaps making even more pools (dividing the zone up a bit, thus making it "true" Blizzlike).

So no, I have no evidence that this is Blizzlike - but I believe it's more Blizzlike than it was and that pools are definitely needed.

If I'm wrong, I'm happy to change the PR to make AC better (after all, that's the goal here).

Issue #4907, opened 2021-03-19
Spell Daze 1604 is not removed by Druid transformations


Stifler82:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/223

##### CURRENT BEHAVIOUR:

When being dazed, a way to counter this is to exit your form, heal, and re-enter. According to the site provided by the format, as well as the ingame tooltip: "The act of shapeshifting frees the caster of Polymorph and Movement Impairing effects." When I did this to remove daze, it was not removed and was simply dazed again attempting it.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

The act of entering the form, such as travel form, should remove daze or any other imparing effects.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Fight a mob and get dazed while in a form, ex: travel form
2. Exit Travel Form
3. Enter Travel Form
4. Notice you are still dazed

Thank you for your report. Please click submit new issue below.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/a32310772487a7a547b1a239dfaf0967b19f7493

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

PR #4908, opened 2021-03-19
fix(DB/Spell_dbc): Spell Daze 1604 is not removed by Druid transformations.

Stifler82:
## Changes Proposed:
-  The spell **Daze 1604** has been modified and added to the **"spell_dbc"** table.

## Issues Addressed:
- Closes https://github.com/chromiecraft/chromiecraft/issues/223
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4907

## Tests Performed:
- Tested in game and working
- Tested on Windows 10 (64)
- **Video Test:** [Druid Shapeshift](https://youtu.be/S6cMSDpnxKk)

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

necropola:
I strongly disagree that this should be *fixed*.
See: https://github.com/chromiecraft/chromiecraft/issues/223#issuecomment-802645816

Stifler82:
> I strongly disagree that this should be _fixed_.
> See: [chromiecraft/chromiecraft#223 (comment)](https://github.com/chromiecraft/chromiecraft/issues/223#issuecomment-802606614) and [chromiecraft/chromiecraft#223 (comment)](https://github.com/chromiecraft/chromiecraft/issues/223#issuecomment-802645816)

@necropola
**Spell:** [wow.gamepedia.com-Link - Daze](https://wow.gamepedia.com/Daze)
```
****Creature daze****
Creature daze is a 50% snare debuff, which often forces its target to dismount if mounted. 
All mobs have a chance of applying daze when they hit a player in melee range that is facing away from the mob. 
The daze mechanic is intended to prevent players from carelessly running past large groups of mobs.

All mob creatures have a chance to daze on each melee hit, ranging from 0% to a maximum 40%. 
Chance to daze is primarily a function of player level - mobs that are below a player's level have less chance to daze, 
while mobs above his level have a higher chance, with equal level mobs having a base 20% chance.

The base chance of 20% is decreased for players under level 30. For example, 
a level 10 player running from a level 10 mob will only have a 7% chance to be dazed.

A player can only be dazed from behind and within melee range of the attacking mob. 
This is most common when running away from it, 
but it can also happen while standing still if the mob approaches from behind, 
or when backing up if the mob overshoots you while following you. 
Due to server latency, it can sometimes appear that a mob has dazed from a long distance, or in front of the player.

It has been confirmed that friendly mobs (for escort quests for example) can daze other mobs that are running away. 
It also seems that the dazed debuff can't be applied by a mob if you are "Immune" 
(such as with a Paladin  [Divine Shield] or Mage  [Ice Block]),
if you "Resist" the hit, or if all the damage from the hit is Absorbed (such as with  [Power Word: Shield]).

While it slows movement speed, creature daze is not considered a movement impairing effect. 
Thus it cannot be removed by the usual means such as druid shapeshifting, 
[Vanish], Faction Insignia, or  [Hand of Freedom]. It is considered a physical debuff, however, 
and can be removed by  [Hand of Protection] and the like. 
Note that it is also removed by immunity granting effects such as  [Divine Shield] or  [Ice Block].

=======================================================================================================================

****Player daze****
Some playable classes have abilities that apply a dazed debuff as well. 
However, player dazes do not dismount others the same way creature daze does, 
nor is the player daze effect immune to skills/items that remove movement impairements or harmful spell effects.

Dazing effects are notorious in PvP settings, especially in Warsong Gulch. 
Flag runners are usually hit with these abilities, 
leaving them vulnerable to attack from the opposing team. 
Player daze effects usually reduce movement speed by either 50% or 70%, depending on the ability.
```


I agree with Necropola, I made a mistake.
I made this mistake because I saw that: **here TOO often**, it happens that **people start commenting on reports and PRs too hastily with **C++ CODES****, without making a thorough investigation as to whether a C++ CODE needs to be changed or not.

So, concerned that someone **(in general, nobodies specific)** could make damage using **CODE C++ on this Spell**, I did a 90% investigation and trusted both the report and the initial comments from the report itself.
So I didn't investigate if this spell was a spell that is (excluded) by this rule **"The act of shapeshifting frees the caster of Polymorph and Movement Impairing effects."**.

So, my bad..
I sinned of "Overprotection" towards the Spells..

PR #4929, opened 2021-03-21
refactor(DB/SQL): Update integer & utf to MySQL's new standards

Kitzunu:
## Changes Proposed:
MySQL gonna do some changes in future updates and throws these warnings when you make tables

- Remove specification of integer
- change utf8 & utf8mb3 to utf8mb4  
- float unsigned to CHECK (`col`>=0) (https://github.com/azerothcore/azerothcore-wotlk/issues/4930)
image

- Change varchar(500) to TEXT(500) in map_dbc due to 
```
Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
```

- Change version.sql engine from MyISAM to INNODB due to
```
ERROR 1071 (42000) at line 2776530: Specified key was too long; max key length is 1000 bytes
```
## Tests Performed:
- None

## How to Test the Changes:

re-make DB and see that everything works as expected

Guessing you haven't done witchcraft with your DB, here you can check all CHECK constraints
```sql
SELECT * FROM information_schema.CHECK_CONSTRAINTS;
```

**Note to ppl with HeidiSQL:** Check constraints works, but viewing them will be fixed with version 11.3 when it releases.

## Need to be tested with

- [x] MariaDB
- [x] MySQL 8.0
- [x] MySQL 5.7

## Target Branch(es):
- [x] Master

 
Note to self RegEx \(.+?\)


Kitzunu:
All CI pass tho :D

pklloveyou:
Compatible with MariaDB?

Kitzunu:
Thats for whoever uses mariaDB to check.

But it should be no difference

claudiodfc:
Doesn't shin have to update the Ci?

FrancescoBorzi:
we should test this on different versions of MySQL (5.7 and 8.0) and MariaDB (recent versions)

Kitzunu:
utf8mb4 supported by mariadb
https://mariadb.com/kb/en/supported-character-sets-and-collations/
and mysql 5.7
https://dev.mysql.com/doc/refman/5.7/en/charset-unicode-utf8mb4.html

Check constraints exist for everyone.

so in theory everything should be good

FrancescoBorzi:
> utf8mb4 supported by mariadb
> https://mariadb.com/kb/en/supported-character-sets-and-collations/
> and mysql 5.7
> https://dev.mysql.com/doc/refman/5.7/en/charset-unicode-utf8mb4.html
> 
> Check constraints exist for everyone.
> 
> so in theory everything should be good

ok good to know, also MySQL 5.7 and 8.0 are already checked in the CI.

But now I'm thinking about something else: @Kitzunu shouldn't we also add an update file for each DB so it updates the table structures for **existing** servers? otherwise your change only affects new servers.

Kitzunu:
Would be optional, but that would mean we have to manually update every single table and that is A LOT of work

Kitzunu:
To be exact, manually updating 377 tables

Kitzunu:
Nah I can't sit and do that for 377 tables...

Winfidonarleyan:
Ha, for other tables create separate PR after merge this :D

Kitzunu:
I was thinking of doing it, but when I came to the characters database and saw 47 tables I gave up q.q

Winfidonarleyan:
- MariaDB 10.5

image

FrancescoBorzi (approved):

FrancescoBorzi:
thanks @Winfidonarleyan for your feedback!

PR #4977, opened 2021-03-24
fix(DB/mechanic_immune_mask): Bosses from Deadmines 

Zoidwaffle:
## Changes Proposed:
-  The bosses are immune to Thunderclap and they shouldn't be

## Issues Addressed:
- Wrong immunity mask - this was last changed in https://github.com/azerothcore/azerothcore-wotlk/pull/4630/files where immunities where carried over from vmangos. However, the MECHANIC_SLOW_ATTACK mask was kept from AC. Now it will be disabled for these specific bosses.
- I have been unable to find other evidence for these changes - it was mentioned here https://github.com/azerothcore/azerothcore-wotlk/pull/4630 and that's the reason for this PR

## Tests Performed:
- Applied SQL
- Not tested in game

## How to Test the Changes:

- Get someone with Thunderclap
- `.go c id 639` (or one of the other guid's).
- Fight and remember to use Thunderclap

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Zoidwaffle:
@JamesGent and @T1ti : these are the changes you mentioned, right?

T1ti:
A bulk update of that flag would probably be a better idea

Si1ker (changes requested):
Do not override flags, add them instead.
See my PR for reference: https://github.com/azerothcore/azerothcore-wotlk/pull/3513

Kitzunu:
Except here he is removing, so should be `&~` :P

Si1ker:
> Except here he is removing, so should be `&~` :P

@zoidwaffle there is the expert! so use the PR linked for reference and `&~` to remove.

Zoidwaffle:
Dammit @Si1ker and @Kitzunu , I learned something. Crap, now I have to forget the last shreds of 2nd grade to make room...

Zoidwaffle:
@T1ti : For some reason pretty much everything from Deadmines had this flag... and then also the racial leaders.

So maybe a bulk update isn't the best way - I guess the racial leaders could/should be immune to a lot?

Besides that, only Somnus (12900)... So if only targeting mobs from #4630, I think it's OK now.

T1ti:
If it's only specific to a few creatures then sounds good. I don't think racial leaders should be immune to it tho ?

A bit unrelated, but now deadmine bosses are immune to stun etc like high level bosses, in low level dungeons bosses should be stunnable/slowable etc

But I don't know at which point bosses become immune to those effects

Si1ker (approved):
/* Affected rows: 8  Found rows: 0  Warnings: 0  Duration for 2 queries: 0.016 sec. */

Issue #4860, opened 2021-03-15
"Ending Their World" quest bugged - despawning NPC, incorrect aggro, incorrect dialogues


Stifler82:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/205

##### ISSUE: FACTION SIDE:

![Alliance](https://user-images.githubusercontent.com/1884642/108204869-3a88d100-711c-11eb-8179-e1b9b73ed450.png)

##### CONTENT PHASE:

Phase 1-19, Bloodmyst Isle

##### SMALL DESCRIPTION:

Quest: [Ending Their World](https://wotlkdb.com/?quest=9759)
NPC_Start: [Demolitionist Legoso](https://wotlkdb.com/?npc=17982)
NPC_End: [Exarch Admetius](https://wotlkdb.com/?npc=17658)

Zone: [Bloodmyst Isle](https://wotlkdb.com/?zone=3525)

There are three bugs I've noted while doing this quest:
1) ~The escort NPC Demolitionist Legoso is DESPAWNING if he gets too far from you, approximately around 40-50 yards away from the player. This is game breaking, as you can't finish the quest without the NPC and you have to abandon the quest, wait \~5-10 minutes for the NPC to respawn and retake the quest. This also makes the quest very difficult to finish, as you cannot stop to heal/cast a spell because of fear of Legoso getting too far away and disappearing.~ https://github.com/azerothcore/azerothcore-wotlk/commit/5a770d294500f74d4eb0d46224e3b314cdeb7fb2
2) Sironas is getting aggro'd before the explosion and somehow is able to attack the NPC Legoso from far distance, even though being a melee mob. This also makes quest additionally difficult to finish, because during my attempt there was four adjacent Sunhawk mobs suddenly getting aggrom which isn't blizzlike behavior.
3) Legoso's dialogue during the quest is all over the place, being out of sync with what is actually happening, i.e. bomb explodes before he finishes counting to 3, encouraging the player for the fight after you already killed the boss and so on and so forth. Below you can find examples attached: first picture is Legoso asking for more time to get the bomb ready, when in reality he's already set it up like 25 seconds ago. Second picture is mentioned previously Legoso saying to get ready when the quest has already been finished, as well as forgetting player's name.
![Screen Shot 03-14-21 at 03 11 PM](https://user-images.githubusercontent.com/80629696/111079769-ba2b7500-84fb-11eb-80b0-5b8f5706abd1.JPG)
[...]

PR #4980, opened 2021-03-24
fix(Core/AI): improve npc_escortAI

Kitzunu:
cp from tc

## Changes Proposed:
- NPC should not go into combat if it has REACT_PASSIVE
- NPC should not go into combat if the place is inaccessible
- NPC should not go into combat if AI cannot attack
- NPC should not go into combat if it is in EVADE mode
- NPC should not go into combat if the enemy is in EVADE mode
- Update DEFAULT_MAX_PLAYER_DISTANCE to 100 instead of 50 to allow players to be further away from the NPC (ref https://github.com/azerothcore/azerothcore-wotlk/issues/4860)

## Tests Performed:
- Build

## How to Test the Changes:

Edit by Shin: check a couple of creatures that use npc_escortAI

For example:

1. Get a creature that uses npc_escortAI
2. check the conditions above, creature should not aggro when they are true
3. Check that the NPC only despawns if you are >100 yards away

## Target Branch(es):
- [x] Master

 

55Honey (commented):
Looks good to me.

FrancescoBorzi (approved):

FrancescoBorzi:
@Kitzunu have you tested this in game? I see only "build" in "Test performed". A check in game is required for such generic C++ change.

Kitzunu:
I have not as I am limited on time, but anyone can easily test it.

But the conditions are pretty logical, for example it should obviously not attack if it is passive

T1ti:
Kinda unrelated but it's been proved in sniffs that on Classic, escort NPCs change their faction to "Escortee" (faction id 40) I wonder if using the correct faction would fix some issues with escorts

Kitzunu:
Testing this would probably require the tester to change code to some script or NPC to make these conditions true.

This is why I don't think it is necessary and instead just look at the logic.
Should an escort NPC be able to attack while in passive state? No. if (me->HasReactState(REACT_PASSIVE))
Should an escort NPC be able to attack a target it cannot get to? No. if (!who->isInAccessiblePlaceFor(me))
Should an escort NPC be able to attack if it cannot attack "who"? No. if (!CanAIAttack(who))
Should an escort NPC be able to attack if it is in evade mode? No. if (me->IsInEvadeMode())
Should an escort NPC be able to attack an enemy that is in evade mode? No. if (who->GetTypeId() == TYPEID_UNIT && who->ToCreature()->IsInEvadeMode())

FrancescoBorzi:
> Testing this would probably require the tester to change code to some script or NPC to make these conditions true.

not necessarily. Once again: when testing, we should first make sure that there is no regression. So testing a couple of escort quests is what we want.

pklloveyou:
Tested

Issue #4868, opened 2021-03-16
[HORDE QUEST] Unexpected Results - Silvermoon, mobs flying in


Stifler82:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/213

##### ISSUE: FACTION SIDE:

 ![Horde](https://user-images.githubusercontent.com/1884642/108204991-63a96180-711c-11eb-9ed9-138c233070a6.png)

##### CONTENT PHASE:

 1-19

##### SMALL DESCRIPTION:

The event mobs fly towards the quest giver NPC at high speed

Quest: [Unexpected Results](https://www.wowhead.com/quest=8488/unexpected-results)

 NPC: [Apprentice Mirveda](https://www.wowhead.com/npc=15402/apprentice-mirveda)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

The mobs should runs towards the npc from a distance, not instantly fly and attack her.

##### CURRENT BEHAVIOUR:

Mobs instantly fly and attack the npc

##### STEPS TO REPRODUCE THE PROBLEM:

Step 1 Go to Eversong woods and complete the 1st quest given by Apprentice Mirveda ( Corrupted soil quest)
Step 2 After completing the previous one, accept the quest "unexpected results" and start the event 
Step 3 See the mobs fly to npc ( has happened on two characters so far)

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/a32310772487a7a547b1a239dfaf0967b19f7493

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

PR #5072, opened 2021-03-31
fix(DB/CreaturesSAI): Quest Unexpected Results

Stifler82:
## Changes Proposed:
- Fixed Stats NPC: [Apprentice Mirveda](https://wowgaming.altervista.org/aowow/?npc=15402)
- Fixed SAI NPC: [Apprentice Mirveda](https://wowgaming.altervista.org/aowow/?npc=15402)
- Fixed Stats NPC: [Gharsul the Remorseless](https://wowgaming.altervista.org/aowow/?npc=15958)
- Fixed SAI NPC: [Gharsul the Remorseless](https://wowgaming.altervista.org/aowow/?npc=15958)
- Fixed Stats NPC: [Angershade](https://wowgaming.altervista.org/aowow/?npc=15656)
- Fixed SAI NPC: [Angershade](https://wowgaming.altervista.org/aowow/?npc=15656)

## Issues Addressed:
- Closes https://github.com/chromiecraft/chromiecraft/issues/213
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4868

## Tests Performed:

- Tested in game and working
- Tested on Windows 10 (64)
- **Video Test:** [Unexpected Results](https://youtu.be/yh8Xm0hK9fw)

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/data/sql/updates/pending_db_world/rev_1617176680426521000.sql
+++ b/data/sql/updates/pending_db_world/rev_1617176680426521000.sql
@@ -1,8 +1,6 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1617176680426521000');
 
-UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15958);
-
-UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15656);
+UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE `entry` IN (15958, 15656);
 
 DELETE FROM `creature_summon_groups` WHERE (`summonerId` = 15402);
 
--- a/data/sql/updates/pending_db_world/rev_1617176680426521000.sql
+++ b/data/sql/updates/pending_db_world/rev_1617176680426521000.sql
@@ -1,0 +1,32 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1617176680426521000');
+
+UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15958);
+
+UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15656);
+
+DELETE FROM `creature_summon_groups` WHERE (`summonerId` = 15402);
+
+DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 15958);
+INSERT INTO `smart_scripts` VALUES
+(15958, 0, 4, 0, 75, 0, 100, 0, 0, 15402, 53, 900, 0, 49, 0, 0, 0, 0, 0, 0, 19, 15402, 53, 0, 0, 0, 0, 0, 0, 'Gharsul the Remorseless - On Distance To Creature - Start Attacking');
+
+DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 15656);
+INSERT INTO `smart_scripts` VALUES
+(15656, 0, 0, 1, 2, 0, 100, 1, 0, 30, 0, 0, 0, 11, 8599, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Angershade - Between 0-30% Health - Cast \'Enrage\' (No Repeat)'),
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/rev_1617176680426521000.sql
```diff
@@ -0,0 +1,32 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1617176680426521000');
+
+UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15958);
+
+UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE (`entry` = 15656);
```
FrancescoBorzi, line 5:
```sql
UPDATE `creature_template` SET `speed_walk` = 0.9, `speed_run` = 1.1 WHERE `entry` IN (15958, 15656);
```

Helias (approved):

Azcobu:
This fix was merged a month ago and would have reached CC a week or so later, correct? Because https://github.com/chromiecraft/chromiecraft/issues/504 mentions very similar behaviour continuing, and just verified it myself, with a single Angershade (and a nearby ghoul) attacking, with no boss spawned:

WoWScrnShot_050221_151459

So wanted to make sure this fix had made it to CC before pursuing the alternative (the fix didn't work) - especially as the test video above shows it working fine.

Stifler82:
@Azcobu 

**So, in a while I will be updating my CORE locally as I am a little bit of "commits" back.**

image

**But!
To my current CORE version which is:**
`AzerothCore rev. f40ef6de4d00+ 2021-04-21 01:01:06 +0000 (master branch) (Win64, Release)`

**The quest works like this:**
**Note:**
`I only made a small change to my fix (currently) that is: I only changed the coordinates of the mobs spawns because I realized that in this fix that I had done the mobs spawned in the same spot, but the fix is the same, only the coordinates change`

https://user-images.githubusercontent.com/1884642/116810563-62dd7480-ab3c-11eb-9c8c-7bd474548f23.mp4

Unfortunately this is the demonstration that: **"There are too many hands inside this jar of jam"**
(And forgive the irony) this situation creates **"Unexpected Results"**

**anyway, now I'll do an update of my CORE and then I'll check what has changed and why the Fix doesn't work anymore..**

Stifler82:
**I did the CORE update, I also updated the DB.**
image

**This is my current AC version:**
`AzerothCore rev. aee1351f1282+ 2021-05-02 00:53:23 +0000 (master branch) (Win64, Release)`

**I have not been able to reproduce the Bug, the only difference between CC and my local is the mobs spawns coordinates.
And anyway I only changed the Coordinates of the minions not the "Boss".
The Boss coordinates are the same as CC and AC**

https://user-images.githubusercontent.com/1884642/116811957-09794380-ab44-11eb-82a5-e5f81b7c881e.mp4

Stifler82:
The only thing I can do **"How to improve coordinates"** and make a Re-work of the previous fix, but nothing changes, the fix is the same as EXACT.
**I repeat:** only the coordinates of the minions change, **(i.e. the 2 identical Angershades that appear on the sides of Gharsul the Remorseless)**

Azcobu:
Just updated my test server as well (to rev. aee1351f1282 2021-05-02 00:53:23, same as yours) and ran this with a level 10 pally belf, and still only got the one Angershade spawning and no Gharsul the Remorseless. I didn't make a nice video like you, just a screenshot:

WoWScrnShot_050221_212913

I wonder if the difference is in how we're accepting the quests or something? Instead of '.quest add 8488', I did it the long way - picked up 'The Wayward Apprentice' from Magistrix Landra Dawnstrider in the village to the west, then travelled to Mirveda, accepted 'Tainted Soil', spawned 8 soil samples, turned them in, and started 'Unexpected Results' from there. i don't know if that could or would make a difference, but can't think of anything else.

Stifler82:
@Azcobu 
Then..
I really don't think the class or race makes a difference, and neither does the chain quest..

Especially because I did the SAI of these MOBS and I know for sure that it has nothing to do with it..

it is Apprentice Mirveda who starts the quest event.
and her SAI says: (Translated): **"As soon as the player gets the quest** `"ID"` **summon the creatures** `"IDs"` **in** `XYZO` **coordinates."**

So, everything that you have hypothesized (for this reason that I have just explained to you), is impossible.

Because the "Trigger" of the event is the player who takes the quest, therefore since at least 1 mob of the trigger spawns, it means that the trigger works, and for some reason the other mob does not.

so now I'm making the RE-WORK, then, we wait for the merge, and after the merge we re-test the quest, and see if it worked..

Stifler82:
@Azcobu 
And btw


https://user-images.githubusercontent.com/1884642/116814407-bf4a8f00-ab50-11eb-92fd-738d4d7ec642.mp4

Issue #524, opened 2017-07-13
Core/Creature: Guards do not attack a non-moving target [$10]


BarbzYHOOL:
**Description**: Neutral guards don't protect players attacked by a player that is not moving.

**Current behaviour**: NEUTRAL guards in booty bay, light's chapel hope etc, won't do anything if you attack another player while not moving yourself. If the guards are near you and standing still, htey won't do anything. But if they move, they will notice it. If you move, they will also notice it.
It has something to do with the guard AI not updating when not moving, and when the player is not moving, something is missing in the AI. (If a guard is moving, he will notice the player though)

It means you can abuse this very easily with ranged attacks, with your pet, or by simply attacking someone in melee without moving.

**Expected behaviour**: The guards should immediately attack you, no matter if they are moving, or standing.

**Steps to reproduce the problem**:

1. Take two characters (alliance and horde)
2. .tele bootybay
3. Go in a place with standing guards, not patrolling.
4. Attack one character without moving.
5. Nothing happens until you move, or until a patrolling guard comes near.

Tested these with @talamortis on my server and on his server.

**Branch(es)**: master (1.0.x+)

[//]: # (This template is for problem reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)
[//]: # (For fixes containing c++ create a Pull Request)

ShinDarth:
2 people tested it (barbz and talamortis) so we can consider this confirmed

BarbzYHOOL:
Same thing on TC demo server (without mmaps enabled but I bet it's the same behaviour).

poszer:
Interesting

poszer:
Looks like issue with movement. Same with other creatures, if you stand near them without moving they wont attack you, ofc until you move.

BarbzYHOOL:
i had said it already

BarbzYHOOL:
It's not the guardAI, it's just about every NPC in the game

If you don't move, it doesn't update. The NPC does not know you exist. You can start casting a long spell, you can buff up, you can eat, you can drink. You can even unstealth right in front of a boss and he won't move. (unless scripted event)

PR #5076, opened 2021-03-31
fix(Core/Unit): Set AI notification when setting/removing contested PvP flags.

Chero:
## Changes Proposed:
-  Set AI notification when player's contested PvP flag changes, so nearby stational guards will attack correctly.

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/524

## Tests Performed:

- Build & Implementation

## How to Test the Changes:
1) Login 1 Horde and 1 Alliance character

2) Go to Booty Bay or another neutral city. Find a stationary guard
**OR**
.tele barrens
.npc add 4624

3) Attack the other player without moving

## Target Branch(es):
- [x] Master

 

Hacki95:
Does this also fix the issue which is named in the first comment here: https://github.com/chromiecraft/chromiecraft/issues/311 ?

Chero:
> Does this also fix the issue which is named in the first comment here: [chromiecraft/chromiecraft#311](https://github.com/chromiecraft/chromiecraft/issues/311) ?

I'm afraid not, but I can imagine a similar fix, calling this on creature spawn. Just-spawned creatures, however, have supposed to have a slight delay between spawn and reaction time (about 5 sec or so)

Hacki95:
> 
> 
> > Does this also fix the issue which is named in the first comment here: [chromiecraft/chromiecraft#311](https://github.com/chromiecraft/chromiecraft/issues/311) ?
> 
> No, if there is such an issue, this will not fix that. This fix is only regarding the contested PvP flag toggle.
> 
>     1. I can imagine a similar fix, calling this on creature spawn. Just-spawned creatures, however, have supposed to have a slight delay between spawn and reaction time (about 5 sec or so).
> 
>     2. As for player resurrection,  UpdateObjectVisibility() is already called there, so in this case, there might be a limitation regarding positioning or something like that. In such cases, a positional condition in visibility updates should be disregarded
> 
> 
> There are various places, where AI & visibility notifications should be triggered (i.e. AI change, resurrection, teleport, movement, etc..) including the contested PvP flag toggle which was completely missing. And there is no oneliner fix for all of them unless you want your CPU usage to skyrocket.

Thanks for the information! Do you think you might be able to make a PR for this bug too?

pklloveyou:
How to test

pklloveyou:
Tested

FrancescoBorzi (approved):

Issue #4996, opened 2021-03-25
[Mining Node] HORDE - 2 nodes on the exact spot


Stifler82:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/290

##### FACTION

- [ ] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39

##### CURRENT BEHAVIOUR:

Two nodes on the exact spot outside orgrimmar, five such instances have been noted. You can see the coordinates 
on the screenshots which show 2 objects on the same coordinate
![Screenshot (9)](https://user-images.githubusercontent.com/80331807/112448743-bbb93080-8d78-11eb-968e-7f2039fbd913.png)
![Screenshot (10)](https://user-images.githubusercontent.com/80331807/112449001-f8852780-8d78-11eb-9699-2ea4fd65b39f.png)
![Screenshot (11)](https://user-images.githubusercontent.com/80331807/112449246-31bd9780-8d79-11eb-8580-88bf782a26a4.png)
![Screenshot (12)](https://user-images.githubusercontent.com/80331807/112449491-72b5ac00-8d79-11eb-94d8-61c24b49a318.png)
![Screenshot (13)](https://user-images.githubusercontent.com/80331807/112449658-a55fa480-8d79-11eb-9454-c62056c6748b.png)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

The nodes should not spawn on top of each other.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Go to the coordinates
2. Mine the node
3. See another node on the same spot after you finish mining it

##### EXTRA NOTES:

Coordinates : 

1) X:503.216003 Y:-3923.429932 Z:23.015200 MapId: 1
2) X:1067.800049 Y:-4590.990234 Z:27.041000 MapId: 1
3) X:1316.969971 Y:-4828.229980 Z:24.025499 MapId: 1
4) X:1178.939941 Y:-4858.160156 Z:24.753000 MapId: 1
5) X:1417.420044 Y:-4711.250000 Z:-0.304713 MapId: 1

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/8bacb39d53349f63066e84bcddb5a7f9300acb0d

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

[...]

Issue #4770, opened 2021-03-09
2 mining nodes inside/on top of eachother in Plaguelands


ghost:
2 mining nodes inside/on top of eachother in Plaguelands

2 locations.

##### CURRENT BEHAVIOUR:

![WoWScrnShot_030921_113450](https://user-images.githubusercontent.com/61194190/110464810-84d1f200-80cb-11eb-9690-68c42ae4d120.jpg)

![WoWScrnShot_030921_113420](https://user-images.githubusercontent.com/61194190/110464816-87cce280-80cb-11eb-8c3e-bd508e4f402f.jpg)

Found another here.

![WoWScrnShot_030921_113842](https://user-images.githubusercontent.com/61194190/110465191-03c72a80-80cc-11eb-84cb-ff4a5c5f34a1.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Only 1 at a time

##### STEPS TO REPRODUCE THE PROBLEM:

1. Go to the shown location in screenshot/map
2. Use Find Minerals
3. Show 2x nodes in same spot

##### AC HASH/COMMIT:

AzerothCore rev. 6a7155eb7e82 2021-03-07 12:37:07 +0000 (master branch)

celestinmetral:
Hello, if possible, I would like to try to resolve this issue for my first one. Please don't hesitate to tell me if I'm doing wrong!

I can confirm that there are two nodes at X=1238 and Y=-1780 (see picture). I found the line in `gameobject` table, and to try I moved the Mithril one at Z=73 for visibility (and to confirm I've got the correct one!).

![image](https://user-images.githubusercontent.com/13820293/110552109-027f1780-8137-11eb-8fb3-2ae57f20648e.png)

To check if we should keep the gold or mithril one, I looked at WotlkDB : [gold vein](https://wotlkdb.com/?object=1734) and [mithril deposit](https://wotlkdb.com/?object=2040) (see "Western plaguelands" map). The problem is that for both objects, there is a spawn node at (51.1, 74.3). Which one should we keep? Where can I can find better info ? Youtube videos ?

Let say we decide to remove the mithril one, the query would be something like `DELETE FROM gameobject WHERE guid=7193;`, right ?

From know, I've got to learn how to make a correct pull request, and understand how the `INSERT INTO `version_db_world` (`sql_rev`) VALUES ('xx');` I saw in other PR works.

Thanks !

Zoidwaffle:
[...]

PR #5078, opened 2021-03-31
fix(DB/gameobject): Remove duplicate mineral nodes

Zoidwaffle:
## Changes Proposed:
-  No minerals of the same type should ever spawn inside each other

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4770
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4996
- Closes https://github.com/chromiecraft/chromiecraft/issues/290
- Partial fix for https://github.com/azerothcore/azerothcore-wotlk/issues/2510

## SOURCE:

- There is no source - this was done using a scripted comparison of DB values for `gameobject` guid's. If an object is present at a location, no other similar object should be in the exact same spot. The object with the lowest guid remains, the higher one is deleted.
- Objects are also removed from `pool_gameobject`.
- I made the SQL take up a lot of lines, to make it easy to understand what is deleted and what remains

## Tests Performed:

- Applied SQL
- Tested some of the objects mentioned in https://github.com/azerothcore/azerothcore-wotlk/issues/4996, it seems fine now
- Not tested for all objects - as some of them are pooled, the issue(s) will not be visible at all times

## How to Test the Changes:

- `.go object <guid>` (check the changes done).

- I'm not aware of any potential issues - anyone who can think of something?

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

pklloveyou:
good job

PR #5023, opened 2021-03-27
fix(DB) Wrong position or multiple placements of herbs

nolt:
## Changes Proposed:
-  Fix wrong position and placement or multiple spawn of herbs

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4984

## Tests Performed:

- SQL applied
- Tested in game

## How to Test the Changes:

- Go to locations reported in issues, check fixed spawn points. (.go object GUID and .gobject near)

## Known Issues and TODO List:

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #5079, opened 2021-03-31
fix(DB/gameobject): Remove duplicate herbs

Zoidwaffle:
## Changes Proposed:
-  No herbs of the same type should ever spawn inside each other

## Issues Addressed:
- No open issues without a PR as far as I can see, but related issues pop up fairly often. Examples include:
- chromiecraft/chromiecraft#272
- chromiecraft/chromiecraft#257
- chromiecraft/chromiecraft#249
- ... and I'm pretty sure more will be found as new zones open up

## SOURCE:

- There is no source - this was done using a scripted comparison of DB values for `gameobject` guid's. If an object is present at a location, no other similar object should be in the exact same spot. The object with the lowest guid remains, the higher one is deleted.
- Objects are also removed from `pool_gameobject`.
- I made the SQL take up a lot of lines, to make it easy to understand what is deleted and what remains

## Tests Performed:

- Applied SQL
- Visited a few locations
- Not tested for all objects - as some of them are pooled, the issue(s) will not be visible at all times

## How to Test the Changes:

- `.go object <guid>` (check the changes done).

- I checked up against  https://github.com/azerothcore/azerothcore-wotlk/pull/5023, removed 1 herb already handled there, so this will not be a complete cleanup before that PR is merged as well
- I'm not aware of any potential issues - anyone who can think of something?

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

pklloveyou:
good job

PR #5103, opened 2021-04-02
fix(Core/Movement): Make sure pet always follows its master.

### LINKED ISSUE #4603: Bug [core] Pet stuck in warsong gulch [$35]
##### ISSUE: FACTION SIDE: 

image
image
**Hunters and Warlocks**
##### CONTENT PHASE:
 Every Content Phases, not linked to a specific level

##### SMALL DESCRIPTION:
If you join WG with your pet, 100% of time when you get down from the hill near the Alliance Graveyard, your pet stuck on the top of the hill and it cannot follow you. (Take a look of the picture below)
WoWScrnShot_021821_133921

Spell: Every PETS
Zone: Warsong Gulch

Pets should follow their masters without stucking on the top of the hill

##### CURRENT BEHAVIOUR:
Pets stuck on the top of the alliance graveyard in **Warsong Gluch**,  moreover in Horde Graveyard they have to climb from the right of the hill and the master have to wait for their arrival.

**Step 1** Join Warsong Gulch with a class that has Pet (hunter or warlock)
**Step 2** Get down from the top of Alliance Graveyard (100% of stuck the pet) or from Horde Graveyard (the pet has to climb from the right of the hill, and it comes to his master very slowly)

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/21f828847f42083f7fad32b18828d254c208baba

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

Related: https://github.com/chromiecraft/chromiecraft/issues/88

<bountysource-plugin>

---
There is a **[$35 open bounty](https://www.bountysource.com/issues/96653456-bug-core-pet-stuck-in-warsong-gulch?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

UltraNix:
Fixed #4603.

## Changes Proposed:
-  Added code that enables pets to always follow its master.

## Issues Addressed:
- Closes #4603

## Tests Performed:

- Tested ingame on Warsong Gulch with warlock's Imp

## Target Branch(es):
- [x] Master

 
### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -174,7 +174,7 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     }
 
     owner->AddUnitState(UNIT_STATE_CHASE_MOVE);
-    i_recalculateTravel = true;
+    i_recalculateTravel = false;
 
     Movement::MoveSplineInit init(owner);
     init.MovebyPath(i_path->GetPath());
@@ -347,7 +347,7 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     owner->AddUnitState(UNIT_STATE_FOLLOW_MOVE);
 
-    i_recalculateTravel = true;
+    i_recalculateTravel = false;
 
     init.MovebyPath(i_path->GetPath());
     init.SetFacing(target->GetOrientation());
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -334,8 +334,10 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     target->GetNearPoint(owner, x, y, z, _range, 0.f, target->ToAbsoluteAngle(tAngle));
 
+    Movement::MoveSplineInit init(owner);
+
     bool success = i_path->CalculatePath(x, y, z, forceDest);
-    if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
+    if ((!success || i_path->GetPathType() & PATHFIND_NOPATH) && !followingMaster) // Exclude pets - must always follow its master
     {
         owner->StopMoving();
         return true;
@@ -345,7 +347,6 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     i_recalculateTravel = true;
 
-    Movement::MoveSplineInit init(owner);
     init.MovebyPath(i_path->GetPath());
     init.SetFacing(target->GetOrientation());
     init.SetWalk(target->IsWalking());
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -157,7 +157,7 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     {
         if (cOwner)
             cOwner->SetCannotReachTarget(true);
-        owner->StopMoving();
+        i_recalculateTravel = false;
         return true;
     }
 
@@ -337,9 +337,11 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     Movement::MoveSplineInit init(owner);
 
     bool success = i_path->CalculatePath(x, y, z, forceDest);
-    if ((!success || i_path->GetPathType() & PATHFIND_NOPATH) && !followingMaster) // Exclude pets - must always follow its master
+    if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
     {
-        owner->StopMoving();
+        if (cOwner)
+            cOwner->SetCannotReachTarget(true);
+        i_recalculateTravel = false;
         return true;
     }
 
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -334,8 +334,6 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     target->GetNearPoint(owner, x, y, z, _range, 0.f, target->ToAbsoluteAngle(tAngle));
 
-    Movement::MoveSplineInit init(owner);
-
     bool success = i_path->CalculatePath(x, y, z, forceDest);
     if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
     {
@@ -349,6 +347,7 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     i_recalculateTravel = false;
 
+    Movement::MoveSplineInit init(owner);
     init.MovebyPath(i_path->GetPath());
     init.SetFacing(target->GetOrientation());
     init.SetWalk(target->IsWalking());
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -152,12 +152,13 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     if (owner->IsHovering())
         owner->UpdateAllowedPositionZ(x, y, z);
 
+    i_recalculateTravel = true;
+
     bool success = i_path->CalculatePath(x, y, z, forceDest);
     if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
     {
         if (cOwner)
             cOwner->SetCannotReachTarget(true);
-        i_recalculateTravel = true;
         return true;
     }
 
@@ -174,7 +175,6 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     }
 
     owner->AddUnitState(UNIT_STATE_CHASE_MOVE);
-    i_recalculateTravel = false;
 
     Movement::MoveSplineInit init(owner);
     init.MovebyPath(i_path->GetPath());
@@ -334,18 +334,17 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
 
     target->GetNearPoint(owner, x, y, z, _range, 0.f, target->ToAbsoluteAngle(tAngle));
 
+    i_recalculateTravel = true;
+
     bool success = i_path->CalculatePath(x, y, z, forceDest);
     if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
     {
         if (cOwner)
             cOwner->SetCannotReachTarget(true);
-        i_recalculateTravel = true;
         return true;
     }
 
     owner->AddUnitState(UNIT_STATE_FOLLOW_MOVE);
-
-    i_recalculateTravel = false;
 
     Movement::MoveSplineInit init(owner);
     init.MovebyPath(i_path->GetPath());
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -157,7 +157,7 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     {
         if (cOwner)
             cOwner->SetCannotReachTarget(true);
-        i_recalculateTravel = false;
+        i_recalculateTravel = true;
         return true;
     }
 
@@ -341,7 +341,7 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     {
         if (cOwner)
             cOwner->SetCannotReachTarget(true);
-        i_recalculateTravel = false;
+        i_recalculateTravel = true;
         return true;
     }
 


src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -334,18 +334,25 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
...
-        owner->StopMoving();
-        return true;
+        // Pet must always follow its master
+        if (followingMaster)
+            init.MoveTo(x, y, z);
+        else
+        {
+            owner->StopMoving();
+            return true;
+        }
```
Yehonal, line 349: little codestyle improvement (instead of using else):

```
        if (!followingMaster) {
            owner->StopMoving();
            return true;
        }
        
        init.MoveTo(x, y, z);
  ```

**Comment by Footman** [comment]
This would be a hack. Problem is with path calculation + slopes. It concerns not only pet but also random npcs. even in dalaran

**Comment by Footman** [comment]
it fixes only one part of problem when pet cant climb up.  often when pet want to come down it run around from the top very long way. also this happens not only with creatures but with normal npcs too. sometime it would be easier for them to "jump" down instead of chosing very long way.

**Comment by UltraNix** [comment]
@Footman Could you tell me where I can observe such problems? I tested in exactly that spot in WSG described in the issue.

**Comment by Footman** [comment]
> @Footman Could you tell me where I can observe such problems? I tested in exactly that spot in WSG described in the issue.

I tested on WG alliance side as well. Before this fix pet could not climb up. Now it can.
but when it want to climb down it doesnt work often and pet runs full way around.

**Comment by Footman** [comment]
> Yep, I checked that. In that cases, the path is normal, so pet decide to build and move by generated path - this PR do not cover that problem. I don't know if pet always should use a shortcut to owner. I'm not sure if it's a bug actually - I saw a lot of videos from retail when pet instead of jumping with owner, decided to run long way and pull mobs.

its still strange at some place pet can go down and and some not and it takes full length path. i think its wrong fore sure at this place in battleground

**Comment by sanctum32** [comment]
I'm not sure about two calls, https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R349 and https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R356

the both could get triggered at same time and MovebyPath probably will overwrite setting of MoveTo

**Comment by Yehonal** [comment]
> Yep, I checked that. In that cases, the path is normal, so pet decide to build and move by generated path - this PR do not cover that problem. I don't know if pet always should use a shortcut to owner. I'm not sure if it's a bug actually - I saw a lot of videos from retail when pet instead of jumping with owner, decided to run long way and pull mobs.

They do when there's a viable path. A viable path in AC is composed of no more than 60 nodes. When a path is longer than that it's flagged as NO_PATH and, if forced, it should create a shortcut instead. I doubt that on retail they always follow a path, I played a hunter for a while and I can assure you that they teleport the pet when there's a non-viable path. 

However, in this case, if @Footman is right and the slope filter causes other unwanted issues, we can remove the filter without any issue. At the moment we have several other checks to avoid climbing the slopes in certain conditions. That specific filter instead is intended to add a higher "difficulty value" for a certain node that contains a high slope, thus trying to avoid that a creature prefers to walk on a log or a rock instead of going around. But there are other pre-calculated solutions to it (e.g. increasing the slope angle threshold during mmap generation, maybe adding a flag in case we still want to keep that information).

**In conclusion**: I think that in ANY case, the pet MUST always have a path. There is no reason why a pet should stop and not follow you anymore, eventually, they must get a shortcut. There's actually something that can be improved in that particular PR but I'll write the feedback there directly

**Comment by sanctum32** [comment]
Well, in trinitycore following units at base keeps making shortcuts https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Movement/MovementGenerators/FollowMovementGenerator.cpp#L145

File: src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp [hunk 2]
```diff
@@ -334,18 +334,25 @@ bool FollowMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
...
-        owner->StopMoving();
-        return true;
+        // Pet must always follow its master
+        if (!followingMaster)
+        {
+            owner->StopMoving();
+            return true;
+        }
+
+        init.MoveTo(x, y, z);
```
Yehonal, line 349: I was wondering: I don't  think you need to use the MoveTo here since the path is calculated in any case (as a shortcut). 

If you follow the pathgenerator logic, you can notice that it generates a shortcut but returns the PATHFIND_NOPATH flag, we can use the calculated shortcut in any case.

Therefore, you can directly add the !followingMaster inside the first condition so the creature stop to move only when it's not a pet otherwise it will use the precalculated shortcut.

It will also solve the double call that has been reported

Yehonal (changes requested):

UltraNix:
> **In conclusion**: I think that in ANY case, the pet MUST always have a path. There is no reason why a pet should stop and not follow you anymore, eventually, they must get a shortcut. There's actually something that can be improved in that particular PR but I'll write the feedback on the code directly
So this is how it works now with this PR - pet always follow master. The existing problem is the "quality" of some paths.

Yehonal:
> I'm not sure about two calls, https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R349 and https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R356
> 
> i think, you could check variable "success" again
> what i mean:
> 
> ```c
>     if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
>     {
>         owner->StopMoving();
>         return true;
>         // Pet must always follow its master
>         if (!followingMaster)
>         {
>             owner->StopMoving();
>             return true;
>         }
> 
>         init.MoveTo(x, y, z);
>         success = false;
>     }
> ```
> 
> and check it:
> 
> ```c
> if (success)
> {
>    init.MovebyPath(i_path->GetPath());
> }
> init.SetFacing(target->GetOrientation());
> init.SetWalk(target->IsWalking());
> ```

I left a review, there's a better way for it

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
UltraNix: I don't get it. Where is that doubled call?

UltraNix:
> I'm not sure about two calls, https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R349 and https://github.com/azerothcore/azerothcore-wotlk/pull/5103/files#diff-17a98564bc33c45ee6563254522143d4285aa660f79f90a59c97bebb820ce626R356
> 
> i think, you could check variable "success" again
> what i mean:
> 
> ```c
>     if (!success || i_path->GetPathType() & PATHFIND_NOPATH)
>     {
>         owner->StopMoving();
>         return true;
>         // Pet must always follow its master
>         if (!followingMaster)
>         {
>             owner->StopMoving();
>             return true;
>         }
> 
>         init.MoveTo(x, y, z);
>         success = false;
>     }
> ```
> 
> and check it:
> 
> ```c
> if (success)
> {
>    init.MovebyPath(i_path->GetPath());
> }
> init.SetFacing(target->GetOrientation());
> init.SetWalk(target->IsWalking());
> ```

And what is the purpose of using `success` again? If not succes, then you have return. Otherwise, the path is good.

UltraNix:
@Yehonal How about now?

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
Yehonal: after the init.MoveTo the code go ahead and then there's the .MovebyPath because you do not return anymore.

so just remove the MoveTo and simplify the condition, the code can safely go ahead and it will use the shortcut since the path is forced

sanctum32:
other option for this, he could add extra check in https://github.com/azerothcore/azerothcore-wotlk/blob/858fb81681047c607f9044110309448f6cbaa2bf/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L269
```c
(owner->IsGuardian() || owner->IsPet())
```

UltraNix:
> other option for this, he could add extra check in
> 
> https://github.com/azerothcore/azerothcore-wotlk/blob/858fb81681047c607f9044110309448f6cbaa2bf/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L269
> 
> ```c
> (owner->IsGuardian() || owner->IsPet())
> ```

You might be right - not only pets, but guardians should always follow target too.

Yehonal:
> other option for this, he could add extra check in
> 
> https://github.com/azerothcore/azerothcore-wotlk/blob/858fb81681047c607f9044110309448f6cbaa2bf/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L269
> 
> ```c
> (owner->IsGuardian() || owner->IsPet())
> ```

I think that followingMaster includes the case when it's a guardian or a pet, so maybe it's not needed

UltraNix:
Pet != guardian.

sanctum32:
> Every pet is guardian.
> Not every guardian is pet.

well, for example, blood worms (DK), they are considered as guardians, not pet but they at base should keep such behavior

UltraNix:
That's right.

Yehonal:
@UltraNix may I change something directly on the PR? It's faster to do the change than explain what to do

UltraNix:
Okay.

Yehonal:
Ok so, what happens is: when you fall down from a cliff the pathgenerator does the following:

1. while you are in air there is no path at all (PATHFIND_NOPATH) because the pet cannot fly, so it must stay still. Meanwhile the recalculation flag is set at true and the SetCannotReachTarget is called (since there's another logic handled inside the AI)

2. When you reach the ground the recalculation happens and then there could be 2 situations: if a path exists (including a shortcut), then the creature will follow it, if it doesn't exist then, after a while, the pet enters in "evade mode". The evade mode for a Pet means that the following motion starts from the scratch. 

Now, it's very unlikely to have a situation where the path does not exist at all unless a temporary one such as you're flying or something like that, or something like: you close a door and your pet is on the other side. In this situation maybe the pet must be teleported because a brutal MoveTo could make it to fall into the texture

UltraNix:
Tested - looks the same, but I got the idea of this update.

FrancescoBorzi (approved):

FrancescoBorzi:
@azerothcore/testers @pklloveyou do you mind testing this?

UltraNix:
What are we waiting here for?

Yehonal:
Has been correctly tested? Does it work properly?

Il mar 13 apr 2021, 09:55 UltraNix ***@***.***> ha scritto:

UltraNix:
Okay. @pklloveyou could you retest it?

FrancescoBorzi:
@azerothcore/testers @mpfans @Footman can you please check this one? it fixes an issue caused by the last movement/pathfinding revamp

Sedatyf:
> @azerothcore/testers @mpfans @Footman can you please check this one? it fixes an issue caused by the last movement/pathfinding revamp

I can check it if you want to

Yehonal:
Whoever can do it, please let us know. A video would be appreciated as well

Sedatyf:
> Anyone can do it, please let us know. A video would be appreciated as well

I'm building this PR. I'll make a video of my walking with my pet

Sedatyf:
Alright, here is my results (with a lot of videos 😄). I tested with an imp and with Treants (Force of Nature talents from Druid) which are guardians. Everything seems to work perfectly. Both (pet and guardiana) take a long path when available. In an extreme way (meaning, stuck in a weird place and then jump down for example), they take shortcuts. They can climb up too. Even if there is no path available for a normal user. 
Please see list below for more details.

- I made 3 different jumps : 1 with a path next to my jump, 1 with no path (alliance respawn point) and 1 with a long path. Pet and guardians follow nicely with those 3 tests ✔️ 

https://user-images.githubusercontent.com/59012307/115004314-f6f5dd80-9ea6-11eb-81a7-fdb0083c1aea.mp4

https://user-images.githubusercontent.com/59012307/115004317-f8270a80-9ea6-11eb-821a-85ceb6b44800.mp4

https://user-images.githubusercontent.com/59012307/115004318-f8bfa100-9ea6-11eb-82a5-d4b0a1c6a8c0.mp4

- For the climb, I invoked my pet, ordered him to stay, changed position, then asked him to follow. Imp and Treants took the long path in Warsong Gulch and they also climbed even if there was no path available (stump video) ✔️ 

https://user-images.githubusercontent.com/59012307/115004848-8bf8d680-9ea7-11eb-8404-9fa5c210b001.mp4

https://user-images.githubusercontent.com/59012307/115004844-8ac7a980-9ea7-11eb-9290-b1c443447aeb.mp4

- Finally, I pushed those tests in an extreme situation to see how pet and guardians behaved. For a jump, they took the shortest path, even if there was no "ground path" (meaning they jumped). For the follow behavior, they took what it seems to be a path for them, even if a normal player couldn't climb those path ✔️ 

https://user-images.githubusercontent.com/59012307/115005438-2527ed00-9ea8-11eb-8cfb-2fddb4616c9c.mp4

https://user-images.githubusercontent.com/59012307/115005450-2822dd80-9ea8-11eb-8f19-30f79396740e.mp4

I only made one video for guardians behavior in order to keep this comment short (I tried, lol) but I can assure you that Treants had exactly the same behavior as Imp

https://user-images.githubusercontent.com/59012307/115005629-57394f00-9ea8-11eb-9314-90ff8ac221ab.mp4

Yehonal:
Outstanding testing job! That's how to do it!
Thank you!

Issue #5135, opened 2021-04-04
.WDT  MWMO chunk should be optional in map extractors


T1ti:
## CURRENT BEHAVIOUR:
.WDT files made by blizzard all have the MWMO chunk, but it is unused for all maps that have terrain. https://wowdev.wiki/WDT#MWMO.2C_MODF_chunks
Currently the extractors require all .wdt files to have this chunk.
The issue is that modding tools to generate custom maps usualy don't generate this chunk(because it's not needed), and many people struggle to use the extractors with custom maps for this reason.
This is the difference with a custom wdt created by noggit(the msot popular map editor) and blizzard's officials .wdt 
![image](https://user-images.githubusercontent.com/40864460/113524262-ee033300-95ad-11eb-9f76-715e4904d0b1.png)

## EXPECTED BLIZZLIKE BEHAVIOUR:
Remove the requirement for the MWMO chunk for maps with terrain.(non wmo only maps). It is unused by the client
Here's the faulty code https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/tools/map_extractor/wdt.cpp#L68 
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/tools/map_extractor/System.cpp#L980
Afaik the extractor deosn't even use that chunk at all

## STEPS TO REPRODUCE THE PROBLEM:

1. 
2. 
3. 

## AC HASH/COMMIT:
https://github.com/azerothcore/azerothcore-wotlk.git

PR #5138, opened 2021-04-05
fix(Core/Tools): extracting custom maps

T1ti:
## Changes Proposed:
-  Comment a check for the MWMO chunk in .wdt files that modding tools don't generate, which is not needed and not even used by the client
-  This allows maps generated by modding tools to be extracted properly

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5135

## Tests Performed:

- Tried extracting maps before and after the change
- Before : (pkr_broken is my custom map)
image
-After : 
image

## How to Test the Changes:

- Extract normal and custom maps

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Faq:
Better way would be implement option to extract also custom maps.

FrancescoBorzi (approved):

Wowkasu:
Thank you so much from the bottom of my heart!!! this is so much needed for the non-standard servers that are becoming more and more numerous.

Issue #5178, opened 2021-04-07
[BUG] Wrong Async queries (especially for pets) and room for improvements


Yehonal:
I've found a technical issue related to the way certain functions are using the AsyncQuery method.

Currently we have 3 ways to execute a query:

1. Enqueue without result (Execute)
    **pro**: it doesn't lock the db so doesn't cause performance issues
    **cons**: doesn't return anything and you cannot write procedural code

2. Execute with sync result (Query)
    **pro**: returns the result of the query and you can write procedural code
    **cons**: it locks the db so can cause performance issues 
   
3. Enqueue with async result (AsyncQuery)
    **pro**: returns the result of the query and you can write async code
    **cons**: it's a bit more complex to use (not 1 line of code)
    
## Issue N.1: Pets etc.

I've noticed that certain methods are using the AsyncQuery as they were synchronous instead.

This is the list of wrongly used async queries:

* https://github.com/azerothcore/azerothcore-wotlk/blob/4b7454574412cfe23bb5902b9b3d42d2313b3541/src/server/game/Handlers/NPCHandler.cpp#L590

* https://github.com/azerothcore/azerothcore-wotlk/blob/4b7454574412cfe23bb5902b9b3d42d2313b3541/src/server/game/Entities/Pet/Pet.cpp#L103

* (huge file, i've attached the screenshot) https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Player/Player.cpp#L28232

![image](https://user-images.githubusercontent.com/147092/113891251-30528d00-97c5-11eb-85c6-97a91af8c130.png)

I'm not 99% sure but I think that every time you run those queries in the best scenario it's a gamble against your I/O speed, in the worst scenario, that logic is completely broken

## Issue N.2: Complexity (optional)

Would be great to find a way to allow devs to easily use the AsyncQuery approach since, in my opinion, it should be vastly used especially for scripts.

Yehonal:
bump

Winfidonarleyan:
#4669

Yehonal:
> #4669

not sure if that is related to this problem

Yehonal:
Any updates here? @Winfidonarleyan were you working on Async recently?

FrancescoBorzi:
not sure if @UltraNix did something about this recently

Winfidonarleyan:
> Any updates here? @Winfidonarleyan were you working on Async recently?

No

[...]

PR #5273, opened 2021-04-13
fix(Core/Spells): Do not allow the use of the master's call while the pet is controlled [port from TC]

ErnestoGonzalezC:
The hunter can use the master's call even if his pet is under the
effects of polymorphs, fears, etc.
This RP prevents the use of the master's call while the pet is under the
effects of stun, fear, disorientation.

## Changes Proposed:
-  fix(Core/Spells): Do not allow the use of the master's call while the pet is controlled.

## Issues Addressed:
- Closes 

## Tests Performed:

- Compiles on windows perfectly.
- 

## How to Test the Changes:
Log in with a hunter and another character, for example, a mage to polymorph. Apply polymorphism or fear to the pet and you will see that the master's call can no longer be used if the pet is under such effects.

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

Kitzunu:
I think it is better if we move it to a spellscript in spell_hunter.cpp

You can look at https://github.com/TrinityCore/TrinityCore/blob/6a46817f5d0a9ec760cc16b0871b74c40033f69a/src/server/scripts/Spells/spell_hunter.cpp#L664

1. Because it makes spell.cpp smaller
2. SPELL_FAILED_NOT_IN_CONTROL would be wrong to use as stun should return SPELL_FAILED_STUNNED, confuse SPELL_FAILED_CONFUSED etc

ErnestoGonzalezC:
@Kitzunu thanks for the suggestion, i will try

ErnestoGonzalezC:
@Kitzunu I could check the code, I'm not sure if I did it the best way.

Kitzunu:
Looks good, if you want to you can change HandleScriptEffect to
```cpp
GetHitUnit()->CastSpell((Unit*)nullptr, SPELL_HUNTER_MASTERS_CALL_TRIGGERED, true);
```
and test in-game it the spell works as it should. (Also move it last)

ErnestoGonzalezC:
> Looks good, if you want to you can change HandleScriptEffect to
> 
> ```c++
> GetHitUnit()->CastSpell((Unit*)nullptr, SPELL_HUNTER_MASTERS_CALL_TRIGGERED, true);
> ```
> 
> and test in-game it the spell works as it should. (Also move it last)

Done i already tested it in game and it works fine. Thanks for the help.

T1ti:
It should also work this way for other spells such as Vodiwalker's Sacrifice, dunno how it works currently, just saying
If the voidwlker is CCd, you can't sacrifice it

pklloveyou:
Tested

Sedatyf:
Hey there.
So I made a hunter on an account that has GM rights. Then I used ``.pet create`` on a boar to have a pet.
Then I created 3 characters : a warlock, a mage, and a rogue.
I dueled my hunter with those characters and made them their control spells : fear for warlock, polymorphe for mage, and blind for rogue.

I can confirm that master's call doesn't break those spells anymore on pet. Please see video attached below.
And finally master's call still works when control debuff ends.


https://user-images.githubusercontent.com/59012307/114939984-7355d580-9e41-11eb-8262-d804f8b1de32.mp4

https://user-images.githubusercontent.com/59012307/114939657-0f331180-9e41-11eb-9ab2-e78325a44d4d.mp4

https://user-images.githubusercontent.com/59012307/114939660-0fcba800-9e41-11eb-969d-606a9e48381b.mp4

ErnestoGonzalezC:
@Sedatyf Excellent review 👍

FrancescoBorzi:
this is copied from TrinityCore: https://github.com/TrinityCore/TrinityCore/commit/361b185ef7bfb0eff3e82c0b12deaf5b597bebee

@ErnestoGonzalezC it's ok to import stuff from TC, just remember to mention the credits to the original author

in this case, credits to @ariel-

@Sedatyf excellent job with testing!

src/server/scripts/Spells/spell_hunter.cpp
```diff
@@ -691,27 +691,54 @@ class spell_hun_masters_call : public SpellScriptLoader
             return ValidateSpellInfo({ SPELL_HUNTER_MASTERS_CALL_TRIGGERED });
         }
 
+        SpellCastResult DoCheckCast()
+        {
+            Guardian* pet = GetCaster()->ToPlayer()->GetGuardianPet();
+            ASSERT(pet); // checked in Spell::CheckCast
```
Yehonal, line 697: why do we need to crash the server here? ...also consider to fix this first: https://github.com/azerothcore/azerothcore-wotlk/issues/5178 otherwise it could crash easily

ErnestoGonzalezC:
@FrancescoBorzi as you can see, this pr did not start as any type of tc port, but because @Kitzunu  suggested that I move all the code to 1 separate script and he recommended me to look at trinitycore because I had to do it like this, I need to change the title :)

src/server/scripts/Spells/spell_hunter.cpp
ErnestoGonzalezC: I will investigate about that

Sedatyf:
Hey @Yehonal, this PR is done and tested, can we have a review and a change in tags? Thanks in advance

Kitzunu (approved):

PR #5277, opened 2021-04-13
fix(SQL): dungeon_access_template min_level

locus313:
## Changes Proposed:
- update min level for each dungeon in dungeon_access_template to use the values from the dbc files

## Issues Addressed:
- Closes 

## SOURCE:

https://wow.tools/dbc/?dbc=lfgdungeons&build=3.3.5.12340#page=1

## Tests Performed:

- apply sql

## How to Test the Changes:

- apply sql
- set `DungeonAccessRequirements.LFGLevelDBCOverride = 1` in your worldserver.conf
- run `.reload dungeon_access_template` & `.reload conf`
- use rdf at different level's, make sure your not placed into dungeons too far above your level.

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 

mdawx:
Closes https://github.com/chromiecraft/chromiecraft/issues/420

Kitzunu:
?

locus313:
Ended up finding out why max_level was keep at 0. :) When i actually tested, it locks players out that are a higher level. Which isn't desired i think :) The min_levels could probably still be updated though (they seemed a little off when i checked them). I just closed it for now, not removing the branch though.

FrancescoBorzi:
@locus313 what's the status of this PR?

locus313:
This pr should be good to go. This should get the min values for each dungeon in line with the dbc values.

FrancescoBorzi:
@azerothcore/testers please review

Nefertumm:
Tested. Seems to be ok, but I found something odd, I was level 23 and still got into Ragefire Chasm, and that dungeon should pop until lvl 18. So idk if that's good or not.

claudiodfc:
### Test Result ✔
## Test case 1
At level 16 got into the 4 available maps: WC, RFC, SFK and DMines.

### Test case 2

At level 19 got into the 5 available maps: WC, RFC, SFK, DMines and Blackfathom Deeps.

### Test Case 3
At level 50 got placed into The Stockades the first queue.

### Notes:
You have not updated the Database with a Maximum level on `dungeon_access_template.max_level` column, inducing the behaviour that @Nefertumm specified. Is the core already using the mention column too?

image

locus313:
> ### Notes:
> 1. You have not updated the Database with a Maximum level on `dungeon_access_template.max_level` column, inducing the behaviour that @Nefertumm specified. Is the core already using the mention column too?

This pr is only meant to update the min level's to be closer to the dbc values. if we set the max_level column it will prevent higher level players from being able to enter the lower level dungeons (even without using RDF). Hence that's why its currently set to 0 for everything. The behavior you mentioned already happens without this pr if you enable 
 `DungeonAccessRequirements.LFGLevelDBCOverride` .


> image

if you go back and look at the tool there is a MinLevel column, in most cases this was higher than the currently set min_level values in the database. So i used the value from this column.

claudiodfc:
> This pr is only meant to update the min level's to be closer to the dbc values. if we set the max_level column it will prevent higher level players from being able to enter the lower level dungeons (even without using RDF). Hence that's why its currently set to 0 for everything. The behavior you mentioned already happens without this pr if you enable

Well, but if we can fix the Max level at the same time then even better no? Why restrict this PR only for Min level, just add the max level and you kill 2 birds with 1 stone.

locus313:
> > This pr is only meant to update the min level's to be closer to the dbc values. if we set the max_level column it will prevent higher level players from being able to enter the lower level dungeons (even without using RDF). Hence that's why its currently set to 0 for everything. The behavior you mentioned already happens without this pr if you enable
> 
> Well, but if we can fix the Max level at the same time then even better no? Why restrict this PR only for Min level, just add the max level and you kill 2 birds with 1 stone.

Because i don't think its a desired effect to lock out higher-level characters from doing lower-level dungeons.

claudiodfc:
Yes, It is.
Otherwise you're gonna have level 80 players queued for Wailing Caverns, and this isn't a desired effect.

locus313:
> Yes, It is. Otherwise you're gonna have level 80 players queued for Wailing Caverns, and this isn't a desired effect.
> 
> Even worse, for people that level solely on RDF, they could be level 40 and still be placed on lower level instances, and this makes it impossible to level up as. That's why there's a threshold for the levels able to queue for those instances.

But like I mentioned it also prevents higher-level characters from just walking into the dungeon as well. (which is the part I think is not desired) I do agree that we dont want this to happen in RDF. But think some other changes are required to address that part.

claudiodfc:
That might be why there's more columns about this. Needs more investigation.

acidmanifesto:
@locus313 what is the current status of this PR?

acidmanifesto:
@locus313 please let us know if this pr is still being worked on or not. If not I will have to Mark it as abandon.

locus313:
As I mentioned before this pr is only targeted at updating the min levels. I do not really have the time right now to investigate the other items brought up.

PR #4369, opened 2021-01-29
feat(CI): restructure docker build

locus313:
## Changes Proposed:
-  restructure the docker build to include additional checks.

## Issues Addressed:
- Closes 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #5294, opened 2021-04-14
Docker Multiple Realms

### LINKED ISSUE #5230: Feature: Support Multiple Realms with Docker Installation
**Describe your feature request or suggestion in detail**

Users should have the ability to add realms to their docker installation.

**Describe a possible solution to your Feature or Suggestion in detail**

I think (and I'm scratching my head because there's a lot to the build process of the docker containers) that basically you'd need to:

1. Update `azerothcore-wotlk/docker/database/generate-databases.sh` to create more (dependent on the number of additional realms you want) `acore_world` and `acore_character` databases (with names appended by a realm number, i.e. `acore_world_realm1`).
2. Add additional services to the `azerothcore-wotlk/docker-compose.yml` for each new realm. Something like:
```yml
# Generate new service names for new realms
ac-worldserver_realm1:
    stdin_open: true
    tty: true
    # Would the image need to change?
    image: azerothcore/worldserver
    restart: unless-stopped
    privileged: true
    build:
      context: ./docker/worldserver
      dockerfile: Dockerfile
    networks:
      - ac-network
    ports:
      # Set external ports to incremented value, avoid duplicates, might have
      # this wrong, not very familiar with what the left/right values denote.
      - ${WORLD_EXTERNAL_PORT:-8086}:8086
      - ${SOAP_EXTERNAL_PORT:-7879}:7879
    volumes:
      # This part is more hazy to me. data should certainly be shared (unless
      # one realm has different content than another), but I believe that
      # only the worldserver.conf and perhaps the log directories have to be
      # different.
      - type: bind
        source: ./docker/realmserver1/bin
        target: /azeroth-server/bin
      - type: bind
        source: ${WORLDSERVER_ETC:-./docker/worldserver/etc}
        target: /azeroth-server/etc
      - type: bind
        source: ${WORLDSERVER_LOGS:-./docker/realmserver1/logs}
        target: /azeroth-server/logs
      - type: bind
        source: ${WORLDSERVER_DATA:-./docker/worldserver/data}
        target: /azeroth-server/data
    depends_on:
      - ac-database
```
3. The `worldserver.conf` found in the `./docker/realmserver1/bin` from the step above would need to include these values:
```bash
WorldServerPort       = 8086
SOAP.IP               = "ac-worldserver-realm1"
SOAP.Port             = 7879
LoginDatabaseInfo     = "ac-database;3306;root;password;acore_auth"
WorldDatabaseInfo     = "ac-database;3306;root;password;acore_world_realm1"
CharacterDatabaseInfo = "ac-database;3306;root;password;acore_characters_realm1"
```
4. Duplicate the entry in your `acore_auth.realmlist` table for the initial `Azerothcore` realm, changing the `name` to whatever you want to call it and the  `port` value to match the realm's port value (as configured above)

**Additional context**

I think that's the gist of it. You'd run the `docker-compose up` command and the services would run as before all at once in their individual containers.

I'm certain someone with more knowledge of how these containers are built and configured could actually add a script to generate this data automatically post-install. Something like an `addrealm.sh` or something.

ReynoldsCahoon:
## Changes Proposed:
-  Update `docker-compose.yml` to include two optional (off by default) profiles for additional realms
-  Add optional `docker/database/generate-databases-with-realms.sh` that creates the needed databases for additional realms
-  Add `realm-one` and `realm-two` configuration defaults
-  Add modifiable environment variables to `.env-dist` for additional realms

## Issues Addressed:
- Closes #5230

## SOURCE:

-  N/A

## Tests Performed:

-  Tested on Linux

## How to Test the Changes:

-  Follow the [documentation to install with Docker](https://www.azerothcore.org/wiki/Install-with-Docker) until you reach Step 3
-  Within `docker/database/Dockerfile`: Comment line 31, uncomment line 34
-  Instead of Step 3's command, run `docker-compose --profile realm-one up`
image
Screenshot from 2021-04-14 14-49-55

**IMPORANT:** Any previous database volumes must be removed, and the database container must be rebuilt, any time changes are made to the database Dockerfile

- [ ] Generating the databases multiple times from the same source files is expensive. Since the database is compiled on build of the database container, it seems there is potential for optimization [via this method](https://github.com/azerothcore/azerothcore-wotlk/issues/5230#issuecomment-819354771) to generate only the databases for the realm containers that you build, however there are issues. This would also require the database container to run once before running the additional realms, so that the database already exists prior to attempts to duplicate it.
- [ ] Documentation

## Target Branch(es):
- [x] Master

I am up for criticism on implementation. I know that some would suggest spinning up another set of containers entirely and then modifying the values as necessary, but I wanted to limit the cost on a single machine (reducing necessary containers) for testing multiple realm functionality, and do it in a way that reduces the amount of steps someone would have to take to implement it.

 

ReynoldsCahoon:
Oh right, additional information:

If you wanted to run two realms, instead of just the one, you'd need to
-  uncomment the realm_two lines in `generate-databases-with-realms.sh`
-  run the following (after removing previous database volumes and rebuilding)
```
docker-compose --profile realm-one --profile realm-two up
```

locus313:
@ReynoldsCahoon i think the idea of having a docker option for multiple realms is great. However @Yehonal is working on a pretty major rework of the docker structure/setup https://github.com/azerothcore/azerothcore-wotlk/pull/4369 .

I think we should hold off on merging this kind of change until after his pr is merged. Since it will require changing the current structure you have purposed.

ReynoldsCahoon:
> @ReynoldsCahoon i think the idea of having a docker option for multiple realms is great. However @Yehonal is working on a pretty major rework of the docker structure/setup #4369 .
> 
> I think we should hold off on merging this kind of change until after his pr is merged. Since it will require changing the current structure you have purposed.

Oh! I wasn't aware of this restructure! I'll review what's being worked on.

Yehonal:
I'm not sure it's the best approach, to create multiple realms you can always use the docker-compose.override.yml to extend the default one: https://docs.docker.com/compose/extends/

We can provide an article on how to do it with AC but basically, the docker official documentation covers 80% of what you need to do. You can already do it with the current configuration but will be even easier after my PR will be merged

Yehonal:
We can also consider to reuse your idea about a script to generate new configurations for new realms. So thank you for your effort :)

ReynoldsCahoon:
> We can also consider to reuse your idea about a script to generate new configurations for new realms. So thank you for your effort :)

For sure. I learned a lot, so no skin off my back if this is discarded. I was responding to a prompt in Discord from a user wondering how easy it was to add multiple realms with Docker. Some suggested just spinning up an entirely separate stack, which is a waste, as the disk space and extra containers that would run would be unnecessary, so I determined to try and get it working more efficiently. ;)

Perhaps something like this can be implemented in a better way once the larger rework is complete.

FrancescoBorzi:
what's the status of this PR?

FrancescoBorzi:
@ReynoldsCahoon  feel free to reopen this when you want to complete it. Closing for now.

PR #5317, opened 2021-04-16
refactor(DB/spell_dbc): Update AttributesEx names

Kitzunu:
## Changes Proposed:
-  Match core

## Target Branch(es):
- [x] Master

 

ErnestoGonzalezC:
tested

Yehonal:
Before doing it we need to upgrade the [DBC_SQL](https://github.com/wowgaming/client-data/releases/tag/dbc_sql_v1) with the new structure.

I think the best solution would be to create a pipeline that upgrades those files automatically with the structural changes. Those files must contain the original DBC data, so we need to process only `ALTER TABLE` queries on it.

Yehonal:
@Kitzunu can you please update the schema here: https://github.com/wowgaming/node-dbc-reader/tree/master/src/schemas ?

Yehonal (approved):

PR #5371, opened 2021-04-20
fix(Core): fix chests not despawning when opened in group

r-o-b-o-t-o:
## Changes Proposed:
-  Deactivate chests when their last item has been rolled by a group

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5362
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5128

Video displaying the current (bugged) behaviour:

https://user-images.githubusercontent.com/24357633/115364584-d211ab00-a1c3-11eb-936a-499f5ec0bc12.mp4

## Tests Performed:

- Tested in-game on Windows 10

## How to Test the Changes:

- Log in on two separate accounts
- With one account, invite the character from the other account in a group
- Go to a chest with both characters (example: `.go o 30950`)
- Make sure that the loot rule is not Free For All
- Open the chest, take all gray/white items, close the chest
- With both characters, roll on the green/blue items
- Verify that the chest is despawning when all items have been rolled for

See example video:

https://user-images.githubusercontent.com/24357633/115364603-d6d65f00-a1c3-11eb-9dd6-565126c5ab38.mp4

## Known Issues and TODO List:
**NOTE**: this is my first C++ PR here, so please review this carefully 😁 if you think it could be fixed in a better way, write a comment and I'll edit the changes.

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #5386, opened 2021-04-21
Revert "fix(Core): fix chests not despawning when opened in group"

FrancescoBorzi:
Reverts azerothcore/azerothcore-wotlk#5371


Sedatyf:
@Footman can you tell me how do you reproduce those crashes? When I tested it, I had no problem

PR #5394, opened 2021-04-21
fix(DB/GameObject): Table `gameobject` cleanup

alchem1ster:
## Changes Proposed:
-  Cleaning the `gameobject` table from duplicates
- Hopefully this can also increase performance or startup time

## Issues Addressed:
- [My own suggestion](https://github.com/azerothcore/azerothcore-wotlk/issues/5379)

## Tests Performed:
- NOT TESTED !!!
- Hopefully, it will be through this PR that we can check it out :)

## How to Test the Changes:
 - I don't know, seriously, because this fix removes above 4,000 duplicate game objects from world
 - Personally, I think this patch should be installed and tested for a long time, because there may be some side changes

## Target Branch(es):
- [x] Master

 

alchem1ster:
`Pool Id 202482 has no equal chance pooled entites defined and explicit chance sum is not 100. The pool will not be spawned.`
I don't know how to fix it automatically. Any ideas? If this is a problem for the pool, there may be several such situations among the 4000 remote objects.

LukasVolgger:
?

PR #5415, opened 2021-04-23
chore(DB/gameobject): table `gameobject` cleanup

alchem1ster:
## Changes Proposed:
- Cleaning the `gameobject` table from duplicates using the formula for calculating the Euclidean distance (`d`) between two objects
  - `d` is less or equal to 0.1
  - `d` is bigger than 0.1 and less or equal to 0.5
- Removing duplicate minerals or herbs in realm
  - `d` is bigger than 0.5 and less or equal to 10.0
- Hopefully this can also increase performance or startup time
  - a total of 1,277 objects were removed

_THIS PR DOES NOT AFFECT GAME OBJECTS THAT ARE IN THE POOL_

## Issues Addressed:
- [My own suggestion](https://github.com/azerothcore/azerothcore-wotlk/issues/5379)

## Tests Performed:
- NOT TESTED!
- Hopefully, it will be through this PR that we can check it out :)

## How to Test the Changes:
- I don't know, seriously, because this fix removes above 1,200 duplicate game objects from realm
- Personally, I think this patch should be installed and tested for a long time, because there may be some side changes
- Of the things that can be checked fairly quickly are the removal of duplicate herbs or minerals. Just look at the GUID that have been removed and what original GUIDs they refer to. Make a `.go` and check that there is no longer a duplicate there

## Target Branch(es):
- [x] Master

 

alchem1ster:
> In case something were to go wrong, could you add a Gist with all entries that you are removing, incase we'd need to restore something in the future ^^

Do you mean something like this? :)

[wayback for 1619185216044797385](https://github.com/azerothcore/azerothcore-wotlk/files/6367790/wayback_1619185216044797385.txt)
[wayback for 1619187862079970140](https://github.com/azerothcore/azerothcore-wotlk/files/6367791/wayback_1619187862079970140.txt)
[wayback for 1619185182789796765](https://github.com/azerothcore/azerothcore-wotlk/files/6367792/wayback_1619185182789796765.txt)

Kitzunu:
Yea, but rather in https://gist.github.com so I don't have to download them x) 

But yes, something like that, but a comment with the Gob name would be useful too

alchem1ster:
> Yea, but rather in https://gist.github.com so I don't have to download them x)
> But yes, something like that, but a comment with the Gob name would be useful too

@Kitzunu 
I only have names for herbs and minerals :)
https://gist.github.com/alchem1ster/3ee374b5365b857973f4f327e17bb89a
https://gist.github.com/alchem1ster/d11eecec8592d344c9a207048f87ab36
https://gist.github.com/alchem1ster/75d3c233cd59bcc89104ff1cc4dac0af

alchem1ster:
And if anyone needs it, here's a list of all the minerals (ores/veins) and herbs that are used in `3.3.5a`
https://gist.github.com/alchem1ster/0ba45ca4c6e9ad7a12a8116a2c7b8367

ErnestoGonzalezC:
I do not know if this is the correct way to test it but I have tried with `` .go gobject guid`` and most of the guid eliminated here, do not exist.

alchem1ster:
> I do not know if this is the correct way to test it but I have tried with ` .go gobject guid` and most of the `guid` eliminated here, exist but are not visible.

Yes and no. This way you can check the removal of duplicate minerals and herbs.
For other game objects affected by this PR, this check will not work properly. For them, it is more likely to assess the fact that no game events have been broken.

Hacki95:
Objects, especially herbs and mining nodes should NOT be removed just because they are close to each other.

Those objects are simply missing pooling data. The only drops which are wrong are the ones which spawn inside each other.

I am not sure this PR should be merged. I have a sense that this will remove a few gameobjects which are not wrong at all.

alchem1ster:
> Objects, especially herbs and mining nodes should NOT be removed just because they are close to each other.

Thank you for the feedback. 
This PR is less about veins and herbs, the main focus has been on objects that spawn within each other (d<0.5). We can remove the part associated with them and leave else.
Also please note `STAGE 3` in the issue to which I am referring. There is a list of veins and herbs that need attention in terms of their pooling.

alchem1ster:
I think that this PR is no longer required in its entirety. I will create a separate PR with identical objects only.

PR #5433, opened 2021-04-24
refactor(Docker): use 'docker compose' instead of 'docker-compose'

FrancescoBorzi:
image

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -24,7 +24,7 @@ services:
     # Update this to wherever you want VS Code to mount the folder of your project
     #- .:/workspace:cached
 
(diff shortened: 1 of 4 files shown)


locus313:
@FrancescoBorzi what version of docker are you running? From the looks of it, the docker version running in the github runners doesn't have this yet.

locus313:
looks like this feature is still in tech preview https://docs.docker.com/compose/cli-command/ and for Linux you actually have to install the tech preview to use it.

Yehonal:
Indeed, as I said @FrancescoBorzi this feature is experimental and ubuntu 20.04 doesn't have it in its package.
So would be better to use the docker-compose instead

Issue #5513, opened 2021-04-28
[MINING] Double Node (Tin/Silver) in Silverpine Forest


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [X] enUS
- [ ] other (specify)

##### FACTION

- [ ] Alliance
- [X] Horde

##### CONTENT PHASE:

- [X] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/474

Double Node of Tin Vein and Silver Vein at [59,50|75,50] in Silverpine Forest.

Node GUIDS are 35547 and 42971.

![WoWScrnShot_042821_201838](https://user-images.githubusercontent.com/81782124/116392144-405b0b00-a85f-11eb-93af-14b01137a06d.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

No double Node spawn.

##### STEPS TO REPRODUCE THE PROBLEM:

1.  .go o 42971
2.  Observe.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/91081f4ad8fed98d7bdc9c0ae13d05c86032df9f

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)
- [lua-CarbonCopy](https://github.com/55Honey/Acore_CarbonCopy)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

ErnestoGonzalezC:
Which of the two should be removed? or just change the location of one?

Azcobu:
[...]

PR #5514, opened 2021-04-28
fix(DB/Gameobject): Remove duplicate mining node.

ErnestoGonzalezC:
## Changes Proposed:
-  Remove duplicate mining node

## Issues Addressed:
- Closes 

- https://github.com/azerothcore/azerothcore-wotlk/issues/5513

## Tests Performed:

- ``.go o 35547``
- Verify that there is no duplicate node

## Target Branch(es):
- [x] Master

 

FrancescoBorzi (approved):

pklloveyou:
Tested

Si1ker:
@ErnestoGonzalezC Don't delete content, if is already in the DB is for a reason, move it.

Any duplicate is seen as "duplicate" just because they are together, so the right fix is to move them away so we eliminate the "duplicate" behaviour (unless is a unique NPC/Gobject)

So please move them instead of deleting them. Take this note also for future references in case you encounter any duplicate.

T1ti:
They weren't duplicate, they had a different ID
image
One is tein vein, the other is silver.

They just need to be pooled so they don't spawn at the same time

FrancescoBorzi:
I agree, I've overlooked this one earlier

PR #5550, opened 2021-04-30
feat(CI): PR Labeler

Kitzunu:
Not yet implemented it on Azerothcore/Github-Actions

So I remember what everything does :D 

https://github.com/dorny/paths-filter
https://github.com/TrinityCore/GitHub-Actions/tree/pr-labeler


Kitzunu:
Actually I want to do this in a fork to avoid clogging the pipeline as it runs both Push and PullRequest CI's

PR #5556, opened 2021-04-30
fix(Core/BattlegroundAB): Add missing Victory near sound

Kitzunu:
Kind of a cp of https://github.com/TrinityCore/TrinityCore/commit/2ac7499f66a80296486b7df6286ff503c8cdf3b9

## Changes Proposed:
-  Add horde sound  

## Tests Performed:

- None 

## How to Test the Changes:

Play Arathi Basin as normal and check if the correct horde sound is playing

## Target Branch(es):
- [x] Master

 

Helias (approved):

Kitzunu:
Assuming it played a sound before the PR, I don't think we need to test this...

Helias:
I agree

PR #5569, opened 2021-05-01
chore(template): Make them less intimidating

Kitzunu:
## Changes Proposed:
Part https://github.com/azerothcore/wiki/pull/454

My goal is to make the size of the files smaller for 2 reasons.

1. To make them less intimidating.
    1. Previously both templates were big and bombarded with information, and some that are not really necessary but just there to fill out the template.
    2. By just having the most vital parts there makes them slimmer and seem more welcoming to a new user, or an old contributor.
    3. We still have the most vital explanations of each topic in the template as well as each topic is very self-explanatory.
    4. We have a further explanation on the forum
2. To reduce the blank space spam in the webhook
image


FrancescoBorzi (approved):

Winfidonarleyan:
Good pr

Kitzunu:
Nobody merge this until wiki pr is merged

Issue #5625, opened 2021-05-04
Copper Ore gives too much Rough Stone


Hacki95:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [X] enUS
- [ ] other (specify)

##### FACTION

- [ ] Alliance
- [X] Horde

##### CONTENT PHASE:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

The [Copper Ore](https://wowgaming.altervista.org/aowow/?object=1731) loot table seems to pull the [Rough Stone](https://wowgaming.altervista.org/aowow/?item=2835) twice, resulting in 2-12 instead of 1-6 rough stone. See the screenshot after aprox. 40 mined ores in Mulgore:

![WoWScrnShot_042021_122144](https://user-images.githubusercontent.com/81629712/115383933-b237b280-a1d6-11eb-878e-efc7f836cb48.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

According to the database of [Copper Ore](https://wowgaming.altervista.org/aowow/?object=1731), only 1-6 Rough Stone should be looted by mining one vein.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Mine several Copper Nodes
2. Keep close attention to the amount of looted Rough Stone

##### EXTRA NOTES:

Linked issue: https://github.com/chromiecraft/chromiecraft/issues/448

Was already reported once in Issue #163, but dropped as it could not be recreated and only occured once. The screenshot above shows several occurences.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/b8b5f68f98ff8771a6ec87ea6805ceced91c406a

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
[...]

PR #5628, opened 2021-05-04
fix(DB/gameobject_loot_template): Correct drop chances for all classic ores

Hacki95:
Co-Author: @necropola

## Changes Proposed:
- Correct the dropchances of all types of "Stones" from classic ores.
- Correct the dropchances of all types of "Ores" from classic ores.
- Correct some groupIds

## Issues Addressed:
- Closes: https://github.com/chromiecraft/chromiecraft/issues/448
- Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/5625

## SOURCE:

- I used a mix of classic wowhead data, math and simple guessing. Since wotlk nodes aren't looted 3 times anymore but only once. So for most numbers I simply took the min/max value and multiplied it by 3. Since some drops already had 2-4 instead of 1-3 I went with similiar numbers for the others.
- The numbers might be guessed but they are way closer to the correct ones I assume. For example most stone types had a drop chance of 100%. Every source points to them being a more moderate 40% so I went with that number.
- I am not 100% sure every min/max number is completely correct now but they should only be off by 1 if they are. Not much of a chance to be 100% sure until wotlk classic hits. Its way better than before atleast.

## Tests Performed:

- Applied the SQL
- Checked the Database

## How to Test the Changes:

- Apply the SQL
- Go do some mining

- No way to be 100% sure about min/max values atm but they are way closer to the correct ones than before. If it turns out they are off by about 1 its a simple fix.

## Target Branch(es):
- [x] Master

 

data/sql/updates/pending_db_world/rev_1620115611701757811.sql
```diff
@@ -0,0 +1,59 @@
+-- INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1620115611701757811');
+
+-- Copper Vein
+UPDATE `gameobject_loot_template` SET `MaxCount`=7 WHERE FIND_IN_SET (`Entry`,'1502,1735,2626,18092') and Item = 2835; -- Rough Stone
```
Kitzunu, line 4: Big AND

Hacki95: Fixed

FrancescoBorzi (approved):

data/sql/updates/pending_db_world/rev_1620115611701757811.sql
```diff
@@ -0,0 +1,59 @@
+-- INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1620115611701757811');
```
FrancescoBorzi, line 1: why commented? we need it

Hacki95: Ohh you are right I always comment this out while testing. Fixed!

necropola:
**An overview of the expected effect on the average drop rate for the different types of Ore & Stone**

image

Hacki95:
Found some better data by using the waybackmachine on wowhead and analyzing the source from the site.

I updated the google sheet. Huge thanks again to @necropola for making it.
grafik

data/sql/updates/pending_db_world/rev_1620115611701757811.sql
```diff
@@ -0,0 +1,74 @@
...
+@HEAVYSTONE = 2838,
+@IRONORE = 2772,
+@SOLIDSTONE = 7912,
+@MITHRILORE = 3858,
+@DENSESTONE = 12365,
+@SOULDARITE = 19774,
+@THORIUMORE = 10620;
+
+-- Copper Vein
+UPDATE `gameobject_loot_template` SET `MaxCount`=6 WHERE FIND_IN_SET (`Entry`,'1502,1735,2626,18092') AND Item = @ROUGHSTONE; -- Rough Stone
```
Kitzunu, line 17: backtick on item

necropola:
I've added an additional column to the sheet which contains the `Loot entry (Data1)` value and I've grouped the rows by this value. This makes it a **lot** easier to match the table with your PR.

image

data/sql/updates/pending_db_world/rev_1620115611701757811.sql
```diff
@@ -0,0 +1,74 @@
...
+
+-- Tin Vein
+UPDATE `gameobject_loot_template` SET `MaxCount`=6 WHERE FIND_IN_SET (`Entry`,'1503,1736,2627,18093') AND `Item` = @COARSESTONE; -- Coarse Stone
+UPDATE `gameobject_loot_template` SET `MinCount`=1 WHERE FIND_IN_SET (`Entry`,'1503,1736,2627,18093') AND `Item` = @COARSESTONE; -- Coarse Stone
+UPDATE `gameobject_loot_template` SET `Chance`=50 WHERE FIND_IN_SET (`Entry`,'1503,1736,2627,18093') AND `Item` = @COARSESTONE; -- Coarse Stone
+UPDATE `gameobject_loot_template` SET `MinCount`=1 WHERE FIND_IN_SET (`Entry`,'1503,1736,2627,18093') AND `Item` = @TINORE; -- Tin Ore
+UPDATE `gameobject_loot_template` SET `MaxCount`=4 WHERE FIND_IN_SET (`Entry`,'1503,1736,2627,18093') AND `Item` = @TINORE; -- Tin Ore
+
+-- Iron Deposit
+UPDATE `gameobject_loot_template` SET `Chance`=50 WHERE Entry = 1505 AND `Item` = @HEAVYSTONE; -- Heavy Stone
```
Kitzunu, line 31: backtick on entry too, all keywords should have them =)

necropola: You mean column names, right?

Kitzunu: yes

Kitzunu: https://www.azerothcore.org/wiki/sql-standards

PR #5611, opened 2021-05-03
feat(Core/Database): port TrinityCore database API

Winfidonarleyan:
## Changes Proposed:
- It's about time, hah
- Port all changes TrinityCore database

## Issues Addressed:
- Closes #4669
- Part #3739 (code only, start support after add boost suport)
- Maybe #4005
- Maybe #3679

## Tests Performed:
- Char (Loading/create/delete)
- Hunter pets (loading, combat, stable master, rename)

## How to Test the Changes:
- Check char create/rename/customize/delete/loading and other
- Check all pets functions (rename/add) and other
- Check character save system

## Known Issues and TODO List:
- [x] License
- [ ] https://github.com/ElunaLuaEngine/Eluna/pull/358
- [ ] https://github.com/azerothcore/mod-transmog/pull/45
- [x] https://github.com/azerothcore/azerothcore-wotlk/pull/5671
- [x] #5950
- [x] Fix `clang-7`
- [ ] change wh
- [x] `.char rename`
- [x] `.char customize`

## Target Branch(es):
- [x] Master
 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #5671, opened 2021-05-06
feat(Cmake/Compiler): deprecated clang6 and gcc 7

Winfidonarleyan:
## Changes Proposed:
- Set `clang` expected version `7.0.0`
- Set `GCC` expected version `8.0.0`

## Issues Addressed:
- Fix for #5611


.github/workflows/core_build.yml
```diff
@@ -29,10 +29,6 @@ jobs:
             compiler: clang9
             modules: without
             extra_logs: false
-          - os: ubuntu-18.04
-            compiler: clang # default in 18.04 is clang 6
-            modules: without
-            extra_logs: false
```
FrancescoBorzi, line 35: please do not just remove the clang6 but **replace** it with clang7 instead

FrancescoBorzi (changes requested):
also don't forget to update the wiki

FrancescoBorzi:
we should also update all modules and replace clang 6 with clang 7

.github/workflows/core_build.yml
Winfidonarleyan: Okay

Winfidonarleyan: Done

FrancescoBorzi:
@Winfidonarleyan also don't forget: https://github.com/azerothcore/azerothcore-wotlk/blob/master/.github/SECURITY.md

Winfidonarleyan:
@FrancescoBorzi done

Winfidonarleyan:
modules, okay

Winfidonarleyan:
Added modules PRs

FrancescoBorzi:
@Winfidonarleyan failing with:

> Unknown compiler clang7

I'm ok if we remove clang6 and 7 from modules and only keep the default ubuntu 20.04 clang version in them. No need to test modules with different clang versions.

Winfidonarleyan:
> @Winfidonarleyan failing with:
> 
> > Unknown compiler clang7
> 
> I'm ok if we remove clang6 and 7 from modules and only keep the default ubuntu 20.04 clang version in them. No need to test modules with different clang versions.

Fixed

FrancescoBorzi (approved):

Kitzunu (approved):

Issue #5661, opened 2021-05-06
quest can't be completed


mpfans:
**CURRENT BEHAVIOUR:**
get quest 12459,When you use item (seeds of nature's write, ID 37887) on `Frost Wyrm` and then kill `weakened realized Frost Wyrm`, the quest cannot be completed

This bug exists in all quests that use items for NPC

**EXPECTED BLIZZLIKE BEHAVIOUR:**
When you use item (seeds of nature's write, ID 37887) on `Frost Wyrm` and then kill `weakened realized Frost Wyrm`, the quest can be completed

**STEPS TO REPRODUCE THE PROBLEM:**
1. .q a 12459
2. use item (seeds of nature's write, ID 37887) on `Frost Wyrm`
3. killed `weakened realized Frost Wyrm`

**EXTRA NOTES:**
You can choose one of the following to complete the quest
1.When you use item (seeds of nature's write, ID 37887) on Frost Wyrm and then run a long way and get out of the fight,Come back and kill it, you can complete the quest
2.Attack Frost Wyrm , and the HP is less than 20%,then use item (seeds of nature's write, ID 37887) on `Frost Wyrm` , and then kill `weakened realized Frost Wyrm`,you can complete the quest

The bug that just appeared from 28ce1e921921a1d04f718056583c9b62dc219edf update to 3f343830fb61aad9ba12c0c908b6b18fc9eafd83 

This bug exists in all quests that use items for NPC

**AC HASH/COMMIT:**
3f343830fb61aad9ba12c0c908b6b18fc9eafd83 

**OPERATING SYSTEM:**
ubuntu 20.04
**MODULES:**
none

Si1ker:
Same behaviour as in #5573

Issue #5573, opened 2021-05-01
[Quest] 'Morbent Fel' Quest Not Completing


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [X] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/494

The quest objective is to kill Morbent Fel (https://wotlkdb.com/?npc=1200) while using a special item, Morbent's Bane. (https://wotlkdb.com/?item=7297)  When using the item, Morbent Fel is changed into "Weakened Morbent Fel".  Killing him does not provide the quest objective kill credit.

Tester's notes - killed him in both weakened and normal states. Neither completed the quest.

![WoWScrnShot_050121_174951](https://user-images.githubusercontent.com/81782124/116776310-d3d24d00-aaa6-11eb-8020-e7f8d864dfa5.jpg)
![WoWScrnShot_050121_175039](https://user-images.githubusercontent.com/81782124/116776311-d9c82e00-aaa6-11eb-8582-59d5988d96b9.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Killing Morbent Fel in his weakened or normal form should provide the quest kill credit.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Have Duskwood quest "Morbent Fel" in your quest log.
2. Equip Morbent's Bane in off-hand.
3. Target Morbent Fel and click Morbent's Bane.
4. Kill Morbent Bane.

##### EXTRA NOTES:

https://classic.wowhead.com/quest=55/morbent-fel

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/91081f4ad8fed98d7bdc9c0ae13d05c86032df9f

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
[...]

PR #5746, opened 2021-05-09
fix(Core/Creature): Fix kill credit for some mobs after UpdateEntry

Raelorasz:
## Changes Proposed:
- Fix an issue that was causing mobs not giving kill credit after call to UpdateEntry
**Note:** this was specific to cases where the original mob has a higher maximum health than the mob it gets updated to. 
In this case  setting m_PlayerDamageReq to the value it was before UpdateEntry prevented players from ever getting kill credit, since this value could be higher than the updated mob's maximum health.
Now, this value will be scaled down based on the maximum health percentage decrease.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5661
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5573
- Closes https://github.com/chromiecraft/chromiecraft/issues/494

## SOURCE:

None, regrettably. TC doesn't have any conditionals for health, which works in this case if `changelevel` is `true` since it also resets m_PlayerDamageReq but not when the maximum health of the updated mob is higher than the former one. 

## Tests Performed:

https://user-images.githubusercontent.com/82320853/117578898-e6750380-b0f0-11eb-8ddd-9dc9cad2d860.mp4

https://user-images.githubusercontent.com/82320853/117578904-ee34a800-b0f0-11eb-8565-3015d08dac96.mp4

https://user-images.githubusercontent.com/82320853/117578911-f68ce300-b0f0-11eb-8829-81eaab065aed.mp4

## How to Test the Changes:

1. `.q a 55`
2. `.go c id 1200`


src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -454,13 +454,18 @@ bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data, bool changele
...
     SelectLevel(changelevel);
     if (previousHealth > 0)
     {
         SetHealth(previousHealth);
-        m_PlayerDamageReq = previousPlayerDamageReq;
+
+        if (previousMaxHealth > GetMaxHealth())
+            m_PlayerDamageReq = (uint32)(previousPlayerDamageReq * GetMaxHealth() / previousMaxHealth);
+        else
+            m_PlayerDamageReq = previousPlayerDamageReq;
```
FrancescoBorzi, line 468: please add brackets

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -454,13 +454,22 @@ bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data, bool changele
...
+    uint32 previousMaxHealth = GetMaxHealth();
     uint32 previousPlayerDamageReq = m_PlayerDamageReq;
 
     SelectLevel(changelevel);
     if (previousHealth > 0)
     {
         SetHealth(previousHealth);
-        m_PlayerDamageReq = previousPlayerDamageReq;
+
+        if (previousMaxHealth > GetMaxHealth())
```
Yehonal, line 465: `if (previousMaxHealth && previousMaxHealth > GetMaxHealth())`

Better to avoid division by zero in some edge-cases

Kitzunu: Can it even be null? Would mean that HP doesnt even exist?

Yehonal: > Can it even be null? Would mean that HP doesnt even exist?

nope, we could even use `> 0`

Si1ker (approved):

Si1ker (approved):

Issue #5709, opened 2021-05-08
(DB/creature): Undermap Moongraze Buck


Si1ker:
Original report at https://github.com/chromiecraft/chromiecraft/issues/568

====================================================

#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [x] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Moongraze Buck (DB GUID 61450, current 191, entry 17201) is walking under the surface of the world, above a cavern.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Should be walking on the surface.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Enter the naga cave south of Bristlelimb Village (Azuremyst Isle at the coordinates (27,76) ).

##### EXTRA NOTES:

![WoWScrnShot_050721_202244](https://user-images.githubusercontent.com/22718174/117519557-0af6a180-af72-11eb-9cb7-359b1ab877a3.jpg)

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/6947789622de45479d2311fd5d1635af3a957b5d

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)
- [lua-CarbonCopy](https://github.com/55Honey/Acore_CarbonCopy)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

Si1ker:
[...]

Issue #5765, opened 2021-05-10
[STORY] Creatures falling into the ground and wrong Z calculation


Yehonal:
Recently several people reported issues with creatures and players falling into the textures

We tried to collect some of them here: https://github.com/azerothcore/azerothcore-wotlk/projects/16 but I'm not sure anymore it's related to the movement/pathfinding since some people also noticed that certain creatures are just spawned under the floor.

In [this PR](https://github.com/pulls?q=is%3Apr+author%3AYehonal+archived%3Afalse+is%3Aclosed) I worked a lot to make the Z calculation working in every case and generating proper normalized paths. One of the dozens of tests I made was with the quest called "Unexpected Results" where some creatures are spawned far away from the target and they had to reach the target that was on a different ground level. Before my PR those creatures were just flying towards the target, while after they calculated the Z coord correctly.

Recently, however, people started to report a lot of issues with the Z coord, and eventually that quest was bugged again and needed to be fixed with "a workaround" I'd say: https://github.com/azerothcore/azerothcore-wotlk/pull/5072

This seems like a generic issue that must be fixed at the core level, we are not supposed to fix all the **coords inside the DB.**

@Footman worked on these PR https://github.com/azerothcore/azerothcore-wotlk/pulls?q=is%3Apr+author%3AFootman+is%3Aclosed

while @UltraNix worked on these ones:
https://github.com/azerothcore/azerothcore-wotlk/pull/4879
https://github.com/azerothcore/azerothcore-wotlk/pull/5114
https://github.com/azerothcore/azerothcore-wotlk/pull/4855

I'm not sure if everyone else also introduced other changes, if so, please report them here.

I think now we should:

1) comment this story with all the evidence of this bug in order to properly test it. Please provide all the instructions!
2) trying to spot which were the PR that caused this issue
3) Provide a PR as soon as possible since this bug is ruining the game experience in the entire world

Unfortunately, I'm not available to work on it at the moment but I can provide all the help needed by chat and here. 

I'd like that everyone contributes to this issue by adding more information

Yehonal:
**By Wotex:**

ultranix made change to Z by removing collisionheight and setting it to 2
i kept on my server collisionheight 


https://github.com/azerothcore/azerothcore-wotlk/pull/4909


i am not sure if its because of this
he takes only 0.5 above ground here
even in old ac it was 2
in tc it is collision height
[...]

PR #5719, opened 2021-05-08
fix(DB/creature): Adjust creature spawn location at Badlands

Hacki95:
## Changes Proposed:
- Update x, y, z and orientation of those 3 NPCS

## Issues Addressed:

- Closes #5696
- Closes https://github.com/chromiecraft/chromiecraft/issues/562

## SOURCE:

- https://github.com/brotalnia/database/blob/master/world_full_12_november_2020.7z

## Tests Performed:

- Applied SQL
- Checked ingame

## How to Test the Changes:

1. Apply SQL
2. .go cr guid

## Known Issues and TODO List:

- /

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #4909, opened 2021-03-19
feat(Core/Maps): Use a fixed offset instead of full collision height …

UltraNix:
…when retrieving floor Z.

Port from TC.
Fixed #4666

## Changes Proposed:
-  
-  

## Issues Addressed:
- Closes #4666.

## Tests Performed:
- Tested with snakes in Wailing Caverns.
-

## Known Issues and TODO List:

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #5775, opened 2021-05-10
fix(Core): Increase Z_OFFSET_FIND_HEIGHT to 2.0

Hacki95:
This addresses the issue of several NPCs falling through the map.

## Changes Proposed:
- Increase Z_OFFSET_FIND_HEIGHT from 0.5 to 2.0. This addresses an issue where after https://github.com/azerothcore/azerothcore-wotlk/pull/4909 certain mobs began falling through the map. 

If we change the constant from 0.5 to 2 many of the mobs who fall through the map right now don't do so anymore:
#5709 doesn't fall through the map anymore.
#5719 GUID 6915 is still falling through the map (possibly wrong values still), the other two are not (with PR).

If we change back the constant to the function everything is the same, those mobs don't fall through the map except 6915 (maybe my fix is just wrong after all).

I tested other values like 1.0 as well and mobs like the Buck fall through the world with this value. 2.0 seems to be the best one for now.

This is most likely not a full fix for this problem as the proper solution seems to be to fix GetCollisionHeight() to not cause problems like #4666.
For now simply increasing the value to 2.0 doesn't trigger #4666 as far as my testing goes.

## Issues Addressed:

- Part: https://github.com/azerothcore/azerothcore-wotlk/issues/5765
- Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/5709
- Closes: https://github.com/chromiecraft/chromiecraft/issues/568

## SOURCE:

- Playing around with different values for the offset.

## Tests Performed:

- Applied the patch.
- Checked a few NPCs

## How to Test the Changes:

1. Apply the patch
2. Check some of the named NPCs in the description
3. Take a look at: https://github.com/azerothcore/azerothcore-wotlk/projects/16 and see if it fixes any problems there.

- Didn't test all issues in: https://github.com/azerothcore/azerothcore-wotlk/projects/16
- This issue needs additional testing.

### DIFF
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -12,7 +12,7 @@
 #include <cassert>
 
 float const GROUND_HEIGHT_TOLERANCE = 0.05f; // Extra tolerance to z position to check if it is in air or on ground.
-constexpr float Z_OFFSET_FIND_HEIGHT = 0.5f;
+constexpr float Z_OFFSET_FIND_HEIGHT = 2.0f;
 
 enum SpellEffIndex
 {


Yehonal:
Locus did the same changes here: https://github.com/azerothcore/azerothcore-wotlk/pull/5775, please find an agreement where to continue working

Hacki95:
Closed in favor of #5775

Issue #5809, opened 2021-05-12
Translation of the menu of the NPC Isirami Fairwind <Innkeeper>.


pangolp:
**CURRENT BEHAVIOUR:**

Currently, in the NPC script the texts are in English, which causes that it does not take the information from the database and does not make the corresponding translations according to the game client used.

**EXPECTED BLIZZLIKE BEHAVIOUR:**

The idea would be to be able to obtain and use the translations, which are already in the database, to make it multilanguage.

**SOURCE:**

**STEPS TO REPRODUCE THE PROBLEM:**

1. Go to the location of the npc
- Alliance: `.go c 110585`. Horde: `.go c 97714`
2. Interact with the npc

**EXTRA NOTES:**

Create the report, then make the pull request and try to fix it.

**AC HASH/COMMIT:**

https://github.com/azerothcore/azerothcore-wotlk/tree/0bc23075bf88e3ce47084a6cb8178e11a924d5a8

**OPERATING SYSTEM:**

- Windows 10

**MODULES:**

**CUSTOMIZATIONS:**

- [mod-npc-promotion](https://github.com/pangolp/mod-npc-promotion)
- [mod-recruit-friend](https://github.com/pangolp/mod-recruit-friend)
- [mod-quest-status](https://github.com/pangolp/mod-quest-status)

![WoWScrnShot_051221_112532](https://user-images.githubusercontent.com/2810187/117992021-fa00b400-b314-11eb-9a3d-0df558720d23.jpg)

![WoWScrnShot_051221_112609](https://user-images.githubusercontent.com/2810187/117992048-fe2cd180-b314-11eb-9f8c-8f2c43ebcd50.jpg)

pangolp:
![WoWScrnShot_051221_124555](https://user-images.githubusercontent.com/2810187/118004939-4e5d6100-b320-11eb-89a1-d1291cbbd0b8.jpg)

![WoWScrnShot_051221_124731](https://user-images.githubusercontent.com/2810187/118004955-53221500-b320-11eb-8b8e-dc4eed3b2508.jpg)

I am having trouble figuring out, which gossip to show you during the Halloween event, since there are so many with the same content.

![trick_or_treat](https://user-images.githubusercontent.com/2810187/118005211-8fee0c00-b320-11eb-98bd-c936c316f8d7.png)

PR #5811, opened 2021-05-12
Fix(Core) Remove hard-coded texts from npc Innkeeper

pangolp:
## Changes Proposed:
-  Use the texts in the database which are already translated.

## Issues Addressed:

- https://github.com/azerothcore/azerothcore-wotlk/issues/5809

## Tests Performed:

- After applying the script, the texts appeared in Spanish (since it is the game client I use) but it should be able to be translated in several languages.

WoWScrnShot_051221_124555

WoWScrnShot_051221_145658

## How to Test the Changes:

- However, it can be tested without the event. It is preferable to activate it: `event start 12`
- If it is an alliance: `.go c 110585`
- If horde: `.go c 97714`

- [x] The menu to be shown in the Halloween event is yet to be defined correctly, since there were many options to use and I don't know which one is the right one.
- [x] I am not sure if I should add a GOSSIP_MENU_OPTION or if it is not necessary. The menu, I assumed it had to be added, but the options, I don't know if they are necessary.

### DIFF
--- a/src/server/scripts/World/npc_innkeeper.cpp
+++ b/src/server/scripts/World/npc_innkeeper.cpp
@@ -12,7 +12,8 @@ enum eTrickOrTreatSpells
     SPELL_TREAT                 = 24715,
     SPELL_TRICKED_OR_TREATED    = 24755,
     HALLOWEEN_EVENTID           = 12,
-    GOSSIP_MENU                 = 9733
+    GOSSIP_MENU                 = 9733,
+    GOSSIP_MENU_EVENT           = 342
 };
 
 class npc_innkeeper : public CreatureScript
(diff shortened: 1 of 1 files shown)


src/server/scripts/World/npc_innkeeper.cpp
```diff
@@ -22,16 +23,24 @@ class npc_innkeeper : public CreatureScript
     bool OnGossipHello(Player* player, Creature* creature) override
     {
         if (IsEventActive(HALLOWEEN_EVENTID) && !player->HasAura(SPELL_TRICKED_OR_TREATED))
-            AddGossipItemFor(player, GOSSIP_ICON_CHAT, "Trick or Treat!", GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + HALLOWEEN_EVENTID);
+        {
+            AddGossipItemFor(player, 342, 0, GOSSIP_SENDER_MAIN, GOSSIP_ACTION_INFO_DEF + HALLOWEEN_EVENTID);
```
FrancescoBorzi, line 27: what is `342` and shall we move it to a constant?

pangolp: The problem is that there are many records for the same phrase, so I don't know which one to use exactly.
That's why I didn't create the constant.

trick_or_treat

pangolp: It seems to work with 342. But since there were so many options, I wasn't sure which one to use.

WoWScrnShot_051221_150827

FrancescoBorzi (approved):

arielcami:
Tested, and worked.

WoWScrnShot_051321_195444

pangolp:
Thanks @FrancescoBorzi and @arielcami

FrancescoBorzi:
thanks @arielcami for testing this!

good job @pangolp

PR #5831, opened 2021-05-13
Fix(Core/Event) Removing hardcore texts in Love is in the Air

pangolp:
## Changes Proposed:
-  Remove texts found in the script
- Use the translations found in DB

## Issues Addressed:

- Closes

## Tests Performed:

- Windows 10 

## How to Test the Changes:

1. Activate the event `event start 8`.
2. Visit the NPCs of the scripts

### DIFF
--- a/src/server/scripts/Events/love_in_air.cpp
+++ b/src/server/scripts/Events/love_in_air.cpp
@@ -52,9 +52,13 @@ public:
             {
                 lock = 0;
                 if (urand(0, 1))
-                    me->MonsterSay("Time is money, friend. Go go go!", LANG_UNIVERSAL, who->ToPlayer());
+                {
+                    me->AI()->Talk(0, who->ToPlayer());
+                }
                 else
-                    me->MonsterSay("That crate won't deliver itself, friend. Get a move on!", LANG_UNIVERSAL, who->ToPlayer());
+                {
+                    me->AI()->Talk(0, who->ToPlayer());
+                }
 
                 if (who->ToPlayer()->GetTeamId() == TEAM_ALLIANCE)
                 {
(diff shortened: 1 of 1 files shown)


src/server/scripts/Events/love_in_air.cpp
```diff
@@ -52,9 +52,13 @@ class npc_love_in_air_supply_sentry : public CreatureScript
             {
                 lock = 0;
                 if (urand(0, 1))
-                    me->MonsterSay("Time is money, friend. Go go go!", LANG_UNIVERSAL, who->ToPlayer());
+                {
+                    me->AI()->Talk(0, who->ToPlayer());
```
Kitzunu, line 56: Use constants not magic numbers

pangolp: > Use constants not magic numbers

Ok, now I remove the magic numbers.
Thank you.

Kitzunu (commented):
Except that the enums could be a bit better, and that some of them have more texts in creature_text than what you use...

You do not need to add nullptr to say. It is enough to use Talk(abc);

pangolp:
> Except that the enums could be a bit better, and that some of them have more texts in creature_text than what you use...
> 
> You do not need to add nullptr to say. It is enough to use Talk(abc);

That was a doubt I had. Thanks, I'll fix it now.
Regarding the enum, if you have any suggestions, I accept them.

pangolp:
I am here to learn, so I have no problem in modifying things that you consider wrong. Feel free to send me the changes you want but as long as you explain me the reasons so I can understand them and learn from them, as in the case of the `nullptr`, which was a doubt I really had, if it was necessary or not.

Helias (approved):

FrancescoBorzi:
@Helias don't forget to apply workflow labels after approving a PR

Issue #5635, opened 2021-05-04
Attacks Against Sitting Enemy Players Do Not Crit


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/524

When you use a melee ability on an enemy player that is sitting it does not guarantee a critical strike.

Tester's notes - had two of my characters duel, one sitting down after every second attack while the other one kept melee attacking.

Auto attacks showed no variation between sitting and standing:
![WoWScrnShot_050421_224918](https://user-images.githubusercontent.com/81782124/117009965-7569d080-ad2b-11eb-8d6c-7c5baa3896aa.jpg)

The report specified melee abilities, which means auto-attack might not count. so tried with Ravage:
![WoWScrnShot_050421_230103](https://user-images.githubusercontent.com/81782124/117011264-cfb76100-ad2c-11eb-9283-34f38812e06d.jpg)

And then Mangle:
![WoWScrnShot_050421_230132](https://user-images.githubusercontent.com/81782124/117011298-d80f9c00-ad2c-11eb-91d2-b2ca9790f50b.jpg)

As you can see, there's no difference between any of the attacks, although every second one was against a sitting target. (Also, feral druids can apparently use Prowl while in combat in duels now, which may need another ticket.)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

When you use a melee ability (something like backstab or shred) or an melee autoattack on a target that is sitting down it should give that one melee strike a 100% crit chance.

From https://wowpedia.fandom.com/wiki/Sit?oldid=1656296:
"Melee attacks against a sitting target automatically hit and cause double damage. The first damage taken while sitting will cause a target to immediately stand. The double damage registers as a critical hit in the combat log, but does not trigger abilities as a normal crit would (such as the Fury Warrior talent: Enrage)"

##### STEPS TO REPRODUCE THE PROBLEM:

best way to reproduce would be dueling i guess
1. Duel another player
2. Tell him to sit down. 
3. Use any melee ability on that target
4. proceed to not get a crit with that ability

[...]

PR #5872, opened 2021-05-15
fix(CORE/UNIT) Fix criticals not working against sitting players.

Maxproz:
**Description:**
The issue is that we set the enemy player into the stand state when we put them into combat.
As a result, when we check to see if they are in the sit state later, its always false.
So the only way you were getting a critical was with the base crit chance.

With this fix if an enemy player is sitting. And you swing at them you will always crit.
This also removes the chance to miss or dodge when sitting down, which I believe is blizzlike, and saves me from passing around a ton of parameters, or having to cache the state inside of the class.

https://github.com/azerothcore/azerothcore-wotlk/issues/5635

## Changes Proposed:
-  Cache the sit state of the enemy player at the start of the attack state sequence.
- Pass the cached sit state into the CalculateMeleeDamage function later on.

## Issues Addressed:

- Closes 
- https://github.com/azerothcore/azerothcore-wotlk/issues/5635

## Tests Performed:

- Builds without errors

## How to Test the Changes:

1. Create 2 Accounts
2. Create 2 Nightelves 
3. Run outside starting zone
4. Duel eachother
5. Have 1 player sit 
6. Auto attack sitting player

An argument could be made about whether it should be possible to miss an attack on sitting player.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -1297,7 +1297,7 @@ void Unit::DealSpellDamage(SpellNonMeleeDamage* damageInfo, bool durabilityLoss)
 }
 
 // TODO for melee need create structure as in
-void Unit::CalculateMeleeDamage(Unit* victim, uint32 damage, CalcDamageInfo* damageInfo, WeaponAttackType attackType)
+void Unit::CalculateMeleeDamage(Unit* victim, uint32 damage, CalcDamageInfo* damageInfo, WeaponAttackType attackType, const bool sittingVictim)
 {
     damageInfo->attacker         = this;
     damageInfo->target           = victim;
@@ -1365,8 +1365,15 @@ void Unit::CalculateMeleeDamage(Unit* victim, uint32 damage, CalcDamageInfo* dam
     else
         damageInfo->damage = damage;
 
-    damageInfo->hitOutCome = RollMeleeOutcomeAgainst(damageInfo->target, damageInfo->attackType);
-
+    if (sittingVictim)
+    {
+        damageInfo->hitOutCome = MELEE_HIT_CRIT;
+    }
+    else
+    {
+        damageInfo->hitOutCome = RollMeleeOutcomeAgainst(damageInfo->target, damageInfo->attackType);
+    }
+    
     switch (damageInfo->hitOutCome)
     {
         case MELEE_HIT_EVADE:
@@ -2198,6 +2205,9 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
     if ((attType == BASE_ATTACK || attType == OFF_ATTACK) && !IsWithinLOSInMap(victim))
         return;
 
+    // CombatStart puts the target into stand state, so we need to cache sit state here to know if we should crit later
+    const bool sittingVictim = victim->GetTypeId() == TYPEID_PLAYER && victim->IsSitState() ? true : false;
+
     CombatStart(victim);
     RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_MELEE_ATTACK);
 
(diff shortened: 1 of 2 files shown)


Kitzunu:
Should it crit for all emote states? Or only sit?

Branel:
Sit would be the only case, I don't think there is a case to be made about not missing if the target is sitting down, as it might sounds reasonable it probably didnt work like that.

Maxproz:
I can easily include missing. Just let me know what production wants and I will modify as needed.

Branel:
Something that needs to be noted is that after TBC it was patched that yes the player would take a crit while taking damage but there would be no proccs from the crit, i.e a fury warrior would not get enraged from taking a crit while sitting down. Is this adressed in your PR?

Maxproz:
Rage generation runs through separate functions and was not part of the code I looked at.

Maxproz:
If the rage thing is still a problem you can assign it to me and I will fix it.

Branel:
Laying down is another case where the player should be affected by a critical strike according to [gamepedia](https://wowwiki-archive.fandom.com/wiki/Critical_strike), might need some more research but it makes sense.

If you can adress the enrage proccs and maybe even there are some similar spells that might be affected that would be great.

Good job so far though!

Maxproz:
Added laying down crits.
Added missing possibility.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1367,6 +1367,10 @@ void Unit::CalculateMeleeDamage(Unit* victim, uint32 damage, CalcDamageInfo* dam
 
     damageInfo->hitOutCome = RollMeleeOutcomeAgainst(damageInfo->target, damageInfo->attackType);
 
+    // If the victim was a sitting player and we didn't roll a miss, then crit.
+    if (sittingVictim && damageInfo->hitOutCome != MELEE_HIT_MISS)
+        damageInfo->hitOutCome = MELEE_HIT_CRIT;
```
FrancescoBorzi, line 1372: brackets please

Kitzunu:
```
Codestyle check script:

  Checking RegEx: '\t'
  Checking RegEx: '[[:blank:]]$'
src/server/game/Entities/Unit/Unit.cpp:1376:    

Remove whitespace at the end of the lines above
```

Maxproz:
> ```
> Codestyle check script:
> 
>   Checking RegEx: '\t'
>   Checking RegEx: '[[:blank:]]$'
> src/server/game/Entities/Unit/Unit.cpp:1376:    
> 
> Remove whitespace at the end of the lines above
> ```

No idea, lol. 

My environment has no whitespace there. 
image

Branel:
The whitespace before switch is what causes the codestyle error.

Branel (commented):
The whitespace before switch is what causes the codestyle error.

Maxproz:
> The whitespace before switch is what causes the codestyle error.

I removed the empty line and it passed, but I feel like this is an error in the code style checker. 
Looks and reads much better with it.

Azcobu:
Tested:
- base case, melee vs sitting target: crits as expected.
- melee vs target lying down: crits as expected.

However, have run into a problem with **stealthed** attacks vs sitting targets. Tested on my druid, and Ravage attacks from stealth vs a sitting target do not automatically crit as they should. Here's the combat log - each Ravage is from stealth against a sitting target:
WoWScrnShot_051821_214827

I also tried stealth attacks vs a lying down target, and they did not automatically crit either.

On the whole I think you've done a great job here, there's just this one odd corner case to address.

Maxproz:
@FrancescoBorzi We should merge this fix and write a new bug for the stealth target not critting.

It will require deeper investigation and I don't have time until the weekend.

FrancescoBorzi (approved):

FrancescoBorzi:
good job @Maxproz and thanks @Azcobu and @Branel for testing/reviewing this one!

Azcobu:
Want me to create a new ticket for the remaining issue, or is that unnecessary?

Branel:
We merge this and create two new reports that details the no enrage procc and the stealth issue separately

PR #5877, opened 2021-05-15
fix(CI): replace clang-9 with clang-8

FrancescoBorzi:
Context: https://stackoverflow.com/questions/67537679/clang9-steady-clocknow-causing-timeout-the-monitored-command-dumped-core

I've tried to reproduce the issue locally with ubuntu 20.04 and clang-9, but it worked fine for me. So it's something wrong with some dependency in our CI. I've opened a report here: https://github.com/github/hub/issues/2765

So let's try replacing clang-9 with clang-8 and see what happens.


locus313:
@FrancescoBorzi I don't think that is the correct repo to report the issue to github. I think its actually this repo it should be reported at https://github.com/actions/virtual-environments.

apps/ci/ci-conf.sh
```diff
@@ -59,6 +59,12 @@ case $COMPILER in
     echo "CCOMPILERCXX=\"clang++-9\"" >> ./conf/config.sh
     ;;
 
+  "clang8" )
+    time sudo apt-get install -y clang-8
+    echo "CCOMPILERC=\"clang-8\"" >> ./conf/config.sh
+    echo "CCOMPILERCXX=\"clang++-8\"" >> ./conf/config.sh
+    ;;
+
```
locus313, line 67: we probably should put clang8 before clang9

apps/ci/ci-install.sh
```diff
@@ -64,6 +64,12 @@ case $COMPILER in
     echo "CCOMPILERCXX=\"clang++-9\"" >> ./conf/config.sh
     ;;
 
+  "clang8" )
+    time sudo apt-get install -y clang-8
+    echo "CCOMPILERC=\"clang-8\"" >> ./conf/config.sh
+    echo "CCOMPILERCXX=\"clang++-8\"" >> ./conf/config.sh
+    ;;
+
```
locus313, line 72: same here

apps/ci/ci-conf.sh
FrancescoBorzi: ah definitely yes, my bad

locus313 (approved):

locus313:
@FrancescoBorzi recently found this:
https://github.com/actions/virtual-environments/issues/3235

Issue #5882, opened 2021-05-16
Dodged and parried strikes do not raise weapon skill


Stump876:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

While I believe the chance to receive a skill up is accurate, dodge and parry do not give skill points. Additionally attacks made on critters do not give points either.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

[https://classic.wowhead.com/guides/classic-wow-weapon-skills](https://classic.wowhead.com/guides/classic-wow-weapon-skills)
[https://wowwiki-archive.fandom.com/wiki/Weapon_skill](https://wowwiki-archive.fandom.com/wiki/Weapon_skill)
 "Each attack that is not evaded has a chance to increase weapon skill" So any miss, dodge, parry, or hit should give a chance at skill up, initially it being at or near 100% dropping to about 50% when reaching 10% of cap. This should include also attacks on critters (rats, rabbits, snakes, etc...) as this is still an attack on something killable.

##### STEPS TO REPRODUCE THE PROBLEM:

1. .setskill of a weapon to 1 
2. Attack a mob roughly the same level of below with that weapon to level it, each swing (block, miss, dodge, parry) should count as 1 skill point initially and get less frequent the closer to the skill cap of the level you get.
3. Pay closer attention to the dodge and parry swings. Testing I did not see those raise a skill point.

##### EXTRA NOTES:
--

Weapon skills were removed in Cataclysm. Any info found would be from 3.3.5 or earlier

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/91081f4ad8fed98d7bdc9c0ae13d05c86032df9f

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #5895, opened 2021-05-16
FIX(CORE/UNIT) Dodges, parries and critters not increasing weapon skills

Maxproz:
**Description:**
This fixes not getting skill ups from attacking an enemy and them parrying your attack or dodging it.

This also fixes attacking a critter and not getting weapon skill ups.

Previously Dodge and Parry were only giving a chance to increase the victims DEFENSE skill.
I changed it so it will proc on victim and attacker. for weapon and defense skills

Also critters were being checked to not give weapon skill ups on purpose but I removed the check.

## Changes Proposed:
-  Make it so dodges and parries have a chance to increase weapon skill not just defense skill
-  Make it so critters have a chance to increase weapon skill.
as far as I know there is no way to abuse this because there is not mass critters in one location.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5882

## How to Test the Changes:

1. Make a nightelf 
2. .character level 20
3. .go 100 100 100
4. .cheat god
5. Find a spider
6. Auto attack it 
7. Observe skill ups when it parries or dodges
8. Find a critter
9. attack it
10. Observe skill up

Not really I analyzed the entire skill ups algorithm and it follows blizzlike behaviour, and works well enough to not modify at the moment.


FrancescoBorzi (approved):

FrancescoBorzi:
@Maxproz thanks for your contribution, have you tested this in game?

Maxproz:
@FrancescoBorzi This compiles, builds, and was working on my machine.

Ready for test.

necropola (commented):
Doesn't that exclude dodge and parry from increasing your character's defense skill? Is this intended? And if it's intended, shouldn't you update the comment, too?

Also it should be (no linebreak
- Closes #5882

so the corresponding issue gets actually linked.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -15504,13 +15504,13 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
...
-            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST))
+            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST | PROC_EX_PARRY | PROC_EX_DODGE))
             {
-                if (target->GetTypeId() != TYPEID_PLAYER && !target->IsCritter())
+                if (target->GetTypeId() != TYPEID_PLAYER)
                     ToPlayer()->UpdateCombatSkills(target, attType, isVictim);
             }
             // Update defence if player is victim and parry/dodge/block
-            else if (isVictim && procExtra & (PROC_EX_DODGE | PROC_EX_PARRY | PROC_EX_BLOCK))
+            else if (isVictim && procExtra & (PROC_EX_BLOCK))
```
necropola, line 15513: Doesn't that exclude dodge and parry from increasing your character's defense skill? Is this intended? And if it's intended, shouldn't you update the comment, too?

Maxproz: > Doesn't that exclude dodge and parry from increasing your character's defense skill? Is this intended? And if it's intended, shouldn't you update the comment, too?
> 
> Or maybe it doesnn't exclude it from increasing defense, because isVictim can be true in the if clause, too. But why not increase weapons skill on target blocks, too (when parry is counted).

@necropola 
I also noticed I needed to update the comments.

For the first part. 
I had the same exact thought - (I was worried if that I changed it like this then there would be limited ways to increase defense skill) But then I considered the fact that we still are getting defense skill ups when we get hit with basic auto attacks. 

This flows into your second part.
isVictim flowing into defense and offense in the first if clause. 
An argument could definitely be made for increasing weapon skill on block also.
The only reason I didn't it because I don't remember it being blizzlike and it doesn't make much sense to me. 

For example: In terms of increasing proficiency with a "Weapon"
When you parry someone you used that weapon in the action. 
When you block someone you only used the shield in the action, the weapon wasn't used at all. 
That's how I thought about it anyway.

Maxproz:
Fixed comments that I forgot to update.

src/server/game/Entities/Unit/Unit.cpp
necropola: > When you block someone you only used the shield in the action, the weapon wasn't used at all.

By that argument you would need to exclude **Dodge**, too.

Anyway, do we have any source on how it worked during WotLK. I did a bit of research, but couldn't find a single line about what counted and what not. What also worries me a bit is that the existing code explicitly excludes skill ups from beating on critters (not that this would make much of a difference), but removing this without evidence doesn't feel good.

Don't get me wrong. I'm all in for making weapon and defense skillups less of an annoyance, e. g. I would love the  `AlwaysMaxSkillForLevel` setting to be activated in **ChromieCraft's** `worldserver.conf`. 😼

Maxproz: Yeah the explicit no critter skillups is slightly worrisome. 

But if there was a bad bug directly related to it. I feel like someone would have left a comment.

poryagholami1374:
https://user-images.githubusercontent.com/80007195/119781262-0e7da880-bee0-11eb-88e4-6e1ff8ef712a.mp4


https://user-images.githubusercontent.com/80007195/119781271-11789900-bee0-11eb-92ea-3c473f4be053.mp4

there is no problem
And I do not approve of this post

tkn963:
Works as intended during my tests 👍

Sedatyf:
Thank you again @tkn963! 🚀

necropola (commented):
@Sedatyf
- We do not know (have a source) how it's supposed to work.
  - Should a dodged/parried attack really increase the weapon skill of the attacker but a blocked attack should not? This does not make any sense at all.
  - Should a player really be able to train weapon skills on critters? 
- We do not know exactly when and with which parameter values this function is called. How do we know which in-game cases we have actually covered?
- We cannot easily test this in game, because of the (low) chance of actually getting a skill point. Without having some debug logging right before invoking `UpdateCombatSkills()` I wouldn't dare comfirming that it works as it should. 
- We deviate further from the TC code base

Instead we should first apply/port https://github.com/TrinityCore/TrinityCore/commit/e641d0c7d71bdb21e443bf52bf508b1581e07839 and then revisit the issue when we have a reliable historic source to justify a change like this.

Also this PR is now severely out of sync with master, i. e. it is missing the changes from PR #5961 by @Kitzunu.

Sedatyf:
Hey @necropola, thank you for the update. I admit, I didn't read through all PR as it was late. Thank you for the heads up! I fixed the status of it

Kitzunu:
Mmm Ariel proc system will be nice. But it will be hell of a job to introduce as it will also introduce a million bugs that has to be fixed individually, but it is a good system =)

Kitzunu:
I dont know what I feel about this change. To me it does not make sense to update weapon skill on doges and parries as you could see it like you are not "getting the effect" of the weapon when you are parried or dodged, just like a miss. Also on critters I guess that could be logical? 

But I can't find anything on the subject at all

necropola:
At least the existing implementation is consistent, i. e. neither dodged nor parried nor blocked attacks can increase the weapon skill of the attacker. And as long as we cannot find any source stating otherwise (couldn't find anything either), I'd just leave it as is.

claudiodfc:
> Mmm Ariel proc system will be nice. But it will be hell of a job to introduce as it will also introduce a million bugs that has to be fixed individually, but it is a good system =)

Can we not port stuff from Developers that don't work on Azerothcore? Otherwise no one will maintain that code specially when we have no idea what it does... Specially from TC....

Kitzunu:
I will close this PR. It doesn't make any sense.

Main reason being to that a mob has a lower chance to evade/parry/dodge whatever when your skill has a higher value, so it makes 0 to no sense to be able to gain weapon skill here.

syssneck:
Hey everyone, here is what I found and remember:
Weapon skill should have the chance to increase on any attack, with the only exception if the target is evading and/or immune to damage (an evading target is always immune to damage).
[https://classic.wowhead.com/guides/classic-wow-weapon-skills](url)

Critter topic:
I also remember that I was able to increase my weapon skill on critters. This was on TBC 2.4.3 back in the day, but the only change to weapon skills in the patch notes up to 3.3.5 I found was this one:
Patch 3.0.8 (2009-01-20): You can no longer skill up weapons on mobs that are immune to damage.

I also found a comment from someone playing on patch 3.0.3 stating:
By visas on 2008/12/08 at 2:40 AM (Patch 3.0.3)
"another option, for alliance anyhow, is to kill rats in the deeprun tram, ironforge side. Critters still give skillups, and those rats respawn after about 5 seconds each."
[https://www.wowhead.com/forums/topic/you-can-train-weapon-skill-on-training-dummy-58091](url)

I know this is not quite set in stone evidence, just what I found and remember. I approve of these changes.

Kitzunu:
Critters I can see sure, but an parry/dodge == No damage dealt, so why would it give skillup, especially when the skillup is used to prevent parries and doges?

necropola:
Well, misses count, too. And you need to count them or you'll have a very hard time getting started.

Kitzunu:
fair point

syssneck:
Just imagine a lvl 80 character that always played with a sword and now gets a mace. According to the "no weapon skill on parry and dodge" model, he would need to fight low level mobs to increase his weapon skill. The problem is, he oneshots every low level mob even with an awful mace and therefore he would need to fight hundreds of low level mobs (steadily increasing in level) to increases his mace weapon skill.

T1ti:
Misses aren't dodges, the weapon skill difference will mostly give misses.

Btw what makes "sense" doesn't matter, it's just about being blizzlike, simply test it on classic

Kitzunu:
Well classic <> Wrath 3.3.5a :shrug: I mean there is literally 0 documentation on how it was. The only way would to have old sniffs, which we don't have. so I honestly don't care if we merge it or not

T1ti:
Afaik the only change wrath did is not allowing to level vs dmg immune mobs

I'm pretty sure dodges and paries should increase it, si will test later

tkn963:
Well, the Wiki has information on this. It might not be equal to firm evidence of how it works but it's there. Weapon skills were removed in Cataclysm so it's Wrath and down.

`Each attack which is not evaded has a chance to increase weapon skill`
```
The base chance for your attack to be parried by the mob:
If the difference between the mob's Defense Skill and your Weapon Skill is less than or equal to 10, your base parry rate is: 5% + (Defense Skill - Weapon Skill)*.1%.
If the difference between the mob's Defense Skill and your Weapon Skill is greater than 10, your base parry rate will be much higher. Exact figures are not known, but base parry rates of Boss mobs (effectively level 73 when attacked by level 70 players) are estimated to be between 10-15%, and may vary from one Boss to the next.

The base chance for your attack to be blocked by the mob is probably the same as the base dodge rate, i.e.: 5% + (Defense Skill - Weapon Skill)*.1%
```

More info: https://wowwiki-archive.fandom.com/wiki/Weapon_skill

tkn963:
I also remember farming weapon skill points on the rats in the deeprun tram back in WotLK, so critters should give skill ups. I have no link to show this is a fact except my own memory so I'm hoping someone can provide such source.

necropola:
Note that the PR in it's current state still excludes blocks from granting skillups, while it allows parries and dodges.

T1ti:
I think we should also add blocks

necropola:
Yep, either leave it as is or go the whole way. But I see it has been merged already ... 😿

Issue #5111, opened 2021-04-03
Warlock's Imp keeps Phase shift in combat


T1ti:
## CURRENT BEHAVIOUR:
-Imp will not lose the phase shift buff when it attacks.
-It will never be attackable by creature even if it has max threat, the creature will always go for the master because phase shift makes the imp immune.
![image](https://user-images.githubusercontent.com/40864460/113466058-10277480-9439-11eb-8d13-3c62c981957f.png)

## EXPECTED BLIZZLIKE BEHAVIOUR:
-Imp should lose phase shift when it attacks and enters combat  
![image](https://user-images.githubusercontent.com/40864460/113466086-4238d680-9439-11eb-9f79-2893bd6bc8bf.png)

## STEPS TO REPRODUCE THE PROBLEM:

1. Summon your imp as a warlock, activate phase shift. 
2. make it attack a creature and see if it loses the buff and if creature can attack it

## AC HASH/COMMIT:
85d21335da425de8f3a1488ad75753e2cfa1529e

markvetx:
Can confirm this on ecfeecb

Kitzunu:
I can see if I can do anything about it

Kitzunu:
Fixed https://github.com/azerothcore/azerothcore-wotlk/pull/5960

PR #5960, opened 2021-05-19
fix(Script/Creature): Script warlock imp

Kitzunu:
## Changes Proposed:
-  Remove Phase Shift when imp goes in combat

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5111

## Tests Performed:

- Builds
- Tested in-game

## How to Test the Changes:

**Run CMAKE**

1. Go in combat with imp


src/server/scripts/Pet/pet_warlock.cpp
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (C) 2016+     AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license, you may redistribute it and/or modify it under version 2 of the License, or (at your option), any later version.
```
FrancescoBorzi, line 2: please use AGPL for new files

Azcobu:
Tested, working perfectly.
- Imp casts Phase Shift when out of combat.
- When entering combat, Imp drops Phase Shift and starts attacking.
- When the imp attacks a creature, that creature is able to attack the imp in return.

Creatures would sometimes still ignore the imp's threat and attack the warlock, even when the imp had much higher threat. So my guess is that imp or pet threat mechanics aren't 100% right yet, but that is outside the scope of this ticket.

Kitzunu:
Threat is whack allover the place. There are some already open issues about it but it is not directly an issue with the imp but the threat system overall.

FrancescoBorzi (approved):

PR #6006, opened 2021-05-22
feat(DB/Version): Add date on update

Kitzunu:
## Changes Proposed:
-  This is good and can open for opportunities where we can output current DB version with commands ingame etc.
- Introduce date column
-  Don't update and check column name
- Start to check latest date column instead
- insert date when importing DB

## How to Test the Changes:

1. run import.sh and db_assambler

WHEN MERGED!!!!

We need to MANUALLY DELETE THIS FOR THE DB FILES:

```sql
SELECT date INTO @COLEXISTS
FROM version_db_db
ORDER BY date DESC LIMIT 1;
IF @COLEXISTS <> '"$oldVer"' THEN LEAVE proc; END IF;
```

when the CI has run the script for ONLY THESE FILES IN THE PR. Otherwise importing with db_assambler will break.

It works as it should after that and we don't need to touch it again.

## Tests Performed:

- Tested locally with import.sh and db_assambler.sh 

| # | C |
| :- | :- | 
| 1 | File that imported these changes (The ones you find in this PR) |
| 2 | A new file generated and imported after these changes |

WORLD:
image

CHARACTERS:
image

AUTH:
image

### DIFF
--- a/apps/db_pendings/import.sh
+++ b/apps/db_pendings/import.sh
@@ -43,7 +43,6 @@ function import() {
             newVer=$dateToday"_"$cnt
 
             startTransaction="START TRANSACTION;";
-            updHeader="ALTER TABLE version_db_"$db" CHANGE COLUMN "$oldVer" "$newVer" bit;";
             endTransaction="COMMIT;";
 
             newFile="$updPath/"$dateToday"_"$cnt".sql"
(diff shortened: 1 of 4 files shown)


FrancescoBorzi:
great idea @Kitzunu , even though I wonder if we could just read it from the column name instead

Kitzunu:
I tried that, but didnt manage to pass it through the core

Kitzunu:
TIP for merging:

After merge this, then in 1 PR you make the removals I described above + that you drop the column that has the name of a date. It is then easy to have the column name and needs less effort :D But that requires back to back merging

apps/db_pendings/import.sh
```diff
@@ -65,14 +64,13 @@ function import() {
...
-                echo "WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'version_db_"$db"' AND COLUMN_NAME = '"$oldVer"';" >> "$newFile";
-                echo "IF @COLEXISTS = 0 THEN LEAVE proc; END IF;" >> "$newFile";
+                echo "SELECT date INTO @COLEXISTS" >> "$newFile";
+                echo "FROM version_db_"$db  >> "$newFile";
+                echo "ORDER BY date DESC LIMIT 1;"  >> "$newFile";
+                echo "IF @COLEXISTS <> '"$oldVer"' THEN LEAVE proc; END IF;" >> "$newFile";
             fi
 
             echo "$startTransaction" >> "$newFile";
-            echo "$updHeader" >> "$newFile";
```
Kitzunu, line 75: If you prefer I can revert this and only have the update part for now... for maybe faster merge :shrug:

locus313 (approved):

PR #5427, opened 2021-04-24
fix(CI): disable mac PCH to allow more caching

Rochet2:
## Changes Proposed:
-  Disable PCH in mac CI to allow ccache to cache more files
-  Decrease mac CI  ccache max size to 1G since that fits a bit more than one build

## Issues Addressed:
- It **seems to me** that regardless of the fixed cache saving the mac builds are taking ~45 minutes even on `master` branch as seen here: 
image

- With the proposed changes the time is hopefully closer to 20 minute mark as seen here: 
image

## Tests Performed:

- 
- 

## Known Issues and TODO List:

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #6008, opened 2021-05-22
feat(CI): separated build pch and no-pch

Winfidonarleyan:
## Changes Proposed:
- Add support pch for all `Ubuntu` build
- Make separated no-pch build with (Ubuntu latest and clang)
- Reduced compile time for all build (Less workload for GitHub actions)


apps/ci/ci-conf-no-pch.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
FrancescoBorzi, line 1: do we really need to duplicate this??

Winfidonarleyan: I don't know how to do otherwise :/

locus313:
Enabling pch is going to make the ci builds take much longer.  We should not enable it for the whole core-build or docker build. Between @Yehonal  and me with purposely disable pch to speed up the build times. If we do enable a pch build it should be in a seprate workflow that does not run automatically. (like the macos, windows, docker workflows)

Winfidonarleyan:
How can it take longer to compile a core with pch?

`succeeded 6 hours ago in 53m 40s`
https://github.com/azerothcore/azerothcore-wotlk/pull/6008/checks?check_run_id=2649285455

locus313:
Take a look at the info in this pr:

https://github.com/azerothcore/azerothcore-wotlk/pull/5427

Build times where greatly  reduced when disabling pch.

When I did some research on the issue it seems to be mainly due to how ccache handles the matching.

.github/workflows/core_build_no_pch.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.compiler }}
+    env:
+      COMPILER: ${{ matrix.compiler }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: /home/runner/.ccache
```
locus313, line 25:
```suggestion
          path: var/ccache
```

locus313: @Yehonal built in support for ccache directly into the scripts so to properly save the cache you need to change the path to this.

Winfidonarleyan:
No need for this repo

Issue #6044, opened 2021-05-25
Crash in WorldSession.cpp:203


FrancescoBorzi:
```
Thread 11 "worldserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ff1795fc700 (LWP 14368)]
0x0000000001291662 in acore::Assert (file=0x13a300c "/root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp", line=203, function=0x13a305d "SendPacket", debugInfo=..., message=0x13a3068 "packet->GetOpcode() != NULL_OPCODE") at /root/env/chromiecraft/azerothcore-wotlk/src/common/Debugging/Errors.cpp:64
64	        Crash(formattedMessage.c_str());
#0  0x0000000001291662 in acore::Assert (file=0x13a300c "/root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp", line=203, function=0x13a305d "SendPacket", debugInfo=..., message=0x13a3068 "packet->GetOpcode() != NULL_OPCODE") at /root/env/chromiecraft/azerothcore-wotlk/src/common/Debugging/Errors.cpp:64
#1  0x00000000008e4c7b in WorldSession::SendPacket (this=<optimized out>, packet=<optimized out>) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:203
#2  0x00000000011140d8 in spell_pvp_trinket_wotf_shared_cd::spell_pvp_trinket_wotf_shared_cd_SpellScript::HandleScript (this=0x7ff0e3c37a80) at /root/env/chromiecraft/azerothcore-wotlk/src/server/scripts/Spells/spell_generic.cpp:2335
#3  0x00000000009b9499 in Spell::CallScriptAfterCastHandlers (this=0x7ff16925ba80) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/Spell.cpp:7877
#4  Spell::_cast (this=0x7ff16925ba80, skipCheck=<optimized out>) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/Spell.cpp:3664
#5  0x00000000009b4e49 in Spell::cast (this=0x7ff16925ba80, skipCheck=255) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/Spell.cpp:3462
[...]

PR #6045, opened 2021-05-25
fix(Core/WorldSession): prevent crash in SendPacket

FrancescoBorzi:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6044

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -200,7 +200,11 @@ ObjectGuid::LowType WorldSession::GetGuidLow() const
 /// Send a packet to the client
 void WorldSession::SendPacket(WorldPacket const* packet)
 {
-    ASSERT(packet->GetOpcode() != NULL_OPCODE);
+    if (packet->GetOpcode() == NULL_OPCODE)
+    {
+        LOG_ERROR("server", "WorldSession::SendPacket(packet) called, but packet->GetOpcode() was NULL_OPCODE");
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


Yehonal:
Not crashing the server could result in corrupted data later though, so maybe we should investigate on why it's happening

Kitzunu:
I agree, something else must be causing this as TC does not have this issue afaik

FrancescoBorzi:
> Not crashing the server could result in corrupted data later though, so maybe we should investigate on why it's happening

I agree we should investigate why it happens but it looks like something that was happening before, because that `ASSERT` has recently been added, while the code causing it to trigger was already there.

With my PR we no longer have the ASSERT but at least we have an error log, so still better than it was before. Plus the players will not crash now and then...

FrancescoBorzi:
@Yehonal isn't there a way to make the LOG ERROR give you the same information as the ASSERT but **WITHOUT** crashing the server ??

Yehonal:
> > Not crashing the server could result in corrupted data later though, so maybe we should investigate on why it's happening
> 
> I agree we should investigate why it happens but it looks like something that was happening before, because that `ASSERT` has recently been added, while the code causing it to trigger was already there.
> 
> With my PR we no longer have the ASSERT but at least we have an error log, so still better than it was before. Plus the players will not crash now and then...

corrupted data and exploit possibilities are not the best things to have, and you can lost such errors within your files. Such kind of issue should be like send directly via email to the Administrator of the server otherwise you will never have any clue that it happened until you do not check the logs

Issue #3679, opened 2020-11-11
server is frozen [$150]


mpfans:
logs:
```
[Warning] Transaction aborted. 4 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 2450 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 7175 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 4527 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 4267 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 2450 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 6976 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 6297 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 499 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
SQL(p): UPDATE character_pet SET slot = 100 WHERE owner = 499 AND slot = 0
 [ERROR]: [1213] Deadlock found when trying to get lock; try restarting transaction
[Warning] Transaction aborted. 3 queries not executed.
[...]

PR #6069, opened 2021-05-27
fix(Core/Shared/DB): transactions not being recommitted on dead-lock error

UltraNix:
## Changes Proposed:
Fix transactions not being recommitted on dead-lock error (error code 1213) because of calling http://dev.mysql.com/doc/refman/5.0/en/mysql-errno.html after sending the ROLLBACK command. This way the returned error code was related to the ROLLBACK command, not the failed transaction.

## Issues Addressed:

- Updates #3679.

## Tests Performed:

-  Not tested 

## How to Test the Changes:

Follow #3679.

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -7,6 +7,8 @@
 #include "DatabaseWorkerPool.h"
 #include "DatabaseEnv.h"
 
+#include <mysqld_error.h>
+
 #define MIN_MYSQL_SERVER_VERSION 50700u
 #define MIN_MYSQL_CLIENT_VERSION 50700u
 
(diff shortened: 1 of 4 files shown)


Viste:
i.e. treating the effects of deadlock rather than its source is an OK solution? @FrancescoBorzi that waht about i talk 
Yeah it's cool we can recommit deadlocked transaction, but why we get it? what code make it possible I think we need to ask ourselves that question in the first place.
@UltraNix good job, no questions to you. 
sorry for write it here, but I couldn't find another suitable place.

UltraNix:
@Viste Original source: https://github.com/TrinityCore/TrinityCore/issues/14024

Viste:
same crutch as here yes cool

FrancescoBorzi (approved):

FrancescoBorzi:
>  I think we need to ask ourselves that question in the first place.

@Viste I agree, the underlying problem should be fixed as well. But this is a step forward

mpfans:
image

image

UltraNix:
These changes should be merge anyway, even if the problem still persists.

FrancescoBorzi:
> These changes should be merge anyway, even if the problem still persists.

I agree, unless they make something worse but so far no one has reported any regression

Issue #6101, opened 2021-05-29
Scarlet Gauntlets/Wristguards Widely Dropped Outside SM


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [ ] Alliance
- [x] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [x] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/725

The Scarlet Gauntlets (ID 10331) dropped of some pigs in RFD. https://imgur.com/a/NVOI9Ep

Tester's notes: Can confirm the Gauntlets appear to be much more widely distributed than they should be:
```
mysql> SELECT COUNT(*)
    -> FROM `creature_template` ct 
    -> JOIN `creature_loot_template` clt ON ct.lootid = clt.entry
    -> JOIN `reference_loot_template` rlt ON rlt.entry = clt.Item
    -> JOIN `item_template` it ON it.entry = rlt.Item
    -> WHERE it.name = "Scarlet Gauntlets";
+----------+
| COUNT(*) |
+----------+
|      440 |
+----------+
1 row in set (0.09 sec)
```
The Gauntlets appear twice in the reference_loot_template tables, in IDs 24056 and 526790. They are a direct drop from the Centurions (2% chance) so should probably be removed from the RLTs.

Update: checked other items in the Scarlet set, and the Scarlet Wristguards (ID 10333)(https://classic.wowhead.com/item=10333/scarlet-wristguards) are even more widely distributed, being dropped from 550 mobs. They are in RLT 24054.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

The Scarlet Gauntlets should only drop in Scarlet Monastery. In fact, only from Scarlet Centurions: https://classic.wowhead.com/item=10331/scarlet-gauntlets

##### SOURCE:

https://www.wowhead.com/item=10331/scarlet-gauntlets#comments 

Last comment.

##### STEPS TO REPRODUCE THE PROBLEM:

Run SQL command above.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/3716ddf3e4e1ef138df334d01980fea019272001

##### OPERATING SYSTEM:

[...]

PR #6102, opened 2021-05-29
fix(DB/loot): Removes Scarlet Set items from rand_loot_temp tables

Azcobu:
## Changes Proposed:
- Remove Scarlet Gauntlets (ID 10331) from reference_loot_templates 24056 and 526790.
- Remove Scarlet Wristguards (ID 10333) from reference_loot_template 24054.

Note that both items continue to be dropped by mobs within SM, as those drops are handled directly and not by loot templates.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/725
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6101

## SOURCE:

links here show that the only mobs that should drop these items are in SM:
https://classic.wowhead.com/item=10331/scarlet-gauntlets
https://classic.wowhead.com/item=10333/scarlet-wristguards

## Tests Performed:

- Ran SQL, checked in Keira, verified items were now gone from referenced RLTs.
- Verified items still existed in normal drop tables for SM mobs.

## How to Test the Changes:

```
SELECT COUNT(*)
FROM `creature_template` ct 
JOIN `creature_loot_template` clt ON ct.lootid = clt.entry
JOIN `reference_loot_template` rlt ON rlt.entry = clt.Item
JOIN `item_template` it ON it.entry = rlt.Item
WHERE it.name = "Scarlet Gauntlets";
```
This should now return 0 rows. Run again with Wristguards in place of Gauntlets.


Azcobu:
Updated the SQL as the automatic review did not like the syntax of the original version.

FrancescoBorzi (approved):

PR #6163, opened 2021-05-31
feat: add argument to .character rename to add name to reserved_name

r-o-b-o-t-o:
## Changes Proposed:
- Add an optional argument `[reserveName]` to the `.character rename` command. If this argument is `1`, then the current player's name is added to the list of reserved names, so that it becomes unavailable.

## Tests Performed:

- Applied SQL and C++ fixes, tested on Windows 10 server and clients

## How to Test the Changes:

1. Select a character in-game and do `.char rename 1` or `.char rename Yournamehere 1`
2. Try to create a character with that name, verify that it says "This name is unavailable"
3. Play around with the command, make sure everything else still works as intended


src/server/scripts/Commands/cs_character.cpp
```diff
@@ -302,10 +302,20 @@ class character_commandscript : public CommandScript
     //rename characters
     static bool HandleCharacterRenameCommand(ChatHandler* handler, char const* args)
     {
+        char* nameStr = strtok((char*)args, " ");
+        char* reserveNameStr = strtok(nullptr, " ");
+
+        if (reserveNameStr == nullptr && nameStr != nullptr && atoi(nameStr) == 1)
```
Kitzunu, line 308:
```suggestion
        if (!reserveNameStr && nameStr && atoi(nameStr) == 1)
```

mpfans:
There is a bug,if ```reserved_name``` have the name of the current character,write failure and error will be reported

You can't guarantee that all player names are in ```reserved_ name``` is not created until it is set, because many players' names are created before they are ```reserved_ name``` table is forbidden

r-o-b-o-t-o:
> There is a bug,if `reserved_name` have the name of the current character,write failure and error will be reported
> 
> You can't guarantee that all player names are in `reserved_ name` is not created until it is set, because many players' names are created before they are `reserved_ name` table is forbidden

I tried to do `.char rename 1` on a character whose name was already in the `reserved_name` table and it didn't give any error. What commit are you on? I probably fixed this problem in c5f49e4.

tkn963:
Using .character rename 1 and changing the name of the character frees up the name and it can be used again but the row still exists in the database. Is this intended? Other than that I can't trigger any issues

tkn963:
My apologies but I encountered a problem. When trying to perform a rename of a character with a name that already exists in the database prints this error to the console:
SQL(p): INSERT INTO reserved_name (name) VALUES ('Revision')
 [ERROR]: [1062] Duplicate entry 'Revision' for key 'PRIMARY'

r-o-b-o-t-o:
@tkn963 

> My apologies but I encountered a problem. When trying to perform a rename of a character with a name that already exists in the database prints this error to the console:
> SQL(p): INSERT INTO reserved_name (name) VALUES ('Revision')
> [ERROR]: [1062] Duplicate entry 'Revision' for key 'PRIMARY'

This probably happened because you inserted the row in the `reserved_name` table manually while the server was already running. When the server starts, it loads this table into memory, and I use this in-memory copy to check for a duplicate before inserting, so this should not happen in a normal scenario. But just to be sure, I changed my query to an `INSERT IGNORE` instead.

> Using .character rename 1 and changing the name of the character frees up the name and it can be used again but the row still exists in the database.

GM accounts are not affected by server-side name restrictions, do you remember if you were using one? If so, that's why it let you re-use the blacklisted name.

Thanks for your feedback!

tkn963:
> 
> 
> @tkn963
> 
> > My apologies but I encountered a problem. When trying to perform a rename of a character with a name that already exists in the database prints this error to the console:
> > SQL(p): INSERT INTO reserved_name (name) VALUES ('Revision')
> > [ERROR]: [1062] Duplicate entry 'Revision' for key 'PRIMARY'
> 
> This probably happened because you inserted the row in the `reserved_name` table manually while the server was already running. When the server starts, it loads this table into memory, and I use this in-memory copy to check for a duplicate before inserting, so this should not happen in a normal scenario. But just to be sure, I changed my query to an `INSERT IGNORE` instead.
> 
> > Using .character rename 1 and changing the name of the character frees up the name and it can be used again but the row still exists in the database.
> 
> GM accounts are not affected by server-side name restrictions, do you remember if you were using one? If so, that's why it let you re-use the blacklisted name.
> 
> Thanks for your feedback!

Both your statements are correct. My bad, I'll re-test it with proper expectations!

tkn963:
I'm having issues using this feature. Using the command, character rename <name> 1, to trigger a rename on a character on a non-GM account - it doesn't add the name to reserved_name at all.

I have checked about 10 times and I'm using the pr-6163 branch which is correct

tkn963:
The issues I was having are no longer there. No proper reason for these issues to even occur, but it works like it should now! 👍

r-o-b-o-t-o:
> The issues I was having are no longer there. No proper reason for these issues to even occur, but it works like it should now! 👍

Great, thanks for the tests!

FrancescoBorzi (approved):

PR #6179, opened 2021-06-01
Don't use bounding radius for aggro distance checks.

Krejza9:
## Changes Proposed:
-  This change is necessary for the correct calculation functioning of the future patch for aggro range
-  

## SOURCE:
https://github.com/vmangos/core/commit/af745365e9bd7f0963734510452c0ac957fb1424

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

## Known Issues and TODO List:

- [ ] Some files need to be traced and added

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #6214, opened 2021-06-03
feat(Core): port aggro distance from vMaNGOS (read detection_range values from creature_template)

### LINKED ISSUE #5455: New database column settings for creatures Aggro.(and rework the leash/aggro systems) [$30]
## CURRENT BEHAVIOUR:

There is not a single settting for creatures aggro at all, they are all hardcoded and use the same generic valeus in the core.

## EXPECTED BLIZZLIKE BEHAVIOUR:

This is based on a screenshot of Blizzard's creature tool editor from late 2009 I was able to find 
image

This confirms that we need settings for : 

- **Detection** : This is the range at which creatures detect and see players

- **Pursuit** : The most hard to guess. I believe this is the initial aggro radius around the creature's original position, in which he will never reset

- **Call for help** : This is the radius of the creature's "call for help" mechanic when it's aggroed, all creatures in 50 yards around it will aggro with it.

- **Timeout** : This is a timer, it sets in seconds how long will it take for the creature to reset (outside the pursuit radius). If set to 0 creature will never stop chasing, it's set to 0 for all creatures in instances.

- **Leash** : I believe this is for the hardset leashes distance for some creatures, this is not widely used, it's mostly used for world bosses so you can't kite them to stormwind, like Kazzak. If 0 then the creature deosn't use this
They do set some premade templates that they reuse, I believe there just are a few generic templates for dungeons mobs, dungeon bosses, world bosses, raid bosses etc...

**### NOTE : Currently the AC leash/Aggro system works very differently, I had already made an issue about how to make it blizzlike, this picture further proofs I was right and Blizz use a timeout system for the leash.**
https://github.com/azerothcore/azerothcore-wotlk/issues/5116
image

## SOURCE:

image
and the research I did in https://github.com/azerothcore/azerothcore-wotlk/issues/5116

## STEPS TO REPRODUCE THE PROBLEM:

1. 
2. 
3. 

## AC HASH/COMMIT:

d11d021f38e0e0db27e7c74da801402e7e5f17d9

## CUSTOMIZATIONS:

<bountysource-plugin>

---
There is a **[$30 open bounty](https://www.bountysource.com/issues/98060852-new-database-column-settings-for-creatures-aggro-and-rework-the-leash-aggro-systems?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Krejza9:
## Changes Proposed:
-  read detection_range values from creature_template
-  contains DB records for unique creatures (for WoW classic only)

## Issues Addressed:

- Closes #5455

## SOURCE:

https://github.com/vmangos/core/commit/9f50e90ac9ed7db233a7343b324c559e021bb5b3
https://github.com/vmangos/core/commit/af745365e9bd7f0963734510452c0ac957fb1424

## Tests Performed:

- can be tested in the game
- 

## How to Test the Changes:

1. Add SQL update
2. "The aggro radius of a mob having the same level as the player is the new column detection_range from the creature_template table"
2. Check if it fits creature range from DB (The formula for the calculation is:  same level as the player is detection_range column )
for each level up minus one yard
for each level down plus one yard


Krejza9:
no longer WIP

FrancescoBorzi:
@azerothcore/developers @azerothcore/testers please check this

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2986,16 +2988,19 @@ float Creature::GetAggroRange(Unit const* target) const
...
+    uint32 playerlevel  = getLevelForTarget(target);
+    int32 levelDiff = int32(creaturelevel) - int32(playerlevel);
 
     // The maximum Aggro Radius is capped at 45 yards (25 level difference)
     if (levelDiff < -25)
         levelDiff = -25;
 
     // The base aggro radius for mob of same level
-    float aggroRadius = 20.0f;
+    float const detectionRange = GetDetectionRange();
```
claudiodfc, line 3000: Isn't this variable kinda useless here? You could simply make:
`float aggroRadius = GetDetectionRange();`

I'm reviewing from smartphone so apologies if this is used below.

Krejza9: Theoretically yes , because in the DB there is a default value  20 on detection_range column, but it is always better to have a default value also within the cpp code, even because of a bad value in the DB or an overload via scriptdev. 
```C++ 
float const detectionRange = GetDetectionRange(); > float GetDetectionRange() const { return m_detectionDistance; } > m_detectionDistance(20.0f)
```

claudiodfc: Maybe I explained myself wrong, I was referring to the variable `detectionRange` not accomplishing anything here. Instead of making this variable, why not simple give `aggroRadius` the default value from function?

Yehonal:
This is a very interesting one! GJ. Could someone (including the author) provide a video test for it?

Also, are you willing to import the pool system from vmangos as well? that's another important feature very easy to import but that would make the difference

Sedatyf:
> This is a very interesting one! GJ. Could someone (including the author) provide a video test for it?
> 
> Also, are you willing to import the pool system from vmangos as well? that's another important feature very easy to import but that would make the difference

I have to find time to test it. @Krejza9 already explained in DM to me this PR. I'll make a video when I'll test it for sure

src/server/game/Entities/Creature/Creature.cpp
Krejza9: I see what you mean, but I followed the AC regulations. You'll find a lot of these functions in AC. I won't argue with the work of the AC developers, I guess there's a reason :)

Krejza9:
> 
> 
> This is a very interesting one! GJ. Could someone (including the author) provide a video test for it?
> 
> Also, are you willing to import the pool system from vmangos as well? that's another important feature very easy to import but that would make the difference
When we merge PR then I'll look into pool system.

src/server/game/Entities/Creature/Creature.h
```diff
@@ -222,6 +223,7 @@ class Creature : public Unit, public GridObject<Creature>, public MovableMapObje
     bool CanStartAttack(Unit const* u) const;
     float GetAggroRange(Unit const* target) const;
     float GetAttackDistance(Unit const* player) const;
+    float GetDetectionRange() const { return m_detectionDistance; }
```
FrancescoBorzi, line 226: please consider `[[nodiscard]]` where applicable

Krejza9: You are right, the return function is not to be discarded

Krejza9: fixed

data/sql/updates/pending_db_world/rev_1622194253034538200.sql
```diff
@@ -0,0 +1,7052 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1622194253034538200');
+
+ALTER TABLE `creature_template` ADD COLUMN `detection_range` FLOAT NOT NULL DEFAULT 20 AFTER `speed_run`;
+
+UPDATE creature_template SET detection_range ='1' WHERE name ="Baron Rivendare";
```
Kitzunu, line 5: https://www.azerothcore.org/wiki/sql-standards lot of things to correct :D

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2986,16 +2988,19 @@ float Creature::GetAggroRange(Unit const* target) const
...
+    float aggroRadius = detectionRange;
+    if (aggroRadius < 1)
+        return 0.0f;
```
Kitzunu, line 3003: https://www.azerothcore.org/wiki/cpp-code-standards#brackets

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2986,16 +2988,19 @@ float Creature::GetAggroRange(Unit const* target) const
     if (aggroRate == 0)
         return 0.0f;
 
-    uint32 targetLevel = target->getLevelForTarget(this);
-    uint32 myLevel = getLevelForTarget(target);
-    int32 levelDiff = int32(targetLevel) - int32(myLevel);
+    uint32 creaturelevel = target->getLevelForTarget(this);
+    uint32 playerlevel  = getLevelForTarget(target);
```
Kitzunu, line 2992: creatureLevel playerLevel

Kitzunu (changes requested):

data/sql/updates/pending_db_world/rev_1622194253034538200.sql
Krejza9: I'll redo it on the entry NPCs, but I won't be responsible if you change them :-D

Kitzunu: Just dont forget to use backticks and not to use single quotes on integers

src/server/game/Entities/Creature/Creature.cpp
Krejza9: This struck me as odd when you don't use them in the file
image

Kitzunu:
```cpp
if (aggroRadius &lt; 1)
{
    return 0.0f;
}

File: src/server/game/Entities/Creature/Creature.cpp (hunk 5 above)
Comment (Krejza9) [reply src/server/game/Entities/Creature/Creature.cpp:2992]: I missed this. I'll fix it.

File: src/server/game/Entities/Creature/Creature.cpp (hunk 4 above)
Comment (Krejza9) [reply src/server/game/Entities/Creature/Creature.cpp:3003]: do you really want to have &lt; instead of < ?
that's not a very good solid

Comment (Kitzunu) [reply src/server/game/Entities/Creature/Creature.cpp:3003]: No, it was weird copy shit on chrome. I want the brackets

File: src/server/game/Entities/Creature/Creature.cpp (hunk 5 above)
Comment (Krejza9) [reply src/server/game/Entities/Creature/Creature.cpp]: done

File: src/server/game/Entities/Creature/Creature.cpp (hunk 4 above)
Comment (Krejza9) [reply src/server/game/Entities/Creature/Creature.cpp]: done

**Comment by Helias** [comment]
I shorted **A LOT** the size of the SQL file.

anyways good job @Krejza9

File: src/server/game/Entities/Creature/Creature.cpp [hunk 6]
```diff
@@ -2991,17 +2993,21 @@ float Creature::GetAggroRange(Unit const* target) const
     if (aggroRate == 0)
         return 0.0f;
 
-    uint32 targetLevel = target->getLevelForTarget(this);
-    uint32 myLevel = getLevelForTarget(target);
-    int32 levelDiff = int32(targetLevel) - int32(myLevel);
+    uint32 creatureLevel = target->getLevelForTarget(this);
+    uint32 playerLevel  = getLevelForTarget(target);
```
Helias, line 2997: if the function getLeveForTarget has the right return type you can use `auto` instead of `uint32` in these two lines

src/server/game/Entities/Creature/Creature.cpp
Helias: I agree with @claudiodfc, I think you can even remove the detectionRange variable and assign to aggroRadius directly `getDetectionRange()`.

you can even use `auto` in this case instead of `float`

Krejza9: fixed

src/server/game/Entities/Creature/Creature.cpp
Krejza9: fixed

Helias (approved):

T1ti:
looking at the code a bit, I doubt the formula is 1yard per level especially since it's not static anymore, I would guess it's percent based, is the code different in vmangos?
Also personal feeling but I always felt like yellow mobs had almsot identical radius but when a mob is red, at like +4/5 levels the aggro radius difference steeply increases

`m_detectionDistance(20.0f)`
means 20 is the default radius if the creature has no data in detection_range right ? 
18 seems to be the new default it's what msot entries use in db, why not use that ?

-we will have to increase the aggro radius of raid bosses, it's definitely higher than 18

The only value I have is the Wotlk Onyxia which is 80, note that this is a template that blizzard reuses for many raid mobs so most raid bosses probably use that ?
image

I would guess raid trash also has higher radius ?

all of that will need some testing from classic servers

T1ti:
image
from a classic video, this is the range at which vanilla onyxia detects a player

Krejza9:
Hi, thank you for the message, it is a 1:1 port from Vmangos, in view of the fact that Vmangos is a WoW classic so only mobs from the classic are modified, BC and WOTLK remain at default values of 20
About your picture:
It was an example , which blizzard presented not necessarily showing the actual values. As for the calculation of the pull aggro, nobody currently knows the exact formula (whether it is a percentage calculation, distance calculation or other) it is just a speculation of how we think it works
The Vmangos code is the closest thing we have at the moment.

Krejza9:
Unfortunately, there is currently no way to determine the exact distance between the player and the mob.
The only possibility is to inject a client, but I don't want to be banned after 16 years of playing :-D

Krejza9:
Hi @Yehonal , I made a test video
https://youtu.be/LlYY2IOeDBM

T1ti:
It looks significantly bigger than on classic imo

T1ti:
Talked to brotalnia (main dev of vmangos)
Apaprently it's wrong and something else needs to be changed
image
"and where here do u see a change to the implementation of IsWithinDistance"
"you are checking distance with combat reach of both objects substracted"
image
image

T1ti:
image
bonus here's some sniffed data of aggror anges from classic

Here the code he linked 
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Creature/Creature.cpp#L1713
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Object/Object.h#L879
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Object/Object.cpp#L1072
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Object/Object.h#L792

Krejza9:
I talked to the brotalnia everything is fine , he didn't know about the second PR #6179 which needs to be deployed at the same time as this one.
My mistake , I should have put both PRs in one to avoid wrong conclusions

FrancescoBorzi:
@azerothcore/developers @azerothcore/testers can anyone please check this?

Krejza9:
@FrancescoBorzi it looks like no one is interested :(

Sedatyf:
It's not that I'm not interested, but we had a lot of things to test which was higher in priority compare to this one. I'll take care of it as soon as I can

Krejza9:
@Sedatyf I have already created a test video as requested.
Unfortunately I can't become an official tester due to ChromieCraft's policy

FrancescoBorzi:
@Krejza9 I've just noticed the test video so I think we can merge this. Just a few questions:

- does it need to be merged together with https://github.com/azerothcore/azerothcore-wotlk/pull/6179 ?
- have you tested the aggro in general, to check if this PR does not introduce any issue? I mean by playing around a bit

Krejza9:
Yes, it needs to be together
I tested Aggro, I didn't find any problem or bug, but of course it's not in my power to test everything in the game.

FrancescoBorzi:
@Krejza9 I think it's handier if you merge the other branch into this one, so it's easier for people to just check it out and test. Can you please do it?

Krejza9:
done

anguaive:
I did some half assed testing on this.

When level diff between player and creature is high enough to decrease the detection range below 5, the minimum value of 5 is used, as expected.
Tested the same interaction with the same mob that was tested in the video. Works as expected.
Went throught Maraudon on a level 49 character and noticed nothing out of the ordinary.

I however noticed that I could sometimes nudge myself a little bit into the detection range without aggroing the creature. 

WoWScrnShot_070621_201307

Here the aggro range is supposed to be `max(5, 18 - abs(60 - 14)) = 5`, but the current distance is 4.32. I wiggled between 4.32 and 5 for a while and eventually the creature noticed me. This might be intended behavior for all I know, and I don't think it's related to the changes this PR introduces, I just thought I would mention it.

Krejza9:
@anguaive I noticed that too, it looks like tick rate problems or miscalculation of vectors in AC.
The same problem also occurs without this PR

FrancescoBorzi:
thanks @anguaive  , I'll flag this as [To Be Merged] soon

FrancescoBorzi (approved):

PR #6350, opened 2021-06-14
feat(doc): changelog system

Yehonal:
## Changes Proposed:
-  new changelog system, instructions are available in this WIKI PR to merge at the same time: https://github.com/azerothcore/wiki/pull/543

This system provides rules and automatizes (via a Deno script) the creation of a changelog file to help developers to adapt their code and know what is new with every (pre)release

### DIFF
--- a/doc/changelog/previous-versions/v3.x.md
+++ b/doc/changelog/previous-versions/v3.x.md
@@ -1,0 +1,1271 @@
+## We suggest that you always use the latest version of our master branch.
+
+Do **NOT** download the source code from the release, always use latest master version by following [the official guide](https://www.azerothcore.org/wiki/Installation).
+
+## Upgrade instructions
+
+For server administrators: instructions about how to upgrade existing servers are available [here](http://www.azerothcore.org/wiki/Upgrade-from-pre-3.0.0-to-latest-master).
+
+## Thanks
+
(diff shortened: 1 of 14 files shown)


doc/changelog/previous-versions/v3.x.md
```diff
@@ -0,0 +1,1271 @@
...
+fix(CI): break on errors in CI scripts
+Core/Eluna: Hotfix call sEluna->OnGameEventStart method arg register
+feat(CI): improve Travis configuration
+Core/Hooks: Adds hooks for start and stop game events
+Core/Hooks: Add new Player hooks
+chore(docs): use new wiki
+fix(CI): Travis triggers the AC bot only once
+fix(DB): sql update chain for all DBs
+DB: Fixing some issues
+```
```
FrancescoBorzi, line 1271: not sure if we should put the full changelog here, as we are deciding to only add breaking changes in the future changelogs... so it might look inconsistent.

Yehonal: We need them as an example of where to put the changelogs of the previous releases and how they should be named.
However, how to write a changelog implies that you follow the wiki guidelines and it's written also into the file that is automatically generated. People should use it as they already do with the SQL in-pending files. You first have to read the wiki otherwise you do not know how to do that just by copying other sql.

FrancescoBorzi (approved):

Yehonal:
Please wait before merge: I want to add a way to write the current commit hash within the section of every new version, in this way people can easily find the commit related to the changes.

Issue #6321, opened 2021-06-11
Theramore Combat Dummies Chase/Attack Players


JamesGent:
**CURRENT BEHAVIOUR:**

![WoWScrnShot_061121_220215](https://user-images.githubusercontent.com/61194190/121748564-1f920080-cb01-11eb-920d-3bd3c14770c0.jpg)

Theramore Combat Dummies Chase/Attack Players which is HILARIOUS but not Blizzlike. They are stationary Target Dummies and should follow Target Dummy behaviour.

**EXPECTED BLIZZLIKE BEHAVIOUR:**

They are stationary Target Dummies and should follow Target Dummy behaviour.

**SOURCE:**

https://wowpedia.fandom.com/wiki/Theramore_Combat_Dummy

This source says they can be attacked but do not take damage and function as a dummy. Given WOTLK changed a lot of Dummy behaviour and added practice DPS dummies to all cities, this is the correct behaviour.

https://classic.wowhead.com/npc=4952/theramore-combat-dummy

However, this Classic source say they cannot be attacked. WOTLK > Classic sources priority for this core.

**STEPS TO REPRODUCE THE PROBLEM:**

1. Go to Theramore.
2. Taunt the dummies with a nondamaging attack and watch them chase you until leashed.
3. Cackle.

**EXTRA NOTES:**

**AC HASH/COMMIT:**
 

AzerothCore rev. cb1aaec 2021-06-10 11:12:23 +0000 (master branch) (Unix, RelWithDebInfo, Static)

**OPERATING SYSTEM:**

**MODULES:**

**CUSTOMIZATIONS:**

Azcobu:
Rather regret fixing this. Perhaps we can bring aggressive target dummies back in an event some day.

![WoWScrnShot_061421_174642](https://user-images.githubusercontent.com/81782124/121893989-c377e800-cd5d-11eb-9d14-60aa1a77a6fb.jpg)

PR #6351, opened 2021-06-14
fix(DB/Creature): Pacifies Theramore Combat Dummies

Azcobu:
## Changes Proposed:
-  Makes the Theramore Combat Dummies into inert targets again, and stops them chasing people.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6321

## SOURCE:

This is not a controversial change.

## Tests Performed:

- Keira3 generated the SQL. Executed with no problems.
- Then checked in-game, dummies behaved properly.

## How to Test the Changes:

1. Attack a Theramore dummy. If it chases you or moves at all, this change did not work.

N/A.


data/sql/updates/pending_db_world/rev_1623675094427536149.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1623675094427536149');
+
+-- Makes Theramore Combat Dummies boring again.
+UPDATE `creature_template` SET `mechanic_immune_mask` = 32, `flags_extra` = 262144, `ScriptName` = 'npc_training_dummy' WHERE (`entry` = 4952);
```
Kitzunu, line 4: https://www.azerothcore.org/wiki/sql-standards#flags-amp-bits

Azcobu: Cheers, rewritten to meet standard.

Azcobu: Actually, just tested this and the combat dummies are dying if you hit them hard enough, so needs more work still.

Azcobu: Was working with refreshed DB for other testing, so the above changes weren't included. Dummies running the new changes don't die, but they do take half-health damage, which the dummies I copied their settings from (Expert's Training Dummy in Org) do not, so still missing a difference somewhere.

FrancescoBorzi (approved):

PR #6419, opened 2021-06-18
fix(Ci): limit concurrency group to pr's

locus313:
## Changes Proposed:
-  only apply the concurrency limits to pr's


.github/workflows/core_build.yml
```diff
@@ -5,8 +5,8 @@ on:
       - 'master'
   pull_request:
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+concurrency:
```
Kitzunu, line 8:
```suggestion
concurrency:
```

locus313: @Kitzunu You were too quick :). I got this updated now.

FrancescoBorzi (approved):

Issue #6452, opened 2021-06-20
Crash in SpellMgr.h


FrancescoBorzi:
```
Thread 16 "worldserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f9c1f5f7700 (LWP 11295)]
0x00000000012d3252 in Acore::Assert (file=0x13bb99e "/root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/SpellMgr.h", line=690, function=0x13bb9e9 "AssertSpellInfo", debugInfo=..., message=0x13bb9f9 "spellId < GetSpellInfoStoreSize()") at /root/env/chromiecraft/azerothcore-wotlk/src/common/Debugging/Errors.cpp:64
64	        Crash(formattedMessage.c_str());
#0  0x00000000012d3252 in Acore::Assert (file=0x13bb99e "/root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/SpellMgr.h", line=690, function=0x13bb9e9 "AssertSpellInfo", debugInfo=..., message=0x13bb9f9 "spellId < GetSpellInfoStoreSize()") at /root/env/chromiecraft/azerothcore-wotlk/src/common/Debugging/Errors.cpp:64
#1  0x00000000005c2317 in SpellMgr::AssertSpellInfo (this=<optimized out>, spellId=<optimized out>) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Spells/SpellMgr.h:690
#2  0x000000000054bb0f in Player::RemoveArenaSpellCooldowns (this=0x7f9bb6e1c6c0, removeActivePetCooldowns=true) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Entities/Player/Player.cpp:4418
#3  0x000000000054abe2 in Player::TeleportTo (this=0x7f9bb6e1c6c0, mapid=559, x=4027.6001, y=2972.78003, z=12.0723, orientation=0, options=0, target=<optimized out>) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Entities/Player/Player.cpp:2437
#4  0x00000000007c4313 in WorldSession::HandleBattleFieldPortOpcode (this=0x7f9be2718200, recvData=...) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Handlers/BattleGroundHandler.cpp:469
[...]

PR #6454, opened 2021-06-20
fix(Core): check spell info and prevent crash

FrancescoBorzi:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6452
- Allow better investigations of the underlying problem using logs

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4417,7 +4417,7 @@ void Player::RemoveArenaSpellCooldowns(bool removeActivePetCooldowns)
     {
         next = itr;
         ++next;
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -4417,7 +4417,7 @@ void Player::RemoveArenaSpellCooldowns(bool removeActivePetCooldowns)
     {
         next = itr;
         ++next;
-        SpellInfo const* spellInfo = sSpellMgr->AssertSpellInfo(itr->first);
```
Kitzunu, line 4420: You will be back son <3

Kitzunu (approved):

Issue #6242, opened 2021-06-05
Fresh DB installation results in a mix of collations error


Dranni21312:
**CURRENT BEHAVIOUR:**
After following the installation guide, the initial attempt at import-all results in an error when processing AUTH updates:

Version 1.9.1 is already installed
1) all: Assemble all                                4) customs: Assemble only customs                   7) import-updates: Assemble & Import only updates
2) bases: Assemble only bases                       5) import-all: Assemble & Import all                8) import-customs:  Assemble & Import only customs
3) updates: Assemble only updates                   6) import-bases: Assemble & Import only bases       9) quit: Exit from this menu
[Please enter your choice]: 5
=====     DB ASSEMBLER MENU     =====
===== STARTING ASSEMBLY PROCESS =====
Generating /home/osboxes/azerothcore/env/dist/sql/auth_base ...
Searching on /home/osboxes/azerothcore/data/sql/base/db_auth/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/auth_updates ...
Searching on /home/osboxes/azerothcore/data/sql/updates/db_auth/ ...
Searching on /home/osboxes/azerothcore/data/sql/updates/pending_db_auth/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/auth_custom ...
Searching on /home/osboxes/azerothcore/data/sql/custom/db_auth/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/characters_base ...
Searching on /home/osboxes/azerothcore/data/sql/base/db_characters ...
Generating /home/osboxes/azerothcore/env/dist/sql/characters_updates ...
Searching on /home/osboxes/azerothcore/data/sql/updates/db_characters/ ...
Searching on /home/osboxes/azerothcore/data/sql/updates/pending_db_characters/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/characters_custom ...
Searching on /home/osboxes/azerothcore/data/sql/custom/db_characters/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/world_base ...
Searching on /home/osboxes/azerothcore/data/sql/base/db_world/ ...
Generating /home/osboxes/azerothcore/env/dist/sql/world_updates ...
Searching on /home/osboxes/azerothcore/data/sql/updates/db_world/ ...
Searching on /home/osboxes/azerothcore/data/sql/updates/pending_db_world/ ...
[...]

PR #6465, opened 2021-06-20
fix(DB/Auth): Collation error

Kitzunu:
## Changes Proposed:
-  Fix collation error 

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6242

## Tests Performed:

- Tested in discord


claudiodfc (commented):
Shouldn't it be utf8_bin?

Kitzunu:
Several reasons not to use utf8 under any circumstances:

1. utf8 is just a synonym for utf8mb3 which is outdated and will be deprecated in later versions of MySQL 8.0
1. utf8mb4 is an updated and supported version and therefore we should use it, always.
1. Error currently is that it is trying to compare utf8mb4_0900_ai_ci and utf8mb4_general_ci, default for utf8mb4 is utf8mb4_general_ci.

claudiodfc:
Got it, thanks

Helias (approved):

PR #6466, opened 2021-06-20
chore(Core/Spell): SpellId's should always be uint32

Kitzunu:
## Changes Proposed:
-  Spell's are not larger than uint32. It is good measure to keep it uint32 when dealing with all spells.


claudiodfc (approved):
Looks good 👍🏼

src/server/scripts/Spells/spell_item.cpp
```diff
@@ -318,7 +318,7 @@ class spell_item_with_mount_speed : public SpellScriptLoader
         void OnRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
         {
             Unit* target = GetTarget();
-            if (uint16 spellId = getMountSpellId())
+            if (uint32 spellId = getMountSpellId())
```
Helias, line 321: you should put here and in the other one above `auto` and change instead in line 291 the return type uin16 to uint32 of the function `getMountSpellId()`

Helias (changes requested):

src/server/scripts/Spells/spell_item.cpp
Kitzunu: Too long I wrote this, completely forgot about it

Helias (approved):

FrancescoBorzi (approved):

Issue #6538, opened 2021-06-24
[Bloodmyst Isle] Aggressive Monster "Nazzivus Summoner" does not react to players walking close


Eldarix:
[Bloodmyst Isle] Aggressive Monster "Nazzivus Summoner" does not react to players walking close

![WoWScrnShot_062321_032215](https://user-images.githubusercontent.com/81888445/123020212-3a525800-d3d2-11eb-8afc-b804e25ea7d1.jpg)

#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [x] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

Original report: https://github.com/chromiecraft/chromiecraft/issues/962

##### CURRENT BEHAVIOUR:
Look on the screenshot, the aggressive monsters "Nazzivus Summoner" on Bloodmyst Isle don't react to low level players walking close by. They should attack the player immediately like every other aggressive monsters.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
They should attack the player immediately like every other aggressive monsters.

##### SOURCE:
Common sense.

##### STEPS TO REPRODUCE THE PROBLEM:
1. Just walk by with a low level Character. In my case it was a level 13 Warrior Draenei named "Excalibur"

##### EXTRA NOTES:
Thank you for fixing! :D 

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/34da0cda5195dede48cb0406b23248330d6249cb

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker),
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[...]

PR #6590, opened 2021-06-26
fix(DB/Creature) Remove Civilian flag from Nazzivus Summoners

Azcobu:
## Changes Proposed:
-  Removes the Civilian flag from the flags_extra field of Nazzivus Summoners (ID 17524) so they will now attack players.
-  Also set one individual spawn (86538) to have a MovementType of 1 (random wandering) instead of 0 (standing still) to match all the others in the area.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6538
- Closes https://github.com/chromiecraft/chromiecraft/issues/962

## SOURCE:

Common sense.

## Tests Performed:

- Ran SQL, no errors.
- Checked in-game, NPCs are now properly aggressive and attack when you go near them.

## How to Test the Changes:

1. .go c 86538, it should attack you instead of ignoring you.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1624689676324157623.sql
+++ b/data/sql/updates/pending_db_world/rev_1624689676324157623.sql
@@ -1,0 +1,9 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1624689676324157623');
+
+-- Removes Civilian flag from Nazzivus Summoners
+UPDATE `creature_template` SET `flags_extra` = `flags_extra`&~(2) WHERE `entry` = 17524;
+
+-- Set MovementType to 1 for spawn 86538 to match all other Nazzivus Summoners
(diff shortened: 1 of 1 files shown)


Helias (approved):

Helias:
I am not 100% sure about this, because if that flags is setted in this NPC, there will be a reason :/, maybe they took it from sniffs? it should stay there and stand? but I didn't find any source, so it looks ok to me

Azcobu:
I did some more searching and there's a comment on TBC WH about how these guys had a script where they would stand around a summon circle and summon a felguard called Tzerak. There's no trace of that script on AC, and Tzerak just spawns by himself. It seems likely the Civilian flag was a leftover from that, though I think most mobs who are busy with summoning should still attack a player who comes near.

Issue #6469, opened 2021-06-21
Midsummer fest double spawning NPC's


Stump876:
Chromiecraft report https://github.com/chromiecraft/chromiecraft/issues/928

#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

There are a lot of NPC's double spawning for the fire festival event. Anyone can feel free to add to this as they encounter more. 
Additionally some humans are labeled as "Goblin Commoner" and have beer mugs and talk about gathering colored eggs. as well as goblins with human masks with beer mugs saying the same thing. I don't think these should be spawned at all. This is all kinds of incorrect. (second screenshot)
![image](https://user-images.githubusercontent.com/68164076/122705301-48ee1300-d20a-11eb-8a8c-29cbb756da00.png)
![image](https://user-images.githubusercontent.com/68164076/122705364-6a4eff00-d20a-11eb-9095-4add6eab12d3.png)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Should not be doubled up. 

##### STEPS TO REPRODUCE THE PROBLEM:

1. Wild shore Area, Stranglethorn Vale: .go creature: 96711, 23791, 16412, 94603, 94550, 94555, 94551, 16433, 94553, 94552, 94786, 94781, 94694, 94653, 94656, 94693, 94489, 94518, 94654, 94711, 94710, 94655
2. Booty Bay (these are general "Goblin commoner" with mugs and talking about gathering eggs, sometimes there are three: .go creature: 93646, 93762, 93763, 93652, 725, 93587, 93913, 724, 93764, 16433, 727, 93734, 93639, 726, 93610, 
3. Burning Steppes: .go creature: 245649, 248581

##### EXTRA NOTES:

Type here.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/34da0cda5195dede48cb0406b23248330d6249cb

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #6659, opened 2021-06-30
fix(DB/creature): added sql procedure to remove npc duplicates

Hacaw:
## Changes Proposed:
-  added sql procedure to remove npc duplicates based on position, map and events and have a precision loss on position.z of 0.1
-  deletes around ~486 entries from `creature` and `game_event_creature` tables 

## Issues Addressed:

This should fix all the reports about NPC duplication, but **NOT** the missing gameobjects (maybe we should split them in separate tasks).
- https://github.com/azerothcore/azerothcore-wotlk/issues/6469
- https://github.com/chromiecraft/chromiecraft/issues/946
- https://github.com/chromiecraft/chromiecraft/issues/957
- https://github.com/chromiecraft/chromiecraft/issues/950
- https://github.com/chromiecraft/chromiecraft/issues/948
- https://github.com/chromiecraft/chromiecraft/issues/945
- https://github.com/chromiecraft/chromiecraft/issues/944
- https://github.com/chromiecraft/chromiecraft/issues/943
- https://github.com/chromiecraft/chromiecraft/issues/942
- https://github.com/chromiecraft/chromiecraft/issues/940
- https://github.com/chromiecraft/chromiecraft/issues/936
- https://github.com/chromiecraft/chromiecraft/issues/937

## SOURCE:

Because there are so many NPCs we need to automate the process of deletion if we want to use only 1 NPC for all the events but with different model-ids.

## Tests Performed:

- Win 10 Docker Setup from https://github.com/azerothcore/azerothcore-wotlk/commit/0f57da93008d20f61d212aac493b48ace35de0e4
- Run procedure
- Check npcs in world
- I used all the test data provided by the contributors from https://github.com/chromiecraft/chromiecraft/issues/935 as an `where guid in ( contributor-test-data )`  80% of the reported GUIDs are identified (see issues for the remaining 20%)

## How to Test the Changes:

1. Make a backup of the `creature` and `game_event_creature` tables before running the procedure to avoid installing the whole DB docker volume with `create table creature_bk as select * from from creature` and `create table game_event_creature_bk as select * from from game_event_creature`
2. Copy paste the procedure in a SQL client
3. Run it and start verifying the GUIDs reported

- The problem is there are some small precisions for `position.x` and `position.y` and some of the GUIDs reported by contributors are not found by this procedure and adding similar conditions to the group-by will actually return 0 results
- Some NPCs are not found as reported on my DB and I assume because there are different GUIDs between ChromieCraft reports and AzerothCore DB generated from sql-revisions (it's enough for a GM to just do a creature add ingame and the whole guids list will not match anymore)
- I think not all the features used in this procedure are supported by all the mysql versions
- I am not sure if there is any SQL code-convention provided, feel free to add any suggestion about formatting, I used CLION and the database client plugin for formatting
- The procedure takes around ~1.5 minute, added a fixme, maybe I should add some indexes ?
- Some of the builds are failing, should I delete from `creature_addon` table too ?
- Other builds are failing with exit code 1 and a semicolon check, not sure what it means
- We could add a LIMIT to the delete query to avoid magically deleting the entire tables
- Should I add the `smart_scripts` table into the inner joins for the creatures ?


UltraNix:
@Hacaw What is the status of this?

Nefertumm:
What's the status for this one? If no response is given, I'll close the PR.

Issue #6563, opened 2021-06-25
Riding tied to level of your character


markvetx:
**CURRENT BEHAVIOUR:**

If you learn riding, it will give the appropriate skill level (75/75, 150/150 etc.)

But, if you re-log, the riding skill points will tie to the level of your character.

**EXPECTED BLIZZLIKE BEHAVIOUR:**

Riding must not be tied to character's level.

**SOURCE:**

**STEPS TO REPRODUCE THE PROBLEM:**

1. Create a char.
2. Learn riding, `.learn 33391` for example.
3. Re-log.
4. Check skill tab.

**EXTRA NOTES:**

**AC HASH/COMMIT:**
 
edf3df5

**OPERATING SYSTEM:**

Arch Linux 5.12

**MODULES:**

None

**CUSTOMIZATIONS:**

Hacaw:
@markvetx I tried to follow your steps, but I think you should add more.

What commands did you use to level up/down, because if you used `character level x` or `level x` indeed I can confirm your report.

As per https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130065/GM+Commands for the `character level` and `level up` command it states that :

> All stats and dependent values recalculated.

I think this might be an issue with the level up commands.

markvetx:
@Hacaw Yes, I did level up my characters with `.lev xx`, but, the same issue is happening on my "non-boosted by command" characters.

One of my characters had 75/75 riding (rank 1), but after some project updates (after 1be561e, I believe), I mentioned this problem.

Onidsouza:
@markvetx It seems that the changes made in project update [1be561e](https://github.com/azerothcore/azerothcore-wotlk/commit/1be561e03b56dc396270335886e59eddad9fa0c6) had even more extra unforeseen consequences. I've checked the changes made to GetSkillRangeType, and at least for me lots of the sql tables that should populate the stores are actually empty in the database (skilllineability_dbc, skillline_dbc, skillraceclassinfo_dbc). 

Moreover, checking the code at DBCStores.cpp, sSkillTiersStore never seems to be actually properly initialized, so my guess is that it is running empty.  Meanwhile, adding the following code to the switch case in line 8038 (inside GetSkillRangeType) to ObjectMgr.cpp  (reverting some changes on that update) seems to be a hotfix for both your bug and another one caused by professions and riding skills max values being affected by level up as well:

``` 
        case SKILL_CATEGORY_SECONDARY:
[...]

PR #6015, opened 2021-05-23
fix(Core/Player): Use SkillLineAbility.dbc to determine player initia…

Kitzunu:
…l spells - skill assignment done in a new table `playercreateinfo_skills`

* Cherry-pick https://github.com/TrinityCore/TrinityCore/commit/2a3546ca36468b81d673bdb27315006144af3165

* Cherry-Pick https://github.com/TrinityCore/TrinityCore/commit/d28b66bca88de621fd00495e4edc7e4d9d01e7eb

* Cherry-Pick https://github.com/TrinityCore/TrinityCore/commit/193408f3351b8ce1fb17dc87e11fe2e18b46ec41

Co-Authored-By: Shauren <shauren.trinity@gmail.com>
Co-Authored-By: Rothend <67004168+Rothend@users.noreply.github.com>
Co-Authored-By: claudiodfc <claudio.daniel.f.c@gmail.com>

## Changes Proposed:
-  Use SkillLineAbility.dbc to determine player initial spells
-  skill assignment done in a new table `playercreateinfo_skills`
- properly scale current value of hidden skills with player level

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1659
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6036
- Closes https://github.com/chromiecraft/chromiecraft/issues/693

## Tests Performed:

- Builds
- Tested in-game

### Case 1

1. Create a new character
2. See that you get all skills that you should.
3. See that you get all spells that you should.

### Case 2

1. Test that you can learn talents

### Case 3

1. Test "Blood Fury" and see that the Tooltip is updated when you level up

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #6672, opened 2021-07-01
Hackfix skill range check

Onidsouza:
Hackfix for not getting profession and riding skill range right, leading to incorrect adjustment of max skill level of those skills in case of login/levelup

## Changes Proposed:
-  Quick fix for checking correctly whether a skill is a profession or a riding skill.

## Issues Addressed:

- #6563 

## Tests Performed:

- Built with CMake on a Windows 10, tested in game by creating characters, leveling them up and logging in and out to check for changes.

## How to Test the Changes:

1. Create a character
2. Learn a profession/riding skill
3. Log out and log in again into the character, the max value for the skill shouldn't change.
4.  Level up your character, the max value for the skill shouldn't change again

- [ ] To find out why professions and riding skills are not being catched by the command `sSkillTiersStore.LookupEntry(rcEntry->SkillTierID)` a few lines prior to the proposed changes in ObjectMgr.cpp, and address this issue.

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -8041,6 +8041,11 @@ SkillRangeType GetSkillRangeType(SkillRaceClassInfoEntry const* rcEntry)
             return SKILL_RANGE_MONO;
         case SKILL_CATEGORY_LANGUAGES:
             return SKILL_RANGE_LANGUAGE;
+        case SKILL_CATEGORY_SECONDARY:
+        case SKILL_CATEGORY_PROFESSION:
+            // not set skills for professions and racial abilities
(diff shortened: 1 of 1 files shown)


necropola:
> ### ToDO
> Find out out why professions and riding skills are not being catched by the command sSkillTiersStore.LookupEntry(rcEntry->SkillTierID) a few lines prior to the proposed changes in ObjectMgr.cpp, and address this issue.

Exactly! 

1. We needed to actually load `SkillTiers.dbc`.
Done by PR #6631
2. We need to know **what** was loaded  to figure out why certain skills don't have the expected entry.

TrinitiyCore (where the original PR #6015) is cherrypicked from does not need this additional check.

Onidsouza:
@necropola I can confirm the issue is already covered by #6631, my bad! Hadn't updated my files since last week before trying to debug this.

Issue #6698, opened 2021-07-03
Player.cpp is too big, it should be split


FrancescoBorzi:
This is causing troubles with several editors and CPU going very high when working on this file. It's definitely not a good practice to have such a huge file.

I still haven't looked at how to split it so suggestions are welcome.

iraizo:
I tested editing both files with micro,nano,visual studio code and i have no problems but nearly 30000 lines should be split, my suggestion would be first of all splitting all setters and getters into 2 files, `player_setters` and `player_getters` (i havent looked at the naming convetion yet so). Another one could be having player_checks where the `has...` functions could be put in like `Player::HasTankSpec`. The same thing could be done with the update timers, same goes with the `Load...` functions and `Send...` functions.

Bogir:
PlayerDefines.cpp

iraizo:
> PlayerDefines.cpp

I mean.. the whole file is just function defenitions

buddiman:
Maybe categorize the functions, when there are more than let's say 20 functions for one category create a new file.
For Example: 

- Quest related functions into "PlayerQuest.cpp" 
- Equipment related functions into "PlayerEquipment.cpp" 
- Spell / Talent related function into "PlayerSpell.cpp"

This would also help to follow Clean Code principles and results in more readable and mostly, more maintainable code.

FrancescoBorzi:
So I've started this process by first extracting the existing categories of methods:

https://github.com/azerothcore/azerothcore-wotlk/pull/6838

Of course, this is not yet ideal and there are still a lot of methods out of place, but it's a start and makes it already way less messy.

If you want to move more methods into those existing categories or create new categories, PRs are welcome.

iraizo:
> 
> 
> So I've started this process by first extracting the existing categories of methods:
> 
> #6838
> 
> Of course, this is not yet ideal and there are still a lot of methods out of place, but it's a start and makes it already way less messy.
> 
> If you want to move more methods into those existing categories or create new categories, PRs are welcome.

should the PR be made on the fork or this repo?

FrancescoBorzi:
> > So I've started this process by first extracting the existing categories of methods:
> > #6838
> > Of course, this is not yet ideal and there are still a lot of methods out of place, but it's a start and makes it already way less messy.
> > If you want to move more methods into those existing categories or create new categories, PRs are welcome.
> 
> should the PR be made on the fork or this repo?

[...]

PR #6804, opened 2021-07-07
refactor(Core/Player): extract KillRewarder, TradeData and PlayerTaxi

FrancescoBorzi:
- Moved 556 lines of code out of `Player.cpp`
- Updates https://github.com/azerothcore/azerothcore-wotlk/issues/6698


Kitzunu:
Rename file from KillerRewarder to KillRewarder. Sounds better and it is what you use within the file. So it should match

FrancescoBorzi:
> Rename file from KillerRewarder to KillRewarder. Sounds better and it is what you use within the file. So it should match

ah yes, good point, thanks

Kitzunu (changes requested):
above

Kitzunu:
no. Why did you merge it without making the changes?

PR #6870, opened 2021-07-10
fix(Core/Common): -Wdeprecated-enum-float-conversion warning

FrancescoBorzi:
- Fixes tons of warnings like:

```
src/server/scripts/EasternKingdoms/zone_undercity.cpp:1567:48: fatal error: arithmetic between floating-point type 'double' and enumeration type 'TimeConstants' is deprecated [-Wdeprecated-enum-float-conversion]
                            JumpToNextStep(1.5 * IN_MILLISECONDS);
                                           ~~~ ^ ~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -43,13 +43,13 @@
 #define MAX_NETCLIENT_PACKET_SIZE (32767 - 1)               // Client hardcap: int16 with trailing zero space otherwise crash on memory free
 
 // TimeConstants
-#define MINUTE 60
-#define HOUR MINUTE * 60
-#define DAY HOUR * 24
-#define WEEK DAY * 7
-#define MONTH DAY * 30
-#define YEAR MONTH * 12
-#define IN_MILLISECONDS 1000
+constexpr auto MINUTE = 60;
+constexpr auto HOUR = MINUTE * 60;
+constexpr auto DAY = HOUR * 24;
+constexpr auto WEEK = DAY * 7;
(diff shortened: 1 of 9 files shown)


Kitzunu:
std::chrono soonTM

src/common/Common.h
```diff
@@ -42,16 +42,14 @@
...
-    IN_MILLISECONDS = 1000
-};
+// TimeConstants
+#define MINUTE 60
+#define HOUR MINUTE * 60
+#define DAY HOUR * 24
+#define WEEK DAY * 7
+#define MONTH DAY * 30
+#define YEAR MONTH * 12
+#define IN_MILLISECONDS 1000
```
Kitzunu, line 52: Isnt it possible to do 
```cpp
enum TimeConstants : uint32
```

Kitzunu: Maybe not with enum I cant remember, but it is possible to do with enum class. Perhaps that avoids the error and lets us avoid #define

FrancescoBorzi: I've tried that, still got:

```
Arithmetic between floating-point type 'double' and enumeration type 'TimeConstants' is deprecated
```

when multiplying with decimal values

Kitzunu: hm.. Good enough I guess until we switch to std::chrono

Winfidonarleyan (approved):

PR #6874, opened 2021-07-10
BREAKING CHANGE(Deps/Boost): Set 1.74 as minimum version

Kitzunu:
## Changes Proposed:
- Update minimum version to 1.74 to be able to proceed with C++20 build

## Tests Performed:

- Tested on windows, build can proceed and does not break on Boost being broken


Kitzunu:
I will test if 1.74 works to fix CI... Hopefully =)

Kitzunu:
https://github.com/azerothcore/wiki/pull/591

FrancescoBorzi:
https://www.conventionalcommits.org/

if it's `BREAKING CHANGE` it ain't `chore` :)

Kitzunu:
True

FrancescoBorzi (approved):

necropola:
This is (currently) for windows only?
Debian stable (buster) has 1.67.
I'll need `buster-backports` which has 1.74 to meet this new requirement on debian stable.

I find thid recent trend of constantly changingbuild requirements extremely annoying.
Did the user experience become any better through any of those updates?

FrancescoBorzi:
@necropola 

> Update minimum version to 1.74 to be able to proceed with C++20 build

this is required to proceed to C++20.

Also, we are requiring a minimum version that is already 1 year old (see https://www.boost.org/users/history/ ). There is no reason to use an even older version.

> This again breaks building on debian stable. which has 1.67.

Version 1.67 has been released **3 years ago**. Despite what we require and do not require, it's not recommended to use such old software unless there is a strong reason for it. Ideally, people should always keep their dependencies up to date with the latest stable versions of the software, in order to not have bugs that have already been fixed in a more recent version.

FrancescoBorzi:
I'm reverting this change (see https://github.com/azerothcore/azerothcore-wotlk/pull/6975 ) as this is causing too many problems, including failing the Docker build on master and I don't want to deal with this now.

Please note this can stay this way forever, in future we will need to require a newer Boost version. It's just for now it's ok to keep it like this.

Also the next Debian version should fix this problem: https://packages.debian.org/bullseye/libboost-dev

necropola:
Offical announcement of the **Bullseye full freeze on 2021-07-17**: https://lists.debian.org/debian-devel-announce/2021/06/msg00000.html
Which makes it quite likely that we are only 2-3 weeks away from release. :tada:

PR #6875, opened 2021-07-10
fix(Core/WheatyExceptionReport): for c++20

Kitzunu:
## Changes Proposed:
- This removes errors from WheatyExceptionReport

## Tests Performed:

- Build with C++20


FrancescoBorzi (approved):

src/common/Debugging/WheatyExceptionReport.cpp
```diff
@@ -1023,7 +1023,7 @@ bool WheatyExceptionReport::FormatSymbolValue(
     // Determine if the variable is a user defined type (UDT).  IF so, bHandled
     // will return true.
     bool bHandled;
-    DumpTypeIndex(pSym->ModBase, pSym->TypeIndex, pVariable, bHandled, pSym->Name, "", false, true);
+    DumpTypeIndex(pSym->ModBase, pSym->TypeIndex, pVariable, bHandled, pSym->Name, (char*)"", false, true);
```
FrancescoBorzi, line 1026: @Kitzunu what about

```suggestion
    DumpTypeIndex(pSym->ModBase, pSym->TypeIndex, pVariable, bHandled, pSym->Name, static_cast<char*>(""), false, true);
```

https://stackoverflow.com/a/18414126/3497671

Kitzunu: Doesnt work, already tried

Kitzunu: Cant static cast a char

FrancescoBorzi: fair enough

lineagedr: Since the argument "" is a const char* instead of casting it to a char* you can replace the "" argument with a nullptr.
Or better yet you can modify the DumpTypeIndex function declaration and change the char* to have a default value of nullptr. Although looking at the function we never use that parameter anyway so why keep it?

Winfidonarleyan (approved):

Helias (approved):

PR #6893, opened 2021-07-11
feat(Core/Logging): add support fmt::format logging

Winfidonarleyan:
## Changes Proposed:
- Add support new format style `fmt::format` (like c++20 `std::format`)

## Tests Performed:
image
image


src/common/Logging/Log.cpp
```diff
@@ -222,6 +222,11 @@ void Log::outMessage(std::string const& filter, LogLevel level, std::string&& me
     write(std::make_unique<LogMessage>(level, filter, std::move(message)));
 }
 
+void Log::_outMessageFmt(std::string const& filter, LogLevel level, std::string&& message)
+{
+    write(std::make_unique<LogMessage>(level, filter, std::move(message)));
```
lineagedr, line 227: Isn't the std::move here redundant? There's no move constructor in the LogMessage and the message parameter is already a rvalue so we're casting rvalue to rvalue

src/common/Logging/Log.h
```diff
@@ -265,4 +271,49 @@ void check_args(std::string const&, ...);
...
+// New format logging
+#define FMT_LOG_EXCEPTION_FREE(filterType__, level__, ...) \
+    { \
+        try \
+        { \
+            sLog->outMessageFmt(filterType__, level__, fmt::format(__VA_ARGS__)); \
+        } \
+        catch (const std::exception& e) \
+        { \
+            sLog->outMessage("server", LogLevel::LOG_LEVEL_ERROR, "Wrong format occurred ({}) at '{}:{}'", \
```
Winfidonarleyan, line 283: .

FrancescoBorzi (approved):

Kitzunu:
thanks

Issue #6892, opened 2021-07-11
new crash


mpfans:
### Current Behaviour

apply https://github.com/azerothcore/azerothcore-wotlk/pull/6805

It crashed after 75 hours of run

and then new crash log

https://gist.github.com/mpfans/61b177c8a1e1882bb25f80ac70366c1f

### Expected Blizzlike Behaviour

none

### Steps to reproduce the problem

none

### AC rev. hash/commit

341d65c1db0d99c0603e04edbe2fc37314cf20dc

+ 

https://github.com/azerothcore/azerothcore-wotlk/pull/6805

### Operating system

ubuntu 20.04

FrancescoBorzi:
@UltraNix  can you please have a look at this ?

UltraNix:
Still some problems with map and teleport. Something fucked up recently.

FrancescoBorzi:
> Something fucked up recently with pets/minions.

@Kitzunu @Winfidonarleyan were there recent changes in that area?

mpfans:
new crash now
https://gist.github.com/mpfans/ff6eafa1fc655d1f1f88b92dee5bc4a2

Kitzunu:
Still valid?

mpfans:
I'm not sure, but I suggest close

PR #6922, opened 2021-07-13
fix(Core/SpellAuras): prevent crash in Aura::GetType()

FrancescoBorzi:
- Updates https://github.com/azerothcore/azerothcore-wotlk/issues/6892
- Prevent crash https://gist.github.com/mpfans/ff6eafa1fc655d1f1f88b92dee5bc4a2 reported by @mpfans


FrancescoBorzi:
not sure if @UltraNix has a better idea to fix this crash

UltraNix:
Seems valid.

Viste:
if i remember right:
https://github.com/WarheadCore/WarheadCore_AC_Arhive/commit/d7f2af413f4fd9e0e708798eb29cb8146de0698e#diff-1921596918c2026118ef02eba3eb6fa8ab14a9a09b4370145facef1cc5c02ebdR31

FrancescoBorzi:
sure @Viste , but also the null check doesn't hurt so we can merge both

PR #6978, opened 2021-07-16
BREAKING CHANGE(Deps/Boost): Set 1.74 as minimum version

Kitzunu:
## Changes Proposed:
- Update minimum version to 1.74 to be able to proceed with C++20 build

## Tests Performed:

- Tested on windows, build can proceed and does not break on Boost being broken


FrancescoBorzi:
we probably want to wait until the next Debian stable version is released, which will have 1.74 in the default Boost package version, as well as waiting until it is available on Ubuntu 20.04. Currently, in GitHub Actions there is 1.74 but not on a clean, standard Ubuntu 20.04.

Kitzunu:
Status?

FrancescoBorzi:
> Status?

still not available in ubuntu 20.04: https://packages.ubuntu.com/search?keywords=libboost&searchon=names

not sure if there are plans to port it

Winfidonarleyan:
Can use this for `Ubuntu 20.04`
https://github.com/WarheadCore/WarheadBand/blob/42b2ee3d3dd241889b6ea3c4e44ce75c759d13e8/apps/installer/includes/os_configs/ubuntu.sh#L11

Kitzunu:
@FrancescoBorzi @locus313

locus313:
> Can use this for `Ubuntu 20.04` https://github.com/WarheadCore/WarheadBand/blob/42b2ee3d3dd241889b6ea3c4e44ce75c759d13e8/apps/installer/includes/os_configs/ubuntu.sh#L11

That requires adding a 3rd party repo, which I don't think we should do. If I remember correctly the `add-apt-repository` command is not installed by default with ubuntu minimal installs that most vps/cloud providers use. So I don't really think we should up the min version yet.

FrancescoBorzi:
exactly, currently it's only available in ubuntu 21.04 or higher:

https://packages.ubuntu.com/search?keywords=libboost1.74-dev

we need to wait for a backport to 20.04, if there will even be there one

LordArcanine:
this will be a problem for debian 10 users like me...

FrancescoBorzi:
I'd say let's reopen this when it's time, no need to keep it hanging as it's a small change.

PR #6987, opened 2021-07-16
refactor(Core/SpellEffects): Rework some of the null checks

Kitzunu:
## Changes Proposed:
- Yes


FrancescoBorzi:
codestyle looks good but have you tested this in game?

Shybuia:
@Kitzunu how to test this PR ?

Kitzunu:
1. Read the code
2. Make every method trigger and make sure it doesn't crash when there is no player
3. Make sure that everywhere it changed from unit->ToPlayer() to player-> that it does the same thing.

There is no easy way of testing it, just go case by case. But I don't think you need to test every case as they are just copy pastas. If it works for one, then it works for all =)

Kitzunu:
@azerothcore/testers please test

Kitzunu:
please review

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -1359,10 +1359,15 @@ void Spell::EffectUnlearnSpecialization(SpellEffIndex effIndex)
     if (effectHandleMode != SPELL_EFFECT_HANDLE_HIT_TARGET)
         return;
 
-    if (!unitTarget || unitTarget->GetTypeId() != TYPEID_PLAYER)
+    if (!unitTarget)
```
FrancescoBorzi, line 1362: why?

Kitzunu: because if you look a few lines under you have this
```cpp
    Player* player = unitTarget->ToPlayer();
    if (!player)
    {
        return;
    }
```
Which checks the same thing

Kitzunu:
so review and test will be done before newyears? 🤔

FrancescoBorzi:
@Kitzunu I think if test instructions start with "read the code" is going to exclude a big part of our testers base. Also what we really want to make sure is that there is no regression, so the developer should mention what in the game could potentially be affected and what to check. I know that's not always possible, but for example this PR is about spells so testers should check that the spell system is not broken.

Kitzunu:
How to test:

- See that you cannot unlearn specs if you are not a player
- See that if a non player tries to get rewarded from BGs it wont work
- See that only players can use the teleporter from Dalaran to the Spires
- See that summon player effects only summons players

here are some test cases.

morale of the story
unitTarget->GetTypeId() != TYPEID_PLAYER == Player* player = unitTarget->ToPlayer();

Kitzunu:
Still, doesnt matter if someone tests it if it is not being reviewed :shrug:

Nyeriah (approved):

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/blob/3632efc07427e7adcf49a3629c0cb4740f060cfb/src/server/game/Entities/Object/Object.h#L273-L274

Nyeriah:
The changes look good, they're harmless 🙂 
Code style is being changed but those changes won't affect anything in anyway

PR #7004, opened 2021-07-17
fix(Core/WorldSocket): error C2102

Kitzunu:
## Changes Proposed:
- for c++20 build

### DIFF
--- a/src/server/game/Server/WorldSocketMgr.cpp
+++ b/src/server/game/Server/WorldSocketMgr.cpp
@@ -62,7 +62,7 @@ bool WorldSocketMgr::StartWorldNetwork(Acore::Asio::IoContext& ioContext, std::s
     if (!BaseSocketMgr::StartNetwork(ioContext, bindIp, port, threadCount))
         return false;
 
-    _acceptor->AsyncAcceptWithCallback<&OnSocketAccept>();
+    _acceptor->AsyncAcceptWithCallback<OnSocketAccept>();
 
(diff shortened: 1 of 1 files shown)


Winfidonarleyan:
Need test this

src/server/game/Server/WorldSocketMgr.cpp
```diff
@@ -62,7 +62,7 @@ bool WorldSocketMgr::StartWorldNetwork(Acore::Asio::IoContext& ioContext, std::s
     if (!BaseSocketMgr::StartNetwork(ioContext, bindIp, port, threadCount))
         return false;
 
-    _acceptor->AsyncAcceptWithCallback<&OnSocketAccept>();
+    _acceptor->AsyncAcceptWithCallback<OnSocketAccept>();
```
FrancescoBorzi, line 65: not sure about this tbh

FrancescoBorzi (changes requested):

src/server/game/Server/WorldSocketMgr.cpp
Kitzunu: So what do you suggest to do instead?

FrancescoBorzi: @Kitzunu no idea.. unfortunately, I didn't have the opportunity to dig into this

Kitzunu:
gonna close, I dont think this is the right approach

Issue #7012, opened 2021-07-18
[Felwood] Olm the Wise (rare) missing movement pattern


iraizo:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [X] 40-49
- [ ] 50-59

### Current Behaviour

Olm the Wise GUID 51897 (https://wowgaming.altervista.org/aowow/?npc=14343) is stationary in his spawn.

### Expected Blizzlike Behaviour

He should be moving around in a pattern per wowhead comment.

By jbricks on 2008/03/22 (Patch 2.3.3)

Re: Olm the Wise"
Like one of the posts above, I hunted him and I got him!<br><br>
...I rode up to where I had found him 
(thanks to the coords found here) just knowing someone had already tamed
 or killed him but...there he was, flying the same pattern I had seen almost 4 hours earlier.

### Source

https://tbc.wowhead.com/npc=14343/olm-the-wise

https://www.youtube.com/watch?v=evs3q3xk_fs
https://youtu.be/WEg1X9Cs6eg?t=9
https://youtu.be/dkk3VDeS-LU?t=2

### Steps to reproduce the problem

1. Teleport to GUID 51897
2. Observe no movement

### Extra Notes

Only videos I found are from cataclysm or later, but since the spawn location of the mob did not change through patches (only model changed) its safe to assume it followed same patrol pattern in early patches. The videos show 3 different spawn locations and Olm is moving around in all of them in sort of circular pattern from his spawn.

Original report: https://github.com/chromiecraft/chromiecraft/issues/1224

#### AC HASH/COMMIT:
https://github.com/chromiecraft/azerothcore-wotlk/commit/c7129fa0dd7b9951589599654fe83cdcd2a55d7a

gurveleg:
Might aswell correct spawn time while at it - its listed as 2h45 min currently where it should be 21-32h.
Source: World of Warcraft Atlas
![Raretimers](https://user-images.githubusercontent.com/80769357/126544677-dd03f9ea-d8c0-43a2-a4eb-3f5c7624e8cf.jpg)
![olmwise](https://user-images.githubusercontent.com/80769357/126544680-7fb81cd2-6a16-4a66-923d-59056f4676c3.jpg)
 Sorry for late addition, just got my hands on the atlas today.

PR #7039, opened 2021-07-20
fix(DB/Creature): Olm the Wise (rare) missing movement pattern

Asthalor:
I added some new spawn points and added movements to all of them

Closes AzerothCore issue#7012

## Changes Proposed:
-  Added 2 new spawn points 
-  Added movement to the creature on all of them

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7012
- Closes https://github.com/chromiecraft/chromiecraft/issues/1224

## SOURCE:

https://tbc.wowhead.com/npc=14343/olm-the-wise

https://www.youtube.com/watch?v=evs3q3xk_fs
https://youtu.be/WEg1X9Cs6eg?t=9
https://youtu.be/dkk3VDeS-LU?t=2

## Tests Performed:

- .go c id 14343
- Observe if it moves around the spawn point

Video test

https://streamable.com/7fwttv

## How to Test the Changes:

1. .go c id 14343
2. Observe if it moves around the spawn point

- Tried to give it the faction of Cenarion circle as the source
-  " Alliance hunters might be puzzled by the fact Olm the Wise is clearly rated "Tameable" via Beast Lore, yet if they try to tame him they get an "invalid target" message. The reason is because Olm has faction with Cenarion Circle. "
- When i added him the faction i couldnt see it, so i dont know if can be possible or not, just to add.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1626780409112630500.sql
+++ b/data/sql/updates/pending_db_world/rev_1626780409112630500.sql
@@ -1,9 +1,10 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1626780409112630500');
 
--- Added 2 more spwn points and movement on them
-DELETE FROM `creature` WHERE (`id` = 14343);
+-- Added movement to the default spawn point he had
+UPDATE `creature` SET `wander_distance` = 20, `MovementType` = 1 WHERE (`id` = 14343) AND (`guid` = 51897);
+
+-- Added 2 more spawn points and movement on them
 INSERT INTO `creature` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `modelid`, `equipment_id`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecs`, `wander_distance`, `currentwaypoint`, `curhealth`, `curmana`, `MovementType`, `npcflag`, `unit_flags`, `dynamicflags`, `ScriptName`, `VerifiedBuild`) VALUES
-(51897, 14343, 1, 0, 0, 1, 1, 0, 0, 5875.65, -1242.83, 403.304, 5.35714, 9900, 20, 0, 3082, 0, 1, 0, 0, 0, '', 0),
 (301303, 14343, 1, 0, 0, 1, 1, 0, 0, 6091.22, -1485.77, 438.13, 0.18, 9900, 20, 0, 1, 0, 1, 0, 0, 0, '', 0),
 (301304, 14343, 1, 0, 0, 1, 1, 0, 0, 6307.9, -1637.18, 474.611, 0, 9900, 20, 0, 1, 0, 1, 0, 0, 0, '', 0);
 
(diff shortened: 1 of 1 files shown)


LukasVolgger (changes requested):
Same as i reviewed here https://github.com/azerothcore/azerothcore-wotlk/pull/7040

Asthalor:
Changed the SQL query to update the one he had and add the 2 new spawn points instead of deleting all and adding the 3

data/sql/updates/pending_db_world/rev_1626780409112630500.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1626780409112630500');
+
+-- Added movement to the default spawn point he had
+UPDATE `creature` SET `wander_distance` = 20, `MovementType` = 1 WHERE (`id` = 14343) AND (`guid` = 51897);
+
+-- Added 2 more spawn points and movement on them
+INSERT INTO `creature` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `modelid`, `equipment_id`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecs`, `wander_distance`, `currentwaypoint`, `curhealth`, `curmana`, `MovementType`, `npcflag`, `unit_flags`, `dynamicflags`, `ScriptName`, `VerifiedBuild`) VALUES
+(301303, 14343, 1, 0, 0, 1, 1, 0, 0, 6091.22, -1485.77, 438.13, 0.18, 9900, 20, 0, 1, 0, 1, 0, 0, 0, '', 0),
+(301304, 14343, 1, 0, 0, 1, 1, 0, 0, 6307.9, -1637.18, 474.611, 0, 9900, 20, 0, 1, 0, 1, 0, 0, 0, '', 0);
```
LukasVolgger, line 9: Please use an ``UPDATE`` here too, precisely because of the transparency and compactness and because the code can then be repeated without an error message

LukasVolgger (changes requested):

data/sql/updates/pending_db_world/rev_1626780409112630500.sql
Asthalor: The problem i see is that the ids i used (301303 and 301304) were not in the database to start with.
I dont know if its possible to use the update without having the primary keys added first, and i need them for the where clause.

LukasVolgger: > The problem i see is that the ids i used (301303 and 301304) were not in the database to start with.
> I dont know if its possible to use the update without having the primary keys added first, and i need them for the where clause.

Ok, in that case you need to delete the entries first and the ``INSERT`` is ok here

Asthalor: No problem at all!
I added the delete clause before the inserts as you said.

LukasVolgger (approved):

PR #7117, opened 2021-07-25
fix(Core/Player): fix Eluna hook in Player::CanUseItem

r-o-b-o-t-o:
## Changes Proposed:
- The whole function (except for the Eluna hook call) was in a big `if`, this PR reverses the `if` and uses an early return instead.
- Add braces to the `if`s that were in there.
- The Eluna hook (`OnCanUseItem`) is now called correctly.

## Tests Performed:

- Compiled C++ changes and tested on Windows 10 client and server

## How to Test the Changes:

* Make sure that items are still functional, use potions for example
* Before pulling the PR, create a lua script that uses `PLAYER_EVENT_ON_CAN_USE_ITEM` ([docs here](https://elunaluaengine.github.io/Global/RegisterPlayerEvent.html)): notice that the callback is not called. After building with this PR, make sure that the event works. Example script:
```lua
local PLAYER_EVENT_ON_CAN_USE_ITEM = 31
local EQUIP_ERR_OK = 0
RegisterPlayerEvent(PLAYER_EVENT_ON_CAN_USE_ITEM, function(event, player, itemEntry)
	print("Item " .. itemEntry .. " used by " .. player:GetName())
	return EQUIP_ERR_OK
end)
```

### DIFF
--- a/src/server/game/Entities/Player/PlayerStorage.cpp
+++ b/src/server/game/Entities/Player/PlayerStorage.cpp
@@ -2294,48 +2294,70 @@ InventoryResult Player::CanUseItem(ItemTemplate const* proto) const
 {
     // Used by group, function NeedBeforeGreed, to know if a prototype can be used by a player
 
-    if (proto)
+    if (!proto)
     {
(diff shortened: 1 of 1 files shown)


Nefertumm (approved):

Nefertumm:
@azerothcore/testers please test this one too :)

PR #6434, opened 2021-06-19
fix(Core/ScriptsMgr): correct hooks name

Winfidonarleyan:
## Changes Proposed:
- Added `virtual` for `CanSendJoinMessageArenaQueue` and `CanSendExitMessageArenaQueue`
- Rename `CanExitJoinMessageArenaQueue` to `CanSendExitMessageArenaQueue`

PR #7136, opened 2021-07-27
Add "OnBeforeFinalizePlayerWorldSession" Hook.

kaytotes:
I'm working on a module where I need to ensure that the clients cache has been invalidated. This can currently of course be done via the `worldserver.conf` however in my case I need to do this with some logic attached that wouldn't fit in to the core very well. I figure this is exactly the use case for adding a new hook so decided to do so. 

I plan on adding a variety of other hooks so i'm starting with something short and sweet to learn the process of contributing to AC.

If i've missed somewhere I need to update to add documentation please let me know.

## Changes Proposed:
-  Adds a `OnBeforeFinalizePlayerWorldSession` to the `WorldScript`.
- Refactor some parts of login and queue popping to avoid duplication.

## Issues Addressed:
- None

## SOURCE:
- None

## Tests Performed:
- Login (with and without queue)
- Compiled with Visual Studio 16 2019
- Added a world script that implements the function eg

```cpp
void MyWorldScript::OnBeforeFinalizePlayerWorldSession(uint32 &version)
{
    sLog->outError("Before Cache Adjustment: %u", version);

    version = sProgression->GetCurrentPatch();

    sLog->outError("After Cache Adjustment: %u", version);
}
```

## How to Test the Changes:

1. Add a custom module from the `skeleton-module` https://github.com/azerothcore/skeleton-module
2. Add a world script with the new hook. 
3. Login to test hook has fired.

- [ ]
- [ ]


Winfidonarleyan:
Look https://github.com/azerothcore/azerothcore-wotlk/pull/6434#issuecomment-864417518

FrancescoBorzi:
@kaytotes welcome to AzerothCore and thanks for your contribution! I hope it will be the first of several!

@Yehonal can you please have a look at this?

Yehonal:
The PR looks good, although I see some duplicated code that maybe can be generalized with an inline function

Yehonal:
I would suggest you to create a function like: `finalizePlayerWorldSession()` with these instruction inside:

```
        uint32 cacheVersion = sWorld->getIntConfig(CONFIG_CLIENTCACHE_VERSION);
        sScriptMgr->OnBeforeFinalizePlayerWorldSession(cacheVersion);
        session->SendAddonsInfo();
        session->SendClientCacheVersion(cacheVersion);
        session->SendTutorialsData();
```

And the hook name also becomes `OnBeforeFinalizePlayerWorldSession` instead.

WDYT?

kaytotes:
@Yehonal I agree i'd be happy to make those changes.

src/server/game/World/World.cpp
```diff
@@ -3477,3 +3474,13 @@ uint32 World::GetNextWhoListUpdateDelaySecs()
...
+
+void World::FinalizePlayerWorldSession(WorldSession* session)
+{
+    uint32 cacheVersion = sWorld->getIntConfig(CONFIG_CLIENTCACHE_VERSION);
+    sScriptMgr->OnBeforeFinalizePlayerWorldSession(cacheVersion);
+
+    session->SendAddonsInfo();
+    session->SendClientCacheVersion(cacheVersion);
+    session->SendTutorialsData();
+}
```
Yehonal, line 3486:
```suggestion
}

```

Yehonal: new line

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -194,6 +194,13 @@ class WorldScript : public ScriptObject
 
     // Called when the world is actually shut down.
     virtual void OnShutdown() { }
+
+    /**
+     * @brief This hook runs before sending the cache version to the Client.
```
Yehonal, line 199:
```suggestion
     * @brief This hook runs before finalizing the player world session. Can be also used to mutate the cache version of the Client.
```

src/server/game/World/World.h
```diff
@@ -384,6 +384,13 @@ class World: public IWorld
 
     void RemoveOldCorpses();
 
+    /**
+     * @brief Executed when a World Session is being finalized. Be it from a normal login or via queue popping.
+     * 
+     * @param session The World Session that we are finalizing.
+     */
+    inline void FinalizePlayerWorldSession(WorldSession* session);
```
Winfidonarleyan, line 392:
```suggestion
    void FinalizePlayerWorldSession(WorldSession* session);
```

Yehonal: Why? isn't it an optimization?

Winfidonarleyan: I'm not sure, but I think such functions should be written entirely in a `.h` file

kaytotes: You are correct if it's to be `inline` I should have moved the implementation into the header file too. 

Unless something is a super small one liner like a getter / setter I rarely like to put the implementation in .h files as a standard so i'd lean more towards simply not making this inline. It's an extremely minor compile time optimisation anyway.

However i'm happy to defer to whatever you to agree on considering this is my first contribution.

Yehonal: It won't change that much though, since some compilers are able to decide which functions to inline when needed (when that keyword is not specified).

In this case I suggested to inline since this function is a private one that is only used internally to avoid code duplication: in few word we're defining a function to avoid writing the same code twice but since it's not used outside its own class, using the inline makes more sense to me. 

The optimization is not regarding the compiler, but the execution of the code. A non inlined function means that the CPU has to access the memory to get the address of that function and then execute it. When inlined (expanded) instead, the code to execute is already there

Yehonal: It's a good practice to put inline functions inside a header only when you want to reuse that function in other cpp files.

Quoting from C++ FAQ:

```Note: It's imperative that the function's definition (the part between the {...}) be placed in a header file, unless the function is used only in a single .cpp file. In particular, if you put the inline function's definition into a .cpp file and you call it from some other .cpp file, you'll get an "unresolved external" error from the linker.```

**Comment by kaytotes** [comment]
@Yehonal @Winfidonarleyan Do we think this can be merged?

File: src/server/game/Scripting/ScriptMgr.h [hunk 4]
```diff
@@ -194,6 +194,13 @@ class WorldScript : public ScriptObject
 
     // Called when the world is actually shut down.
     virtual void OnShutdown() { }
+
+    /**
+     * @brief This hook runs before finalizing the player world session. Can be also used to mutate the cache version of the Client.
+     *
+     * @param version The cache version that we will be sending to the Client.
+     */
+    virtual void OnBeforeFinalizePlayerWorldSession(uint32& /*version*/) {}
```
Kitzunu, line 203:
```suggestion
    virtual void OnBeforeFinalizePlayerWorldSession(uint32& /*cacheVersion*/) {}
```

Maybe this is a better param name?

kaytotes: @Kitzunu Agreed and pushed.

src/server/game/World/World.h
Yehonal: @kaytotes @Winfidonarleyan so what do you think?

Yehonal: Switched to unresolved since I just noticed this function is inside the "public" list, it should be a private function since is not used outside and, based on what I said before, imho also inline (unless you have a technical reason why not doing it, other than a code-style opinion)

kaytotes: I'll make the appropriate tweaks.

FrancescoBorzi:
> @Yehonal @Winfidonarleyan Do we think this can be merged?

ping @Yehonal @Winfidonarleyan

src/server/game/Scripting/ScriptMgr.cpp
```diff
@@ -408,6 +408,11 @@ void ScriptMgr::OnAfterConfigLoad(bool reload)
     FOREACH_SCRIPT(WorldScript)->OnAfterConfigLoad(reload);
 }
 
+void ScriptMgr::OnBeforeFinalizePlayerWorldSession(uint32& cacheVersion)
```
lineagedr, line 411: For primitive data types like in this case int you should just pass them by value there's no need to pass by reference here.

src/server/game/World/World.cpp
```diff
@@ -3477,3 +3474,13 @@ uint32 World::GetNextWhoListUpdateDelaySecs()
 
     return uint32(ceil(t / 1000.0f));
 }
+
+void World::FinalizePlayerWorldSession(WorldSession* session)
+{
+    uint32 cacheVersion = sWorld->getIntConfig(CONFIG_CLIENTCACHE_VERSION);
```
lineagedr, line 3480: const uint32

src/server/game/World/World.h
kaytotes: @Yehonal I have moved the FinalizePlayerWorldSession function to be private since it won't be used elsewhere. I have also marked it as inline however based on the above quote from C++ FAQ since it will only used in World.cpp i've left the implementation in there. 

I did try to move the implementation into the `.h` file however the `sScriptMgr` instance usage made it a bit of a pain. Pulling `ScriptMgr.h` into `World.h` causes compilation issues. I also attempted to forward declare `ScriptMgr` so that I could instead do `ScriptMgr::instance()->` but that meant I'd then have to go around and include `ScriptMgr.h` everywhere that includes `World.h` and that became an issue.

src/server/game/Scripting/ScriptMgr.cpp
kaytotes: Normally I would agree but in this specific case pass by reference is required as far as I'm aware as I want to be able to modify what the cacheVersion actually is that gets sent to clients.

src/server/game/World/World.cpp
kaytotes: For the same reason as your other comment this is intentional. Normally I'd agree but i'm explicitly wanting to be able to modify this value from the hook.

lineagedr: I see then all should be good.

FrancescoBorzi:
@kaytotes @Yehonal what's the status of this PR? can we merge it?

kaytotes:
@FrancescoBorzi I'd love for this to be merged in been waiting on it for a project i'm working on. Think this still needs approval.

Yehonal (approved):
Can be merged

PR #7439, opened 2021-08-18
fix(Core/Spells): Sacred Cleansing range

Kitzunu:
* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/c8f2056b177ec5f72ff0b183e5bdbbba1db9b4de)

Co-Authored-By: Lucas Nascimento <keader.android@gmail.com>

## Changes Proposed:
* By default Sacred Cleansing is set to range 0 (self) which means it does not work on friends and pets.

* Make range 40yds, the same as Cleanse (4987)

## Tests Performed:

- none

## How to Test the Changes:

1. See that Sacred Cleansing can proc on friendly targets that is not you


Shybuia:
Isn't this Cata pre-patch spell ? I cant remember this from Wotlk at all.

Kitzunu:
As we use 3.3.5a DBC's, no this is a wrath spell =)

Shybuia:
lol sorry, I suppose I was bad holy paladin then, I just checked ingame :) will test it

Shybuia:
I just tested this before PR and it is working, have you tried to validate if this is real issue before making this PR ?? 

https://youtu.be/vIzTz_3VlR8
https://youtu.be/e3ui7EwbFj0

Kitzunu:
No but that is not neccessary.

As you can see here https://wowgaming.altervista.org/aowow/?spell=53659, this is the DBC data displayed, by default the range of this spell is set to `Range 0 yards (Self Only)` which means it is impossible for Cleanse to trigger Sacred Cleansing on any other target than the caster

Shybuia:
Ok I understand you, I am just questioning my test before PR, have I did it incorrectly? Because as you say "it is impossible for Cleanse to trigger Sacred Cleansing on any other target than the caster". Then it shouldn't work in my test before PR, right ? I'm bit confused now.

Kitzunu:
It shouldn't, just for the sake of it try it on a friendly pet.

Also it might be due to that there already exist a hackfix somewhere in the core that "fixes" it but not documented. Which makes it hard to find.

pklloveyou:
Just in case

Kitzunu:
I think I found the hack, will investigate

Kitzunu:
There is a hack within Spell::CheckRange that makes spells change range from only self to not self if you target someone else.

This can cause an exploit for some spells to get longer range than they should.

So while this PR doesn't change anything in-game for now, it will be necessary for whenever we decide to remove the hackfix

FrancescoBorzi:
what's the status of this? has it been properly tested in game so we can merge it?

Shybuia:
sorry I havent looked into this for longer time, I can re-test it ingame and give feedback so we can continue with this

Kitzunu:
> what's the status of this? has it been properly tested in game so we can merge it?
> 
> please update the labels accordingly, let's not leave pending PRs forever

This PR has no effect on the spell ingame as long as the hackfix in Spell::CheckRange is present. However this fix is necessary for when someone removes the hackfix in Spell::CheckRange see https://github.com/azerothcore/azerothcore-wotlk/pull/7439#issuecomment-901866583 for details

Kitzunu:
So it is gtg

Shybuia:
i will be not able to test it anytime soon, sorry

pklloveyou:
Tested

FrancescoBorzi (approved):

Issue #7100, opened 2021-07-24
Crash in WorldSession::SendPacket


Kitzunu:
### Current Behaviour

https://gist.github.com/Kitzunu/5320e4372612b285718ce066acf33f5a

Unhandled exception thrown: read access violation.
**this** was nullptr.

(L#206)

https://github.com/azerothcore/azerothcore-wotlk/blob/59c8f7276eb7a578789360c5cfc8087c9f549969/src/server/game/Server/WorldSession.cpp#L197-L207

### Expected Blizzlike Behaviour

not null :D

### Steps to reproduce the problem

1. .reload page

### AC rev. hash/commit

59c8f7276eb7a578789360c5cfc8087c9f549969

### Operating system

win11

lineagedr:
https://github.com/azerothcore/azerothcore-wotlk/blob/59c8f7276eb7a578789360c5cfc8087c9f549969/src/server/game/Server/WorldSession.cpp#L692

You could try.
```c++
if (format && format[0]) 
```

Kitzunu:
dont work

PR #7451, opened 2021-08-19
fix(CORE/Crash): reload page GetSession check to avoid crash

peti446:
## Changes Proposed:
-  Add check before sending notifications as console does not have a session
- 
## Issues Addressed:

- Closes Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7100

## Tests Performed:

-  Compile without error
- Started server and performed .reload page in console and as gm in game

## How to Test the Changes:

1. Compile
2. Start world server
3. Issue reload page from console
4. Log into gm character
5. Issue .reload page 
6. Check no crash and in game you got notification in middle of screen


src/server/scripts/Commands/cs_reload.cpp
```diff
@@ -856,7 +856,10 @@ class reload_commandscript : public CommandScript
         LOG_INFO("server.loading", "Re-Loading Page Texts...");
         sObjectMgr->LoadPageTexts();
         handler->SendGlobalGMSysMessage("DB table `page_texts` reloaded.");
-        handler->GetSession()->SendNotification("You need to delete your client cache or change the cache number in config in order for your players see the changes.");
+        if (handler->GetSession())
+        {
+            handler->GetSession()->SendNotification("You need to delete your client cache or change the cache number in config in order for your players see the changes.");
```
Kitzunu, line 861: Isnt there one for PSendNotification or similar instead that can be used?

peti446: As fas as I know there is no PSendNotification or similar, this function already does the formatting tho.

Winfidonarleyan: Maybe just use `SendGlobalGMSysMessage`?

Kitzunu: You can use `PSendSysMessage`, I believe it only sends to players

Winfidonarleyan: Why send this message to the players? Players should not see this kind of message, this command is for the gm who is fixing something, he can execute it several times

Kitzunu: Yes, but a player still needs to delete client cache, and I know Barbz wanted it that way. but for me idc, I don't mind either

peti446: I would make sense to send to all players online at the time, however that would still not fix the problem, players offline? Not paying attention to chat ... some kind of announcement would need to be made over several days to get everyone to delete the cache.
I would keep this for GM only, so as a GM you can debug it and once you done make an announcement for all to do so.
`SendGlobalGMSysMessage` does make sense the notification shows a text on screen while this will show it in gm chat and should work fine. It depends on what is the expectation of this output really if it needs to be a notification or not, I do like GMSysMessage better tbh

Kitzunu: use it

peti446: Also on a side note, as the message mentioned you can change the server ClientCacheVersion = 0 to force a cache reset on all clients.

Kitzunu (approved):

Winfidonarleyan (approved):

Winfidonarleyan:
👍

Viste (approved):

PR #7466, opened 2021-08-20
chore(Github/Issue): Correct label for feature requests

Kitzunu:


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -1,7 +1,7 @@
 name: Feature request
 description: Suggest an idea for this project
 title: "Feature: "
-labels: [Suggestion]
+labels: [Priority-Suggestion]
```
Helias, line 4: why priority?

Kitzunu: Because it is cool.

No but it is easy to see if you look at all the priorities as a suggestion never really has any priority.

I like it

FrancescoBorzi: o_O I don't get it, can you provide some examples?

iraizo: i also dont get how this is supposed to be a priority subcategory, it should just be a `suggestion` and `priority` label. 
How are suggestions and priorities even related in any case

Issue #7428, opened 2021-08-18
Warrior Slam and Execute Not Appearing In Combat Log/Floating Text On Miss


Azcobu:
### What client do you play on?

enGB

### Faction

- [x] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1432

Slam or Execute won't show up anywhere when the ability is missing/dodged/parried (combat log, floating text etc).
https://www.youtube.com/watch?v=EuV8iV1A5fo

As shown in the video, testing is done on a high level dummy (for a lot of misses) and none if them show in the combat log or on the floating text addon, it just disappears. Also if you look closely you sometimes get the slam hit animation particle(which you shouldn't get if you miss) and your rage gets dumped(which it shouldn't if you miss) but it still doesn't show up anywhere.
I tried to reacreate this on a low lvl dummy(so your attacks don't miss) and every slam hit shows up in the combat log.
Execute has the exact same bug but can't reacreate it on the dummy for obvious reasons.

Tester: confirmed for both.

Demo. First few seconds shows normal misses and how they display in combat log/floating text. Then comes a lot of Slam spam, which for the most part is entirely invisible in the log apart from one or two hits. For this I had the cooldown cheat on, but power cheat off, so I'm using rage. Then Execute tested on a hyena. This is harder to see because unlike Slam, Execute sets auto-attack on, so there's a lot of autoattacks messing things up. But I'm spamming Execute in there too, and it's not showing at all, apart from hits.

https://user-images.githubusercontent.com/81782124/129918414-c643e717-cc77-4d73-8ee0-21b43a3853ef.mp4

### Expected Blizzlike Behaviour

Your misses should show up in the combat logs/floating text and you shouldn't get the particle effect of the ability hitting and your rage dumped (as it sometimes happen)

### Source

This shouldn't be controversial, as the combat log should treat these abilities like any others.

### Steps to reproduce the problem

Get a high level mob/dummy and spam slam/execute on it and check the combat log/floating text.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/2d77fe9274005a99f8445b8dac9dd0fcd3faf3c5

### Operating system

Ubuntu 20.04

### Modules

[...]

PR #7468, opened 2021-08-20
fix(CORE/Spells): Warrior's slam and execute spells miss bug

jestermaniac:
## Changes Proposed:
aye whas good. this shit is not even a complete fix cuz there's still one problem but i still finna try a PR on this cuz at least the most annoying bug was resolved i think.

so the problem is this: when u execute or slam the server doesn't cast only 1 spell but two. example: if u cast rank 1 slam (id = 1464) the program finna call a script inside ```Spell::CallScriptEffectHandlers``` that is located in ```src/scripts/Spells/spell_warrior.cpp``` that is gonna execute ANOTHER slam spell (id = 50783, don't ask me where this comes from prolly something in dbc?) that is the actual spell that gon send the dmg back to the client.

**note** that both these spells can miss (block,parry,miss,reflect) so u with one cast u got double chance to fail (lmao). 

the problem with this is that if u cast slam, the sever is going to execute spell with id = 1464 and if it miss it gon send back the client a ```SMSG_SPELL_GO``` packet with inside the miss condition **and** the spell id which for some reason won't show up into the combat log or floating combat text either. matter of fact if u change, in the packet, that spell id with 50783 the miss will show up in the log. it seems like it doesn't recognize 1464  as a spell id or some shit.

so that's the first problem. 

the other problem is that if errythang went gucci the first time with spell id = 1464 when the program will execute the final spell with id = 50783 going through spell scripts the spell can still fail cuz ```targetInfo.missCondition``` is calculated every time u call a spell. if it doesn't then errythang is lit and we get dmg packet ```SMSG_SPELLNONMELEEDAMAGELOG``` back in the client otherwise **NOTHING** will be sent back to the client bc the ```SMSG_SPELL_GO``` is sent only once when spell id = 1464.

to fix this shit i did a dirty ass trick and introduced a call to ```Unit::SendSpellMiss``` to let the client know the attack went shit. this fixed both combat log + floating combat text.

## Issues Addressed:

- https://github.com/azerothcore/azerothcore-wotlk/issues/7428

## SOURCE:

None, cuz it common sense

## Tests Performed:

- Tried locally and apart from one problem still present i didnt find anythang else. would be good tho if  some other tester could provide feedback as well.

## How to Test the Changes:

1. pick up a war (i did lvl 49) nothing but a 2h sword
2. spawn high lvl training dummy
3. slam it
4. to test execute pick a low lvl mob and have fun

the elephant in the room that remains to fuck with is the particle effect. this shit is popped when the ```SMSG_SPELL_GO``` packet is received by the client and so basically (in the clase of slam) when spell id =1464 and the cast can succeed so in the client we gon see the particle effect but the next cast can still fail so we can still get like a miss in the combat log.

to fix this latter issue i think we gotta redo how the spell works or dome some spellscript-fu shit to workaround this shit. ideally we would only need to call the spell once no matter what and not call-chaining this shit that create only confusion. 

also forgot to mention: this problem affect a shit ton of other abilities. right now i only worried bout the spells in the mentioned in the issue but ideally a lot of spells got this double-chance-to-fail type of stuff.

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2430,13 +2430,19 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         }
     }
 
+    bool TmpSendSpellMissSent = false;
+
     if (spellHitTarget)
     {
         SpellMissInfo missInfo2 = DoSpellHitOnUnit(spellHitTarget, mask, target->scaleAura);
         if (missInfo2 != SPELL_MISS_NONE)
         {
             if (missInfo2 != SPELL_MISS_MISS)
+            {
                 m_caster->SendSpellMiss(spellHitTarget, m_spellInfo->Id, missInfo2);
+                TmpSendSpellMissSent = true;
+            }
+
             m_damage = 0;
             spellHitTarget = nullptr;
 
@@ -2645,6 +2651,20 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
             m_caster->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_TALK);
             if (unitTarget->ToCreature()->IsAIEnabled)
                 unitTarget->ToCreature()->AI()->AttackStart(m_caster);
+        }
+
+        // send log for spells that wouldn't make it into the combat log as missed or blocked or w/e (ex. slam/execute of warrior)
+        // send only if not already done above
+
+        if (!TmpSendSpellMissSent)
+        {
+            if (missInfo != SPELL_MISS_NONE)
+            {
+                if (CheckIfMissedSpellNeedToBeSent())
+                {
+                    caster->SendSpellMiss(unitTarget, GetTranslatedSpellIdForClientCombaLog(), missInfo);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Viste:
its global problam and this fix is hack :(

Kitzunu:
This seems very hacky

jestermaniac:
it is but u gotta redo a lotta shit if u wanna make it global. like i said in the original PR arework of the logic of the class is needed imo but who gon test everything? doing so u're touching all spells (players + npcs). anyway if y'all got inputs im all ears.

Viste:
you can just port global fix from TC ;)

jestermaniac:
i dont understad @Viste . which branch of tc u be talkin to? 3.3.5 branch got the same issue as AC just tested rn. the other branch is SL i didn't try but that's a whole another core though

immersionnn:
Does this fix the rage issue too? Do all your misses return your rage instead being rng?

Kitzunu:
For me this is too hacky to go into the core. We need a generic fix instead. But good try

Viste:
> i dont understad @Viste . which branch of tc u be talkin to? 3.3.5 branch got the same issue as AC just tested rn. the other branch is SL i didn't try but that's a whole another core though

335 but specific commit ;) they fuck this again :D

jestermaniac:
> 
> 
> > i dont understad @Viste . which branch of tc u be talkin to? 3.3.5 branch got the same issue as AC just tested rn. the other branch is SL i didn't try but that's a whole another core though
> 
> 335 but specific commit ;) they fuck this again :D

idk what that means

Kitzunu:
Thank you for trying but I will close the PR as https://github.com/azerothcore/azerothcore-wotlk/pull/7494 is more generic <3

Issue #7268, opened 2021-08-09
Ghostlands - Dar'Khan Clipping/Pathing Issues


Azcobu:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1363

Killing the enemy Dar'Khan for the Ghostlands quest "The Traitor's Destruction" is complicated by him running through the wall to his immediate left if you do not stay within melee range of him. Attempting to kite him around his room and using LOS to stop his Shadowbolt casting will result in him running through a wall or up into the ceiling and onto a different floor. He will then be in proximity to other enemies hostile to the player yet he does not end up pulling them to the player. Being a Paladin I had to run behind the railing immediately to your left as you enter his room, pop out and run over to the wall he goes into, cast Consecration, and run back behind the railing. Dar'Khan would then run back through the wall, take ~5% damage total HP from the Consecration and I had to rinse and repeat until he died.

Tester: confirmed, Dar'Khan is clipping through the rear wall pretty frequently.

Demo at 2x speed:

https://user-images.githubusercontent.com/81782124/128718174-ff9d027b-2a3d-4025-a592-45eb6aca45c4.mp4

### Expected Blizzlike Behaviour

Dar'Khan does not run through walls or other level geometry. Enemy engages the player in combat and respects normal LOS rules.

### Source

This is not controversial.

### Steps to reproduce the problem

1. .go c 84102
2. Engage Dar'Khan in combat.
3. Run behind the railing immediately to your left after entering his room while still in combat with him.
4. Witness Dar'Khan's true form of wall.

### Extra Notes

https://wotlkdb.com/?quest=9167

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/36a1777b99c400484537c358af8fa6b3a42b13c7

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
[...]

Issue #5410, opened 2021-04-23
Aldaron the Reckless Falling Through Floor [$15]


Azcobu:
## CURRENT BEHAVIOUR:

Aldaron the Reckless (GUID: 56994) in Eversong (target of Wretched Ringleader quest, https://wotlkdb.com/?quest=9076) can be made to fall through the floor he's standing on by hiding behind a pillar while he's casting. He'll run around a bit to get line of sight, and clip through the floor. It did take a while to get it to happen, though.

The original poster on CC did say they could still kill him and collect his head for the loot, so it's not game-breaking, just aesthetic.

Original CC report is https://github.com/chromiecraft/chromiecraft/issues/453

## EXPECTED BLIZZLIKE BEHAVIOUR:

Mobs shouldn't fall through the floor.

## SOURCE:

![WoWScrnShot_042321_134246](https://user-images.githubusercontent.com/81782124/115872638-301ee680-a481-11eb-9f77-981914b929cf.jpg)

## STEPS TO REPRODUCE THE PROBLEM:

1. .go c  56994 and start combat with Aldaron.
2. Use pillar to obstruct line of sight.
3. Watch Aldaron fall through the floor.

## AC HASH/COMMIT:

rev. b385878e4a5b 2021-04-22 22:25:19.

## OPERATING SYSTEM:

Ubuntu 20.10

## MODULES:

N/A.

## CUSTOMIZATIONS:

N/A.

Azcobu:
Quick note, Aldaron is a lvl 8 NPC, so guessing 1-19 tag is more appropriate, not sure why the original CC poster had it on 20-29.

FrancescoBorzi:
thanks @Azcobu , done!

UltraNix:
@Azcobu Do you have mmaps enabled?
I cannot reproduce it with mmaps - without them, it's doable.

Azcobu:
Sorry, I don't know - is there a command to find out? 

My AC instance is vanilla, with no mods or other changes, just what you get from working through the instructions here - https://www.azerothcore.org/wiki/ac-dashboard-core-installation.

If it helps, I was under the impression that a lot of 'mobs falling through floors' issues had been mitigated, although not entirely fixed, by https://github.com/azerothcore/azerothcore-wotlk/pull/5778, so problems reproducing this issue might be because of that.

UltraNix:
Just type ingame command `.mmap loc`

Azcobu:
[...]

Issue #3740, opened 2020-11-22
[NPC] Incorrect casters behaviour


markvetwow:
##### CURRENT BEHAVIOUR:

There is a bug with caster NPCs behaviour. If caster NPC was interrupted or was LoS'ed, they continue to move out from current aggro or trying to move to the point, where they can cast on current aggro again.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

A vast majority of caster NPCs MUST go in melee to current aggro, if they were interrupted or if current aggro breaking its LoS.

##### STEPS TO REPRODUCE THE PROBLEM:

1. Create a character with interrupt ability/spell (rogue, shaman etc.)
2. Learn it, of course
3. .tele brackwell
4. Pull Surena Caledon from the inside
5. Try to avoid her LoS or interrupt her casts

##### EXTRA NOTES:

I'm applying a link to the video [here](https://youtu.be/B9Mfh1gtOQ8)

##### AC HASH/COMMIT:

17afcaf

##### OPERATING SYSTEM:

Windows 10 x64

##### MODULES:

It's a clean build

Quickson:
Can reproduce AzerothCore rev. 0bee5af962c0

markvetwow:
Any guesses what might cause this? :(

markvetwow:
Any guesses on what might cause this? :( It's kinda gamebreaking. You can basically kite the casters all around the building etc. without even being touched.

Hacaw:
How should we handle this ? Let them cast through walls, or make them move faster to get out of LOS ?

aradep:
Can also confirm, something is odd with ranged mobs and LoS.  
[Example of blizzlike behavior in Classic](https://youtu.be/pKCr1DY12XE?t=17740)

markvetx:
This issue is still present even after #4894

esanhueza:
Yeah, it seems casters must change to melee if the target is out of sight and only try to cast again after successfully reaching the target. Right now they try to cast constantly (even when moving) without accounting for their last action.
About the delay, it seems that there is some kind of delay between actions, so if you stop (interrupt, silecence, LoS, etc) one action, they wait for their "next turn" to decide what to do.
I will try to fix this issue during the week. But it would be great if someone can digest the behavior of casters or at least compile some videos of how they should behave.

markvetx:
Could this be related to SAI?

aradep:
[...]

PR #7472, opened 2021-08-20
Fix incorrect pathing of caster mobs when their casts fail due to LOS

srdja:
## Changes Proposed:
Changes the way caster mobs chase you after you move out of their line of sight. Originally when one would hide behind something while still not being too far away, the caster mob would usually move left or right, seemingly unpredictably, to find a spot that is in range and with a clear line of sight. They would only ever try to close in on you if you were completely out of their range.

The change here is that it makes the caster mobs run straight at you after hiding behind something until a clear line of sight is reached. Which is more in line with blizzlike behavior.

## Issues Addressed:
- #3740
- #7268
- #5410

## Tests Performed:

- Tested in places referred to by the issues above.
- Moonbrook mages
- Bloodmyst blood elves
- Inside Karazhan

## How to Test the Changes:

1. Pull a caster mob anywhere in the game and run behind something that would block the line of sight. They should run straight towards you until they reach a clear line of sight.


src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -114,7 +114,13 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     if (PositionOkay(owner, target, maxRange, angle) && !owner->HasUnitState(UNIT_STATE_CHASE_MOVE))
         return true;
 
-    bool moveToward = !owner->IsInDist(target, maxRange);
+    float tarX, tarY, tarZ;
+    target->GetPosition(tarX, tarY, tarZ);
+
+    bool withinRange = owner->IsInDist(target, maxRange);
+    bool withinLOS   = owner->IsWithinLOS(tarX, tarY, tarZ);
+    bool moveToward  = !withinRange || (withinRange && !withinLOS);
```
UltraNix, line 122: `bool moveToward  = !(withinRange && withinLOS);`

srdja: It's more terse, but it took me a moment to parse and unpack the different cases. The other one spells it all out which, in my opinion, makes it easier to read. So I'm not sure what the gain here would be since these things get optimized by the compiler anyway. Any thoughts on this?

FrancescoBorzi:
@srdja thanks for your contribution and welcome to AzerothCore!

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
UltraNix: Just readbility.

srdja: Alright, let's not get stuck on preferences then :-)

UltraNix (approved):

FrancescoBorzi:
@azerothcore/testers please check this one

stoudtlr:
**PR Resolves issue**
tested on Dar'Khan from linked issue #7268
1. ran around left side of stairs out of LOS.
2. Dar'Khan ran forward instead of to the side until he had LOS around the corner
image
3. then ran around other side of the stairs to get out of LOS again
4. Dar'Khan ran forward again until in LOS.  This time was also within melee range
image

Next tested Surena from linked issue #3740 
1. agro'd Surena
2. ran to the left to get out of LOS
3. Surena then ran straight forward to the doorway which put me back in LOS with her and she resumed casting
4. next used Arcane Torrent on rogue to interrupt her casting and ran out of LOS around the house
5. Surena ran straight at me to get within melee range

Tested Aldaron from linked issue #5410
1. kept running around different pillars and he never fell of the edge by running in an odd direction when i was out of LOS
2. he always ran at me until within LOS
image

No issues were found in any of these three tests.

syssneck:
Caster mobs don't melee attack, like here (cast interrupted because of LoS, runs to the player, three auto attacks and starts casting again): https://www.youtube.com/watch?v=pKCr1DY12XE&t=17740s

But with this PR it now looks like this:

https://user-images.githubusercontent.com/68868567/132388998-f969dafb-e4c7-44cd-acf6-1858cb533d5f.mp4

This seems easily exploitable. Just aggro lots of caster mobs as melee and do it like I did, without taking a single hit (because they don't melee hit you).

Asthalor:
We are going to check the smartsAIs for the cooldowns, beacuse as syssneck say a lot of those mobs have problems attacking after casting a spell.
In this case adding 1 or 2 secods more between casts we can reach a more blizzlike feeling.
Example after changing from a 3 to 5 seconds to a 6 to 8 seconds cooldown.

https://streamable.com/zs7avp

For example is this case the defias have a cooldown after casting of 3 to 5 seconds. Taking into account that fireball is a 3 seconds cast they will rarely attack, because they have a 2 seconds BaseAttackTime.
Speaking of the speed of the attacks, on the video they attack way faster than 2 seconds. So we should check that too.
Here we can see it well https://youtu.be/pKCr1DY12XE?t=17750. No parries i think.

patou01:
@Asthalor I believe the behaviour where the mobs stop and wait before they cast is part of the issue. Visible for instance here https://drive.google.com/file/d/1ngQDv4k7-TCSCyV8Z-bD78WGKF-RSWHq/view?usp=sharing at 00:35 and onwards (but happens a few times in that video)

Issue #7497, opened 2021-08-22
Goblin Bruisers Attacking After Player Zones In After Recent PVP


Azcobu:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [X] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1474

I did fight in STV arena, used HS, died.
Same happened after BG got finished, they killed me in their city :D
Happened now in Gadgetzan, first screenshots in Booty Bay.

![WoWScrnShot_082021_203123](https://user-images.githubusercontent.com/80590226/130286555-4bc840db-f584-49d5-991f-ee812fa91684.jpg)
![WoWScrnShot_082021_215250](https://user-images.githubusercontent.com/80590226/130286747-81b7d034-6d6a-4371-81ec-162c8e20b288.jpg)

Tester: confirmed. Tried this two ways - one was killing someone in world PVP and then hearthing back to Booty Bay:

https://user-images.githubusercontent.com/81782124/130342804-0a81e937-e0bb-4d89-967e-0152f3477fbe.mp4

The other was to join WSG, kill someone there, and then /afk out:

![WoWScrnShot_082121_145958](https://user-images.githubusercontent.com/81782124/130342816-8718cb28-3a69-4f41-9af1-cc40a16ccf2a.jpg)

In both cases the active PVP flag is being preserved across teleports. I'm not sure if the context is the same, but possibly related to https://github.com/azerothcore/azerothcore-wotlk/pull/7257 

### Expected Blizzlike Behaviour

After you finish PvPing outside (far away) from goblins they shouldn't aggro you.
They should only aggro if you attack someone in their city or near city where they can see you.

### Source
"If any members of the Alliance and Horde cause disturbances **on city grounds**, the Bruisers are quick to react, assaulting the aggressors and running them out of town, killing them if necessary." - https://wowwiki-archive.fandom.com/wiki/Goblin_Bruiser

I couldn't find a more direct source on goblin bruiser behaviour, but I think it's fair to say they shouldn't be aggroing onto players for things that happen on another continent.

### Steps to reproduce the problem

1. PvP somewhere far away from goblins.
2. HS to them after finishing fight
3. Or be in the city, go BG, exit BG and die :D

### AC rev. hash/commit

[...]

PR #7509, opened 2021-08-22
fix(Core/Entities): contested guards attacking after bg/recent pvp

jestermaniac:
## Changes Proposed:
aye, so always setting PLAYER_FLAGS_CONTESTED_PVP flag when player go in pvp was fucking us cover cuz this is the flag that announce to the contested guards to whoop our ass which imo shouldn't be set always but only when needed like the issue states

what i did was to not set the contest pvp timer if player in bg (extend maybe in arenas?) or there are no contested guards that can see or detect me nearby (not sure about this part maybe better smth else?).

i used MAX_AGGRO_RADIUS for the search of nearby units but idk if that's correct tbh. maybe i should put some greater constant u dig. im taking suggestions cuz idk if in retail there is a fixed radius for this bih

word

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7497

## SOURCE:
look issue referenced

## Tests Performed:

- tested both world pvp and bg

## How to Test the Changes:

1. open two client and log in two different chars
3. with one account the other (or enter bg and leave)
4. .go c 76
5. booty bay goblins shouldn't be after ur ass

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -379,7 +379,6 @@ void Unit::Update(uint32 p_time)
     // WARNING! Order of execution here is important, do not change.
     // Spells must be processed with event system BEFORE they go to _UpdateSpells.
     // Or else we may have some SPELL_STATE_FINISHED spells stalled in pointers, that is bad.
-
     m_Events.Update(p_time);
 
     if (!IsInWorld())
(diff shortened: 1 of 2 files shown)


FrancescoBorzi:
@jestermaniac build fails

jestermaniac:
imma close this tbh. idk tf its wrong. it compiles fine locally

PR #7510, opened 2021-08-22
feat(Script/Command): Allow GMs to disable getting mail

Kitzunu:
## Changes Proposed:
- Allow GMs to accept/disallow others from sending mail to them

## Tests Performed:

- Built
- Tested .mail (on/off) ingame

## How to Test the Changes:

1. use `.mail (on/off)` and try to send mail to that character
2. Log out and in and see if the state is saved on your character. `.mail` to see current state
3. Change values in config to Disable/Enable/Last save state by default


data/sql/updates/pending_db_world/rev_1629659877965039000.sql
```diff
@@ -0,0 +1,14 @@
...
+(30099, 'Accepting Mail: %s'),
+(30100, 'Accepting Mail: ON'),
+(30101, 'Accepting Mail: OFF');
+
+UPDATE `acore_string` SET `content_default` = 'ON' WHERE `entry` = 38;
+UPDATE `acore_string` SET `content_default` = 'OFF' WHERE `entry` = 39;
+
+DELETE FROM `command` WHERE `name` = 'mail';
+INSERT INTO `command` (`name`, `security`, `help`) VALUES
+('mail', 1, 'Syntax: .mail on|off\r\nEnable/disable accepting mail by GM from players');
```
steenburgh, line 14: [Minor] Does this apply to all GMs or just the GM that ran the command? I can't tell from the description

steenburgh: Another minor thing - this command does not say what happens when run with no arguments. Looking at the code though, it looks like it's intended to print out the current status

src/server/game/Entities/Player/PlayerStorage.cpp
```diff
@@ -5499,6 +5499,22 @@ bool Player::LoadFromDB(ObjectGuid playerGuid, CharacterDatabaseQueryHolder cons
                     SetAcceptWhispers(true);
                 break;
         }
+
+        switch (sWorld->getIntConfig(CONFIG_GM_MAIL_TO))
+        {
+            default:
```
Helias, line 5505: never see default in the first line of a switch

steenburgh: Looks like it has to be there because the default is to run case 0

data/sql/updates/pending_db_world/rev_1629659877965039000.sql
Kitzunu: Just copied .whisper description :D I can improve it

src/server/game/Entities/Player/PlayerStorage.cpp
Kitzunu: It's used the same for the 4 switches above, so I just copied it

src/server/game/Entities/Player/Player.cpp
```diff
@@ -2216,6 +2219,8 @@ void Player::SetGMVisible(bool on)
         m_ExtraFlags |= PLAYER_EXTRA_GM_INVISIBLE;          //add flag
 
         SetAcceptWhispers(false);
+
+        SetAcceptMail(false);
```
lineagedr, line 2223: Seems kind of silly to be to block recieving mails when the GM uses the invisibility command.

Kitzunu: Mm maybe yes.

But then I'd argue the same for whisper. Perhaps both should be decided by the player and not when we go invis or not. Thoughts?

lineagedr: @Kitzunu Whisper is fine since the GM goes invis and wishes to not be noticed by other players. With SetAcceptMail he can be noticed when they try to send him a mail and receive an error message

src/server/game/Entities/Player/PlayerStorage.cpp
lineagedr: This may seem off-topic but all those switch statements accumulate to 75 lines of code which can be shorten down to 17 lines

```c++

// GM state
const auto ApplyExtraFlags = [&](WorldIntConfigs configIndex, uint32 flags, void (Player::*func)(bool), bool val)
{
    if (!AccountMgr::IsPlayerAccount(GetSession()->GetSecurity()))
    {
        switch (sWorld->getIntConfig(configIndex))
        {
            case 1: (this->*func)(val); break;
            case 2: if (extraflags & flags) (this->*func)(val); break;
        }
    }
};

ApplyExtraFlags(CONFIG_GM_CHAT,          PLAYER_EXTRA_GM_CHAT,         SetGMChat,         true);
ApplyExtraFlags(CONFIG_GM_MAIL_TO,       PLAYER_EXTRA_ACCEPT_MAIL,     SetAcceptMail,     true);
ApplyExtraFlags(CONFIG_GM_INVITES_TO,    PLAYER_EXTRA_ACCEPT_INVITES,  SetInvites,        true);
ApplyExtraFlags(CONFIG_GM_LOGIN_STATE,   PLAYER_EXTRA_GM_ON,           SetGameMaster,     true);
ApplyExtraFlags(CONFIG_GM_VISIBLE_STATE, PLAYER_EXTRA_GM_INVISIBLE,    SetGMVisible,      false);
```

Thoughts?

steenburgh: I like this, but remember that if we do this, now we have to test every if statement we modified to be sure we didn't break something. May not be worth it

steenburgh: But actually it might not be so bad either, given that the code will be easier to maintain in the future

Kitzunu: I will not change the logic in this PR. This is better for a refactor in another pr

Kitzunu:
now we gtg

Kitzunu:
I probably have to check so that the acore strings are colliding with existing, but codestyle is ready

Kitzunu:
Ready for test and review.... Let's go

@azerothcore/testers

tkn963:
I don't know how to link a specific file that has been changed but is the sql correct?

> DELETE FROM `acore_string` WHERE `entry` IN (30099, 30100, 30101);
> INSERT INTO `acore_string` (`entry`, `content_default`) VALUES
> (30098, 'Accepting Mail: %s'),
> (30099, 'Accepting Mail: ON'),
> (30100, 'Accepting Mail: OFF');

It works fine, however I had issues with the command not existing and tried to import it again to make sure and it throws errors about duplicate entries.

tkn963:
To add on to my issues, using the command .mail only triggers the command .mailbox no matter what I do.

> AzerothCore rev. 2f449326e0ca+ 2021-09-15 17:50:28 +0200 (pr-7510 branch) (Unix, RelWithDebInfo, Static) (worldserver-daemon) ready...
> Starting up Auction House Listing thread...
> AC> AC> Table `command` have non-existing command 'mail', skip.

Kitzunu:
I will test again tonight and see if I can reproduce

tkn963:
> 
> 
> I will test again tonight and see if I can reproduce

Please do. I don't think I'm doing something wrong, the script I'm using is the one I made myself and have been using with all other PRs with no issues. Let me know if I need to have a closer look at what I'm doing, meaning it works on your end.

Kitzunu:
Works for me.

Maybe you forgot to import the SQL file

image
image

src/server/game/Miscellaneous/Language.h
```diff
@@ -1360,6 +1360,9 @@ enum AcoreStrings
     // 30087-30095 reserved for passive anticheat
 
     LANG_DEBUG_LFG_ON                             = 30096,
-    LANG_DEBUG_LFG_OFF                            = 30097
-};
+    LANG_DEBUG_LFG_OFF                            = 30097,
+    LANG_COMMAND_ACCEPT_MAIL                      = 30098,
+    LANG_COMMAND_ACCEPT_MAIL_ON                   = 30099,
+    LANG_COMMAND_ACCEPT_MAIL_OFF                  = 30100,
+  };
```
Kitzunu, line 1367:
```suggestion
};
```

tkn963:
It does complain about the command .mail found in the command table being non-existing and version_db_world contains '1629659877965039000'. I could do a full reset, cloning the repo again and switching to the PR, but the console does say I'm using the pr-7510 branch which is correct. If nothing else, I guess I can't test this PR.

Kitzunu:
1. You need to compile the code
2. You need to import the SQL, but for your own sanity don't import the SQL Rev part in PR's
3. This command is only for GM chars

tkn963:
I just noticed that the sql-file included in this PR is not located in the pending folder so something is not right. I'll figure this out and test the PR properly.

tkn963:
> 
> 
>     1. You need to compile the code
> 
>     2. You need to import the SQL, but for your own sanity don't import the SQL Rev part in PR's
> 
>     3. This command is only for GM chars

I appreciate the suggestion but I have tested several PRs previously without issues, the issue is only that something somewhere screwed up and I'll redo everything from scratch and make sure everything works before properly testing the PR.

tkn963:
I tested the PR. I had one issue. I set .mail off on the GM character, tried to send a mail to that character and it works as it should. I then tried to log the GM account out, without closing the mail window on the second account, and back to the character selection screen and then press Send on the second account again and that account got disconnected and the console says 'Segmentation fault'.

Kitzunu:
Can you do a step by step how I can reproduce that?

Because for me .mail off blocks mail from being sent to that char

tkn963:
The mail is blocked while the GM character is online.

1. Log a GM character and a non-GM character in (Not sure if it matters if the second account is GM or not but mine wasn't)
2. Type .mail off on the GM account
3. Try to send a mail from the second account, it will say the recipient does not exist. DO NOT close the mail window
4. Log the GM character out, back to the character selection screen
5. Try pressing Send on the second character again

Kitzunu:
Ah, that actually sounds believable as we check receive->GetSession()->GetSecurity() and while the character is logged out we dont have a session

tkn963:
It seems to happen no matter if steps are taken on the GM account just before trying to send the mail. A GM character that is offline with .mail off seems to trigger a crash every time.

tkn963:
> 
> 
> Ah, that actually sounds believable as we check receive->GetSession()->GetSecurity() and while the character is logged out we dont have a session

Everything works exactly as it should except for the issue with the session then. So fixing that and I give my 👍 for it being perfect.

Kitzunu:
reason for the crash is because it is trying to find a session of a character that is not online resulting in null, I will take a look at how to improve it in the weekend

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -173,6 +173,14 @@ void WorldSession::HandleSendMail(WorldPacket& recvData)
             mails_count = playerData->mailCount;
         }
     }
+
+    // Check if a GM is blocking mail.
+    if (receive->GetSession()->GetSecurity() && !receive->IsAcceptingMail())
```
lineagedr, line 178:
```c++
if (receive && receive->GetSession()->GetSecurity() && !receive->IsAcceptingMail())
```

Winfidonarleyan (approved):

src/server/game/Handlers/MailHandler.cpp
Kitzunu: ah true, why didnt I just do that lol

Kitzunu, line 178:
```suggestion
    if (receive && receive->GetSession()->GetSecurity() && !receive->IsAcceptingMail())
```

Kitzunu:
someone test it pls

tkn963:
I tested it. There's no crashing this time, obviously. If I use .mail off, try to send a mail it'll say recipient not found - if I log the GM character out right there and then I can then send it even if .mail off is used just before logging out.

My point is: The value doesn't seem to be saving so if a GM isn't online it doesn't care.

tkn963:
I feel like a .mail off variable should be permanent until .mail on is used. I know that's not how the other GM options work because they only affect a character that's online. If I'm wrong in the idea behind this, go ahead and take my acceptance because then it works as it should.

Kitzunu:
Can you test again just for the sake of it. Otherwise I have to come up with something else

tkn963:
> 
> 
> Can you test again just to see if this works. Otherwise I have to come up with something else

I'll get right on it, give me a few minutes to re-compile.

tkn963:
That had no effect. If I had to take a guess, the issue is that the GM isn't actually an active session, so the if will return false either way. I don't know if something like this could be done without altering character/auth databases?

Kitzunu:
Yep I know. Trying to figure out a way to get security without session D:

Kitzunu:
As we force players to have mails always on here https://github.com/azerothcore/azerothcore-wotlk/pull/7510/files#diff-baadebd8cd1117ca48225f316a5ab3fd5fd55b20963394d302341147183db067R166 I think I can safely remove the check from MailHandler

Kitzunu:
I think I found the issue

Kitzunu:
Now this works. But I dont know how pretty it is lol

tkn963:
> 
> 
> Now this works. But I dont know how pretty it is lol

I wish I could give a comment but I'm not very fluent in C++. I can test it though.

tkn963:
I tested with mail on and off, both having the GM online and offline, and I ran into no issues.

Kitzunu:
@Winfidonarleyan could you check again please

Winfidonarleyan:
One moment

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -173,6 +173,36 @@ void WorldSession::HandleSendMail(WorldPacket& recvData)
...
+
+    // Check if a GM is blocking mail.
+    if (receive)
+    {
+        if (receive->IsNotAcceptingMail())
+        {
+            player->SendMailResult(0, MAIL_SEND, MAIL_ERR_RECIPIENT_NOT_FOUND);
+            return;
+        }
+    }
```
Winfidonarleyan, line 185: Can make 1 condition? And add `esle if`

Winfidonarleyan (approved):
Not bad, can merge

src/server/game/Handlers/MailHandler.cpp
```diff
@@ -173,6 +173,28 @@ void WorldSession::HandleSendMail(WorldPacket& recvData)
...
         }
     }
+
+    // Check if a GM is blocking mail.
+    if (receive && receive->IsNotAcceptingMail())
+    {
+        player->SendMailResult(0, MAIL_SEND, MAIL_ERR_RECIPIENT_NOT_FOUND);
+        return;
+    }
+    else if (QueryResult result = CharacterDatabase.Query("SELECT guid FROM characters WHERE extra_flags & 2048 = 2048");) // PLAYER_EXTRA_ACCEPT_MAIL
```
Kitzunu, line 183:
```suggestion
    else if (QueryResult result = CharacterDatabase.Query("SELECT guid FROM characters WHERE extra_flags & 2048 = 2048")) // PLAYER_EXTRA_ACCEPT_MAIL
```

jamadaha:
**Tested**

- [X] Sent mail to someone with ``.mail on`` | **Works as expected** 
- [X] Sent mail to someone with ``.mail on`` while the receiver is logged out | **Works as expected**
- [X] Sent mail to someone with ``.mail on`` after login (Logged out/in to check whether it saves) | **Works as expected** 
- [X] Sent mail to someone with ``.mail off`` | **Works as expected** (Sender gets error cannot find mail recipient)
- [X] Sent mail to someone with ``.mail off`` while the receiver is logged out | **Works as expected** (Sender gets error cannot find mail recipient)
- [X] Sent mail to someone with ``.mail off`` after login (Logged out/in to check whether it saves) | **Works as expected** 
- [X] Config with option 1 (Mail on by default) | **Works as expected**
- [ ] Config with option 0 (Mail off by default) | **Not sure whether it is working**
     * Set option, started server
     * Made new char, set gm on
     * On another char, could send to the new GM char (This is the same behaviour as for config option 1)
- [ ] Config with option 2 (Mail to last save state) | **Not sure whether it is working**
     * With option 0 it still saves the mail off/on state between log off/in
     * So what is the difference between this and the others? Presumably, it will still have a default state, right? I assume that it will default to mail on, but it doesn't state so. 
     * Honestly have no idea what this is, as opposed to the others.

**Works, with exceptions (I think)**

Kitzunu:
I will fix this, I rearanged some things and forgot about that one. Will do it in the week

FrancescoBorzi:
@Kitzunu what's the status of this?

Kitzunu:
irl

Kitzunu:
Idno I cant fix the config

Winfidonarleyan:
Ping me if need help with this, or PR in Discrod @Kitzunu

FrancescoBorzi:
any news about this?

Kitzunu:
nope, I dont have the time to prioritize this over other stuff

PR #7103, opened 2021-07-25
refactor(Scripts/Northrend): code cleanup (part 11)

FrancescoBorzi:
- Just a minor code cleanup, follow up of https://github.com/azerothcore/azerothcore-wotlk/pull/7065
- Cleaning `src/scripts/Northrend/**` source files

PR #7520, opened 2021-08-23
refactor(Core/Common): code cleanup (part 12)

FrancescoBorzi:
- Just a minor code cleanup, follow up of https://github.com/azerothcore/azerothcore-wotlk/pull/7103
- Cleaning `src/common/**` source files

### DIFF
--- a/src/common/Collision/Management/MMapManager.cpp
+++ b/src/common/Collision/Management/MMapManager.cpp
@@ -185,13 +185,9 @@ namespace MMAP
             LOG_DEBUG("maps", "MMAP:loadMap: Loaded mmtile %03i[%02i,%02i] into %03i[%02i,%02i]", mapId, x, y, mapId, header->x, header->y);
             return true;
         }
-        else
-        {
-            LOG_ERROR("maps", "MMAP:loadMap: Could not load %03u%02i%02i.mmtile into navmesh", mapId, x, y);
-            dtFree(data);
-            return false;
-        }
 
+        LOG_ERROR("maps", "MMAP:loadMap: Could not load %03u%02i%02i.mmtile into navmesh", mapId, x, y);
+        dtFree(data);
         return false;
     }
 
@@ -228,15 +224,11 @@ namespace MMAP
             LOG_ERROR("maps", "MMAP:unloadMap: Could not unload %03u%02i%02i.mmtile from navmesh", mapId, x, y);
             ABORT();
         }
-        else
-        {
-            mmap->loadedTileRefs.erase(packedGridPos);
-            --loadedTiles;
-            LOG_DEBUG("maps", "MMAP:unloadMap: Unloaded mmtile %03i[%02i,%02i] from %03i", mapId, x, y, mapId);
-            return true;
-        }
 
-        return false;
+        mmap->loadedTileRefs.erase(packedGridPos);
+        --loadedTiles;
+        LOG_DEBUG("maps", "MMAP:unloadMap: Unloaded mmtile %03i[%02i,%02i] from %03i", mapId, x, y, mapId);
+        return true;
     }
 
     bool MMapManager::unloadMap(uint32 mapId)
--- a/src/common/Collision/Models/WorldModel.cpp
+++ b/src/common/Collision/Models/WorldModel.cpp
@@ -160,7 +160,7 @@ namespace VMAP
 
         // check if tile shall be used for liquid level
         // checking for 0x08 *might* be enough, but disabled tiles always are 0x?F:
-        if ((iFlags[tx + ty * iTilesX] & 0x0F) == 0x0F)
+        if (iFlags && (iFlags[tx + ty * iTilesX] & 0x0F) == 0x0F)
         {
             return false;
         }
@@ -181,6 +181,11 @@ namespace VMAP
             x---------x---> dx
           0           1
         */
+
+        if (!iHeight)
+        {
+            return false;
+        }
 
         const uint32 rowOffset = iTilesX + 1;
         if (dx > dy) // case (a)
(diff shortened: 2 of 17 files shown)


src/common/Collision/Models/WorldModel.cpp
```diff
@@ -621,7 +626,7 @@ namespace VMAP
         {
             //if (fread(&chunkSize, sizeof(uint32), 1, rf) != 1) result = false;
 
-            if (result && fread(&count, sizeof(uint32), 1, rf) != 1) { result = false; }
+            if (fread(&count, sizeof(uint32), 1, rf) != 1) { result = false; }
             if (result) { groupModels.resize(count); }
```
steenburgh, line 630: The condition for this if statement is also redundant as we're already inside the `if (result && readChunk(rf, chunk, "GMOD", 4))` if statement

steenburgh:
What all is included in these code cleanup PRs? Is it mostly just stuff you see, or is there some tool tossing up warnings that you're using?

src/common/Utilities/StartProcess.cpp
```diff
@@ -182,21 +182,21 @@ namespace Acore
...
                 }
                 catch (...)
                 {
-                    result = EXIT_FAILURE;
+                    res = EXIT_FAILURE;
                 }
 
                 my_child.reset();
-                return was_terminated ? EXIT_FAILURE : result;
+                return was_terminated ? EXIT_FAILURE : res;
```
steenburgh, line 199: Personally, I'd prefer to spell out `result` - is naming it `res` a C++ and/or AzerothCore coding standard or something?

steenburgh: Similar comment about the line where `appenderName` was changed to `appendName`

src/common/Collision/Models/WorldModel.cpp
FrancescoBorzi: @steenburgh no, because if it gets inside the first `if`` then `false` will be assigned to `result`

FrancescoBorzi:
> What all is included in these code cleanup PRs? Is it mostly just stuff you see, or is there some tool tossing up warnings that you're using?

I've been fixing warnings coming from CLion

src/common/Utilities/StartProcess.cpp
FrancescoBorzi: @steenburgh me too, but the problem here is that there is already a `result` variable in the outer scope, so this would be shadowing it, which is bad

src/common/Collision/Models/WorldModel.cpp
steenburgh: 👍Right, I missed that

FrancescoBorzi: @steenburgh thanks for checking my PR!

src/common/Utilities/StartProcess.cpp
steenburgh: Maybe it could be renamed to `exitCode` instead?

src/common/Collision/Models/WorldModel.cpp
```diff
@@ -343,7 +348,7 @@ namespace VMAP
         delete iLiquid;
         iLiquid = nullptr;
 
-        if (result && fread(&iBound, sizeof(G3D::AABox), 1, rf) != 1) { result = false; }
+        if (fread(&iBound, sizeof(G3D::AABox), 1, rf) != 1) { result = false; }
```
steenburgh, line 351: Does C++ not do exceptions? Looks like those would make this code a lot more readable

FrancescoBorzi: @steenburgh not sure if I got what you meant, but basically `result` is always truthy there

src/common/Utilities/StartProcess.cpp
FrancescoBorzi: @steenburgh good idea, done!

steenburgh (approved):
Not familiar with the code base, but the changes seem ok

Winfidonarleyan (approved):

src/common/Collision/Models/WorldModel.cpp
steenburgh: Nah that part earlier makes sense. I'm just commenting on the code style in general

```
result = true
if(doThing1() != 1) { result = false }
if(result && doThing2() != 1) { result = false }
if(result && doThing3() != 1) { result = false }
...
return result
```

... is a lot uglier than
```
try {
    doThing1()
    doThing2()
    doThing3()
    return true;
} catch () {
    return false;
}
```

src/common/Collision/Management/MMapManager.cpp
```diff
@@ -185,13 +185,9 @@ namespace MMAP
...
         }
-        else
-        {
-            LOG_ERROR("maps", "MMAP:loadMap: Could not load %03u%02i%02i.mmtile into navmesh", mapId, x, y);
-            dtFree(data);
-            return false;
-        }
 
+        LOG_ERROR("maps", "MMAP:loadMap: Could not load %03u%02i%02i.mmtile into navmesh", mapId, x, y);
+        dtFree(data);
```
Kitzunu, line 190: Should this error always be sent?

src/common/Collision/Management/MMapManager.cpp
```diff
@@ -228,15 +224,11 @@ namespace MMAP
...
-            --loadedTiles;
-            LOG_DEBUG("maps", "MMAP:unloadMap: Unloaded mmtile %03i[%02i,%02i] from %03i", mapId, x, y, mapId);
-            return true;
-        }
 
-        return false;
+        mmap->loadedTileRefs.erase(packedGridPos);
+        --loadedTiles;
+        LOG_DEBUG("maps", "MMAP:unloadMap: Unloaded mmtile %03i[%02i,%02i] from %03i", mapId, x, y, mapId);
+        return true;
```
Kitzunu, line 231: same as above

src/common/Collision/Management/MMapManager.cpp
FrancescoBorzi: @Kitzunu this PR is not changing the behaviour, the `if` is returning so either you use `else` or not, the result is exactly the same

src/common/Collision/Management/MMapManager.cpp
FrancescoBorzi: same answer as above :)

src/common/Collision/Management/MMapManager.cpp
FrancescoBorzi: image

Winfidonarleyan: :/

Issue #3329, opened 2020-08-23
(Windows 10) MeshExtractor Fails to build when compiled via Visual Studio 2019


FiftyTifty:
**##### TITLE:**

MeshExtractor Fails to build when the **TOOLS** and **WITH_MESHEXTRACTOR** flags are enabled in CMake, when building on Windows 10.

**##### CURRENT BEHAVIOUR:**

When following the instructions on the AC wiki, as the tools are not provided, you are required to build them from AC's source after enabling the option in CMake. The compiler used is Visual Studio Community Edition 2017.

All the tools, except MeshExtractor, compile successfully.

**##### STEPS TO REPRODUCE THE PROBLEM:**

1. Download the latest AC repo via git
2. Follow the instructions on the Wiki posted here: https://www.azerothcore.org/wiki/Installation
3. When it comes to "Compiling on Windows", when configuring CMake, enable two flags: **TOOLS** and **WITH_MESHEXTRACTOR**.
4. After the build files have been generated, open **AzerothCore.sln**.
5. Building the solution via Visual Studio's menu **Build** -> **Build Solution** will only fail for **MeshExtractor.vcxproj**.

**##### EXTRA NOTES:**

27/09/20: The latest commit resolved the ACE errors, but now there are a multitude of new errors: https://pastebin.com/LZndMMjM

<details>
<summary>OLD</summary>
Here are the relevant lines from the build log: https://pastebin.com/KWCvwjhN
</details>

**##### AC HASH/COMMIT:**

<details>
<summary>As of 27/09/2020 the issue remains, but with different errors (no ACE errors now), with the following **git log** output:</summary>
The command **git log** outputs the following:
commit 91f4671d07b40a5316268bb4b2096dcd2bdf20d9 (HEAD -> master, origin/master, origin/HEAD)
Author: Stefano Borzì <stefanoborzi32@gmail.com>
Date:   Sun Sep 27 16:31:41 2020 +0200

    fix(Core/Spells): prevent client crash on use spell Neural Needle + ESC (#3383)

commit 1910b4ab35bf6ecc65f02df3e06cba4b6805318d
Author: Stefano Borzì <stefanoborzi32@gmail.com>
Date:   Sat Sep 26 14:46:27 2020 +0200

    chore(Core/Player): remove useless SendQuestComplete (#3380)

commit 8b3a1c1fe58488912f3d3b1a244664e7385b46a1
[...]

PR #7545, opened 2021-08-25
fix(Maps/MeshExtractor): build on windows

Kitzunu:
## Changes Proposed:
* Add missing includes

* Remove duplicate struct from utils

* Remove faulty ASSERT and use already defined.

* Comment out code that has no counterpart

- I mean it cant break it anymore than it already is as wait() and activate() doesn't exist. So this just fixes the build, I don't know if it works.

But presumably it's working the same as it did before.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3329

### DIFF
--- a/src/tools/mesh_extractor/ContinentBuilder.cpp
+++ b/src/tools/mesh_extractor/ContinentBuilder.cpp
@@ -180,7 +180,7 @@ void ContinentBuilder::Build()
                     if ((*_th)->Free)
                     {
                         (*_th)->SetData(itr->X, itr->Y, MapId, Continent);
-                        (*_th)->activate();
+                        // (*_th)->activate(); //@todo: activate is not a member of BuilderThread, where did it come from? Not present in intial commit.
                         next = true;
                         break;
                     }
@@ -196,7 +196,7 @@ void ContinentBuilder::Build()
     // Free memory
     for (std::vector<BuilderThread*>::iterator _th = Threads.begin(); _th != Threads.end(); ++_th)
     {
-        (*_th)->wait();
+        // (*_th)->wait(); //@todo: wait is not a member of BuilderThread, where did it come from? Not present in intial commit.
         delete *_th;
     }
 }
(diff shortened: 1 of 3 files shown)


FrancescoBorzi:
what's the status of this?

Kitzunu:
The status of this is "what should activate() and wait() do?", they don't exist

Viste:
> The status of this is "what should activate() and wait() do?", they don't exist

for what you need this meshext?

patou01:
My guess based on a quick look at the [history](https://github.com/azerothcore/azerothcore-wotlk/commits/master/src/tools/mesh_extractor/ContinentBuilder.cpp) is that it's a call to a removed mother class.
If you look at the [first commit](https://github.com/azerothcore/azerothcore-wotlk/blob/e8e94a0a663be0a1cefe704d8c6860a526b2d7e9/src/tools/mesh_extractor/ContinentBuilder.cpp), you can see `class BuilderThread : public ACE_Task_Base`
Up until this [commit ](https://github.com/azerothcore/azerothcore-wotlk/blob/b9d14372fd1139423d9be8d2952f2249f93c8d83/src/tools/mesh_extractor/ContinentBuilder.cpp )by Francesco it's there.
And then with a [refactor to remove ACE](https://github.com/azerothcore/azerothcore-wotlk/blob/4dd9b92759ad7e2a7e8e4ffce83cb4d90b8afa0c/src/tools/mesh_extractor/ContinentBuilder.cpp): `class BuilderThread`.

If I checkout the commit before "remove ACE from tools" I find:
image
image

Apologies for the white background, haven't fixed np++ yet.
Content of the 2 functions is here in case you don't want to checkout the old commit: https://pastebin.com/4c79hfbQ 



Does this help in any way?

Kitzunu:
Ah yea, ACE is removed

Kitzunu:
@Winfidonarleyan could you take a look?

Viste:
i still don't understand what you want to extract with this mesh extractor

Winfidonarleyan:
> i still don't understand what you want to extract with this mesh extractor

Agree

Issue #7552, opened 2021-08-25
Quest "Gaining Acceptance" Missing Reputation Limit


Azcobu:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [X] 40-49
- [X] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1503

The repeatable quest "Gaining Acceptance" is available at 600/12000 honored.
![image](https://user-images.githubusercontent.com/22718174/130338819-9583ff8f-75bd-4828-b8a2-cb5d6b6264f9.png)

Tester: confirmed. checking the quest (ID 13662) shows no upper limit set for RequiredMaxRepValue, which means it can be taken forever. Also tested this in game:

Exalted rep and still able to accept:
![WoWScrnShot_082521_214950](https://user-images.githubusercontent.com/81782124/130789699-16566875-8c15-4815-8ae7-1226e6d95e3e.jpg)

### Expected Blizzlike Behaviour

The repeatable quest "Gaining Acceptance" should no longer be available once the player has hit 0/12000 honored.

### Source

https://wowpedia.fandom.com/wiki/Thorium_Brotherhood?oldid=2592601
It guides the player to get as far as they can into Honored (600/12000) only with Dark Iron Residue by going all the way up to 5975/6000 friendly with a combinaison of "Gaining Acceptance" and "Gaining Even More Acceptance".

Once Honored, it suggest to do a few quests (but the links are broken) since you would normally no longer be able to hand over Dark Iron Residue for some reputation.

https://classic.wowhead.com/quest=7737/gaining-acceptance
Classic WoWHead comments also suggest once at honored to hand-in materials that are far more complicated to obtain in order to progress.

@gadaotada adds more info:

Yep I can confirm this is a bug as of WOTLK 3.3.5a patch. Here is some examples of it
1. Using this wotlk db u can find this https://wotlkdb.com/?faction=59 read up the guide
2. https://wowwiki-archive.fandom.com/wiki/Quest:Gaining_Acceptance a bit old stuff now
3. https://www.wowhead.com/news/the-exalted-an-overview-of-all-reputations-192128#reputation-classic check the comments on the reputation
[...]

PR #7563, opened 2021-08-26
fix(DB/Quest): "Gaining Acceptance" Reputation limit

iraizo:
## Changes Proposed:
-  Change `Req.MaxRepValue` to 6000 instead of 0

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7552
- closes https://github.com/chromiecraft/chromiecraft/issues/1503

## Tests Performed:

- None

## How to Test the Changes:

Look at https://github.com/azerothcore/azerothcore-wotlk/issues/7552

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1629974973789158607.sql
+++ b/data/sql/updates/pending_db_world/rev_1629974973789158607.sql
@@ -1,3 +1,3 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1629974973789158607');
 
-UPDATE `quest_template_addon` SET `RequiredMaxRepValue` = 6000 WHERE (`ID` = 13662);
+UPDATE `quest_template_addon` SET `RequiredMaxRepValue` = 6000, `RequiredMaxRepFaction` = 59 WHERE (`ID` = 13662);
--- a/data/sql/updates/pending_db_world/rev_1629974973789158607.sql
+++ b/data/sql/updates/pending_db_world/rev_1629974973789158607.sql
@@ -1,0 +1,3 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1629974973789158607');
+
+UPDATE `quest_template_addon` SET `RequiredMaxRepValue` = 6000 WHERE (`ID` = 13662);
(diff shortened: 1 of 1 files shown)


baranzadeoglu:
> @iraizo We need 3000 rep for neutral to friendly and another 6000 rep for friendly to honored, so the max rep value should be 9000 here! I didn't know, and just found this out myself. Please change the value to 9000 otherwise the quest will no longer be available halfway through friendly.

I believe the right max value should be 8375. Here is why.

There is basically 2 quests. From friendly to honored these two are the only way of gaining rep (apart from quests).
One is gaining acceptance, requires 4 residue, rewards 25 rep.
other one is gaining even more acceptance, requires 100 residue, rewards 625 rep.
Now latter already cuts out at max rep 9000 but if you cut the former at 9000 rep as well one can get up to 8975 rep with the first, then turn in the second for 9600 total rep whereas residue turn-ins should have been out of the picture at 9000 rep marker.

syssneck:
@baranzadeoglu That works as intended. The quest only checks, what your reputation is at the moment you "take" the quest and not with the reputation gained after completing the quest. Of course, you could do that and, in fact, the whole point of the quests is, to get to honored (and beyond). The only reason for the "Gaining Even More Acceptance" quest is, to make it less tedious.
Also, from honored to revered you need another 12000 reputation and 600 is very little compared to that. That was only the easy part of the rep grind for the Thorium Brotherhood, the hard part starts from honored on.

The max. value of 8375 renders the problem, that players could get into the situation of not having enough rep to be honored (<9000) but can't do the quests anymore, because their rep might be over 8375 (for example you have 8000 rep and do one of the "Gaining Even More Acceptance" quest = 8625)

There are people in the comments, from patch 3.1.0 & 3.1.3, who did the exact same thing you described back in the day: https://www.wowhead.com/quest=7737/gaining-even-more-acceptance#comments

FrancescoBorzi:
what's the status of this?

syssneck:
@FrancescoBorzi it is all good and can be merged. I'm not sure, if I should set the "to be merged" label without any reviewer approving the changes, so that is why I waited with it. I don't want to skip any steps, but I'm feeling confident enough, to review easy SQL changes myself.

FrancescoBorzi:
> I'm not sure, if I should set the "to be merged" label without any reviewer approving the changes, so that is why I waited with it

yes, you can apply [To Be Merged] as soon as the PR has been tested. The code review approval is also required so GitHub won't let people merge without it ;)

FrancescoBorzi (approved):

PR #1789, opened 2019-05-01
fix(Core): Fixed join multiple guilds exploit

old-intelligentquantum-boycott:
##### TESTS PERFORMED:
Yes, imported without errors. Checked and tested in game, everything looks fine.

##### HOW TO TEST THE CHANGES:
First: Clone WowAddin (WoW Console) from https://github.com/tomrus88/WowAddin
Then: Use below code as a method and register the command in order to function

`bool GuildLeave(char const*, char const* args)
{
  CDataStore Packet(CMSG_GUILD_DECLINE);
  Packet.Finalize();
  ClientServices::SendPacket(&Packet);

  Console::Write("Packet Sent", ECHO_COLOR);
  return true;
}`

After That: Try to compile and Inject it to Wow.exe Process using a DLL Injector
Then: See this video >> https://ufile.io/3wpn1xnl

##### KNOWN ISSUES AND TODO LIST:

- [ ]
- [ ] 

##### Target branch(es):

Master

PR #7631, opened 2021-08-30
Fixed join multiple guilds exploit

IntelligentQuantum:
OLD PR: https://github.com/azerothcore/azerothcore-wotlk/pull/1789

Tested And It works fine for me

First: Clone WowAddin (WoW Console) from https://github.com/tomrus88/WowAddin
Then: Use below code as a method and register the command in order to function

```
bool GuildLeave(char const*, char const* args)
{
  CDataStore Packet(CMSG_GUILD_DECLINE);
  Packet.Finalize();
  ClientServices::SendPacket(&Packet);

  Console::Write("Packet Sent", ECHO_COLOR);
  return true;
}
```

After That: Try to compile and Inject it to Wow.exe Process using a DLL Injector

Then see this video without this PR: https://vimeo.com/594412340


lineagedr:
Why not just do https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/GuildHandler.cpp#L79

```diff
LOG_DEBUG("guild", "CMSG_GUILD_DECLINE [%s]", GetPlayerInfo().c_str());

+if (GetPlayer()->GetGuild())
+    return;

GetPlayer()->SetGuildIdInvited(0);
```

IntelligentQuantum:
> Why not just do https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/GuildHandler.cpp#L79
> 
> ```diff
> LOG_DEBUG("guild", "CMSG_GUILD_DECLINE [%s]", GetPlayerInfo().c_str());
> 
> +if (GetPlayer()->GetGuild())
> +    return;
> 
> GetPlayer()->SetGuildIdInvited(0);
> ```

:+1: Yeah why not, i'm going to test this

IntelligentQuantum:
> Why not just do https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/GuildHandler.cpp#L79
> 
> ```diff
> LOG_DEBUG("guild", "CMSG_GUILD_DECLINE [%s]", GetPlayerInfo().c_str());
> 
> +if (GetPlayer()->GetGuild())
> +    return;
> 
> GetPlayer()->SetGuildIdInvited(0);
> ```

it works fine, Can you make PR ?

src/server/game/Guilds/Guild.cpp
```diff
@@ -2197,6 +2197,15 @@ bool Guild::AddMember(ObjectGuid guid, uint8 rankId)
...
         rankId = _GetLowestRankId();
 
+    ObjectGuid::LowType guidLow = guid.GetCounter();
+
+    CharacterDatabasePreparedStatement* memberGuidQuery = CharacterDatabase.GetPreparedStatement(CHAR_SEL_GUILD_MEMBER_GUID);
+    memberGuidQuery->setUInt32(0, guidLow);
+    PreparedQueryResult memberGuidResult = CharacterDatabase.Query(memberGuidQuery);
+
+    if (memberGuidResult)
+        return false;
```
FrancescoBorzi, line 2207:
```suggestion
    if (memberGuidResult)
    {
        return false;
    }
```

FrancescoBorzi:
@IntelligentQuantum thanks for this PR.

Please don't forget to:

- specify what is the status of it and apply the Label(s) accordingly (waiting to be tested? ready to be reviewed?)
- also specify whether you tested this in-game already or not

lineagedr:
Correct me if I'm wrong but do we even need GetPlayer()->SetInGuild(0); ?

So if we decline a guild invite it will set our current guildId to 0 which doesn't make any sense we should either remove it
or check if we're not in a guild and then set it to 0 which again wouldn't make any sense because we're not in a guild removing it might be the best option
```c++
void WorldSession::HandleGuildDeclineOpcode(WorldPacket& /*recvPacket*/)
{
    LOG_DEBUG("guild", "CMSG_GUILD_DECLINE [%s]", GetPlayerInfo().c_str());
    GetPlayer()->SetGuildIdInvited(0);
}
```

IntelligentQuantum:
@lineagedr I removed this in old PR https://github.com/azerothcore/azerothcore-wotlk/pull/1789/files

lineagedr:
@IntelligentQuantum Any reason that you reverted that fix? Because I don't see any harm in removing it

IntelligentQuantum:
@lineagedr They said it has crash, But i have used it from almost 5 years and there is no problem with it.

FrancescoBorzi:
@IntelligentQuantum is this ready to be tested?

IntelligentQuantum:
> @IntelligentQuantum is this ready to be tested?

Yes

src/server/game/Handlers/GuildHandler.cpp
```diff
@@ -77,6 +77,9 @@ void WorldSession::HandleGuildDeclineOpcode(WorldPacket& /*recvPacket*/)
 {
     LOG_DEBUG("guild", "CMSG_GUILD_DECLINE [%s]", GetPlayerInfo().c_str());
 
+    if (GetPlayer()->GetGuild())
+        return;
```
Helias, line 81:
```suggestion
    if (GetPlayer()->GetGuild())
    {
        return;
    }
```

Helias (approved):

FrancescoBorzi:
@IntelligentQuantum if you tested the normal guild actions and they still work fine (e.g. try to create a guild, invite someone, remove someone, etc...), I'd say we can merge this.

Issue #7572, opened 2021-08-26
Gordunni Shamans cast Healing Wave too often (SmartAI)


syssneck:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [X] 40-49
- [ ] 50-59

### Current Behaviour

Original report: [Chromie issue#1548](https://github.com/chromiecraft/chromiecraft/issues/1548)

Link: https://wowgaming.altervista.org/aowow/?npc=5236
Mob heals themselves without timer.

https://user-images.githubusercontent.com/31207969/130994735-55795623-614c-487b-b1d6-5bd9fc1a1c43.mp4

### Expected Blizzlike Behaviour

 It should be heal themselves below %40 health at every 14 - 21 seconds.

### Steps to reproduce the problem

NPC Info
```
DB GUID: 50243, current GUID: 13160.
Faction: 45.
npcFlags: 0.
Entry: 5236.
DisplayID: 11559 (Native: 11559).
Level: 44.
EquipmentId: 1 (Original: 1).
Health (base): 1712. (max): 1712. (current): 1712.
Unit Flags: 0.
Unit Flags 2: 2048.
Dynamic Flags: 0.
Faction Template: 45.
SpawnTime: Full:5m Remain:0s
Loot: 5236 Pickpocket: 5236 Skinning: 0
InstanceID: 0
Phasemask: 1
Armor: 2527
Position: -5467.840820 1280.123413 22.367174.
AIName: SmartAI ScriptName: 
MechanicImmuneMask: 0
SpellSchoolImmuneMask: 0
```

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/1f44732417968953c965e392c373cc453f79cf02

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
[...]

PR #7647, opened 2021-08-30
fix(DB/Creature): Reduced casts of healing from Gordunni Shamans

Asthalor:
They had 2 events that triggered a heal.
One checked a friendly unit, and the other himself. Because he could heal him with both i deleted the one that checked friend and leave the one with the self healing.
## Changes Proposed:
-  Removed the  event 14 - Friendly Health

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7572
- Closes https://github.com/chromiecraft/chromiecraft/issues/1548

## SOURCE:

None

## Tests Performed:

- .go c id 5236 
- Attack him and check the timers, they should heal themselfs once every 14 to 21 seconds when below 40% hp

Video

https://streamable.com/htcjt5

## How to Test the Changes:

1. .go c id 5236 
2. Attack him and check the timers, they should heal themselfs once every 14 to 21 seconds when below 40% hp


Asthalor:
I had some messing with master branch that should not mess with this pr. :P

syssneck:
Okay, but supposedly you pull multiple mobs, then the shamans can't heal their friends anymore. Is this a good way to deal with the issue? This might be an overarching issue with the event itself in SmartAI: `14 - FRIENDLY_HEALTH`.

Si1ker:
image


Against

[TrinityCore - Gordunni Shaman SAI](https://github.com/TrinityCore/TrinityCore/blob/f759809d9d4364bc1d988e4390d3d5a33d5469e9/sql/old/6.x/world/02_2015_11_08/2015_07_20_01_world.sql#L159)


😿

Asthalor:
Yeah. maybe that was a bigger change that needed. The thing is that the code shared by  Si1ker is exactly the same as we have, so if we pick it by the source it means that this issue is expected.

image

The other option was to remove the self healing, given that the smartAI option 'Gordunni Shaman - Between 0-40% Health - Cast \'Healing Wave\'' heals himself, and the option 'Gordunni Shaman - Friendly At 450 Health - Cast \'Healing Wave\'' can target him.

FrancescoBorzi:
what's the status of this?

Asthalor:
We dont know if the issue here is blizzlike or not. The event 4 Frienly NPC within 40m is at 450HP trigger the heal, but it trigger the heal on himself too. I tought of deleteing the event 5, which is the self healing so he only have 1 healing, but its only my opinion and not anything related to blizzlike, so we need more data.

Si1ker:
I just came from the video posted in the chromiecraft issue, there are other mobs with healing spells around Azeroth that have the same behavior. He is using a shaman character that can interrupt spell casting as well as other classes have "fear, kicks, mind freeze, stuns, any other kind of spells that could interrupt casting". Example on Shamans:
image


Those spells that cut spell casting are in WoW not only meant to be used in PvP, also in PvE scenarios like this one.

All sources point to the same place: "this is blizzlike behavior, not a bug, use your spells bruh."

Asthalor:
Then i will close this PR as its intended.

PR #7664, opened 2021-08-31
feat(DB): adds location information to game objects 1/17

jamadaha:
## Changes Proposed:
- Adds map, zone, and area id to the first 5000 game objects

## Tests Performed:

- Verified that the correct amount of rows was changed — 4999 rows — as GUID 3716 is already correct
- Checked every 200'th game object location — should be fine as the data was gathered automatically.
- Spot checked some data points in newly inserted data, to make sure that it was inserted correctly

## How to Test the Changes:

1. Run SQL query
2. Should change 4999 rows (Missing 3716, as it's already correct)
3. Check some data points. No need to check all. (If you want to go ahead, but it would take 2h+)
4. Make certain that each map, zone, and area ID is in the correct column in DB

## Known Issues and TODO List:

- The data is imprecise, not as in wrong. Sometimes the area is set to the zone instead e.g. where it should say Ironforge it would instead say Dun Morugh. I don't why it does this, as it doesn't occur when manually checking the data.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #7731, opened 2021-09-05
fix(DB): adds location information to game objects 2/17

jamadaha:
## Changes Proposed:
- Adds map, zone and area id to game objects
- This is a continuation of #7664 
## Tests Performed:

- Verified that the correct amount of rows was changed — 5000 rows
- Checked every 100'th game object location — should be fine as the data was gathered automatically.
- Spot checked some data points in newly inserted data, to make sure that it was inserted correctly

## How to Test the Changes:

1. Run SQL query
2. Should change 5000 rows
3. Check some data points. No need to check all. (If you want to go ahead, but it would take 2h+)
4. Make certain that each map, zone, and area ID is in the correct column in DB


Nyeriah:
How exactly are you extracting this information?

For example, **mapId** is not a cosmetic field, like the other two. It indicates where the object is supposed to be spawned. There's a quite significant difference between setting a GO to map 0 (Eastern Kingdoms) and 532 (Karazhan) https://github.com/azerothcore/azerothcore-wotlk/blob/7e2e6f8ee85f82f57a3c6e0f6f645b4c42292e7a/data/sql/updates/db_world/2021_09_13_08.sql#L1513

That introduce this issue: https://github.com/azerothcore/azerothcore-wotlk/issues/8319
And probably several others...

stoudtlr:
> How exactly are you extracting this information?
> 
> For example, **mapId** is not a cosmetic field, unlike the other two. It indicates where the object is supposed to be spawned. There's a quite significant difference between setting a GO to map 0 (Eastern Kingdoms) and 532 (Karazhan)
> 
> https://github.com/azerothcore/azerothcore-wotlk/blob/7e2e6f8ee85f82f57a3c6e0f6f645b4c42292e7a/data/sql/updates/db_world/2021_09_13_08.sql#L1513
> 
> That introduced this issue: #8319 And probably several others...
> 
> But without knowing how exactly you're extracting this information it's impossible to tell what to expect. For now I suppose those instance related objects need to be rechecked, if we're able to identify them

You confused me at first.
1. this PR didn't modify that GO
2. i didn't see any instances of double objects with any of the 50 that I tested
3. the issue you mentioned was opened a month before this was commited

I'm tracking now though.  :-)
It was PR#[ 7815](https://github.com/azerothcore/azerothcore-wotlk/pull/7815) that makes these same changes but to different GO's that you were referencing.  

You're right.  I reverted to a commit prior to that checked ingame and there was no double door.  Then manually edited that one GO and the double door appeared.
image

stoudtlr:
idk where i got the time from.  not enough coffee this morning i guess, but like i said, you're right.  It was one of the commits from this group of PR's that caused the issue.

jamadaha:
@Nyeriah @stoudtlr The manner of gathering was through a script that travels to all the gameobjects ingame, where it then writes ```.gps```. I didn't know this was anything but cosmetic for the gameobjects, just wanted it for grouping objects for queries in the DB.

Nyeriah:
> @Nyeriah @stoudtlr The manner of gathering was through a script that travels to all the gameobjects ingame, where it then writes `.gps`. Where the gm log file then contains all the information, which was then converted to CSV. I didn't know this was anything but cosmetic for the gameobjects, just wanted it for grouping objects for queries in the DB.

Cool, that explains why that gate showed up inside the instance. You get teleported inside of it once you try to port to it using .gob object. Thanks for the information.

It's great, really, but it would probably be best if it skipped altering mapId in the next iterations. I dare say mapId is something that can't be wrong, because once you set it to a certain value you already got in mind where exactly you want the object to be spawned.

Issue #6239, opened 2021-06-05
Minions not engaging in combat, again [$20]


arielcami:
**CURRENT BEHAVIOUR:**
- Pet isn't engaging in combat.

**EXPECTED BLIZZLIKE BEHAVIOUR:**
- Pets should rush right at it's master's attacker.

**STEPS TO REPRODUCE THE PROBLEM:**

1.   .go xyz -7702.6 -4217.43 9.76 1 5
2.   Stand 30 yards away.
3.  Attack Wastewander Shadow Mage, ignore his pet.

**EXTRA NOTES:**
-This issue has been addressed and merged in the past(#5166), but it's still not working in this particular case.
-Video:
https://user-images.githubusercontent.com/71854140/120896981-9e5ed900-c5e9-11eb-9cf5-c2ebb3444e6c.mp4

**AC HASH/COMMIT:**
- f819cf86fd7f

**OPERATING SYSTEM:**
-Win10 64bits

**MODULES:**
-No modules

<bountysource-plugin>

---
There is a [$20 open bounty](https://www.bountysource.com/issues/98988587-minions-not-engaging-in-combat-again) on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).

Azcobu:
Confirmed on rev. f819cf86fd7f

https://user-images.githubusercontent.com/81782124/120897603-2a164b00-c666-11eb-8fad-595f853f74f4.mp4

Kitzunu:
That PR looks very hacky. That should be handled by the core, imagine if you had to add that to all minions/guardians' SAI

T1ti:
yes this should obviously be a general core fix

mserajnik:
Corresponding ChromieCraft issue: https://github.com/chromiecraft/chromiecraft/issues/862 (the PR that closes this issue should close that one too)

Kitzunu:
Also the one who fixes this should remove the hack fixes from the DB

syssneck:
This is still an annoying bug and makes warlock mobs way too easy to kill. Is someone on this, or can we maybe add a bounty for this?

Azcobu:
Agreed, this is coming up rather a lot and it's affecting quite a lot of summoner mobs. Tag added.

TheRiddermark:
[...]

PR #5166, opened 2021-04-06
fix(DB/smart_scripts): Minions not engaging

Stifler82:
## Changes Proposed:
-  Fixed SAI NPC: [Imp Minion](https://wowgaming.altervista.org/aowow/?npc=12922)
-  Fixed SAI NPC: [Voidwalker Minion](https://wowgaming.altervista.org/aowow/?npc=8996)

## Issues Addressed:
- Closes https://github.com/chromiecraft/chromiecraft/issues/266
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4969
- Related PR https://github.com/azerothcore/azerothcore-wotlk/pull/4626 **(Little Correction of this PR)**

## Tests Performed:

- Tested in game and working
- Tested on Windows 10 (64)
- **Video Test:** [Npc Minions not Engaging](https://youtu.be/wXh_WH4I4PA)

## Known Issues and TODO List:

- [ ]
- [ ] 

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #7836, opened 2021-09-12
fix(Core/Unit): Improve pets behavior

patou01:
Attacking pets does not put owner in combat and vice-versa.
This attempts fixing it.

Fixes issue AzerothCore #6239 / ChromieCraft 862

## Changes Proposed:
-  CombatStart() on a target with a pet will set the pet in combat 
-  CombatStart() on a target that is a pet will set the owner in combat

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6239 

## Tests Performed:

- Tested in game, videos available: https://drive.google.com/drive/folders/1HBOJ1Qb-9YQcGCPLRHh8Qy1t2Y1EiG7k?usp=sharing 
- Moved to different zones with known pets to test behaviour in both directions (as the bug is symmetric)

## How to Test the Changes:

1. recompile core with change
2. Move to any mob with a pet and attack the pet or the owner in separate pulls. Example mobs are at positions:
.go -42.97 -2935.17 93.46
.go 1484.946045 -4852.268066 12.592702
.go xyz -7702.6 -4217.43 9.76 1 5
These are however mobs I have tested on
3. Both mobs should come each time.

- Confirm hunter pets have not been affected.
- Modify in case of multiple pets, I don't know if these mobs exist and couldn't figure it out with m_Controlled.
- **Is the logic right? This is my first time writing code for an emulator core and could not find the documentation I wanted. I highlight this here to avoid a non-working fix like https://github.com/azerothcore/azerothcore-wotlk/pull/5166, this needs a proper review.**
- The fix didn't work in Slave Pens (as per the last video), but I think the ray might not be considered as a controlled creature (DB error?). Input appreciated on that. I couldn't find a pack like this outside of these. Should I open an issue for these, or is my fix lacking?


Si1ker:
@patou01 https://github.com/azerothcore/azerothcore-wotlk/pull/5166/files
Seems a hack fix and not a proper one, also it helped manually with only the NPCs in the DB update, not all in the world.

Thanks for the contribution, let's wait for an additional tester to apply these changes at local and provide feedback.

Si1ker:
@azerothcore/testers

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12734,7 +12734,31 @@ void Unit::CombatStart(Unit* target, bool initialAggro)
...
+                if (Unit* controlled = target->GetFirstControlled())
+                {
+                    controlled->SetInCombatWith(this);
+                    SetInCombatWith(controlled);
+                    controlled->AddThreat(this, 0.0f);
+                }
+            }
+
+            // if unit has an owner, put owner in combat.
+            if (Unit* enemyOwner = target->GetOwner())
```
Kitzunu, line 12749:
```suggestion
            if (Unit* targetOwner = target->GetOwner())
```

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12734,7 +12734,31 @@ void Unit::CombatStart(Unit* target, bool initialAggro)
...
+            {
+                if (!(enemyOwner->IsInCombatWith(this)))
+                {
+                    /* warding off to not take over aggro for no reason
+                    Using only AddThreat causes delay in attack */
+                    if(!enemyOwner->IsInCombat())
+                        enemyOwner->ToCreature()->AI()->AttackStart(this);
```
Kitzunu, line 12756: brackets

Kitzunu (dismissed):

src/server/game/Entities/Unit/Unit.cpp
patou01: Changed in the new commit, as I couldn't figure out how to get the full change here.

src/server/game/Entities/Unit/Unit.cpp
patou01: Should be fixed. Shouldn't this be caught by a linter?

Helias (approved):

src/server/game/Entities/Unit/Unit.cpp
Kitzunu: probably

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12734,7 +12734,33 @@ void Unit::CombatStart(Unit* target, bool initialAggro)
             if (target->IsPet())
                 target->ToCreature()->AI()->AttackedBy(this); // PetAI has special handler before AttackStart()
             else
+            {
                 target->ToCreature()->AI()->AttackStart(this);
+                // if the target is an NPC with a pet or minion, pet should react.
+                if (Unit* controlled = target->GetFirstControlled())
```
Kitzunu, line 12740: is there some better name to name it? like targetControlledUnit or something. imo controlled is just weird 🤷

stoudtlr:
Testing results:
Went to the Barrens and tested Razormane Hunters.  If I attacked the Razormane or his pet the other would start attacking.  Works as expected now.
Next went to the Blasted Lands to test Shadowsworn Warlocks.  When you attack the warlock his Voidwalker minion will NOT initially attack.  Voidwalker minion only starts to attack once Warlock uses all mana and moves within melee range or you move to within melee range.  Expected behavior is for Voidwalker to move to you.

From my limited testing this seems to have fixed pets, but minions might still need work.

Shadowsworn Warlocks can be found here:
.go -11486 -2691 7

patou01:
I suspect the issue you found is "good testing finds other bugs". While I can reproduce the issue, my best guess is that the link between the 2 is broken for the mob you found.

Here's my findings for now, but I'm very deep in unknown territory so someone who's good with databases could confirm.
Burning blade apprentice (3198) casts spell 12746 to summon a voidwalker of id 8996.
Shadowsworn Warlock (6008) casts spell 5108 to summon a voidwalker of id 8996. 8996 is defined in code as GENERIC_VOIDWALKER:
image

I suspect the cast of 5108 doesn't enable the link to the owner. Interestingly, 5108 is also showed on wowhead for a mob (defias conjurer, id 619) in Westfall, which doesn't use that spell on our server.

**It seems editing 5108 to 12746 in the smartAI for the shadowsworn warlock fixes the issue.** However, unless someone is good with that and can confirm that is the way to go, I will not include it here, that is because it is not a core fix.

However it highlighted another one, which is that social aggros aren't working properly with pets. So I'll do some more digging...

stoudtlr:
More testing and you are correct.  The voidwalker minions in Durotar DO attack as expected.  So it may be an unrelated issue I found with the Shadowsworn Warlock's and their minions.

However, please see if you can recreate this consistent crash I'm getting now.
1. create hunter character
2. level up to 10
3. do hunter quests to get tame, mend, feed, etc spells
4. tame a pet
5. wander around hostile area until your pet is attacked

If I am attacked first everything works fine, however, if enemy target my pet first instead of me instant crash of world server

patou01:
Potential root cause here (ToCreature() and AI() should be more guarded than that), but I need to look into it again tomorrow or later to confirm what's up.
```
    if (!targetOwner->IsInCombat())
    {
        targetOwner->ToCreature()->AI()->AttackStart(this);
    } 
```
Thanks for the report!

Si1ker:
We really need this PR asap.

Pets on current core not even engaging in combat by proximity as any hostile mob would do, also if master is attacked/killed it never gets in combat:
2021-09-14 22_19_01-World of Warcraft

This is on ChromieCraft, the AzerothCore's server.

patou01:
If it's urgent, I suggest doing it in 2 steps.

1) I fix the crash cause on this one and we merge
2) we figure out the pet not entering combat issue from social aggro and
make another PR

I'll try to update it tonight, but the issue has been known for months by
now, so...

patou01:
To the best of my knowledge, this last one addresses the issues we have.

However, I found other bugs for which I absolutely need some support because the aggro system seems to be... dogshit to me. Pets for some reason **do not aggro by range**. They will juts start chasing after you. CombatStart() does not seem to call AttackStart() and vice-versa. I really would need to understand how in the code:
- aggro distance is checked
- how a combat is started.

stoudtlr:
tested your latest commit and saw no issues.  Pets still attack and the crashes went away.  It's not my call, but I would agree with your recommendation of merging this and creating a new issue and pr for the other bugs that were found.  Your fix at least gets things heading in the right direction.

Si1ker:
image

Left: I'm the target at the pet but is not attacking me -> The owner is not even there.
Right: The pet whether is not around the owner or does not have an owner.

As no other contributor has any complaints about these changes, I'm approving this. Other issues related can be fixed along the way.

Si1ker (approved):

patou01:
@Si1ker , how should I proceed for the follow-up fix however? Do I make a new issue detailling the remaining issues? (It's mainly on social and distance aggro of pets). Thanks.

Si1ker:
> @Si1ker , how should I proceed for the follow-up fix however? Do I make a new issue detailling the remaining issues? (It's mainly on social and distance aggro of pets). Thanks.

@patou01  That's correct, new issues/PRs from here addressing directly the remaining stuff.

PR #7852, opened 2021-09-13
fix(DB/Gameobject): adds location information to game objects 4/17

jamadaha:
## Changes Proposed:
-  Adds map, zone and area id to game objects
- This is a continuation of #7815 

## Tests Performed:

- Verified that the correct amount of rows was changed — 5000
- Checked every 100'th game object location — should be fine as the data was gathered automatically.
- Spot checked some data points in newly inserted data, to make sure that it was inserted correctly

## How to Test the Changes:

1. Run SQL query
2. Should change 5000 rows
3. Check some data points. No need to check all. (If you want to go ahead, but it would take 2h+)
    * Check by typing .gps at the gameobject, then compare that to DB (After query)
4. Make certain that each map, zone, and area ID is in the correct column in DB

* Admin objects can have wrong the location. The cause is that you can't teleport to some objects via ``.go object``


Helias (approved):

Helias:
query executed without issues

PR #7863, opened 2021-09-13
fix(Core/Spell): PLAYER_FIELD_BYTE2_INVISIBILITY_GLOW applies only wi…

Kitzunu:
…th invisibility type INVISIBILITY_GENERAL (0)

Co-Authored-By: Meji <2695278+meji46@users.noreply.github.com>
* cherry-pick commit <https://github.com/TrinityCore/TrinityCore/commit/47a4d58a982d5c22fbcdd07cac9dae5fd1bc52e4>

## Changes Proposed:
-  title
-  Also add some enums I found while searching TC
- Less magic numbers pog

## SOURCE:

check https://github.com/TrinityCore/TrinityCore/issues/26688 for the original author explanation

## Tests Performed:

- none

## How to Test the Changes:

1. I don't think this needs to be tested. The code kinds speaks for it self in this one...
2. Check that only spells with INVISIBILITY_GENERAL get glow vision

### DIFF
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -477,6 +477,49 @@ enum PlayerFlags
     PLAYER_FLAGS_UNK31             = 0x80000000,
 };
 
+enum PlayerBytesOffsets //@todo: Implement
+{
+    PLAYER_BYTES_OFFSET_SKIN_ID         = 0,
+    PLAYER_BYTES_OFFSET_FACE_ID         = 1,
+    PLAYER_BYTES_OFFSET_HAIR_STYLE_ID   = 2,
+    PLAYER_BYTES_OFFSET_HAIR_COLOR_ID   = 3
+};
+
+enum PlayerBytes2Offsets //@todo: Implement
+{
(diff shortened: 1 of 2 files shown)


stoudtlr:
I tested spells/potions referenced in the linked TC issue and in-game they all appear the same with the PR as they do without the PR.  Is that what's expected? Are there spells to test that should have invisibility without the glow effect?

Helias:
it's a refactoring, so the behavior of before should be the same now but with a better code quality, it's normal that nothing changed @stoudtlr  :wink:

Helias (approved):

meji46:
> I tested spells/potions referenced in the linked TC issue and in-game they all appear the same with the PR as they do without the PR. Is that what's expected? Are there spells to test that should have invisibility without the glow effect?

@stoudtlr Yes, it's expected. In general, the invisibility auras that the player can intentionally apply (items and class spells) are of type INVISIBILITY_GENERAL (so PLAYER_FIELD_BYTE2_INVISIBILITY_GLOW is still applied).

There are other types of invisibility auras (associated with quests or other mechanics) that were applying the screen effect when they shouldn't. For example: https://www.wowhead.com/spell=38544/coax-marmot

PR #7869, opened 2021-09-13
fix(Core/Unit): Improve check for players to automatically stand if a…

Kitzunu:
…ttacked

## Changes Proposed:
-  title  

## Tests Performed:

- none

## How to Test the Changes:

1. Go in combat with a player that is not standing


IntelligentQuantum (approved):

tkn963:
I can't seem to test this because I don't have any issues using the latest commit of the master branch. I tried sitting down as alliance and attacking as horde and I tried in duels but every time the player being attacked stands up. Melee attacks and spells that hit and that misses causes the player to stand up for me.

Kitzunu:
that's correct

Kitzunu:
if attacked, a player should stand up

tkn963:
> 
> 
> if attacked, a player should stand up

Yes, but they always do for me when not using this PR. Has this already, somehow, been merged into the master branch or are there specific circumstances I need to work with?

Kitzunu:
Ah no, it already works as expected, yes.

I just improved the logic to just check for players 🗡️ 

So if it works the same before the PR and after the PR then it is good =)

tkn963:
> 
> 
> Ah no, it already works as expected, yes.
> 
> I just improved the logic to just check for players 🗡️
> 
> So if it works the same before the PR and after the PR then it is good =)

Oh, alright. My mistake. It's currently compiling the PR so I'll make sure in a little bit then.

tkn963:
Works like it should 👍

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -12718,38 +12718,41 @@ void Unit::SetInCombatWith(Unit* enemy, uint32 duration)
     SetInCombatState(false, enemy, duration);
 }
 
-void Unit::CombatStart(Unit* target, bool initialAggro)
+void Unit::CombatStart(Unit* vicitm, bool initialAggro)
```
Efymer, line 12721: Shouldn't be `victim` instead of `vicitm`?

Efymer (changes requested):

src/server/game/Entities/Unit/Unit.cpp
Kitzunu: Ah yes typo, thanks

FrancescoBorzi:
@Kitzunu conflicts again

Efymer (approved):

Issue #5342, opened 2021-04-18
[Fleeing for Assist]: NPCs on Low HP need 1.666% reduced speed per 1% health below 30% HP [$30]


Stifler82:
**EDITED**: to add the outcome of the discussion here, provided by @T1ti 

So TLDR how fleeing works on classic : 

- Fleeing doesn't change the creature's speed at all, it should still have the same speed as if it was not fleeing

- **But currently AC is missing the low health speed reduction mechanic, which is about -2.5% reduced speed per 1% hp below 20%.** 
This is probably the msot important point

- If you FEAR a creature which doesn't have reduced move speed(above 20% hp) like with warlock's fear, the creature will run at full speed. 

- **as a recap, fleeing state never affects the creature's speed, creature's speed is only affected by external factors like the low health speed reduction mechanic, or other slwos applied by players**

- Creatures seem to run to other creatures while fleeing if they get close to them, but the original fleeing direction seems completely random(my personal analysis). When they do this, they stop at the creature and trigger a call for help, after a small delay the creatures will aggro and the fleeing creature will stop fleeing exactly at the same time the new creatures aggro.

- basically, fleeing movement is compeltely RNG, UNTIL they get close to another creature, then they should? run to it and do a call for help.(previous point)

-----------
Originally reported https://github.com/chromiecraft/chromiecraft/issues/183

##### ISSUE: FACTION SIDE:
BOTH

##### CONTENT PHASE:
ALL

##### SMALL DESCRIPTION:
Half the times, when you're fighting a humanoid mob that will flee at low health, once they reach the threshold, they will IMMEDIATELY take off at a very high speed, even faster than regular running, towards an enemy pack. Once they get anywhere near another mob, they will INSTANTLY pull the mob.

This is blatantly wrong, even by Classic standards.

This is really frustrating and punishing to deal with. Sure, you can say "just save your burst for when they get really low", but that's not a luxury that can always be afforded, especially by certain classes. And when that happens: death.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
https://www.youtube.com/watch?v=CuvYOo1jG1k

I don't think ALL the mobs are supposed to run at ludicrous speeds. Even for the ones that do, they seem way too fast.

What's 100% wrong is that the mobs pull way too quickly once they get to other mobs. If you're ranged, the fleeing mob can get into kissing range with the non-pulled mob, and you can still kill the fleeing mob without the other pulling. Here, not quite...

[...]

PR #7932, opened 2021-09-16
Fix issue 5342 mob speed health

patou01:
## Changes Proposed:
-  call Unit::UpdateSpeed() more often for NPCs
-  in UpdateSpeed(), reduce speed based on health

## Issues Addressed:

- Fixes part of  https://github.com/azerothcore/azerothcore-wotlk/issues/5342

## SOURCE:

See 5342

## Tests Performed:

- builds ok. Tested on my own machine (win10 x64)
- Get in game, find a mob, aggro, reduce hp and check his speed.
- video of my testing : https://drive.google.com/file/d/18Gq1EjtGGVFzkdCg9HNVSNZFDVLnTIuu/view?usp=sharing 
https://drive.google.com/file/d/1v8iruv8V3jqWyLwBR2HdT2ZysesTwRbO/view?usp=sharing 

## How to Test the Changes:

Use macro to read speed. The following one seems to work
/script DEFAULT_CHAT_FRAME:AddMessage(string.format("%f", GetUnitSpeed("target")), 1.0, 1.0, 1.0);
.modify speed 2.5 on yourself will help get away from mobs.
.damage X will be useful to get mobs to accurate hp.
1. Find any mob. Get in combat, have him run at normal speed towards you. Read speed, should be 8 for average mob
2. Damage him to 31%, make sure he's running with no obstacles. Read speed, should be same as above.
3. Damage to anything below 30%, mob starts slowing down. At 1% he should be at about 45% slow (speed of ~4.5 if base is 8)
4. Heal mob, you can use .aura 33763 to cast lifebloom rank 1. speed should increase back to 100% as life reaches > 30%.
5. Do same as above but cast a frostbolt or other slowing spell, slow should be more than before

Maybe review we didn't affect players or pets in any wrong way, but the code should be warded against that.

Second part of the bug:
1. Find a mob that runs away at 15% hp.
2. Get him to ~17%, read speed.
3. Drop him just below 15 so he runs away, read speed. Should be comparable (within 5%).
4. Regardless of the mob finding someone or not to assist, his speed shouldn't change.

No known issue, but I made a change down in movement manager (of something that should probably not have been there in the first place), could maybe cause odd things.
- [ ] 
- [ ]


Helias (changes requested):
remove the lines of code directly without comment them

UltraNix (changes requested):
I'm completely against these changes. `UpdateSpeed` in every `Update` tick in combat - no.

patou01:
Do we have a flag to signify health has changed then?

UltraNix:
I think you should moved that code from `Unit::Update` to `Unit::DealDamage` and check there if victim is below that 30% - tnen call update speed.

patou01:
I'll move it to there and figure out a way. Checking for HealthBelowPct(30) will probably cause issues because heals are instant. If a mob is at 20 and gets healed to 80% from an ally npc, checking for that will mean the slow sticks, and I don't think that's the way it's intended. Though I'll admit I don't remember. 

Which is why it should be "on health change". 

=> Proposal:
- Add a call to updateSpeed in DealDamage towards the end, with proper warding.
- Add a call to UpdateSpeed in HealBySpell warded by a check to previous health.
```
uint32 prevHealth = GetHealth()
// do the healing here
if (prevHealth < 30)
UpdateSpeed()
```

UltraNix:
Then move it to `Unit::SetHealth`

patou01:
Behaviour should be unchanged based on my tests:
https://drive.google.com/file/d/1ro-w2CNxg0FWqoXtmVxGdrrw5b8XOICR/view?usp=sharing
https://drive.google.com/file/d/1U8zg89EkeKV1ktDJiQINiO09E2Gti4sg/view?usp=sharing

This should be suitable now?

patou01:
Apologies for the spam. Should be okay now. As I had already implemented something else on top I had to undo some commits and didn't realize I forgot to remove part of it.
Updated videos, though there's no difference visible: https://drive.google.com/file/d/1zlW_IDMfXbDPL-5hh2gwDpyS2qI1Oysa/view?usp=sharing 
https://drive.google.com/file/d/1qRBHnpHihjmDWRQoKfehAaxMcwH87kW4/view?usp=sharing

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -14521,8 +14527,16 @@ void Unit::SetHealth(uint32 val)
             val = maxHealth;
     }
 
+    float prevHealthPct = GetHealthPct();
+
     SetUInt32Value(UNIT_FIELD_HEALTH, val);
 
+    // mobs that are now or were below 30% need to update their speed
+    if (GetTypeId() == TYPEID_UNIT && (prevHealthPct < 30.0 || HealthBelowPct(30)))
```
UltraNix, line 14535: You need here two checks:
1. if mob were below 30 and now is above it
2. if mob were above 30 and now is below it

patou01: I don't see why? The formula is continuous but breaks at 30. So unless we want to make it by steps, eg 30-25-20-15-10, I think this is correct?
If a mob was previously at 30, if he got lower than before I need to recompute. If he got higher, I need to recompute as well (my understanding is they gain speed back). Therefore `prevHealthPct < 30.0` 
If a mob is now below 30, he needs to be recomputed. So `HealthBelowPct(30)`.

And as it's both of these cases that need to re-evaluate, we use `||`. 

Did I misunderstand what we try to achieve?

UltraNix: Okay - you are right.

Nyeriah:
I don't think all creatures should be affected by this. I see that you're exempting pets from the slow, but what about dungeon & raid bosses? Vehicles? SoTA & Wintergrasp for example. I don't think this is correct. Personally, I never heard of this feature, but I see some extensive research has been done about it, so I could definitely be wrong, but I don't think it can be applied globally as it is.

patou01:
Dungeon mobs for sure. Bosses I'm not confident and maybe not. I'll look into it a bit more and add a few wards if needed. My feel is that pretty much everything slows down, but that's just my feeling of 5 years of playing.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13443,11 +13442,18 @@ void Unit::UpdateSpeed(UnitMoveType mtype, bool forced)
             break;
     }
 
-    // for creature case, we check explicit if mob searched for assistance
-    if (GetTypeId() == TYPEID_UNIT)
+    // added to fix azerothcore issue 5342
```
Kitzunu, line 13445: unnecessary comment.

patou01:
> Player controlled vehicles to be more precise. But also things like https://www.wowhead.com/npc=36609/valkyr-shadowguard or https://tbc.wowhead.com/npc=20064/thaladred-the-darkener and other creatures with specific mechanics that involve speed/movement.


I would then suggest we skip vehicles, WB and mobs immune to snares. I'm not that sure about worldbosses not having this, but I don't have any video where we can estimate the speed at which eg a nefarian crosses the room to go kill someone for a 1% wipe.

syssneck:
What about escort NPCs? Are they affected, and should they be affected?

patou01:
Hmm I don't know if there's a tag specifically for them, but don't they heal back up once they're out of combat anyway? Could also exclude questgivers but it's starting to sound dodgy.

Nyeriah:
> Edit: Should be noted however that affecting a few more mobs than strictly needed is probably not even half as bad as today's situation of having none.

I agree, but since we don't have enough data to guarantee that everything should be affected I'll advocate for a controlled approach. I think I would settle with player controlled vehicles, pets and dungeon bosses being exempt from being slowed. But having a flag_extra would be really awesome for any other cases that may seem to not be affected by this, like the val'kyrs for example, or any other creature that we don't have enough data to confirm.

T1ti:
> Dungeon mobs for sure. Bosses I'm not confident and maybe not. I'll look into it a bit more and add a few wards if needed. My feel is that pretty much everything slows down, but that's just my feeling of 5 years of playing.
> 
> @T1ti , as you had done most of the research, do you know if this applies to eg mechanical mobs, bosses etc?

I think a simple solution is not applying it to mobs that have slowing immunity, all bosses should have it.

patou01:
Warded as follows.
``` 
    Creature* creature = ToCreature();
    Unit* owner    = GetOwner();
    // ignore pets, player owned vehicles, and mobs immune to snare
    if (GetTypeId() == TYPEID_UNIT
        && !IsPet()
        && !(owner && owner->IsPlayer() && IsVehicle())
        && !(creature && creature->HasMechanicTemplateImmunity(MECHANIC_SNARE)))
    {
```

I'm not sure if the vehicle part is correct, as I have no experience with vehicles.

I tested on the beast and he wasn't slowed by speed, while a random trash mob next door was.

patou01:
Updated to match your proposal. It seems in other places in the code we do
```
GetTypeId() == TYPEID_UNIT && ToCreature()->
```
But I don't think there's a difference between that and assigning the pointer to a variable.

Nyeriah:
> Updated to match your proposal. It seems in other places in the code we do
> 
> ```
> GetTypeId() == TYPEID_UNIT && ToCreature()->
> ```
> 
> But I don't think there's a difference between that and assigning the pointer to a variable.

ToCreature() checks for TYPEID_UNIT before returning, so not point checking for it.

https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Object/Object.h#L287

Also, since you're using that variable anyway, may as well use it to check if we're dealing with a creature

Nyeriah:
This looks fine for me, could use some help testing @azerothcore/testers

patou01:
@azerothcore/testers , if someone could have a look here please? :)

temperrr:
Since nobody tested it again, and I heard its relatively small change, I revoke my original statement that it needs additional testing and fix the label accordinly.

Thanks again for this PR @patou01 !

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -13496,11 +13495,22 @@ void Unit::UpdateSpeed(UnitMoveType mtype, bool forced)
             break;
     }
 
-    // for creature case, we check explicit if mob searched for assistance
-    if (GetTypeId() == TYPEID_UNIT)
+    int32 slowFromHealth = 0;
+    Creature* creature = ToCreature();
+    // ignore pets, player owned vehicles, and mobs immune to snare
+    if (creature
```
UltraNix, line 13501: It should work for bosses too?

patou01: Is it a remark saying "it should work for bosses too", or is it "but this will slow down bosses"? Bosses should be covered by the `hasMechanicTemplateImmunity(MECHANIC_SNARE)` part. At least it seemed to work when I tested on some?

UltraNix: Okay. Not all bosses are immuned to snare effects.

patou01: Hmm. Then it needs discussion, but are you sure about that?
`SELECT * FROM `creature_template` WHERE `rank` = 3 AND NOT (`mechanic_immune_mask` & 1024) AND `mechanic_immune_mask`!= 0;`
Doesn't seem to return much, and most of those are likely wrong (ie they have 8 388 624 which matches only immune horror and immune fear. I doubt there's that many bosses one can sheep/sap/ban/freeze/stun/snare/slow/grip/charm.

Of course it could be that I have the SQL request wrong. I can make it more explicit if we want and add a check for `isWorldBoss` and/or `isDungeonBoss`

UltraNix: Hmm. Then I might be wrong about that. But in case that some bosses will be not immuned to snare, it would be better to use `isDungeonBoss` check.

patou01:
Removing the to be merged label until I've made that additional change for checking worldbosses. I guess it's fine if I test it myself after that.

T1ti:
if world bosses aren't immune to snare, then there's somethnig wrong with their data

src/server/game/Entities/Unit/Unit.cpp
patou01: Added now

patou01:
@ultranix are we okay to merge like this then? (assuming the ci builds)

patou01:
@ultranix @helias you both have requested changes that I think are adressed, could you check please? :)

UltraNix:
Looks okay.

patou01:
> 
> 
> Looks okay.

What I meant is 
image

UltraNix (approved):

Helias (approved):

Issue #7921, opened 2021-09-16
Various NPCs Incorrectly Dropping Black Lotuses


Azcobu:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1740

66 creatures drop black lotus

Tester: confirmed. For all but one of the NPCs the drop rate is 0.02% (the exception being Baron Kazum at 0.24%) but it shouldn't be any at all.

<details>
<summary>Full NPC List</summary>

UnknownTable
---
| entry | name | chance | 
| ---: | --- | ---: | 
| 7032 | Greater Obsidian Elemental | 0.02 | 
| 7033 | Firegut Ogre | 0.02 | 
| 7034 | Firegut Ogre Mage | 0.02 | 
| 7035 | Firegut Brute | 0.02 | 
| 7036 | Thaurissan Spy | 0.02 | 
| 7037 | Thaurissan Firewalker | 0.02 | 
| 7038 | Thaurissan Agent | 0.02 | 
| 7039 | War Reaver | 0.02 | 
| 7040 | Black Dragonspawn | 0.02 | 
| 7041 | Black Wyrmkin | 0.02 | 
| 7042 | Flamescale Dragonspawn | 0.02 | 
| 7043 | Flamescale Wyrmkin | 0.02 | 
| 7046 | Searscale Drake | 0.02 | 
| 7434 | Frostsaber Pride Watcher | 0.02 | 
| 7435 | Cobalt Wyrmkin | 0.02 | 
| 7436 | Cobalt Scalebane | 0.02 | 
| 7438 | Winterfall Ursa | 0.02 | 
| 7439 | Winterfall Shaman | 0.02 | 
| 7440 | Winterfall Den Watcher | 0.02 | 
| 7441 | Winterfall Totemic | 0.02 | 
| 7446 | Rabid Shardtooth | 0.02 | 
| 7452 | Crazed Owlbeast | 0.02 | 
| 7453 | Moontouched Owlbeast | 0.02 | 
| 7458 | Ice Thistle Yeti | 0.02 | 
| 7460 | Ice Thistle Patriarch | 0.02 | 
| 8523 | Scourge Soldier | 0.02 | 
[...]

PR #7948, opened 2021-09-18
fix(DB/creature_loot): Remove Black Lotus drops from NPCs

Azcobu:
## Changes Proposed:
Deletes [Black Lotus](https://tbc.wowhead.com/item=13468/black-lotus) (ID 13468) from any NPC drop table where it appears. This is because it shouldn't drop from any NPC (with the possible exception of Mushgog, who does not drop it on AC in any case) but is only acquired through picking it directly.

The check on the comment field in the SQL is strictly unnecessary, but it's better to be extra careful when running mass deletes against the entire loot table based on a single key.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7921
- Closes https://github.com/chromiecraft/chromiecraft/issues/1740

## SOURCE:

https://tbc.wowhead.com/item=13468/black-lotus#dropped-by
https://wowpedia.fandom.com/wiki/Black_Lotus?oldid=2379075

## Tests Performed:

- Ran SQL, no errors/warnings. 66 rows affected as expected.
- Ran following query to check:

```SQL
SELECT ct.entry, ct.name, clt.chance
FROM `creature_template` ct
JOIN `creature_loot_template` clt ON ct.lootid = clt.Entry
WHERE clt.Item = 13468
```
No results returned, as expected.

## How to Test the Changes:

1. Run the query above. No results should be returned.

N/A.


IntelligentQuantum (approved):

data/sql/updates/pending_db_world/rev_1631940505188780442.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1631940505188780442');
+
+-- Deletes Black Lotus from all NPC loot tables
+DELETE FROM `creature_loot_template` WHERE `item` = 13468 AND `comment` LIKE '%Black Lotus%';
```
Helias, line 4: what about adding and entry != 11447? 
https://tbc.wowhead.com/npc=11447/mushgog

Azcobu: That's because Mushgog doesn't drop Black Lotus on AC right now, so there's no need to exclude him.

Also, I have some doubts about the Wowhead page that says he does, as the arena he spawns in (in Dire Maul) also has a Black Lotus spawn point that he may be linked to. So it seems likely that the Wowhead client just got confused between someone picking the herb and looting his corpse and erroneously linked the two. But that's supposition on my part.

Helias (approved):

Issue #7776, opened 2021-09-08
[Blackrock Depths - Boss] Summoner's Tomb/ The Seven Event too fast


temperrr:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [x] 50-59

### Current Behaviour

When you get to Summoner's Tomb and when you pull the event. Bosses that should spawn every 45 seconds or so, but here they spawn every 20 sec or so, might be wrong, never counted. But everyone in multiple runs (over 20) stated that they spawn faster than expected leading to fight 3 bosses at once instead max 2.
https://classic.wowhead.com/npc=9039/doomrel#drops;mode:normal

### Expected Blizzlike Behaviour

https://www.youtube.com/watch?v=wW84027e2JU&ab_channel=Chou-chiChang

At lower levels in BRD you on Blizzlike WoW have time to kill 1 by 1 boss, while here bosses spawns so fast that you get 3 at some point even with BiS lvl 49 gear and so fast killing with WoTLK talents. Our spawn times are insane.

### Steps to reproduce the problem

1. Enter BRD and get to Summoner's Tob (Near the end of BRD full run)
2. Start the event by talking to one of bosses (straight forward when you enter door)
3. Bosses spawning so fast, literally 20 sec or 25, get 3 bosses at some point to handle.

### Extra Notes
Original Report: https://github.com/chromiecraft/chromiecraft/issues/1659
Triager notes:
https://www.youtube.com/watch?v=QoEwUz7Bdq8

24:41 Anger'rel 
25:18 Seeth'rel (killed before next one activated)
25:40 Dope'rel (killed before next one activated)
26:03 Gloom'rel (killed before next one activated)
26:26 Vile'rel (killed before next one activated)
26:51 Hate'rel (killed before next one activated)
27:12 Doom'rel 

Right now, they spawn in 14-20 seconds on CC.
They should either spawn on a timer, or when they are too quickly die, the next one should activate.
https://wowpedia.fandom.com/wiki/The_Seven
`All of the mobs in the encounter are immune to curses. Also the event is entirely on a timer. After a set amount of seconds, the next mob will release and aggro the first player it sees.`

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/595bb6adccbabc714469f3935541978283b8bdfb

[...]

PR #7952, opened 2021-09-18
fix(Core/BRD): timer for tomb of seven

patou01:
The timers were not correct. The first one took too long to engage.
The other ones spawned too quickly.

Closes azerothcore 7776
closes chromiecraft 1659

## Changes Proposed:
-  Add new define TIMER_TOMB_START to enable a different time before mobs spawn and between mobs. Set to 5s
-  Change define TIMER_TOMBOFTHESEVEN from 15000 to 30000.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7776
- Closes https://github.com/chromiecraft/chromiecraft/issues/1659

## SOURCE:

Source in issue, not a vanilla-era source however.

## Tests Performed:

-  Built and run the event in successful case.
- 

## How to Test the Changes:

1. .teleport blackrock
2. enter dungeon, take left ( .add 11000  to get the key). Then right to incendius, at incendius fly up (.gm fly) take the door that is north-west, walk down, take left where the ambassador flamelash is, then just follow the room to the other door. There might be a .teleport or .go but I couldn't figure it out.
3. Talk to Doom'rel to start the event. Try the event in the various ways it's supposed to be tested.

- [ ] 
- [ ] 


Helias (approved):

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockDepths/instance_blackrock_depths.cpp
```diff
@@ -8,7 +8,8 @@
 #include "InstanceScript.h"
 #include "ScriptMgr.h"
 
-#define TIMER_TOMBOFTHESEVEN    15000
+#define TIMER_TOMBOFTHESEVEN    30000
```
Helias, line 11: this `TIMER_TOMBOFTHESEVEN` is still used in the project? If not you can remove it

patou01: Yes, it's used line 606 in same file.
I had to split the timers into 2: one before the event (after having talked to the npc), and one in between the mobs. Before this it was 1 timer for all of it. So everything was on a 15s delay. 15s after talking to the npc the mob would attack, and then every 15s a mob. Now it's 5 and then every 30.
15s might have "still felt okay", but 30 is way too far into the "okay guys let's leave this is not happening" field that it had to be split. It's also more accurate.

PR #7996, opened 2021-09-20
fix(Core/Unit): Perform melee attack when unit has not enough power for melee spell

r-o-b-o-t-o:
## Changes Proposed:
- Perform a melee attack instead of doing nothing when trying to cast a melee spell which fails because the user doesn't have enough power. For example, if a warrior queues Heroic Strike but loses their rage before the Heroic Strike hits, a regular melee attack will be performed instead of skipping this swing completely.

## Issues Addressed:
No issue about this yet as far as I can tell.

## SOURCE:
Tested on classic TBC

## Tests Performed:
- Compiled C++ changes
- Tested on Windows 10 client and server

## How to Test the Changes:
1. Create a warrior
2. `.learn 772`
3. Attack an enemy
4. When you have 15 rage, queue Heroic Strike, but use Rend before HS is cast
5. Verify that when you have the error that says "Not enough rage", your character does a regular melee attack instead of waiting for the next swing

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -2211,10 +2211,35 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType, bool extr
     if (attType != BASE_ATTACK && attType != OFF_ATTACK)
         return;                                             // ignore ranged case
 
+    bool meleeAttack = true;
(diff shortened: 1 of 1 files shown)


Helias (approved):

FrancescoBorzi:
@azerothcore/testers can anyone test this please?

Issue #6276, opened 2021-06-09
Contagion of Rot spell self spreading and stacking


Stump876:
Original report from Chromiecraft [https://github.com/chromiecraft/chromiecraft/issues/811](https://github.com/chromiecraft/chromiecraft/issues/811)

#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [x] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Fetid corpse casts [Contagion of Rot](https://classic.wowhead.com/spell=7102/contagion-of-rot#used-by-npc) "5% chance to spread to allies" However, when infected it can spread to yourself with this version of [Contagion of Rot](https://www.wowhead.com/spell=7102/contagion-of-rot#used-by-npc) "10% chance to spread to allies" This version appears to stack on itself and the version cast by the Fetid Corpse. I'm not exactly sure which version is correct, or if they are both correct and this is how the spell acts. But the bug is the stacking and self spreading.
![image](https://user-images.githubusercontent.com/68164076/121285293-fbc98300-c892-11eb-81d8-98b0459697ed.png)
![image](https://user-images.githubusercontent.com/68164076/121285342-169bf780-c893-11eb-9574-9c9ecbe86f4e.png)
![image](https://user-images.githubusercontent.com/68164076/121285252-eb190d00-c892-11eb-94c6-9a8a2a019fd0.png)
Tested by attacking a Spider.
![image](https://user-images.githubusercontent.com/68164076/121287622-be66f480-c896-11eb-8d51-b9db99389961.png)
##### EXPECTED BLIZZLIKE BEHAVIOUR:

The disease should just refresh, as it does with the version casted by Fetid Corpse, not stack. And it should only spread to allies as the tooltip says, not to self.

##### STEPS TO REPRODUCE THE PROBLEM:

1. .go creature 6077
2. Aggro and wait for the disease to be applied.
3. go attack something, preferably something other than a fetid corpse to prove that it's proccing from you and not just being recast.
4. Wait for it to spread to you.

##### EXTRA NOTES:

NPC: Fetid Corpse (npc=1270)
SPELL: Contagion of Rot (spell=7102)

##### AC HASH/COMMIT:

[...]

PR #8021, opened 2021-09-21
fix(Script/Spells): Contagion of rot

jestermaniac:
## Changes Proposed:

sup

there's two spells

7102 -> casted by the mobs
7103 -> triggered by aura for 7102

7102 gets removed and applied errytime
7103 never gets removed and kept being applied by new mobs

had to implement a spell script for 7102 to remove 7102,7103 auras when applied and had to insert new conditions for 7103 to not spread to targets that have already 7102 or 7103.

decided to this way cuz conditions for 7103 make sense cuz u gotta spread only to specific targets (that got no aura) and made a script for 7102 to remove also 7103 when reapplied

gang

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6276
- Closes https://github.com/chromiecraft/chromiecraft/issues/811

## How to Test the Changes:

1. .go c 6077
2. attack lossa corpse there and u should always be havin only 1 debuff active 


Asthalor:
Tested all OK,.

Before the PR you could infect yourself with 7103 multiple times.
After testing its not possible, you can see the effect going on when attacking, but it wont apply to you.

However it will be applied to a friendly near, as expected.
If you keep attacking instead of applying another one it will be refreshed, being the 5% one or the 10% one.

Examples:

Before, lots of debuffs on yourself

acumulacion debuff

After

acumulacion despues

On this one we can see that i have the 5% one, and i applied to the other player, so she now have 10%. After fighting more mobs there when the debuff was applied instead of adding one more it refreshed, so all good.

Good job!

IntelligentQuantum (approved):

Issue #5199, opened 2021-04-08
Spawn at incorrect graveyard when dying in Frostmane Hold, Dun Morogh


nolt:
- [ ] enGB
- [X] enUS
- [ ] other (specify)

##### FACTION

- [X] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39

##### CURRENT BEHAVIOUR:

When you die in Frostmane Hold, you get ported to the graveyard in Coldridge Valley.

##### EXPECTED BLIZZLIKE BEHAVIOUR:

When you die in Frostmane Hold, you should be ported to the graveyard south of the Iceflow Lake.
Here is a wowhead comment from 3.0.9 that sais this:
https://www.wowhead.com/quest=287/frostmane-hold#comments:id=215696:reply=84315

##### STEPS TO REPRODUCE THE PROBLEM:

1. Die in Frostmane Hold, in DUn Morogh.
2. 
3. 
4.

##### EXTRA NOTES:

According to the wowhead comments, you should definetly not be ported to that graveyard.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/ee1918a3e9bc355fbd6f8d58d77760fb61fdbff7

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)
- [lua-CarbonCopy](https://github.com/55Honey/Acore_CarbonCopy)

##### OTHER CUSTOMIZATIONS:

None.

##### SERVER:

ChromieCraft

nolt:
Tested

When you .tele frostmane hold and .die, you will be teleported south of the Lake. But when you enter inside the cave and ,die you will be teleported to (screen attached).
[...]

PR #8022, opened 2021-09-21
fix(Core/Graveyards): Rework the graveyard selection logic to allow l…

Nyeriah:
…inking graveyards to areas, not just zones

## Changes Proposed:
-  Rework the graveyard selection logic to allow linking graveyards to areas as well, not only zones.
- Fully compatible with the entirety of the previously linked entries - if no linked graveyard is found to the player's current area, it'll just fallback to checking for zone as it would before.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5199

## SOURCE:

There's apparently some faulty logic in the graveyard selection code we have.

For example, it's worth noting that while the wiki mentions the entries from graveyard_zone are linked following the values from **AreaTable.dbc**, the code only looks up the player Zone, and searches for links according to that.

I searched about this for a bit and ended up finding this https://github.com/TrinityCore/TrinityCore/commit/9c220e8f55eaa8cfc93ed26ea7318430e5c6749b where TC allowed areas to be linked to graveyards in the early 6.x branch, but that code was never backported to 3.3.5.

While I'm unsure why that never happened, it seems clear to me that graveyards should optimally be linked by area and not simply by zone as it's done currently, which causes strange results like seen in the issue #5199, where it just abritrary chooses the closest graveyard regardless of it being "acessible" or not.

Furthermore, the video from classic TBC presented in that issue (https://youtu.be/wyaBOODuwqs) shows that there's clearly more to it than just a mere distance calculation, seeing how the player was sent to Kharanos and not to Coldridge Valley, as happens currently.

## Tests Performed:

- Builds successfully in Win11.
- Killed myself inside caves, inside instances, works fines.
- KIlled myself in Frostmane Hold and got ported to Kharanos.

## How to Test the Changes:

1. .tele frostmane hold
2. .die self
3. Release

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1632245299092353800.sql
+++ b/data/sql/updates/pending_db_world/rev_1632245299092353800.sql
@@ -1,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1632245299092353800');
+
+DELETE FROM `graveyard_zone` WHERE `ID` = 101 AND `GhostZone` = 135;
+INSERT INTO `graveyard_zone` (`ID`, `GhostZone`, `Faction`, `Comment`) VALUES
+(101, 135, 469, 'Frostmane Hold, Kharanos GY - Dun Morogh');
--- a/src/server/game/Misc/GameGraveyard.cpp
+++ b/src/server/game/Misc/GameGraveyard.cpp
@@ -82,14 +82,15 @@ GraveyardStruct const* Graveyard::GetDefaultGraveyard(TeamId teamId)
 
 GraveyardStruct const* Graveyard::GetClosestGraveyard(float x, float y, float z, uint32 MapId, TeamId teamId)
 {
-    // search for zone associated closest graveyard
+    uint32 areaId = sMapMgr->GetAreaId(PHASEMASK_NORMAL, MapId, x, y, z);
+
     uint32 zoneId = sMapMgr->GetZoneId(PHASEMASK_NORMAL, MapId, x, y, z);
 
-    if (!zoneId)
+    if (!zoneId && !areaId)
     {
         if (z > -500)
         {
-            LOG_ERROR("sql.sql", "ZoneId not found for map %u coords (%f, %f, %f)", MapId, x, y, z);
+            LOG_ERROR("sql.sql", "GetClosestGraveyard: unable to find zoneId and areaId for map %u coords (%f, %f, %f)", MapId, x, y, z);
             return GetDefaultGraveyard(teamId);
         }
     }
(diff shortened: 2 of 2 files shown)


jamadaha:
@Nyeriah Is this ready for review/test?

Nyeriah:
@jamadaha Yes, it can be tested/reviewed

jamadaha:
Also did a quick test, found that I spawned at the wrong graveyard:
image
Should spawn around the red square (Is closer graveyard)

Nyeriah:
@jamadaha I'll check what's wrong with the horde graveyard. As for the ally, aren't you supposed to spawn in Kharanos? It's what this video from retail shows: https://www.youtube.com/watch?v=wyaBOODuwqs&ab_channel=wowinsights

jamadaha:
Huh... Seems that is right then. Thought, you would spawn at the closer graveyard.

Nyeriah:
I solved the issue with horde characters. The issue was that with this commit we're adding a graveyard link to the Frostmane Hold area for allies, while expecting horde characters to just fall to the previous default. 

It just happens that I forgot to check the player's team, so in the Frostmane Hold case specifically, as a horde player, the core would find a graveyard assigned but wouldn't port you there because the faction check would fail. That issue is gone now.

FrancescoBorzi:
@jamadaha can you please give this another try?

jamadaha:
**Tests**
- [X] Tested dying as alliance in Frostmane Hold (Area in zone) | **Works as expected**
- [X] Tested dying as horde in Frostmane Hold (Area in zone) | **Works as expected**
- [X] Tested dying as alliance in Deadmines (Instance) | **Works as expected**
- [ ] Tested dying as horde in Deadmines (Instance) | **I think it works as expected**
image
- [ ] Tested dying as horde off coast (Coast) | **Not sure, doesn't seem like the closest graveyard**
image
- [ ] Tested dying as horde underwater (Underwater) | **This seems really far away**
image
- [ ] Tested dying as alliance underwater (Underwater) | **Better but still far off**
image

**I am whether it is working as expected**

Nyeriah:
What is expected to happen after this PR:
1) If we assign a graveyard to an area, the core will first check for that graveyard link before checking for any graveyards linked to the zone. For example:
 Frostmane Hold (Area) and Dun Morogh (Zone).
Player dies in Frostmane Hold (Area) -
-- Core finds a link --> Sends player to Kharanos (the graveyard we linked)

2) If no graveyards are assigned to the area (and after this PR there's only one graveyard assigned to any area still), then it'll fall back to how it behaved before (nothing changes at all).

This PR doesn't aim at solving calculations, that wasn't affected, but if we deem necessary it can be changed afterwards by linking graveyards to the specific areas where they feel wrong.

jamadaha:
Ah, okay. Tested the weird graveyard choices on the master branch, and those are the same as on this branch. So in that sense this PR works as expected.

Nyeriah:
Awesome, thank you very much for taking the time to test this! @jamadaha

FrancescoBorzi:
@jamadaha @Nyeriah thanks for your contributions, please don't forget to update the workflow labels (i.e. remove [Waiting to be tested] and apply [To Be Merged]) when you feel this has been tested enough.

DavuKnight:
Think this can be closed. 
Its been tested over at https://github.com/chromiecraft/chromiecraft/issues/370
which likely can be closed as well.

PR #8064, opened 2021-09-24
fix(Core/Commands): Partially fix an issue with mails sent from conso…

Nyeriah:
…le never showing up the 'Customer Support' sender in-game

## Changes Proposed:
-  Assign the target (receiver) as sender when mails are sent from console, so they properly display the "Customer Support" sender-tag in-game (same as with any other CS mail, like item recovery and such, that also use receiver as sender).

## SOURCE:

It's the same thing done by other support mail deliveries, so I don't see a reason why we shouldn't also do it here
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Player/Player.cpp#L11610

## Tests Performed:

- Builds successfully on Win11.
- Tested sending mails from console, works (but check the notes bellow on how to test it)

## How to Test the Changes:

1. .send items / .send mail from console, twice (once to load the mail list, the other to actually show the notice).
2. Proper notification from "Customer Support" will show

We have issues loading the mail packets that handle the notifications. 

The call that handles the notifications loads all of the character mails as well, clearing the cache that is then repopulated from database - in this line: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/MailHandler.cpp#L780

But by the time the iteration happens (to send the unread mails data), the mail cache isn't populated yet. https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Handlers/MailHandler.cpp#L790

So the first part of the packet, that tells there are unread mails, is sent, but the actual information isn't.

I'll submit another PR to address this, as I feel like it escapes the scope of this issue even if it's related to it somehow.

### DIFF
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -2680,8 +2680,10 @@ public:
         std::string subject = msgSubject;
         std::string text    = msgText;
 
-        // from console show not existed sender
-        MailSender sender(MAIL_NORMAL, handler->GetSession() ? handler->GetSession()->GetPlayer()->GetGUID().GetCounter() : 0, MAIL_STATIONERY_GM);
+        ObjectGuid::LowType senderGuid;
+        // If the message is sent from console, set it as sent by the target itself, like the other Customer Support mails.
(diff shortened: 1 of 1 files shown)


IntelligentQuantum:
Tested

IntelligentQuantum (approved):

PR #8138, opened 2021-09-28
fix(Core/Spells): Fixes charge ending on and/or facing away from the enemy

jamadaha:
## Changes Proposed:
-  Changes the path calculation for charge. Instead of making a path to a point near the creature, it now makes a path to the creature, but then reduces the length of the path. 
-  

## Tests Performed:

- Successful build
- Charged a bunch at multiple creatures: on hills (Works), underwater (Doesn't work, charges onto the creature), surface of water to underwater (Same as underwater), ground to underwater (Works)

## How to Test the Changes:

NOTE: First core PR, so please double check
1. Not quite sure how to test this one, other than charging a bunch of creatures
2.
3.

- Fix falling into the ground at the end of charge (Happens both before and after)
- Fix standing in the ground after charge (Not sure whether it happens before)

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -5734,15 +5734,8 @@ SpellCastResult Spell::CheckCast(bool strict)
                         if (!target)
                             return SPELL_FAILED_BAD_TARGETS;
 
-                        Position pos;
-                        target->GetChargeContactPoint(m_caster, pos.m_positionX, pos.m_positionY, pos.m_positionZ);
-
                         if (m_caster->GetMapId() == 618) // pussywizard: 618 Ring of Valor
-                            pos.m_positionZ = std::max(pos.m_positionZ, 28.28f);
-
-                        float maxdist = m_caster->GetMeleeRange(target);
-                        if (!target->IsInDist(&pos, maxdist))
-                            return SPELL_FAILED_NOPATH;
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -5734,15 +5734,8 @@ SpellCastResult Spell::CheckCast(bool strict)
...
-
                         if (m_caster->GetMapId() == 618) // pussywizard: 618 Ring of Valor
-                            pos.m_positionZ = std::max(pos.m_positionZ, 28.28f);
-
-                        float maxdist = m_caster->GetMeleeRange(target);
-                        if (!target->IsInDist(&pos, maxdist))
-                            return SPELL_FAILED_NOPATH;
+                            target->m_positionZ = std::max(target->m_positionZ, 28.28f);
 
                         if (m_caster->GetMapId() == 618) // pussywizard: 618 Ring of Valor
```
Goatform, line 5740: Move it under this check then, no need to have double `m_caster->GetMapId() == 618` checks

jamadaha: That was something I was looking at, but I would rather just remove these if possible. Looking at Trinity core, they don't have these checks.

Goatform: If those hackfixes are no longer needed then I'd say it is fine imo, but you would have to test it out. (Before and after)

Issue #8011, opened 2021-09-21
Spell: Frost Trap procs Lock and Load from creatures with snare immunity flag.


avarishd:
### Current Behaviour

When you place a Frost Trap under a creature (with immune to snare flag) it triggers and the caster gains Lock and Load buff.

### Expected Blizzlike Behaviour

When you place the Frost Trap under a creature (with immunity to snare flag) it should still trigger, but without giving the caster Lock and Load buff.

### Source

https://wowpedia.fandom.com/wiki/Lock_and_Load

Wrath of the Lich King Patch 3.2.0 (2009-08-04):
Now has a 22 second cooldown. The Lock and Load effect cannot be obtained on targets immune to snare effects when Frost Trap is used.

### Steps to reproduce the problem

1. Make a hunter
2. Put 3/3 points into Lock and Load 
3. Go to any creature with a snare immunity flag (any boss)
4. .cast 13809 under them

### Extra Notes

**This bug affects both bosses and non-boss creatures with the snare immunity flag.**

Ignore the immune text above them (its from entrapment talent).

Boss(immune to snare):
https://i.gyazo.com/f5292731ee82ceeffb2f1b75c9804cb7.mp4

Non-boss creature (immune to snare):
https://i.gyazo.com/7f677d9d1dbf04a54f977dcb8183ac26.mp4

The code that handles that part.
 https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Spells/Spell.cpp#L2109-L2116

### AC rev. hash/commit

AzerothCore rev. 5fba063fdf67

### Operating system

Windows 10, Docker

### Custom changes or Modules

mod-cfbg

PR #8177, opened 2021-09-30
fix(Core/Spells): Correct hunter's Lock and Load procs

Nyeriah:
- Move related code to spell scripts
- Fixes it proc'ing when the target is immune to snares
- Partial port of https://github.com/TrinityCore/TrinityCore/commit/746b838870e93d9843a089613ce7f1cd160b6ea5

## Changes Proposed:
- Move related code to spell scripts
- Fixes it proc'ing when the target is immune to snares
- Partial port of https://github.com/TrinityCore/TrinityCore/commit/746b838870e93d9843a089613ce7f1cd160b6ea5

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8011

## Tests Performed:

- Tested using immolation traps against immune & non immune targets: no procs happened, as expected.
- Tested using frost & freezing traps against immune & non immune targets: no procs happened when the targets were immune, otherwise the spell proc'ed normally, as expected.

## How to Test the Changes:

1. Create a hunter and learn the Lock and Load talent from the Survival spec tree
2. .npc add temp 9938 (Magmus, boss, immune) / npc add temp 26829 (Frenzied Worgen, random creature, no immunities)
3. Use the trap spells and lure them individually into the trap, see if the spell procs or not 

- [ ]
- [ ]


IntelligentQuantum:
Tested

IntelligentQuantum (approved):

Kitzunu (approved):

Issue #8186, opened 2021-09-30
[Dun Morogh] Copper Vein Spawned Underground


gurveleg:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original Report: https://github.com/chromiecraft/chromiecraft/issues/1904

There seems to be an invisible copper node in Dun Morogh between Gol'Bolar Quarry and Helm's Bed Lake (see screenshot).

At first I was looking for the node inside the cave - there wasn't any - and then checked outside, but there wasn't any either.

![Bildschirmfoto 2021-09-30 um 21 39 49](https://user-images.githubusercontent.com/91610030/135521074-94ce582e-167e-4644-8938-c6e6173bcf78.png)

Tester Confirmed:
![WoWScrnShot_100121_005133](https://user-images.githubusercontent.com/80769357/135540212-8b2f34d5-2361-4fbb-af71-9f2f8ec4e263.jpg)
Copper Vein GUID 73532
AzerothCore rev. 792b99c

### Expected Blizzlike Behaviour

Nodes should be spawned in reachable locations.

### Source

This is not controversial.

### Steps to reproduce the problem

Check the copper node between Gol'Bolar Quarry and Helm's Bed Lake in Dun Morogh.

Teleport to GUID 73532 - pan the camera around a bit so it zooms out

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b2f08934bcc60491f8fa8971f98837b2d48a489e

### Operating system

macOS Catalina + Wine

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[...]

PR #6647, opened 2021-06-30
fix(DB): Importing Copper nodes from VMangos

claudiodfc:
### Improved Copper Vein spawns and pooling for the following zones:
1. Duskwood
2. Westfall
3. Darkshore
4. Loch Modan
5. Wetlands
6. Hillsbrad
7. Silverpine Forest
8. Barrens
9. Thousand Needles
10. Mulgore
11. Durotar
12. Ashenvale
13. Desolace
14. Stonetalon Mountains
15. Redridge Mountains
16. Tirisfal Glades
17. Elwynn Forest
18. Dun Morogh

## Issues Addressed:

- Closes #6167
- Closes #6679

## Tests Performed:

Tested on Windows 10

## How to Test the Changes:

1. Roam around these zones trying to find unusual behaviors on nodes.
2. Make sure The Draenei and Blood elf zones still have their Copper veins

## TODO
- [X] Still need to re-insert the TBC Copper nodes after being deleted

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #8188, opened 2021-10-01
fix(DB/GO): Repositioning a Copper Vein

syssneck:
## Changes Proposed:
-  Slightly changes the position of a Copper Vein that was unreachable

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8186

## Tests Performed:

- Ran SQL, no errors
- checked in-game
- is reachable now

## How to Test the Changes:

1. `.go o 73532`
2. if not spawned, check with `.gobject near 1` that it is there

- [ ]
- [ ]


Hacki95:
Where did you get the coordinates from?

I really dislike that we merge more and more things which have been changed by "feeling" when we could use the classic sniff database from brotalnia.

syssneck:
It's kind of a temporary fix to make it work for now (with the closest position I found in-game that is reachable).
It would be nice, if we could work on this PR: https://github.com/azerothcore/azerothcore-wotlk/pull/6647 so these temporary fixes wouldn't be necessary anymore, but apparently no one is really interested in it.

Hacki95:
I started the work on the vmangos port months ago but had no time to finish it properly. Claudiofc then started the same task and also stopped.

Problem with the temp fixing is, in the end nobody will remember which ones are temp fixes and which ones aren't. The ores and herbs are "pretty easy" to reimport but there also have been many changes to NPC spawns etc.

Ofcourse classic sniff data is imcomplete and also might be incorrect but its definitely better than randomly changing cords.

Maybe we can get a freelance dev to reimport the spawns.

syssneck:
I completely agree with you, and I will abstain from creating more temporary fixes.
My favorite method would be to import from Classic Sniff what works (Temper tested this a lot) and then fix the rest with sources we find or at least make the ores/herbs reachable.

syssneck:
Lmao I found the position of the Copper Vein in Classic sniff:
grafik
and this is basically what we had before, the Ore is stuck behind the wall and not reachable.

Hacki95:
Might still be blizzlike. In the end blizzard had some wrongly placed Ores as well. 

But to really confirm it somebody would need to check on classic.

Kitzunu:
Thanks for the PR and anyone that tested =)

PR #8189, opened 2021-10-01
refactor(Core/CreatureAI): Turn hardcoded entries into flagsExtra

Kitzunu:
## Changes Proposed:
- Turn hardcoded stuff into DB
-  Introdoce CREATURE_FLAG_EXTRA_HARD_RESET
- implement HasFlagsExtra 

## Tests Performed:

- None

## How to Test the Changes:

1. See if these bosses still despawn at evade

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -185,38 +185,9 @@ void CreatureAI::EnterEvadeMode()
         }
     }
 
-    // @todo: Turn into a flags_extra in creature_template
     // despawn bosses at reset - only verified tbc/woltk bosses with this reset type - add bosses in last line respectively (dungeon/raid) and increase array limit
-    static constexpr std::array<uint32, 24> bosses = {
-        /* dungeons */
-        28684, /* Krik'thir the Gatewatcher */
-        36502, /* Devourer of Souls */
-        36658, /* Scourgelord Tyrannus */
-        /* raids */
-        32871, /* Algalon */
-        39863, /* Halion */
-        33186, /* Razorscale */
-        36626, /* Festergut */
-        32867, /* Steelbreaker - Assembly of Iron */
-        32927, /* Runemaster Molgeim - Assembly of Iron */
-        32857, /* Stormcaller Brundir - Assembly of Iron */
-        33350, /* Mimiron */
-        16060, /* Gothik the Harvester */
-        36678, /* Professor Putricide */
-        15990, /* Kel'Thuzad */
-        33993, /* Emalon the Storm Watcher */
-        17257, /* Magtheridon */
-        25315, /* Kil'jaeden */
-        15928, /* Thaddius */
-        32930, /* Kologarn */
-        32906, /* Freya */
-        36597, /* The Lich King */
-        36853, /* Sindragosa */
-        36855, /* Lady Deathwhisper */
-        37955  /* Blood-Queen Lana'thel */
-    };
-
-    if (std::find(std::begin(bosses), std::end(bosses), me->GetEntry()) != std::end(bosses))
+    CreatureTemplate const* cInfo = sObjectMgr->GetCreatureTemplate(me);
+    if (cInfo && cInfo->HasFlagsExtra(CREATURE_FLAG_EXTRA_HARD_RESET))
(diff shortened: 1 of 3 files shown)


Nyeriah (approved):

Winfidonarleyan (approved):

src/server/game/AI/CreatureAI.cpp
```diff
@@ -185,38 +185,9 @@ void CreatureAI::EnterEvadeMode()
...
-        32930, /* Kologarn */
-        32906, /* Freya */
-        36597, /* The Lich King */
-        36853, /* Sindragosa */
-        36855, /* Lady Deathwhisper */
-        37955  /* Blood-Queen Lana'thel */
-    };
-
-    if (std::find(std::begin(bosses), std::end(bosses), me->GetEntry()) != std::end(bosses))
+    CreatureTemplate const* cInfo = sObjectMgr->GetCreatureTemplate(me);
```
Winfidonarleyan, line 189: Need creature id

Winfidonarleyan:
Hah, no create, need number creature :D

Kitzunu:
I know x) Trying to figure out the right way lol

Winfidonarleyan:
> I know x) Trying to figure out the right way lol

https://github.com/azerothcore/azerothcore-wotlk/blob/5908f1f2bf78bc6b8964cde2561d36364162ccac/src/server/game/Entities/Object/Object.h#L101

No this?

src/server/game/AI/CreatureAI.cpp
```diff
@@ -185,38 +185,10 @@ void CreatureAI::EnterEvadeMode()
...
-        36597, /* The Lich King */
-        36853, /* Sindragosa */
-        36855, /* Lady Deathwhisper */
-        37955  /* Blood-Queen Lana'thel */
-    };
-
-    if (std::find(std::begin(bosses), std::end(bosses), me->GetEntry()) != std::end(bosses))
+    Creature* creature = GetCreature(me);
+    CreatureTemplate const* cInfo = sObjectMgr->GetCreatureTemplate(creature);
+    if (cInfo && cInfo->HasFlagsExtra(CREATURE_FLAG_EXTRA_HARD_RESET))
```
Nyeriah, line 191: Suggestion: why don’t you add the helper to Creature instead of the template? Then you can just do 

me->HasFlagsExtra()

Winfidonarleyan: Maybe better, hmm

Kitzunu: Because flags_extra (flags_extra & flag) only exist in CreatureTemplate and I don't know how I would export that to creature

Nefertumm (approved):

src/server/game/AI/CreatureAI.cpp
Nyeriah: Check those https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Creature/Creature.h#L74-L76 🙂

Kitzunu: I'll look at it tomorrow

Kitzunu: But perhaps it is also good to keep the helper in template and not remove it completely?

Nyeriah: I think it's up to you, wouldn't hurt. But it seems to me that retrieving the creature template is uncommon, mostly used in abstract cases where we don't have a creature yet (e.g loading). People will try to create the creature to access the template when they need it, I think, which isn't optimal as we could just access it directly from the creature.

But if you decide to keep it in the CreatureTemplate, I could also suggest trying:
```cpp
me->GetCreatureTemplate()->HasFlagsExtra(CREATURE_FLAG_EXTRA_HARD_RESET)
```

Should also work. CreatureTemplate can't be null here - creatures can't exist without a template, and we are using an already spawned creature, so it will never be null in this case

FrancescoBorzi:
can anyone test this in game please?

Nyeriah:
No wiki reference about this :(

Kitzunu:
Good catch

PR #8224, opened 2021-10-02
Fix(Core/Map): Fix encounters list for heroic difficulties

Nefertumm:
## Changes Proposed:
-  Brief explaining... There's no EncounterList for heroic raid difficulties like ICC or Ruby Sanctum, so the hook never gets executed in those difficulties.  It's not a fancy way to fix it, but we can't just add encounter lists to DB. I have encounter this bug while I was trying to use the OnAfterUpdateEncounterState hook for a module. If anyone has another solution for this, please let me know and I'll do it ASAP!

## Issues Addressed:

- None

## Tests Performed:

- Built and tested in-game with multiple instances runs with different characters.

## How to Test the Changes:

1. Create a module with a simple OnAfterUpdateEncounterState hook, i.e. a simple cout to console with some msg.
2. Look at the differences before and after the pr. Now it should be executed when you do an heroic raid like ICC or RS.

- [ ]
- [ ]


Nyeriah:
Perhaps this has something to do with the fact ICC doesn't have a heroic mode? Raid difficulty was supposed to be dynamic (per boss basis), with both "difficulties" sharing the same lockout. But we don't handle that properly yet.

Nefertumm:
Mmm but ICC and RS both have heroic mode, 10H and 25H. Also TOC has heroic mode and it has its proper encounter list for all difficulties.

qwe900:
> 
> 
> Mmm but ICC and RS both have heroic mode, 10H and 25H. Also TOC has heroic mode and it has its proper encounter lists for all difficulties. You're right about the dyn diff

icc 10 and 25 has heroic mode in 3.3.5, aftter cata prepatch it got removed and boss based

Nyeriah (approved):

Issue #611, opened 2017-09-07
Core/Script: Upper Blackrock Spire Father Flame event does not start.


StygianTheBest:
**Description**:  Looting Father Flame from the obelisk does not start the event to fight  Solakar Flamewreath.

**Expected**:
> In order to start the event to summon Solakar Flamewreath you will need to loot the Father Flame on the western side of the room. After you loot the Father Flame it will start an event and waves of Dragonkin will come out at you. After you fight off all the waves Solaka Flamewreath will yell out then come running.

**Branch:**
Master

**Commit Hash:** 
c15f15ebb0a113a5cf20fc11afb89b2bcff9ec6d

gtao725:
The issue still exists as of commit hash 7eb703e. almost 2 years lol...

Pondaveia:
@gtao725 It ain't forgotten, I'll give an update soon.

Zoidwaffle:
Maybe some inspiration/info can be here: https://github.com/TrinityCore/TrinityCore/issues/19370 but it's not something that can, currently, be used directly.

Shineslippers:
Some vids
https://youtu.be/usD4PsDX1N4?t=603
https://youtu.be/srDqhx5oGF0?t=923

Wowhead [comment](https://www.wowhead.com/npc=10264/solakar-flamewreath#comments:id=1091101)
>By Kasa24 on 2010/07/29 (Patch 3.3.5)
Walk in and head to the left - it doesn't matter whether you kill the two guards or not - and you will see a glowing crystal near the wall. Click it as to loot, and you can get Father Flame. By right clicking it you'll see a yell, and two dragonkin will come to attack you. There are 6 waves of two dragonkin each, with about 30 seconds of downtime between them (as I did this on my 80 pally not sure if its 30 seconds from the time you kill them or the time that they spawn, since I killed them almost instantly). After the 6th wave Solakar will yell at you and spawn.

PR #8236, opened 2021-10-03
Fix(core/script): UBRS Solakar / father flame 

patou01:
**Latest commit should be kind of okay, but still very "private server obvious".** 

## Changes Proposed:
-  Add script for solakar and Hatchers (too hard to make in SAI)
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/611

## Tests Performed:

- basic run through event
- 

## How to Test the Changes:

1. Pick up father flame from object in UBRS, verify event runs properly and resets correctly when needed.
2.
3.

- [ ] Spawn times
- [ ] Spells for Hatchers? I believe they're supposed to cast stuff other than just slap around.


patou01:
The eggs seem to be a bit of a mess. The database has this
image

It seems that deleting the bottom one doesn't influence the behaviour in game...

temperrr:
Tested,

emotes work
resetting works
After killing drake, rend nowhere to be found
event doesnt reset anymore.
image

@Nyeriah

patou01:
@temperrr I was very confused. My response was about to be "thanks for testing this WIP test, I'm unsure how I affected the Rend event, but good to know I'll look into it". Guess I will still make sure I won't break rend :D

patou01:
Help needed:

- Number of waves and what kind of waves, should they be random, 1 of each mob, can they be the same?
- spell timers for the boss

patou01:
@ultranix, remaining questions/remarks based on your review from a while ago:

- `me->GetMotionMaster()->MovePoint(0, me->GetVictim()->GetPosition());`, to me it makes no sense to define ZERO = 0 here, so just leaving it as is. Ideally I'd want to not use MovePoint() at all, but it seems the mob doesn't want to start moving if I just use some attack function after having called `StopMovingOnCurrentPos`.  Do you have an alternative?
- I replaced `SelectNearestTarget` with doZoneInCombat. We'll see how that goes once the other PR goes in.
- I believe the positions are *not* "magic numbers", they're just constants that I cannot define as constant. They're defined at the top like anybody would expect, I think they're okay.

Other than that, conflicts removed now, the CI build should be okay.

Nefertumm:
It's looking good now! Sorry for the no-response, I didn't get the github notification :(

Nefertumm (approved):

temperrr:
Tested;

works as expected and close enough to blizzlike as we can from videos.

Issue #8241, opened 2021-10-03
[Dungeon Finder] Being able to kick someone within the first few seconds of the instance


syssneck:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/1907

After joining a group with dungeon finder (in my case, a group of 2 queued up), we could kick as soon as we've got inside the dungeon with the help of the dungeon finder.
![WoWScrnShot_093021_182320](https://user-images.githubusercontent.com/22718174/135539260-8724592c-8d6c-41f0-b4c4-f99ac3ea7384.jpg)

### Expected Blizzlike Behaviour

> A vote to kick a member may not be initiated in the first 15 minutes of a group.

I think I should get an error message when I am attempting to open the window asking why am I trying to vote kick a player.

### Source

On October 31th 2010 https://wowpedia.fandom.com/wiki/Vote_Kick?oldid=2442479

### Steps to reproduce the problem

Queue up with the dungeon finder.
Get inside the dungeon with the dungeon finder.
Attempt to open the window asking the reason why you wish to kick someone.

### Extra Notes

I do think the duration of the timer should be a configurable option.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b2f08934bcc60491f8fa8971f98837b2d48a489e

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[...]

PR #8243, opened 2021-10-03
fix(Core/LFG): Implement ERR_PARTY_LFG_BOOT_NOT_ELIGIBLE_S and lower …

Nyeriah:
…the max number of kicks from 3 to 2

## Changes Proposed:
-  Check if the player has the dungeon cooldown area to allow kicking (gets applied whenever a new group has just been formed, dungeon deserter also checks for this)
-  Lower max kicks allowed from 3 to 2 (maybe put this into a server config in the future?)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8241

## SOURCE:

https://wowpedia.fandom.com/wiki/Vote_Kick?oldid=2442479

## Tests Performed:

- Tested kicking a player right after joining a LFG dungeon group

Skärmbild 2021-10-03 154859

## How to Test the Changes:

1. Log 5 characters
2. Queue lfg
3. Try to initiate vote kick

**IMPORTANT: .debug lfg sadly can't be used, it doesn't apply  the dungeon cooldown aura we use to check if we can kick people.**

### DIFF
--- a/src/server/game/DungeonFinding/LFGGroupData.h
+++ b/src/server/game/DungeonFinding/LFGGroupData.h
@@ -25,7 +25,7 @@ namespace lfg
 
     enum LfgGroupEnum
     {
-        LFG_GROUP_MAX_KICKS                           = 3,
+        LFG_GROUP_MAX_KICKS                           = 2,
     };
 
     /**
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -12117,7 +12117,7 @@ Player* Player::GetNextRandomRaidMember(float radius)
     return nearMembers[randTarget];
 }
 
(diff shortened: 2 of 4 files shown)


Hacki95:
Very nice PR!

Nefertumm (approved):

syssneck:
Tested. I can't vote kick anybody for the first 15 min when a new group was formed through the dungeon finder. The only exception is an offline player, who can be kicked immediately after they went offline (dc, crash or log out).

I noticed one small thing (I don't know how it was before the PR): After I finished the dungeon, I was still not able to kick anybody and I got the following message:
grafik

Nyeriah:
> likeblizzard: max kicks =3

Hey,

The wiki says there's a maximum of 2 kicks and that information was not contested. However, I plan to move this to world settings so users may do what suits them best.

PR #8290, opened 2021-10-05
fix(CI): Don't run builds when PRs are labeled

Nyeriah:
## Changes Proposed:
-  Don't run builds when PRs are labeled
--- There's no point in running new builds whenever a new label is added, builds are already fired when new commits are pushed, and we are always updating the branch before merging anyway, so firing new builds whenever a label is added just creates a resource bottleneck
--- I suppose this also didn't account that we would ever have over 100+ labels added in a single day, there's probably a few days of builds queued by now that end up failing anyway.

Skärmbild 2021-10-05 174824

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -4,7 +4,6 @@ on:
     branches:
       - 'master'
   pull_request:
-    types: ['labeled']
 
 concurrency:
   group: ${{ github.head_ref }} || concat(${{ github.ref }}, ${{ github.workflow }})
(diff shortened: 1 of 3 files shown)


Nefertumm (approved):

Nyeriah:
I know we normally wait a few days but I'll go ahead and merge this so we can start working on labeling the PRs without the CI getting stuck again.

Nefertumm:
I agree. 👍

PR #8338, opened 2021-10-07
fix(DB/Quest): When Smokey sings, I get violent

Nyeriah:
- Port conditions/SAI from TC
- Explosion effect works, gameobjects despawn

## Changes Proposed:
- Port conditions/SAI from TC
- Explosion effect works, gameobjects despawn  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8297
- Closes https://github.com/chromiecraft/chromiecraft/issues/1992

## SOURCE:

TrinityCore

## Tests Performed:

- Tested in-game
- 

## How to Test the Changes:

1. go gob 45698
2. quest add 6041
3.Use item provided near the marks

## Known Issues and TODO List:

- [ ] CRITICAL - Gameobjects never respawn, SAI needs to be updated (https://github.com/azerothcore/azerothcore-wotlk/pull/8340)

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #8340, opened 2021-10-08
fix(Core/SAI): Update SMART_ACTION_FORCE_DESPAWN to allow temporarily despawning gameobjects

Nyeriah:
## Changes Proposed:
-  Currently gameobjects are **deleted** from world and never respawn when that action is called.
-  Creature/GO respawn implementation is precarious - the object handles it's own respawn timer, each instance is permanently spawned and the spawn/respawn process is merely fictional (the same instance of the creature/go is spawned at all times, "respawning" merely resets its variables).
- If we delete an object from world (as the current implementation does), it simply never respawns

## Issues Addressed:

- Closes 

## SOURCE:

Partially ported from TrinityCore, adapted to our current conditions (no dynamic spawns)

## Tests Performed:

- Tested the quest from here: https://github.com/azerothcore/azerothcore-wotlk/pull/8338
- 

## How to Test the Changes:

Easiest way: apply the changes from here https://github.com/azerothcore/azerothcore-wotlk/pull/8338

go gob 45698
quest add 6041
3.Use item provided near the marks
4. Notice that the marks despawn and respawn after 3 minutes

This will require some testing to ensure there's no adverse effects to the preexisting SAI scripts.
```sql
SELECT * FROM smart_scripts WHERE action_type = 41;
```
But I really doubt there are many cases where the despawn was meant to be permanent, because that basically would imply those scripts would only be able to be executed once per server restart...

**IMPORTANT: this only affects SAI action 41 when GAMEOBJECTS are despawned! No changes made to NPCs being force-despawned.**


Asthalor:
Tested, all good

Applied this changes and #8338. Went to the place and bomb it and it despawed. After 3 minutes it respawned again.

Video
https://streamable.com/3ql84w

After 3 minutes

image

Good job!

Kitzunu:
Ty for test and PR

PR #8377, opened 2021-10-09
feat(Core/Spells): implemented two new custom attributes to handle aura saving rule

sanctum32:
In this PR i implemented two new custom spell attributes and fixed implicit conversation in generic SpellInfo attributes checks

## Changes Proposed:
-  Implemented initial attributes to handle aura saving (SPELL_ATTR0_CU_REJECT_AURA_SAVING, SPELL_ATTR0_CU_FORCE_AURA_SAVING)
-  Applied some code style to old code
-  Added mask stacking rules for SPELL_ATTR0_CU_NEGATIVE_EFF and SPELL_ATTR0_CU_POSITIVE_EFF
-  If Error happens on certain attributes, only attribute which errored will get skipped. This allows other flags to continue loading
-  Raise Ally control aura saving handling been moved to DB and applied SPELL_ATTR0_CU_REJECT_AURA_SAVING to this spell
## Issues Addressed:

- Closes 

## Tests Performed:

- Tested both attributes with "Retribution Aura" (rank 1 but rest of aura ranks should work same too)
- Tested both attributes with "772 Rend" (rank 1 but rest of aura ranks should work same too)

## How to Test the Changes:

Add any aura spell id with one of these attributes and check if aura gets saved or non saved (noting, SPELL_ATTR0_CU_FORCE_AURA_SAVING and SPELL_ATTR0_CU_REJECT_AURA_SAVING doesn't stack, you can pick only one)

- [ ]
- [ ]


Kitzunu:
What is aura saving?

sanctum32:
@Kitzunu there can be situations where certain aura maybe should be saved but generic check indicates that it can't be saved to characters db. Same happens with spells which maybe shouldn;t saved but same system tells that it can be saved. These attributes can make exceptional forced rules for such behaviors without marking certain spells as negative and similar

sanctum32:
Yet if new collumn is not necessary, i can remake this

FrancescoBorzi:
> I want to preserve possible ports from trinitycore if something may will get ported.

sounds good to me, @azerothcore/testers can anyone double-check this please?

Nyeriah:
The code is good, but I agree with @Kitzunu. While the new column may make it easier for future ports, it breaks normalization as the rows we got there will be filled of lots of 0s and it seems that will be the case most of the time. It's something we tried to avoid countless times @ TC (take for example the several tables that were normalized, creature/creature_addon, quest_template/addon, etc.). 

This is pure tech speech tho, _**in practice**_, the new column would be better as that allows developers to create custom flags as they wish without ever worring about override TC flags.

A solution that doesn't break normalization? We could use an entirely new table `spell_custom_attr_ac`, that way we part away from TrinityCore's implementation while also keeping useless data off the DB.

I found this while looking around: https://github.com/TrinityCore/TrinityCore/commit/b4aa698acb88e1a74bc22d97279148a567b88cad

sanctum32:
@Nyeriah whole spell_custom_attr is more as data holder for optional data, we can always contain null entries there to avoid 0s... either tho, if these 0s are there and they aren't loaded to relative data, it doesn't consumes too much resources (except size on storage). Also we always could use nulls instead. Either tho, duplicating tables would mean to make more trash in database structure (ac structure already is trashed by useless *dbc data)... but yet i would rather add ac prefix into custom attribute collumn

sanctum32:
@Nyeriah i totally wasn't aware that this been added in trinitycore master branch... alright i will update this PR then

src/server/game/Spells/SpellInfo.h
```diff
@@ -323,6 +324,7 @@ class SpellInfo
     uint32 AttributesEx6;
     uint32 AttributesEx7;
     uint32 AttributesCu;
+    uint32 AttributesCu1;
```
Kitzunu, line 327: not needed

src/server/game/Spells/SpellInfo.h
```diff
@@ -201,6 +201,7 @@ enum SpellCustomAttributes
     SPELL_ATTR0_CU_POSITIVE_EFF0                 = 0x02000000,
     SPELL_ATTR0_CU_POSITIVE_EFF1                 = 0x04000000,
     SPELL_ATTR0_CU_POSITIVE_EFF2                 = 0x08000000,
+    SPELL_ATTR0_CU_REJECT_AURA_SAVING            = 0x10000000,
```
Kitzunu, line 204: I would prefer to have them together. So pls move SPELL_ATTR0_CU_FORCE_AURA_SAVING to be at 0x20000000

Kitzunu (changes requested):

src/server/game/Spells/SpellInfo.h
UltraNix: Not needed - not used in core.

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -783,6 +783,7 @@ SpellInfo::SpellInfo(SpellEntry const* spellEntry)
     AttributesEx6 = spellEntry->AttributesEx6;
     AttributesEx7 = spellEntry->AttributesEx7;
     AttributesCu = 0;
+    AttributesCu1 = 0;
```
UltraNix, line 786: Not need  - not used by core.

sanctum32: removed

src/server/game/Spells/SpellInfo.h
sanctum32: Removed

src/server/game/Spells/SpellInfo.h
sanctum32: Applied

Issue #8397, opened 2021-10-10
[Scholomance] Kirtonos floating mid-air


patou01:
### Current Behaviour

![image](https://user-images.githubusercontent.com/2592673/136685682-c5aa0068-a3e7-494b-bffb-5b15e497867b.png)

Probably missing a "setFlight" or something on phase change.

### Expected Blizzlike Behaviour

Undeads don't fly, gargoyles do.

### Steps to reproduce the problem

Go to kirtonos, summon him, wait for human phase.

### AC rev. hash/commit

198552766ddbbccdb8232e0575693ef8306d29ac

### Operating system

win10 x64

Issue #8386, opened 2021-10-09
Kirtonos the Herald can be summoned twice


jamadaha:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/2041

When using the blood of innocents on Kirtonos' brazier in Scholomance, you can summon him twice by having two people use the brazier at once. This has to happen before the brazier is lit, so two players will have to use it at the same time. 

This makes two Kirtonos appear in the sky, and you have to fight them both. They both drop items as well.

![WoWScrnShot_100921_174508](https://user-images.githubusercontent.com/90989261/136666939-e2559933-4a6c-4e6d-b104-bfeb72ec3589.jpg)
![WoWScrnShot_100921_174507](https://user-images.githubusercontent.com/90989261/136666940-501a22a9-b7d3-4f44-b8bc-f037aaf38eb4.jpg)
 
Testers note:

![image](https://user-images.githubusercontent.com/53914190/136670110-b16ce37c-9f9e-48b8-8cd8-d3a7389e712d.png)
![image](https://user-images.githubusercontent.com/53914190/136670155-b7a01894-f964-4713-91de-92ee65bc8eda.png)

### Expected Blizzlike Behaviour

Only one Kirtonos is supposed to appear, regardless of how many people use the brazier. After the brazier is lit, it i

### Source

N/A, can't find that there is any specific discourse on this, but I know this is not what is supposed to happen, having played the dungeon quite a few times in classic and other versions. 

### Steps to reproduce the problem

1. Enter the Scholomance Dungeon
2. Make your way to Kirtonos area/brazier
3. Have two (or more?) players use the brazier with the blood of innocents at the same time.
4. Now, more than one Kirtonos will appear and fight the party.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b2f08934bcc60491f8fa8971f98837b2d48a489e

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
[...]

PR #8401, opened 2021-10-10
Fix(Script/scholo): Improvements for Kirtonos

patou01:
## Changes Proposed:
- Removed most SAI from brazier, it now only sets the instance data
- Event starts at the end of the gameobject use, not at the start. ie he spawns when the brazier is lit
- Gate and brazier control is moved from kirtonos.cpp to scholomance.cpp
- Retry is now possible after a wipe, granted one has enough blood.
- Kirtonos doesn't swim in the air in humanoid form anymore

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8397
- closes https://github.com/azerothcore/azerothcore-wotlk/issues/8386
- closes https://github.com/azerothcore/azerothcore-wotlk/issues/8852 
- closes https://github.com/chromiecraft/chromiecraft/issues/2221
- closes https://github.com/chromiecraft/chromiecraft/issues/2041

## Tests Performed:

- tested on my own machine. https://drive.google.com/file/d/1h2Sc_UCuLoGhe9s-WOogzoRG_WeluJnq/view?usp=sharing
- tested fail event (enabling GM mode), door opens and brazier reset
- tested 2 players clicking at the same time, only one spawns and the last clicker gets "invalid target" upon completion, which afaik is blizzlike
- tested killing the boss, door opens brazier doesn't reset

## How to Test the Changes:

1. Go to scholo at kirtonos spawn, use blood of innocent (.add 13523)
2. Test the event in usual ways: wipe, boss kill, etc (note that if the boss is killed and you respawn him, the event probably wont work anymore)
3. Test what happens if you get thrown out the window too, should just cause a reset but good to be sure.

- [ ] Might have broken the "load after crash" feature.
- [ ] I broke the flying path slightly towards the end where he lands I believe, but it's still good enough.


patou01:
I think this PR is fully ready now... Only remaining issue that I'm aware of about Kirtonos is that the path kinda sucks, but I think that's a core issue on pathing and not on the boss itself. https://drive.google.com/file/d/1hlkrFheGlin4BPSZBFVWHNBDtLxpaNQ3/view?usp=sharing

patou01:
@azerothcore/testers if anyone could have a look that the event happens correctly?
Should be checked that:
- all 3 issues are fixed
- no obvious bug between summon and end of fight.

Nyeriah (approved):

Synful-Syn:
Also closes these two original reports.
https://github.com/chromiecraft/chromiecraft/issues/2041
https://github.com/chromiecraft/chromiecraft/issues/2221

Helias:
@patou01 merge conflicts, please fix them to allow the test :wink:

patou01:
@Helias , done

patou01:
@azerothcore/testers , since I believe the boss is currently 100% broken unless someone fixed it behind our back, might be worth to have a look at it.

temperrr:
Tested:

- [x] No errors
- [x] He flies outside for a bit, then flies inside to the staircase
- [x] He uses abilities, kicks you back, goes in and out of bat form.
- [ ] When he kicks you out the window, you get a loading screen and go back to where you entered RDF (is this correct?)
- [x] Can only summon once
- [x] Can summon again if you wipe

patou01:
In vanilla and TBC, you'd get ported out if  you fell out the window. I assume that's the same for rdf. I'll do some checks during the rest of the week and add the label if I find it.

patou01:
> Tank him at the top of the stairs. He has a knockback attack and if he sends you out the window, you will zone out of the instance.  

From https://wowwiki-archive.fandom.com/wiki/Kirtonos_the_Herald . As far as I'm aware zoning out will put you back to where you entered, so RDF joining location in this case. I don't see anything that says it changed between introduction and removal.

PR #8464, opened 2021-10-13
fix(Core/Globals): Check if creatures got any valid `InhabitType` dur…

Nyeriah:
…ing load to prevent broken creatures

## Changes Proposed:
-  Check if creatures got any valid `InhabitType` during load to prevent broken creatures from going unnoticed.
-  We are currently only checking if InhabyType = 0 or InhabitType = absurd, but we gotta also check if the creature actually has any valid inhabittype
- This is an easy mistake to make but an uncommon one, and as an uncommon one when it happens it's usually not spotted right away (e.g https://github.com/azerothcore/azerothcore-wotlk/pull/7426#issuecomment-941741277)
- The goal here is not that the creature works in-game if a broken template is loaded, but that the user is alerted that wrong data was used and he should see it fixed

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested starting up with intentionally messed up templates.
- Tested in-game.

Creatures that only got InhabitType 8 currently will reset once pulled, as they never find a valid path to their target.

I applied the default correction here, since I saw that was used in the other cases of wrong inhabittype, but the result is still a creature behaving very badly.

## How to Test the Changes:

```sql
UPDATE `creature_template` SET `InhabitType` = 8 WHERE (`entry` = 30455); -- Previously 1
```

1. run the query above
2. start the server
3. check db logs
4. revert the broken query

```sql
UPDATE `creature_template` SET `InhabitType` = 1 WHERE (`entry` = 30455); -- Previously 1
```

- [ ]
- [ ]


Nyeriah:
> Why cant the creature be rooted?
> And why dont we fix the root instead if it is broken?

1) ROOTED isn’t an actual inhabit type. It was a bad design choice to add something completely unrelated there. TC moved It to a flag_extra in time.

As mentioned  in the report, ROOTED only means it can’t walk. But first it needs to know where it can’t.

Without any valid inhabit types (Ground, water, air) and just the rooted flag set, the creature will be stuck on reset, because it never finds a valid path to target.

it works fine if it has ROOTED + GROUND, for example.

2) The root isn’t broken, vide the PR description.

Kitzunu:
ty for pr

Issue #8479, opened 2021-10-13
Item "Plans: Corruption" Incorrectly Dropping From Various NPCs


Azcobu:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [x] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1860

drops from 11 mobs

![image](https://user-images.githubusercontent.com/84640316/134785444-257894be-7f68-4487-a881-5ab56aa079f5.png)

Tester: confirmed. It is also contained in RLT 24709, which isn't used by any NPC directly, but is itself contained in RLT 24739, which is used by Red Dragonspawn ID 1045.

### Expected Blizzlike Behaviour

its obtainable through blacksmithing plans in stratholme

"These plans are picked up while opening Blacksmithing [10-50] Blacksmithing Plans while in Stratholme. "

### Source

https://tbc.wowhead.com/item=12830/plans-corruption
https://wowpedia.fandom.com/wiki/Plans:_Corruption?oldid=2227987

### Steps to reproduce the problem
```SQL
SELECT ct.name, clt.chance, ct.maxlevel, it.ItemLevel
FROM `creature_template` ct
JOIN `creature_loot_template` clt ON ct.lootid = clt.entry
JOIN `item_template` it ON clt.item = it.entry
WHERE it.entry = 12830;
```

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/7e2e6f8ee85f82f57a3c6e0f6f645b4c42292e7a

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[...]

PR #8482, opened 2021-10-13
fix(DB/creature_loot): Remove Plans: Corruption from various NPCs

Azcobu:
## Changes Proposed:
- Deletes [Plans: Corruption](https://tbc.wowhead.com/item=12830/plans-corruption) from the drop tables of 11 NPCs.
- Removes the same item from RLT 24709, which is indirectly used by [Red Dragonspawn](https://tbc.wowhead.com/npc=1045/red-dragonspawn) by way of RLT 24739. This NPC shouldn't have access to the item either.

This removal is because the item is not supposed to be a world drop, but is instead contained inside an item [Blacksmithing Plans](https://tbc.wowhead.com/object=176327/blacksmithing-plans) in Strath, and should only be available from it. Also checked the plans and they do drop the item correctly.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8479
- Closes https://github.com/chromiecraft/chromiecraft/issues/1860

## SOURCE:

https://tbc.wowhead.com/item=12830/plans-corruption
https://wowpedia.fandom.com/wiki/Plans:_Corruption?oldid=2227987

## Tests Performed:

- Ran SQL, no errors/warnings, correct number of lines changed.
- checked with SQL queries below.

## How to Test the Changes:

Run queries below, they should both return no results.
```SQL
SELECT ct.name, clt.chance, ct.maxlevel, it.ItemLevel
FROM `creature_template` ct
JOIN `creature_loot_template` clt ON ct.lootid = clt.entry
JOIN `item_template` it ON clt.item = it.entry
WHERE it.entry = 12830;
```

```SQL
SELECT rlt.entry, it.name
FROM `reference_loot_template` rlt
JOIN `item_template` it ON rlt.Item = it.entry
WHERE rlt.item = 12830;
```

N/A.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1634130719161646149.sql
+++ b/data/sql/updates/pending_db_world/rev_1634130719161646149.sql
@@ -1,0 +1,8 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1634130719161646149');
+
+-- Delete Plans: Corruption from various NPCs
+DELETE FROM `creature_loot_template` WHERE `Entry` IN (10398, 10399, 10400, 10406, 10407, 10408, 10409, 10412, 10413, 10463, 10464) AND `Item` = 12830;
+
+-- Delete Plans: Corruption from RLT 24709
+DELETE FROM `reference_loot_template` WHERE `Entry` = 24709 AND `Item` = 12830;
+


IntelligentQuantum:
Tested

Issue #8507, opened 2021-10-14
[Western Plaguelands] Issues with Blood of Heroes


gurveleg:
### What client do you play on?

enGB

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59

### Current Behaviour
Original Report: https://github.com/chromiecraft/chromiecraft/issues/2089

They are spawning underground it seems and some are even double spawns i am using .gobject near to see them. Unable to loot.
![unknown](https://user-images.githubusercontent.com/28974715/137271555-0c8bf281-f88f-404c-ae2d-654d24296d4f.png)
![unknown (1)](https://user-images.githubusercontent.com/28974715/137271584-a80ad121-e77c-402e-9179-9be454ef2719.png)
![unknown (2)](https://user-images.githubusercontent.com/28974715/137271598-588b9bc0-afa4-4734-8f54-7edbbe714f98.png)
![unknown (3)](https://user-images.githubusercontent.com/28974715/137271606-a4539f74-3482-4bb9-9dea-b0b9f108896a.png)
![unknown (4)](https://user-images.githubusercontent.com/28974715/137271615-13d6d88f-8a98-4cd9-887b-51c1f55a9c05.png)

Tester: Confirmed AzerothCore rev. 4539833833cd

5281 and 5276 are spawned under/partialy under the ground
![WoWScrnShot_101421_092443](https://user-images.githubusercontent.com/80769357/137273112-0d9ec794-3174-45c2-ad7c-ee7fecf49597.jpg)
![WoWScrnShot_101421_092517](https://user-images.githubusercontent.com/80769357/137273115-1f1a33ad-bd54-45ac-91fb-7b4f3cab8f50.jpg)

GUIDs 45515, 45513, 45935, 45937, 45854, 45869 have too short spawn rate, they should be same as normal blood.
https://wowgaming.altervista.org/aowow/?object=176214
![blood2](https://user-images.githubusercontent.com/80769357/137277913-33b48000-f208-49f7-8a1c-4f69d97a7061.jpg)

https://wowgaming.altervista.org/aowow/?object=176213
![blood](https://user-images.githubusercontent.com/80769357/137273760-feae6788-0fd2-4b55-9cf6-ab4332c6c7c9.jpg)

### Expected Blizzlike Behaviour

Be above ground.

### Source

[...]

PR #8519, opened 2021-10-14
fix(DB\GameObject):Blood of Heroes respawn time, position and data corrections

Ragelezz:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8507
- Closes https://github.com/chromiecraft/chromiecraft/issues/2089

## Tests Performed:

- SELECT * FROM gameobject WHERE id IN (176213,176214); and see that there are duplicate spawns
- same query, respawn time is not the same to all objects
- guids 5281 and 5276 are underground

## How to Test the Changes:

1. SELECT * FROM gameobject WHERE id IN (176213,176214); no more duplicates
2. same query, respawn time is the same and adjusted gameobject_template data to match the original
3. guids 5281 and 5276 are above ground

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1634231663747169200.sql
+++ b/data/sql/updates/pending_db_world/rev_1634231663747169200.sql
@@ -1,15 +1,3 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1634231663747169200');
 
-UPDATE `gameobject` SET `spawntimesecs`=7200 WHERE `guid` IN (45513,45515,45854,45869,45935,45937);
-UPDATE `gameobject` SET `position_z`=47.4206 WHERE `guid`=5281;
-UPDATE `gameobject` SET `position_z`=65.9905 WHERE `guid`=5276;
--- Correct object data to match the original
-UPDATE `gameobject_template` SET `displayId`=4211 WHERE `entry`=176214;
-UPDATE `gameobject_template` SET `type`=3 WHERE `entry`=176214;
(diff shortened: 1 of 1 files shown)


Kitzunu:
? thart liknk gthisx PR

Nyeriah:
There are two gameobjects related to this, a chest and a trap. The trap is supposed to spawn on top of the chest but our core doesn't handle that yet (https://github.com/azerothcore/azerothcore-wotlk/pull/8572 will address it).

The trap gameobject shouldn't be spawned @ database, it gets spawned together with the chest it is linked to (linked through gameobject_template).

I appreciate the time you've taken to look into this, but those changes are incorrect and they will end up breaking it in the future

The coords and spawn time need to be adjusted for the chest object, and it seems you're changing it for the trap, but the trap is going away soon enough so that part needs correction too

Ragelezz:
So just delete the trap object?

Nyeriah:
Hello,

Gameobject 176214 is not spawned anymore, the query in this PR returns no results. Please check if any changes are still necessary after the latest updates.

Thank you for your time

Ragelezz:
I just replaced trap object with the proper lootable object as needed, no need to do other things now.

Nyeriah:
Okay, so if nothing else is required, is it safe to close this PR? The SQL query here doesn't change any rows anymore.

Ragelezz:
Yes.

Issue #8484, opened 2021-10-13
Rank 1 Spell Lock Not Silencing Target


Azcobu:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/2026

Warlock Felhunter pet ability Rank 1 [Spell Lock](https://tbc.wowhead.com/spell=19244/spell-lock#comments) is currently only interrupting spellcasting and not silencing the target.

I recorded a short video with demonstration: https://www.youtube.com/watch?v=KGab_piyI58

Tester: confirmed for Rank 1. The only other rank for Spell lock, rank 2 at lvl 52, seems to work fine.

For reference: correctly working Rank 2 Spell Lock looks like this from the target's view. Target is debuffed, silenced, and combat log shows 'is afflicted by'
![WoWScrnShot_101421_001953](https://user-images.githubusercontent.com/81782124/137148059-3cc1a181-044f-425d-975f-427b67dbc1a4.jpg)

Now rank 1 - no debuff shown, combat log does not show 'is afflicted by', target is able to cast spells with no problems:
![WoWScrnShot_101421_003224](https://user-images.githubusercontent.com/81782124/137148384-bd5a9b81-d7fa-4272-9b1e-ade7865aef91.jpg)

Target casting Frost Nova without any problems well within the 3 second silence. It does interrupt spell casting in progress though.
![WoWScrnShot_101421_004250](https://user-images.githubusercontent.com/81782124/137150500-a2bb0b54-61ad-4097-aa14-bce13ed4fdf3.jpg)

### Expected Blizzlike Behaviour

"Spell Lock" should silence it's target alongside interrupting current spellcast.

Current ability tooltip: https://i.imgur.com/acd4fhB.png
Wowhead comment talking about the silence effect: https://i.imgur.com/XA9w3tV.png

### Source

https://www.wowhead.com/spell=19647/spell-lock
+in game tooltip https://i.imgur.com/acd4fhB.png

https://wowpedia.fandom.com/wiki/Spell_Lock?oldid=2378800

### Steps to reproduce the problem

1. Cast Rank 1 "Spell Lock" on any target
2. Silence effect does not appear

### AC rev. hash/commit

[...]

Issue #2026, opened 2019-06-29
code corresponding to the missing options in the configuration file


w5860363:
options in config file: worldserver.conf.dist

Current behaviour: invalid in the file world.cpp.

Expected behaviour: Configuration File Options Work

Steps to reproduce the problem:

1. https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/worldserver/worldserver.conf.dist
MaxWhoListReturns = 20
2.open https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/World/World.cpp
No discovery MaxWhoListReturns
3.open https://github.com/TrinityCore/TrinityCore/blob/317b0a4fb3169962e84aea24e98cdd59de08646c/src/server/game/World/World.cpp#L1344 
Can find relevant content
Branch: Master.

AC hash/commit: https://github.com/azerothcore/azerothcore-wotlk/commit/391b4ba95413c7055b33c18cdfd7feaf67c7511b

Operating system: Linux, Ubuntu 14.04 and Windows 7.

Modules: No.

Other Customizations: No.

PR #8526, opened 2021-10-15
fix(DB/Spell): Spell Lock

grimgravy:
_spell removed via script and added by database (spell_linked_spell)_ 
## Issues Addressed:

- Closes [#8484](https://github.com/azerothcore/azerothcore-wotlk/issues/8484)
- Closes [#2026](https://github.com/chromiecraft/chromiecraft/issues/2026)


Kitzunu:
This is not the right way of fixing it. 

This data already exist in the DBC files.

The issue is with the execution order of effects. This is as hacky as the core hack

IntelligentQuantum:
```
 ERROR 1062 (23000) at line 147389: Duplicate entry '1633953386512414400' for key 'version_db_world.PRIMARY'
```

PR #7518, opened 2021-08-23
fix(Core/Entities): contested guards attacking after bg/recent pvp

jestermaniac:
## Changes Proposed:
aye, so always setting PLAYER_FLAGS_CONTESTED_PVP flag when player go in pvp was fucking us cover cuz this is the flag that announce to the contested guards to whoop our ass which imo shouldn't be set always but only when needed like the issue states

what i did was to not set the contest pvp timer if player in bg (extend maybe in arenas?) or there are no contested guards that can see or detect me nearby (not sure about this part maybe better smth else?).

i used MAX_AGGRO_RADIUS for the search of nearby units but idk if that's correct tbh. maybe i should put some greater constant u dig. im taking suggestions cuz idk if in retail there is a fixed radius for this bih

word

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/7497

## SOURCE:
look issue referenced

## Tests Performed:

- tested both world pvp and bg

## How to Test the Changes:

1. open two client and log in two different chars
3. with one account the other (or enter bg and leave)
4. .go c 76
5. booty bay goblins shouldn't be after ur ass

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #8558, opened 2021-10-16
fix(Core/Entities): fix contested flag timer when logging out and in

ghost:
## Changes Proposed:

PR to fix a bug introduced here https://github.com/azerothcore/azerothcore-wotlk/pull/7518

basically what happened is that if u get flagged contested pvp u got a timer of 30 sec after which the flag finna be dropped. if u log out in that time window and log back in the timer won't start again (unless u near nuff contested guards) and u'll be stuck with the flag until u attack anotha player and the guards get mad atchu again.

## Tests Performed:

- errythang i tested in that PR plus the bug introduced


Nyeriah:
**Tested, seems to be working as expected.**

Tests performed: I logged the broken character I was using to debug this, and after applying this PR the contested flag was removed correctly when I logged in with no guards in sight.

Thanks for the quick response

PR #8561, opened 2021-10-16
fix(Core/DungeonFinder): Random Dungeon Finder always removes no perm…

### LINKED ISSUE #8222: [Dungeon Finder] Queueing multiple times with a group can lead to dungeons not resetting
### Current Behaviour

Queueing multiple times with a group (without disbanding in between the runs) can lead to dungeons not resetting. We got teleported to the same dungeon that we just did (BRD) and we immediately saw, that the mobs at the entrance were missing.

Somehow we didn't get a new ID through the dungeon finder and got teleported to the "old" dungeon. Upon disbanding the group (because we did this dungeon already) we got the deserter debuff.

Dungeons that were queued and entered through the dungeon finder should always be "new" (reset).

### Source

_No response_

1. run a lot of rdfs with a group
2. have bad luck?
3. and see that the dungeon didn't reset
(rare bug, hard to reproduce)

### Extra Notes

_No response_

### AC rev. hash/commit

6c953e9333d8ff23

### Operating system

Win 10 x64

### Custom changes or Modules

_No response_

ghost:
… binds.

Fixes #8222

## Issues Addressed:

- Closes #8222

## Tests Performed:

- Tested ingame

## How to Test the Changes:

1. You need 5 people on 15 lvl (there are only 3 instance on that lvl range)
2. Queue for RDF.
3. Finish RDF dungeon.
4. Queue for the next RDF.
5. The next dungeon must be different than the 1st one.
6. Finish that dungeon.
7. Queue for the next RDF.
8. If you are now teleported to the 1st dungeon, you should get new instance id and the instance should be reset. If it is not the 1st dungeon, repeat 5-7.


UltraNix:
@syssneck @Nefertumm Please, test it.

Issue #6053, opened 2021-05-26
Laughing Sisters PVP flagged on PVE servers


Stump876:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [X] enUS
- [ ] other (specify)

##### FACTION

- [X] Alliance
- [X] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [x] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Laughing Sister NPC is PVP flagged on Chromiecraft (RP PVE server)
Link to Chromiecraft issue: [https://github.com/chromiecraft/chromiecraft/issues/700](https://github.com/chromiecraft/chromiecraft/issues/700)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

I believe depending on the nature of the server would determine if the Laughing Sister is flagged or not. PVP=flagged PVE=No flag.
These are from WOW classic. It's about all I could find showing a flag status.
PVE server
https://youtu.be/8meigtRY3Ow
PVP server
https://youtu.be/iNkBpfB5e8g?t=54

##### STEPS TO REPRODUCE THE PROBLEM:

1. .Go creature 34212
2. If you are on a PVE server neither your character will not be flagged for PVP. The Laughing Sister should not be as well.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/e50754d804987ec93792b6a4d113913ab018189b

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-LevelUpReward](https://github.com/55Honey/Acore_LevelUpReward)
- [lua-CarbonCopy](https://github.com/55Honey/Acore_CarbonCopy)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #8563, opened 2021-10-16
(Core/creature) Fix pvp tags for mobs

patou01:
Not sure where to put this, main author of the logic and stuff is https://github.com/neifion-00000000 , I just wrote the code for now to simplify and validate that it could work.

## Changes Proposed:
-  Use the flags "guard" and "civilian" to make NPC pvp targets, as well as some factions by default.
-  Apply "guard" tag where needed for the NPCs that are still missing (eg Astranaar Sentinels)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6053

## SOURCE:

Most of the research done is in the file here. But basically it seems the most logical indication of PVP flagging is not the faction alone (which it cannot be) but a combination of guards and civilian. Also we're using PVP factions because as far as we know these are neither guards nor civilians but always PVP flagged.
[faction_setpvp_write_up.txt](https://github.com/azerothcore/azerothcore-wotlk/files/7358044/faction_setpvp_write_up.txt)

## Tests Performed:

- 
- 

## How to Test the Changes:

1. Go to laughing sister: not pvp flagged (.go c 34210)
2. Go to Astranaar sentinel: pvp flagged (.go c 32429) (not yet, once we have decided on the way to go for the flags)
3. Other mobs around the world should now also show the correct flag.

- [ ] Some mobs need to have a PVP flag added but we haven't figured out which yet. The wiki suggests we could use UNIT_FIELD_BYTES_2, but as far as I can tell that gets read *after* setting the PVP (which is written on the same entry as far as I can tell). 
image


neifion-00000000:
> 
> 
> Isnt there a faction flag for pvp factions? I believe there should be, no?

not that I know of, and I've known of this bug since years ago messing around on mangoszero 

The only function for pvp flagging in the core was related to assist behavior, which was why it was so buggy/wrong to begin with.

T1ti:
some unit flags are supposed to set pvp

patou01:
I would prefer a faction flag too, but I have no idea... This is not to be merged at this state anyway (we should add a sql change too to fix some mobs) and was more of a way to bring up the discussion saying "look we think this could do the trick and improve the behaviour". Maybe there's room to add a flag on factions?

patou01:
@Kitzunu how can we decide if it is okay to use this guard flag that we use here for the pvp tag?
Also for the factions in the array, I don't think it warrants making an entry in the database as it's only ever useful here as far as we're aware and it's pretty well contained so the risk of messing something up is basically zero.

src/server/shared/DataStores/DBCStructure.h
```diff
@@ -948,6 +948,19 @@ struct FactionTemplateEntry
         return hostileMask == 0 && friendlyMask == 0;
     }
     [[nodiscard]] bool IsContestedGuardFaction() const { return (factionFlags & FACTION_TEMPLATE_FLAG_ATTACK_PVP_ACTIVE_PLAYERS) != 0; }
+    [[nodiscard]] bool IsPVPFaction() const
+    {
+        // lists the various factions that are linked to PVP. Warsong, defilers, silverwing, etc.
+        const uint32 pvpFactions[] = {1515, 1641, 412, 1598, 1514, 1642, 1577, 1599, 1214, 1215, 1335, 1554, 1597, 1216, 1217, 1334, 1534, 1596};
```
Winfidonarleyan, line 954:
```suggestion
        constexpr std::array<uint32, 16> pvpFactions = { 1515, 1641, 412, 1598, 1514, 1642, 1577, 1599, 1214, 1215, 1335, 1554, 1597, 1216, 1217, 1334, 1534, 1596 };
```

FrancescoBorzi:
what's the status of this PR?

patou01:
@FrancescoBorzi as we're taking over a flag that is no longer used (after my PR on FD is merged), I'd like to know how we go about that.

The goal of this PR is to use the flags Civilian and Guard to mark mobs as pvp. It also means a number of mobs need to be marked as guards (all smaller cities have their guards not flagged atm). However I want to be sure that we all agree that being a "guard" has basically no influence on the npcs (as FD is not dependent on the guard status).

acidmanifesto:
What is the status of this pr and is it still revelavent or abandoned?

PR #8078, opened 2021-09-25
fix(DB/gossip_menu): Add vendor option to Innkeeper Shaussiy

Azcobu:
## Changes Proposed:
- Adds the missing option to buy goods from Innkeeper Shaussiy. This NPC already had a vendor items table, but had no gossip menu to make use of it, so added that.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8058
- Closes https://github.com/chromiecraft/chromiecraft/issues/1829

## SOURCE:

See https://wotlkdb.com/?npc=6737 or any other database website, she does sell stuff. The fact that the NPC had a pre-existing vendor table supports this.

## Tests Performed:

- Ran SQL, no errors/warnings, 1 row affected per cmd.
- Checked in game, vendor menu is now working:

WoWScrnShot_092521_164640
Buying items:
WoWScrnShot_092521_165820

## How to Test the Changes:

1. .go c 37069
2. Try to buy something from her.

## Known Issues and TODO List:

N/A.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #8573, opened 2021-10-17
Revert "fix(DB/gossip_menu): Add vendor option to Innkeeper Shaussiy"

Azcobu:
Reverts azerothcore/azerothcore-wotlk#8078 as per https://github.com/azerothcore/azerothcore-wotlk/pull/8078#issuecomment-944929149


Nyeriah:
Something seems wrong, it shows no files were created

Azcobu:
I was wondering about that, but haven't ever reverted anything before so wasn't sure what was going on. This was made by just hitting 'Revert' on the other issue, is there more that needs to be done?

Nyeriah:
> I was wondering about that, but haven't ever reverted anything before so wasn't sure what was going on. This was made by just hitting 'Revert' on the other issue, is there more that needs to be done?

Yep, to revert database changes generally we need to reverse what we've done, that button only works for c++ changes. I can help you with that if you'd like

Azcobu:
Thanks, I'll see how it goes. Will close this PR then.

Azcobu:
This is turning out to be trickier than I thought. If I just delete the gossip menu option I originally added (```DELETE FROM `gossip_menu_option` WHERE `MenuID` = 347 AND `OptionID` = 3;```), leaving behind the gossip menu option to set the hearthstone, the gossip disappears altogether. Now, when you interact with her, you just get an immediate 'set hearthstone?' prompt with no gossip at all:

WoWScrnShot_101721_145212

But this isn't correct, as your screenshot from Classic shows - it should show some gossip text with an option there to set the hearth. There's a remaining gossip_menu_option still set for setting the hearthstone in gossip (her gossip menu ID is 347), but it's not shown. There's a flag in the creature_template `type_flags` field called FORCE_GOSSIP and even that doesn't make her display the hearthstone gossip option. Also tried changing her flags, as she has 2/Questgiver and 7/Vendor set - tried removing those, leaving just 0/Gossip and 16/Innkeeper set, but that didn't help. So I'm a bt puzzled about how to make her display the gossip menu option she has remaining. Any thoughts?

On a side note, she may also need Trick or Treat mechanics added - maybe that's why she has the questgiver flag.

Issue #8564, opened 2021-10-16
Core: "Ignore FD" is not consistent in creature.cpp 


patou01:
### Current Behaviour

_canDetectFDOf:
![image](https://user-images.githubusercontent.com/2592673/137598722-fb3519a7-12e1-4ab2-9f04-42cdb4f8deec.png)
It uses isGuard(). 
And emperor in AQ40 use it:
![image](https://user-images.githubusercontent.com/2592673/137598711-46917bcc-cad8-4b9c-9f40-38e2582d7999.png)

There's also "CanIgnoreFD"
![image](https://user-images.githubusercontent.com/2592673/137598752-7b506903-dced-42b5-87e9-e5fc1247925a.png)

Which returns something different, and is used elsewhere.

### Expected Blizzlike Behaviour

There should be only one, and as far as I can tell it's purely mob related, not the target. We should probably remove the _canDetectFD and either correct all to canIgnoreFD or rewrite the logic of whatever uses it.

Part of this was introduced here
https://github.com/azerothcore/azerothcore-wotlk/commit/e29acf44d03793f4f12193dd0f7622ea5ddf6938#diff-2222cde2ff898034f0ce38bba2d5755cbdeef050c328b2d56599bc6d6664afc1 

I guess we're just missing some clean-up?

### Steps to reproduce the problem

search for any of the functions written above in the code and where they're used

### AC rev. hash/commit

c4cbd6161d25701c28ac03c19275782f0c72ffa8

### Operating system

not relevant

PR #5360, opened 2021-04-19
feat(Core/Creature): Create & use CREATURE_FLAG_EXTRA_IGNORE_FEIGN_DEATH

Kitzunu:
Co-Authored-By: offl <11556157+offl@users.noreply.github.com>
CP https://github.com/TrinityCore/TrinityCore/commit/642796f3f2c66c596358d7376125c99599e73b67

## Changes Proposed:
-  Create & use CREATURE_FLAG_EXTRA_IGNORE_FEIGN_DEATH which has been found in difficulty flags leaks

## Tests Performed:

- None yet

## How to Test the Changes:

Check that these creatures ignore feign death

## Target Branch(es):
- [x] Master

 
## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #8586, opened 2021-10-17
Core refactor: Clean-up after FD

patou01:
## Changes Proposed:
-  remove _CanDetectFeignDeathOf(Unit) and adapt necessary code.
-  Added a check for FD in CanCreatureAttack() as these were basically always paired.
- that function was not doing what the name suggests, or at least not in a clear way. It's also supposed to not be used and should have been removed when https://github.com/azerothcore/azerothcore-wotlk/pull/5360 was introduced.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8564

## Tests Performed:

- Built and tested the shaman elementals. Mobs still seem to aggro normally.
- 

## How to Test the Changes:

1. Test if behaviour of pets, shaman elementals or mobs overal changed. There should be no change.
2.
3.

- [ ]
- [ ]


patou01:
@Nyeriah any input on how to test this? Is it a matter of eh... "it doesn't seem to do anything gamebreaking and the mobs still attack" ?

FrancescoBorzi:
> @Nyeriah any input on how to test this? Is it a matter of eh... "it doesn't seem to do anything gamebreaking and the mobs still attack" ?

@patou01 well, you wrote this PR so you know better what can be potentially affected. Can you write a summary so people can double-check things in game?

patou01:
> 
> @patou01 well, you wrote this PR so you know better what can be potentially affected. Can you write a summary so people can double-check things in game?

That's a bit the issue here... this is a left-over cleanup from others. In theory, all mobs and FD are affected. For some reason, some bosses were using `_canDetectFDOf()`, so I guess we want to check they still behave as they should (but bosses should never ignore FD afaik).

I assume running a test of FD should do the trick? So checking that pets disengage when the target FDs, checking that bosses and mobs disengage, checking that city guards **don't** disengage, and I think that should do the trick?

Nyeriah:
As far as I know, the easiest example of creatures that ignores FD is indeed city guards, I'm not sure what else. Perhaps truesight auras do something about it too?

patou01:
Note that we're in theory not changing anything, just unifying. So if anyone is detecting FD based on the removed function but not on CanIgnoreFD, then the issue is a missing db flag for that entry (though it would be nice to catch such cases before they go live)

FrancescoBorzi:
what's the status of this PR?

patou01:
Fixed merge conflicts.

What should be tested:
- pets stop chasing when target FDs
- pets don't start attacking someone that was in combat with them after they FD. So: hunter + pet engage on 2 targets (let's say another hunter, hunter B). If both hunter and pet B has attacked pet A, if hunter B FDs and pet B dies, pet A shouldn't engage on hunter B by himself.
- Fire and Earth elementals of shamans should work as before (engaging on spawn is where the difference might have happened)
- mobs shouldn't engage targets that are in FD. FD should still work normally.
- guards should still ignore FD
- Twin emperors should still reset on FD. Same for Hadronox, Lady Deathwhisper

claudiodfc:
Tested.
- Player X has pet in defensive mode. Player Y has pet on passive mode. When Player Y hits player X, player x's pet starts attacking player Y. On Feign Death, pet resets combat.
- Pets don't start attacking someone that was in combat with them after they FD.
- Orc Hunter teleports into Stormwind and aggros a Guard. On Feign Death the guard resets. If close enough to the guard when they reset, they start attacking the player. They probably shouldn't even reset?
- Teleported to Icecrown near packs of adds while in gm mode and casted Feign Death. When gm mode was off, adds never aggroed while wandering around the player.
- Using Feign Death after aggroing Lady Deathwhisper makes her reset after 20s approximately. Hadronox and Emperors reset immediately.
- Elementals look fine. Started attacking while in combat.

The only possible problems are in how Lady Deathwhisper resets and maybe guards ignoring FD altogether. Overall, It's good to go 👍🏼

PR #8653, opened 2021-10-20
feat(CI/CheckPending): add check version db line

Winfidonarleyan:
## Changes Proposed:
- Move all check code to separated file `ci-pending.sh`
- Add new check version db line


Winfidonarleyan:
> > > over all seems ok, but the echo from each check dont seem to be making it into the output for the check. Maybe try removing the `>` from the echo lines to see if that solves it.
> > 
> > 
> > No, check work if exist SQL files
> 
> Humm, ill try doing some local testing with it.

Thanks

.github/workflows/check_pending_sql.yml
```diff
@@ -4,25 +4,8 @@ on:
 
 jobs:
   check-pending-sql:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
FrancescoBorzi, line 7: I'm against this, because we will not realise when `ubuntu-latest` will point to a different version and build will suddenly start failing.

I'm in favour of always using the latest Ubuntu LTS, it's just with this change we will not know when exactly that happens :)

FrancescoBorzi (changes requested):

.github/workflows/check_pending_sql.yml
Kitzunu: for this it shouldnt matter as it is only checking the SQL file and not building the entire solution, no?

FrancescoBorzi: true

FrancescoBorzi (approved):

PR #8654, opened 2021-10-20
fix(Core/AI): fixed possible "self interrupts" in BossAI::ExecuteEvent calls

sanctum32:
This PR fixed possible issue where for example, we get 2 events, at first event unit started to channel some spell but at same event loop, at same time there another event is about to get triggered which maybe is also about to trigger another spell. Result for this is clear - one of spells will not get executed or both casts will get interrupted.
This PR fully addresses possible "self interrupt" case. Instead of having two spells dodging between themselves - next event spell will get delayed once unit state gets changed (casting finished). 
Instant spells doesn;t have any impact for this.

## Changes Proposed:
-  adds additional casting state check in BossAI::UpdateAI after ExecuteEvent.
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- Made custom creature script which has two identical channeled spells cast events and examined if they gets casted properly.
- 

## How to Test the Changes:

This mainly can be tested by C++ developers, choose any boss script, select two or more events which triggers channeled spells and set identical event timers.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
@@ -622,15 +622,25 @@ void BossAI::SummonedCreatureDespawnAll()
 void BossAI::UpdateAI(uint32 diff)
 {
     if (!UpdateVictim())
+    {
         return;
+    }
 
     events.Update(diff);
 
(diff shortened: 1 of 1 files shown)


sanctum32:
of course you can also fix such case in ExecuteEvent override itself but thats more generic

Issue #6232, opened 2021-06-05
Warrior/Druid Charge Occasionally Clipping Through World


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [ ] Generic
- [x] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/774

When I charged the Redridge Mystic on top of a hill in Redridge, I fell under the texture.

![WoWScrnShot_060421_112201](https://user-images.githubusercontent.com/84775498/120810442-57baa180-c54b-11eb-96fd-251da6c2420e.jpg)
![WoWScrnShot_060421_112115](https://user-images.githubusercontent.com/84775498/120810477-6012dc80-c54b-11eb-8e58-b0729c6c03cb.jpg)

Tester's notes: The NPC is question has a GUID of 28327, and it is standing on a sharp little hill. The clipping issue is not common, but it can happen:

https://user-images.githubusercontent.com/81782124/120879921-01f6ff80-c606-11eb-8af0-28ab351022d6.mp4

##### EXPECTED BLIZZLIKE BEHAVIOUR:

It shouldn't happen.

##### STEPS TO REPRODUCE THE PROBLEM:

1. .go c 28327
2. Charge the Redridge Mystic if he stands on the little hill like on the screenshot
3. Keep trying, it took ages to verify.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/40cf7b46d221e562a20264316d660318af213690

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[...]

PR #8692, opened 2021-10-22
fix(Core/Spells): Fix charge pathing issues 

sschepens:
## Changes Proposed:
- Port from TC https://github.com/TrinityCore/TrinityCore/pull/24010
- Uses Path Generator to generate Path but shortens it to not land directly on target
- Path shorten should normalize the path to correct Z position
- Fix `Unit::canSwim` which was incorrectly checking player control
- Fix `PathGenerator` to generate shortcuts whenever creatures are *in* water instead of *under* water, this fixes charging whenever any or both involved units are on the top of the water.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6232

## Tests Performed:
- Tested multiple times on reported mobs `.go c 10904` and `.go c 28327` and never clipped through terrain
- Tested on Ring of Valor and didn't see any problems, it seems the hacks in place are no longer required, they should theoretically be handled by PathGenerator probably.

Haven't still tested on Ruins of Lordaeron.

Just a comment, Z position seems to end up off sometimes, but it's actually correctly generated in the server spline movement. It seems something is causing the client to stop the movement earlier or to inform a bad position, but this hasn't generated any serious clipping for me, but you still sometimes see that the character is a slighlty under ground.

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


sschepens:
@jamadaha mi last two commits fix the issues you reported, though I would actually like for someone with more experience review the changes.

@temperrr the issues should now be fixed and ready for review

FrancescoBorzi:
cc @Yehonal @Footman

jamadaha:
Tested it again, and so far as I can find it works perfectly.

sschepens:
There is still one issue remaining, when both caster and target are floating at the top of the water it fails to generate path.

I think the problem is actually that `sourceUnit->CanSwim()` is returning false, when it seems kind of odd because we are actually swimming.

acidmanifesto:
I was not aware we had a possible related attempt pr going on. Currently my map calculations solves it but we have issues with the vmaps needing to be revised which is out of my scope of expertise. https://github.com/azerothcore/azerothcore-wotlk/pull/8775

sschepens:
My last two commits fix top of water charging.

It seems that `Unit::canSwim` had a bug where it was checking for the wrong flag which was renamed some time ago in TC.

It also seems that when pathing we care if the unit is **in** the water, not necessarily **under** it.

sschepens:
The only issue I see remaining with charge is related to transports AC is intentionally aborting Charge when units are on different transports https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Spells/Spell.cpp#L5805.

This is weird because in Ring of Valor one can't charge against an enemy even though the transport is on ground level, also, when on top of an elevator, you can't charge downwards to a target which seems weird as well. 
Is there someone who knows the exact blizzlike mechanics regarding transports?

sschepens:
@FrancescoBorzi fixed style check, this should be ready for review and merge

laasker:
Not sure if its related to this PR, but in Dalaran Arena and Blade Edge's Arena, you can charge targets above the floor/bridge and then you stop before reaching the target. Not sure if it is suposed to act like that

https://user-images.githubusercontent.com/79991768/140234288-23166c1d-2d2b-4de7-a532-af3ff8bd501a.mp4

https://user-images.githubusercontent.com/79991768/140235070-933fadaf-f21d-41da-8797-9e27bc035237.mp4

src/server/game/Movement/MovementGenerators/PathGenerator.cpp
```diff
@@ -211,20 +211,21 @@ void PathGenerator::BuildPolyPath(G3D::Vector3 const& startPos, G3D::Vector3 con
     {
         bool buildShotrcut = false;
 
-        bool isUnderWaterStart = _source->GetMap()->IsUnderWater(_source->GetPhaseMask(), startPos.x, startPos.y, startPos.z, _source->GetCollisionHeight());
-        bool isUnderWaterEnd = _source->GetMap()->IsUnderWater(_source->GetPhaseMask(), endPos.x, endPos.y, endPos.z, _source->GetCollisionHeight());
-        bool isFarUnderWater = startFarFromPoly ? isUnderWaterStart : isUnderWaterEnd;
+        bool isWaterStart = _source->GetMap()->IsInWater(_source->GetPhaseMask(), startPos.x, startPos.y, startPos.z, _source->GetCollisionHeight());
```
Yehonal, line 214: We can't use "IsInWater" here, this shortcut is made for those creatures which are fully under water (where the shortcut is possible), but when you are on the shore (just with the feet in water) the path **must not** use a shortcut otherwise they can have issues with textures.

What are you trying to solve here?

sschepens: I checked on shore and IsInWater returns false.
It's implementation is checking if the unit is swimming, so this should be OK.

The issue happens when you're floating on the surface of the water and attempt to charge a creature underwater, or the other way around. Also if both units are floating on the surface, path generation fails as well.

sschepens:
@laasker that's definitely not intended, I'll check.

sschepens:
@laasker I knew this would come, it's the change i introduced about the path limit, I'm gonna revert that.

@jamadaha this will revert back to failing charge at max distance on the redrige mob you were testing.

From what I can see, Trinity Core is allowing the path the exceed by 50% the range of the spell (spellRange * 1.5), but AC was actually allowing some degree more, [about 60%](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Spells/Spell.cpp#L5813) it was allowing a path limit of 40 which is 60% more than 25 which is Charge range.

I don't really know what's blizzlike behaviour here, should we keep our 60% or keep TC's 50%?

src/server/game/Movement/MovementGenerators/PathGenerator.cpp
sschepens: @Yehonal what do you think? maybe it's a vmap/mmap related issue, i've seen several reports that TC is handling these things better.

sschepens: @Yehonal what do you think of this? I have checked several water positions, such as on the shore, where only feet are in water, and behavior is OK because we're checking if the unit is swimming.

Yehonal: TrinityCore does not handle the long paths that contain both underwater and on-ground sections. For instance, if a creature chases you for a long path, once it has to go back to its own original position, the creature might swim to the shore, then run on the ground, then swim into a lake, then run again on the ground. This code above actually checks if the paths are composed of a mix of these 2 by correctly using shortcuts when it's in water and normal paths when the creature can walk.

Without that code (the trinitycore case) the creature would literally fly above the water in most of the cases and it continues to fly on the shore. TC applied some "tricks" with a combination of MMAPS + creature flags in order to allow the creatures not to do crazy stuff underwater. This code there is very generic instead and allows all creatures that can swim to move flawlessly.

Probably, we need to investigate a bit more on the **charge** issue when it's in water, but I'd not change this code if it breaks the behaviour I've just mentioned

sschepens: @Yehonal i just did a few tests pulling murlocs on the shore down to the water, and the path seems OK, they walk to the water, then swim towards me. When returning it does the same, swim towards the shore, then walking to it's original position.

Also tried pulling one under water, make it travel through an island again to the water on the other side, and then reset them, and also the path generated seems ok.

Is this the case you were talking about?
Is there any other case that should be checked?

sschepens: I mean, i see some weird walking in water, but it seems to be the same on master branch.
With this changes: https://youtu.be/WB9sjc27TDk
master: https://youtu.be/Uh7AFwUMfHU

It does seem pretty similar, with both I can see some degree of weird movement with creatures walking in the water instead of swimming, but they do climb the island accordingly.

sschepens: @Yehonal I can revert this particular change if you want, I have an incoming PR to update our maps to TC versions and it should fix this issues and some blink related ones as well.

sschepens:
@Yehonal I reverted the underwater change, I will fix this charge issue separately once this and https://github.com/azerothcore/azerothcore-wotlk/pull/9124 are merged.

This should be ready for another review

Yehonal:
> @Yehonal I reverted the underwater change, I will fix this charge issue separately once this and #9124 are merged.
> 
> This should be ready for another review

ok! maybe you can test again right after that PR and try with or without the underwater change. I've seen your video and it's a good example though. If you notice, with your changes the Murlocs tends to "hover/fly" upon the small hill, while without the underwater change they walk correctly. Indeed, I can still see some weird behaviours (they should run, but probably a bug because of the stun?) but I think it's way worse with the changes

sschepens:
> > @Yehonal I reverted the underwater change, I will fix this charge issue separately once this and #9124 are merged.
> > This should be ready for another review
> 
> ok! maybe you can test again right after that PR and try with or without the underwater change. I've seen your video and it's a good example though. If you notice, with your changes the Murlocs tends to "hover/fly" upon the small hill, while without the underwater change they walk correctly. Indeed, I can still see some weird behaviours (they should run, but probably a bug because of the stun?) but I think it's way worse with the changes

With #9124 I think we can safely port TC's PathGenerator code to ours, probably some of your underwater stuff is no longer necessary. Particularly this case: https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/PathGenerator.cpp#L213 makes charging to an underwater creature work OK, we're missing that case.

Yehonal:
> > > @Yehonal I reverted the underwater change, I will fix this charge issue separately once this and #9124 are merged.
> > > This should be ready for another review
> > 
> > 
> > ok! maybe you can test again right after that PR and try with or without the underwater change. I've seen your video and it's a good example though. If you notice, with your changes the Murlocs tends to "hover/fly" upon the small hill, while without the underwater change they walk correctly. Indeed, I can still see some weird behaviours (they should run, but probably a bug because of the stun?) but I think it's way worse with the changes
> 
> With #9124 I think we can safely port TC's PathGenerator code to ours, probably some of your underwater stuff is no longer necessary. Particularly this case: https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/PathGenerator.cpp#L213 makes charging to an underwater creature work OK, we're missing that case.

No, the PathGenerator should not be migrated. Because I already did it in past and removed all the TC code that was bugged, including that one (from 4 years ago). I do not think it's a good idea just importing TC stuff without checking again all the improvements I made in my PR to that part of code

sschepens:
@Yehonal of course I will check behaviors before porting code. What I was meaning is that with the maps update come a lot of better handling of water surfaces and slopes, and maybe some of your code regarding those can be safely eliminated.

claudiodfc:
Do you need to test something currently or is this WIP?

sschepens:
@claudiodfc I'm currently waiting for #9124 to be merged so I can introduce some more changes

UltraNix:
@sschepens What is the status of this?

UltraNix:
@sschepens ??

Kitzunu:
@IntelligentQuantum feel free to take over <3

PR #8720, opened 2021-10-24
fix(Core/RealmList) Fix display of realm population

Meltie2013:
## Changes Proposed:
-  Fix realm population not reflecting real levels of players on given realm
-  Fix realms displaying `Medium` levels when there are 2 or more realms when `population` remains `0`
-  Fixed 2 realm flags that were displaying incorrectly when used

**Without Changes**
image

**With Changes**
image
This last images reflects the values updates to the `population` field in the `realmlist` database.

**Notice**
Make sure you have a set max population when testing this, since the flags are dependent on max realm population and total connected players.

## Tests Performed:

- Builds without errors
- View images above (keep in mind, the flags `New` and `New Players` are flipped in the image.
  This been fixed with current PR.

## How to Test the Changes:

1. Make 2 realms (can be a realm that is not used)
2. Log into an account
3. You can either log into the world or sit on the realmlist screen
4. Wait for the server to update the population or change `population` field to any value between `0` and `1`
5. Watch the population levels change according to the values given per realm


Nyeriah:
I did some research about this and I've found something:

>Each realm has a population listed on the server selection screen. Each realm is characterized as having Recommended, Low, Medium, High, and Full population. **Recommended servers are what Blizzard thinks will give a new player the best experience, while the rest are categorized based on how many players are currently logged into the realm.** These servers change throughout the day, so you may want to decide on a server at the time you would normally play, so that you'll know how full it is normally. You cannot create a character on a Full server unless you already have characters there. Some High population servers have wait queues during peak hours. This means you cannot sign in at all to them until other people have completely logged off of that server.

Apparently the "recommended" flag should be static and not change depending on the amount of players logged in.

FrancescoBorzi:
@azerothcore/testers can anyone please look into this?

Meltie2013:
Closing this PR out and looking at a different methods like IPC.

Maurowin:
Let me know when you want another tests like the last time.

Issue #5649, opened 2021-05-05
The Green Tower Shield - Proc Does Not Work


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [x] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/531

The Equip on the lvl 36 item The Green Tower does not proc. https://wowgaming.altervista.org/aowow/?item=1204

Tester's notes - The proc is only a 1% rate anyway, but tried this with an 80 warrior wearing the shield vs a fast-hitting lvl 20 druid cat, and let the cat melee the warr for about 10 mins, delivering about 600 melee attacks. No procs were seen before the warr died, so it does seem pretty likely the proc is not working.

![WoWScrnShot_050521_231513](https://user-images.githubusercontent.com/81782124/117151050-db238e80-adf7-11eb-9d99-a27156f82d7a.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

" When struck in combat has a 1% chance of raising a thorny shield that inflicts 3 Nature damage to attackers when hit and increases Nature resistance by 50 for 30 sec." 

The spell it should cast is https://wowgaming.altervista.org/aowow/?spell=18097

##### STEPS TO REPRODUCE THE PROBLEM:

1. Have a high level character with the shield duel a low level character (so they don't do much damage)
2. Gather a few hundred melee strikes.

##### EXTRA NOTES:

https://wowgaming.altervista.org/aowow/?item=1204

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/91081f4ad8fed98d7bdc9c0ae13d05c86032df9f

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #8745, opened 2021-10-26
fix(Core\Spells):The Green Tower passive aura duration

Ragelezz:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/5682
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5649
- Closes https://github.com/chromiecraft/chromiecraft/issues/531

## Tests Performed:

- .aura 18097 and after .list aura, observe it has a duration of 30s, passive auras from items which gives a benefit (like procs or so) should be unlimted
- 

## How to Test the Changes:

1. .aura 18097 and after .list aura, observe Duration to -1, unlimited
2.
3.

- [ ]
- [ ]


Ragelezz:
@T1ti

Ragelezz:
@azerothcore/testers

Nyeriah:
> Shouldn't u fix how the cores handles the data rather than modifying the data ?

Hey, do you have any suggestion on how we could address this one?

jamadaha:
**Tested**
Bot tested the outlined method before and after, and this PR does indeed change it to -1 / unlimited.
After:
image
**Seems to work**

Helias:
Probably, @T1ti means that there is a way to edit that data, the DurationIndex, through the SQL table `spell_dbc`, you could do it even Keira3 too if I remember correctly

Ragelezz:
So how we should do it exactly ?

Helias:
read in detail this https://www.azerothcore.org/wiki/spell_dbc
installing node-dbc-reader (https://github.com/wowgaming/node-dbc-reader) you could create the spell_dbc rows that you need and later you can edit them using Keira3

T1ti:
> Probably, @T1ti means that there is a way to edit that data, the DurationIndex, through the SQL table `spell_dbc`, you could do it even Keira3 too if I remember correctly.
> 
> EDIT: you can, but you need to first populate the row of `18097` in spell_dbc with the original data (if it is not already present) and later you can use Keira3 to use the UPDATE sql query and edit the DurationIndex to -1.
> 
> I know, it can seem complicated but it's more long than complicated, you can use the DB to make this fix following this guide and using Keira3 https://www.azerothcore.org/wiki/spell_dbc
> 
> I am fine to keep this in the core, but I would like to move all the spells edit to the DB in the future if we can

No I meant that I doubt we need to modify the data, and that if AC needs it to be set to unlimited instead of the blizzlike data (the cactual duration), it's probably because AC doesn't handle the spell data correctly.

claudiodfc:
I agree that this shouldn't be merged. If we keep adding band-aids to problems like these, one day when we have to remove the band-aids the infection will be out of proportions.
Fixing how the spell is handled and not hack fixing should be the way to go.

PR #8771, opened 2021-10-27
Fixxing the gunship battle cannons in icecrown

qwe900:
Getting energy on cast not on hit

## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## SOURCE:

[https://youtu.be/nSDn2m0ThsI?t=33](https://youtu.be/nSDn2m0ThsI?t=33)
## Tests Performed:

- 
- 

## How to Test the Changes:

1. go to gunship battle
2. enter gun
3. shot

may theres is a cleaner way to do it
- [ ]
- [ ]


Ragelezz:
Works.

Ragelezz (changes requested):
break not return at line 1955

Kitzunu:
Hacks... Try making a generic spellscript instead please

Ragelezz:
Or a spellscript inside gunship file but I doubt none of the bosses require more attention.

Kitzunu:
I'd prefer it to ve generic incase we find more cases like this

qwe900:
> 
> 
> break not return at line 1955

if we break we just leave the switchcase and generate power on cast and hit so we get double energy.

A Generic spellscript will also not work i think because the spelleffect is handled before and will automaticly generate the power.

qwe900:
> 
> 
> Or a spellscript inside gunship file but I doubt none of the bosses require more attention.

There are few more bugs in IceCrownCitadel.
Gunship artillery does not work , the target selection is wrong and the spell is wrong i think. 
There is also no knockback on artilly hit.

Professor debuff on tank does not stack in last phase. 

And as Healer you dont have a chance to win the fight against the mob where u will be teleported to. Because the NPC u heal dont get the dmg buff.

Ragelezz:
As said, ICC might need a rework later.
To this PR, you can remove the Energize effect from spell via SpellMgr and make a spellscript to handle properly on cast in gunship file. Generic spellscripts SHOULD NOT be involved in instance scripts at all.

Kitzunu:
Fair point.

Cc @Nyeriah do you know of a better way of doing it without hardcoding stuff?

Nyeriah:
@Kitzunu this may have to do with that issue with spelleffects being executed in the wrong order, eff1 should execute independently of the result from eff0

Kitzunu:
Oh another one of those. Then we shouldnt touch it until a generic fix is made imo

qwe900:
theres not spelleffectapplyfix on azerothcore so how to overite this.

Kitzunu:
It is a general bug in the core, that covers more spells. eg. https://github.com/azerothcore/azerothcore-wotlk/issues/8484 

So it needs a general fix in the core, to make the execution order work as it should. Per Nyeriah's comment

qwe900:
so editing 
boss_icecrown_gunship_battle.cpp

`class spell_igb_cannon_blast : public SpellScriptLoader
{
public:
    spell_igb_cannon_blast() : SpellScriptLoader("spell_igb_cannon_blast") { }

    class spell_igb_cannon_blast_SpellScript : public SpellScript
    {
        PrepareSpellScript(spell_igb_cannon_blast_SpellScript);

        bool Load() override
        {
            return GetCaster()->GetTypeId() == TYPEID_UNIT;
        }

        void CheckEnergy()
        {
            if (GetCaster()->GetPower(POWER_ENERGY) >= 100)
            {
                GetCaster()->CastSpell(GetCaster(), SPELL_OVERHEAT, true);
                if (Vehicle* vehicle = GetCaster()->GetVehicleKit())
                    if (Unit* passenger = vehicle->GetPassenger(0))
                        sCreatureTextMgr->SendChat(GetCaster()->ToCreature(), SAY_OVERHEAT, passenger);
            }
        }
        void PowerGain()
        {
            if (GetCaster()->GetPower(POWER_ENERGY) < 100)
            {
                GetCaster()->SetPower(POWER_ENERGY, GetCaster()->GetPower(POWER_ENERGY) + urand(6, 10));
               
            }
        }

        void Register() override
        {
            OnCast += SpellCastFn(spell_igb_cannon_blast_SpellScript::PowerGain);
            AfterHit += SpellHitFn(spell_igb_cannon_blast_SpellScript::CheckEnergy);
        }
    };

    SpellScript* GetSpellScript() const override
    {
        return new spell_igb_cannon_blast_SpellScript();
    }
};`

and 
SpellMgr.cpp
`    //canon fix icc
    ApplySpellFix({69399, 70172}, [](SpellEntry* spellInfo)
    {
        spellInfo->Effect[EFFECT_1] = 0;
    });`

is better yes ?

Nyeriah:
> so editing boss_icecrown_gunship_battle.cpp
> 
> `class spell_igb_cannon_blast : public SpellScriptLoader { public: spell_igb_cannon_blast() : SpellScriptLoader("spell_igb_cannon_blast") { }
> 
> ```
> class spell_igb_cannon_blast_SpellScript : public SpellScript
> {
>     PrepareSpellScript(spell_igb_cannon_blast_SpellScript);
> 
>     bool Load() override
>     {
>         return GetCaster()->GetTypeId() == TYPEID_UNIT;
>     }
> 
>     void PowerGain()
>     {
>         GetCaster()->SetPower(POWER_ENERGY, GetCaster()->GetPower(POWER_ENERGY) + urand(6, 10));
>         if (GetCaster()->GetPower(POWER_ENERGY) >= 100)
>         {
>             GetCaster()->CastSpell(GetCaster(), SPELL_OVERHEAT, true);
>             if (Vehicle* vehicle = GetCaster()->GetVehicleKit())
>                 if (Unit* passenger = vehicle->GetPassenger(0))
>                     sCreatureTextMgr->SendChat(GetCaster()->ToCreature(), SAY_OVERHEAT, passenger);
>         }
>   
> 
>     }
> 
>     void Register() override
>     {
>         OnCast += SpellCastFn(spell_igb_cannon_blast_SpellScript::PowerGain);
>        
>     }
> };
> 
> SpellScript* GetSpellScript() const override
> {
>     return new spell_igb_cannon_blast_SpellScript();
> }
> ```
> 
> };`
> 
> and SpellMgr.cpp ` //canon fix icc ApplySpellFix({69399, 70172}, [](SpellEntry* spellInfo) { spellInfo->Effect[EFFECT_1] = 0; });`
> 
> is better yes ?

I don't understand why the spellmgr part is necessary. But changing that spell script from AfterHit to OnCast will probably solve it, yes? I didn't know those spells already had a spell script, but it's likely the issue here, yeah (perhaps the only issue).

qwe900:
> 
> 
> > so editing boss_icecrown_gunship_battle.cpp
> > `class spell_igb_cannon_blast : public SpellScriptLoader { public: spell_igb_cannon_blast() : SpellScriptLoader("spell_igb_cannon_blast") { }
> > ```
> > class spell_igb_cannon_blast_SpellScript : public SpellScript
> > {
> >     PrepareSpellScript(spell_igb_cannon_blast_SpellScript);
> > 
> >     bool Load() override
> >     {
> >         return GetCaster()->GetTypeId() == TYPEID_UNIT;
> >     }
> > 
> >     void PowerGain()
> >     {
> >         GetCaster()->SetPower(POWER_ENERGY, GetCaster()->GetPower(POWER_ENERGY) + urand(6, 10));
> >         if (GetCaster()->GetPower(POWER_ENERGY) >= 100)
> >         {
> >             GetCaster()->CastSpell(GetCaster(), SPELL_OVERHEAT, true);
> >             if (Vehicle* vehicle = GetCaster()->GetVehicleKit())
> >                 if (Unit* passenger = vehicle->GetPassenger(0))
> >                     sCreatureTextMgr->SendChat(GetCaster()->ToCreature(), SAY_OVERHEAT, passenger);
> >         }
> >   
> > 
> >     }
> > 
> >     void Register() override
> >     {
> >         OnCast += SpellCastFn(spell_igb_cannon_blast_SpellScript::PowerGain);
> >        
> >     }
> > };
> > 
> > SpellScript* GetSpellScript() const override
> > {
> >     return new spell_igb_cannon_blast_SpellScript();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > };`
> > and SpellMgr.cpp ` //canon fix icc ApplySpellFix({69399, 70172}, [](SpellEntry* spellInfo) { spellInfo->Effect[EFFECT_1] = 0; });`
> > is better yes ?
> 
> I don't understand why the spellmgr part is necessary. But changing that spell script from AfterHit to OnCast will probably solve it, yes? I didn't know those spells already had a spell script, but it's likely the issue here, yeah (perhaps the only issue?).

Because the effect1 will gain the power after hit which is wrong. So this override it and let handle the spellscript(which oldway just checks the power for overheating ) also gains power now.

PR #8810, opened 2021-10-28
fix(Core/Spells): Potential crashes in spell_generic and spell_item

Nefertumm:
## Changes Proposed:
-  Add missing checks to avoid potential crashes in the future.

## Issues Addressed:

- Closes 

## Tests Performed:

- Compiled

## How to Test the Changes:

1. As this PR is only added checks, all should be working the same way as before, but without potential crashes.


Ragelezz:
No idea if this can be tested until its happening, if its the case.

Kitzunu:
No need to test

Nyeriah (approved):

PR #8849, opened 2021-10-30
(Core/SmartAI): Improve SMART_EVENT_GOSSIP_HELLO

acidmanifesto:
## Changes Proposed:
Add another value to event_param1 for SMART_EVENT_GOSSIP_HELLO to support all cases of GossipHello/reportUse in GameObjects:
- event_param1 set to 0: execute the action for both GossipHello and reportUse. This might result in the action being executed twice when clicking the GameObject
- event_param1 set to 1: execute the action for only GossipHello
- event_param1 set to 2: execute the action for only reportUse

## Issues Addressed:
TC Issue that is replicable on AZ core, that was reported in 2018.
https://github.com/TrinityCore/TrinityCore/issues/21678

## SOURCE:
TC Commit to address the issue that was provided in 2019.
https://github.com/TrinityCore/TrinityCore/commit/9e58ef6940c78d92f1553cf7ab75e87a6e2cf112

## Tests Performed:
Builds without issues.

## How to Test the Changes:

Steps to reproduce the problem before the PR:

1. Execute this sql:
```
UPDATE gameobject_template SET AIName="SmartGameObjectAI" WHERE entry=181147;
DELETE FROM smart_scripts WHERE entryorguid=181147 AND source_type=1;
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(181147, 1, 0, 0, 64, 0, 100, 0, 0, 0, 0, 0, 72, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 'Gob - GOSSIP_HELLO - CLOSE_GOSSIP'),
(181147, 1, 1, 0, 64, 0, 100, 0, 0, 0, 0, 0, 12, 322, 2, 60000, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 'Gob - GOSSIP_HELLO - SUMMON_CREATURE');
```
2. .Gob add temp 181147
3. Interact with the gob once
4. notice 2 npcs are spawned instead of one

After this PR the issue is no longer a problem.


FrancescoBorzi:
> Tests Performed:
> Builds without issues.

@acidmanifesto do you mind testing it in-game as well? so we are sure this does not cause any regression

acidmanifesto:
> > Tests Performed:
> > Builds without issues.
> 
> @acidmanifesto do you mind testing it in-game as well? so we are sure this does not cause any regression

No regression from what i can see. Since the default case is set to zero currently in our database nothing gets broke as we had nothing else for event_para1 except 0 default. And as the event_para1 is and always have been marked as zero for Gossip Select ( see line fore reference:https://github.com/azerothcore/azerothcore-wotlk/blob/2cf4d0586746432cd7e307fbd1556fb7d4dd2226/src/server/game/AI/SmartScripts/SmartScriptMgr.h#L165) , we can then easily adjust to either 1 or 2 as the issues get discovered over time in the event we find one that is discovered to fire off twice.

Nyeriah (approved):

PR #8876, opened 2021-11-01
feat(Core/Unit): Implement DoCastRandomTarget() helper

Nyeriah:
## Changes Proposed:
-  Implement DoCastRandomVictim() helper

```cpp
if (Unit* target = SelectTarget(SELECT_TARGET_RANDOM, 0, 100, true))
{
    DoCast(target, SPELL_IMMOLATE);
}
```
Becomes:
```cpp
DoCastRandomVictim(SPELL_IMMOLATE, 0, 100.0f);
```

Or simply:
```cpp
DoCastRandomVictim(SPELL_IMMOLATE);
```

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested in-game.
- 

## How to Test the Changes:

1. Check if The Beast @ UBRS still casts immolate on a random target
2.
3.

- [ ]
- [ ]


Kitzunu:
Isn't "victim" always referred to as the current target with the highest threat? Could be confusing. Perhaps rename it to DoCastRandomTarget?

Also what about adding a parameter like "bool allowVictim" or similar, to allow or disallow throwing the spell on the target with the highest/second-highest threat. Usually the one or two tanks. Or perhaps uint32 threatTablePosition does that?
But I do believe that there are some mechanics where a boss should send a random spell to any character that isn't the tank

Nyeriah:
> Isn't "victim" always referred to as the current target with the highest threat? Could be confusing. Perhaps rename it to DoCastRandomTarget?
> 
> Also what about adding a parameter like "bool allowVictim" or similar, to allow or disallow throwing the spell on the target with the highest/second-highest threat. Usually the one or two tanks. Or perhaps uint32 threatTablePosition does that? But I do believe that there are some mechanics where a boss should send a random spell to any character that isn't the tank

Yes, threatTablePositon does that, it follows the same logic as the previous method. By default it selects any random target (tanks included), but you can change that as you wish like before

```cpp
DoCastRandomVictim(SPELL_IMMOLATE, 1, 100.0f);
```

will skip the tank

I'll change it to RandomTarget

Kitzunu:
>Yes, threatTablePositon does that, it follows the same logic as the previous method. By default it selects any random target (tanks included), but you can change that as you wish like before
>
>DoCastRandomVictim(SPELL_IMMOLATE, 1, 100.0f);
will skip the tank

nice I did not know this =)

Kitzunu (approved):

Kitzunu:
looks good

Nefertumm (approved):
Excellent, this is nice. In the future, maybe we can accept a predicate as parameter too. :)
Also, I think this is a really good method since we can skip checking if the target is valid within the creature script.

Issue #8853, opened 2021-10-31
Loot while stealthed (rogue + druid)


lingxDEV:
### Current Behaviour

You can enter Stealth / Prowl and loot a mob without Stealth / Prowl being cancelled.

### Expected Blizzlike Behaviour

When looting Stealth should be cancelled.

### Source

![image](https://user-images.githubusercontent.com/62337627/139578167-5d5ca600-07f5-4614-adf0-92f6769e1430.png)

https://wowpedia.fandom.com/wiki/Stealth

Test on any retail version of the game. It's always been the case.

### Steps to reproduce the problem

1. Kill an NPC
2. Enter Stealth / Prowl
3. Loot NPC
4. Stealth / Prowl is not cancelled

### AC rev. hash/commit

39bfe880f49f

### Operating system

Win 10 x64

### Custom changes or Modules

None

Shineslippers:
Indeed



https://user-images.githubusercontent.com/85256266/139581436-0545a7e1-cd21-41a8-b755-9e87657aee48.mp4


,

PR #8982, opened 2021-11-05
Fix [Core/loot] remove stealth/fd on loot

patou01:
## Changes Proposed:
-  Remove a list of auras if present on a player upon looting

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8853

## Tests Performed:

- Tested on a rogue and hunter (I believe they shouldn't be able to loot in FD)
- 

## How to Test the Changes:

1. find a mob, kill it, stealth, loot.
2. Test that opening a bag (eg bag of gems or some rdf reward) doesn't break stealth.
3. Same with FD, though I'm not sure what should happen to FD when opening a bag.

- [ ]
- [ ]


patou01:
I'm not too sure about how the loot system works, so maybe this should be done elsewhere? I believe any attempt at looting something should remove stealth? Also not sure this is the right place to make this check.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,16 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    const AuraType toRemove[] = {SPELL_AURA_MOD_STEALTH, SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH};
+    for (const auto& aura : toRemove)
+    {
+        if (HasAuraType(aura))
```
Nyeriah, line 7541: Checking if they have the aura is not necessary, if they don’t when you try to remove it nothing happens

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,16 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    const AuraType toRemove[] = {SPELL_AURA_MOD_STEALTH, SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH};
```
Winfidonarleyan, line 7538: maybe use `std::vector` ?

src/server/game/Entities/Player/Player.cpp
Winfidonarleyan: I agree with Nyeriah

src/server/game/Entities/Player/Player.cpp
patou01: Done, though I'm not sure I initialized that properly, I never use vectors for stuff that's known ahead of time.

patou01:
@nyeriah also if you have any idea of specifically which auras should be removed? These 3 came to my mind, but there might be others?

Nyeriah:
> @Nyeriah also if you have any idea of specifically which auras should be removed? These 3 came to my mind, but there might be others?

Can’t think of any other, it’s safe to go with those

UltraNix:
@patou01 Test it with pickpocket looting.

Nyeriah:
> @patou01 Test it with pickpocket looting. Stealth auras in that case should not be removed.

To add to this: also check opening a container (loot bag, RDF reward for example) and see if it destealths you... it should not :P

patou01:
Shite. Then I need to think a bit more. I'll fix that tomorrow.

src/server/game/Entities/Player/Player.cpp
Winfidonarleyan: Hmm, need learn this, maybe need use `constexpr std::array`

Winfidonarleyan: Yes, need to use `std::array`, thanks for the new knowledge

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,17 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    std::vector<AuraType> toRemove{ SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH};
```
Winfidonarleyan, line 7538:
```suggestion
    constexpr std::array<AuraType, 2> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,17 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    constexpr std::array<AuraType, 2> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
```
Winfidonarleyan, line 7538:
```suggestion
    constexpr std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,17 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    constexpr std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
+    if (loot_type != LOOT_PICKPOCKETING)
+    {
+        toRemove.push_back(SPELL_AURA_MOD_STEALTH);
```
Winfidonarleyan, line 7541:
```suggestion
        toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH, SPELL_AURA_MOD_STEALTH };
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,17 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    constexpr std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
```
Winfidonarleyan, line 7538:
```suggestion
    std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,17 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
...
+
+    std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
+    if (loot_type != LOOT_PICKPOCKETING)
+    {
+        toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH, SPELL_AURA_MOD_STEALTH };
+    }
+    for (const auto& aura : toRemove)
+    {
+        RemoveAurasByType(aura);
+    }
```
Winfidonarleyan, line 7546:
```suggestion
    std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
    if (loot_type != LOOT_PICKPOCKETING)
    {
        toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH, SPELL_AURA_MOD_STEALTH };
    }

    for (const auto& aura : toRemove)
    {
        if (aura == SPELL_AURA_NONE)
        {
            continue;
        }

        RemoveAurasByType(aura);
    }
```

Winfidonarleyan:
@patou01, I rework to `std::array` for save speed core. Can retest local?

Winfidonarleyan (approved):

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7534,6 +7534,23 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
     PermissionTypes permission = ALL_PERMISSION;
 
     LOG_DEBUG("loot", "Player::SendLoot");
+
+    std::array<AuraType, 3> toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH };
+    if (loot_type != LOOT_PICKPOCKETING)
+    {
+        toRemove = { SPELL_AURA_MOD_INVISIBILITY, SPELL_AURA_FEIGN_DEATH, SPELL_AURA_MOD_STEALTH };
```
Nyeriah, line 7541: The logic here looks more complex than it needs to be, why don't we put the FOR inside this IF? We don't really need to filter specific aura types, you can't pickpocket while invisible or in feign death f.e

Winfidonarleyan: Hm, good idea

UltraNix:
@patou01 I suggest to move this code from `Player::SendLoot` to loot handler, `WorldSession::HandleLootOpcode`, as I believe that's the only case (looting corpse) when these auras should be removed

patou01:
As far as I know, these auras should be removed in at least the following cases:
- use of a gameobject (but is that always a loot?)
- looting a creature corpse
- looting a player corpse (insignia)

I'm actually not sure if lockpicking should unstealth? I believe it shouldn't actually.

UltraNix:
When you are using gameobjects, it's already handled by core to remove steath (as far I can remember - you can test it).
Not use about insignias.
```
 LOOT_CORPSE                 = 1,
    LOOT_PICKPOCKETING          = 2,
    LOOT_FISHING                = 3,
    LOOT_DISENCHANTING          = 4,
    // ignored always by client
    LOOT_SKINNING               = 6,
    LOOT_PROSPECTING            = 7,
    LOOT_MILLING                = 8,

    LOOT_FISHINGHOLE            = 20,                       // unsupported by client, sending LOOT_FISHING instead
    LOOT_INSIGNIA               = 21,                       // unsupported by client, sending LOOT_CORPSE instead
    LOOT_FISHING_JUNK           = 22                        // unsupported by client, sending LOOT_FISHING instead
````
These are the types.
Corpse should remove steatlh (gameobjects should be already handled by core - to test)
Pickpocket should not.
Fishing - ??
Disenchanting - should not.
Skinning - ??
Prospecting - should not
Milling - should not
Insignia - ??

**Comment by patou01** [comment]
> When you are using gameobjects, it's already handled by core to remove steath (as far I can remember - you can test it). Not use about insignias.

Hmm then I might have to clean up under there, this code I've added here is just above the check based on the object type. I don't think it makes sense to have a call to "removeAuras" inside each if/else. 

> So the only loot type to make stealth be dropped is LOOT_CORPSE. So I would leave that code in that place, but limited that if only to `LOOT_CORPSE` :)

As far as I can tell then it should be `if (loot_type == LOOT_CORPSE && !guid.IsItem())`, based on a first test on my side.

**Comment by patou01** [comment]
Meh, I'll work on it a bit and push something later today.

File: src/server/game/Spells/SpellMgr.cpp [hunk 9]
```diff
@@ -7421,6 +7421,11 @@ void SpellMgr::LoadDbcDataCorrections()
         spellInfo->AttributesEx3 |= SPELL_ATTR3_ALWAYS_HIT;
     });
 
+    ApplySpellFix({ 13262, 31252 }, [](SpellEntry* spellInfo)
+    {
+        spellInfo->AttributesEx |= SPELL_ATTR1_ALLOW_WHILE_STEALTHED;
+    });
```
Kitzunu, line 7427: No.

just add the spell id's to L#4099 `spellInfo->AttributesEx |= SPELL_ATTR1_ALLOW_WHILE_STEALTHED` already exists there. No need for duplicate entries

Kitzunu (changes requested):

src/server/game/Spells/SpellMgr.cpp
patou01: Done, but we still need to know if these lines are actually needed.

FrancescoBorzi:
what's the status of this?

Nyeriah:
Are we sure we should be able to disenchant while stealthed? Disenchant is that long-ass cast time, it's kinda weird for us to be able to do that while stealthed, isn't it?

patou01:
> Are we sure we should be able to disenchant while stealthed? Disenchant is that long-ass cast time, it's kinda weird for us to be able to do that while stealthed, isn't it?

That's my feeling too. I added it as I haven't played wotlk that much and ultranix said it shouldn't.

From what I recall, only lockpicking shouldn't remove stealth, or is that not correct? Else I'll just remove that entire part and leave only the looting so we can merge.

FrancescoBorzi:
> Else I'll just remove that entire part and leave only the looting so we can merge.

In general, smaller PRs are always better as they are easier to be checked and merged

patou01:
What I've tested:
- rogue loses stealth on loot
- druid loses stealth on loot
- hunter loses FD on loot
- rogues stay stealthed when looting a bag (item)

But I realized that an invis mage cannot see the mobs, so do they matter?

src/server/game/Spells/SpellMgr.cpp
patou01: Is not part of the PR anymore.

claudiodfc:
Tested Results: ✔
- Rogue loses stealth on loot
- Druid loses stealth on loot
- Hunter loses FD on loot
- Rogues stay stealthed when looting a bag (item) and Pickpocketing

Nyeriah (approved):

Kitzunu (approved):

PR #9008, opened 2021-11-06
feat(CI): skip matrix builds for non-C++ changes

FrancescoBorzi:


.github/labeler.yml
```diff
@@ -1,3 +1,9 @@
+cpp:
```
Kitzunu, line 1: could we rename the lable to "file-cpp" or something? just "cpp" seems like a weird label to have

.github/labeler.yml
```diff
@@ -1,3 +1,9 @@
...
+  - src/**/*.cpp
+  - src/**/*.h
+  - deps/**/*.cpp
+  - deps/**/*.h
```
Kitzunu, line 5: You could do ./*.cpp

Kitzunu: to target all cpp files

.github/workflows/core_matrix_build.yml
```diff
@@ -0,0 +1,61 @@
...
+            compiler: gcc8
+          - os: ubuntu-20.04
+            compiler: gcc # default in 20.04 is gcc 9
+          - os: ubuntu-20.04
+            compiler: gcc10
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }}-${{ matrix.compiler }}
+    env:
+      COMPILER: ${{ matrix.compiler }}
+    if: github.repository == 'azerothcore/azerothcore-wotlk' && (github.event.label.name == 'cpp' || github.ref == 'refs/heads/master')
```
FrancescoBorzi, line 35: for some reasons this doesn't work... if I remove the part after `&&` it works, but with the label condition it will never trigger

.github/labeler.yml
FrancescoBorzi: ok

.github/labeler.yml
FrancescoBorzi: ok

Nyeriah (approved):

Issue #9034, opened 2021-11-07
LBRS Overlord Wyrmthalak's adds disapear after death of boss


Shineslippers:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59

### Current Behaviour

[Overlord Wyrmthalak](https://wowgaming.altervista.org/aowow/?npc=9568)'s adds disapear after death of boss.

https://user-images.githubusercontent.com/85256266/140652522-7136a9f2-4d90-4f54-b597-b5b3824b4660.mp4

### Expected Blizzlike Behaviour

They should stay alive

### Source

https://youtu.be/bcMPIdeqa74?t=3482
https://youtu.be/nA8emnau2sg?t=3101
https://youtu.be/jAY2FAWUngk?t=3322

### Steps to reproduce the problem

.go c 45757

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/420ca2e2b57216b90f93e2deb870946b420a72df

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
[...]

PR #9036, opened 2021-11-07
LBRS Wyrmthalak, adds no longer despawn after death

patou01:
## Changes Proposed:
- Rework "JustDied" 
- Add a Talk(0) when he summons mobs
- Add an entry for him in creature_text

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9034

## Tests Performed:

- tested on my local, adds don't despawn on death, but do despawn later on if they are not in combat for whatever time was set before.
- 

## How to Test the Changes:

1. .go c 45757
2.
3.

- [ ]
- [ ]


src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_overlord_wyrmthalak.cpp
```diff
@@ -77,7 +77,12 @@ class boss_overlord_wyrmthalak : public CreatureScript
 
         void JustDied(Unit* /*killer*/) override
         {
-            _JustDied();
+            events.Reset();
+            if (instance)
+            {
+                instance->SetBossState(DATA_OVERLORD_WYRMTHALAK, DONE);
```
Nyeriah, line 83: this is the only line actually required

patou01: I tried to keep the change with "minimal impact". I believe only the SaveToDB could have any use, the boss state being unused as far as I'm aware...

Nyeriah: I understand, but SetBossState() already saves to database: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Instances/InstanceScript.cpp#L331

And events.Reset() has no use because the creature is dead and events reset when they respawn on Reset()

Furthermore, boss scripts never run outside instances, so check instance for nullptr is also trivial

patou01: > Furthermore, boss scripts never run outside instances, so checking instance for nullptr is also trivial.

Hmm I doubt that part (WBs in open world come to my mind), but else we can simplify bossAI. That's where I got the code from. (since I removed the call to _JustDied() )
image

Nyeriah: See this  GetBlackrockSpireAI macro?

https://github.com/azerothcore/azerothcore-wotlk/pull/9036/files#diff-9a42647dacaa111473022e06147c4c23dcddcd80d42a63c0a86bc896ab559ee7R54

It ensures his script is only executed inside BlackrockSpire. I was wrong, he's fine, no updating required, we can get rid of the instance nullptr check.

patou01: Ah, but then we *can* remove it if we know we're in an instance, or do I get that wrong?

Nyeriah: > Ah, but then we _can_ remove it if we know we're in an instance, or do I get that wrong?

I think maybe we got confused here. We can't remove it from BossAI::_JustDied() because as you said they can be world bosses.

But we don't need to check for instance script validity when bosses are registered using that kind of function (I believe every instance has, at least should have, one of those functions to check if the creature is in the right map).

So yes, we can remove it from his script

patou01: Ok, then I remove it here while changing up a few other things in there.

Nyeriah (approved):

patou01:
@nyeriah can you check I didn't mess up the sql part please?

Nyeriah:
> @Nyeriah can you check I didn't mess up the sql part please?

It’s just missing broadcast text but I’m not sure which one it should be, there’s several for that same line

src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_overlord_wyrmthalak.cpp
```diff
@@ -41,8 +41,13 @@ enum Adds
     NPC_SMOLDERTHORN_BERSERKER      = 9268
 };
 
-const Position SummonLocation1 = { -39.355f, -513.456f, 88.472f, 4.679f };
-const Position SummonLocation2 = { -49.875f, -511.896f, 88.195f, 4.613f };
+enum Misc
+{
+    CALL_HELP   = 0,
+};
```
Winfidonarleyan, line 47: Just use `constexpr` for this

Issue #9088, opened 2021-11-09
UBRS skinning The Beast froze the game


Shineslippers:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59

### Current Behaviour

When you trying to skinning The Beast game froze and you have to do server restart.

https://user-images.githubusercontent.com/85256266/140931814-183e89d5-ea82-4e62-aff7-904987852c4d.mp4

### Expected Blizzlike Behaviour

X

### Source

X

### Steps to reproduce the problem

.additem 12709 - skining dagger
.setskill 173 400 - dagger skill if needed
.setskill 393 400 - skinning
.go c id 10430
.cast 10768

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/12e92b58082dfda8b73b50aa97b98b9479bb7629

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-carbon-copy](https://github.com/55Honey/Acore_CarbonCopy)
[...]

PR #9090, opened 2021-11-09
Fix(core/Creature) initialize AI before calling OnCreatureCreate

patou01:
## Changes Proposed:
-  Move OnCreatureCreate call to the end of the AddToWorld(), so that a user gets a creature that is initialized beforehand.
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9088

## SOURCE:

TC has the same (discovered after trying) https://github.com/TrinityCore/TrinityCore/blob/04909bada85d4770a58a5b2b6f2180f5aacb1b81/src/server/game/Entities/Creature/Creature.cpp#L274
Since the code base has diverged, we still need to check that we're not affecting anything else

## Tests Performed:

- Tested on The beast, he seemed to work. Using a win10 x64 build relwithdebinfo.
- 

## How to Test the Changes:

1. First test is to verify it fixes the issue (steps taken from the issue report)
.additem 12709 - skining dagger
.setskill 173 400 - dagger skill if needed
.setskill 393 400 - skinning
.go c id 10430
.cast 10768

2. Try to find where it could fail. I have no clue there, to me this is logical that it's called at the end (similar to the eluna call)
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -222,13 +222,11 @@ void Creature::AddToWorld()
         // it's also initialized in AIM_Initialize(), few lines below, but it's not a problem
         Motion_Initialize();
 
-        if (GetZoneScript())
-            GetZoneScript()->OnCreatureCreate(this);
-
         GetMap()->GetObjectsStore().Insert<Creature>(GetGUID(), this);
         if (m_spawnId)
+        {
             GetMap()->GetCreatureBySpawnIdStore().insert(std::make_pair(m_spawnId, this));
-
+        }
         Unit::AddToWorld();
 
         SearchFormation();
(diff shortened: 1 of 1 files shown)


Nyeriah:
> @Nyeriah did you test? :P

Doesn't really need testing

acidmanifesto:
> @Nyeriah did you test? :P

Great job. 👍

patou01:
Would still makes sense, just so we're sure it's not a fluke on my side.

Nyeriah:
> Would still makes sense, just so we're sure it's not a fluke on my side.

This is just a script hook that calls the instance script methods, we are just moving the code from one place to another, so nothing really unexpected can come out of this.

Furthermore, as you mentioned, it's the same logic used by TrinityCore.

Moreover, it's the same logic used by Eluna, that also calls the AddedToWorld() hook after AI has been initialized.

This is a crash fix so getting this merged as soon as possible is in all of our best interests.

Issue #9133, opened 2021-11-11
Alterac Valley - Enemy Player Corpses Not Dropping Loot


Azcobu:
### Current Behaviour

Enemy players killed in Alterac Valley BG currently do not drop correct loot. You can take their insignia and loot a small amount of money from them, but none of the items that should drop do.

Note this is similar to https://github.com/azerothcore/azerothcore-wotlk/issues/5261, which is for NPCs in AV.

This was discussed on Trinity, and a solution was proposed in https://github.com/TrinityCore/TrinityCore/issues/3515#issuecomment-124840008. Presumably this could also address #5261.

### Expected Blizzlike Behaviour

Enemy players should drop a range of items:

Alliance drops:
- [Stormpike Soldier's Blood](https://tbc.wowhead.com/item=17306/stormpike-soldiers-blood)
- [Stormpike Soldier's Flesh](https://tbc.wowhead.com/item=17326/stormpike-soldiers-flesh)
- [Stormpike Lieutenant's Flesh](https://tbc.wowhead.com/item=17327/stormpike-lieutenants-flesh)
- [Stormpike Commander's Flesh](https://tbc.wowhead.com/item=17328/stormpike-commanders-flesh)
- [Armor Scraps](https://tbc.wowhead.com/item=17422/armor-scraps)
- [Autographed Picture of Foror & Tigule](https://tbc.wowhead.com/item=18228/autographed-picture-of-tigule) - note the name of this item differs from retail as it was changed only a few months ago.

Horde drops:
- [Storm Crystal](https://tbc.wowhead.com/item=17423/storm-crystal)
- [Frostwolf Soldier's Medal](https://tbc.wowhead.com/item=17502/frostwolf-soldiers-medal)
- [Frostwolf Lieutenant's Medal](https://tbc.wowhead.com/item=17503/frostwolf-lieutenants-medal)
- [Frostwolf Commander's Medal](https://tbc.wowhead.com/item=17504/frostwolf-commanders-medal)
- [Armor Scraps](https://tbc.wowhead.com/item=17422/armor-scraps)
- [Autographed Picture of Foror & Tigule](https://tbc.wowhead.com/item=18228/autographed-picture-of-tigule)

### Source

https://wowpedia.fandom.com/wiki/Alterac_Valley?oldid=2378043
https://classic.wowhead.com/guides/alterac-valley-quests-pvp-battleground-classic-wow

[...]

PR #9198, opened 2021-11-15
feat(Core/Loot): Implement player_loot_template

Nyeriah:
* Also fix AV player loot

## Changes Proposed:
-  Implement player_loot_template table (entry = Team ID)
-  This is a cleaner approch than just hacking a creature_loot_template and sending it and should also allow modules to do cool stuff.
- Implement CONDITION_SOURCE_TYPE_PLAYER_LOOT_TEMPLATE (28), same as all the other loot conditions

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9133

## Tests Performed:

- Tested in-game.
- 

## How to Test the Changes:

1. Join Alterac Valley & check if removing insignia gives you items
2. Join another BG and check if removing insignia gives you items
3.

- [ ] Wintergrasp quest item loot (currently broken) will still need to be fixed (outside the scope of this PR, but should be simple enough)
- [ ]


data/sql/updates/pending_db_world/rev_1636980654557070200.sql
```diff
@@ -0,0 +1,50 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1636980654557070200');
+
+DROP TABLE IF EXISTS `player_loot_template`;
+CREATE TABLE IF NOT EXISTS `player_loot_template` (
+  `Entry` MEDIUMINT UNSIGNED NOT NULL DEFAULT 0,
```
Kitzunu, line 5: can be tinyint

Kitzunu:
Also shouldn't we delete the hacked creature_template entries?

Nyeriah:
> Also shouldn't we delete the hacked creature_template entries?

They don't exist :D
Sunwell did the core stuff but it seems the database counterpart wasn't in the release (select * from creature_loot_template where entry =1; returns nothing)

data/sql/updates/pending_db_world/rev_1636980654557070200.sql
Nyeriah: I'm just afraid that if it's not exactly the same as the other tables we'll have issues during load (the table needs to be an exact copy of the others otherwise creating a entire loading method just for this table would be required :P )

Kitzunu:
kek

src/server/game/Conditions/ConditionMgr.cpp
```diff
@@ -1003,7 +1004,7 @@ void ConditionMgr::LoadConditions(bool isReload)
             cond->ErrorTextId = 0;
         }
 
-        if (cond->SourceGroup)
+        if (cond->SourceGroup || cond->SourceType == CONDITION_SOURCE_TYPE_PLAYER_LOOT_TEMPLATE)
```
Kitzunu, line 1007: why? I guess as 0 (Horde) would count as no SourceGroup

Kitzunu (approved):

src/server/game/Conditions/ConditionMgr.cpp
Nyeriah: Yes, the IF fails because sourcegroup = 0 gets interpreted as false

PR #9217, opened 2021-11-16
fix(Core/Spells): Increased Arcane Bolt effect radius to 40 yards.

### LINKED ISSUE #8930: LBRS Urok Ogre Magus has range issues
### Current Behaviour

If you are in a "deadzone", he casts the spell but it never fires nor seems to hit.

The spell should hit, or the mob move closer.

### Source

_No response_

.go c 40508
.additem 12533
.additem 12534

Start the event, stay far enough from a mage.

### Extra Notes

_No response_

### AC rev. hash/commit

cde621f9ebf0edc012eeed53f0b1e28a0985b909

### Operating system

win10 x64

### Custom changes or Modules

_No response_

UltraNix:
Fixes #8930

## Issues Addressed:

- Closes #8930

## Tests Performed:

- Not tested

## How to Test the Changes:

More info [here](https://github.com/azerothcore/azerothcore-wotlk/issues/8930)


T1ti:
Sorry but that's a bad fix

The proper fix is removing this flag from the SAI cast
image

Here's a vid after removing it, they move and melee attack the player between casts as intended instead of being immobile 
https://i.gyazo.com/d338cf85d5f43d9f3f46efc743628f1e.mp4

reference from Classic https://youtu.be/AY-m_3lGiqg?t=112

UltraNix:
Your problem is something problem. The problem here was the spell  - spell range was 40y, but effect radius was only 20y. So the mage was casting spell, but none got hit.
Just test it.

T1ti:
Ok i'm testing it, but the issue wasn't that the spell didn't it, it was that the mob just deosn't move from his circle

UltraNix:
@T1ti Watch the video from the issue once again. The mag is not mowing, because his target is still in 40y range, so he's casting the spell. But the problem is that the player is not hit by the spell, because the damage effect (which is an AoE effect) is only 20y.

T1ti:
But if you watch the vid, his AOE only hits very close players, it's def not 40 yards

T1ti:
image
The mage deosn't get hit there

T1ti:
but tbh to me it sounds like the targeting shouldn't be "victim", but cast the spell if there is a player within 20 yards

image
 change that to" player in range 20 yards"

T1ti:
Ok finally got it compiled, here's a test of your PR, the NPCs are still bugged and static https://i.gyazo.com/e537676175cf055ce0c12a87b1a27849.mp4

UltraNix:
@T1ti So it's okay now. I can lower that to 20y, it 40y is too big.

T1ti:
tested again, The mobs are still broken, they stand still instead of gonig in melee to attack the player.
This is 100% due to the cast flag

UltraNix:
@T1ti But that PR is not going to fixit and the problem linked to it is not about it.

IntelligentQuantum:
@UltraNix Isn't it better to add it in spell_dbc?

UltraNix:
It's the same.

T1ti:
> @T1ti But that PR is not going to fixit and the problem linked to it is not about it.

well it's also a "range issue" so it can be fixed 2 in one. But the latest fix looks ok to fix the spell not hitting

UltraNix:
But are you sure these mobs should melee attack? I think they should behave just like casters only casting spells.

T1ti:
> But are you sure these mobs should melee attack? I think they should behave just like casters only casting spells.

I posted a reference video above... https://www.youtube.com/watch?v=AY-m_3lGiqg&t=112s

UltraNix:
@T1ti Yep - you are okay with mob. I'm changing that. Thx a lot for that.

UltraNix:
@T1ti Please, test it again.

claudiodfc:
#### Tested and working properly ✔

Nyeriah (approved):

Kitzunu (approved):

PR #9242, opened 2021-11-18
feat(Scripts/Commands): Implement restart/shutdown reasons

Nyeriah:
## Changes Proposed:
-  Add reason param to the server restart/shutdown commands
-  

## Issues Addressed:

- Closes 

## SOURCE:

TrinityCore

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1.Log a character
2. .server restart 120 updates
3. Check that it displays reason ingame

- [ ] Commands can't take more than a single word as param
- [ ]


IntelligentQuantum (approved):

Nyeriah:
@Winfidonarleyan I need those commands to accept the "reason" param as an optional param, which type should I be using? Tail? Can Tail be an <Optional>? Could use some help here, as it currently only prints a single word. I'll run some trial & error later but if there's hints you could share that would save me some time :)

Winfidonarleyan:
> @Winfidonarleyan I need those commands to accept the "reason" param as an optional param, which type should I be using? Tail? Can Tail be an _Optional_? Could use some help here, as it currently only prints a single word. I'll run some trial & error later but if there's any hint you could share that would save me some time :)

Okay, i try make this

src/server/scripts/Commands/cs_server.cpp
```diff
@@ -267,57 +267,57 @@ class server_commandscript : public CommandScript
         return true;
     }
 
-    static bool HandleServerShutDownCommand(ChatHandler* /*handler*/, int32 time, Optional<int32> exitCode)
+    static bool HandleServerShutDownCommand(ChatHandler* /*handler*/, int32 time, Optional<std::string> reason, Optional<int32> exitCode)
```
Winfidonarleyan, line 270:
```suggestion
    static bool HandleServerShutDownCommand(ChatHandler* /*handler*/, int32 time, Optional<int32> exitCode, Tail reason)
```

Winfidonarleyan: @Nyeriah need use `Tail`. This argument must necessarily be at the end, because it covers the whole text

Nyeriah:
> > @Winfidonarleyan I need those commands to accept the "reason" param as an optional param, which type should I be using? Tail? Can Tail be an _Optional_? Could use some help here, as it currently only prints a single word. I'll run some trial & error later but if there's any hint you could share that would save me some time :)
> 
> Okay, i try make this

Thank you very much, this works

PR #9243, opened 2021-11-18
fix(Core/Scripts): add support using whitespace for cs_lookup

Winfidonarleyan:
## Changes Proposed:
- Add support using whitespace for cs_lookup
- Replace args `std::string_view` to `Tail` 

## Tests Performed:
image

## How to Test the Changes:
1. Just use `.look` commands with 2+ word


Nyeriah:
Thanks for the quick fix

Kitzunu:
Will it also work with special characters like `'`? because that was broken too

Winfidonarleyan:
> Will it also work with special characters like `'`? because that was broken too

It should work.

Issue #9003, opened 2021-11-06
General Drakkisath - MOB ID - 10363


Kurthos:
### Current Behaviour

Incorrectly level 62 and not "Boss" level - 
![image](https://user-images.githubusercontent.com/1082884/140612354-1080f0e7-7739-4156-a083-ac4fb0d939b1.png)

When boss resets after a wipe, Chromatic Elite Guards do not properly respawn.

Drakkisath currently spams his Conflagrate ability. permanently CCing your tank. 

### Expected Blizzlike Behaviour

Drakkisath should be a Boss level mob.

Chromatic Elite Guards should respawn if the encounter resets.

There should be a reasonable cast order gap between spells - maybe 15 second cd? Need to find logs for this.

### Steps to reproduce the problem

Engage combat on Drakkisath

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/97effd83d141419c00227833c7887c58e6598113

### Operating system

Windows 10 x64

Kitzunu:
Add hash/commit

Nyeriah:
Some information about conflagration:

The first cast happens after 15 seconds, and then it's repeated every 18 seconds.
It currently targets a random player.

patou01:
Can I mention an additional issue with the adds here @Nyeriah ? The adds are linked to each other but slow to aggro drakkisath.

https://drive.google.com/file/d/1N3v6NXwSdFhDjHwajvzXX_W3sUQnBjuW/view?usp=sharing

Nyeriah:
> Can I mention an additional issue with the adds here @Nyeriah ? The adds are linked to each other but slow to aggro drakkisath.
> 
> https://drive.google.com/file/d/1N3v6NXwSdFhDjHwajvzXX_W3sUQnBjuW/view?usp=sharing

I will rework that with creature_formations and use minion data so they respawn on wipes, but that part might be interesting for you to research since you were looking into that a while ago - the call happens on EnterCombat(), so it seems EnterCombat() is not called on aggro right away.

Nyeriah:
Scrap that, I found this while looking:

![Skärmbild 2021-11-06 201620](https://user-images.githubusercontent.com/47818697/140626604-8a7a0e8e-f99c-48f1-8274-3dc6ed7e5d9e.jpg)

His script used the same creature ID when trying to search for him & the guards...

T1ti:
[...]

PR #9299, opened 2021-11-23
fix(Scripts/UBRS): Drakkisath's conflagrate should drop threat tempor…

Nyeriah:
…arily

also adjust it's timer slightly

## Changes Proposed:
-  Reset threat temporarily for Conflag. targets
-  Adjust timer according to feedback

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9003

## Tests Performed:

- Tested in-game.
- 

## How to Test the Changes:

1. engage drakkisath with 2 characters
2. notice threat drops when he casts conflag and he goes to the next target
3. notice that threat is readded once conflag expires (according to feedback it should be temporary)

- [ ]
- [ ]


Nefertumm (approved):

Nefertumm:
Tested and working as expected.

T1ti:
Isn't target switching when applying a CC breakable by dmg global mob AI rather than ahrdcoded threat drop ?

Nyeriah:
> Isn't target switching when applying a CC breakable by dmg global mob AI rather than hardcoded threat drop ?
> 
> The same behavior should happen when any mob applies a polymorph or whatever, they ignore the cc'd player and switch to secondary

This isn't a breakable CC.

T1ti:
Confuse should be the same as disorient afaik, it's just the aura effect that is different  and it does break on DMG.

Confusion was renamed to disorient in patch 1.9 https://wowpedia.fandom.com/wiki/Disorient 

Dragon's breath also uses spell aura Confuse.

T1ti:
Also Blizz would have 100% included the threat drop in a spell effect, unless there's a dummy/script effect, in this case it could be coded

Nyeriah:
Conflagrate:
Skärmbild 2021-11-25 053338

Dragon's breath:
Skärmbild 2021-11-25 053454

Right now, you can take as many hits and Conflagrate never breaks. So as I see it, it's the MECHANIC_DISORIENT that breaks on damage and not the spell itself, it has nothing to do with the aura but the mechanic that it applies. It could be that the core is doing it wrong, but that's how it is right now.

T1ti:
Yeah drakkisath's conflag doesn't seem to apply any mechanic effect 
I'll look into it more, maybe there's an hidden server side spell that drops threat

Plasmashield:
Threat shouldn't reset temporarily. You shouldn't get back your threat after conflagration over

Nyeriah:
> Threat shouldn't reset temporarily. You shouldn't get back your threat after conflagration over

Please open an issue with the sources for your claim and I'll look into it

Plasmashield:
> > Threat shouldn't reset temporarily. You shouldn't get back your threat after conflagration over
> 
> Please open an issue with the sources for your claim and I'll look into it

Oky. But i don't understand why it was implemented like this without any sources

Nyeriah:
> > > Threat shouldn't reset temporarily. You shouldn't get back your threat after conflagration over
> > 
> > 
> > Please open an issue with the sources for your claim and I'll look into it
> 
> Oky. But i don't understand why it was implemented like this without any sources

I was told by a retail player this is how it should work, and this is how this kind of mechanic usually works. If you think it's wrong you're free to make your claim, not a problem to me at all.

Plasmashield:
So... Just someone told it to you? The same as it was with Elemental Focus? Is it enough that someone just said it to you to change it? By the fact that in 5 minutes of searching for a video, you can find how it actually works on a classic for now.

Nyeriah:
> So... Just someone told it to you? The same as it was with Elemental Focus? Is it enough that someone just said it to you to change it? By the fact that in 5 minutes of searching for a video, you can find how it actually works on a classic for now.

Please find such evidence and post it as already suggested. As I said, I have no issues reworking this if it's not how it should work, but there's no room for drama in projects like this, you were already told what needs to be done.

Issue #9329, opened 2021-11-25
[MC] Lava Surger: Surge not working correctly


temperrr:
### Current Behaviour

Surge targets a random target in range

### Expected Blizzlike Behaviour

Surge should target the furthest target on its threat list 

Clarifying : he should target the furthest player by distance that has threat, not the lowest threat.

### Source

https://www.youtube.com/watch?v=gFyl6G-PoUM
https://wowpedia.fandom.com/wiki/Lava_Surger

### Steps to reproduce the problem

1. go to MC
2. Have multiple chars
3. Engage a surger
4. all spread out
5. all damage him
5. he should target the furthest threat list target, but he doesn't.

### AC rev. hash/commit

3b1b80cbaae3

### Operating system

Unix

### Custom changes or Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
- [lua-carbon-copy](https://github.com/55Honey/Acore_CarbonCopy)
- [lua-custom-worldboss](https://github.com/55Honey/Acore_CustomWorldboss)
- [lua-level-up-reward](https://github.com/55Honey/Acore_LevelUpReward)
- [lua-recruit-a-friend](https://github.com/55Honey/Acore_RecruitAFriend)
[...]

PR #9338, opened 2021-11-25
fix(DB/SAI): Lava Surger should target the farthest player

Nyeriah:
## Changes Proposed:
-  Lava Surgers should target the player farthest away instead of any random target
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9329

## Tests Performed:

- Tested in-game
- 

## How to Test the Changes:

1. Log 3 characters
2. Engage Lava Surgers
3. Position the characters in a way they are at different ranges, notice the surger will prefer to hit the one farthest away

Keep in mind the spell has a 40 yard range, so if the farthest away one is out of range the other will be chosen instead.
Keep in mind they're level 60 mobs, if the spell misses there won't be any charge, so be level 60 when testing this.

- [ ]
- [ ]


Kitzunu (approved):

0xCiBeR:
Tested, Lava Surger now correctly targets the farthest away player in a 40 yard range 👍

Nyeriah:
> Tested, Lava Surger now correctly targets the farthest away player in a 40 yard range 👍

Thanks for testing!

PR #9375, opened 2021-11-27
fix(DB/Pathing) Add missing Springpaw pathing in Sunstrider Isle

malcrom:
## Changes Proposed:
-  Add missing pathing

## SOURCE:
TC but corrected follow angle

## Tests Performed:
- Installs and checked pathing in game 
- 
## How to Test the Changes:
1. No pathing currently exist for these creatures in AZ but do on Official.


Nefertumm (approved):

savowe:
Looks great!

- [x] followed the path for all three lynx/cub pairs
- [x] killed single cats and watched if they would regroup if one is respawned

PR #9389, opened 2021-11-27
fix(DB/Creature) Add missing random movement to Sarkoth

malcrom:
## Changes Proposed:
-  Creature Sarkoth has no movement in AZ. Added random movement.

## SOURCE:
My eyes

## Tests Performed:
Compared to Official. Looks close enough. How can I know Random movement is 100% correct?


acidmanifesto:
> How can I know Random movement is 100% correct?
My opinion is the fact that the is no repeating waypoints or a constant consistent pattern. Waypoints also tend to be always consistent after the creature is spawned\respawned. So far with Sarkoth, the 3 times killed on retail, the npc does not move to the same spot consistently as its first waypoint, while also still maintaining its consistent spawn spot.

Nyeriah (approved):

Issue #9438, opened 2021-12-01
LFG/RDF Teleport Allowing Exit From Dungeon While In Combat


Azcobu:
### Current Behaviour

It is currently possible to use the LFG/RDF teleport out of a dungeon while in combat. This is being used to grief players - see support ticket 690 on the CC Discord.

Example:

Our designated griefer is Aith - he pulls some mobs, then uses RDF teleport to exit the dungeon:
![WoWScrnShot_120121_235302](https://user-images.githubusercontent.com/81782124/144242645-e481a1ec-666d-413f-98fe-f7828342abc4.jpg)

They all aggro onto the other party member. Fortunately, escape is equally easy:

https://user-images.githubusercontent.com/81782124/144242566-6afeb395-45bd-415e-b55c-4e5c7e314499.mp4

### Expected Blizzlike Behaviour

Telporting in or out of a dungeon via LFG/RDF should be blocked during combat.

### Source

"Players may teleport in and out of the dungeon at any time **if they are not in combat.**" - https://wowpedia.fandom.com/wiki/Dungeon_Finder?oldid=2357590 - from August 2010

### Steps to reproduce the problem

1. .debug lfg
2. Join solo or with someone to grief.
3. Get into combat, then use the LFG/RDF icon on the minimap to teleport out.

### AC rev. hash/commit

94c0a3a72996

### Operating system

Linux Mint 20.2

### Custom changes or Modules

N/A.

PR #9466, opened 2021-12-02
fix(Core/LFG): Prevent leaving LFG if in combat

Eronox:
## Changes Proposed:
-  Emit `LFG_TELEPORTERROR_COMBAT `(was `LFG_TELEPORTERROR_CHARMED`) if player is in combat when trying to leave LFG
-  Rename `LFG_TELEPORTERROR_CHARMED ` to  `LFG_TELEPORTERROR_COMBAT  `since the message is shown as below

> You can't do that while in combat

## Issues Addressed:

- Closes  https://github.com/azerothcore/azerothcore-wotlk/issues/9438

## SOURCE:

"Players may teleport in and out of the dungeon at any time if they are not in combat." - https://wowpedia.fandom.com/wiki/Dungeon_Finder?oldid=2357590 - from August 2010

## Tests Performed:

- Message is shown and you don't get removed from group/dungeon if in combat attempting to leave
image

- Build Pass
> ========== Build: 4 succeeded, 0 failed, 14 up-to-date, 1 skipped ==========

## How to Test the Changes:

1. `.debug lfg`
2. Join solo or with someone to grief.
3. Get into combat, then use the LFG/RDF icon on the minimap to teleport out.

- See `LFGMgr.h`, when/if the **existing** todo is completed it's necessary to make sure the message output isn't changed for this scenario

> LFG_TELEPORTERROR_COMBAT                     = 8       // FIXME - It can be 7 or 8 (Need proper data)

### DIFF
--- a/src/server/game/DungeonFinding/LFGMgr.cpp
+++ b/src/server/game/DungeonFinding/LFGMgr.cpp
@@ -2104,15 +2104,25 @@ namespace lfg
         LfgTeleportError error = LFG_TELEPORTERROR_OK;
 
         if (!player->IsAlive())
+        {
             error = LFG_TELEPORTERROR_PLAYER_DEAD;
+        }
         else if (player->IsFalling() || player->HasUnitState(UNIT_STATE_JUMPING))
+        {
             error = LFG_TELEPORTERROR_FALLING;
+        }
         else if (player->IsMirrorTimerActive(FATIGUE_TIMER))
+        {
             error = LFG_TELEPORTERROR_FATIGUE;
+        }
         else if (player->GetVehicle())
+        {
             error = LFG_TELEPORTERROR_IN_VEHICLE;
-        else if (player->GetCharmGUID())
-            error = LFG_TELEPORTERROR_CHARMING;
+        }
+        else if (player->GetCharmGUID() || player->IsInCombat())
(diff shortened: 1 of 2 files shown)


Nyeriah:
LGTM, the messages are hardcoded client side so there's no error to this, also, charmed state (mind control) should keep them in combat, not sure if it does right now but it's good to keep it like this

Nyeriah (approved):

src/server/game/DungeonFinding/LFGMgr.cpp
```diff
@@ -2104,15 +2104,25 @@ namespace lfg
...
+        {
             error = LFG_TELEPORTERROR_FATIGUE;
+        }
         else if (player->GetVehicle())
+        {
             error = LFG_TELEPORTERROR_IN_VEHICLE;
-        else if (player->GetCharmGUID())
-            error = LFG_TELEPORTERROR_CHARMING;
+        }
+        else if (player->GetCharmGUID() || player->IsInCombat())
```
Eronox, line 2122: I don't know if you can get charmed without being put in combat, but if that's the case then the message thrown has always been wrong. 

Keeping as is just incase, rather than only using  `player->IsInCombat()`

PR #9483, opened 2021-12-03
feat(Core/Player): Implement player specific settings

Nyeriah:
## Changes Proposed:
-  Implement a feature to allows saving specific character settings, that can expand certain server features and can also be used by modules to store character specific options.
-  Implement a command to allow disabling/enabling battleground/arena queues, as well as autobroadcast messages.
- Implement two new world configs, to enable this system and to decide whether players should be allowed to disable autobroadcast or not (can do this by setting min required level).

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested in-game.
- 

## How to Test the Changes:

1. Have CFBG OFF (important, they use their own queue announcer, needs to be updated)
2. Enable queue announcers (world.config file)
3. run .settings announcer bg off
4. queue to a bg, notice the announcer won't fire

- [ ]
- [ ]


Nyeriah:
Data is stored like this:
|  GUID | SOURCE   | DATA   | 
|---|---|---|
|  1 |0   | 0 1 1   |  
|   1| 1  | 0 1   | 
|   1| 2  |  3 2 |  

Where SOURCE 0 = AC default (configs that extend the core's default features).
If anyone intends to use this inside a module, all that they need to do is register a new source in their module files.

And then, for example:
```cpp
player->UpdatePlayerSetting(SOURCE_TRANSMOG, TRANSMOG_OPTIONAL_FLAG, enabled);
```

src/server/game/Entities/Player/PlayerSettings.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (result)
+    {
+        do
+        {
+            Field* fields = result->Fetch();
+
+            uint32 source = fields[0].GetUInt32();
+            std::string data = fields[1].GetString();
+
+            Tokenizer tokens(data, ' ');
```
Winfidonarleyan, line 42: Use new Tokenizer

src/server/game/Entities/Player/PlayerSettings.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+            PlayerSettingVector setting;
+            setting.resize(tokens.size());
+
+            uint32 count = 0;
+
+            for (auto token : tokens)
+            {
+                PlayerSetting set;
+                set.value = atoi(token);
```
Winfidonarleyan, line 52: StringConvert

src/server/game/World/World.cpp
```diff
@@ -2570,6 +2573,34 @@ void World::SendWorldText(uint32 string_id, ...)
...
 }
 
+void World::SendWorldTextOptional(uint32 string_id, uint32 flag, ...)
+{
+    va_list ap;
+    va_start(ap, flag);
+
+    Acore::WorldWorldTextBuilder wt_builder(string_id, &ap);
+    Acore::LocalizedPacketListDo<Acore::WorldWorldTextBuilder> wt_do(wt_builder);
+    for (auto itr : m_sessions)
```
Winfidonarleyan, line 2583: Auto const&

src/server/scripts/Commands/cs_player_settings.cpp
```diff
@@ -0,0 +1,81 @@
...
+            {
+                handler->SetSentErrorMessage(true);
+                handler->PSendSysMessage(LANG_CMD_AUTOBROADCAST_LVL_ERROR, sWorld->getIntConfig(CONFIG_AUTOBROADCAST_MIN_LEVEL_DISABLE));
+            }
+
+            on ? setting.RemoveFlag(ANNOUNCER_FLAG_DISABLE_AUTOBROADCAST) : setting.AddFlag(ANNOUNCER_FLAG_DISABLE_AUTOBROADCAST);
+            player->UpdatePlayerSetting(SETTING_SOURCE_AC, SETTING_ANNOUNCER_FLAGS, setting.value);
+        }
+
+        handler->SetSentErrorMessage(false);
```
Winfidonarleyan, line 72: no need this line if `return true`

Nyeriah:
@Yehonal

Yehonal:
Amazing feature!!!

We should do something similar for the server configurations IMHO. Allowing modules to store confs in the database when needed

SUGGESTION: What if we use the JSON field instead of the text?

https://github.com/azerothcore/azerothcore-wotlk/discussions/8428

That would help us to extend even more this functionality in future by creating indexable options and allowing better parsing by web applications

Yehonal:
Another note: the "source" should be a string IMHO. It would avoid collisions between modules just forcing them to use the module name. Let's say this scenario:

1. Module "Private Custom Something" uses source 14 for its own conf, but they can't share it
2. Module "Open New System" uses source 14 for its own conf and they share it in the public list
3. Now the "Private Custom Something" is in conflict with the other one and they can't be installed in the same system

Of course, we can offer some high-enough "ranges" for public and private modules, but it's an ugly and limited solution that still doesn't solve the collisions

While it's nearly impossible that if you call your source key in a "smart" way using a good namespace such as: `mod_azerothshard_pvp_something` , most likely no one else would use your organization name together with the name of the module itself (similar to the npm strategy, or the Wordpress plugins)

Nyeriah:
> Another note: the "source" should be a string IMHO. It would avoid collisions between modules just forcing them to use the module name. Let's say this scenario:
> 
> 1. Module "Private Custom Something" uses source 14 for its own conf, but they can't share it
> 2. Module "Open New System" uses source 14 for its own conf and they share it in the public list
> 3. Now the "Private Custom Something" is in conflict with the other one and they can't be installed in the same system
> 
> Of course, we can offer some high-enough "ranges" for public and private modules, but it's an ugly and limited solution that still doesn't solve the collisions
> 
> While it's nearly impossible that if you call your source key in a "smart" way using a good namespace such as: `mod_azerothshard_pvp_something` , most likely no one else would use your organization name together with the name of the module itself (similar to the npm strategy, or the Wordpress plugins)

I've considered this. I just wasn't sure if a search by string would be acceptable, I think it solves our problems.

claudiodfc:
@Nyeriah missing some SQL table update perhaps

```
In mysql_stmt_prepare() id: 453, sql: "SELECT source, data FROM character_settings WHERE guid = ?"
Table 'acore_characters.character_settings' doesn't exist
In mysql_stmt_prepare() id: 454, sql: "REPLACE INTO character_settings (guid, source, data) VALUES (?, ?, ?)"
Table 'acore_characters.character_settings' doesn't exist
In mysql_stmt_prepare() id: 455, sql: "DELETE FROM character_settings WHERE guid = ?"
Table 'acore_characters.character_settings' doesn't exist
```

claudiodfc:
### Tested and working. 👍🏽  

#### Notes
using command `.settings announcer bg off` displays message `<error>` and toggles the settings off
using command `.settings announcer battleground off` displays message `<error>` and settings remain off

after using the command off 3 times, i got the ´<error>´ message and after i turn them on, I got the following message:
image

claudiodfc:
Got this message on the worldserver console:

Acore string entry 5073 not found in DB.

Nyeriah:
Thanks for testing!

Nefertumm:
Still wip?

Nefertumm (approved):

data/sql/updates/pending_db_characters/rev_1638643807174948000.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO `version_db_characters` (`sql_rev`) VALUES ('1638643807174948000');
+
+DROP TABLE IF EXISTS `character_settings`;
+CREATE TABLE IF NOT EXISTS `character_settings` (
+  `guid` INT UNSIGNED NOT NULL,
+  `source` TEXT NOT NULL,
+  `data` TEXT NULL,
+  PRIMARY KEY (`guid`, `source`)
```
Nefertumm, line 8: You should specify a length to `source` in order to make it primary key. Use varchar instead of text for that

Nyeriah:
Didn't have time to learn how to do the JSON thing & to do the required changes after that

Nyeriah:
After much pondering I decided I won't be using JSON for this. The reason is that the select query needs to be static, and we can't do that if we don't know the atributes we'll be using, so each module would need their own load script, which, at this time, sounds counterproductive to me.

Nyeriah:
Queries fixed, tested again and everything's working. Ready to sail.

Nyeriah:
https://github.com/azerothcore/mod-cfbg/pull/88

PR #9484, opened 2021-12-04
core(misc/conf): change default option & add description for setting Instance.SharedNormalHeroicId

Voxstrasza:
## Issues Addressed:

Replaces default option for Instance.SharedNormalHeroicId. The current default setting enables a total of 4 lockouts for ICC10/25 Normal/Heroic and Ruby Sanctum10/25 Normal/Heroic, when blizzlike is two (normal/heroic share lockouts).  This setting being changed to Enabled as default mimics what blizzard intended for ICC and it allows players to (as an example) kill the first 3 bosses on Normal, and then swap to Heroic mode for Gunship (if they have heroic unlocked by killing normal Lich King) and then back to Normal - without leaving the instance.

The setting does not effect Trial of the Crusader, which will remain 4 separate lockouts (10 normal, 10 heroic, 25 normal, 25 heroic), nor does it effect earlier raids that do not have a Heroic setting (Naxx, Ulduar, etc)

Also added description for this setting.

## SOURCE:

https://wowpedia.fandom.com/wiki/Icecrown_Citadel_(instance)

`Icecrown Citadel is home of the updated instance difficulty setting which can be changed on a boss per boss basis. It is also able to be switched at any time during a raid without leaving and reentering the instance. `

https://www.mmo-champion.com/threads/731139-Ruby-Sanctum-lockout?p=8219455&viewfull=1#post8219455

Note that any date after October 12, 2010 is the Cataclysm pre-patch and the lockout systems were completely redesigned.  If you search for more info/proof on this, any date after Oct 12 2010 is referring to the Cataclysm lockout system.

## Tests Performed:

I've had this setting enabled for two years.

## How to Test the Changes:

1. You should still be able to do Trial of the Crusader 4 times a lockout (10 normal, 10 heroic, 25 normal, 25 heroic)
2. You should only be able to do Icecrown Citadel 2 times a lockout (10 normal or heroic, 25 normal or heroic)
3. You should be able to swap between heroic and normal ICC 10/25 if you have heroic unlocked
4. Ruby Sanctum follows the ICC lockout system: repeat 2 and 3 for RS.


Nyeriah (approved):

AggrDog:
blocking two modes when passing one difficulty level still works. Gone 10 normal mode. got CD for heroic

Issue #9408, opened 2021-11-29
Gnomish Battle Chicken not procing any effects


IntelligentQuantum:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59

### Current Behaviour

Original Post: https://github.com/chromiecraft/chromiecraft/issues/2429

The trinket - https://wowgaming.altervista.org/aowow/?item=10725
The chicken - https://wowgaming.altervista.org/aowow/?npc=8836
Haste proc - https://wowgaming.altervista.org/aowow/?spell=23060
Enrage proc - https://wowgaming.altervista.org/aowow/?spell=13168

Currently, from my experience using the trinket all day today and yesterday, I have yet to see either the enrage proc or the haste proc activate.

The chicken is currently hitting for 22 damage a hit w/o enrage.

### Expected Blizzlike Behaviour

The trinket - https://wowgaming.altervista.org/aowow/?item=10725
The chicken - https://wowgaming.altervista.org/aowow/?npc=8836
Haste proc - https://wowgaming.altervista.org/aowow/?spell=23060
Enrage proc - https://wowgaming.altervista.org/aowow/?spell=13168

From my experience in Vanilla Classic and TBC Classic - as long as the chicken got good auto attack up time you would expect most uses of the trinket to proc the enrage, the haste proc however was a little less common, and it's hard to recall from memory since we often popped our battle chickens together to guarantee a proc.

WoWHead comments from the time are inconsistent on the proc rate, some people say 50%, some people say once every 5-6 uses (15-20%), some people reference getting procs every in dungeons. It's difficult to know exactly because it depends on the uptime of the chicken's auto attack, how quickly it enrages (attacking quicker, for a more likely proc), and other factors.

You would also expect the chicken to auto attack for >10 w/o enrage proc.

Recent experimentation on TBC Classic has brought up a theory that being hit by a hostile action, guarantees a enrage and haste proc instantly on summon, I've included a video on this theory in sources. This behaviour is not happening on ChromieCraft.

### Source

4.3.0 100% proc claim: https://www.wowhead.com/item=10725/gnomish-battle-chicken#comments:id=1603075
3.3.2 50% proc claim: https://www.wowhead.com/item=10725/gnomish-battle-chicken#comments:id=1003276:reply=155451
[...]

PR #9485, opened 2021-12-04
fix(DB/SAI): Add Gnomish Battle Chicken combat SAI

Voxstrasza:
## Changes Proposed:
-  Battle Chicken casts Battle Squawk w/ 25% chance in combat on self
-  Battle Chicken casts Chicken Fury w/ 50% chance in combat on self

## Issues Addressed:
#9408

## Tests Performed:
Untested

## How to Test the Changes:
1. Use trinket Gnomish Battle Chicken
2. Attack mob with Chicken out
3. Wait for procs.
4. Invite some friends to see if they get the Haste buff.

I pulled 25% and 50% out of my ass.  There is no concrete information on the proc rates - but the Haste proc has historically procced less than the Enrage proc.


Nyeriah (approved):

Voxstrasza:
@Nyeriah I didn't test - did anyone confirm testing to you?

Kitzunu:
90% of the time sai dont need to be tested as long as the comments are good

Nyeriah:
I checked the database and saw they had no scripts at all whatsoever. As for the queries themselves, as Kitzunu said, there's usually no reason to expect any adverse effects from SAI - it just does what the numbers tell it to do

Nyeriah:
There's only one thing here I would change: that spell already has TARGET_UNIT_CASTER_AREA_PARTY as target type, so setting it to be cast on the invoker's party is not necessary, the amount of targets is also not necessary. It works fine with self as target, since it's an aoe spell, the spell target selectors will take care of the rest.

Voxstrasza:
> There's only one thing here I would change: that spell Squawk already has TARGET_UNIT_CASTER_AREA_PARTY as target type, so setting it to be cast on the invoker's party is not necessary, the amount of targets is also not necessary. It works fine with self as target, since it's an aoe spell, the spell target selectors will take care of the rest.

i'll edit it, thanks for the feedback

Eronox:
If I'm not mistaken, I believe the event type should probably be `UPDATE_IC` (id: 0) instead of `AGGRO` (id: 4), because otherwise the spells may only be cast on aggro with the given chance instead of during combat

KaneoHunter:
https://www.youtube.com/watch?v=TSywF4MrPBQ

I wanted to make note of this recent revelation, we might want to account for this behaviour?

PR #9538, opened 2021-12-06
feat(Scripts/Commands): Item restoration command

Eronox:
## Changes Proposed:
-  New command: `.item restore list <playerName>`
-  New command: `.item restore <restoreId> <playerName>`
-  Move `.itemmove` wip command into this new context

One will need to retrieve the id from the `recovery_Table` in order to restore the item. 
That's meant to be achieved by using the restore list command, then using the given id (not item id) in the restore command

This is for a more robust systematical performance and to easier adapt it to SOAP

## Tests Performed:

- Build works
- SQL queries gets applied, can get re-applied without errors
- Tested ingame
- Tested through console

## How to Test the Changes:

1. Enable either `ItemDelete.Method` or `ItemDelete.Vendor` in `worldserver.conf`

Make sure to pay attention to the `ItemDelete.Quality `and `ItemDelete.level` configs incase you want to try selling/deleting some lvl 5 milk or something

2. Restart worldserver if it's already on
3. Login, buy something from a vendor, sell it or destroy it
4. Run `.item restore list <characterName>`
5. Run `.item restore <id from list> <characterName>`
6. Visit mailbox to see a mail returning the item by the postmaster

You can also do the same steps in console, meaning you don't have to be online to run the commands or to get a mail sent to you.

- Garbage recycle _recovery_items_ table, like a config setting expiration time before it can't be recovered any longer


src/server/scripts/Commands/cs_item.cpp
```diff
@@ -0,0 +1,180 @@
...
+        char const* param1 = strtok((char*)args, " ");
+        if (!param1)
+            return false;
+
+        char const* param2 = strtok(nullptr, " ");
+        if (!param2)
+            return false;
+
+        uint8 srcSlot = uint8(atoi(param1));
+        uint8 dstSlot = uint8(atoi(param2));
```
Nyeriah, line 152:
```suggestion
    static bool HandleItemMoveCommand(ChatHandler* handler, uint8 srcSlot, uint8 dstSlot)
    {
```

Nyeriah (approved):

PR #9543, opened 2021-12-06
fix(Core/Spell): Core not handling AURA_INTERRUPT_FLAG_TELEPORTED

### LINKED ISSUE #9542: Core not handling spell aura interrupt flag AURA_INTERRUPT_FLAG_TELEPORTED
### Current Behaviour

If an aura is applied from a spell with the aura interrupt flag `AURA_INTERRUPT_FLAG_TELEPORTED`, it does **NOT** get removed when the wearer is teleported.

This flag is **DIFFERENT** from `AURA_INTERRUPT_FLAG_CHANGE_MAP` which only triggers when the target enters a new map (teleporting to different map, logging in...). Actually, `CHANGE_MAP` flag is **correctly handled** by AC.

The flag `AURA_INTERRUPT_FLAG_TELEPORTED` exists in spells and it might be handleable by core (spoiler: [yes it is possible to handle it](https://youtu.be/bvTgvyiQwhA)).

Here is a video with custom spells created specifically to test this. One spell with aura interrupt flag CHANGE_MAP and the other with aura interrupt flag TELEPORTED to show the difference, and how one is currently correctly handled while the other one is not: https://youtu.be/QI_32qTx1SU

Auras applied from spells with `AURA_INTERRUPT_FLAG_TELEPORTED` should be removed **when wearer gets teleported**, regardless if it's on same/different map. However, it must not remove the aura when entering a map (such as when a character is logged in) unlike auras from spells with aura interrupt flag `CHANGE_MAP`. This should trigger on teleport only.

Here is a video of the same situation but with the **problem fixed** by handling this flag on teleportation: https://youtu.be/bvTgvyiQwhA

### Source

_No response_

1. Cast a spell that applies an aura and has flag `AURA_INTERRUPT_FLAG_TELEPORTED` ([spell 42146](https://wowhead.com/spell=42146) makes your character slowly walk forward and unstoppable)
2. Teleport yourself (e.g. use `.tele <location name>` gm command)
3. (do `.unaura all` if your character is still moving forward after teleporting because aura did not get removed on teleport)

### I made an edit to fix this and am currently working on making a PR for it (:
Still made this issue report if anyone else wants to work on it in a different way than me, and to discuss of it.

### AC rev. hash/commit

* AzerothCore revision            : 1b5c4d82dba7 2021-12-06 19:39:04 -0300 (azerothcore branch)
* AzerothCore buildtype           : RelWithDebInfo

### Operating system

Ubuntu 20.04.3

### Custom changes or Modules

  |   +- mod-autobalance
  |   +- mod-bg-reward
  |   +- mod-cfbg
  |   +- mod-character-tools
  |   +- mod-eluna-lua-engine
  |   +- mod-learn-spells
  |   +- mod-npc-buffer
  |   +- mod-pvp-titles
  |   +- mod-skip-dk-starting-area
  |   +- mod-solocraft
  |   +- mod-transmog

TheLegendaryKingdom:
Actually auras from spells with `AURA_INTERRUPT_FLAG_TELEPORTED` are **not removed
on teleportation**. This change makes it handled at teleportation on same map **AND**
if map changes, but it does not remove the aura when entering world to avoid
such auras being removed at character login.

Closes AzerothCore issue #9542

## Changes Proposed:
- Remove auras with interrupt flag `TELEPORTED` **whenever the wearer is teleported**
- This should not remove such auras on entering world (like at character login) but only on teleportation (flag name is explicit)

## Issues Addressed:

- Closes #9542

## Tests Performed:

- Built correctly
- Tested with custom spells specifically created with the flags TELEPORTED and CHANGE_MAP to see the difference.
- Here is a video before this PR: https://youtu.be/QI_32qTx1SU
- Here is after the fix: https://youtu.be/bvTgvyiQwhA

## How to Test the Changes:

1. Cast a spell that applies an aura and has flag `AURA_INTERRUPT_FLAG_TELEPORTED` ([spell 42146](https://wowhead.com/spell=42146) makes your character slowly walk forward and unstoppable)
2. Teleport yourself (e.g. use `.tele <location name>` gm command) within the same map and then to another map (passing through dark portal changes map)
3.

- [ ]
- [ ]


src/server/game/Entities/Player/Player.cpp
```diff
@@ -1417,6 +1417,9 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
         if (!(options & TELE_TO_NOT_LEAVE_COMBAT))
             CombatStop();
 
+        //remove auras before teleporting...
+        RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_TELEPORTED | AURA_INTERRUPT_FLAG_MOVE | AURA_INTERRUPT_FLAG_TURNING);
```
UltraNix, line 1421: I don't thinks it's a good change. It will remove auras with these flags on teleport on the same map, e.g. blink or shadowstep.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -1494,7 +1497,7 @@ bool Player::TeleportTo(uint32 mapid, float x, float y, float z, float orientati
                     InterruptNonMeleeSpells(true);
 
             //remove auras before removing from map...
-            RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_CHANGE_MAP | AURA_INTERRUPT_FLAG_MOVE | AURA_INTERRUPT_FLAG_TURNING);
+            RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_CHANGE_MAP | AURA_INTERRUPT_FLAG_TELEPORTED | AURA_INTERRUPT_FLAG_MOVE | AURA_INTERRUPT_FLAG_TURNING);
```
UltraNix, line 1500: As I said earlier - that flag is not related to teleport.

UltraNix (changes requested):

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: Then I could make it removed only  if teleport to more than a specified distance (like 100) to avoid it being removed from spells such as blink ?

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: I ensure you it is a different flag than CHANGE_MAP, if you watch the two videos you'll spot it (:
More, its name is explicit imo, and it is recognized by the core and correctly doing its job when I use it.

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: More more:
I Will look forward in dbcs to find out if such spells have this aura interrupt flag, which seems at my sense illogic

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: @UltraNix after checking within Stone Harry's Spell Editor, these spells do not have aura interrupt flag TELEPORTED (se this [video](https://youtu.be/lOW207cbQME) : https://youtu.be/lOW207cbQME)

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: @UltraNix I moved it to the clause where it will only be removed if teleports on another map OR teleport at least 100 yards away.
I overall deleted the other flags I should not remove in this case.

Sorry for my confusion about blink and shadowstep, if you read my previous messages before I deleted them. I did not understand and thought you were saying this edit was gonna remove blink aura or shadowstep aura, I was really confused about it (x
I see better now, you were talking about blink or shadowstep were gonna remove auras they probably shouldn't. So it is avoid by the 100yards condition then (neither blink or shadowstep can send the player so far away).

src/server/game/Entities/Player/Player.cpp
TheLegendaryKingdom: @UltraNix I reverted this part of the PR. You're right, the case here might not be regular teleport. It seems more related to map id and/or map instance. Did not totally understand, but stopped messing with it.

Then the PR now is only one line (:

After testing it with my two spells it works correctly. 

STILL: as I only remove auras with interrupt flag TELEPORTED and no longer the others, should I keep int under the 100 yards condition ? Or is it better to be absolute about teleport and remove auras with interrupt flag TELEPORTED just above the 100 yards condition ? (second option is more logic to me, as the flag name only mentions teleporting, not "teleporting far" or whatever)

Here is one more video with the previous commit, removing auras with flag TELEPORTED within the 100 yards condition:
https://youtu.be/5Qe81YA63Ec

I'll try outside the 100 yards condition and post another video of it.

TheLegendaryKingdom: @UltraNix outside the 100 yards condition, so removes auras with this interrupt flag whenever teleporting:
https://youtu.be/_dCTi4kMsqs

UltraNix:
Still not good. Now you are using that on every teleport. So e.g. if you have Frost Nova on self and then use Shadowstep, Frost Nova is removed - you think it's okay?

TheLegendaryKingdom:
> Still not good. Now you are using that on every teleport. So e.g. if you have Frost Nova on self and then use Shadowstep, Frost Nova is removed - you think it's okay?

No it's not, you're totally right. I moved it again under the 100 yards condition above, so it only removes auras with this flag if teleports more than 100yards away. It should avoid shadowsteplike and blinklike spells to do this

UltraNix (approved):

PR #9593, opened 2021-12-08
fix(Core/Spell) Green Beam spells should only hit one target

malcrom:
## Changes Proposed:
-  Limit  Green Beam spells to one target

## SOURCE:
Official and sniff

## Tests Performed:
- Before this spell hit all entry in range now hits one.

## How to Test the Changes:
1. Working perfectly atm. SAI is complete but missing spawns needed for testing not yet pushed.


Eronox:
Is this used by a toy or something? I can't find sources around where it's used, but it seems like it's a toy since it shouldn't be used in combat 

https://tbc.wowhead.com/spell=31628/green-beam

Eronox:
ok i got info about it's being used in belf starting zone :)

> small crystals around the room cast beam on crystal on center, but I think there's one on the second floor with the same entry

malcrom:
image

malcrom:
Working on SAI and missing spawns.

Nyeriah (approved):

Javilico92:
For me beams get interrupted after few seconds and the it recast again and again, its just me ? Tested both beams (tested on BCC and it get interrupted after some minutes)
WoWScrnShot_120821_214045

malcrom:
> For me beams get interrupted after few seconds and the it recast again and again, its just me ? Tested both beams (tested on BCC and it get interrupted after some minutes) WoWScrnShot_120821_214045

I checked again. Works perfectly.

PR #9649, opened 2021-12-11
fix(Core/Instances): Always put boss minions in combat with zone

Nyeriah:
## Changes Proposed:
-  Always put minions in combat with all players, even if they already have a target.
-  Previously this call was void if they were already attacking something, causing them to not be put in combat with every player, eventually leading them to reset if their victim died before another player engaged them.

## Issues Addressed:

- Closes 

## Tests Performed:

- Not tested.
- 

## How to Test the Changes:

1. Engage Gehennas with two characters, attack the boss with both characters but only hit the add with the tank
2. Kill the tank that pulled
3. Notice if the adds stay in combat or reset without attacking the other player

- [ ]
- [ ]


Nefertumm (approved):
Nice!

PR #9651, opened 2021-12-11
chore(Core/Cache): Remove an assert breaking 1v1 arena module

Nyeriah:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


acidmanifesto (approved):

Kitzunu:
Add log_error then.
The assert was put there for a reason

acidmanifesto:
> Add log_error then. The assert was put there for a reason

 I am curious about the logical inconsistency being detected with it being a assert. i wonder if this assert was suppose to be a log to begin with from whatever source it was refactor from.

acidmanifesto:
> Add log_error then. The assert was put there for a reason

This is to correct the issue of the 1v1 module breaking. If it was with out the 1v1 module then this will make sense to have that assert. But since the 1v1 module is a catelog module, this assert really has no purpose.

Kitzunu:
Yea. But if you dont use it this might be valuable information

Nyeriah:
Not really, I removed the asserts from everywhere else, I had just forgotten this one. It makes sense for Trinity to have this because they don't have wild settings like we do, but since mods have freedom to do anything, it's pointless to log this while things are operating as expected.

Winfidonarleyan (approved):

PR #9674, opened 2021-12-12
fix(Core/ScriptMgr): Correct whisper hook parameters

Rochet2:
## Changes Proposed:
-  Correct the language and type of the whisper chat hook
- The other chat hooks behave the same way as the corrected one
- Seems that the old code was a copy paste mistake from emote code

## Issues Addressed:

- Closes https://github.com/azerothcore/mod-eluna-lua-engine/issues/74

## Tests Performed:

- Builds
- Works

## How to Test the Changes:

1. You could try make a playerscript that uses whisper hook and then you could check the parameter values that are being passed to the hook. The values should be CHAT_MSG_EMOTE for type and LANG_UNIVERSAL for language which is wrong.
2. Alternatively set up Eluna and AIO and try out the test window script, wich should show up as you log in or as you use .test command https://github.com/Rochet2/AIO/tree/master/Examples/TestWindow The script fails because eluna expects the language to be addon and its hardcoded to be universal.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -8844,7 +8844,7 @@ void Player::Whisper(std::string_view text, Language language, Player* target, b
 
     std::string _text(text);
 
-    if (!sScriptMgr->CanPlayerUseChat(this, CHAT_MSG_EMOTE, LANG_UNIVERSAL, _text, target))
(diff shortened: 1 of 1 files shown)


Kitzunu:
Please, next time create branches on your fork. Making them on the repo is polluting the discord hook.

Nyeriah (approved):

Winfidonarleyan:
My bad, thanks for PR.

PR #9675, opened 2021-12-12
refactor(Core/DB): Deprecate MariaDB 10.4 and below

Kitzunu:
## Changes Proposed:
- Deprecate older versions of MariaDB as they cannot run queries that MySQL 5.7 and 8.0 can run.


src/server/database/Database/DatabaseWorkerPool.cpp
```diff
@@ -40,8 +40,8 @@
 #endif
 
 #if MARIADB_VERSION_ID >= 100600
-#define MIN_MYSQL_SERVER_VERSION 100200u
-#define MIN_MYSQL_CLIENT_VERSION 30203u
+#define MIN_MYSQL_SERVER_VERSION 100500u
+#define MIN_MYSQL_CLIENT_VERSION 30503u
```
Winfidonarleyan, line 44: I'm not sure about this line.

Winfidonarleyan (approved):

src/server/database/Database/DatabaseWorkerPool.cpp
Kitzunu: me neither, but I could not find where it is located

Winfidonarleyan: in `MariaDB 10.6` only

Winfidonarleyan:
And need wiki changes

Kitzunu:
> And need wiki changes

There is no info bout it on the wiki, but I will fix it

src/server/database/Database/DatabaseWorkerPool.cpp
Kitzunu: then I will just change back client thing? Or do you have time to test it fast with your setup?

I only run mysql

Issue #2211, opened 2019-08-14
Gameobject pooling for herbs and mines not working correctly


Malow:
##### SMALL DESCRIPTION:
The gameobject pooling for herbs and mines is supposed to only spawn some nodes of the available ones within a zone if I understand it correctly. Currently it seems to be spawning all, or at least way too many, of them at the same time.

##### EXPECTED BLIZZLIKE BEHAVIOUR:
There should be fewer herb and mining nodes in the world, and especially mining nodes where either a rare-version or a common-version can spawn (Truesilver/Mithril, Iron/Gold, Tin/Silver) should only have 1 of them spawned, and not both at the same time spawned inside of each-other.

##### CURRENT BEHAVIOUR:
![Image1](https://i.imgur.com/MTjTT9E.png)
![Image2](https://i.imgur.com/5lJVY2z.jpg)
![Image3](https://i.imgur.com/u9KAChR.jpg)
![Image4](https://i.imgur.com/2MBNs8j.jpg)
![Image5](https://i.imgur.com/GuLvtZz.png)

##### EXTRA NOTES:
I've only tested the Vanilla-world, but the problem exists in all zones I've tested (10+ I think). Black Lotuses in Burning Steppes is a good example, Black Lotuses should be pretty rare, but doing a full sweep in Burning Steppes you can easily find 10+ Black Lotus nodes.

##### BRANCH(ES):
master

##### AC HASH/COMMIT:
178315f8e263394063a602cda17d6eca1bd294da

##### OPERATING SYSTEM:
Win 10

##### MODULES:
None

##### OTHER CUSTOMIZATIONS:
Completely fresh non-modified server.

andoys:
Hmm in my live server,  players also reported same thing for Classic World. But it's a different story for Northrend there's too few herbs and mining nodes. 
Might be related. Which for me fixed some of it here: https://github.com/azerothcore/azerothcore-wotlk/issues/1857#issuecomment-496941534

BarbzYHOOL:
Do we have to fix the DB one by one or is it core related? I don't know but that's kinda annoying

Si1ker:
Could this be related? #2510

Zoidwaffle:
[...]

PR #9676, opened 2021-12-12
feat(core/pool): Adjust GameObject pools in world via config

dantav:
## Changes Proposed:
Based on issues reported with mining and herb nodes being too close or too many. I have created a way of targeting specific pools or many pools with directives in `worldserver.conf` 

The issues reported are apparently database related have mostly the issue of not being associated to a pool. As the will spawn  and respawn without minding the `max_limit` of the pool of that area.

So this feature will serve two purposes:

1. A means of production game servers to tweak their pools
i.e the following config: `Pool.GameObject = "Tin,0.2,50002;"` Adjust all tin in Jangolode Mine {50002} of westfall to 20% // Originally 15 now 3 mines max appear
2. Can be used to debug and look for issues in certain areas
i.e. `Pool.GameObject = "*,0.0,50002;" `Disables all GameObjects in Pool to see what is not in that pool, or `Pool.GameObject = "*,0.0;" `to see all GameObjects in world thats not in a pool, also added additional logging for interacting with a gameobjects, `Logger.spells.objects`, it will display the Object template id, guid of object and type. You can then confirm its existence in `pool_gameobject`  table

## Issues Addressed:

- Closes (kinda, its a work around until database is cleaned up)
https://github.com/azerothcore/azerothcore-wotlk/issues/2211

## Tests Performed:

Tested multiple config options in game, heres some I documented
```
Pool.GameObject = "*,1.5" ==> Adjust all GameObjects pools to 150% more, needs comma to signify no description
Pool.GameObject = "Tin,0.2,50002;" ==> adjust all tin in Jangolode Mine of westfall to 20% // Orginally 18 now 3 mines
Pool.GameObject = "Copper,0.1;Tin,0.01,920;Silver,0.0;Gold,0.2;Iron,0.2;Mithril,0.2;Truesilver,0.2;Thorium,0.2;Dark Iron,0.2;Colbalt,0.2;"
```

It still needs more testing

## How to Test the Changes:

1. Read the new config setting at bottom of `src/server/worldserver/worldserver.conf.dist`
2. Build your own config in your config
3. Restart game server
4. Get a herb or mining GM (if you want data on nodes you will need to set your characters_skill to appropriate level, for instance character of GUID of 10 find, value `UPDATE characters_skill SET value = 150 WHERE guid = 10 AND skill = 186` this sets your characters mining skill to 150. I do not know of a GM command for this.)
5. Look at he nodes around the area and notice they change before and after the config changes. Any node not disappearing or respecting your directive you can verify if its missing from the `pool_gameobjects` table. Its expected to be either missing or wrong data in pool tables.
6. You can also fix gameobjects by associating them to a pool near by and then watch them be affected by your directives.

I notice chests are not in pools, it would be nice if someone can find one with a pool to test
Need more herb testing, I mostly tested mines in Westall and Elwynn Forest.

### DIFF
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -749,7 +749,7 @@ void PoolMgr::LoadFromDB()
                 {
                     std::vector<std::string> directives = explode(sWorld->getGameObjectConfig(), ';');
                     std::string configDescription;
-                    
+
                     uint32 target_pool_id;
 
                     bool found = false;
@@ -779,7 +779,7 @@ void PoolMgr::LoadFromDB()
                         // check if correct pool, target_pool_id=0 means no pool specified, its global
                         if (target_pool_id != 0 && target_pool_id != pool_id)
                             continue;
-                        
+
                         std::string description = fields[3].GetString();
 
                         // check if correct description or empty (which means global, any description matches)
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -686,8 +686,7 @@ void PoolMgr::LoadFromDB()
     {
         uint32 oldMSTime = getMSTime();
 
-        //                                                 1        2         3
-        QueryResult result = WorldDatabase.Query("SELECT guid, pool_entry, chance FROM pool_gameobject");
+        QueryResult result = WorldDatabase.Query("SELECT guid, pool_entry, chance, description FROM pool_gameobject");
 
         if (!result)
         {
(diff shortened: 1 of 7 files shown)


Kitzunu:
I don't see the point of this. I mean great job of figuring this out, but it is things you should be able to do from the database.

So if there is an issue with the max_limit in the database we should fix that and not work around it like this.

dantav:
> I don't see the point of this. I mean great job of figuring this out, but it is things you should be able to do from the database.
> 
> So if there is an issue with the max_limit in the database we should fix that and not work around it like this.

Yeah this should be a database fix, but honestly this should make it easier for troubleshooting. Also allows game servers to be tweaked if they want a more challenging or easier way of finding resources.

Kitzunu:
I dont like this approach for several reasons

1. this should be fixed so the DB actually works
2. Finding unpooled gobs can be done with SQL queries
3. relying on comments is bad as
    - they can be misspelled
    - 975 gob pools don't even have a comment
4. if the DB is fixed, then the modifying of how many gobs in a pool can simply be altered by the user in the database.

dantav:
> I dont like this approach for several reasons
> 
> 1. this should be fixed so the DB actually works
> 2. Finding unpooled gobs can be done with SQL queries
> 3. relying on comments is bad as
>    
>    * they can be misspelled
>    * 975 gob pools don't even have a comment
> 4. if the DB is fixed, then the modifying of how many gobs in a pool can simply be altered by the user in the database.

It gives power to the user to tweak things via config. Anyways if its so easy to fix why has it sat unresolved for 2 years? I'm just providing a tool. The description check is just an option, you can bypass it with * to match anything.

Kitzunu:
it has been broken since 2015 on TC and as this is just a fork of TC, let's say it's been broken for 6 years :D

Never said it was easy, I have never looked at it. Never been interested. I'm just saying that if the more we work around things in the core, the less likely the actual issue will be fixed.

dantav:
Actually this would better serve as a mod.

Kitzunu:
that I can support :D

Kitzunu:
Please ping any of us in either discord or this pr if you decide to make a module and we can fork it under the AC org

PR #9760, opened 2021-12-18
fix(db): 4 NPCS in sw shadowglen with "wait" movetype location fixed

neifion-00000000:
2 Grell and 2 Grellkin were incorrectly placed.

Uploaded sniff that was taken while working/exploring to submissions.

## Changes Proposed:
-  Place mobs in the correct location-  

## SOURCE:
Packets for area 04 have been submitted prior to this PR
image

## Tests Performed:
- Mobs checked against source.
- SQL tested 

## How to Test the Changes:

1.  .go c 47286 (47284 is nearby)
     .go c 47334 (47344 is nearby)
2.  Compare to classic servers  (read below if they do not match exactly)

Some interesting behavior was discovered, a "scare" mechanic.  It can make the mobs move to new locations as they run away from any active combat.  Both Grell and Grellkin have this mechanic.  If you go to check these mobs against classic and do not see them in the correct location, be sure to clear the camp (or at least kill the ones with the behavior we emulate as "stay" movement), its likely they had the scare mechanic go off and were never killed afterward.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1639812684867509600.sql
+++ b/data/sql/updates/pending_db_world/rev_1639812684867509600.sql
@@ -1,9 +1,7 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1639812684867509600');
 
 -- Source: TBC Classic 2.5.2.41446
-
 DELETE FROM creature WHERE guid IN (47284, 47286, 47334, 47344);
-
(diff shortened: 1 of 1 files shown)


Kitzunu:
Dont replace into.
Use delete + insert

neifion-00000000:
> I agree. We need to do a delete and then insert or a update. With replace we may end up with undesire outcomes.

fixed it up :)

Kitzunu:
You can make these into one query.

Check https://www.azerothcore.org/wiki/sql-standards to learn more about SQL stuff

neifion-00000000:
I am including this in a different upcoming PR as part of a larger spawn point revision

PR #9771, opened 2021-12-18
feat(Apps/CI): Add check for broadcast_text

Kitzunu:
## Changes Proposed:
-  why not be too sure that people don't fuck with the table unless they know what they are doing :D


apps/ci/ci-pending.sh
```diff
@@ -31,5 +31,16 @@ for i in `find data/sql/updates/pending* -name "*.sql" -type f`; do
     fi
 done
 
+for i in `find data/sql/updates/pending* -name "*sql" -type f`; do
+    if $(cat "$i"|sed "s/'.*'\(.*\)/\1/g"|grep -q -i -E "broadcast_text"); then
+        echo "> broadcast_text check - OK"
+    else
```
Kitzunu, line 37: could someone just proof check that I have the if/else in the right order and that the regex is correct? :D

Winfidonarleyan: just add test file

Kitzunu: True. Didnt think of that hahaha

Kitzunu:
ok works

Winfidonarleyan (approved):

Nyeriah (approved):

PR #9874, opened 2021-12-24
fix(Script) Quest "Deactivating the Spire"

malcrom:
## Changes Proposed:
- Update quest to deactivate the beams for 10 seconds on click like official.  
- I chose C++ over SAi for this scripting due to DespawnOrUnsummon forced respawn timer fails the first time in SAI and I wanted this to work correctly.

## SOURCE:
Sniff

## Tests Performed:
- Tested in game

## How to Test the Changes:
1. .quest add 8889
2. .go xyz 9334.2 -7880.1899 74.9103
3. do quest

### DIFF
--- a/src/server/scripts/World/go_scripts.cpp
+++ b/src/server/scripts/World/go_scripts.cpp
@@ -1942,7 +1942,7 @@ class go_duskwither_spire_power_source : public GameObjectScript
 public:
     go_duskwither_spire_power_source() : GameObjectScript("go_duskwither_spire_power_source") {}
 
-    bool OnGossipHello(Player* /*player*/ player, GameObject* go) override
+    bool OnGossipHello(Player* /*player*/, GameObject* go) override
     {
         if (Creature* bunny = go->FindNearestCreature(NPC_POWER_SOURCE_INVISIBLE_BUNNY, 1.0f))
         {
--- a/src/server/scripts/World/go_scripts.cpp
+++ b/src/server/scripts/World/go_scripts.cpp
@@ -1928,6 +1928,30 @@ public:
     }
 };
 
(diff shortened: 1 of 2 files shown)


src/server/scripts/World/go_scripts.cpp
```diff
@@ -1928,6 +1928,30 @@ class go_bells : public GameObjectScript
...
+{
+    NPC_POWER_SOURCE_INVISIBLE_BUNNY = 17984
+};
+
+class go_duskwither_spire_power_source : public GameObjectScript
+{
+public:
+    go_duskwither_spire_power_source() : GameObjectScript("go_duskwither_spire_power_source") {}
+
+    bool OnGossipHello(Player* player, GameObject* go) override
```
acidmanifesto, line 1945:
```
bool OnGossipHello(Player* /*player*/, GameObject* go) override
```

acidmanifesto (changes requested):

src/server/scripts/World/go_scripts.cpp
acidmanifesto: commit out the **player** and that should fix the compile testing.

malcrom: Thanks, fixed

acidmanifesto: You forgot to delete player so it shows as Player* /*player*/ player.
Needs to show as Player* /*player*/

acidmanifesto:
```
You forgot to delete player so it shows as Player* /*player*/ player.
Needs to show as Player* /*player*/
```

acidmanifesto (approved):

PR #9928, opened 2021-12-30
fix(SQL): Remove wrong comment

Kitzunu:


malcrom (approved):

Nyeriah:
I think this doesn't work like this unfortunatelly, we need to check if the following files applied successfully even though this one failed, and if they did, it' probably better we don't touch this since then it'll reapply to this revision. Not entirely sure though

Kitzunu:
I guess what we could do is nuke the file

Kitzunu:
💥

Nyeriah:
why was this merged? did we test if it doesn't break the database revision?

Issue #9745, opened 2021-12-17
NPCs from time to time doesn't show vendor window


Shineslippers:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/2604

When talking to a vendor, sometimes their menu doesn't appear, but the game still thinks you are "shopping" with them, and you need to walk away from them until the menu would normally close, and then initialize a conversation with them again. sometimes happens more than once, seems to only happen when initializing conversation when moving

Tester: 

The issue is confirmed. Sometimes when you running to NPCs and right-click on them it doesn't show vendor's window. NPCs is sticking to player like when vendor's window is open. `Esc` or `Rgihr-click again` doesn't help. Only out of range can reset it. It's hard to reproduce. Didn't find pattern.

https://user-images.githubusercontent.com/85256266/146576715-1f34cb9f-99e9-4fec-b5d0-9184891df209.mp4

### Expected Blizzlike Behaviour

Talking to a vendor should pull up their menu all the time without failiure.

### Source

X

### Steps to reproduce the problem

.go c 37605
run and right-click on vendor

### Extra Notes

Does not happen with just Kaita Deepforge, thats just the NPC i tested it with and noticed it first.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/d87d0a2ae892286a44a85cc230fa66d44479d4bd

### Operating system

Windows 10

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[...]

PR #9889, opened 2021-12-26
Fix (Core) issue in Creature::UpdateEntry

acidmanifesto:
This was a devil to debug. This should now load the scriptid if any and then force clear any stuck menu status before preparing and sending a gossip to a player.

## Changes Proposed:
-  We must update last scriptId or it looks like we reloaded a script, breaking some things such as gossip temporarily.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/2604
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9745

## Tests Performed:

- Builds and Tested in Game.

## How to Test the Changes:

1. .go c 37605
2. run and right-click on vendor

image

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #9962, opened 2022-01-01
fix(Core): Suppress status opcode after gossip menu

dantav:
This issue seems to just affect only NPCs who are quest givers who are also vendors/trainers. 

## Changes Proposed:
-  Suppress opcode query for gossip of quests

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/2604
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9745

## SOURCE:

This is debug while walking to NPC and right clicking
gossip

## Tests Performed:

- Built and tested on my local environment, no issues

## How to Test the Changes:

1. Checkout the master
2. Goto any vendor/trainer that is also a quest giver, (does not need to have any active quests)
3. Walk towards the npc while spam right clicking (don't need to go crazy here, just typical game playing)
4. You will now notice the menu appears from the NPC, however clicking on any menu item does nothing

Repeat test with standing still
You will notice it works fine.

Repeat walking test with this PR, you will see issue is gone

- More testing, interacting with other NPCs


acidmanifesto (approved):
Let see if we can get some consistant  testering

acidmanifesto:
@IntelligentQuantum can you test this.
May be a better solution then my pr attempt where i ended up i na dead end https://github.com/azerothcore/azerothcore-wotlk/pull/9889

dantav:
Yeah this one was trick

> @IntelligentQuantum can you test this. May be a better solution then my pr attempt where i ended up in a dead end #9889

Yeah this one was super tricky because of the packet timing

IntelligentQuantum (approved):

IntelligentQuantum:
@acidmanifesto works fine

acidmanifesto:
Thank you for the pr.

Issue #8632, opened 2021-10-20
Hallows end, replace hardcoded strings.


arnaucastellvi:
### Current Behaviour

The event of Hallows end (`hallows_end.cpp`) have hardcoded strings, would be good to replace it for text of broadcast_text and broadcast_text_locale.

### Expected Blizzlike Behaviour

You can modify and locate the text via DDBB.

### Source

Example of hardcoded text in src/server/scripts/Events/hallows_end.cpp

`me->MonsterYell("Prepare yourselves, the bells have tolled! Shelter your weak, your young and your old! Each of you shall pay the final sum! Cry for mercy; the reckoning has come!", LANG_UNIVERSAL, 0);`

` player->MonsterSay("Horseman rise...", LANG_UNIVERSAL, 0);`

### Steps to reproduce the problem

1- Modify text in acore_world.broadcast_text and acore_world.broadcast_text_locale where id =22695 (example text)
2- Restart worldserver and start event..
3- The modification is not appear in game.

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/ba719f0feb70827beaebb40ff5095f75e2fd4302

### Operating system

Debian 10

Kitzunu:
Fyi for the one taking this on. There are a lot of spells that needs to be scripted for this. Look at TC's file for reference =)

v-mstrs:
Completed as of https://github.com/azerothcore/azerothcore-wotlk/commit/6e39ffa64e323f8b2cf3bb24b34c0ae41a2b701d

Nyeriah:
Thanks!

PR #9976, opened 2022-01-02
fix(CORE/Events): Hallows end,  hardcoded strings replaced.

febaut:
### Changes Proposed

The event of Hallows end (hallows_end.cpp) had hardcoded strings. Were replaced according to broadcast_text.

### Issues Addressed:
Closes:  https://github.com/azerothcore/azerothcore-wotlk/issues/8632

### DIFF
--- a/src/server/scripts/Events/hallows_end.cpp
+++ b/src/server/scripts/Events/hallows_end.cpp
@@ -960,6 +960,16 @@ enum hhSounds
     SOUND_DEATH                                     = 11964,
 };
 
+enum hhTexts // broadcast_text.ID
+{
+    HH_SAY_DEATH                                    = 23455,
+    HH_SAY_PLAYER_DEATH                             = 40546,
+    HH_SAY_REJOINED                                 = 22271,
+    HH_SAY_ENTRANCE                                 = 22261,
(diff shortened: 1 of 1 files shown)


Kitzunu:
Add the entries in creature_text and reference to that instead

febaut:
Is that necessary? I saw some examples and they call from broadcast_text. Also, if I'm not wrong, broadcast table has priority.

Kitzunu:
Yea, you reference broadcast_text in creature_text. Better doing it that way so that everything is done the same.

febaut:
I've double checked and creature_text does have the texts for Headless Horseman already.

`SELECT * FROM creature_text WHERE CreatureID = 23682`

On the other hand, Say and Yell, they both refer to Talk function which uses broadcast text so no way to refer directly creature_text.

acidmanifesto (approved):

temperrr:
I can't get the headless horseman to spawn in rdf with .event start 12 and joining the headless horseman queue.... :/ (with this pr or without.)

Nyeriah:
The changes requested by Kitzunu were not addressed. Waiting feedback from the author @febaut

Issue #9860, opened 2021-12-23
[Quest] Poisoned Water - Transformed elementals do not attack


ishidinho:
### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/2665

During the quest "Poisoned Water" in Eastern Plaguelands, you use the item [Aspect of Neptulon] to transform elemental in Discordant Surges. After the transformation the elementals do not attack you.

![WoWScrnShot_122321_161336](https://user-images.githubusercontent.com/62797536/147259657-deb8ac35-fd3e-44cc-aea6-f11fa321f165.jpg)

Tester:

The issues with the quest is confirmed.

![WoWScrnShot_122321_163845](https://user-images.githubusercontent.com/62797536/147263071-0bdaf28c-c399-4e04-8cf4-19597bde1221.jpg)

### Expected Blizzlike Behaviour

After the transformation the elementals should attack you. 

### Source

Video with timestamp: https://youtu.be/mu9t2sjrh2A?t=92

### Steps to reproduce the problem

1. `.go c 92084`
2. `.quest add 6804`
3. Use [Aspect of Neptulon] on elementals

### Extra Notes

Quest: https://wowgaming.altervista.org/aowow/?quest=6804

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/c70cc855ff3d2724a0e0a3866b6a43db1f2faddb

### Operating system

Ubuntu

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #9988, opened 2022-01-03
fix(DB/Quest/Creature): Poisoned Water: Transformed elementals.

febaut:
During the quest "Poisoned Water" in Eastern Plaguelands, you use the item [Aspect of Neptulon] to transform elemental in Discordant Surges. After the transformation the elementals didn't attack you. Now they do as it supposed.

## Changes Proposed:
-  Add a row to related creatures' SMART_SCRIPTS that pulls the creature on spellhit. 

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9860

## Tests Performed:

- 
- 

## How to Test the Changes:

First apply the SQL scripts. Then restart the server (.reload smart might not apply the changes to spawned creatures).
1. .go c 92084
2. .quest add 6804
3. Use [Aspect of Neptulon] on elementals

- [ ]
- [ ]


data/sql/updates/pending_db_world/rev_1641244560298659050.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1641244560298659050');
+
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES (8522, 0, 2, 0, 8, 0, 100, 0, 21332, 0, 0, 0, 0, 49, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 0, 'Plague Monstrosity - On Spellhit \'Aspect of Neptulon\' - Start Attacking');
```
acidmanifesto, line 3:
```
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES 
(8522, 0, 2, 0, 8, 0, 100, 0, 21332, 0, 0, 0, 0, 49, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 0, 'Plague Monstrosity - On Spellhit \'Aspect of Neptulon\' - Start Attacking'),
(8521, 0, 3, 0, 8, 0, 100, 0, 21332, 0, 0, 0, 0, 49, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 0, 'Blighted Horror - On Spellhit \'Aspect of Neptulon\' - Start Attacking');
```

acidmanifesto: Condense two seperate queries into one query.
Line Break after VALUE

acidmanifesto (changes requested):

Nyeriah:
Hey, 

First of all, thank you very much for your contribution. However, this looks wrong to me as creatures shouldn't need an event call to start attacking players, this likely has to do with them having the wrong flags or wrong react states.

I'm not sure if we'll be able to get some sniffing on this, but that's most definitely the case for them.

Regards

febaut:
> 
In this case the spell that does the creature template update has a maximum range of 30 yards which is more than threat range. If you are close enough to the mob it will attack, as a normal mob does.

Kitzunu:
no

Nyeriah:
Well, what I understood of this issue is that the transformed elementals never engaged players, but looking into that report the second image has a level 81 character so that would explain why it doesn't try to engage him.

>> In this case the spell that does the creature template update has a maximum range of 30 yards which is more than threat range. If you are close enough to the mob it will attack, as a normal mob does, then is not flags or something else.

This is likely wrong then, because if the creature should be put in combat with the player after being hit by it then the spell itself would handle that, but it seems this spell has no initial threat by design so that never happens.

febaut:
Yep, I tried with lvl 56. I'm curious about the spell because it does basically nothing. The transformation is through Smart Script as well.

acidmanifesto:
Added Need Sniff label for further clarification in how it should be handled.

PR #10045, opened 2022-01-06
refactor(Core/Creature): Reworked creature aggro radius calculation

Kitzunu:
* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/980047763a685217e712f58b8713d48523d90ccb)

* Formulas are taken from WoW Wiki

Co-Authored-By: Ovah <18347559+Ovahlord@users.noreply.github.com>

## Changes Proposed:
-  Rework creature aggro radius calculations with formulas from WoW Wiki

## Tests Performed:

- builds

## How to Test the Changes:

1. Check general aggro radius of creatures above and below your level

for example
> The aggro radius for creatures with equal level as the player is 20 yards.

> WoW Wiki: the minimum radius seems to be 5 yards, while the maximum range is 45 yards

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -3328,43 +3328,52 @@ void Creature::ReleaseFocus(Spell const* focusSpell)
 float Creature::GetAttackDistance(Unit const* player) const
 {
     float aggroRate = sWorld->getRate(RATE_CREATURE_AGGRO);
-
     if (aggroRate == 0)
         return 0.0f;
 
     if (!player)
         return 0.0f;
 
-    uint32 playerLevel = player->getLevelForTarget(this);
-    uint32 creatureLevel = getLevelForTarget(player);
+    // WoW Wiki: the minimum radius seems to be 5 yards, while the maximum range is 45 yards
+    float maxRadius = (45.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
+    float minRadius = (5.0f * sWorld->getRate(RATE_CREATURE_AGGRO));
 
-    int32 levelDiff = static_cast<int32>(playerLevel) - static_cast<int32>(creatureLevel);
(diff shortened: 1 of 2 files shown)


acidmanifesto (approved):

T1ti:
Does this override the new creature detection_range parameter we recently added in creature_template ?

Kitzunu:
This fixes the general aggro radius calculations to be as blizzlike as we know.

I don't know where detection_range is used, but none of the removed code has any correlation with it

T1ti:
Looks like another issue then if it's not used

Kitzunu:
Do you know where detection_range is used? Because it wasn't in the previous code either for Creature::GetAttackDistance()

Kitzunu:
Ah it is used in Creature::GetAggroRange()

Kitzunu:
what is the use-case of detection_range

Kitzunu:
I just found GetAggroRange and GetAttackDistance used very weirdly in AC. I need to investigate it. Because right now they are both a hot mess.

But what I want to know to make my life a tad bit easier is.

is detection_range a random guessed value, or is it a sniffed value?
Because in DB atm we have 1, 10, 18, 20.

Kitzunu:
Ok so I have confirmed it is not sniffed, and on wowwiki it states 20yd.

So right know detection_range just looks like something made up to me with values someone took from the air

T1ti:
> Ok so I have confirmed it is not sniffed, and on wowwiki it states 20yd.
> 
> So right know detection_range just looks like something made up to me with values someone took from the air

https://github.com/azerothcore/azerothcore-wotlk/issues/5455

wow wiki is wrong, the range isn't **_always_**  20, exceptions are rare but they exist

Kitzunu:
Okay so basically a guessed value.

But now I know what I can work with.

Kitzunu:
I really can't be bothered to look into this

Issue #10099, opened 2022-01-09
Feature: allow to queue both RDF & BGs together


FrancescoBorzi:
### Describe your feature request or suggestion in detail

By default one can't join both BG and RDF at the same moment. It would be nice to have a conf to allow it. Of course this conf should be disabled by default.

PR #10102, opened 2022-01-09
feat(Core/Player): allow to queue both RDF & BGs together

IntelligentQuantum:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10099

## Tests Performed:

- Not Tested
- 

## How to Test the Changes:

1. https://github.com/azerothcore/azerothcore-wotlk/issues/10099
2.
3.

- [ ] My english is shit So please fix description in worldserver.conf.dist ^^
- [ ]


acidmanifesto (approved):
Description is fine with me.

Nyeriah:
What happens if they accept the BG call during the LFG role check proposal?
Also what happens if they are inside a BG and get a LFG call, vice versa?
Might want to remove them from the other queue once they accept one of the queues?

src/server/worldserver/worldserver.conf.dist
```diff
@@ -3622,6 +3622,14 @@ LFG.KickPreventionTimer = 900
 
 EnablePlayerSettings = 0
 
+#
+#     Allow.Join.BG.And.LFG
+#        Description: Allow join BG and LFG in the same time.
```
Nyeriah, line 3627:
```suggestion
#        Description: Allow queueing for BG and LFG at the same time.
```

FrancescoBorzi:
@IntelligentQuantum thanks for this PR in record time! :) did you handle cases like:

- the player is already in BG and asked to join RDF
- the player is already in RDF and asked to join BG

or does the other queue cancel as soon as you join one of them?

acidmanifesto:
> What happens if they accept the BG call during the LFG role check proposal?
> Also what happens if they are inside a BG and get a LFG call, vice versa?
> Might want to remove them from the other queue once they accept one of the queues?

I know and can only speak for sl retail in this case. On sl retail you will remain in queue for the other while in the other and if u accept u will be in the group/bg with a deserter aura. Blizzlike at its finest

Nyeriah:
> > What happens if they accept the BG call during the LFG role check proposal?
> > Also what happens if they are inside a BG and get a LFG call, vice versa?
> > Might want to remove them from the other queue once they accept one of the queues?
> 
> I know and can only speak for sl retail in this case. On sl retail you will remain in queue for the other while in the other and if u accept u will be in the group/bg with a deserter aura. Blizzlike at its finest

Yes, that's not exactly what I meant though. It was more in the sense of, does the queue still work if that happens. The RDF queue is prone to jamming from what I can tell, so we must ensure it doesn't break once they queue to multiple things.

But I think it should be fine

temperrr:
It does exactly what we described.

You can queue for both, whichever pops up first you get queued into, then removes the other queues.

Do we want to extend the description to let players stay in queue as on hold whilst they do a battleground and vice versa?

src/server/worldserver/worldserver.conf.dist
```diff
@@ -3622,6 +3622,14 @@ LFG.KickPreventionTimer = 900
 
 EnablePlayerSettings = 0
 
+#
+#     Allow.Join.BG.And.LFG
+#        Description: Allow queueing for BG and LFG at the same time.
+#        Default:     0 - Disabled
+#                     1 - Enabled
+
+Allow.Join.BG.And.LFG = 0
```
Winfidonarleyan, line 3631: `JoinBGAndLFG.Enable = 0`

Nyeriah: @IntelligentQuantum could you check this one please?

Nyeriah:
There's unexpected logic here

Skärmbild 2022-01-25 094658

If the setting is **disabled**, it will fall into that IF and leave the conditional segment. In other words, if the setting is disabled, no other RDF requirement is ever checked, leading level requirements to be bypassed among other things.

This breaks the RDF queue. Could you check this @IntelligentQuantum ?

PR #10170, opened 2022-01-14
fix(DB): Teldrassil herbalism gameobject re-construction

neifion-00000000:
All objects are pulled from sniff data.
Behavior observed while gathering data used to form pools.

## Changes Proposed:
Adds a significant amount of missing herbalism gameobjects for Teldrassil.
Introduces area based pooling to prevent herb extinction.
image
image

Overall herbs increased based on count (fully spawned zone)

## Issues Addressed:

- Removes old incorrect pooling.

## SOURCE:

Teldrassil zone packets, tediously gathered

## Tests Performed:

- Every single node inserted has been checked as part of the process I did to pool them

## How to Test the Changes:

1.  Load the SQL and pick a pool you'd like to check.
2.  Check the spawns in that pool and make sure only the number assigned are active.

Note:  It looks a bit heavily spawned due to view distance defaults on AC; turn down view distance in config and it will look more natural, the actual spawns are fine and working properly

There are likely less than 5, possibly less than 3 missing nodes (possibly none, but I don't give myself that much credit and I feel like I was missing some when I went to check, despite having gathered the packets) in the entire zone now, and they spawn as much as they should on a blizzard server.  There may be minor inaccuracies in the pooling but it is closer than it has ever been to correct.


neifion-00000000:
and of course, an Easter Egg map from when I was gathering packets 
image

acidmanifesto (approved):

PR #10176, opened 2022-01-14
Fix(Core/BWL): Vaelastrasz reset after wipe

### LINKED ISSUE #10153: [BWL] Vaelastrasz the Corrupt doesn't reset properly
CURRENT BEHAVIOUR:

When main aggro target dies, boss becomes stuck in evade mode
image

EXPECTED BLIZZLIKE BEHAVIOUR:

Boss should not be in evade mode.

SOURCE:

Go to bwl second  boss, with two players
wipe on tank, boss just casually stands there.

NOTES

AC HASH/COMMIT:
0c35369
OPERATING SYSTEM:
Windows 11
MODULES:
mod-progression-system
CUSTOMIZATIONS:

Nefertumm:
## Changes Proposed:
-  Change react state when the intro is done, as it was passive all the time so it would never had changed the initial target.
-  We shouldn't talk to him every time we wipe.
-  Separate the timers.

## Issues Addressed:

- Closes #10153

## Tests Performed:

- Tested in-game

## How to Test the Changes:

1. .go c id 13020
2.  use the gossip, wait the intro and let the tank die
3.  it should reset or change the target.

- [x]  We shouldn't have to talk to him after every wipe. [[confirmed by [this](https://www.youtube.com/watch?v=ik6jreLykMQ) , already fixed]]


Nefertumm:
> The logic works as it should, I do however have one issue with this update.
> 
> I tried this by nuking the boss down to roughly 20 percent, using 2 characters to make sure there were no issues, and the boss reset as it should except the health doesn't reset until the next engagement. The health stayed at 20% and as soon as I attacked it again the health was reset. I waited several minutes after dying to make sure it wasn't being reset at any point before me engaging it again.
> 
> I wasn't able to trigger any issues with the reset except what I mentioned above. I don't know the retail behavior so I can't personally say if it should update when reset or not when engaged.

Strange... Check if the issue is present without the pr, please!

Nefertumm:
Ah yes, I see the issue, the health reset is in entercombat instead of reset... I'll make an update on that too soon

tkn963:
If that is indeed the issue, if I'm not required to test it again (which I can if asked) I can say it works as it should.

Nefertumm:
Check again whenever you can :)

tkn963:
Just sending this to make sure it's known I'm not ignoring it or missed it but I'll test it later tonight, sadly though probably not for another 8 hours (work).

tkn963:
Tested, works as it should! 👍

Nyeriah (approved):

acidmanifesto (approved):

Issue #10214, opened 2022-01-17
Knot Thimblejack's Cache loot - some items should not be dropped by mobs


nolt:
### Original issue: 
https://github.com/chromiecraft/chromiecraft/issues/2802

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [X] 60

### Current Behaviour

![image](https://user-images.githubusercontent.com/84640316/149617492-e881f160-a467-4bb4-96a9-eb4346b052c4.png)

### Expected Blizzlike Behaviour

these should only drop from the cache in DM. Not from a random lvl 80 mob.

### Source

https://tbc.wowhead.com/item=18519/pattern-shifting-cloak

### Steps to reproduce the problem

keira 3 or mysql.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/c1ebda66bae939e7151fc9f43176dd897e8c01b7

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
[...]

PR #10215, opened 2022-01-17
fix(DB/CreatureLoot) Felhunter (417)

nolt:
## Changes Proposed:
-  Deletes drop from Felhunter

## Issues Addressed:

- Partially Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10214

## SOURCE:

https://tbc.wowhead.com/item=18519/pattern-shifting-cloak

## Tests Performed:

-  sql applied, test drop

## How to Test the Changes:

1. Spawn Felhunter (entry 417), kill it, pattern will not drop

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1642432018580300313.sql
+++ b/data/sql/updates/pending_db_world/rev_1642432018580300313.sql
@@ -1,4 +1,4 @@
 INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1642432018580300313');
 
-#closes ##10214
-DELETE FROM `creature_loot_template` WHERE `Entry` = 417 and `Item` = 18519;
+#partially closes ##10214
+DELETE FROM `creature_loot_template` WHERE `Entry` = 417;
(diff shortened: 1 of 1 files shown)


nolt:
PR is not ready yet.

acidmanifesto (approved):
its funny, because entry 417 is the Warlock's summon Felhunter. Which would of made it impossible to drop at all unless looting from a player's pet is possible. I concur with the changes.

nolt:
@acidmanifesto 
if that so, I delete all loot from it cuz now it has 12 items that can be dropped :o

acidmanifesto:
> @acidmanifesto if that so, I delete all loot from it cuz now it has 12 items that can be dropped :o

even better. usually when it comes to loot audit, we will audit either just the loot item itself or the npc\gameobject\item itself.
in this case i suggest making it a npc loot audit so we can keep it less chaotic and simple.

acidmanifesto:
image
@nolt looks like the dry run caught something. can u correct it?

acidmanifesto:
I recommend adding the line to the sql
```UPDATE `creature_template` SET `lootid`='0' WHERE  `entry`=417;```

PR #6555, opened 2021-06-24
feat: reorder character selection screen

r-o-b-o-t-o:
## Changes Proposed:
-  Add an `order` column to the `characters` table, allowing players to reorder their characters in the character selection screen

## Issues Addressed:

None

## Tests Performed:

- Applied SQL and C++ patches
- Tested on Windows 10 client and server

## How to Test the Changes:

1. Create multiple characters in an account, eg:
image
2. Change their `order` column, for example:
```sql
UPDATE `characters` SET `order` = 1 WHERE `name` = "One";
UPDATE `characters` SET `order` = 2 WHERE `name` = "Two";
UPDATE `characters` SET `order` = 3 WHERE `name` = "Three";
UPDATE `characters` SET `order` = 4 WHERE `name` = "Four";
```
3. Log out and log back in
4. The characters should be reordered according to the `order` column:
image
5. (optional) If you want to reset back to the default order, use
```sql
UPDATE `characters` SET `order` = NULL WHERE `name` IN ("One", "Two", "Three", "Four");
```

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #10243, opened 2022-01-18
fix(Core/Database): use char guid if character order is null

r-o-b-o-t-o:
## Changes Proposed:
- Use the character's guid if the order value is null to sort the character selection list (see https://github.com/azerothcore/azerothcore-wotlk/pull/6555)

## Issues Addressed:

- This makes newly created characters go to the bottom of the characters list instead of the top

## Tests Performed:

- Built and tested on Windows 10 64bits, mariadb 10.5.10

## How to Test the Changes:

1. Log into an account that already had characters which use the `order` column
2. Make sure the characters are still in the same order as before
3. Create a new character, make sure the new character is at the bottom of the list


temperrr:
Tested:

image
(Eamath oldest char, I ordered them before applying the PR to be like this)
image

Works

acidmanifesto (approved):

PR #10297, opened 2022-01-22
fix (core/logging): log money

Bogir:
## Changes Proposed:
-  Correction of texts in the topic and the type of logged action. Getting rid of `<>` tags, which can cause certain difficulties when processing logs by the administrator. 
-  Now we can sort entries by the type of logged actions by `type` column if need

## Tests Performed:

- Tested in game by sending gold, trading between 2 characters, deposit and wifdraw gold from/to Guildbank and sending by mail.

## How to Test the Changes:

1. make transactions with money with amount about 100 gold betwen characters by trade or mail, deposit to and withdraw from  GuildBank or receive profit from AuctionHouse. Type of action must be placed in `type` field of `log_money` table. 
2. If you deposit money to GB - check that field `receiver_acc` get the Guild ID and `receiver_name` get Guild Name instead 0 as it was before.
3. If you wifdraw money from GB - check that sender is Guild and receiver is character instead 0 as it was before.
4. Now you can process logged actions with filtering by action type. If in PHP - you don't need to parce `topic` field and make `htmlspecialchars($topic)`-clearing to obtain recorded action type. Just filter or search by `topic` column and you'll find what you want.

For updating `log_money` table on live server you can use this function via php:

[fix_live_table_php.txt](https://github.com/azerothcore/azerothcore-wotlk/files/7918282/fix_live_table_php.txt)


Bogir:
I don't understand why old commit present here (ObjectMgr.cpp )

Viste:
о нихуя ты живой :DD я думал ты умер(в контексте эмуляции)

Bogir:
> о нихуя ты живой :DD я думал ты умер(в контексте эмуляции)
нет же это не ты кудыйта запропастилсо? я всё ботов допиливаю. ща мне надо чтоб на сервере это гумно было синхронно с моим, для того и прю

data/sql/updates/pending_db_characters/rev_1642821969869259834.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_characters` (`sql_rev`) VALUES ('1642821969869259834');
+
+-- column `type` contains type of logged action
+ALTER TABLE `log_money` ADD COLUMN `type` char(255) NOT NULL AFTER `date`;
```
Nyeriah, line 4: Hey, thanks for your PR, it's much appreciated. However, the design here could be improved as I do believe a tinyint field would be better than a char one. It serves the same purpose but there's less data stored overall.

0 - mail
1 - trade
...etc

Nyeriah (changes requested):

data/sql/updates/pending_db_characters/rev_1642821969869259834.sql
Bogir: You mean that I must to replace the words with numbers here and at the core?
type 1 = mail, type 2 =trade ... ?
Then we will have to describe it somewhere in the wiki...

Bogir: Fixed and checked in game all 6 types.

Nyeriah (approved):

data/sql/updates/pending_db_characters/rev_1642821969869259834.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_characters` (`sql_rev`) VALUES ('1642821969869259834');
+
+-- column `type` contains type of logged action
+ALTER TABLE `log_money` ADD COLUMN `type` tinyint(1) NOT NULL COMMENT '1=COD,2=AH,3=GB DEPOSIT,4=GB WITHDRAW,5=MAIL,6=TRADE' AFTER `date`;
```
Kitzunu, line 4:
```suggestion
ALTER TABLE `log_money` ADD COLUMN `type` TINYINT NOT NULL COMMENT '1=COD,2=AH,3=GB DEPOSIT,4=GB WITHDRAW,5=MAIL,6=TRADE' AFTER `date`;
```

Kitzunu (changes requested):

dualcorex:
I did an in-game test by sending some money between characters and depositing/withdrawing money from the guild bank while watching the updates to the `log_money` database. 

[x] Guild name shows up instead of "0" in `receiver_name` when depositing to guild bank and `sender_name` when withdrawing from the guild bank.
[x] A number is populate in the `type` column corresponding to the type of money transaction (i.e 6 for trades and 5 for mails).

NOTE: I did notice that amounts deposited/withdrawn from the guild bank did NOT register in the `log_money` database. Example: 10g00s00c or less was not recorded. 10g00s01c or more was recorded. Perhaps that is normal for the guild bank.

Capture

Kitzunu (approved):

Issue #10315, opened 2022-01-22
Command: Character Change Faction


htc16:
### Current Behaviour

When switching factions using the command:

`.character changefaction $name
`

unlocks virtually all flight points

![unknown](https://user-images.githubusercontent.com/58153179/150649843-0337bc25-ce9f-4aa0-a2d8-bd5e5f6f076d.png)

### Expected Blizzlike Behaviour

Above

### Steps to reproduce the problem

Above

### AC rev. hash/commit

AzerothCore rev. b5ab409614bf+ 2022-01-18 20:25:17

### Operating system

Debian 10 x64

Kitzunu:
confirmed on https://github.com/azerothcore/azerothcore-wotlk/commit/378539ce7f0fedf5c5a631ca724aef530c664f1a

acidmanifesto:
Fixed via https://github.com/azerothcore/azerothcore-wotlk/pull/9729

PR #9729, opened 2021-12-15
fix(Core/Handlers): Faction Change service properly reset and give ne…

UltraNix:
…w level-appropriate fly points.

Fixes #7981
Fixes #8732

## Issues Addressed:

- Closes #7981
- Closes #8732
- Closes https://github.com/chromiecraft/chromiecraft/issues/1788
- Closes https://github.com/chromiecraft/chromiecraft/issues/2192

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

change character level to something low, e.g. `char level 45`
`.character factionchange`
Change faction
See that only level-appropriate flight paths are available.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #10352, opened 2022-01-25
Disables the "all flight paths unlocked" after changefaction.

febaut:
## Changes Proposed:
-  CharacterHandler.cpp has a weird conditional structure that unlocks all the flight paths for a player after a faction change. It even does have a comment `//this doesn't seem to be 100% blizzlike... but it can't really be helped.`. However, I'm not sure about the Blizzlike behaviour. After some research, some private servers forums says that it unlocks all the paths wheter or not you had them before. There's no info as far as I saw about how it works blizzzlike. 

- This pull request disables the unlocking.  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10315

## SOURCE:

This is the only info I managed to find but is not totally useful 

https://us.battle.net/support/en/article/178364

## Tests Performed:

-  I checked Trinity Core's sourcer and it has something similar. The result is the same: unlocking all the paths.

## How to Test the Changes:

1. Comment the conditional structre as this PR suggests
2. Create a new char and do .level 30 (The condition requires lvl 7+ but there are some other things one may want to avoid, like money conditions)
3. .char changefaction 
4. Visit a Flight master

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -2193,7 +2193,7 @@ void WorldSession::HandleCharFactionOrRaceChangeCallback(std::shared_ptr<Charact
             stmt->setUInt32(0, lowGuid);
             trans->Append(stmt);
 
-            if (level > 7)
+            /*if (level > 7)
             {
                 // Update Taxi path
                 // this doesn't seem to be 100% blizzlike... but it can't really be helped.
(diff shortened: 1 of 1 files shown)


IntelligentQuantum:
Tested @Nyeriah review please

Nyeriah:
Hey,

Thanks for the PR, but I don't think this is something we want as default. Having them unlock every fly path is a problem but having without any fly path at all is even more of a problem.

febaut:
Yep, thought so. I think the current code aims to unlock some of them based on player's level, but turns out unlocking all.

acidmanifesto:
Superceeded by https://github.com/azerothcore/azerothcore-wotlk/pull/9729

PR #10402, opened 2022-01-28
Fix (DB/spell_proc_event) Lawbringer's T1 8 Set Bonus Fix

TheDdraig:
## Changes Proposed:
-  This fixes the Paladin's Lawbringer 8 set bonus exploit (currently procs on every hit)
-  

## Issues Addressed:

- Was reported from Chromiecraft that the 8 set for paladins tier 1 would constantly proc

## SOURCE:
If you go onto azerothcore test paladin tier 1 8 set it will constantly go off, you can also see this isnt the case from youtube vids, this is also a formula for it

## Tests Performed:

- Performed
- Build

## How to Test the Changes:

1. Make a Paladin add tier 1 set and with 8 set bonus and go attack any mob and you will see it go off constantly, seems no % chance was defined, check combat log and circle of healing is the proc spell for that set bonus.

2. The proc chance is a guess, as players have commented it was only 3%, however that seems too low so I have set it to 20% with a 50 sec cooldown as melee/auto attack procs have a cooldown to prevent exploiting. Please test it fully, if it feels off  you can easily adjust, this may be a discussion between the staff as a sniff would not provide the actual chance, however there was a formula in: https://tbc.wowhead.com/spell=21747/lawbringer#comments


acidmanifesto (approved):

data/sql/updates/pending_db_world/rev_1643386602617683649.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1643386602617683649');
+-- Fixes Paladins Lawbringer's 8 set bonus
+INSERT INTO `spell_proc_event` (`entry`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `procFlags`, `procEx`, `ppmRate`, `CustomChance`, `Cooldown`) VALUES
```
Kitzunu, line 3: always delete before an insert.

Kitzunu (changes requested):

malcrom (approved):

Kitzunu (approved):

55Honey (approved):

claudiodfc:
##### Test
The following timestamps are the times when the spell procced during a 5 minutes test:
- 0.25
- 1.17
- 2.09
- 3.00
- 3.55
- 4.48

##### Results
Overall, It seems like the proc chance is low and doesn't make the set bonus much viable, but I believe from wowhead comments that this was also the case for Vanilla.

claudiodfc (approved):

Issue #3648, opened 2020-11-05
Stats Limits is missing


w5860363:
Current behaviour: Stats Limits
Expected behaviour: players stat limit 75%
Steps to reproduce the problem:
1.update item 10000 stat
.additem
Players Dodge <100%
Also, I found a trinitycore commit that was deleted. Is that intentional? https://github.com/TrinityCore/TrinityCore/commit/e18571346400281861490b36c4d35a381bc80dd8
Branch: Master.
AC hash/commit: https://github.com/azerothcore/azerothcore-wotlk/commit/2e809055a07e5965f832a989839d4758d734a5a4
Operating system: Linux, Ubuntu 14.04 and Windows 7.
Modules: No.
Other Customizations: No.

Kitzunu:
That commit would have been for Mop

placemolle:
I imported the referenced commit [here](https://github.com/placemolle/azerothcore-wotlk/tree/fix-3648).
But before creating a pull request, i'm wondering if this is the correct Blizz-like behavior.
Right now, dodge/parry/block/crit are capped at 95%, but the caps are disabled by default.
If this is the right behavior for WOTLK, why disabled by default?

Helias:
@placemolle, unfortunately, I don't know the answer, but I think that you can open the PR and may create a discussion there ;-)

PR #10403, opened 2022-01-28
feat(Core/Player): Stats limitation

Kitzunu:
* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/e18571346400281861490b36c4d35a381bc80dd8)

* closes https://github.com/azerothcore/azerothcore-wotlk/issues/3648

Co-Authored-By: Corbi <12052933+corbi@users.noreply.github.com>

## Changes Proposed:
-  Add option to choose stat limitation if one wants
- Disabled by default

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3648

## Tests Performed:

- None

## How to Test the Changes:

1. Enable the config
2. Set the limitation % to what you like
3. See that Crit/Dodge/Parry/Block percentages cannot exceed the limitation


Nefertumm (approved):

Goatform:
Why not simply set it to `100.0` and in case someone for some reason really wants to apply cap for stat(s) then all they would need to do is decrease the number (enable the cap feature)

src/server/game/Entities/Unit/StatSystem.cpp
```diff
@@ -612,6 +612,12 @@ void Player::UpdateBlockPercentage()
         value += GetTotalAuraModifier(SPELL_AURA_MOD_BLOCK_PERCENT);
         // Increase from rating
         value += GetRatingBonusValue(CR_BLOCK);
+
+        if (sConfigMgr->GetOption<bool>(CONFIG_STATS_LIMITS_ENABLE))
```
acidmanifesto, line 616: need to add #include "Config.h" to top

acidmanifesto (changes requested):

src/server/game/Entities/Unit/StatSystem.cpp
Kitzunu: ah yea, forgot about that kek

Winfidonarleyan (approved):

Nyeriah (approved):

acidmanifesto (approved):

PR #10464, opened 2022-01-31
fix(DB): Fire elemental invasion event start NPC

claudiodfc:
## Changes Proposed:
-  Removed wrong entry added to the event
- Added the correct GUID for the Baron in Fire elemental Invasion

## Issues Addressed:

- Closes not reported issue

## SOURCE:

not needed

## Tests Performed:

- Not tested but I have other 3 PRs for the other elemental bosses with the correct GUID so if it works for them, it works for Baron as well.

## How to Test the Changes:

1. Go to the middle of Ungoro, you can use command `.tele ungoro` or `.go c 247200`
2. Kill the boss that spawns there
3. Use the command to start the Elemental Invasion event, which should be something like `.event start 13``
4. The boss has to re-spawn

None

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1643502379399332300.sql
+++ b/data/sql/updates/pending_db_world/rev_1643502379399332300.sql
@@ -1,9 +1,0 @@
-INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1643502379399332300');
-
-DELETE FROM `command` WHERE `name` = 'go quest';
-INSERT INTO `command` (`name`, `security`, `help`) VALUES
-('go quest', 1, 'Syntax: .go quest <starter/ender> <quest>.\nTeleports you to the quest starter/ender creature or object.');
-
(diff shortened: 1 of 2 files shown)


acidmanifesto (approved):

Lixar91:
Hello!

This is the first time I'm testing a PR. All recommendations are welcomed :smile: .

AC version where test was made: `AzerothCore rev. 210029095af0 2022-02-10 14:50:33 +0100 (master branch) (Unix, RelWithDebInfo, Static)`
OS where compiled and used: Linux Mint 20.3

Before the test the boss can be found and killed but when you activate the event with `.event start 13` the boss doesn't re-spawn as mentioned. Here you can see a screenshot as proof where I killed the boss and after that, I started the event:

antes

When I apply the fix, I am not able to find the boss unless the event is active. If it is active, the boss spawns correctly. 

Nevertheless, I'm not sure what is supposed to happen if you stop and start the event. What happened to me was that the corpse of the boss was re-spawned, as you can see here:

final

PR #10471, opened 2022-01-31
Refactor(Core/Hook): OnPlayerReputationChange

Nefertumm:
## Changes Proposed:
-  Make the hook bool instead of void.
-  Move the hook call, so if the hooks call return false, the default behavior wont execute. i.e. this would allow to avoid rep change.

## Issues Addressed:

- Closes none

## Tests Performed:

- Tested in-game with https://github.com/azerothcore/mod-cfbg/pull/92 and without it.

## How to Test the Changes:

1. The rep gain should not be affected, so go around and gain rep.


temperrr:
Breaks Eluna

/mnt/data/azerothcore/modules/mod-eluna-lua-engine/src/ElunaLuaEngine_SC.cpp:660:10: fatal error: virtual function 'OnReputationChange' has a different return type ('void') than the function it overrides (which has return type 'bool')
    void OnReputationChange(Player* player, uint32 factionID, int32& standing, bool incremental) override
    ~~~~ ^
/mnt/data/azerothcore/src/server/game/Scripting/ScriptMgr.h:1008:18: note: overridden virtual function is here
    virtual bool OnReputationChange(Player* /*player*/, uint32 /*factionID*/, int32& /*standing*/, bool /*incremental*/) { return true; }

Nefertumm:
https://github.com/azerothcore/Eluna/pull/18
https://github.com/azerothcore/mod-cfbg/pull/92

temperrr:
Getting the same error.

temperrr:
Tested

- [x] CFBG Rep gain in av normally and playing as normal faction, works as expected now
- [x] Quest rep gain (in world questin) not affected
- [x] world kill rep gain not affected

Nyeriah (approved):

PR #10535, opened 2022-02-05
Fix(Core/MC): Ragnaros dying while submerging

Nefertumm:
## Changes Proposed:
-  On Ragnaros death, emerge him immediately.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/2969

## Tests Performed:

- Not tested.

## How to Test the Changes:

See Honey's comment on the issue.


Nyeriah:
Most bosses that I've seen (Kael'thas, Ionar, Slabhide) that got some sort of transition phase don't die until the transition has finished. I haven't checked Ragnaros, but it's likely to be the case there as well. They still take damage, but their HP won't drop below 1.

Nyeriah (approved):

temperrr:
tested; 
he will now get down to 1 hp only, and will not die during the phase. You will need to deal 1 damage to get him down after submerge.

Issue #10429, opened 2022-01-29
[BWL] Nefarian Issues


temperrr:
### Current Behaviour

Class calls effect all classes
Mage class call does not work
shaman class call doesnt work
priest class call doesnt work
No tail swipe being used
rogue class call no teleport
nef is spawned on his throne whilst we a re killing him
nef loot table is wrong (dropped 2 swords of akhandi)

### Expected Blizzlike Behaviour

Class calls should only affect said class.
More class calls could be not working (Couldnt test them all)
Nef should not sit on his throne when you are fighting his dragon form 

### Source

https://classic.wowhead.com/guides/nefarian-blackwing-lair-strategy

### Steps to reproduce the problem

.go c id 10162

### AC rev. hash/commit
fb8a38c7b3067414a8aaa4450756cc9f9143239a

### Operating system

Linux

### Custom changes or Modules

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
- [lua-custom-worldboss](https://github.com/55Honey/Acore_CustomWorldboss)
- [lua-send-and-bind](https://github.com/55Honey/Acore_SendAndBind)
- [lua-temp-announcements](https://github.com/55Honey/Acore_TempAnnouncements)
- [lua-zone-debuff](https://github.com/55Honey/Acore_ZoneDebuff)

PR #10536, opened 2022-02-05
Fix(Core/BWL): Nefarian

Nefertumm:
## Changes Proposed:
-  Implement class calls.
-  Fix Victor Nefarius being visible while we fight the drake form.
-  Class call should only affect the class called (almost all, except dk).
-  Delete double Ashkandi drop from loot table.
-  Class call infernals (warlocks) should not drop gold on dying.
-  Nefarian script moved to the new script system.
-  Nefarian should use Tail Lash.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10429

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. .go c id 10162
2. Engage

- [ ] There might be more issues not reported. I only fixed what was reported.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1643946548773673200.sql
+++ b/data/sql/updates/pending_db_world/rev_1643946548773673200.sql
@@ -13,9 +13,9 @@ INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
 (23414, 'spell_class_call_handler'), -- Rogue
 (23424, 'spell_corrupted_totems'),
 (23603, 'spell_class_call_polymorph'),
-(23397, 'aura_class_call_berserk'),
 (23410, 'aura_class_call_wild_magic'),
-(23418, 'aura_class_call_siphon_blessing');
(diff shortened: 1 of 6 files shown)


temperrr:
Tested

- [ ] Reset in phase 1 is broken  (I got MCed and fight reset) image
- [ ] Resetting in phase 2 gives 
image

Won't check class calls yet, but with a next raid testing.

temperrr:
https://github.com/chromiecraft/chromiecraft/issues/2973

temperrr:
image


After killing the amount of drakonids, he is just chilling there.

Nefertumm:
Only thing to improve: Do not teleport to the same position again.

temperrr:
Other stuff works, new issues:

Dying during phase transition results in no reset, you can also spawn multiple nefs this way to farm him possibly?

image
image


Minions during phase 1 should be selected randomly but only 2 kinds. 2 kinds per side + chromatics on both side.
image
https://classic.wowhead.com/guides/nefarian-blackwing-lair-strategy

Issue #10540, opened 2022-02-05
server crash 


mpfans:
### Current Behaviour

https://gist.github.com/mpfans/113c9e78751faa6c9e613f7b78eac7ac

### Expected Blizzlike Behaviour

none

### Steps to reproduce the problem

none

### AC rev. hash/commit

AC> AzerothCore rev. 7b041946e163+ 2022-02-04 23:39:05 +0000 (master branch) (Unix, RELWITHDEBINFO, Static)
Using SSL version: OpenSSL 1.1.1f  31 Mar 2020 (library: OpenSSL 1.1.1f  31 Mar 2020)
Using Boost version: 1.71.0
Using MySQL version: 80028
Using CMake version: 3.16.3
Compiled on: Linux 5.4.0-91-generic
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /home/polly/data/maps. Total size: 252191207 bytes
vmaps directory located in /home/polly/data/vmaps. Total size: 658130721 bytes
mmaps directory located in /home/polly/data/mmaps. Total size: 2199390740 bytes
Using zhCN DBC Locale as default. All available DBC locales: zhCN 
Using World DB: ACDB 335.6-dev
Using World DB Revision: 2022_02_04_13
Using Character DB Revision: 2022_01_29_00
Using Auth DB Revision: 2022_01_23_00
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0

### Operating system

ubuntu 20.04

### Custom changes or Modules

> List enable modules:
- mod-anticheat
- mod-multi-client-check

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/pull/10546

UltraNix:
@mpfans Test #10546

PR #10546, opened 2022-02-05
fix(Scripts/Scholomance): Darkmaster Gandling crash

Kitzunu:
## Changes Proposed:
-  fix nullptr crash

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10540

## Tests Performed:

- none

## How to Test the Changes:

Unsure as I do not know how this boss works. But it is just a null check


src/server/scripts/EasternKingdoms/Scholomance/boss_darkmaster_gandling.cpp
```diff
@@ -302,7 +302,7 @@ class boss_darkmaster_gandling : public CreatureScript
                 SetGate(room, CLOSED);
                 SpawnMobsInRoom(room);
                 DoCast(target, GandlingPortalSpells[room], true); // needs triggered somehow.
-                if (target->GetGUID() == me->GetVictim()->GetGUID())
+                if (target->GetGUID() && target->GetGUID() == me->GetVictim()->GetGUID())
```
pklloveyou, line 305:
```
if (target->GetGUID() && me->GetVictim() && target->GetGUID() == me->GetVictim()->GetGUID())
```

Nyeriah:
Hm this won't fix the crash though :P

src/server/scripts/EasternKingdoms/Scholomance/boss_darkmaster_gandling.cpp
Nyeriah, line 305:
```suggestion
                if (target && me->GetVictim() && target->GetGUID() == me->GetVictim()->GetGUID())
```

Kitzunu: This won't do it either because we already check vicitm in the first if statement, L#299
```cpp
            if (spellinfo && spellinfo->Id == SPELL_SHADOW_PORTAL && target && me->GetVictim())

UltraNix (approved):

Nyeriah (approved):

Nyeriah:
Victim() is the unit being chased, while target is the spell target, so for aoe spells they won't be the same
target can't be null in that hook, because the spell will always hit something, the victim however... is not ensured to be valid, there might even not be a victim at all

Kitzunu:
makes sense, I forgot that it was a spell hook

PR #10549, opened 2022-02-05
fix(Core/Build): correct macos build

Winfidonarleyan:
## Changes Proposed:
- Fix build for macos


Kitzunu:
drop mac support when? :P

Winfidonarleyan:
> drop mac support when? :P

I'm curious, too.

Kitzunu:
I wonder if someone actually runs AC on mac, or if they run it in docker

iraizo:
most of them dont even use an Intel cpu and ac does not even support arm64e so why bother

Nyeriah (approved):

Issue #10051, opened 2022-01-07
Dreadnaught' Battlegear Last Set Bonus (Cheat Death)


jads147:
### Current Behaviour

Dreadnaught' Battlegear last set bonus effect proc's at 100% health.

### Expected Blizzlike Behaviour

Dreadnaught' Battlegear last set bonus effect should proc at 20% health and should only proc once.

### Steps to reproduce the problem

Happens everytime.

### AC rev. hash/commit

AzerothCore rev. 017772043d9d 2022-01-05 23:22:43 +0000 (master branch) (Win64, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1m  14 Dec 2021 (library: OpenSSL 1.1.1m  14 Dec 2021)
Using Boost version: 1.74.0
Using MySQL version: 80027
Using CMake version: 3.20.2
Compiled on: Windows 10.0.19043
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in C:/build/bin/RelWithDebInfo/data/maps. Total size: 252191207 bytes
vmaps directory located in C:/build/bin/RelWithDebInfo/data/vmaps. Total size: 658130721 bytes
mmaps directory located in C:/build/bin/RelWithDebInfo/data/mmaps. Total size: 2196365032 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS
Using World DB: ACDB 335.5-dev
Using World DB Revision: 2022_01_05_04
Using Character DB Revision: 2021_12_26_01
Using Auth DB Revision: 2021_10_14_00
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
> List enable modules:

### Operating system

Windows 10 x64

Shineslippers:
[Cheat Death](https://wowgaming.altervista.org/aowow/?spell=28845) proc at any % of health and resresheing every time on damage. YPlayers can't get T3, just in case.

https://user-images.githubusercontent.com/85256266/148584985-e6f448c7-0142-4bdd-8226-99d2599173ec.mp4

Nefertumm:
```diff
diff --git a/src/server/scripts/Spells/spell_warrior.cpp b/src/server/scripts/Spells/spell_warrior.cpp
[...]

PR #10561, opened 2022-02-06
Fix (Core/Scripts): Warrior T3 8P Bonus Fix

TheDdraig:
## Changes Proposed:
-  This fixes the tier 3 bonus fix on the 8 piece "cheat death" however this will need testing and the only thing that it doesnt fix is the proc once, however it doesnt really matter as soon as warrior is above 20% it will no longer proc, however let me know if you would rather a cooldown for it as the current system I dont believe it's possible to make it just proc once.

## Issues Addressed:
- Currently Dreadnaught set will proc it's 8 set bonus everytime you take damage, however it should only proc at 20% hp. I was unable to make it only proc once though, as currently it will keep on procing at 20%, however when you are above 20% the proc will stop and I believe this is the currently behaviour on TC.
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10051

## SOURCE:

It will clearly state on the item set bonus: https://tbc.wowhead.com/transmog-set=688/dreadnaughts-battlegear

## Tests Performed:

- Build
- Performed

## How to Test the Changes:

1. Make a warrior, go to level 60 or above and do .additem set 688 and then go take damage, try without this fix and then try with this fix.


acidmanifesto (approved):
//Approved by PuzzyWisardBrigade Commodore MDic

malcrom:
//PussyWizardEliteMalcrom also approves this fix .

PR #10565, opened 2022-02-06
feat(Core/Player): Implement option to convert excess honor points int…

Nyeriah:
…o money

## Changes Proposed:
-  Implement an option to conver excess honor points into gold after changing the honor cap
- 

## Tests Performed:

- tested ingame
- 

## How to Test the Changes:

1. UPDATE characters SET money = 0, totalhonorpoints = 76000;
2. log ingame
3. check if your honor converted into gold
4. tip: set MaxHonorPointsMoneyPerPoint to 10 and you'll get 1 gold after logging in


src/server/game/Entities/Player/Player.cpp
```diff
@@ -6105,7 +6105,15 @@ bool Player::RewardHonor(Unit* uVictim, uint32 groupsize, int32 honor, bool awar
 void Player::SetHonorPoints(uint32 value)
 {
     if (value > sWorld->getIntConfig(CONFIG_MAX_HONOR_POINTS))
+    {
+        if (int32 copperPerPoint = sWorld->getIntConfig(CONFIG_MAX_HONOR_POINTS_MONEY_PER_POINT))
```
Kitzunu, line 6109: use GetOption<int32>()

Nyeriah: Wrong. That's used when the configs are read from the text file, not from the storage when they are already loaded.
https://github.com/azerothcore/azerothcore-wotlk/pull/10565/files#diff-d88fa29cc529859e675b29c15bcddddb5fe915116df44cac561a52773107d94cR847

Kitzunu: According to Winfi it should be used like this https://github.com/azerothcore/azerothcore-wotlk/commit/b50984e9bfb66b496190a7b10839d1e1331dc311#diff-9dbcc2f32c3db710fb2f3843093a5aa606c9e5716599e5f540b4d1078951ef62R617

Nyeriah: There are two methods that can read configs, one is used to read the text file directly and the other reads the server memory.

To not be confused with the old load method that was replaced with the GetOption<type>

I don't really think it's efficient to load the text file for the config string every time the config is called. Modules do that because there's no other option, but the memory storage is probably more performance friendly and the correct way to access them during runtime.

Nyeriah: That said... I believe those configs you added on that other PR should've been added to the server storage (like the rest of them)

Kitzunu: Now I am no fan of the way the "new" configs work, that you define the default everywhere. And I do much more prefer the way it always have been done to be defined only once.

I made the suggestion only because I trust Winfi as he reworked the system

Nyeriah: I think there was a mistake when your PR got reviewed. It's not new, nothing changed, you can check and you'll notice every single config variable is still loaded the way it is in this PR.

You're right, it's also not good to define defaults more than once in case the same config is used multiple times.

temperrr:
Tested

Seems to work as described. Noticed that if you keep the config active, and honor gained when on cap will also be instantly transferred to silvers.

temperrr:
Tested,

Now no longer gives coppers when gaining honor when at cap.
Other behavior is still working as intended

PR #10633, opened 2022-02-11
Refactor(DB/spell): Rename fields according to their DBC equivalent

buddiman:
## Changes Proposed:
While working on https://github.com/azerothcore/Keira3/pull/1623 I noticed that in the table `spell_dbc` the EffectBonusMultiplier fields were named `Field227`, `Field228` and `Field229`. The name change would make the columns easier to identify and work with. (Also it makes using them in Keira easier). The new names would then be the same as in the DBC structure / the name commonly used.

## Issues Addressed:

- N/A

## SOURCE:

- [wowdev wiki](https://wowdev.wiki/DB/Spell)
- DBCStructure.h
https://github.com/azerothcore/azerothcore-wotlk/blob/2933a535ff0c037f00ec2c93e27ecef91ce5f849/src/server/shared/DataStores/DBCStructure.h#L1670-L1672

## Tests Performed:

- Query applied without errors
- Server is still usable without errors in the console/log.
- Casting a spell from the `spell_dbc` table worked fine.

## How to Test the Changes:

1. Apply the sql script
2. Look at the structure of the `spell_dbc` table, `Field227`, `Field228` and `Field229` should be no longer existent and replaced by the fields `EffectBonusMultiplier_1`, `EffectBonusMultiplier_2` and `EffectBonusMultiplier_3`

- The query only runs successful when the table structure has the exact fields `Field227`, `Field228` and `Field229`. When using updates in order, it is no problem. If the structure was modified manually then the script will fail.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1644590484974206900.sql
+++ b/data/sql/updates/pending_db_world/rev_1644590484974206900.sql
@@ -1,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1644590484974206900');
+
+ALTER TABLE `spell_dbc` CHANGE COLUMN `Field227` `EffectBonusMultiplier_1` FLOAT DEFAULT 0 NOT NULL;
+ALTER TABLE `spell_dbc` CHANGE COLUMN `Field228` `EffectBonusMultiplier_2` FLOAT DEFAULT 0 NOT NULL;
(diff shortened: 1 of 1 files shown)


FrancescoBorzi (approved):

FrancescoBorzi:
To be merged as soon as the `node-dbc-reader` is ready too

PR #10720, opened 2022-02-17
fix(DB/npc_text): Fix Kraz Gossip Text

ZhengPeiRu21:
## Changes Proposed:
-  Fix gossip text for Kraz in Zul'Aman
- Also switched object IDs for two reward satchel to match correct NPC; before Kraz and Ashli were backwards and give eachother's bag

## Issues Addressed:

- Closes unreported issue

## SOURCE:

Text already in DB, found in video here: https://youtu.be/xmNa4tDoYQc?t=201
See more discussion here: https://github.com/TrinityCore/TrinityCore/issues/19527

## Tests Performed:

- Before:
Capture
WrongPackage

- After:

## How to Test the Changes:

1. .npc add temp 24024
2. Talk to NPC, correct gossip and fixed bag

These NPC are still very hack. In correct behavior, they should be in a cage before boss is defeated, and when boss is defeated, cage is open, NPC does RP event, and gives reward. They should have different gossip while in cage, after boss is defeated, and after treasure is rewarded. Text for these event are in the DB, but instance script right now does not track individual treasure rewarded, and only spawns NPC after boss is defeated, so instance needs to be reworked to have all correct behavior. But this improvements is at least better than before!

TrinityCore scripts are about the same state as AzerothCore with these problems. Maybe is best to wait until TBC Classic is release this dungeon so all correct data can be sniffed then do a proper rewrite.


acidmanifesto (approved):

acidmanifesto:
This is just a fyi incase another review comes here. the failure is on the custom module build test that is not related at all to this PR.

PR #10724, opened 2022-02-17
fix(Core/Guild): Guild bank handling issues

Nefertumm:
## Changes Proposed:
-  After de13bf426e162ee10cbd5470cec74122d1d4afa0 we found an overflow issue within the core where the client is sending -1 for guild bank rights and slots if the player is the guild master. Previous this commit, the core forced the overflow. This caused DB issues when we tried to save those values (as the DB expects unsigned too), causing the items to not be saved.

## Issues Addressed:

- Closes none addressed

## Tests Performed:

- Built, test needed.

## How to Test the Changes:

1. Create a guild
2. Buy a tab.
3. You shouldn't receive a SQL error in worldserver.


Nyeriah (approved):

acidmanifesto:
image
bought numerous tabs. no errors. sorry. i thought i posted the follow up and well... not seeing it on this pr.

PR #10742, opened 2022-02-19
fix (core): server.log dump

acidmanifesto:
When Logger.entities.player.dump=4,Console Server is turned on you get alot of weird stuff happening when you delete a character.
image
looking at the server.log you will then understand that error code.
image
This gets rid of that insanity. It didnt even back up all the things that pdumps does so this one is just a really odd out of date feature/functon.

## Changes Proposed:
-  prevents the server.log from doing some weird sql query in server.log dealing with some alternative pdump when Logger.entities.player.dump=4,Console Server is turned on. 

- This does not effect the pdump command.

## Issues Addressed:

- Closes 

## Tests Performed:

- builds
- performs

## How to Test the Changes:

1. have logger Logger.entities.player.dump=4,Console Server
2. delete character
3. notice sql query is not filling up server.log file.

- [ ]
- [ ]


Nyeriah:
I suggested just adding a config in case someone relied on this feature instead of completely removing it.

acidmanifesto:
> I suggested just adding a config in case someone relied on this feature instead of completely removing it.

i am a bit against that idea. it doesnt dump all that is dump by the pdump command. It over inflates the log, leads to confusion. Currently there is a setting for it and thats the Logger.entities.player.dump=4,Console where they can jsut null it out by putting a # in front of it on the config.

Nefertumm:
I'm not sure about the commenting. I'd rather delete it and if something goes wrong just restore it.

acidmanifesto:
> I suggested just adding a config in case someone relied on this feature instead of completely removing it.



> I'm not sure about the commenting. I'd rather delete it and if something goes wrong just restore it.

let me know what you decide. I will follow it up from there

Nefertumm:
Delete it, we can always restore it.

acidmanifesto:
> Delete it, we can always restore it.

done

Nefertumm (approved):

PR #10757, opened 2022-02-20
fix(DB/Spells): Added internal proc cooldown to Nature's Grasp.

### LINKED ISSUE #9475: Nature's Grasp doesnt have ICD
### Current Behaviour

Nature's Grasp (rank 8) does not have an Internal Cooldown for the charges at the moment. If you get attacked very fast, you lose all stacks and the target gets rooted by 5s, instead of 10 (it refresh the root and get Diminish Return).

Nature's Grasp should have a small 1~2 seconds of internal cooldown.

### Source

https://www.bluetracker.gg/wow/topic/us-en/23425763051-natures-grasp-question/

''Nature's Grasp is a self buff that causes things that hit you to become rooted. That effect now has 3 charges. There is a small internal cooldown in between roots to keep say a rogue who hits you twice very quickly from wasting two charges. In your example, both the DK and pet should get rooted. If the DK breaks the roots and gets on you again, they will get rooted again (subject to the 45 sec duration).''

1. cast nature's grasp .cast 53312 with one char
2. get a rogue with 2 fast daggers (1.8 and 1.4 speed)
3. on the rogue, cast hemmorhage (.cast 48660)
4. watch nature's grasp stacks and root duration on rogue

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/f0a10fd3c10a9063713dd60a943be59cf7b6e757

### Operating system

Windows 10 x64

### Custom changes or Modules

_No response_

UltraNix:
Fixes #9475

## Issues Addressed:

- Closes #9475

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

cast nature's grasp .cast 53312 with one char
get a rogue with 2 fast daggers (1.8 and 1.4 speed)
on the rogue, cast hemmorhage (.cast 48660)
watch nature's grasp stacks and root duration on rogue

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1645377551356909900.sql
+++ b/data/sql/updates/pending_db_world/rev_1645377551356909900.sql
@@ -1,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1645377551356909900');
+
+DELETE FROM `spell_proc_event` WHERE `entry`=-16689;
+INSERT INTO `spell_proc_event` VALUES
+(-16689,0,0,0,0,0,0,0,2,0,0,1000);
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9049,22 +9049,6 @@ bool Unit::HandleProcTriggerSpell(Unit* victim, uint32 damage, AuraEffect* trigg
             ToPlayer()->RemoveCategoryCooldown(1248);
         }
     }
-    else if (auraSpellInfo->SpellFamilyName == SPELLFAMILY_DRUID)
-    {
-        // Nature's Grasp
-        if( triggerEntry->SpellIconID == 20 )
-        {
-            if (!victim)
-                return false;
-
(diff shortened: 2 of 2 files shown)


avarishd:
Tested before and after, everything **works as stated** in the issue/pr.

After digging a bit further I've stumbled upon this video: https://www.youtube.com/watch?v=FO-ULuoTuKk
Can't really judge from that (slow hitting melee), but at times it seems that ICD is 1,5s, but at other times it also looks like he consumes 2 stacks at the same time, Unfortunately nothing official.

Andrevs096:
It works. Though is it possible for the cooldown to be only on one target/target that wasted one charge?
example: if a rogue is hitting you, then it works perfectly

But in a scenario like, a Death Knight and his ghoul are hitting you, the nature's grasp proc roots on the DK and ghoul keeping hitting you without getting rooted, the internal cooldown resets and dk hits you before his ghoul does, so the DK get rooted again, while the ghoul is free

''There is a small internal cooldown in between roots to keep say a rogue who hits you twice very quickly from wasting two charges. In your example, both the DK and pet should get rooted. If the DK breaks the roots and gets on you again, they will get rooted again'' 
According to this i feel like it was different in retail

PR #10793, opened 2022-02-23
BREAKING CHANGE(Core/Config): Individual XP rate per bg

Nefertumm:
## Changes Proposed:
-  One rate per BG. This would be useful if we would like to encourage people to do certain BG.

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. Add the new configs to worldserver.conf
2. Set GiveXPForKills = 1 
3. Put some random number to Rate.XP.BattlegroundKill (or leave it as it is) and the other options.
4. See that your exp should be variable depending on the BG.

- [ ] Better description, really open to improvements there :P


src/server/worldserver/worldserver.conf.dist
```diff
@@ -2294,9 +2294,16 @@ Rate.XP.Pet       = 1
...
 #        Default:     1
 
-Rate.XP.BattlegroundKill = 1
+Rate.XP.BattlegroundKill     = 1
+Rate.XP.BattlegroundKillAV   = -1
+Rate.XP.BattlegroundKillWSG  = -1
+Rate.XP.BattlegroundKillAB   = -1
+Rate.XP.BattlegroundKillEOTS = -1
+Rate.XP.BattlegroundKillSOTA = -1
+Rate.XP.BattlegroundKillIC   = -1
```
Kitzunu, line 2306: Why not set all to 1? 

1 * default will still be default

Nyeriah: да, кажется лучше

Winfidonarleyan: Определённо

Nefertumm: @Kitzunu I developed this way so you can set them to 1 if BattlegroundKill = X where X > 1, so you can always override it. I tried with 0, but 0 and floats don' get along too well.

Kitzunu: I'd propose dropping the general XP rate and just setting all rates to 1.0f as we do the same on several other occasions.

Kitzunu: Because a general xp rate is not needed when you can twek them anyways

Nefertumm: Yeah, I can do that, I didn't want to because of repentance and to leave the config opt as it is for retro-compatibility.

FrancescoBorzi: I think it's better like it currently the PR is

src/server/game/Miscellaneous/Formulas.cpp
```diff
@@ -93,7 +93,37 @@ uint32 Acore::XP::Gain(Player* player, Unit* unit, bool isBattleGround /*= false
             xpMod *= creature->GetCreatureTemplate()->ModExperience;
         }
 
-        xpMod *= isBattleGround ? sWorld->getRate(RATE_XP_BG_KILL) : sWorld->getRate(RATE_XP_KILL);
+        if (isBattleGround)
+        {
+            switch (player->GetMapId())
+            {
+                case 30:  // AV
```
FrancescoBorzi, line 100: isn't there an enum mapping all these IDs? if not, shall we create one?

src/server/worldserver/worldserver.conf.dist
Nyeriah: I disagree:
1) the code looks so much cleaner with the 1x as default and it's more intuitive as that's how every other rates setting is currently implemented
2) the override in this scenario becomes obsolete because each setting can be changed individually, so if one wishes to keep them as they used to be they need to do nothing

Nefertumm: Fair enough, I'll change it then

src/server/game/Miscellaneous/Formulas.cpp
Nefertumm: Sadly no, there is no enum :( I would but I don't know where exactly, hehe

Kitzunu: map.h or something

FrancescoBorzi: ping ping :)

Nefertumm: Done, I added it on Battleground.h since it makes more sense than Map.h, more specific...

FrancescoBorzi (approved):

Nefertumm:
Added

Lynethris:
I got this issue with worldserver : 
Screenshot 2022-03-03 082329

Nyeriah:
> I got this issue with worldserver : Screenshot 2022-03-03 082329
> 
> I have try to extract a fresh data folder, but same thing.

This doesn't have anything to do with this PR. Please open a new issue about it.

Lynethris:
So test done and it's work: 

- [X]  Rate.XP.BattlegroundKillAV
- [X] Rate.XP.BattlegroundKillWSG
- [X] Rate.XP.BattlegroundKillAB
- [X] Rate.XP.BattlegroundKillEOTS
- [X] Rate.XP.BattlegroundKillSOTA
- [X] Rate.XP.BattlegroundKillIC

Images: 
<details>
Rate x1:
WoWScrnShot_030322_193819
Rate x5:
WoWScrnShot_030322_192855
</details>

PR #10823, opened 2022-02-25
fix(Scripts/Kalimdor): fix Azuregos respawn timer not persisting afte…

Nyeriah:
…r restarts

## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -3523,3 +3523,12 @@ void Creature::SetRespawnTime(uint32 respawn)
 {
     m_respawnTime = respawn ? GameTime::GetGameTime().count() + respawn : 0;
 }
+
+void Creature::SaveRespawnTimeToDB()
+{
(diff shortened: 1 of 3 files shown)


Bogir:
What is the difference between the old method and the new one? 
`void Creature::SaveRespawnTime()`
`void Creature::SaveRespawnTimeToDB()`
Both methods do same actions - saving same data to DB

maybe we should only add check to old method and remove new?
изображение

Bogir:
[Nefertumm](https://github.com/Nefertumm) 
Why do you merge it so fast? 1 day is not enough

Nyeriah:
No idea how I didn't see that, yes, it does the same thing, thanks

Bogir:
But something goes wrong with timer
it must be 10 days, but I have 1 day after Boss was killed
изображение

PR #10868, opened 2022-03-01
fix(DB): Gossip Option to Restore Missing Totem

ZhengPeiRu21:
## Changes Proposed:
-  Add gossip logic to restore missing totem if destroyed
-  Original issue was for Alliance version, Horde version was also missing so is also added

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3141

## SOURCE:

See original issue

## Tests Performed:

Capture

## How to Test the Changes:

1. .quest add 14111 or 14100
2. .quest complete (same quest)
3. .go quest ender (same quest)
4. Complete quest
5. Destroy totem
6. See new gossip option

The only spell that creates this item in the spell_dbc is the original version that requires reagents, so I believe adding item directly like this is correct logic.
I am not sure that this is the correct broadcast text for the gossip option, but I chose the one that looked correct. Please change if you can find another correct text!

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1646152616244410200.sql
+++ b/data/sql/updates/pending_db_world/rev_1646152616244410200.sql
@@ -1,0 +1,42 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1646152616244410200');
+
+SET @NPC := 35073;
+SET @GOSSIP_MENU_ID := 10631;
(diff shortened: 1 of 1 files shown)


temperrr:
Tested, works as described. Thanks for your pr!

Nefertumm (approved):

PR #10406, opened 2022-01-28
fix(Core): Save gameobject state on instances 

claudiodfc:
## Changes Proposed:
Currently, azerothcore doesn't save gameobject states on instances.
Whenever there's a re-start or crash, the instance's gameobjects and their states aren't saved, producing un-wanted behaviours and blocking instances at times.

Implemented CRUD for new table `instance_saved_go_state_data` that holds the states of gameobjects.

- When worldserver launches and gameobjects are loaded, this will check if this object's state exists on the DB and sets the previous state.
- On instance deletion (reset) these states are also removed based on the instance ID.
- Whenever a gameobject state changes inside a dungeon or raid, we save on the database the set state.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9422
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9468
- Closes part of https://github.com/azerothcore/azerothcore-wotlk/issues/9414
- Closes part of https://github.com/azerothcore/azerothcore-wotlk/issues/9529

## SOURCE:

No source

## Tests Performed:

- Tested in-game

## How to Test the Changes:

The following test case should be executed when in a party and solo, to see if the instances are resetting properly.

If possible testing some dungeons like the ones mentioned on the issues would be appreciated.

#### Test case 1
1. .tele culling
2. .go gameobject 67455
3. .gobject activate 67455
4. .ser res 1
5. Re-launch Worldserver
6. Check if book-case is still open
7. Repeat steps 1, 2 and 3 but Reset all instances after repeating the first step.

## Known Issues and TODO List:

- [x] There might be some crash so if possible doing other dungeon test such as the Deadmines would help out. I wasn't able to reproduce a crash again.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #269, opened 2016-11-30
Fix Deadmines door lever not opening door.

devovh:
**Changes proposed:**

-  Fix Deadmines door lever not opening door.

**Target branch(es):** 1.x/2.x etc.
master

**NOTE** You no longer need to squash your commits, on merge we will squash it for you. (GitHub added a new feature)

PR #10874, opened 2022-03-02
fix(Core): Save gameobject state on instances.

devovh:
https://github.com/azerothcore/azerothcore-wotlk/pull/10406 - Fix this PR.


acidmanifesto:
That pr never got merged and appears abandon. Essentially there is nothing to fix

acidmanifesto (changes requested):
Please replace each tab used with 4 spaces:
image

devovh:
I have tested everything and it works correctly, the objects that perform the opening or closing functions on the door work flawlessly. In most cases, it works even better because it retains the door status. I have already solved this with another method, but there was no integration with the database when it comes to the status of opening or closing a given door or other objects related to this function. Here is a link to a patch that was already resolved, it works the same, only here the status in the current PR is included.

Old fix, link:
https://github.com/azerothcore/azerothcore-wotlk/pull/269

PR #10879, opened 2022-03-02
(DB/auth) Delete auto incrementable in account

pangolp:
## Changes Proposed:
-  Previously, about 10 accounts were created by default, so when they were removed, the auto incrementing option was left in the table structure.

## Issues Addressed:

- Closes 

## Tests Performed:

- Generally, when I create a server, I have to remember to remove that option, because otherwise, the accounts start to be generated from id 11.

## How to Test the Changes:

1. If you remove the option, the beads start from id 1. Otherwise, they start at id 11.


Nyeriah (approved):

pangolp:
Thanks @Nyeriah

Issue #10864, opened 2022-03-01
quest id 9967 is can not finished


5266869:
### Current Behaviour

quest id 9967 is can not finished
this quest is the quest id 9962's next step

### Expected Blizzlike Behaviour

it should finish when kill the brothers.

### Steps to reproduce the problem

i used the 13days ago version.
81f6de5
finish the quest id 9962 ,
the lead the next step quest id 9967

### AC rev. hash/commit

.

### Operating system

ubuntu 21

Kisskam:
I confirm that the mission is not completed by killing the 2 brothers.

*I don't speak English, I use a translator to be able to communicate.

5266869:
yes, please fix it. thanks

PR #10889, opened 2022-03-03
fix(DB/Scripts): Fix Ring of Blood Quest Credit

ZhengPeiRu21:
## Changes Proposed:
-  Add party that begins quest as target to receive quest credit (target was previously missing)
-  Quest can now be completed

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10864

## Tests Performed:

- Completed quest, now successfully marks as complete when both brothers are killed
- Killing only one brother does not complete quest, as intended

## How to Test the Changes:

1. .go quest starter 9967
2. Accept quest
3. Defeat enemies
4. Receive credit

I think this should be best way of assign credit only to correct party. There is possibility that if two parties are start this quest at the same time, that credit will not go to first party, but I don't know how Blizz-like way to handle this. I think Blizz may prevent other parties from accept same quest while it is running to avoid this issue, but that I think is a separate issue.


Kisskam (approved):
Tested and confirmed the changes if they solve the problem.

Nefertumm (approved):

PR #10917, opened 2022-03-06
fix(Core/Mails): Do not load expired mails from database.

UltraNix:
## Tests Performed:

- Not tested.

## How to Test the Changes:

Get 140 AH mails


Kitzunu:
Why dont we delete expired mails?

Nyeriah:
> Why dont we delete expired mails?

We do, every 6 hours from what I understand. However, if you have a gazillion AH mails it is not fast enough and you'll find yourself stuck with too many pending sale mails.

Kitzunu:
Then we can just delete them on load instead of skipping them? Or will that be too big of a load for the DB to handle everytine?

UltraNix:
@Kitzunu We can't, because sometimes expired mails must be returned to sender.

Kitzunu:
Ah, fair point, and I guess that happens every 6 hrs then

UltraNix:
Yes.

Kitzunu (approved):

temperrr:
Tested:

After buying 140 items (or selling 140 items) I get the message when inbox is full. After clearing the inbox I can instantly sent mail again to these characters.

However, this is now spamming console after logging in on the char that sold all the items on AH, and I dont think we want this spam in wordconsole, if ever?

image

UltraNix:
@temperrr Done.

PR #10526, opened 2022-02-04
feat(Core/Creature): Implement quest_greetings table

Kitzunu:
* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/a64d1ec51eb11fd01a26d08a09de3d8d7dd89675)

* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/6dda09818d32fc715549cda3e8bd6fa806425b27)

Co-Authored-By: ForesterDev <11771800+ForesterDev@users.noreply.github.com>
Co-Authored-By: Giacomo Pozzoni <giacomopoz@gmail.com>

## Changes Proposed:
-  Implement quest_greetings table
-  https://github.com/azerothcore/wiki/pull/686

## Issues Addressed:

- Closes 

## Tests Performed:

- Builds
- Tested in-game

image

## How to Test the Changes:

1. 
```sql
INSERT INTO `quest_greeting` (`ID`, `Type`, `GreetEmoteType`, `GreetEmoteDelay`, `Greeting`, `VerifiedBuild`) VALUES (341, 0, 0, 0, 'I don\'t have much time for idle talk, $N.  I\'ve got to get this bridge rebuilt before the rains come.  I\'ve finished every project on-time and under budget and I\'m not about to start slipping now.', 0);
```

2. .go c id 341
3. Open gossip and see his quest_greeting is showing if there are 2 or more options there

to get it to show:

.q add 345

Also check that other gossips and quest givers works as they should

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #10928, opened 2022-03-07
fix(DB/quest_greeting): Populate quest_greeting table

ZhengPeiRu21:
## Changes Proposed:
-  Add quest_greeting data from TrinityCore and Vmangos

## SOURCE:

TrinityCore and Vmangos

## Tests Performed:

- Several NPCs spot-test, all seem to work as intended:
Capture

## How to Test the Changes:

1. Go to any NPCs mentioned in PR when multiple quest or gossip available
2. Test that Quest Greeting is appear correctly

There are some NPCs that have both quest_greeting text and a gossip menu with gossip flag. This is consistent with Vmangos and TrinityCore, so I think it is okay, but it does seem like there would be no reason to have quest_greeting text in these case because it would be override by the gossip_menu text. An example is creature 240.

I have also not added the quest_greeting_locale data yet. Both TrinityCore and Vmangos are very incomplete with locale data, so I think we will need a separate sniff effort to collect this.


data/sql/updates/pending_db_world/rev_1646669102474392500.sql
```diff
@@ -0,0 +1,200 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1646669102474392500');
+
+DELETE FROM `quest_greeting` WHERE `ID` BETWEEN 234 AND 186420;
```
acidmanifesto, line 3: this makes me a bit nervous as that is a pretty huge range between. are we certain you got everything between this range in the insert below?

acidmanifesto (changes requested):

data/sql/updates/pending_db_world/rev_1646669102474392500.sql
ZhengPeiRu21: Yes, the table is currently empty. It was just introduced by https://github.com/azerothcore/azerothcore-wotlk/pull/10526 this past weekend. However, if we would prefer I can have it only remove the entries explicitly added. Should I make that change?

ZhengPeiRu21: I have updated PR to only remove the IDs only specifically added beneath. Thank you for the suggestion! I think it is a good change.

acidmanifesto (approved):

Kitzunu (approved):

Bogir:
Can we find localized strings in data-files extracted from localized clients?

ZhengPeiRu21:
> Can we find localized strings in data-files extracted from localized clients?

From what I can know, I think quest greeting data is not in client data and needs to be sniffed, but I could be mistake.

PR #10942, opened 2022-03-08
fix (core/logger) Corrected count of Broadcast Text Locales shown by logger on server start

Bogir:
## Changes Proposed:
-  Previously, the server showed the number of lines for enUS instead of the amount of other localizations.
  
## Tests Performed:

- Tested by starting server. Now it shows correct value of records in `broadcast_text_locale`
- 

## How to Test the Changes:

1. Apply changes, compile and start the server. 
If something went wrong, you'll se old results:
изображение

In other case it must show correct result, something like this:
изображение
 

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -9246,6 +9246,7 @@ void ObjectMgr::LoadBroadcastTextLocales()
         return;
     }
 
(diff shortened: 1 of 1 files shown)


acidmanifesto (approved):
Whoa. Good Catch.

acidmanifesto:
Tested on my end. PR is good to go.
image

PR #10969, opened 2022-03-10
feat(db/locales): fix zhTW texts, part 2

Bogir:
Fix missing zhTW CompletedText locale text for world event quest
Taken from TrinityCore/issues/27866


LukasVolgger (approved):
I like the use of `REPLACE INTO` instead of `DELETE` an `INSERT`. Nice catch!

data/sql/updates/pending_db_world/rev_1646937861709527999.sql
```diff
@@ -0,0 +1,34 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1646937861709527999');
+
+REPLACE INTO `quest_template_locale` VALUES
```
Nefertumm, line 3: As the other PRs, please put the table columns :)

Nefertumm (changes requested):

data/sql/updates/pending_db_world/rev_1646937861709527999.sql
```diff
@@ -0,0 +1,35 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1646937861709527999');
+
+REPLACE INTO `quest_template_locale` (ID, locale, Title, Details, Objectives, EndText, CompletedText, ObjectiveText1, ObjectiveText2, ObjectiveText3,  ObjectiveText4, VerifiedBuild)
```
Nefertumm, line 3: backticks

ZhengPeiRu21:
This seem like the correct zhTW text.

data/sql/updates/pending_db_world/rev_1646937861709527999.sql
Bogir: fixed

PR #11039, opened 2022-03-15
fix(DB/Locale): Add missing zhCN/zhTW quest_offer_reward_locale

Bogir:
Taken from TrinityCore/commit/6b92aa3da4e252f315daa55b7a8b23c2b43f5bfe


ZhengPeiRu21:
The zhCN quest data is correct and include a few that is missing from AzerothCore, although many is already present. The zhTW data appears to be a copy of the zhCN data and is incorrect - although maybe having incorrect text is better than currently totally missing?

PR #11040, opened 2022-03-15
fix(DB/Locale): Add missing zhCN/zhTW quest_request_items_locale

Bogir:
TrinityCore/commit/747ee281dd2f22b1cc0c44683854c87b2e869d14


ZhengPeiRu21:
This data looks incorrect. The zhCN values are already present in AzerothCore, and the zhTW values are just copy of the zhCN changed to Traditional Chinese values instead of proper localization.

Bogir:
> This data looks incorrect. The zhCN values are already present in AzerothCore, and the zhTW values are just copy of the zhCN changed to Traditional Chinese values instead of proper localization.

I have noticed it. If in this case zhCN is enough, then I will close the PR ?

ZhengPeiRu21:
> > This data looks incorrect. The zhCN values are already present in AzerothCore, and the zhTW values are just copy of the zhCN changed to Traditional Chinese values instead of proper localization.
> 
> I have noticed it. If in this case zhCN is enough, then I will close the PR ?

If player is use the zhTW client, it is maybe better for them to see the string show in the PR then fall back to English. However, there are trouble, like NPCs and game place names having different localization, so it could be confusing. The best solution is have correct zhTW data, but it is many efforts to collect all of the data. I don't know if the team feels having incorrect but usable data is better for short term. Most important thing is to remember later to fix it!

PR #11080, opened 2022-03-19
fix(DB/Loot): Molten Core

Nefertumm:
## Changes Proposed:
-  Fixed a lot of issues with loot rules on MC.
-  Not UDB friendly, but this feels so much better.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3056
- Closes https://github.com/chromiecraft/chromiecraft/issues/3062
- Closes https://github.com/chromiecraft/chromiecraft/issues/3063
- Closes https://github.com/chromiecraft/chromiecraft/issues/3064
- Closes https://github.com/chromiecraft/chromiecraft/issues/3173
- Closes https://github.com/chromiecraft/chromiecraft/issues/3174
- Closes https://github.com/chromiecraft/chromiecraft/issues/3175
- Closes https://github.com/chromiecraft/chromiecraft/issues/3176
- Closes https://github.com/chromiecraft/chromiecraft/issues/3177
- Closes https://github.com/chromiecraft/chromiecraft/issues/3218

## SOURCE:

Several data from classic and corrections from VMangos.

## Tests Performed:

- Testing in progress.

## How to Test the Changes:

1. Check every loot from bosses. You can do this killing the boss and .respawn it to do it quickly.

- [ ] There is some odd rule that if you already have an epic unique item, you can't see it on the loot window unless you are the ML. This is not the correct behaviour since you should be able to see the loot always, but if you have you shouldn't be able to loot it.


data/sql/updates/pending_db_world/rev_1647689221083701700.sql
```diff
@@ -0,0 +1,430 @@
...
+(12000, 18808, 0, 0.0, 0, 1, 2, 1, 1),
+(12000, 18811, 0, 0.0, 0, 1, 2, 1, 1),
+(12000, 19139, 0, 0.0, 0, 1, 2, 1, 1);
+
+-- Ragnaros
+DELETE FROM `creature_loot_template` WHERE `Entry` = 11502;
+INSERT INTO `creature_loot_template` (`Entry`, `Item`, `Reference`, `Chance`, `QuestRequired`, `LootMode`, `GroupId`, `MinCount`, `MaxCount`) VALUES
+(11502,7078 ,0    ,40 ,0,1,0,1,9),
+(11502,17204,0    ,3  ,0,1,0,1,1),
+(11502,17982,0    ,95 ,0,1,0,1,1),
```
temperrr, line 313: this item needs to be removed  (11502,17982,0    ,95 ,0,1,0,1,1),

Nefertumm:
I'll fix that and the startup errors soon.

j-hdt:
@Nefertumm keep in mind that creating loot groups which have been 4 classes per faction in classic each, will now result in 1 loot set containing 4 classes and one containing 5 classes, not being equally sized any more. In classic, paladin and shaman werent available to both factions, thus resulting in even groups.

For MC, Ony and BWL have 1 pool with each item having an equal drop chance, and 2 items dropping from that pool sounds more reasonable. that way, the classes from the 5 class pool, which including shamans and paladins for both factions and 3 more classes, aren't in disadvantage here

temperrr:
> @Nefertumm keep in mind that creating loot groups which have been 4 classes per faction in classic each, will now result in 1 loot set containing 4 classes and one containing 5 classes, not being equally sized any more. In classic, paladin and shaman werent available to both factions, thus resulting in even groups.
> 
> For MC, Ony and BWL have 1 pool with each item having an equal drop chance, and 2 items dropping from that pool sounds more reasonable. that way, the classes from the 5 class pool, which including shamans and paladins for both factions and 3 more classes, aren't in disadvantage here

Not too sure about this, 

For example on garr: Paladin/Shaman pool is 16.66% chance for your item, other pool is 20%. It's a small difference.
If you put them in equal pools (remember they also have 2 filler no tier items) you'll end up both being  18,18% 

You'll end up with a small % more fair for all classes, but it would go further away from "blizzlike" in one way right.

j-hdt:
> > @Nefertumm keep in mind that creating loot groups which have been 4 classes per faction in classic each, will now result in 1 loot set containing 4 classes and one containing 5 classes, not being equally sized any more. In classic, paladin and shaman werent available to both factions, thus resulting in even groups.
> > For MC, Ony and BWL have 1 pool with each item having an equal drop chance, and 2 items dropping from that pool sounds more reasonable. that way, the classes from the 5 class pool, which including shamans and paladins for both factions and 3 more classes, aren't in disadvantage here
> 
> Not too sure about this,
> 
> For example on garr: Paladin/Shaman pool is 16.66% chance for your item, other pool is 20%. It's a small difference. If you put them in equal pools (remember they also have 2 filler no tier items) you'll end up both being 18,18%
> 
> You'll end up with a small % more fair for all classes, but it would go further away from "blizzlike" in one way right.

yes, pools with additional items shouldnt really be touched - they end up having different amount of items anyway.
but if you have 2 separate pools for tier pieces only, you would create 2 uneven pool. In those cases (2 pools for tier pieces only, one pool 4, one pool 5 items, both pools are a 100% drop), I would suggest to go with one pool, all equal drop chances, and having 2 items from that pool (exclusive or inclusive of duplicate drops) happen. This would be closer to the original intention of 2 equal pools as it was in classic with 8 classes per faction.

temperrr:
Don't think they ever had two pools with tier pieces only

j-hdt:
may be more applicable to ragnaros, onyxia and bwl.

temperrr:
> may be more applicable to ragnaros, onyxia and bwl.

They use 1 pool with all items in it already that gets rolled twice.

Nyeriah:
what's with all the extra spaces :\

Nyeriah (approved):

PR #11081, opened 2022-03-19
fix(DB/Loot): Blackwing Lair

Nefertumm:
## Changes Proposed:
-  Fixed a lot of issues with loot rules on BWL.
-  Not UDB friendly, but this feels so much better.

## Issues Addressed:

- Closes none

## SOURCE:

Data gathering from classic and VMangos.

## Tests Performed:

-  Testing in progress.

## How to Test the Changes:

1. Go to BWL and kill every boss and check their loot, you can kill the boss and then do a .respawn to kill it again.

- [ ] Same as MC PR, There is some odd rule that if you already have an epic unique item, you can't see it on the loot window unless you are the ML. This is not the correct behaviour since you should be able to see the loot always, but if you have you shouldn't be able to loot it.


Nefertumm:
I'll fix the startup errors soon.

temperrr:
Loot is looking good, no issues (apart from the startup errors)

Nyeriah (approved):

PR #11130, opened 2022-03-24
fix(DB/gossip_menu): add missed gossip menu for  Thrall Warchief

Bogir:
Used by npc Thrall Warchief  (Zone Orgrimmar)

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1648131745456322279.sql
+++ b/data/sql/updates/pending_db_world/rev_1648131745456322279.sql
@@ -1,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1648131745456322279');
+
+-- for Zone Orgrimmar
+REPLACE INTO `gossip_menu` (`MenuID`, `TextID`) VALUES
+('3665', '5733'),
+('3666', '5734'),
+('3667', '5735'),
+('3668', '5736'),
+('3669', '5737'),
+('3670', '5738');


Lynethris:
Tested

data/sql/updates/pending_db_world/rev_1648131745456322279.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1648131745456322279');
+
+-- for Zone Orgrimmar
+REPLACE INTO `gossip_menu` (`MenuID`, `TextID`) VALUES
```
IntelligentQuantum, line 4: Please Use Insert Instead of replace

IntelligentQuantum (changes requested):

data/sql/updates/pending_db_world/rev_1648131745456322279.sql
Bogir: > Please Use Insert Instead of replace

Thanks for replying!

But... What for? For complicate the syntax and double the SQL-query? Usage of REPLACE is the best variant instead DELETE + INSERT
I never use it and don't recommend it to you. This synthax is a complete replacement for the double syntax.
If you want to control what wil be changed - you can see it all in (???, ???). So, don't need to delete it befor inserting. Just use REPLACE and be happy :)

Kitzunu: Because it is what AzerothCore has decided, see https://www.azerothcore.org/wiki/sql-standards#insert-amp-delete

data/sql/updates/pending_db_world/rev_1648131745456322279.sql
```diff
@@ -0,0 +1,10 @@
...
+('3665', '5733'),
+('3666', '5734'),
+('3667', '5735'),
+('3668', '5736'),
+('3669', '5737'),
+('3670', '5738');
```
Kitzunu, line 10: Also, these are ints, so don't use quotes around them

PR #11132, opened 2022-03-24
feat (DB/gossip_menu): add missed Watcher of Norgannon texts

Bogir:
## Changes Proposed:
- add missed pages and dialogues 

- [ ] - **We have duplicated gossip_menu_options with those texts: 57000 - 57005. Does them used somewhere?**


Nyeriah (approved):

Nyeriah (changes requested):
Actually, there are quotes around the numbers, those need to be removed

PR #11226, opened 2022-03-29
Core/Object: Fix WorldObject::SetZoneScript unsafe cast to InstanceMap

Kitzunu:
* cherry-pick commit (https://github.com/TrinityCore/TrinityCore/commit/e9992066a5ccd7f085a56de09ea027240e8da873)

Co-Authored-By: HelloKitty <5829095+HelloKitty@users.noreply.github.com>

## Changes Proposed:
-  Change unsafe casting in WorldObject::SetZoneScript
-  Not seen a case where it causes a crash, but better to prevent a maybe edge case than not :D

## Issues Addressed:

- Closes 

## Tests Performed:

- None

## How to Test the Changes:

1. When we create a Transport/Gob/CreatureFromProto we set SetZoneScript();
2. It's an edge case so there is no real test case as 99% of the times it would work with the old code.

### DIFF
--- a/src/server/game/Entities/Object/Object.cpp
+++ b/src/server/game/Entities/Object/Object.cpp
@@ -2272,8 +2272,8 @@ void WorldObject::SetZoneScript()
 {
     if (Map* map = FindMap())
     {
-        if (map->IsDungeon())
-            m_zoneScript = (ZoneScript*)map->ToInstanceMap()->GetInstanceScript();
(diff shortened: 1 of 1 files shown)


HelloKitty:
Won't ever crash in normal operation of Mangos/Trinitycore forks but in the rare case you rewrite Map::Update you may encounter the issue and crash outlined in the original PR: https://github.com/TrinityCore/TrinityCore/pull/27877

Nyeriah (approved):

PR #10971, opened 2022-03-11
fix(Scripts/BlackwingLair): Razorgore improvements

Nyeriah:
## Changes Proposed:
-  Script his evade modes after retail research and some other improvements
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1. Pull razorgore
2. .damage him
3. notice he kills all mobs if he dies during p1

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #11253, opened 2022-04-02
fix(Core/Spells): Razorgore's Explosion ignores LoS.

### LINKED ISSUE #11180: [BWL] Razorgore explosion doesn't kill the hidden players
### Current Behaviour

Not fixed by #10971, if players kill razorgore before that all eggs are destroyed, he explose. But the explosion doesn't impact the hidden players by a gameobject.

When razorgore explose, all players are killed.

### Source

_No response_

1. Hide a player with a columns or an other gameobject
2. Kill razorgore during the P1/P2

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. 9e7d1e2a3af7

### Operating system

Windows 11

### Custom changes or Modules

_No response_

UltraNix:
Fixes #11180

## Issues Addressed:

- Closes #11180

## Tests Performed:

- Not tested.

## How to Test the Changes:

Hide a player with a columns or an other gameobject
Kill razorgore during the P1/P2

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1648906248199850600.sql
+++ b/data/sql/updates/pending_db_world/rev_1648906248199850600.sql
@@ -1,0 +1,5 @@
+INSERT INTO `version_db_world` (`sql_rev`) VALUES ('1648906248199850600');
+
+DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId`=13 AND `SourceEntry`=20038;
+INSERT INTO `conditions` (`SourceTypeOrReferenceId`, `SourceGroup`, `SourceEntry`, `SourceId`, `ElseGroup`, `ConditionTypeOrReference`, `ConditionTarget`, `ConditionValue1`, `ConditionValue2`, `ConditionValue3`, `NegativeCondition`, `ErrorType`, `ErrorTextId`, `ScriptName`, `Comment`) VALUES
+(13,1,20038,0,0,31,0,4,0,0,0,0,0,'','Explosion targets players');
--- a/src/server/game/Spells/SpellInfoCorrections.cpp
+++ b/src/server/game/Spells/SpellInfoCorrections.cpp
@@ -4252,6 +4252,7 @@ void SpellMgr::LoadSpellInfoCorrections()
     ApplySpellFix({ 20038 }, [](SpellInfo* spellInfo)
     {
         spellInfo->Effects[EFFECT_0].RadiusEntry = sSpellRadiusStore.LookupEntry(EFFECT_RADIUS_50000_YARDS);
+        spellInfo->AttributesEx2 |= SPELL_ATTR2_IGNORE_LINE_OF_SIGHT;
     });
 
     for (uint32 i = 0; i < GetSpellInfoStoreSize(); ++i)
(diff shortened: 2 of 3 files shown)


Lynethris:
@UltraNix It's doesn't work because it's ignore totally the wall of the room and kill everything in the raid. We can maybe adjust the radius to the room (around 100y),  but I think that we will got the problem if the boss is kill near the vael door.

Nyeriah:
The spell initially had a shorter range, but it was reported as not enough. I didn't add the ignore LOS flag because I predicted this could happen. The spell targets can be conditioned to only kill the encounter mobs, however, then it no longer kills players...

Lynethris:
I'm not a dev but why not just use/create a function to kill only the players in the raid and then despawn the adds during the reset like with nef ? We would keep the explosion only for the visual.

Nyeriah:
The point is recreating scripts the way they're in the base game and there a spell is used to do it, we try to perform tasks using proper game mechanics instead of raw code. What you propose would be a simple solution but yes it's not the way it should be scripted, although it's still a solution...

Lynethris:
I understand. For the moment, this issue is minor a problem. Don't break a game mechanic and it's not a source of bug abuse. So we can wait to find a better solution.

UltraNix:
@Lynethris Why it's such a problem? When you fight with this boss, all gates are closed and you cannot escape. If all players should be dead, then I see no other way.

Nyeriah:
> @Lynethris Why it's such a problem? When you fight with this boss, all gates are closed and you cannot escape. If all players should be dead, then I see no other way.

I think he means it kills every mob inside the instance

Lynethris:
Yes, it is. The explosion kill players and mobs/bosses.

UltraNix:
Oh, that's a big problem ;)

UltraNix:
@Lynethris Try again now with the sql.

Lynethris:
It's work but it's break the boss reset (PR #10971).

UltraNix:
@Lynethris What is now the problem with that boss reset?

Lynethris:
> @Lynethris What is now the problem with that boss reset?

Sorry, not specify. When the boss explose, he is stay dead.

UltraNix:
@Lynethris Should be okay now.

Lynethris:
Test done and it's work. 

- [X] Razorgore reset properly
- [X] When Razorgore is killed before P3, he explodes and kill all players
- [X] Players hidden by a gameobject are hitted by the explosion

Just two minors issues:

- [ ] Spells like Divine shield can immune the player to the explosion
- [ ] When the boss explodes, it's take 30sec before the reset. Can we down this delay ?

Nyeriah:
> * [ ]  When the boss explodes, it's take 30sec before the reset. Can we down this delay ?

This is done after retail research, it's not exact but an aproximate value

Lynethris:
> > * [ ]   When the boss explodes, it's take 30sec before the reset. Can we down this delay ?
> 
> This is done after retail research, it's not exact but an aproximate value

Ok

UltraNix:
@Lynethris 
` Spells like Divine shield can immune the player to the explosion`
Should be okay now.

Lynethris:
Work. All is finally good. I pass it in "to be merged". Thanks @UltraNix !

Nyeriah (approved):

PR #11323, opened 2022-04-07
fix(DB): Many Missing and Fixed zhCN Translations

ZhengPeiRu21:
## Changes Proposed:
-  Fix many or incorrect zhCN locale text
-  Many thanks to @mpfans for providing the data! I only format it into commands and compared it with current database to see what was missing or updated.

## Issues Addressed:

- Closes unreported issue

## Tests Performed:

- SQL file runs without errors, spot checked many text to confirm it is correct zhCN text and not just zhTW in Simplified Chinese

## How to Test the Changes:

1. Apply update
2. Test with zhCN client


acidmanifesto (approved):

mpfans:
A large part of the data here comes from @Remberance .I have verified some data and think it can be merged. If there are other problems, I will submit PR for correction.thank you @ZhengPeiRu21

Nyeriah:
The file seems rather large, we tend to have issues loading large files, it gets stuck when loaded by the auto updater

mpfans:
> The file seems rather large, we tend to have issues loading large files, it gets stuck when loaded by the auto updater, needs testing



> 

There are no errors when I execute SQL statements manually, but I don't know if there will be problems with automatic update

Nyeriah:
> > The file seems rather large, we tend to have issues loading large files, it gets stuck when loaded by the auto updater, needs testing
> 
> > 
> 
> There are no errors when I execute SQL statements manually, but I don't know if there will be problems with automatic update

The problem is that unless users have a high SQL packet treshhold it causes issues, we cannot guarantee everyone has that so we try to use smaller files as otherwise people start having issues.

mpfans:
> > > The file seems rather large, we tend to have issues loading large files, it gets stuck when loaded by the auto updater, needs testing
> > 
> > 
> > > 
> > 
> > 
> > There are no errors when I execute SQL statements manually, but I don't know if there will be problems with automatic update
> 
> The problem is that unless users have a high SQL packet treshhold it causes issues, we cannot guarantee everyone has that so we try to use smaller files as otherwise people start having issues.

You're right. 
CC @ZhengPeiRu21 帮忙分隔下这个PR文件吧,建议用表来分隔,每个表为一个PR提交.谢谢你!

Nyeriah:
I’ll try importing it to see if MySQL errors

Nyeriah (approved):

ZhengPeiRu21:
> CC @ZhengPeiRu21 帮忙分隔下这个PR文件吧,建议用表来分隔,每个表为一个PR提交.谢谢你!

嗯，虽然这个好像已经merge了但是我下次会先好好分隔。

Nyeriah:
@ZhengPeiRu21 It's fine, tbh, this is difficult, ideally we'd prefer them to be kept together as much as possible, but if files get too large we start having issues. So it's hard to tell. I loaded this one using the autoupdater and it was fine, so I merged it.

Nyeriah:
It turns out that despite working fine for me, it is giving people issues :\

mpfans:
> It turns out that despite working fine for me, it is giving people issues :\

Does this commit need to revert and resubmit the split PR?

Kosiak:
Gives the error when I start worldserver. Closing connection to mysql. I probably could do some tinkering with mysql but we should not use large files like that maybe.

PR #11384, opened 2022-04-12
fix(Core/BWL): Nefarian class call spells

Nefertumm:
## Changes Proposed:
-  Make the class selection uniform.
-  Extend the class call filter to all the spell effects, this was causing some weird behaviours.

## Issues Addressed:

- Closes 

## Tests Performed:

- Not tested. Only built.

## How to Test the Changes:

1. Go to Nefarian with different classes (example: 2 hunters and 1 mage) and make sure the chances of getting a class call is the same for every class present (on the example before, hunters should have a chance of 50% instead of 66%)
2. Check that the effects are not applied to another class.

- [ ]
- [ ]


acidmanifesto (approved):

temperrr:
tested it work

PR #11476, opened 2022-04-22
fix (core): Null check the helpers

acidmanifesto:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -19154,7 +19154,10 @@ void Unit::ExitVehicle(Position const* /*exitPosition*/)
         return;
 
     GetVehicleBase()->RemoveAurasByType(SPELL_AURA_CONTROL_VEHICLE, GetGUID());
-    ToPlayer()->SetCanTeleport(true);
+    if (Player* player = ToPlayer())
+    {
+        player->SetCanTeleport(true);
+    }
     //! The following call would not even be executed successfully as the
     //! SPELL_AURA_CONTROL_VEHICLE unapply handler already calls _ExitVehicle without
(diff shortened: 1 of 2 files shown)


Nyeriah:
I'm not sure if this helps at all, you're still converting not checking object type, so if a creature casts a charge spell ToPlayer() will return null and it will lead to a crash.

```
if (ToPlayer())
   ToPlayer()->SetCanTeleport();
```
or
```
if (Player* player = ToPlayer())
   player->SetCanTeleport();
```

Should help

Nyeriah:
Yes, sure, it should be ```m_caster->ToPlayer()```

It was just an example

acidmanifesto:
> Yes, sure, it should be `m_caster->ToPlayer()`
> 
> It was just an example. Probably the same needs to be checked in the other PRs

im looking over the other, should still be fine already. I am still a bit curious how that crash was triggered.

acidmanifesto:
@Nyeriah besides my sleep depervation you were right. good catch.

RS-Degen:
This one too?

https://github.com/azerothcore/azerothcore-wotlk/blob/14eb63c2715a665be972aa771bd99cf48c816a0f/src/server/game/Entities/Unit/Unit.cpp#L13829

acidmanifesto:
> This one too?
> 
> https://github.com/azerothcore/azerothcore-wotlk/blob/14eb63c2715a665be972aa771bd99cf48c816a0f/src/server/game/Entities/Unit/Unit.cpp#L13829

Nope.

Kitzunu (approved):

Nyeriah (approved):

PR #11682, opened 2022-05-08
fix(Docker): unsafe repository ('/azerothcore' is owned by someone else)

avarishd:
Unsure if this is the correct way/location to do this, perhaps someone has a better idea on how to solve it.
This issue happens on `docker-desktop 4.7.1 / git version 2.36.0.windows.1` and it started happening after the recent [git security update](https://github.blog/2022-04-12-git-security-vulnerability-announced/)

## Changes Proposed:
-  Add the /azerothcore directory as a safe directory in docker

## Issues Addressed:

- Closes

> fatal: unsafe repository ('/azerothcore' is owned by someone else)
> To add an exception for this directory, call:
>
>       git config --global --add safe.directory /azerothcore
> --
    Could not find a proper repository signature (hash) - you may need to pull tags with git fetch -t
    Continuing anyway - note that the versionstring will be set to "unknown 1970-01-01 00:00:00 (Archived)"

## SOURCE:

https://stackoverflow.com/questions/72153342

## Tests Performed:

- Docker builds and runs without the error showing

## How to Test the Changes:

1. build with docker

- [ ]
- [ ]

### DIFF
--- a/apps/docker/docker-build-prod.sh
+++ b/apps/docker/docker-build-prod.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
 cd /azerothcore
+git config --global --add safe.directory /azerothcore
 
 bash acore.sh compiler build


apps/docker/docker-build-prod.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
 cd /azerothcore
+git config --global --add safe.directory /azerothcore
```
locus313, line 4: Instead of adding this into the docker-build-prod.sh file I would instead add it to the Dockerfile within the base image stage. Probably somewhere around this line: 
https://github.com/azerothcore/azerothcore-wotlk/blob/713cddc1ba534cd29b5d62c3f2fb4ed2626003c9/apps/docker/Dockerfile#L65  

It would need to be with RUN in front of the command.

acidmanifesto:
Is this PR abandon?

acidmanifesto:
PR marked as Abandon. Its been two months since request status change with no response.

PR #11735, opened 2022-05-14
(DB/Fix) Position Quest Body and Heart

pangolp:
## Changes Proposed:
-  The main idea of the pull request is to show the player the place where the mission should be performed. After investigating and testing several values in the `quest_poi` and `quest_poi_points` tables, it occurred to me, to check other missions that are similar, and I found that the ones that worked correctly, had the flag number 10. I did the test, and indeed, after applying the flag, the position within the map changes, showing the player, where to perform the mission.

## Issues Addressed:

- Closes

## Tests Performed:

- Inside the game
- Windows 10

## How to Test the Changes:

1. Try to perform the mission, before applying the changes, and you will notice that it does not indicate the position where you have to go, on the contrary, the mission points to the same NPC that gives you the quest.
2. Then apply the SQL and you will see what I leave in the images below.

## Screenshots

quest_druid_horde

quest_druid_alliance

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1652493536788924600.sql
+++ b/data/sql/updates/pending_db_world/rev_1652493536788924600.sql
@@ -1,0 +1,2 @@
+
+UPDATE `quest_template` SET `Flags`=`Flags`|10 WHERE `ID` IN (6001, 6002);


Kitzunu:
This should be sniffed.

pangolp:
> This should be sniffed.
> 
> You are adding QUEST_FLAGS_SHARABLE QUEST_FLAGS_PARTY_ACCEPT
> 
> Which definetly not all quests should have. So just blindly adding flags is wrong.

Then look for a solution to the problem, because the truth is that I can't find one. It's a temporary solution, it's not perfect, but at least it works. If not, you find a better solution.

pangolp:
Since the solution is not blizzlike, nor a sniff of the official one, but nevertheless, it solved the problem, I decided to close the topic, and then let someone else look for a sniff type solution. This way we do not damage the emulator, with false data.

PR #11736, opened 2022-05-14
(DB/AcoreString) Modifying message entry 474 475 476

pangolp:
### Changes Proposed:
-  The idea of the pull request is to modify the messages that provide information about the status of the missions, by using the `.lookup quest` command. Currently they are displaying some parameters, which do not receive information, so it is better to modify them.

### Issues Addressed:

- Closes

### Tests Performed:

- Inside the game
- Windows 10

### How to Test the Changes:

1. Use the command `.lookup quest $name` and you will see that it has a format, where it should include some parameters, but they are not passed to the message. For example:

```
6001 - [Body and Heart] Quest %s (%u) is already active.
6002 - [Body and Heart]
```

### Screenshot

acore_string

This is the message, after applying the SQL. It is shown in Spanish, because I am using that client, and I took the opportunity to add the translations.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1652503003334756400.sql
+++ b/data/sql/updates/pending_db_world/rev_1652503003334756400.sql
@@ -1,0 +1,4 @@
+
+UPDATE `acore_string` SET `content_default`=' [rewarded]', `locale_esES`=' [recompensada]', `locale_esMX`=' [recompensada]' WHERE `entry`=474;
+UPDATE `acore_string` SET `content_default`=' [complete]', `locale_esES`=' [completada]', `locale_esMX`=' [completada]' WHERE `entry`=475;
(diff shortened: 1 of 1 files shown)


acidmanifesto:
I dont have a Es or Mx client. Can u show what it was showing before the pr? Just so we can better understand the comparison?

Nyeriah:
This will break the message that shows once you do .quest complete

Kitzunu:
Just fix the script if it is broken.

pangolp:
> I dont have a Es or Mx client. Can u show what it was showing before the pr? Just so we can better understand the comparison?

The message currently displayed, with no change applied, is this:

```
6001 - [Body and Heart] Quest %s (%u) is already active.
6002 - [Body and Heart]
```

A %s and %u appear, but these parameters will never be passed to the message.

pangolp:
> This will break the message that shows once you do .quest complete, .quest add and others

In the `.quest complete` and `.quest add`, the parameters also appear without values.

pangolp:
> Just fix the script if it is broken.

Ok, then I close the PR.

Issue #11485, opened 2022-04-22
Potential Smartscript broken


ghost:
### Current Behaviour

Today, I have tried to fix this issue: https://github.com/azerothcore/azerothcore-wotlk/issues/10340

After many tries with smartscript and other values, nothing worked. So I started to do some research. On TC, all same fixes use on lever the smartscript "state_changed to 2" with a "gameobject_activation" to the target door. It's strange because it's actually the case.

And I found this: https://github.com/azerothcore/azerothcore-wotlk/issues/4784
This PR was fixed by UltraNix, and even it's not the same source after the fix the lever was able to open/close the door. I tested the actual situation, and it's broke again. The lever reset properly but after using the lever, you can't reopen the door. So I think that it's bound to a change with gameobject_state_changed or gameobject_activation.

If it's the case, there are maybe several unknown issues

Example  of one lever that doesn't work:
![image](https://user-images.githubusercontent.com/100411212/164766134-92345e4c-2b2a-4005-bc26-30f74a322fea.png)

### Expected Blizzlike Behaviour

I'm not totally sure, how this smartscripts work really. But IG, levers should be able open/close the doors.

### Steps to reproduce the problem

Look DB with  id: 101832
Look IG, the lever  can only close a door.

### AC rev. hash/commit

4ec69ba

### Operating system

Windows 11

Kosiak:
Pain in ass this :D

Issue #10340, opened 2022-01-24
Cannot open door in Deadmines


ishidinho:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/2855

After killing the first boss and opening the door, If you enter the door and use the lever to close the door, it stays stuck and the inside and outside levers wont open the door.

Tester:

The issue is confirmed. The door does not open after using the lever.

![WoWScrnShot_012422_101135](https://user-images.githubusercontent.com/62797536/150755949-17311cc5-d1d7-484b-a0dc-ee9894766f28.jpg)

There was a PR in 2016 to address this issue: https://github.com/azerothcore/azerothcore-wotlk/pull/269.

### Expected Blizzlike Behaviour

For the door to open/close when using the levers.

### Source

I don't believe a source is needed regarding a stuck door.

### Steps to reproduce the problem

1. Enter deadmines
2. Kill first boss
3. Open door
4. Go inside door
5. Use lever right after door to close the door
6. boom stuck

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/259b9133f68ef0d740fc871d59fab3d2791f33b4

### Operating system

Windows 10 21H1 19043.1415

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[...]

PR #11739, opened 2022-05-14
fix(DB/Smart Scripts) - Door Lever - On Game Object State Changed.

devovh:
**Changes Proposed:**

-  Fix all door's lever, Now Opening and Close all Door's - working.

**Issues Addressed:**

- Closes : Solving the problem with all doors regarding opening and closing.

**Tests Performed:**

- Inside the game
- Windows Server 2022 Datacenter

**How to Test the Changes:**
- .tele deadmines, kill first boss and test lever (door open and close)

**How to Test AzerothCore PRs:**

When a PR is ready to be tested, it will be marked as [WAITING TO BE TESTED].

REMEMBER: when testing a PR that changes something generic (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but especially check that the PR does not cause any regression (i.e. introducing new bugs).

For example: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but we should test Y and Z as well.


Nyeriah:
Please don't delete the PR template and include the necessary information.

devovh:
I understand, I quickly added this amendment that it is a serious bug related to opening and closing doors, for example through levers. At the beginning, in the old azerothcore revie based on sunwell core there was a similar problem, then there was a different solution to this problem, because it was enough to uncomment a certain entry in the source code regarding the ready object. Later, after updating this solution in the source code based on trinitycore, someone probably missed adding a smart script to the database, and therefore this problem still remained like it used to be with this uncommenting.

Maelthyrr:
Closes:
- https://github.com/azerothcore/azerothcore-wotlk/issues/10340
- https://github.com/azerothcore/azerothcore-wotlk/issues/11485

acidmanifesto:
Can you do a fetch upstream so we can see if that failure is already taken care of?

Issue #6129, opened 2021-05-30
Commander Felstrom Infinite Loot Farm


Azcobu:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [ ] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [x] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Original report: https://github.com/chromiecraft/chromiecraft/issues/735

Commander Felstrom, a rare mob found in Duskwood (https://wowgaming.altervista.org/aowow/?npc=771), drops his loottable upon succesfully casting his rebirth skill. 

Tester's notes: Confirmed. Every time he goes through the ressurection spell, he leaves behind a corpse filled with his loot - I found 4 greens in 4 corpses.

Screenshots showing looting successive Felstrom corpses while he's still alive:
![WoWScrnShot_053021_233329](https://user-images.githubusercontent.com/81782124/120107384-05bafa00-c1a0-11eb-8667-98a4a6267422.jpg)
![WoWScrnShot_053021_233334](https://user-images.githubusercontent.com/81782124/120107385-06ec2700-c1a0-11eb-810a-7261e80f622e.jpg)
![WoWScrnShot_053021_233339](https://user-images.githubusercontent.com/81782124/120107389-081d5400-c1a0-11eb-86c9-6838e2ef0548.jpg)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

If succesfully rebirthing, Commander Felstrom's corpse should be unlootable. Only an actual kill should yield loot.

##### SOURCE:

https://wowgaming.altervista.org/aowow/?spell=3488

##### STEPS TO REPRODUCE THE PROBLEM:

1.  .go c 5126
2.  Bring him to low HP: 1-120 Hp.
3.  Let him succesfully cast Felstrom Resurrection.
4.  Loot corpse, repeat steps 2 - 4.

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/3716ddf3e4e1ef138df334d01980fea019272001

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #11744, opened 2022-05-14
fix(DB/Creature): Commander Felstrom Infinite Loot Farm

avarishd:
Unsure if this is the proper way to do this, but it works.
Additionally, **I don't know how to stop giving EXP through SAI,** without messing with the spell's properties.
## Changes Proposed:
-  Set dynamicflag 128 on spell hit and despawn corpse after 2s

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6129

## SOURCE:

https://www.youtube.com/watch?v=DaMBxAdcslw - Corpse stays for 1-2 seconds
## Tests Performed:

- Tested in-game

## How to Test the Changes:

1. .go c 5126 - and look for the NPC in the rooms around (multiple spawns) /tar Commander
2. Bring him to low HP: 1-120 Hp. / .damage 1050 
3. Let him finish his cast
4. Notice he is unlootable and corpse despawns
5. Kill him without letting the the cast to go off, he is lootable 

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1652563880706540800.sql
+++ b/data/sql/updates/pending_db_world/rev_1652563880706540800.sql
@@ -1,0 +1,7 @@
+-- Commander Felstrom, set unlootable if resurrection goes through and despawn body after
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 771;
+DELETE FROM `smart_scripts` WHERE `entryorguid` = 771;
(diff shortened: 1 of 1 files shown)


Gultask:
Tested and working 👍

mpfans:
tested and work.
```
I don't know how to stop giving EXP through SAI
```
If the spell is successful, do you not calculate the kill, so you don't need to give experience value?

avarishd:
> If the spell is successful, do you not calculate the kill, so you don't need to give experience value?

Perhaps SPELL_EFFECT_INSTAKILL should reset all damage calculations so it doesn't give any credit, but this is unknown.
Also there is no option in SAI (Unless I'm blind) to disable EXP (maybe can be added?)

mpfans:
> > If the spell is successful, do you not calculate the kill, so you don't need to give experience value?
> 
> Perhaps SPELL_EFFECT_INSTAKILL should reset all damage calculations so it doesn't give any credit, but this is unknown. Also there is no option in SAI (Unless I'm blind) to disable EXP (maybe can be added?)

Is it better to restore health after successful casting?

avarishd:
> Is it better to restore health after successful casting?

No, https://db.rising-gods.de/?spell=3488

Nefertumm (approved):

Issue #11863, opened 2022-05-24
Licensing terms


grabthar:
### Current Behaviour

Hello dear community,
I think the project have some licenses issues, from what I understand:

All files in azerothcore are licensed under the AGPLv3 but most of them come from project like TC, mangos wich are under GPL.
The whole project can be licensed under AGPL but individual files must keep their original license. Even if you create a file under AGPL if you pull code that is under GPL the license of the file fall under GPL.

A file can't have dual license AGPL/GPL, but the project may have some files under GPL and other under AGPL

### Expected Blizzlike Behaviour

Correct license. Most of the core should be under GPL.

### Source

https://www.gnu.org/licenses/license-list.en.html#AGPLv3.0

> Please note that the GNU AGPL is not compatible with GPLv2. It is also technically not compatible with GPLv3 in a strict sense: you cannot take code released under the GNU AGPL and convey or modify it however you like under the terms of GPLv3, or vice versa. However, you are allowed to combine separate modules or source files released under both of those licenses in a single project, which will provide many programmers with all the permission they need to make the programs they want. See section 13 of both licenses for details.

you cannot take code released under the GNU AGPL and convey or modify it however you like under the terms of GPLv3, **_`or vice versa.`_**

 you are allowed to **_`combine separate`_** modules or source files released under both of those licenses in a single project

### Steps to reproduce the problem

Any

### AC rev. hash/commit

#a4af1136856604e836a4ed6d9f8a65fbca9357bc

### Operating system

All

55Honey:
Hello @grabthar 
I appreciate your concerns regarding our license. I suggest you read up on part 13, covering a mix of both licenses.

```Notwithstanding any other provision of this License, you have permission to link or combine any covered work with a work licensed under version 3 of the GNU General Public License into a single combined work, and to convey the resulting work. The terms of this License will continue to apply to the part which is the covered work, but the work with which it is combined will remain governed by version 3 of the GNU General Public License.```

Source: https://www.gnu.org/licenses/agpl.html

**Comment by grabthar** [comment]
Yes you can license azerothcore project under AGPL without problem, but each file in src/ state they are under agpl.

```
...
 * This program is free software; you can redistribute it and/or modify it
 * under the terms of the GNU Affero General Public License as published by the
 * Free Software Foundation; either version 3 of the License, or (at your
 * option) any later version.
```
**_you can redistribute it and/or modify it under the terms of the GNU Affero General Public License_**

[...]

PR #11873, opened 2022-05-26
fix(LICENSE): GPLv2 or later

Kitzunu:
This is an attempt to fix the licensing issue that has been present on AzerothCore.
If there are any references to AGPL or the Affero GPL left in the repository that is not intentional and the GPLv2 is the license to be used.

For more information about the case you can check up on https://github.com/azerothcore/azerothcore-wotlk/issues/11863

If there are any other issues with this PR, or if this PR does not fix the issue of AGPL being applied wrongly please let me know so we can correct it.

Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11863

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -166,8 +166,8 @@ This project exists thanks to:
 
 ## License
 
-- The new AzerothCore source components are released under the [GNU AGPL v3](https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-AGPL3)
-- The old sources based on MaNGOS/TrinityCore are released under the [GNU GPL v2](https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE-GPL2)
+- The AzerothCore source is released under the [GNU GPL v2 or later](https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE)
+- The old source based on MaNGOS/TrinityCore are released under the [GNU GPL v2 or later](https://github.com/azerothcore/azerothcore-wotlk/blob/master/LICENSE)
 
 
 ## Special thanks
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,8 +2,8 @@
 
 ## Warning
 The code of AzerothCore is shipped as it is without any form of warranty,
-and - except for third party libraries licensed under the AGPL 3,
-which you can read from the file "LICENSE-AGPL3".
+and - except for third party libraries licensed under the GPLv2 or later,
+which you can read from the file "LICENSE".
 
 ## Point of current development
 The project is currently hosted at https://www.azerothcore.org/ and developed under https://github.com/azerothcore
--- a/LICENSE
+++ b/LICENSE
@@ -1,620 +1,281 @@
-                    GNU AFFERO GENERAL PUBLIC LICENSE
-                       Version 3, 19 November 2007
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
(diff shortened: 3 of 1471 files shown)


Nyeriah (approved):

acidmanifesto (approved):

Nyeriah:
#first

ZhengPeiRu21:
This is maybe good opportunity to change this line, which incorrectly say that changing the banner message is against the terms of the license. It can still have threat, but is not against license terms technically, so is will be more scary when correct. Haha.

The line is here:
https://github.com/azerothcore/azerothcore-wotlk/blob/76d57065e266f2642035291fac13982d18751761/src/server/worldserver/worldserver.conf.dist#L1240

ihm-tswow:
Friends, you cannot undo a license just like that (even if it was incompatible!). 

If this only affects the files that was changed in the recent PR ~6 months ago, you might get away with it if you ask _every contributor to those files since then_. If this is intended to completely remove the AGPL from azerothcore, you'd need permission from every contributor to the AGPL modules since 2016. The alternative is to drop any AGPL-infected code, which you won't likely ever be able to do.

There is no completely pain-free way out of something like this, but if you just want to be on the good end of other cores, I think we're just happy you don't leave clauses keeping us hostage from editing our own files.

If you want any advice with this, you can reach out here or on discord. While I'm not a lawyer, I'm probably the person that's discussed this topic the longest in this scene to date, so I have a decent grasp on GPL conventions and how other projects use it.

Nyeriah:
> Friends, you cannot undo a license just like that (even if it was incompatible!).
> 
> If this only affects the files that was changed in the recent PR ~6 months ago, you might get away with it if you ask _every contributor to those files since then_. If this is intended to completely remove the AGPL from azerothcore, you'd need permission from every contributor to the AGPL modules since 2016. The alternative is to drop any AGPL-infected commits, which you won't likely ever be able to do.
> 
> There is no completely pain-free way out of something like this, but if you just want to be on the good end of other cores, I think we're just happy you don't leave clauses keeping us hostage from editing our own files.
> 
> If you want any advice with this, you can reach out here or on discord. While I'm not a lawyer, I'm probably the person that's discussed this topic the longest in this scene to date, so I have a decent grasp on GPL conventions and how other projects use it.


> Friends, you cannot undo a license just like that (even if it was incompatible!).
> 
> If this only affects the files that was changed in the recent PR ~6 months ago, you might get away with it if you ask _every contributor to those files since then_. If this is intended to completely remove the AGPL from azerothcore, you'd need permission from every contributor to the AGPL modules since 2016. The alternative is to drop any AGPL-infected commits, which you won't likely ever be able to do.
> 
> There is no completely pain-free way out of something like this, but if you just want to be on the good end of other cores, I think we're just happy you don't leave clauses keeping us hostage from editing our own files.
> 
> If you want any advice with this, you can reach out here or on discord. While I'm not a lawyer, I'm probably the person that's discussed this topic the longest in this scene to date, so I have a decent grasp on GPL conventions and how other projects use it.

I'm not sure what you're proposing. I understand the implications of this, I don't see other feasible way to achieve it.

Kitzunu:
I will leave this as-is.

Personally, I do not have the time to go through every commit or every file to see if or what was licensed correctly or not. I also do not really care about the license to the extent that I will research it anymore, but did this as I and others came to understand that the licensing issue that came before I ever contributed here was wrong. Perhaps we just leave this here and hope all contributors approve this change, which is also unlikely.

We are all out for the same thing, fair open sourcing and mutual respect amongst the wow emulation community.

If you know the way to correct it the right way, feel free to open a PR.

Kitzunu:
we can't just relicense everything. That would also break a lot of things.

Issue #11875, opened 2022-05-26
Tauren Shaman - Call of Earth Ritual Salve drop rate


Dejanos:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour
Upon taking the [Call of Earth](https://classic.wowhead.com/quest=1519/call-of-earth#videos:id=166856) quest as Tauren Shaman from [Seer Ravenfeather ](https://classic.wowhead.com/npc=5888/seer-ravenfeather) players gets to go and obtain 2 [Ritual Salves](https://classic.wowhead.com/item=6634/ritual-salve) from [Bristelback Shamans](https://classic.wowhead.com/npc=2953/bristleback-shaman) in the  Brambleblade Ravine. The drop rate currently appears to be too low, as out of 50 kills, I got the quest item to drop 9 times.

### Expected Blizzlike Behaviour
Upon looking the Classic WoW footage, all videos appeared to have 100% drop chance as every Shaman killed dropped the quest item on first try. WoWHead states the drop chance to be 14% and 12.34% on WoWWiki, but that is highly unlikely. Official AoWoW WoWgaming search engine also states the drop chance to be 80%.

### Source
[WoW Classic Salve drop on first kill 1 ](https://youtu.be/XOcfcDDR_Ok?t=195)
[WoW Classic Salve drop on first kill 2](https://youtu.be/XOcfcDDR_Ok?t=241)
[WoW Classic Salve drop on first kill 3](https://youtu.be/D93OxZ2hBQU?t=302)
[WoW Classic Salve drop on first kill 4](https://youtu.be/D93OxZ2hBQU?t=369)
[WoW Classic Salve drop on first kill 5 ](https://youtu.be/EBIcxB8iGMg?t=462)
[WoW Classic Salve drop on first kill 6 ](https://youtu.be/EBIcxB8iGMg?t=483)

WoWHead TBC Comment:
![2 shamans drop](https://user-images.githubusercontent.com/100415626/170602413-2377a301-577f-44f2-9ef4-44ed9e16c4f3.png)
[
AoWoW WoW Gaming database currently stating the drop rate to be 80%: ](https://wowgaming.altervista.org/aowow/?item=6634)

### Steps to reproduce the problem
[...]

PR #9379, opened 2021-11-27
fix(DB/Loot) Fix all starting area loot to work like Official

malcrom:
## Changes Proposed:
-  Update Current loot for starting area and modify creature loot Primary keys to allow double reference.

## SOURCE:
Days of killing stuff on official and days of writing SQL.

## Tests Performed:
- I have been testing and all seems well but would like more testers.

Will include a back out file below.

## How to Test the Changes:
1. Make a toon, Kill stuff

PR #11906, opened 2022-05-29
fix(DB/Creature_loot): Increase drop rate of Ritual Salve

Maelthyrr:
## Changes Proposed:
-  Update drop rate to 72 % (based with test  on retail. Look below)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11875

## SOURCE:

Drop rate confirm also with https://wotlkdb.com/?item=6634

## Tests Performed:

- None

## How to Test the Changes:

1.Create a Tauren Shaman
2. Start Call of Earth questline from [Seer Ravenfeather](https://classic.wowhead.com/npc=5888/seer-ravenfeather)
3. See you are having difficulties gathering 2 Salves after more than 5 Shamans killed.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1653810747543320600.sql
+++ b/data/sql/updates/pending_db_world/rev_1653810747543320600.sql
@@ -1,2 +1,2 @@
 -- Adjust drop rate of Ritual salve to 72% (based by a sniffer with 325 kills)
-UPDATE  `creature_loot_template` SET `Chance` = 72 WHERE `Entry` = 2953 AND `Item` = 6634;
+UPDATE `creature_loot_template` SET `Chance` = 72 WHERE `Entry` = 2953 AND `Item` = 6634;
--- a/data/sql/updates/pending_db_world/rev_1653810747543320600.sql
+++ b/data/sql/updates/pending_db_world/rev_1653810747543320600.sql
@@ -1,9 +1,2 @@
---
-DELETE FROM `creature_loot_template` WHERE (`Entry` = 2953);
-INSERT INTO `creature_loot_template` (`Entry`, `Item`, `Reference`, `Chance`, `QuestRequired`, `LootMode`, `GroupId`, `MinCount`, `MaxCount`, `Comment`) VALUES
-(2953, 0, 11111, 0.2, 0, 1, 0, 1, 1, 'Bristleback Shaman - (Small Pouch ReferenceTable)'),
-(2953, 0, 20000, 30, 0, 1, 0, 1, 1, 'Bristleback Shaman - (Grey 1-5 EXP 0 ReferenceTable)'),
-(2953, 0, 20014, 30, 0, 1, 0, 1, 1, 'Bristleback Shaman - (Food 1-5 EXP 0 ReferenceTable)'),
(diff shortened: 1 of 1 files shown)


IntelligentQuantum (approved):

Undeadhunter:
Tested OK  high drop rate but not 100% for sure on the Bristleback shamans

Maelthyrr:
> Tested OK high drop rate but not 100% for sure on the Bristleback shamans

It's intend. Sources say 80%.

Undeadhunter:
I badly worded that then, 
I didn't do a complete statistical analysis but out of the 25 kills I had 4 with no drop.

acidmanifesto:
Be advise your link for reference states the site is from pserver data

Maelthyrr:
I didn't know for woltkdb. But stay the same thing, videos show that the drop is near 100% but not 100%. And it's definitely not 14% as it is the case in the DB and WoWHead. What do we do ?

acidmanifesto:
> I didn't know for woltkdb. But stay the same thing, videos show that the drop is near 100% but not 100%. And it's definitely not 14% as it is the case in the DB and WoWHead. What do we do ? 4/25 it's make 86%.

I will leave that for your discretion. This is where we will have to rely on what you feel is right. Im not home to check or match up with udb which has been a differ debate in itself as it doesnt cover conditions if any. Wowhead is diluted with their method of datamining thru addon use. But if you feel your approach from the video and other references would put it more towards a possible blizz like rate then go for it.

Gultask:
See https://github.com/azerothcore/azerothcore-wotlk/pull/9379 as they had changed the drop rate for Ritual Salve. Now, I don't really know for sure but I think there was a mistake, as the table contained in the comments says 80% and in the sql it changed to 15%
Can't really be sure since I'm not on my PC right now, but I think 80% is the correct chance.

neifion-00000000:
Currently working on this drop rate, how high confidence do we want to be on the rate?

ATM I am 99% confident that its drop rate is between 65 and 87%, and 90% confident it is between 69 and 83% but I am working on getting those numbers better.

Maelthyrr:
> Currently working on this drop rate, how high confidence do we want to be on the rate?
> 
> (Updated)
> ATM I've mined enough data to know its 65-79% with 98% confidence  67-77 with 90% confidence
> 
> I'm currently 2/5ths of the amount I want to gather (500 samples) at which point I'd have a solid recommendation.

Thanks you si much 😊

data/sql/updates/pending_db_world/rev_1653810747543320600.sql
```diff
@@ -0,0 +1,2 @@
+-- Adjust drop rate of Ritual salve to 72% (based by a sniffer with 325 kills)
+UPDATE  `creature_loot_template` SET `Chance` = 72 WHERE `Entry` = 2953 AND `Item` = 6634;
```
Nefertumm, line 2:
```suggestion
UPDATE `creature_loot_template` SET `Chance` = 72 WHERE `Entry` = 2953 AND `Item` = 6634;
```

Nefertumm (changes requested):

Nefertumm (approved):

PR #11927, opened 2022-06-01
fix(DB/creature_template): Vanilla gold drop values (credit Vmangos)

ZhengPeiRu21:
## Changes Proposed:
-  Adjust the gold drop values for many vanilla creatures based on sniffs
- Because the SQL will check that existing gold drop values match before the changing, this will not affect creatures that were changed from Elite to Normal or this type of thing. It will only change creatures that have the same incorrect Vanilla gold values as Vmangos is fix.

## Issues Addressed:

- Closes unreported issue

## SOURCE:

Vmangos: https://github.com/vmangos/core/commit/37b12caeb9becf4473096b319e1d69a105e2233f

## Tests Performed:

- N/A

## How to Test the Changes:

1. Gold drop values are changed

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1654095984571268900.sql
+++ b/data/sql/updates/pending_db_world/rev_1654095984571268900.sql
@@ -1,0 +1,2789 @@
+--
+-- Update gold dropped by mobs based on classic sniffs.
+UPDATE `creature_template` SET `mingold`=3 WHERE `entry`=36 && `mingold`=13; -- Harvest Golem (sample size 30)
+UPDATE `creature_template` SET `maxgold`=24 WHERE `entry`=36 && `maxgold`=21; -- Harvest Golem (sample size 30)
+UPDATE `creature_template` SET `mingold`=1 WHERE `entry`=38 && `mingold`=2; -- Defias Thug (sample size 134)
+UPDATE `creature_template` SET `maxgold`=5 WHERE `entry`=38 && `maxgold`=7; -- Defias Thug (sample size 134)
+UPDATE `creature_template` SET `mingold`=1 WHERE `entry`=40 && `mingold`=5; -- Kobold Miner (sample size 65)
+UPDATE `creature_template` SET `mingold`=2 WHERE `entry`=46 && `mingold`=9; -- Murloc Forager (sample size 26)
+UPDATE `creature_template` SET `maxgold`=15 WHERE `entry`=46 && `maxgold`=17; -- Murloc Forager (sample size 26)
+UPDATE `creature_template` SET `mingold`=1 WHERE `entry`=80 && `mingold`=2; -- Kobold Laborer (sample size 47)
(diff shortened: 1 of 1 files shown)


T1ti:
You really shouldn't commit that, first of all gold values are very different between vanilla and Wrath, like 20/30% of the creatures had their level modified and a lot of creatures used tob e elite in vanilla
And second, it's really bad sample size(lot of creztures with udner 10 kills, the current sources collected from addons is as reliable as sniffs but with much bigger sample sizes

Just awful idea overall

ZhengPeiRu21:
> You really shouldn't commit that, first of all gold values are very different between vanilla and Wrath, like 20/30% of the creatures had their level modified and a lot of creatures used to be elite in vanilla And second, it's really bad sample size(lot of creatures with under 10 kills, the existing sources collected from addons/wowhead is as reliable as sniffs but with much bigger sample sizes. Aaand the gold drop values could have been changed between vanilla and wrath, who knows
> 
> Just awful idea overall, always be careful when using Vanilla values especially if it's related to creatures levels/rank

Sorry for not explain this very clearly in the original description. Because the `WHERE` clause is select both the creature ID and the existing min or max gold, this SQL will only change values where we are currently using the same values that Vmangos was using before implement this change. Any creature that was having the gold value changed after Vanilla would not have a matching value, so it will not be affected (unless Vmangos before had the correct WotLK gold values and is correcting them to Vanilla value, in which case this is true would be wrong.)

However, on the other hand, I am understanding that Vmangos could be a wrong source for us here, so if we are more confident in our existing values, then this change is not necessary - I do not know where our original data is come from or how accurate it is for gold drop values; I do only know that our loot drop tables have many problems and thought it may be same source, so this may provide step in the right direction.

acidmanifesto:
@ZhengPeiRu21 I dropped the PR link into the staff room for discussion. Hopefully some feedback is provided

acidmanifesto:
> > At the moment, where it stands. We need to know if these values were changed at any point from vanilla to wotlk. Are you able to provide any supporting references for this change? Although i am 100% confident it is correct for vanilla, we are uncertain if there was any adjustments when wotlk was retail.
> 
> Sorry, I am not having any references other than the Vmangos source. I have reviewed patch notes and never find a case where it is mention that gold values were changed, and I would be surprising to see Blizzard taking time to update low-level gold values without a reason, but it is always possible that there is some change that no one has before noticed. Maybe this PR can be a reference guide that can be confirmed in WotLK Classic after it is launch.

I would like to keep this open as well as the other pr. So far all i can confirm is packet changes and sindragosa roar is muted on login screen in classic wotlk. Lol. Well let u know when we can login.

ZhengPeiRu21:
With WotLK Classic now available for confirmation, we should be able to check that these are correct values.

acidmanifesto:
> With WotLK Classic now available for confirmation, we should be able to check that these are correct values.

Process in confirming information. Please give me sometime. this is alot to sort thru.

acidmanifesto (approved):

acidmanifesto:
Was a massive audit. All checked out with other core developers and their sniffer members. Thank you for your PR

PR #11965, opened 2022-06-04
feat(Core/Command): Use timestring with all commands

### LINKED ISSUE #11920: Durations for commands `mute` and `deserter [bg/instance] add` should be the same format as `ban`

Make all the commands use the same duration format that the `ban` command uses such as  `4d20h6m9s` for 4 days, 20 hours, 6 minutes and 9 seconds.

If you wish to mute Angrygnome for a day, which is 1440 minutes then you will have to type `.mute Angrygnome 1440 Being a gnome`. Instead, I would like to only type `.mute Angrygnome 1d Being a gnome` because it is a lot easier to read.

There is also the `deserter bg add` and `deserter instance add`. The duration is in seconds. If you wish to grant a gnome a 13 days battleground deserter then type the following command when having them targetted: `.deserter bg add 1123200`
This number is ridiculous and unreadable unlike `.deserter bg add 13d`.

Make all the commands use the same duration format that the `ban` command uses such as  `4d20h6m9s` for 4 days, 20 hours, 6 minutes and 9 seconds.

### Additional context

Looking on the [GM commands wiki](https://www.azerothcore.org/wiki/gm-commands), there are these two commands that might need a similar treatment.

`.npc add move #creature_guid [#waittime]`
`.npc set spawntime #time`

The implementation of the feature may require an update on the [GM command wiki](https://www.azerothcore.org/wiki/gm-commands).

For anything that might **needs** to work with the old durations, if there are letters in the duration, it is the `4d20h6m9s` duration format otherwise, it would be in seconds for the `deserter` command and minutes for the `mute` command.

Gozzim:
## Changes Proposed:
-  TODO
-  

## Issues Addressed:

- Closes #11920

## Tests Performed:

- TODO
- 

## How to Test the Changes:

1. TODO
2.
3.

- [x] `.mute`
- [ ] `.deserter`
    - [ ] add
    - [ ] remove all
- [ ] `.server`
    - [ ] shutdown
    - [ ] restart
    - [ ] idleshutdown
    - [ ] idlerestart
- [ ] `.npc set spawntime`
- [ ] `.bf timer`

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1654372051309044400.sql
+++ b/data/sql/updates/pending_db_world/rev_1654372051309044400.sql
@@ -1,0 +1,11 @@
+--
+DELETE FROM `acore_string` WHERE `entry` IN (283,300,301,11003);
+INSERT INTO `acore_string` (`entry`, `content_default`, locale_deDE) VALUES
+(283, "%s has disabled %s's chat for %s, effective at the player's next login. Reason: %s.", "Ihr habt dem Spieler %s das chatten für %s gesperrt, beginnend mit dem nächsten Login des Spielers. Grund: %s."),
+(300, "Your chat has been disabled for %s. By: %s, Reason: %s.", "Euer Chat wurde für %s abgeschaltet. Von: %s, Grund: %s"),
+(301, "%s has disabled %s's chat for %s. Reason: %s.", "Ihr hab den Chat von %s für %s abgeschaltet. Von: %s, Grund: %s"),
+(11003, "Server: %s has muted %s for %s, reason: %s", NULL);
+
+DELETE FROM `command` WHERE `name` IN ("mute");
+INSERT INTO `command` (`name`, `security`, `help`) VALUES
+("mute", 2, "Syntax: .mute [$playerName] $mutetime [$reason]\nDisible chat messaging for any character from account of character $playerName (or currently selected) at $mutetime time. Player can be offline.\n$mutetime: use a timestring like \"1d15h33s\".");
--- a/src/server/scripts/Commands/cs_misc.cpp
+++ b/src/server/scripts/Commands/cs_misc.cpp
@@ -2374,9 +2374,14 @@ public:
     }
 
     // mute player for some times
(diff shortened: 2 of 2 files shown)


src/server/scripts/Commands/cs_deserter.cpp
```diff
@@ -92,7 +93,7 @@ class deserter_commandscript : public CommandScript
     * .deserter bg add 1h30m
     * @endcode
     */
-    static bool HandleDeserterAdd(ChatHandler* handler, Optional<PlayerIdentifier> player, std::string time, bool isInstance)
+    static bool HandleDeserterAdd(ChatHandler* handler, Optional<PlayerIdentifier> player, Optional<std::string> time, bool isInstance)
```
Winfidonarleyan, line 96: How do you imagine it?

src/server/scripts/Commands/cs_deserter.cpp
```diff
@@ -106,17 +107,17 @@ class deserter_commandscript : public CommandScript
...
         }
 
         Player* target = player->GetConnectedPlayer();
-        int32 duration = TimeStringToSecs(time);
+        int32 duration = TimeStringToSecs(time.value());
 
         if (duration == 0)
         {
-            duration = atoi(time.c_str());
+            duration = atoi(time.value().c_str());
```
Winfidonarleyan, line 120: don't use `atoi` we have `Acore::StringTo`

Gozzim: Oh didn't know that. Thanks :)
I'll change it later

src/server/scripts/Commands/cs_deserter.cpp
Gozzim: What do you mean exactly?
The idea is that it adds the default length of the debuff if no parameter for the time is given.
So it checks if time exists, if it does not, it defaults to 15min for BG and 30min for RDF.
If one is given, it first assumes it to be an Acore TimeString and tries to convert it into seconds. If that fails, it next assumes it to be a pure number and tries to convert it. If that's still 0, false is returned.

Maelthyrr:
Tested and work fine.

Gozzim:
Closing due to the mistake and reopening

PR #11972, opened 2022-06-06
Uldaman item 7666 Alliance only drop

CyberHanz:
This item drops in Uldaman, should only drop for Alliance

## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1. Run uldaman as horde and kill stuff, item should not drop 
2. Run Uldaman as Alliance, kill stuff, item should drop sometime
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/Uldaman7666allyonly
+++ b/data/sql/updates/pending_db_world/Uldaman7666allyonly
@@ -1,0 +1,1 @@
+UPDATE item_template SET flagsextra ="2" WHERE NAME = "shattered necklace"; #


Gultask:
What about this horde quest? Would this break it or is there a condition for it to drop while on the quest?

https://classic.wowhead.com/quest=2283/necklace-recovery

CyberHanz:
> What about this horde quest? Would this break it or is there a condition for it to drop while on the quest?
> 
> https://wowgaming.altervista.org/aowow/?quest=2283

Good one, with my fix it wouldn't drop at all.. So it should only drop when you are on the particular quests as a Horde.
Without quest, no drop I think?

Gultask:
Horde getting this item outside the quest is blizz-like:
>I tossed this item after I got it, right-clicked and got the "This quest is not available to your race" message.
If you're Horde, there's a quest to obtain this in Orgrimmar, so don't throw it away. Just don't bother right-clicking on it ([source](https://classic.wowhead.com/item=7666/shattered-necklace#comments:id=3025568))

There is another Shattered Necklace item (ID: 7716) which I believe could be modified to be used in the Horde quest but that would be a custom change.

CyberHanz:
If it's blizzlike, how annoying, we should keep it that way.
Thanks for pointing it out, I missed that part.
I'll close the ticket.

PR #11977, opened 2022-06-07
Fix(DB/Creature): Zul'Gurub boss damage

temperrr:
## Changes Proposed:
-  Reduce damage from ZG bosses as it was way too high. (normal values were all 35 (higher than BWL)

## Issues Addressed:

- None
- 
## SOURCE:

Cmangos, compared with Warcraft logs to assure validity. 

## Tests Performed:

- Tested all bosses and compared damage.

## How to Test the Changes:

1. Go to ZG
2. Kill all the bosses and get hit by them ;)  

- None


ratkosrb:
@T1ti Some info about Hakkar here https://github.com/cmangos/issues/issues/2189

The values in the book are with all stacks of the buff (all mini bosses alive).

Gultask:
Also might need to tweak damages of Hazza'rah's Nightmare Illusions (15163) They are supposed to deal massive damage but currently deal very little. Their damage modifier in Keira is 1.

temperrr:
I'll look into that

temperrr:
> Also might need to tweak damages of Hazza'rah's Nightmare Illusions (15163) They are supposed to deal massive damage but currently deal very little. Their damage modifier in Keira is 1.

From cmangos its the same value, where did you find it should do massive damage?

ratkosrb:
In my classic sniff parses, creature 15163 is seen doing damage in the 4785 - 5992 range. This is from combat log, not mind control, so it's not the exact damage range, but it gives you an idea what the damage multiplier should be.

Gultask:
yeah, what ratksrb said. ^^
Their gimmick is that they have like 500-700 HP and are slow to start attacking, but they deal immense amounts of damage since there's also 3 of them at once.
[In this WotLK-era video](https://youtu.be/nN4-ecE0FnE?t=124) they deal 1000+ damage, and around double of the boss' damage

temperrr:
I'll see if I can find a good source for it as I cant start guessing :3

PR #11958, opened 2022-06-04
feat(Core/Command): Deserter Command send feedback

Gozzim:
## Changes Proposed:
-  Using the `.deserter` command will now always give feedback
    - Player or amount of players effected
    - Time factor if relevant
    - The kind of deserter processed

## Issues Addressed:

- Closes #11936

## Tests Performed:

- Added and removed BG Deserter from PlayerChar/Console to/from an online player
- Added BG Deserter from PlayerChar/Console to multiple online players and removed BG Deserter from all players 
- Added and removed BG Deserter from PlayerChar/Console to/from an offline player
- Added BG Deserter from PlayerChar/Console to multiple offline players and removed BG Deserter from all players 
- Added BG Deserter from PlayerChar/Console to a mix of online and offline players and removed BG Deserter from all players 

## How to Test the Changes:

1. Add a Deserter debuff to an offline player and remove it (before and after logging out)
2. Add a Deserter debuff to an online player and remove it (before and after logging in)
3. Add a Deserter debuff to a mix of online and offline players and remove it from all players

## Known Issues and TODO List:

- [ ] Removing deserter from all players will not display the correct value of the amount of players if a player hasn't been saved to the DB after gaining the Debuff
- [x] #11955

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #12002, opened 2022-06-10
feat(DB): MySQL PreparedStatement Affected Row Count

Gozzim:
**This PR is not just meant as a change to the MySQL Query function for Premared Statements but mostly as a base for an open discussion about the change.**
@Nyeriah and I talked about this a bit but a change to functions associated with the Database should not be made unless it is certain that it will not cause any issues.

## Changes Proposed:
-  Correctly count the affected Rows on `UPDATE`, `DELETE` and `INSERT` queries using Prepared Statements

## Description

In #11958 I am using a `SELECT` Query to determine the amount of deserter debuffs with certain constraints in the Database, just to have feedback for the number of affected rows by the following `DELETE` Query.
I have tried to use a `DELETE` query only with the plain SQL query function as well as with the Prepared Statement query function. Sadly neither give feedback about the amount of affected rows.
As @Nyeriah pointed out correctly, this is highly inefficient, especially with a table associated with characters just for some feedback on the removed deserter debuffs.

So after looking into the matter a bit deeper, it seems like I have found the issue.

### Query Function with plain SQL Queries
The whole function can be found below
https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L329-L368
which is called by
https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L313-L327

returning false will lead to the function calling it to return a `nullptr`

The number of affected rows is gained by using `mysql_affected_rows`
<details>
<summary>Code for Affected Rows</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L351-L353
</details>

Since a `DELETE` query, for instance, will not have a result, it will always return a `nullptr`, meaning the whole ResultSet, including the actual number of affected rows (which might be > 0) will not exist.
<details>
<summary>Code of Result Check</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L356-L357
</details>

Because basically all uses of this function (including those in modules) will check whether the result exists, changing this would only cause chaos though.

Interestingly enough the Function for Prepared Statements works a bit differently.

### Query Function with Prepared Statements
The whole function can be found below
https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L260-L311
which is called by
https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L535-L551

Here the issue is a completely different one.
We try to get the values for the affected rows by calling `mysql_stmt_num_rows`
<details>
<summary>Code for Affected Rows</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/448f9494c5ead8761f7f1effec7faef90ddeb36f/src/server/database/Database/MySQLConnection.cpp#L306-L308
</details>

The problem is that the affected rows will always be `0` because `mysql_stmt_num_rows` only returns a value when`mysql_stmt_store_result` is used to buffer the entire result set in the statement handler (Source: [MySQLDocs/mysql-stmt-num-rows](https://dev.mysql.com/doc/c-api/8.0/en/mysql-stmt-num-rows.html)). Though I'm not entirely sure if I didn't miss anything here.

For that reason changing `mysql_stmt_num_rows` to `mysql_stmt_affected_rows` will actually return the correct value for a `DELETE` query.
But now there is an issue with the number of affected rows in `SELECT` queries, which returns a value of `-1`. This results in finally returning `18446744073709551615` as the number of affected rows as it is a `uint64`.
Of course, we can just make a check whether or not the value is `rowCount < 0` and return 0 if so.

I can't find an issue with this though, to be honest, since I couldn't find any function actually using `GetRowCount()` with Prepared Statements in the code.
<details>
<summary>Code for GetRowCount</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/de13bf426e162ee10cbd5470cec74122d1d4afa0/src/server/database/Database/QueryResult.h#L81
</details>

Also I found it a bit weird that a few checks for the affected row count will return a nullptr if failed even though it is always 0 if i'm not mistaken.
<details>
<summary>Check 1</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/336166c83e209d77c339ffeaef96f9fe77ed19e8/src/server/database/Database/DatabaseWorkerPool.cpp#L203-L220
</details>
<details>
<summary>Check 2</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/5969df4e303121cca77166df28aa0e9ff2f17e9f/src/server/database/Database/QueryHolder.cpp#L46-L57
</details>

## Tests Performed:

- Ran and debugged various SQL queries using the `Query(PreparedStatementBase* stmt)` function

## How to Test the Changes:

1. Run the server
2. Extensively test actions which cause a SQL query with a Prepared Statement to be called (`SELECT`, `UPDATE`, `DELETE` and `INSERT`)

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -304,7 +304,7 @@ bool MySQLConnection::_Query(PreparedStatementBase* stmt, MySQLPreparedStatement
     m_mStmt->ClearParameters();
 
     *pResult = reinterpret_cast<MySQLResult*>(mysql_stmt_result_metadata(msql_STMT));
-    *pRowCount = mysql_stmt_num_rows(msql_STMT);
+    *pRowCount = mysql_stmt_affected_rows(msql_STMT);
     *pFieldCount = mysql_stmt_field_count(msql_STMT);
 
     return true;


Nyeriah:
I need some opinions on this @Kitzunu @Winfidonarleyan

Kitzunu:
Reading MySQL Docs this seems legit. As long as nothing breaks im good with it

Gozzim:
Any further opinions/objections @Winfidonarleyan @Yehonal ?

Maelthyrr:
@FrancescoBorzi @Helias

Winfidonarleyan:
for test

https://github.com/azerothcore/azerothcore-wotlk/blob/02fa20b6380b991634198de9a295abac11152d60/src/server/database/Database/QueryResult.h#L42

Gozzim:
I can also add a change or link a branch for testing if you guys want where I intend to use this.
To be exact right now the `.deserter bg remove all` command uses a `select` before the `delete` query to get the number of lines, which is highly inefficient. So this is a usecase.

acidmanifesto:
Marking it as sensitive wip until we get one of the requested reviewers to look at this.

acidmanifesto:
i'm trying to grasp the idea here. we dont want it to return with the results of all the set rows but instead we want it to return the results of affected rows?

Yehonal:
Shouldn't we keep `mysql_stmt_num_rows` for SELECT and use `mysql_affected_rows` for the others?

I think is pretty safe, if we do not use the rowCount non-SELECT queries, it should not impact any logic.

Another way would be to allow to switch between the 2 via a parameter maybe

Gozzim:
Sorry, it's been a while for me as well and I am quite busy even though I'd love to finally get back into ACore :/
So I need to read into it more carefully as well again but from what I can remember right now (I might correct myself in the coming weeks when i can finally find the time again):

> i'm trying to grasp the idea here. we dont want it to return with the results of all the set rows but instead we want it to return the results of affected rows?

@acidmanifesto Basically yes. The idea is to actually return the number of affected rows for `UPDATE`, `DELETE` and `INSERT` queries as they will never return a result with `mysql_stmt_num_rows`. (I just noticed that my references seem to point at the code of the PR and not the current code). The reason for this is further explained above.

https://github.com/azerothcore/azerothcore-wotlk/blob/437d93926f64819c764a935cfb59540ef903ac25/src/server/database/Database/MySQLConnection.cpp#L307



> Shouldn't we keep `mysql_stmt_num_rows` for SELECT and use `mysql_affected_rows` for the others?
> 
> I think is pretty safe, if we do not use the rowCount non-SELECT queries, it should not impact any logic.
> 
> Another way would be to allow to switch between the 2 via a parameter maybe

@Yehonal 
I have not found any instance where this causes issues but we it's true that it would probably be the safer method to check what kind of request it is and determine the use of `mysql_affected_rows` or `mysql_stmt_num_rows` from that.
A parameter would in my opinion not have any benefits. The returns of non-SELECT queries would still be broken by default (considering the default is `mysql_stmt_num_rows`) and barely anyone is going to be aware of the parameter or the actual difference between the two. I remember having to do quite alot of research to fully understand how all the functions, libraries and parameters work together and why it is the way it is. And I still hadn't understood every side-effect quite yet.

PR #12024, opened 2022-06-12
Fix(DB/Quest): "Burning Blood", "Iron Coal" and "Sunscorched Shells" …

temperrr:
…should be available after "Fire Hardened Mail"

## Changes Proposed:
-  Set prerequisite quest for "Burning Blood", "Iron Coal" and "Sunscorched Shells" to  "Fire Hardened Mail"

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3372
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12007

## Tests Performed:

- not tested

## How to Test the Changes:

1. `.quest add 1701` `.quest complete 1701` `.go q ender 1701`
2. turn in the quest, depending on your race one crumb quest should have unlocked.
3. `.go q starter 1705` `.go q starter 1708` `.go q starter 1710`
4. check if the quests Burning Blood, Iron Coal and Sunscorched Shells are available to be taken.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

Issue #12007, opened 2022-06-11
Warrior Quest Chain for lvl20+ Gear Cannot Be Completed


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3372

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [X] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

Completing Fire Hardened Mail (1701) does not unlock quests Burning Blood (1705) Iron Coal (1708) and Sunscorched Shells (1710)

It does unlock the race-exclusive crumb quests [Grimand Elmore](https://wowgaming.altervista.org/aowow/?quest=1700), [Klockmort Spannerspan](https://wowgaming.altervista.org/aowow/?quest=1704) and [Mathiel](https://wowgaming.altervista.org/aowow/?quest=1703) to their respective races.

### Expected Blizzlike Behaviour

After turning in Fire Hardened Mail, quests 1705, 1708 and 1710 should all be unlocked and available by going to their questgivers, ignoring the crumb quests if they aren't available.

### Source

Sniffs and Classic research from @neifion-00000000 

Also thanks to Tebec for pointing this out.
>The race tags on the above is just which NPC Furen will send you to after he gives you your armor. The other two NPC's will give you their quest if you locate them manually, so you can get an (almost) full set of Fire Hardened if you are willing to run to them and do their errands.
>
>About Draenei, it seems they don't get a note from Furen to give to one of the other smiths. But the quests should get unlocked anyway. The Human smith is around the corner from Furen so it is a quick check for a curious Draenei.:) ([source](https://www.wowhead.com/quest=1701/fire-hardened-mail#comments:id=152454))

### Steps to reproduce the problem

1. `.quest add 1701` `.quest complete 1701` `.go q ender 1701`
2. turn in the quest, depending on your race one crumb quest should have unlocked.
3. `.go q starter 1705` `.go q starter 1708` `.go q starter 1710`
4. check if the quests Burning Blood, Iron Coal and Sunscorched Shells are available to be taken.

### Extra Notes

From @neifion-00000000:

![image](https://user-images.githubusercontent.com/91289495/173171295-d6a3cff0-415b-407f-b133-f0abb1f6927d.png)
[...]

PR #12060, opened 2022-06-14
(DB/loot): Addressed drop style to be proper for Fire Hardened Mail

neifion-00000000:
Quest drops properly drop individual stacks in loot window.

What's left of #12007 that isn't fixed by #12024 (Extra Notes section)

## Changes Proposed:
-  Makes the loot appear as it should (unstacked drops).  

## Issues Addressed:

- Closes 

## SOURCE:
TBC Classic / packet sniff

## Tests Performed:

- Tested as explained below, works as expected.

## How to Test the Changes:

1.  .quest add 1701
2.  .tele to wetlands and goto spider cave, get a few drops
3.  .tele charred (vale) and kill a few chimaeras 

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1655244193594772800.sql
+++ b/data/sql/updates/pending_db_world/rev_1655244193594772800.sql
@@ -1,0 +1,16 @@
+--
+DELETE FROM `creature_loot_template` WHERE item=6838;
+INSERT INTO `creature_loot_template` (`Entry`, `Item`, `Reference`, `Chance`, `QuestRequired`, `LootMode`, `GroupId`, `MinCount`, `MaxCount`, `Comment`) VALUES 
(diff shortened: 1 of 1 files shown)


temperrr:
Curious question, what did u actually change to the loot that now causes it to stack instead of spread?

neifion-00000000:
> Curious question, what did u actually change to the loot that now causes it to stack instead of spread?

I had to use groupids because otherwise you get key error for having the same entry / lootid / groupid

each stack has its own drop chance, with one chance being 100%

for the 1-2s, its 100% / 50%, for the 1-3s its 100%/34%/34%, so in effect this did nothing (except that it was incorrectly not 100% already for this quest)

acidmanifesto (approved):

acidmanifesto:
Thank you for your PR.

PR #12071, opened 2022-06-17
fix(DB/gameobject): Overhaul of Gameobject 106318 in Mulgore

neifion-00000000:
Adds a lot of previously missed chests in Mulgore.
Pools them on sniffed spawns so they are not always the same.

## Changes Proposed:
-  Removes existing chests in Mulgore, for organizational purposes
-  Insert all Mulgore chests, then pools them where appropriate.

## Issues Addressed:

- Makes them correct and makes them complete.

## SOURCE:
Sniff variables and hard work looting them on blizz servers long enough to understand them.

## Tests Performed:

- After installing, to test, I set the respawn time to 30 seconds.    UPDATE `gameobject` set spawntimesecs='30' WHERE id=106318;
- That makes it easy to check.  Since I found all these it was easy for me, but I also created a map.  The numbers on this map will match the GUID offset I've written
image

## How to Test the Changes:

1.  See above

- Make sure GM is off when testing this or it will not work.


acidmanifesto (approved):

acidmanifesto:
Thank you for the PR

Issue #12107, opened 2022-06-19
Characters below level 10 should not be allowed to join a raid group


temperrr:
### Current Behaviour

Any char can join a raid group

### Expected Blizzlike Behaviour

a char below level 10 cannot join a raid group

### Source

https://wowpedia.fandom.com/wiki/Patch_3.3.0
![image](https://user-images.githubusercontent.com/20357406/174472458-a91b73e2-4773-4b69-b952-602c39fa1a40.png)

### Steps to reproduce the problem

1. make raid group with 2  level 10+
2. try inviting a level 10-

### AC rev. hash/commit

82cfd1b8d4147f09564e692f13effc4142178d72

### Operating system

ubuntu x64

Kitzunu:
This Opcode handles raid convertion

https://github.com/azerothcore/azerothcore-wotlk/blob/82cfd1b8d4147f09564e692f13effc4142178d72/src/server/game/Handlers/GroupHandler.cpp#L612-L631

Kitzunu:
https://github.com/azerothcore/azerothcore-wotlk/pull/12110 @temperrr

PR #12110, opened 2022-06-19
fix(Player/Groups): Level restriction on allowing raid groups

Kitzunu:
## Changes Proposed:
-  And it's configurable, pog. 
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12107

## Tests Performed:

- tested ingame
- 

## How to Test the Changes:

1. Try to create a raid group with a member below level 10

### DIFF
--- a/src/server/game/Groups/Group.cpp
+++ b/src/server/game/Groups/Group.cpp
@@ -18,6 +18,7 @@
 #include "Group.h"
 #include "Battleground.h"
 #include "BattlegroundMgr.h"
+#include "Config.h"
 #include "DatabaseEnv.h"
 #include "GameTime.h"
 #include "GroupMgr.h"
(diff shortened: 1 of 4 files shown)


temperrr:
Will test tomorrow

temperrr:
Tested, works as described
config works as described.

src/server/game/Groups/Group.cpp
```diff
@@ -278,6 +279,16 @@ void Group::ConvertToLFG(bool restricted /*= true*/)
     SendUpdate();
 }
 
+bool Group::CheckLevelForRaid()
+{
+    for (member_citerator citr = m_memberSlots.begin(); citr != m_memberSlots.end(); ++citr)
+        if (Player* player = ObjectAccessor::FindPlayer(citr->guid))
```
Nyeriah, line 285: FindConnectedPlayer is probably better. Iirc findplayer only checks the current map storage and only accounts for players currently in world (so no loading screen while teleporting from one place to another, for example)

Kitzunu: Used the same as Group::ConvertToRaid uses. But I can redo

Nyeriah: probably fine then

Nyeriah (approved):

PR #12128, opened 2022-06-20
feat(Core/Formations): Implement GROUP_AI_FLAG_DONT_RESPAWN_LEADER_ON…

Nyeriah:
…_EVADE flag

## Changes Proposed:
-  Add a flag so the leader doesn't respawn on evade, so it can be used with bosses (see thekal)
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1. Pull thekal
2. kill cat 1, kill thekal, reset
3. notice thekal doesn't respawn, while both cats do
4. kill both cats & reset, notice both cats respawn

- [ ]
- [ ]


Nyeriah:
@UltraNix opinions? I need this to be able to use the flag with bosses, otherwise the bosses respawn once I reset the adds (e.g kill the boss, leave adds alive, reset, boss respawns).

UltraNix (approved):

PR #12061, opened 2022-06-15
cute and fluffy AQ20

Si1ker:
image

### Changes:
**==== Kurinnaxx ====**
- Cleanup/refactor
- Migrate trash spell handling from script to db
- Guides state that cleave (WideSlash) applies the 'Mortal Wound' debuff but videos demonstrate that it does not apply them. Those are separate events.
source: https://youtu.be/EgazIWGM0n0?t=30 AND https://youtu.be/idACansl4-0?t=3 / https://youtu.be/idACansl4-0?t=14

**==== General Rajaxx ====**
- Cleanup/refactor

**==== Ossirian ====**
- Cleanup/refactor
- Add Immunity to Taunt

**==== Buru ====**
- Cleanup/refactor
- Schedule EVENT_CREEPING_PLAGUE
- Adjust timer for EVENT_CREEPING_PLAGUE
- Add Immunity to Taunt
- Let "Creeping Plague" tick 3 times before stacking up
- Remove EVENT_DISMEMBER on Phase2  |  https://youtu.be/D0nktm78pII?t=375
- Make all visual spell effects triggered to skip unit egg death
- Remove all 'Creeping Plague' stacks on Buru death
- Make the remaining eggs hatch at Phase 2 (won't respawn)  |  https://youtu.be/TyRheH9Cys0?t=382
- Disable experience gain on "Buru Egg" and "Hive'Zara Hatchling"
- Disable visual effects on eggs at "Phase2 forced hatch"
- Despawn "Hive'Zara Hatchling" after 10sec if out of combat

**==== Moam ====**
- Cleanup/refactor

**==== Ayamiss ====**
- Cleanup/refactor
- Add Immunity to Taunt

### Known issues:

**Kurinnaxx** 
* Sand traps require scripting.
* Timers require adjustment.

**General Rajaxx**
* Waves and dialogues are gonna need scripting.
* Timers require adjustment.
* Lt. General should be immune to Rajaxx's knockback attack

**Ossirian**
* Sand vortex triggers visible.
* Crystal object init is delayed (intended?).
* Timers require adjustment.

**Buru**
* Timers require adjustment.
* Egg is targeting the player.
* Eggs repositions if target within 0-3 yards.
* Buru missing damage reduction in Phase 1  |  [idea](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/Northrend/Ulduar/Ulduar/boss_hodir.cpp#L335)
[...]

PR #12138, opened 2022-06-21
fix(scripts/RuinsOfAhnQiraj): Improve Ossirian

Si1ker:
Originated from #12061

## Changes Proposed:
- Cleanup/refactor
- Immune to Taunt
- Remove SummonPlayerMechanic (already in core)
- Make Sand Vortex not selectable
- Change the amount of Sand Vortex NPCs

## Issues Addressed:

- Updates #12473

* Crystal object init is delayed (intended?). | Reported at #12472


Gultask:
I took a look in retail and it seems that the Summon Player mechanic isn't based on distance but rather mmaps. The boss was rather fast anyway so I couldn't kite it even when mounted, and the arena is complex enough that when the knockback happens the player might get stuck on the pillar debris, which does trigger the summon spell.

Nyeriah (approved):

Issue #12206, opened 2022-06-26
[ZG] Jin'do the Hexxer - Teleport should teleport players into the air instead of at ground level.


Gultask:
### Current Behaviour

The boss teleports players to ground level.

### Expected Blizzlike Behaviour

The boss should teleport players into the air, but not high enough to cause fall damage.

### Source

https://youtu.be/rTFP5VYf3zw?t=126
https://youtu.be/p_oWqxRhJCI?t=213 shows, although only visually, the exact height of which the player is teleported to.

### Steps to reproduce the problem

`.go c id 11380` with at least 2 or 3 characters to avoid resets from Hex and the totems.
wait for a teleport to see if characters are teleported into the air on at ground level

### Extra Notes

May need a sniff to determine exact distance?

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/4cbe846d2221dc861598c473394e810942cde210

### Operating system

Windows 10

Issue #12205, opened 2022-06-26
[ZG] Jin'do the Hexxer - Sacrificed Trolls spawn behaviour is not blizzlike.


Gultask:
### Current Behaviour

1. A group of Sacrificed Trolls spawn when a player is teleported.
2. The respawn timer of the present Sacrificed Trolls is wrong.
3. Sacrificed Trolls' bodies take too long to despawn.
4. There is a separate respawn timer for every Sacrificed Troll.

### Expected Blizzlike Behaviour

1. Sacrificed Trolls should not be spawned when a player is teleported.
2. Sacrificed Trolls should have a respawn timer of ~20s
3. Their corpses should despawn after ~5s
4. All Sacrificed Trolls should respawn together, regardless of when they were killed.

### Source

https://youtu.be/Lz8daXkGGYY?t=144
https://youtu.be/udQAHdjHi4A?t=173 https://youtu.be/udQAHdjHi4A?t=212 https://youtu.be/udQAHdjHi4A?t=258
https://youtu.be/I3oveygDBY4?t=90
https://youtu.be/I3oveygDBY4?t=580 corpse despawn
https://youtu.be/Zg-3BgmCLmA?t=119 They respawn together

### Steps to reproduce the problem

1. `.go c id 11380` with at least 2 or 3 characters to avoid resets from Hex and the totems.
2. kill the sacrificed trolls immediately after they spawn and see if a teleport will spawn another set of them.
3. kill the sacrificed trolls and count down their respawn timer and how long it takes for their corpses to disappear.
4. kill the sacrificed trolls one at a time, and see if all the dead ones respawn together, regardless of in which order they were killed

### Extra Notes

Sacrificed Troll (14826)

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/4cbe846d2221dc861598c473394e810942cde210

### Operating system

Windows 10

Issue #12204, opened 2022-06-26
[ZG] Jin'do the Hexxer - Brain Wash Totem is missing beam visuals.


Gultask:
### Current Behaviour

It's missing visuals

### Expected Blizzlike Behaviour

It should have a green beam directed to the mind controlled player.

### Source

https://youtu.be/rTFP5VYf3zw?t=15

### Steps to reproduce the problem

`.go c id 11380` with at least 2 or 3 characters to avoid resets from Hex and the totem.
when the Brain Wash Totem is cast, see if it has a green beam directed towards the mind controlled character

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/4cbe846d2221dc861598c473394e810942cde210

### Operating system

Windows 10

Issue #12188, opened 2022-06-26
[ZG] Jin'do - Some desilusion do nothing


Maelthyrr:
### Current Behaviour

There are few desilusion of jhin'do that do nothing and don't attack players.

### Expected Blizzlike Behaviour

Should always attack players.

### Steps to reproduce the problem

1. Engage Jhin'do
2. Get the debuff to see desilusion
3. Look if some desilusion stay at their position and don't attack

### AC rev. hash/commit

76e9e19

### Operating system

Windows 11

PR #12233, opened 2022-06-29
fix(Core/ZulGurub): Jin'do the Hexxer rewrite

Nefertumm:
## Changes Proposed:
-  A lot of code improvements.
-  Brain wash totem should use the correct spell now.
-  Corrected respawn timer and corpse despawn for sacrificed trolls, and also added to a formation.
-  Use the proper spell for teleporting players and summoning creatures (Shades)
-  Now shades should properly engage after summoned.
- #12324 
- #12330 

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12188
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12204
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12205
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12206

## SOURCE:

Run and walk speed and spell ids from CMangos.

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. .go c id 11380 with 2 or 3 characters to avoid combat reset.
2. Test the issues above.

- [ ] There is some odd behaviour if SAI calls a force despawn and after calls a respawn. This is making the creature to despawn anyway, despite being respawned. (This sounds odd and might be fixed on a different PR)

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1656469500232857400.sql
+++ b/data/sql/updates/pending_db_world/rev_1656469500232857400.sql
@@ -1,0 +1,34 @@
+--
+DELETE FROM `spell_target_position` WHERE `id` = 24466;
+INSERT INTO `spell_target_position` (`ID`, `EffectIndex`, `MapID`, `PositionX`, `PositionY`, `PositionZ`, `Orientation`) VALUES
+(24466, 0, 309, -11582.9, -1251.15, 90, 5.04179);
+
+UPDATE `creature` SET `spawntimesecs` = 20 WHERE `id1` = 14826;
+
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 14826;
+
+DELETE FROM `smart_scripts` WHERE (`entryorguid` = 14826) AND (`source_type` = 0) AND (`id` IN (0, 1));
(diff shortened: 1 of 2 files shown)


Gultask:
Some Shades were turning idle, other than that everything works except for that issue in the TODO list.

I think I found what's making some Shades idle:
They were attacking one character but as soon as they got Charmed from Brain Wash, the Shades lost the target and stopped moving or doing anything at all.

They were all attacking the character 'Two', but once they got Brain Wash'd they became idle.

WoWScrnShot_063022_194514

Other than that there is another minor thing that might be interesting. When the Brain Wash ends but the totem is not destroyed, it will stay there as a prop without doing anything.
And if I Ice Block I can escape the Brain Wash, and that does the same to the totem. It renders it 'spent' and it won't cast Brain Wash on any other target.

However, I had found a video some time ago that showed that Brain Wash Totem would choose another target if the charm had broken. I'll try to look for it again.

IntelligentQuantum (approved):

Nefertumm:
@Gultask I've updated the shade reset stuff, about the brain wash totem, I won't address that with this PR probably for now since we are short on time :)

Gultask:
- [ ] I think I missed this before but the charmed players are attacking the Sacrificed Trolls and the neighbouring critters before the players. I think that's a faction issue? Sacrificed Troll (14826) is faction 16, Jin'do the Hexxer (11380) is faction 28.

Another thing: about pr 12330 (the dance on evade) I cannot attack Jin'do while he's dancing. However, Brain Wash totems are also not working if I'm running solo. So that means that both problems cancel each other out, funnily enough.

So besides the Sacrificed Troll friendly fire it's all functional 👍

Nefertumm:
@Gultask I would greatly appreciate if you could open a new issue for what's missing right now! I'm merging since it improves a lot of things right now.

Issue #12171, opened 2022-06-25
Greater Wind Stones and Abyssal Lords are broken.


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3675

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [X] 60

### Current Behaviour

We summ the boss like normal, but once the combat phase started, the boss instantly despawns, we did this 3 times, and the 3 times the same thing happend, we lost the mats for the summs aswell.

It was a team of ally and horde, and we did 2 summs with ally, 1 with horde, same result.
Also it was 2 fire lords and 1 wind lord, so its not certain lord lf the 4.

### Expected Blizzlike Behaviour

For the boss to not despawn unless we wipe.

### Steps to reproduce the problem

1. `.additem set 492`
2. `.additem 20422`
3. `.additem 20451`
4. `.go xyz -6751 1649 8 1 5.5`
5. Use Greater Wind Stone
6. see if boss despawns immediately after it's emote/rp event

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/6983903a89c556c0ca1239101cffc38098d6b081

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[...]

PR #12247, opened 2022-06-30
fix(DB/Scripts): Greater Windstone Bosses

ZhengPeiRu21:
## Changes Proposed:
-  Change Greater Windstone Bosses to begin combat after 5 seconds instead of 10 seconds, because they are set to despawn after being out of combat for 5 seconds
- Resolves issues with despawning after summoning
-  If we prefer to have wait 10 seconds, a other option is to make the despawn out of combat timer longer instead. Vmangos appears to use a 5 second timer despawn as well, though, so I think the implementation of this PR is Blizz-like.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3675
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12171

## Tests Performed:

- Summoned all bosses, now correctly start combat and do not despawn

## How to Test the Changes:

1. Summon boss at greater windstone, ensure combat starts and boss does not despawn

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1656632586022126700.sql
+++ b/data/sql/updates/pending_db_world/rev_1656632586022126700.sql
@@ -1,0 +1,2 @@
+--
+UPDATE `smart_scripts` SET `event_param1` = 5000, `event_param2` = 5000 WHERE `entryorguid` IN (15203, 15204, 15205, 15305) AND `event_type` = 1;


Gultask:
Tested and works.

IntelligentQuantum (approved):

PR #12253, opened 2022-07-01
Small improvements for weapon/defence skill handling

schell244:
## Changes Proposed:
Small improvements for weapon/defence skill handling.
- blocked attacks should also have a chance increase weapon skill for attacker
- fix assigning moblevel (see comment: if defense than victim == attacker)
- intellect bonus for weapon skill should apply to all classes
- small code cleanup

## Issues Addressed:
- Closes 

## SOURCE:
https://wowwiki-archive.fandom.com/wiki/Weapon_skill (mentions paladins/shaman for intellect bonus)

## Tests Performed:
- build
- tested ingame

## How to Test the Changes:
1. create new character and and check weapon/defence skill increase. Should basically work as before, except that blocked attacks can also increase weapon skill.

- Formula for calculating increase chance is probably wrong.
https://de.classic.wowhead.com/guides/classic-wow-weapon-skills
_Each attack that is not evaded has a chance to increase weapon skill (evading enemies will not increase weapon skill, unfortunately). With lower weapon skill, the skill gain rate is at or near 100%, however the skill gain change begins to sharply decrease once the weapon skill cap is approached. At 10% away from the cap, the skill gain is roughly 50%, and gaining the final skill points can take some time (at which point your attacks should not be missing nearly as often)._ 
- Defence skill should increase slower compared to weapon skill (VMangos research)
- There's is a chance that skill can increased twice per hit (because proc is handled on attacker and victim?)

### DIFF
--- a/src/server/game/Entities/Player/PlayerUpdates.cpp
+++ b/src/server/game/Entities/Player/PlayerUpdates.cpp
@@ -980,42 +980,54 @@ void Player::UpdateWeaponSkill(Unit* victim, WeaponAttackType attType, Item* ite
 
 void Player::UpdateCombatSkills(Unit* victim, WeaponAttackType attType, bool defence, Item* item /*= nullptr*/)
 {
-    uint8 plevel    = getLevel(); // if defense than victim == attacker
+    uint8 plevel = getLevel();
+    uint16 currentSkillValue = defence ? GetBaseDefenseSkillValue() : GetBaseWeaponSkillValue(attType);
+    uint16 currentSkillMax = 5 * plevel;
+    int32  skillDiff = currentSkillMax - currentSkillValue;
+
+    // Max skill reached for level.
+    // Can in some cases be less than 0: having max skill and then .level -1 as example.
+    if (skillDiff <= 0)
+    {
+        return;
+    }
+
     uint8 greylevel = Acore::XP::GetGrayLevel(plevel);
-    uint8 moblevel  = victim->getLevelForTarget(this);
+    uint8 moblevel = defence ? victim->getLevelForTarget(this) : victim->getLevel(); // if defense than victim == attacker
     /*if (moblevel < greylevel)
         return;*/
     // Patch 3.0.8 (2009-01-20): You can no longer skill up weapons on mobs that are immune to damage.
 
     if (moblevel > plevel + 5)
+    {
         moblevel = plevel + 5;
+    }
 
     uint8 lvldif = moblevel - greylevel;
     if (lvldif < 3)
+    {
         lvldif = 3;
+    }
 
-    uint32 skilldif = 5 * plevel - (defence ? GetBaseDefenseSkillValue()
-                                            : GetBaseWeaponSkillValue(attType));
-    if (skilldif <= 0)
-        return;
-
-    float chance = float(3 * lvldif * skilldif) / plevel;
+    float chance = float(3 * lvldif * skillDiff) / plevel;
     if (!defence)
-        if (getClass() == CLASS_WARRIOR || getClass() == CLASS_ROGUE)
-            chance += chance * 0.02f * GetStat(STAT_INTELLECT);
+    {
+        chance += chance * 0.02f * GetStat(STAT_INTELLECT);
+    }
 
-    chance =
-        chance < 1.0f ? 1.0f : chance; // minimum chance to increase skill is 1%
+    chance = chance < 1.0f ? 1.0f : chance; // minimum chance to increase skill is 1%
 
     if (roll_chance_f(chance))
     {
         if (defence)
+        {
             UpdateDefense();
+        }
         else
+        {
             UpdateWeaponSkill(victim, attType, item);
+        }
     }
-    else
-        return;
 }
 
 void Player::UpdateSkillsForLevel()
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -15690,15 +15690,14 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
                 !target->IsPet()
            )
         {
-            // On melee based hit/miss/resist/parry/dodge need to update skill (for victim and attacker)
-            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST | PROC_EX_PARRY | PROC_EX_DODGE))
+            // On melee based hit/miss/resist/parry/dodge/block need to update skill (for victim and attacker)
+            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST | PROC_EX_PARRY | PROC_EX_DODGE | PROC_EX_BLOCK))
             {
                 if (target->GetTypeId() != TYPEID_PLAYER)
+                {
                     ToPlayer()->UpdateCombatSkills(target, attType, isVictim, procSpell ? procSpell->m_weaponItem : nullptr);
+                }
             }
-            // Update defence if player is victim and we block
-            else if (isVictim && procExtra & (PROC_EX_BLOCK))
-                ToPlayer()->UpdateCombatSkills(target, attType, true);
         }
         // If exist crit/parry/dodge/block need update aura state (for victim and attacker)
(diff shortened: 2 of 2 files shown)


acidmanifesto:
marking as needing feedback as i am not famliar with skill gains.

src/server/game/Entities/Player/PlayerUpdates.cpp
```diff
@@ -980,42 +980,56 @@ void Player::UpdateWeaponSkill(Unit* victim, WeaponAttackType attType, Item* ite
...
 
-    uint32 skilldif = 5 * plevel - (defence ? GetBaseDefenseSkillValue()
-                                            : GetBaseWeaponSkillValue(attType));
-    if (skilldif <= 0)
-        return;
-
-    float chance = float(3 * lvldif * skilldif) / plevel;
+    float chance = float(3 * lvldif * skillDiff) / playerLevel;
     if (!defence)
-        if (getClass() == CLASS_WARRIOR || getClass() == CLASS_ROGUE)
```
UltraNix, line 1004: Why this check is removed?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -15684,21 +15684,13 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
...
         // only when you are not fighting other players or their pets/totems (pvp)
-        if (GetTypeId() == TYPEID_PLAYER &&
-                target->GetTypeId() != TYPEID_PLAYER &&
-                !(target->IsTotem() && target->ToTotem()->GetOwner()->IsPlayer()) &&
-                !target->IsPet()
-           )
-        {
-            // On melee based hit/miss/resist/parry/dodge need to update skill (for victim and attacker)
-            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST | PROC_EX_PARRY | PROC_EX_DODGE))
+        if (IsPlayer() && !target->IsCharmedOwnedByPlayerOrPlayer())
```
UltraNix, line 15687: And what about charmed creature?

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -15684,21 +15684,13 @@ void Unit::ProcDamageAndSpellFor(bool isVictim, Unit* target, uint32 procFlag, u
...
+            // On melee based hit/miss/resist/parry/dodge/block need to update skill (for victim and attacker)
+            if (procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_MISS | PROC_EX_RESIST | PROC_EX_PARRY | PROC_EX_DODGE | PROC_EX_BLOCK))
             {
-                if (target->GetTypeId() != TYPEID_PLAYER)
-                    ToPlayer()->UpdateCombatSkills(target, attType, isVictim, procSpell ? procSpell->m_weaponItem : nullptr);
+                ToPlayer()->UpdateCombatSkills(target, attType, isVictim, procSpell ? procSpell->m_weaponItem : nullptr);
             }
-            // Update defence if player is victim and we block
-            else if (isVictim && procExtra & (PROC_EX_BLOCK))
-                ToPlayer()->UpdateCombatSkills(target, attType, true);
```
UltraNix, line 15701: Why removed? Victims should not skill up their defense skill on block?

UltraNix (changes requested):
Need answers.

src/server/game/Entities/Unit/Unit.cpp
schell244: What makes this `else if` statement different from the `if` above is that it checks for blocked attacks `PROC_EX_BLOCK` and that it additionally checks for `isVictim == true`.

With the condition `isVictim == true`, it will only allow a defence skill up for the player who blocked the attack. The case that player was attacker (`isVictim == false`) is not covered with weaponskill, so currently no chance so skill up weapon when attack was blocked.

To fix that, removing  `isVictim == true` and replacing the hard coded `true` for bool defence with `ToPlayer()->UpdateCombatSkills(target, attType, isVictim );` would cover both cases.

Or just remove the whole block and add `PROC_EX_BLOCK` to the if condition above.

src/server/game/Entities/Player/PlayerUpdates.cpp
schell244: Due to https://wowwiki-archive.fandom.com/wiki/Weapon_skill intellect bonus for weapon skill should apply to all classes (it mentions paladin and shaman) and not only warriors and rogues:
_Similarly, equip any gear you have that has an Intellect bonus on it—if you're a paladin, put on your Holy gear and cast Blessing of Kings; if you're a shaman, put on your Elemental or Restoration gear._

src/server/game/Entities/Unit/Unit.cpp
schell244: `!target->IsCharmedOwnedByPlayerOrPlayer()` should prevent any skill ups for targets that are players or any creatures owned by a player.
-> No weapon skill during pvp.

UltraNix: `IsCharmedOwnedByPlayerOrPlayer` means also charmed creature, not only owned, e.g. using Mind Control on some mob. The previous check did not prevent from that.

src/server/game/Entities/Unit/Unit.cpp
UltraNix: `ToPlayer()->UpdateCombatSkills(target, attType, isVictim, procSpell ? procSpell->m_weaponItem : nullptr);` Because now the `isDefense` parameter is always true if it is a victim, But should be true if it is victim AND it's PROC_EX_BLOCK .

schell244: The if condition above also handles both weapon and defence skill up for `PROC_EX_PARRY | PROC_EX_DODGE`. 
isVictim == true and player did parry/dodge -> defence skill
isVictim == false -> and player's attacks was parried/dodged -> weapon skill

_Each attack which is not evaded has a chance to increase weapon skill, with low weapon skill, ..._

So I think bock should be handled the same way as parry/dodge?

UltraNix: Alright. Looks okay

src/server/game/Entities/Unit/Unit.cpp
schell244: I can't prove that skill increase should not work on mobs that are temporarily owned by players.
I was using `!IsCharmedOwnedByPlayerOrPlayer()` becasue it makes the code look cleaner (also vmangos is using this check).

UltraNix: Okay.

UltraNix (approved):

Maelthyrr:
I have a doubt about intellect.  I'm sorry, I don't reminder how it was during woltk. I know that it's present in classic and tbc, but woltk comes with some stats modifications. And on your ownedcore source, I see a comment where someone say that was remove in woltk.

https://www.ownedcore.com/forums/world-of-warcraft/world-of-warcraft-guides/267567-intellect-raises-weapon-skill-faster.html

schell244:
> I've no Idea if blocked attacks should have a chance to increase weapo' skills but for others points, it's look fine to me.

Thanks for reviewing.
Regarding blocked attacks, both wowhead and wowwiki are writing:

_Each attack that is not evaded has a chance to increase weapon skill (evading enemies will not increase weapon skill, unfortunately)._

So I'm assuming blocked attacks should be handled the same way like missed/dodged/parried, ...

schell244:
Is the question whether blocked attacks should increase weapon skill the only remaining point of concern? In that case I could revert this change and add a TODO for this in the code?

Maelthyrr:
The main problem is that nobody have the knowledge to confirm this changes. Before to pass "in to be merge", I just want to check on classic ptr now we have woltk. But as you can see, we are really busy with AQ release on CC, and lot of PR need to be tested. But I will try to do this fast.

schell244:
Thanks for the update. No rush. I was just wondering if this was the only matter of concern. Love to see how much work and passion goes into AQ or any raid that has been reworked ❤️

Maelthyrr:
Sorry for the delay. Not easy to check this but it's always true on woltk classic.

schell244:
Sure, thanks again for checking.

acidmanifesto:
Thank you for the PR

PR #12290, opened 2022-07-03
fix(Core/Gossips): Properly send packet with questgiver query handler.

### LINKED ISSUE #12215: You need to log out to return the quest
### Current Behaviour

Many quests need to be logged out before they can be returned. They cannot be repeated before. It is accidental Find a way to reproduce today

https://youtu.be/TBJkX6gfqFg

It can be seen from the video that the task can be completed when I accept it, but I fly away on the mount provided by NPC and then fly back You can't return the task You need to log in again to return the task

none

### Source

_No response_

1. .go xyz 504.003845 2008.118896 113.079597 530
2. .q a 10382
3. talk `Gryphoneer Leafbeard` fly to `Shatter Point`
4. talk `Gryphoneer Windbellow` Fly back
5. talk `Field Marshal Brock` 

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/38889324b2bd4a77b47bf4774753198e280723ec

### Operating system

Ubuntu20.04

### Custom changes or Modules

_No response_

UltraNix:
Fixes #12215

## Issues Addressed:

- Closes #12215

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

`.go xyz 504.003845 2008.118896 113.079597 530`
`.q a 10382`
talk Gryphoneer Leafbeard fly to Shatter Point
talk Gryphoneer Windbellow Fly back
talk Field Marshal Brock

Also check #9962 if there is no regression on this.

### DIFF
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -98,6 +98,12 @@ void WorldSession::HandleGossipSelectOptionOpcode(WorldPacket& recv_data)
     if (_player->PlayerTalkClass->IsGossipOptionCoded(gossipListId))
         recv_data >> code;
 
+    // Prevent cheating on C++ scripted menus
+    if (_player->PlayerTalkClass->GetGossipMenu().GetSenderGUID() != guid)
+    {
+        return;
+    }
+
     Creature* unit = nullptr;
     GameObject* go = nullptr;
     Item* item = nullptr;
@@ -527,6 +533,16 @@ void WorldSession::HandleSetSelectionOpcode(WorldPacket& recv_data)
 {
     ObjectGuid guid;
     recv_data >> guid;
+
+    if (!guid)
+    {
+        // Clear any active gossip related to current selection if not present at player's client
+        GossipMenu& gossipMenu = _player->PlayerTalkClass->GetGossipMenu();
+        if (gossipMenu.GetSenderGUID() == _player->GetTarget())
+        {
+            _player->PlayerTalkClass->SendCloseGossip();
+        }
+    }
 
     _player->SetSelection(guid);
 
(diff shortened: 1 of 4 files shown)


mpfans:
tested and work

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -98,6 +98,12 @@ void WorldSession::HandleGossipSelectOptionOpcode(WorldPacket& recv_data)
     if (_player->PlayerTalkClass->IsGossipOptionCoded(gossipListId))
         recv_data >> code;
 
+    // Prevent cheating on C++ scripted menus
+    if (_player->PlayerTalkClass->GetGossipMenu().GetSenderGUID() != guid)
```
acidmanifesto, line 102: would this not break alot of exsisting custom npc gossips that are not based on guid of the character due to them being the need to be manaully spawned?

acidmanifesto (changes requested):

src/server/game/Handlers/MiscHandler.cpp
UltraNix: No. It's not a player's guid, but the guid of NPC you are talking to. You cannot select any option from other NPCs that you are not talking to.

acidmanifesto: > No. It's not a player's guid, but the guid of NPC you are talking to. You cannot select any option from other NPCs that you are not talking to.

I apologize, that is what i ment, the guid of the npc is not matching the gossip menu id. We have a few custom gossip npcs. Transmog npc being one wher we have to manually add to the playing field which will then give it a different guid from it's gossip.

UltraNix: `I apologize, that is what i ment, the guid of the npc is not matching the gossip menu id` Then it's wrong.

Issue #12279, opened 2022-07-02
[Emerald Dragons] Ysondre - Number of summoned adds is not blizzlike.


Gultask:
### Current Behaviour

When Ysondre summons the druid spirits, the number doesn't change depending on how many people are in the raid.

### Expected Blizzlike Behaviour

When Ysondre summons the druid spirits, the number of adds spawned should be proportional to the size of the raid, up to a maximum of 15 adds. See Extra Notes for Vmangos' application and some of my observations.

### Source

https://www.youtube.com/watch?v=BE00BpMb5NQ WotLK 2-Man, 1 add spawns
https://www.youtube.com/watch?v=cheP7oKQdN4 WotLK Solo, 1 add spawns
https://www.youtube.com/watch?v=_pxAPLeOroc Nov 23, 2008 6 people in the raid, I see 2-3 adds spawn.
https://youtu.be/5_IMr6hOoHg?t=133 Hunter Solo, 1 add spawns
https://www.youtube.com/watch?v=awCoyQ3WGnM Ditto
https://www.youtube.com/watch?v=grGTVFDN8ss WotLK 2-man, 1 add spawns
https://www.youtube.com/watch?v=aFmJVWsmElA WotLK 5-man, I see only 2 adds spawn
https://www.youtube.com/watch?v=jQNltxUYT8Q Classic
https://www.youtube.com/watch?v=caSlXK5KGmU Classic
https://www.youtube.com/watch?v=qd2C6xfEQfs Vanilla

### Steps to reproduce the problem

1. `.go c id 14887`
2. engage Ysondre with different number of raiders
3. see how many adds spawn

### Extra Notes

Dragons of Nightmare.
From: https://github.com/chromiecraft/chromiecraft/issues/3617

~~The WotLK footage shows only 1 add spawning at each phase. That seems to indicate that it is not based on a number of people in the raid but in how many people are in melee range instead, with a minimum of 1 add spawned.~~ Uncomfirmed. Vmangos scales the number of adds proportionally to the number of raiders.

Vmangos has the number of adds scaling with a formula of raiders×0.75. However, from the provided video sources here, I'd say that the scaling is closer to raidersx0.5, rounded down, and with a minimum of 1 to a maximum of 15 adds.

vmangos: https://github.com/vmangos/core/blob/development/src/scripts/world/dragons_of_nightmare/boss_ysondre.cpp

### AC rev. hash/commit

[...]

PR #12352, opened 2022-07-10
fix(Scripts/World: Spawn adds should scale with raid size

Maelthyrr:
## Changes Proposed:
-  Spawn adds will scale with raid size. Max is set to 15 adds.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12279

## SOURCE:

Cherry-pick: https://github.com/vmangos/core/blob/development/src/scripts/world/dragons_of_nightmare/boss_ysondre.cpp

## Tests Performed:

- None

## How to Test the Changes:

1. .go c id 14887
2. engage Ysondre with different number of raiders
3. .damage 90000
4. Count the number of adds

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/World/boss_emerald_dragons.cpp
+++ b/src/server/scripts/World/boss_emerald_dragons.cpp
@@ -273,9 +273,7 @@ public:
 
                 for (const auto attacker : attackers)
                 {
-                    if (Player* pPlayer = attacker->GetSource()->GetOwner()->ToPlayer())
-                        if (pPlayer && pPlayer->IsAlive() && !pPlayer->IsGameMaster())
-                            ++attackersCount;
+                    ++attackersCount;
                 }
 
                 uint8 amount = attackersCount < 20 ? attackersCount * 0.75f : 15;
--- a/src/server/scripts/World/boss_emerald_dragons.cpp
+++ b/src/server/scripts/World/boss_emerald_dragons.cpp
@@ -276,7 +276,7 @@ public:
                     ++attackersCount;
                 }
 
-                uint8 amount = attackersCount < 20 ? attackersCount * 0.75f : 15;
+                uint8 amount = attackersCount < 30 ? attackersCount * 0.5f : 15;
                 amount = amount < 1 ? 1 : amount;
 
                 for (uint8 i = 0; i < amount; ++i)
--- a/src/server/scripts/World/boss_emerald_dragons.cpp
+++ b/src/server/scripts/World/boss_emerald_dragons.cpp
@@ -277,7 +277,7 @@ public:
                 }
 
                 uint8 amount = attackersCount < 20 ? attackersCount * 0.75f : 15;
-                amount = amount < 3 ? 3 : amount;
+                amount = amount < 1 ? 1 : amount;
 
                 for (uint8 i = 0; i < amount; ++i)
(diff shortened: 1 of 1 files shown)


Maelthyrr:
@UltraNix can I get your review ? I'm really not sure about the line 276, but I don't find another way to use the condition isAlive and isGamemaster.

acidmanifesto (approved):

src/server/scripts/World/boss_emerald_dragons.cpp
```diff
@@ -268,7 +268,20 @@ class boss_ysondre : public CreatureScript
             {
                 Talk(SAY_YSONDRE_SUMMON_DRUIDS);
 
-                for (uint8 i = 0; i < 10; ++i)
+                auto const& attackers = me->GetThreatMgr().getThreatList();
+                uint8 attackersCount = 0;
+
+                for (const auto attacker : attackers)
+                {
+                    if (Player* pPlayer = attacker->GetSource()->GetOwner()->ToPlayer())
```
UltraNix, line 276: `Player* player` Say `No` to hungarian notations

src/server/scripts/World/boss_emerald_dragons.cpp
```diff
@@ -268,7 +268,20 @@ class boss_ysondre : public CreatureScript
...
+                        if (pPlayer && pPlayer->IsAlive() && !pPlayer->IsGameMaster())
```
UltraNix, line 277: `pPlayer` check needed - done above.

UltraNix (changes requested):
`if (Player* pPlayer = attacker->GetSource()->GetOwner()->ToPlayer())`
`attacker->GetSource()->GetOwner()` Are you sure it will never crash? `GetOwner` is always there?

Maelthyrr:
> `if (Player* pPlayer = attacker->GetSource()->GetOwner()->ToPlayer())` `attacker->GetSource()->GetOwner()` Are you sure it will never crash? `GetOwner` is always there?

It's was I asked, I don't see what use using the threatlist to check the state of players. Maybe isOnline ?

Nyeriah:
Just one consideration: game masters (gm mode on) should never be in a creature's threat list as well as dead players (but I'm not entirely sure about the second ones)

Maelthyrr:
I ask myself. If a player is in threatlist, he should be alive? If he die, he gone. So the two conditions are just useless ?

Nyeriah:
The GM one yes I'm sure, they're not valid targets so they can't be added to the threatlist. As for the dead players, I am not entirely sure, but I think we don't store references for dead players, yes

temperrr:
- [ ] 1 raider - 3 adds
- [ ] 5 raiders - 3 adds

Gultask:
I had the same results as temperrr but I tested it before the latest updates. I'll check again.

Also, I think you can remove the minimum spawn of 3 and set it to 1. In WotLK there wasn't a minimum of 3.
https://youtu.be/cheP7oKQdN4?t=39 here spawns only one.

Maelthyrr:
Need to add a check to be sure that a player is target. But I don't how treat with threatref to apply these checks. So I let someone else fix this.

temperrr:
tested, works

PR #12370, opened 2022-07-12
fix(Scripts/Gossip): Trial of the crusader

pangolp:
## Changes Proposed:
-  Select the gossip and npc text from the database, instead of always displaying the same gossip and hardcore texts.
-  Write the npc script to look like the blizzlike content.

<details>
<summary>Before</summary>

toc_default

</details>

<details>
<summary>After</summary>

toc_1

toc_2

</details>

## Issues Addressed:

- Closes 

## Tests Performed:

- Windows 10
- In-game

## How to Test the Changes:

1. `.go c 67185`
2. Set the raid in normal 10 mode
3. Perform the band normally, even dying, to verify that the npc shows the gossip again when re-entering.

## Topics to test

- [ ] In each fight, a different gossip and text should be shown on the npc.
- [ ] The npc that triggers the event remains inactive until the previous event ends.
- [ ] The door closes at the start of the event (I have to reverse this, now I really like it much better and it makes sense this way, but it doesn't seem to work this way).
- [ ] The last fight could be invoked in 2 ways. Waiting on the platform for 60 seconds, or through the gossip. According to the video, the fight is only activated by the gossip, so the automatic start after 60 seconds is disabled.
- [ ] Upon death, please check that everything is back to normal.
- [ ] The translation of a final text of Tirion was missing, in fact, it was missing to relate it, because the translation already exists.
- [ ] The second boss, had a hardcore text, although I have to keep working on it, because as it has 2 texts, I have to randomly select one of them.

- [x] Restructure the script, because although it works, it could be written in a different way, so as not to repeat so much code.
- [x] Move the IDS, both from the Gossip Menu and the Npc text, inside an enum.
- [x] Some npc text is also missing to be added to the gossip_menu. Actually, to link them. Although it works without that, the information must be in the database.
- [x] When the gossip is selected, the door should close to prevent the player from leaving the instance and the event from running normally. Currently, in some instances, it allows the player to exit even when the event has started.
- [x] The gossip flag should be available when the event ends in its entirety. To avoid that the player can start an event without finishing the previous one.


pangolp:
There are still some things to do, but at least the script is functional. I have tested it with a Spanish client, and it would be nice if other people could test it with other clients. And I have to improve, the things that I mention inside the TODO List.

pangolp:
I'm going to do a couple of commits, or things, that may be "extras" for some, but I want to use the pull request, to have some sort of "enhancements" to the commits. That show how the code is evolving, starting for example, with a series of if, and then, ending with a case. And that way, people who want to see the example and then replicate it (if they don't know how to do it), have an example to see.

pangolp:
The last event runs by itself after a while, that's why the NPC, at the end of the fourth battle, does not show in gossip. However, if the player exits and re-enters the instance, surely what happens is that the event is cut, and that is why it allows him to execute the gossip to call the event.

pangolp:
@Maelthyrr thanks.

Maelthyrr:
Thanks for you PR. It's tested and work fine :)

-[X] In each fight, a different gossip and text should be shown on the npc.
-[X] The npc that triggers the event remains inactive until the previous event ends.
-[X] Upon death,  everything is back to normal.

No new issues find.

pangolp:
This PR, besides changing the gossip and some other things, also changes the way the fight is done. Because as I mentioned in a moment, previously, the player could execute the event and go out the door. Now, when starting the event, he is trapped inside, having to do the same, die, or disconnect. I don't understand why someone would start an event and leave the place, but it really seemed absurd and I wanted to change it. I don't know if that kind of change, it's something blizzlike, it's something more common sense.

src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/instance_trial_of_the_crusader.cpp
```diff
@@ -879,8 +875,25 @@ class instance_trial_of_the_crusader : public InstanceMapScript
...
                         {
-                            c->Yell("Banished to the Nether!", LANG_UNIVERSAL);
-                            c->PlayDirectSound(16146, 0);
+                            uint8 textID = urand(0, 1);
+                            switch (textID)
+                            {
+                                case 0:
+                                {
+                                    c->AI()->Talk(9);
+                                    c->PlayDirectSound(16145, 0);
```
Kitzunu, line 884: why are we using magic numbers, and why isnt the sound tied to the text?

pangolp: > why are we using magic numbers, and why isnt the sound tied to the text?

I have no idea, that's how it was done, ask the person who wrote the script at the time, I only made some modifications to the gossip and then I started to look at some things related to the mechanics of the event.

Kitzunu: but you are changing this entire codeblock. If you didnt write this code, then where did you pull it from?

pangolp: > but you are changing this entire codeblock. If you didnt write this code, then where did you pull it from?

Yes, I changed the code, now I understand what you mean. The thing is that the NPC has 2 different dialogs, but in the database, it has only 1. For that reason, the original author left the code that way, and the truth is that I didn't edit it in that aspect. But I agree, that you could add the sound inside the creature_text, but for that, you would have to do a sniff, because it seems that there is an error in the text. One of them is missing.

pangolp:
This code, I don't know if it's really worth to merge it, beyond the gossip and the npc text. Because I make several changes in the logic of the fight, but reviewing videos, then I saw that these changes are not seen in other videos. It was more a matter of common sense. That the player, could not start the event, and leave the raid. But well, I leave it up to you, for me, it would be good to add it this way, but it will depend on you, if you want to do it or not.

pangolp:
Sorry for the late reply, I was traveling.

pangolp:
This pull request, I would close it. Because I made changes, in the logic, that I don't know if they correspond. The texts, if they would have to be added and those things, but I'll make another pull request with that, but the rest of the things, I think, affect the "mechanics". I really like it better this way, making that the player can't leave, if he triggers the event, but I don't know if it's correct.

Maelthyrr:
> This pull request, I would close it. Because I made changes, in the logic, that I don't know if they correspond. The texts, if they would have to be added and those things, but I'll make another pull request with that, but the rest of the things, I think, affect the "mechanics". I really like it better this way, making that the player can't leave, if he triggers the event, but I don't know if it's correct.

@Nefertumm @Nyeriah if someone get the time to review this before to merge in the coming days :)

pangolp:
> @Nefertumm @Nyeriah if someone get the time to review this before to merge in the coming days :)

I really don't know if it would be good to merge it, because it "changes" the logic a little bit. As I mentioned before, this makes the event, leaving the player locked, which seemed fine to me, but I don't know if it is the usual mechanics.

acidmanifesto:
Even though this is already tested quit some time ago, i would feel confident of a re-test to ensure there is no issues. The author has some concerns and i really do not want to invalidate this PR.

pangolp:
The concerns I have, are with respect to the mechanics, although the code clearly could be improved. But so much time has passed since I made it, that in the worst case, we could close it and do it later as a last resort. The translations part was interesting, but then, the rest of the mechanics implies that once the event is triggered, the player or players involved will not be able to leave the event unless they die or successfully complete the event. It is more than anything a mechanism to prevent a player from talking to the npc and leaving through the door, preventing the event from being triggered unnecessarily.

pangolp:
Then also, incorrect ways of working in the document were found, but they were reused, based on the code that already existed. For example, the execution of the sound, from the code, instead of being called from the database. The problem is that if all the creature_text / broadcast_text, are correct, there is something that I do not quite understand. The second Boss, in one of its dialogues, uses 2 sentences, in a random way, while in the emulator, there is only one of them.

src/server/scripts/Northrend/CrusadersColiseum/TrialOfTheCrusader/instance_trial_of_the_crusader.cpp
Kitzunu: just don't use hardcoded numbers in the core.

IntelligentQuantum:
@pangolp

1. create 2x account rank 0
2. create 2x hero horde
3. you can't click on npc

PR #12371, opened 2022-07-12
fix(db/creature): ZG entrance creature improvement 2

neifion-00000000:
Implements snake pack by entrance

## Changes Proposed:
-  Creates a pool for the first (and only) snake 3-4 pack in ZG  this will be the most complicated pack in the instance ironically. 

## Issues Addressed:

- Snake packs will spawn 1-3 of each type, never 4 of the same type, and about half the time will spawn only as 3

## SOURCE:
Blizzard servers

## Tests Performed:

- Tried several repetitions

## How to Test the Changes:

- Goto ZG
- Reboot server, goto ZG in new instance

- You should not need to reboot the server to have changes occur, this has nothing to do with data being put in this PR but rather with the way pooling is handled in instances somehow.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1657669676037619800.sql
+++ b/data/sql/updates/pending_db_world/rev_1657669676037619800.sql
@@ -1,0 +1,283 @@
+--
+/* Maintenance on ZG Entranceway mobs part 2 Pooling */
+DELETE FROM `creature` WHERE `guid` IN (49116, 49117, 49118, 49119);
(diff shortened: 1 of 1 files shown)


neifion-00000000:
Impacted area.

image

P.S. While I got this working locally, I got the PR pushed with 5 mins to spare, I'll be back to check its correct and it should be but its worth noting incase I forget to come back.

acidmanifesto (approved):

Maelthyrr:
Many try, and seems to fine.

Issue #12152, opened 2022-06-23
Ghost Mushroom at Skulk Rock in Hinterlands spawn above ground


ishidinho:
### What client do you play on?

enUS

### Faction

- [x] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

Origin report: https://github.com/chromiecraft/chromiecraft/issues/3660

They spawn above ground

### Expected Blizzlike Behaviour

They spawn in the cave below

### Source

Ghost Mushroom only spawn in caves, general knowledge

Comment by tester:

> As far as Azeroth goes, Ghost Mushrooms can be found only in [The Hinterlands](https://wowwiki-archive.fandom.com/wiki/The_Hinterlands) (in the cave beneath Skulk Rock, and in the cave near the altar in Jintha'Alor), [...]
(source: https://wowwiki-archive.fandom.com/wiki/Ghost_Mushroom#Source)

### Steps to reproduce the problem

go there
![image](https://user-images.githubusercontent.com/50233983/175050565-e71a0d5b-b965-4952-af63-d2db17f0317a.png)

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/6983903a89c556c0ca1239101cffc38098d6b081

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[...]

PR #12382, opened 2022-07-15
fix(DB/gameobject): Ghost Mushroom Position

ZhengPeiRu21:
## Changes Proposed:
-  Move Ghost Mushroom to be within underground cave, as close to existing position as possible
-  Co-ordinates are not sniffed but at least look more correct than existing values - before was incorrectly above ground which should not happen with Ghost Mushrooms

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12152
- Closes https://github.com/chromiecraft/chromiecraft/issues/3660

## Tests Performed:

- In-game:

## How to Test the Changes:

1. .go gameobject 16425
2. See if position looks correct (may need several attempts because it is pooled)

- [ ] If we are able to get sniffed values for nodes, that is better than do this solution, but getting all correct node sniffs is a large future project


Maelthyrr:
Tested and the position is fine now.

Nefertumm (approved):

Issue #12391, opened 2022-07-16
Inoculation - Quest: 9303


pangolp:
### Current Behaviour

Template change of moonkin plays levelup animation
Inoculated Nestlewood Moonkin are friendly to the player (green)

### Expected Blizzlike Behaviour

Template change of moonkin doesn't play levelup animation
Inoculated Nestlewood Moonkin are neutral to the player (yellow)

### Source

https://www.youtube.com/watch?v=sKz8eQbenjI

### Steps to reproduce the problem

1. .go c 75965
2. .quest add 9303

### Extra Notes

- https://wowgaming.altervista.org/aowow/?quest=9303
- https://wowgaming.altervista.org/aowow/?npc=16518

- https://www.wowhead.com/quest=9303/inoculation
- https://www.wowhead.com/npc=16534/inoculated-nestlewood-owlkin
- https://www.wowhead.com/item=22962/inoculating-crystal
- https://www.wowhead.com/spell=29528/inoculate-nestlewood-owlkin
- https://www.wowhead.com/npc=16518/nestlewood-owlkin

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/02fa20b6380b991634198de9a295abac11152d60

### Operating system

Ubuntu 20.04

### Custom changes or Modules

[mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
[mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
[mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
[mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
[mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
[mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
[mod-eluna](https://github.com/azerothcore/mod-eluna)
[mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
[...]

PR #12393, opened 2022-07-16
[WIP] Fix (Core/SmartScript.cpp) Control the change of faction

pangolp:
### Changes Proposed:
-  Currently, the npc, performs a level increase effect, due to the fact that inside the script, previously, there was an update of the template.
-  At the same time, the npc must remain neutral, and currently when using the item, it becomes friendly (due to the template change).

### Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3775
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12391

### SOURCE:

https://youtu.be/g3Qz_liqHR0

### Tests Performed:

- Windows 10
- In-game

### How to Test the Changes:

1. .go c 75965
2. quest add 9303

---


pangolp:
The script is wrong. I just realized, that the name of the NPC, should change, so the template change was ok. The problem then, must be in the template faction. But it strikes me, that there might be something wrong inside `creature_template`, anyway, I will check it. Or, the faction change may be in the spell, so I will make changes in this pull request.

pangolp:
Name: [Inoculated Nestlewood Owlkin](https://tbc.wowhead.com/npc=16534/inoculated-nestlewood-owlkin)
- Quick Facts
- Type: Humanoid
- Added in patch 2.5.1.38043
- Source: tbc.wowhead.com

Name: [Inoculated Nestlewood Owlkin](https://wowgaming.altervista.org/aowow/?npc=16534)
- Quick Facts
- Level: 3 - 4
- React: A H
- Faction: Friendly
- Source: wowgaming

pangolp:
I think the script is fine, maybe at most, the level change effect, but according to the source, it looks like it would have to be friendly. Anyway, I will try to look for more information about it.

Maelthyrr:
Tested and work fine.

pangolp:
Although it works, it is wrong, since when deleting the template, the name of the npc does not change. I think the problem is not in the smart_scripts, but in the spell. Thanks anyway for trying it, but this is not the solution we need. When I made the change, I thought it was the right thing to do, but then, the person who made the report, made me notice some points, that I had not taken into consideration. Like the fact that the npc changes its name after the spell is used on it.

pangolp:
I will check, if I find this spell inside the core, or if it is a dummy, we will have to create a spell, because it is doing some things that should not or rather, that it changes to friendly is fine, because the npc in creature_template is friendly, it would be a matter of modifying the behavior, only on that occasion.
https://www.wowhead.com/spell=29528/inoculate-nestlewood-owlkin

pangolp:
src/server/game/Spells/Auras/SpellAuraEffects.cpp:5994:

I'm going to check this out, see if we have any luck.

pangolp:
@Maelthyrr With this new change, it seems to be working.
It no longer does the level change effect.
It no longer changes the faction (in this case).

The faction is only changed if there is a level change.
This is a condition that does not convince me at all.
But if you want to make a SmartAI that changes faction
You would only have to set the `bool changelevel` to `True`.

WoWScrnShot_072622_105222

WoWScrnShot_072622_105229

pangolp:
I am not quite sure how I should resolve conflicts. If anything, we could create a new pull request, because now some things have changed. I'll wait for your confirmation to proceed.

pangolp:
I don't know if the conflict is well resolved. What happens, is that I didn't realize to update the file before making the change. But I think with that, it should be more than enough. Anyway, please, that someone receives it and confirms me to make the change.

Nyeriah:
I believe the merge conflict resolution went wrong. 

This is removing SMART_ACTION_SET_INST_DATA64 completely.

Also, the bottom change doesn't make sense. Creature level update doesn't have anything to do with faction...

pangolp:
> I believe the merge conflict resolution went wrong.
> 
> This is removing SMART_ACTION_SET_INST_DATA64 completely.
> 
> Also, the bottom change doesn't make sense. Creature level update doesn't have anything to do with faction...
> 
> image

It is true that it has no relation, but I needed some boolean variable, so that it is not always. My goal, is that as I pass false in the SmartAI call, then, by default, the faction change is not carried out. But in the case, if I passed true, then it would still work. That is the analysis I tried to make of the situation. I was not 100% convinced, but it occurred to me that it could become an alternative. But then came the conflict, and I resolved it badly. I think I am going to close the PR and reopen a new one, updating the information and from there on, we will see how we solve it. Because that's the way it works.

pangolp:
How do we make sure that the faction does not change, for example in this case? The only alternative I could think of is that, some kind of condition. I am not 100% satisfied with the solution, but as in the rest of the calls of the method, I never pass the `changelevel`, it occurred to me that it could be an alternative. Anyway, the compilation doesn't fail because of that, it fails because I solved the conflict wrong.

pangolp:
This pull request, I have to close it. I tried to fix it, but on the way, I made a lot of mistakes, so I will create another one later. Thanks to everyone who participated in it.

Issue #12447, opened 2022-07-20
Exploit: Instance boss farming upon logout on instances


WORISX:
### Current Behaviour

You can easily abuse and exploit the next steps to kill some specific bosses upong making a new band with a character that wasn't at the instance at the time of the kill with the other 2 characters that are being used to kill the boss, character 1 and 2 are killing the boss inside of the instance, then leaving group, and then being invited by 3rd character to a raid group, then both characters just logout and log-in again and the boss will be spawned and both characters that were inside of the instance will remain at the same place with the boss spawned again which allows to farming it over and over endlessly, and this 3rd character will be just outside all the time.

Currently there is no prompt or info output that tells that the reset has been done upon the new invitation from the 3rd character.

### Expected Blizzlike Behaviour

If a new invitation is being sent from a player that wasn't at the instance (3rd character), when log out is done by the 2 characters that were at the instance were the boss was killed, both should be sent to entry point of that instance or remain at the same place but without any reset happening (since i guess a reset is being made becouse the 3rd character is not binded to instance and is considered as a new instance id, and then the 2 characters that were previously at instance are at a new instance id) not allowing them to farm over and over that specific boss.

Only a new instance id must happen to character 1 and 2 if the new leader (character 3) is doing a manual reset, and if that 2 characters logout, the boss must remain killed, if no manual reset happens by new leader, character 1 and 2 will still be bounded to previous instance id until they go out of the dungeon and enter it again, and if character 3 enters instance, will be on a different instance id than the other 2 characters unless character 1 and 2 exit and enter again the instance.

If a manual reset happens, character 1 and 2 should exit and enter instance again to see character 3 inside of the instance of character 3, instance id's must not reset for players inside a previous instance upong receiving any kind of invitation.

### Steps to reproduce the problem

1º Create a group that consists of at least 2 characters and make a 3rd character that will remain outside NOT IN GROUP with player 1 and 2, just outside of the instance, that will be used later.
2º Go to an instance that is not making a permanent bind to instance, IE Ragefire Chasm, Stratholme or any other on Normal difficulty.
3º Go to the last boss of the instance with the 2 characters that are in group, and then, kill the boss you want to exploit at.
4º When the boss is killed, then leave the group with both 2 characters that were previously on the group (both will be prompted for teleport outside of the instance in 1 minute)
5º With your 3rd character be quick and invite both characters that are still at instance were boss was killed.
[...]

PR #12459, opened 2022-07-20
fix(Core): Instance Reset Exploit Fix

ZhengPeiRu21:
## Changes Proposed:
-  When a player logs out during a "not your instance" homebind timer, they are sent to beginning of the instance
-  Also an optimization to only check map players when logging out within an instance - previously would unnecessary fetch all players on non-instance maps when anyone in a group logged out.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12447

## Tests Performed:

- Not tested yet! Please help test. Thanks!

## How to Test the Changes:

1. Follow steps in original issue report
2. Ensure no longer can re-log in at boss location

- I do not know proper Blizz-like behavior when logging out with a homebind timer. It is possible correct behavior should be to go to homebind location or graveyard rather than instance entrance. If someone can confirm this behavior is different, we can change it, but I think we should merge this first to fix the farming exploit.

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2959,7 +2959,7 @@ bool InstanceMap::AddPlayerToMap(Player* player)
                 return false;
             }
         }
-        else if (playerBind && playerBind->save != mapSave)
+        else if (player->GetSession()->PlayerLoading() && playerBind && playerBind->save != mapSave)
         {
             // Prevent "Convert to Raid" exploit to reset instances
             return false;
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2959,6 +2959,11 @@ bool InstanceMap::AddPlayerToMap(Player* player)
                 return false;
             }
         }
+        else if (playerBind && playerBind->save != mapSave)
+        {
+            // Prevent "Convert to Raid" exploit to reset instances
+            return false;
+        }
         else
         {
             playerBind = sInstanceSaveMgr->PlayerBindToInstance(player->GetGUID(), mapSave, false, player);
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -673,11 +673,6 @@ void WorldSession::LogoutPlayer(bool save)
             }
         }
 
-        // If the player has a countdown to be removed from instance and logs out, send them to the beginning of the instance
-        // Prevents instance farming exploit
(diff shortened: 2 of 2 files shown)


WORISX:
Tested, not working so far, you can still reproduce and exploit even with the PR

Maelthyrr:
Tested in ragefire and deadmine but doesn't work.

ZhengPeiRu21:
Okay, thank you for testing. I will try a different way.

WORISX:
If i can suggest something, ill try checking if the instance id of the players in the group is the same instance id, if not, then when log out and log in again, teleport to entry point but not reset and still keep in the previous instance id where the boss is killed, only by exit instance and enter again is where the reset should happen, not on log out or log in, the current behaviour is not correct

ZhengPeiRu21:
I think this should be now work correctly. Would someone please test again?

benjymansy123:
Upon logging back in as per the original report, the two characters were teleported outside the instance (RFC). Upon walking back into the instance, it looks like the characters were sent to their homebinding rather than actually being teleported inside the instance. Don't know if that's intended behavior, but that's what happened.

ZhengPeiRu21:
> Upon logging back in as per the original report, the two characters were teleported outside the instance (RFC). Upon walking back into the instance, it looks like the characters were sent to their homebinding rather than actually being teleported inside the instance. Don't know if that's intended behavior, but that's what happened.
> 
> image

Thank you, that is helpful to know! It seems we are at least part working now. I will take another look.

ZhengPeiRu21:
I think it now should be working! Would you please mind to test again?

benjymansy123:
> I think it now should be working! Would you please mind to test again?

Okay, so upon logging back in, the two characters are just outside the instance (In Orgrimmar outside RFC) still in the raid group. Walking back in the instance loads RFC with all the mobs respawned, trash and boss.

ZhengPeiRu21:
> > I think it now should be working! Would you please mind to test again?
> 
> Okay, so upon logging back in, the two characters are just outside the instance (In Orgrimmar outside RFC) still in the raid group. Walking back in the instance loads RFC with all the mobs respawned, trash and boss.

Thank you so much for the multiple testings! This is the intended behavior.

If we are wanting to change other behavior, such as the player should return to their own old instance when re-entering, I can change that as well, but I think that is design change and not a bug fix. Right now, any time a group member who is not the leader enters an instance, it will enter the leader's instance. If the leader does not have one, a new instance will be created.

benjymansy123:
I did not do extensive testing outside of the described issue, so if it causes an issue outside of that test case, I didn't see it.

Gultask:
Tested as well.

Nyeriah (approved):

PR #12460, opened 2022-07-20
fix(DB/SAI): Zulian Crocolisk has scripted movement

neifion-00000000:
-  Adds id 2 and 3 into creature 15043's smart scripts.
-  Movement loops 30 seconds of 0 movement, 10 seconds of random movement in a small radius

## SOURCE:

Sniffed timers and estimated movement values
## Tests Performed:

- Tested in ZG.  Reset looks a bit abrupt but needed to keep from wandering all over the instance.

## How to Test the Changes:

1. Goto ZG
2. Watch any croc pack for a couple mins


neifion-00000000:
special thanks @55Honey  for the radius suggestion, saved time over my first guess.

ZhengPeiRu21:
Thank you Neifion! VMangos is having a field in their waypoint data table to set wander distance when a creature is at a waypoint with a long delay. If we see many more creatures that should be having movement patterns like this, I think we should implement that feature as well to save lots of manual scripting!

neifion-00000000:
> Thank you Neifion! VMangos is having a field in their waypoint data table to set wander distance when a creature is at a waypoint with a long delay. If we see many more creatures that should be having movement patterns like this, I think we should implement that feature as well to save lots of manual scripting!

that won't impact this one but yes I agree, it would cut down on the what I feel is pointless need for SAI in a lot of other packs.

waypoint_data also has issues with emotes firing off in weird ways too so pretty much everything that isn't a strict path needs dealt with in SAI atm

neifion-00000000:
found a bug in this, the mobs will now drift due to SAI random waypoint function

neifion-00000000:
fixed, this requires #12468 to function

data/sql/updates/pending_db_world/rev_1658354450598559500.sql
```diff
@@ -0,0 +1,12 @@
+--
+/* Crocs movement is a scripted action that occurs about every 30 seconds, not normal random movement */
```
acidmanifesto, line 2: no /* */ you sinner

acidmanifesto (changes requested):
image

acidmanifesto (approved):
Checks should be good. Too high on sugar booger to stay up for work flow to turn green.
**Snorting happy line noises**

Gultask:
Tested and it really does look much more like retail 👍

data/sql/updates/pending_db_world/rev_1658354450598559500.sql
```diff
@@ -0,0 +1,12 @@
+--
+-- Crocs movement is a scripted action that occurs about every 30 seconds, not normal random movement
+UPDATE `creature` SET `wander_distance`=0, `MovementType`=0 WHERE `id1`=15043;
+
+DELETE FROM `smart_scripts` WHERE `entryorguid`=15043;
```
Nyeriah, line 5: missing sourcetype, without it we risk deleting gameobject scripts for example

Nyeriah (changes requested):

Nyeriah (approved):

PR #12552, opened 2022-07-26
fix(Scripts/RuinsOfAhnQiraj): Implement Rajaxx waves gear check

Nyeriah:
## Changes Proposed:
-  Waves will engage the raid within 2 minutes if the previous one is not killed 
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1. pull pack
2. wait 3 min
3. check the next one comes

- [ ]
- [ ]


Nefertumm (approved):

Gultask:
- [X] Pulling and waiting 3 minutes pulls the next wave.
 
- [X] Pulling the intended second wave before the timer runs out seems to reset it.
 
- Pulling Drenn after Qeez before the 3 minutes pass will not reset the timer and will pull Tuubid. (Also made 2 packs, Yeggeth and Xurrem, pull at the same time when the 3 min timer passed)

- Pullind Drenn first instead of Qeez will cause Tuubid to pull after the 3 min.

We might need to check in retail if it's possible to 'sequence break' this encounter (what if I pull a different pack instead of Qeez first instead? what if I pull in the wrong order?). I think retail has a 2min timer between waves as well, but the only written source I found for a 2 minute timer is from a Classic Season of Mastery Guide so the 3min is probably correct for WotLK.

Gultask:
Ok I went to check in retail and it seems my suspicions were wrong. Pulling and killing Drenn's group first caused Xurrem to be pulled next, also causing him to pull 2 more packs due to assist mechanics. So you can disregard those points I mentioned in the last comment.
And for me the timer was 2 minutes.

PR #12560, opened 2022-07-27
fix(DB/creature_template): Remove Incorrect Loot

ZhengPeiRu21:
## Changes Proposed:
-  Remove incorrect loot from several creatures that should not have
-  Note that this is slightly different from original TC PR - I did not remove pickpocket loot from 12352 because it appears it should have it is correct (https://www.wowhead.com/wotlk/npc=12352/scarlet-trooper)
- Co-authored by offl

## Issues Addressed:

- Closes unreported issue

## SOURCE:

From TrinityCore https://github.com/TrinityCore/TrinityCore/commit/74ed066aa34fbd30161136d5989e99bbe34ab72d

## Tests Performed:

- SQL runs, no DB issues

## How to Test the Changes:

1. Test that creatures listed no longer drop loot

- If someone wants to confirm that Scarlet Trooper cannot be pickpocketed on retail, we can remove that loot as well, but otherwise I think we can remain as is - we are currently consistent with WoWHead.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1658940484513440700.sql
+++ b/data/sql/updates/pending_db_world/rev_1658940484513440700.sql
@@ -1,0 +1,3 @@
+--
+DELETE FROM `creature_loot_template` WHERE `Entry` IN (4625,11374,12352,27213,27414);
+UPDATE `creature_template` SET `lootid` = 0 WHERE `entry` IN (4625,11374,12352,27213,27414);


Maelthyrr:
Sniff from Woltk Classic PTR. Can't be pickpocketed.

Gultask:
Works. I think the Wowhead page has bad data, because there's only 9 counts of player pickpocketing them and it could very well be pserver data.
The Scarlet Warder, for comparison, has 7k counts.

Kitzunu (approved):

PR #12573, opened 2022-07-28
fix(Scripts/EmeraldDragons): fix summon player spam

Nefertumm:
## Changes Proposed:
-  Repeat the event every 500ms instead of every update.
-  Check for his actual target and don't use the other method <.<

## Issues Addressed:

- Closes 

## Tests Performed:

- Built.

## How to Test the Changes:

1. Go to any of the emerald dragons and engage with 2 chars
2. With the tank stay in melee range while you run away with the other, the one who isn't tanking should not be summoned ever.
3. Later try kiting with the tank, you should be ported as soon as you reach 50yds


temperrr:
test work

Nyeriah (approved):

PR #12632, opened 2022-08-04
feat(Core/Utilities): add event emitter class

r-o-b-o-t-o:
## Changes Proposed:
Adds an `EventEmitter` class that allows you to store multiple callbacks that you can execute later, with the ability to pass arguments to the functions.

**Example**:
* Add an `EventEmitter` to one of your classes:
```cpp
class Calculator
{
public:
    EventEmitter<void(int, int, int)> OnThingComputed;

    int ComputeThing(int a, int b)
    {
        int result = a * b;
        OnThingComputed(a, b, result);
        return result;
    }
};
```
* Then subscribe to the event with the `+=` operator:
```cpp
Calculator calc;
calc.OnThingComputed += [](int a, int b, int result)
{
    std::cout << "The result of ComputeThing(" << a << ", " << b << ") is " << result << std::endl;
};

calc.ComputeThing(3, 4);
calc.ComputeThing(2, -10);
```

This prints:
```
The result of ComputeThing(3, 4) is 12
The result of ComputeThing(2, -10) is -20
```

I intend to use this mainly as a way of adding communication between modules. This adds a good alternative to the hooks that the core already has, which in this case would be overkill and not modular at all.
Concrete example here: https://github.com/azerothcore/mod-anticheat/pull/79

## Issues Addressed:

N/A

## SOURCE:

N/A

## Tests Performed:

- Compiled and tested on Windows 10
- Compiled on Manjaro Linux

### DIFF
--- a/src/common/Utilities/EventEmitter.h
+++ b/src/common/Utilities/EventEmitter.h
@@ -1,0 +1,44 @@
+/*
+ * This file is part of the AzerothCore Project. See AUTHORS file for Copyright information
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by the
+ * Free Software Foundation; either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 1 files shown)


acidmanifesto (approved):

acidmanifesto:
Thank you for the pr

PR #12650, opened 2022-08-07
chore(CI): remove ubuntu-18.04

locus313:
## Changes Proposed:
-  remove ubuntu-18.04 in the process of being deprecated https://github.com/actions/virtual-environments/issues/6002
-  update SECURITY.md
-  remove duplicate labeled type

## Issues Addressed:

- Closes 


acidmanifesto (approved):

heyitsbench:
Due to the fact that OpenSSL 3.0 still isn't supported, won't the CI fail every time?

locus313:
> 

That shouldn't be a problem. we can just work around it like we do on macos.

locus313:
> > 
> 
> That shouldn't be a problem. we can just work around it like we do on macos.
going to break out adding 22.04 into a separate pr. Need to dig more into the overrides required for cmake

PR #12683, opened 2022-08-08
fix(Core/TempleOfAhnQiraj): Ouro rewrite

Nefertumm:
## Changes Proposed:
-  Properly implement submerge phase.
-  Script dirt mounds.
-  Added missing visuals for Ouro.
-  Properly implemented its berserk and boulder.

## Issues Addressed:

- Closes -

## SOURCE:

Sniffs.

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. Engage Ouro

- [ ] Cleanup the header file and sort data bosses to make SetBossState work
- [ ] Spells like Sand Blast and Sweep seemed to be kinda odd, we might need to recheck those.


pklloveyou:
Tested

Nyeriah (approved):

Gultask:
- [ ] These aren't despawning when Ouro submerges:

image

- [ ] The first time Ouro spawns it doesn't perform the Birth animation.

- [ ] During Phase 2, Ouro should periodically summon Dirt Mounds and those should summon scarabs periodically as well

Otherwise it looks great, but I can see now a potential exploit, which I can't confirm if it's blizzlike or not. If any character stays within melee range, or repeatedly moves in and out of melee range, the timer for the submerge for no melee targets seems to reset, making Ouro unable to deal much damage to ranged DPS if there's even 1 target within melee range.
So this raises 2 questions for me:

1. If the tank moves out of melee range, does Ouro start attacking another melee target?
2. Does the timer reset? furthermore, does the timer start counting for if the tank moves out of melee range or  for any melee character?

If you want me to I'll open issues about this so more work can be put towards AQ40.

Nefertumm:
@Gultask test again whenever you can

Gultask:
It looks good but somehow the gobs still aren't despawning when the boss submerges.
When the boss evades, however, the gobs despawn perfectly.

Nefertumm:
There we go

Gultask:
It works 👍

PR #12689, opened 2022-08-09
chore(Core/GameObject): Clear an error that often doesn't mean anything

Nyeriah:
## Changes Proposed:
-  Happens once f.e a death knight leaves a death gate open and logs off/dissappears/gets deleted
-  No reason for the crash scare as it often doesnt mean anything ill (have yet to see a crash from this in the past 12 months)

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -131,7 +131,7 @@ void GameObject::RemoveFromOwner()
         return;
     }
 
-    LOG_ERROR("entities.gameobject", "Delete GameObject ({} Entry: {} SpellId {} LinkedGO {}) that lost references to owner {} GO list.",
+    LOG_DEBUG("entities.gameobject", "Delete GameObject ({} Entry: {} SpellId {} LinkedGO {}) that lost references to owner {} GO list.",
         GetGUID().ToString(), GetGOInfo()->entry, m_spellId, GetGOInfo()->GetLinkedGameObjectEntry(), ownerGUID.ToString());
 
     SetOwnerGUID(ObjectGuid::Empty);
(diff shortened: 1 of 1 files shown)


IntelligentQuantum (approved):

acidmanifesto (commented):
Why not LOG_DEBUG?
I see this with dk death gates and a few profession GO's being summoned when the player logs off.
Why can it to a error log if you are clearing the error message?

Nyeriah:
> Why not LOG_DEBUG? I see this with dk death gates and a few profession GO's being summoned when the player logs off. Why can it to a error log if you are clearing the error message?

I am just removing the "This will probably crash" notice as it frightens people over something that is very unlikely to crash. I haven't moved it to the debug because I am not sure if there is something obscure here that I am not seeing and for some reason it would be important to log/notice this. I don't oppose the idea, however

acidmanifesto:
> > Why not LOG_DEBUG? I see this with dk death gates and a few profession GO's being summoned when the player logs off. Why can it to a error log if you are clearing the error message?
> 
> I am just removing the "This will probably crash" notice as it frightens people over something that is very unlikely to crash. I haven't moved it to the debug because I am not sure if there is something obscure here that I am not seeing and for some reason it would be important to log/notice this. I don't oppose the idea, however

the only way it will crash is if it hits the assert when there is no owner's guid. really only a couple cases exsist. Mainly with portals. The senior counter part has it going to debug for logging as well with a line of somewhat of a small explaination i guess dealing with the mage, im not really a mage player at all so i can not confirm that one note liner independently, can can confirm it with the death gate when the dk summon the death gate then logs off.

acidmanifesto (approved):

PR #12728, opened 2022-08-13
fix(Scripts/World): Emerald Dragon's Dream Fog should chase switch ta…

Nyeriah:
…rgets once reaching their chase target

## Changes Proposed:
-   Emerald Dragon's Dream Fog should chase switch targets once reaching their chase target
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- Not tested
- 

## How to Test the Changes:

1. Engage dragons with 2-3 characters
2. let the fogs pop up
3. see if they switch targets once they reach you

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/World/boss_emerald_dragons.cpp
+++ b/src/server/scripts/World/boss_emerald_dragons.cpp
@@ -36,7 +36,8 @@ enum EmeraldDragonNPC
     DRAGON_EMERISS                  = 14889,
     DRAGON_TAERAR                   = 14890,
 
-    GUID_DRAGON                     = 1
+    GUID_DRAGON                     = 1,
+    GUID_FOG_TARGET                 = 2
 };
 
 //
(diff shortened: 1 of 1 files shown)


Gultask:
Hmm I also had the threat table issue where the Dream Fogs were only detecting one target.

Tested with 3 characters, one tank was always attacking the boss.
Most of the time the Fogs would spawn and 1 or 2 would settle on the tank. Sometimes one would stay put.
The best way of attracting a Fog was healing since it would force me into its threat list. Then the Fog would move away from its stationary position to the healer. Healing the healer would move the fog again.
When they were moving they'd sometimes stop dead in its tracks.

these 2 fogs are idle:
image

Nyeriah:
Should be good now

Gultask:
Looks good 👍 
They are constantly travelling around now, targeting other players once they reach their previous target.

Nefertumm (approved):

Issue #12711, opened 2022-08-12
Weather system broken


ChibaPet:
### Current Behaviour

In Winterspring, it should never rain:

~~~
MariaDB [acore_world]> select * from game_weather where zone = '618';
+------+--------------------+--------------------+---------------------+--------------------+--------------------+---------------------+------------------+------------------+-------------------+--------------------+--------------------+---------------------+------------+
| zone | spring_rain_chance | spring_snow_chance | spring_storm_chance | summer_rain_chance | summer_snow_chance | summer_storm_chance | fall_rain_chance | fall_snow_chance | fall_storm_chance | winter_rain_chance | winter_snow_chance | winter_storm_chance | ScriptName |
+------+--------------------+--------------------+---------------------+--------------------+--------------------+---------------------+------------------+------------------+-------------------+--------------------+--------------------+---------------------+------------+
|  618 |                  0 |                 25 |                   0 |                  0 |                 20 |                   0 |                0 |               20 |                 0 |                  0 |                 25 |                   0 |            |
+------+--------------------+--------------------+---------------------+--------------------+--------------------+---------------------+------------------+------------------+-------------------+--------------------+--------------------+---------------------+------------+
~~~

and yet, it's raining right now. I've also observed it raining in Tanaris recently, which should only have sandstorms.

### Expected Blizzlike Behaviour

https://wowpedia.fandom.com/wiki/Weather_effects#Kalimdor_and_Eastern_Kingdoms

### Steps to reproduce the problem

Wait in a zone that should never see rain, and eventually it'll rain.

### AC rev. hash/commit

AzerothCore rev. 59e45c251eb5 2022-04-27 20:18:44 -0300 (master branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1n  15 Mar 2022 (library: OpenSSL 1.1.1n  15 Mar 2022)
Using Boost version: 1.74.0
Using MySQL version: 100515
Using CMake version: 3.18.4
Compiled on: Linux 5.10.0-13-amd64
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /opt/acore/data/maps. Total size: 291218471 bytes
vmaps directory located in /opt/acore/data/vmaps. Total size: 658130721 bytes
[...]

PR #12729, opened 2022-08-13
fix(Core/Weather): Wrong change type algorithm

SoglaHash:
There's a rare bug that occurs where it starts raining in zones that
 shouldn't like Winterspring and Tanaris.

https://user-images.githubusercontent.com/74299960/184515714-7074c2e7-201d-4c7e-b4ff-bcddf75accb1.mp4

## Changes Proposed:
-  Change range to not include chance 0

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12711#issuecomment-1214226996

## SOURCE:

https://wowpedia.fandom.com/wiki/Weather_effects#Kalimdor_and_Eastern_Kingdoms

## Tests Performed:

I changed the setting in `worldserver.conf.dist` and managed to get weather updates more frequently and was able to reproduce the issue.
```
#
#    ChangeWeatherInterval
#        Description: Time (in milliseconds) for weather update interval.
#        Default:     600000 - (10 min)
ChangeWeatherInterval = 20
```
I used vscode to set a conditional breakpoint when the rnd hits 0. And as you can see chance1 is the chance to rain in the zone and it'll change the type to rain.

chance0_0_set_type_rain_bug

The randomness is also inaccurate. For example, for a zone with 5% chance to rain. The rnd value ranges [0,99]. With the `<=` operator,  it gives true on values `0,1,2,3,4,5` which is 6%. Changing the range to [1,100] would also correct the randomness to 5%.

## How to Test the Changes:

Change `worldserver.conf.dist` to a low value 
1. Observe ingame to see if it rains in Tanaris, Winterspring
2. or set a breakpoint and investigate
3.

- [ ] Isle of Quel'danas and Western Plaguelands both have rain_chances > 0 
- [ ]

### DIFF
--- a/src/server/game/Weather/Weather.cpp
+++ b/src/server/game/Weather/Weather.cpp
@@ -146,11 +146,11 @@ bool Weather::ReGenerate()
     uint32 chance3 = chance2 + m_weatherChances->data[season].stormChance;
 
     uint32 rnd = urand(0, 99);
-    if (rnd <= chance1)
+    if (rnd < chance1)
         m_type = WEATHER_TYPE_RAIN;
-    else if (rnd <= chance2)
+    else if (rnd < chance2)
         m_type = WEATHER_TYPE_SNOW;
-    else if (rnd <= chance3)
+    else if (rnd < chance3)
         m_type = WEATHER_TYPE_STORM;
     else
         m_type = WEATHER_TYPE_FINE;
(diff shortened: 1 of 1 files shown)


heyitsbench:
😆  Good find!

heyitsbench:
Spent half an hour in both Tanaris and Winterspring with the config option set to 200, both only ever snowstormed/snowed respectively.

Side note, there are a few other instances of `urand(0, 99)` in the code, those should also probably get looked at to get changed to 1, 100 as well in the future.

acidmanifesto (approved):

PR #12749, opened 2022-08-15
refactor(Core/TempleOfAhnQiraj): Temple of Ahn Qiraj header

Nefertumm:
## Changes Proposed:
-  Move some bosses from scriptedai to bossai
-  Sort datas for the header, this allows us to use set boss state properly to handle doors, etc.

## Issues Addressed:

- Closes -

## Tests Performed:

-  Tested some bosses.
-  Tested ouro spawner respawn on fail.

## How to Test the Changes:

1. Go to Temple of Ahn Qiraj and see the boss states for each boss.


Nyeriah (approved):

Gultask:
Couldn't find any errors. All bosses are still casting spells. Events such as the C'Thun swallowing still work like they did before.

PR #12846, opened 2022-08-25
Minor CMake Update to Modules Cmake File to Allow Custom Modules to make changes without changing Core.

DavidStrickland0:
## Changes Proposed:
- Enable module developers to include a cmake file into the CmakeList.txt of the Modules Project so they can update dependencies or perform other module specific tasks without altering Modules CMakeList.txt. There are already several module specific changes in Core that this approach should alleviate moving forward.  

## Tests Performed:
It Does build given this is a CMAKE only change without any payload changes the only thing that might break is the compile which appears to be working.

## How to Test the Changes:
Pull the Repo and Compile. This is a CMake only change and a minor one at that.


AnchyDev:
I agree this would be a great addition for module developers.
Tested successfully with CMake 3.23.2 on Windows 10 (19044.1826) and was able to successfully include & link & compile with my module that has a mono dependency.

Kitzunu (approved):

PR #12855, opened 2022-08-26
fix(Core/TempleOfAhnQiraj): Twin emperors rewrite

Nefertumm:
## Changes Proposed:
-  Rewrote the whole encounter.
-  Fix Intro.
-  The Master's Eye should despawn before the intro.
-  Set proper spell immunity flags for each boss.
-  Use proper spells for each boss.
-  Added a new overload for NearTeleportTo, now it accepts Position as parameter.
-  Added missing mechanics from mutate bug.
-  Added missing texts.
-  They must share their hp but when they heal they don't. It's an odd behaviour but it's shown at it is in sniffs.

## Issues Addressed:

- Closes 

## SOURCE:

Sniffs

## Tests Performed:

- Tested in-game, but it's hard to test alone.

## How to Test the Changes:

1. Go to Twin Emperors and engage them.

- [x] Missing sounds/says. 
- [x] Critters respawn time.
- [ ] Handle doors (probably for another PR, since they need to be moved to their proper system)
- [ ] For some reason Master's Eye keeps falling to the ground sometimes, it should be always hovering.


Gultask:
- [ ] A lot of the times only one of the Twins will drop loot. Probably because players aren't actually dealing damage to the other emperor.
- [ ] Mutate Bug/Explode Bug can be used on a creature that already has one of the auras, making them grow enormous
- [ ] After the teleport there is a NE female sound?
- [ ] Explode Bug doesn't work. The aura is removed instantly.
- [ ] Vek'lor is casting way too much it looks like. Maybe a rescheduling issue?
- [ ] I see no speech bubbles from them, at least when standing far away.

Nyeriah (approved):

PR #12881, opened 2022-08-28
chore(Core): Debloat Mail.cpp

Synful-Syn:
## Changes Proposed:
- Debloat the following by process of elimination since `receiver` won't suddently change and avoid duplicate checks "are we missing `receiver`?".
image
image
image
- Don't do `ObjectAccessor::FindPlayerByLowGUID()` if we don't need it.
- ~~Don't calculate the value of `index` since it is always the same value for each `stmt->SetData(...)`.~~ (cancelled)
- Use directly the value of `mailItemIter->second` instead of putting it into an `Item*`. Same with `CharacterDatabaseTransaction(nullptr)`.

## Issues Addressed:

- Closes Nothing.

## Tests Performed:

- Got mails from player.
- Got mails from player with items
- Got mails from GM.
- Got mails from GM with items.
- Got mails from AH.
- Returned mails.
- Deleted mails.
- Checked the returned mails.

## How to Test the Changes:

Do mail stuff.

- [ ]

### DIFF
--- a/src/server/game/Mails/Mail.cpp
+++ b/src/server/game/Mails/Mail.cpp
@@ -153,13 +153,7 @@ void MailDraft::SendReturnToSender(uint32 /*sender_acc*/, ObjectGuid::LowType se
 {
     Player* receiver = ObjectAccessor::FindPlayerByLowGUID(receiver_guid);
 
-    uint32 rc_account = 0;
-    if (!receiver)
-    {
-        rc_account = sCharacterCache->GetCharacterAccountIdByGuid(ObjectGuid(HighGuid::Player, receiver_guid));
-    }
-
-    if (!receiver && !rc_account)                            // sender not exist
+    if (!receiver && !sCharacterCache->GetCharacterAccountIdByGuid(ObjectGuid(HighGuid::Player, receiver_guid)))
     {
         deleteIncludedItems(trans, true);
         return;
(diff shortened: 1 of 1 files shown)


acidmanifesto (approved):

acidmanifesto:
We need a Return to Sender Mail Test for this PR

src/server/game/Mails/Mail.cpp
```diff
@@ -219,43 +212,41 @@ void MailDraft::SendMailTo(CharacterDatabaseTransaction trans, MailReceiver cons
...
-    stmt->SetData(++index, sender.GetSenderId());
-    stmt->SetData(++index, receiver.GetPlayerGUIDLow());
-    stmt->SetData(++index, GetSubject());
-    stmt->SetData(++index, GetBody());
-    stmt->SetData  (++index, !m_items.empty());
-    stmt->SetData(++index, uint32(expire_time));
-    stmt->SetData(++index, uint32(deliver_time));
-    stmt->SetData(++index, m_money);
-    stmt->SetData(++index, m_COD);
-    stmt->SetData (++index, uint8(checked));
```
Kitzunu, line 250: I prefer this as it is impossible to mess up if tables are added or removed.

Synful-Syn: I have "reverted" it. I think your point is absolutely valid.

Gultask:
All of the provided above work. Also tested replying and returning mails and it works.

Nefertumm:
Ty for your PR Syn!

Issue #5366, opened 2021-04-19
Desolace Starving Bonepaws don't path


mdawx:
Originally reported https://github.com/chromiecraft/chromiecraft/issues/445
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [x] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

https://wowgaming.altervista.org/aowow/?npc=4689
https://wowgaming.altervista.org/aowow/?npc=4690
https://wowgaming.altervista.org/aowow/?npc=4688

These hyena's don't path around desolace.

(Not sure if they path, but don't path after they respawn? Just was randomly going through desolace and noticed they don't path)

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Those hyena's should path around desolace.

##### EXTRA NOTES:

.go c 27973 (Rabid Bonepaw)
.go c 27970 (Rabid Bonepaw)

.go c 27953 (Bonepaw Hyena)
.go c 27946 (Bonepaw Hyena)
.go c 27939 (Bonepaw Hyena)

.go c 27956 (Starving Bonepaw)
.go c 27963 (Starving Bonepaw)

Source: https://classic.wowhead.com/npc=4688/bonepaw-hyena#comments
> (..) they also travel in a pach of 5-6. They roam the western edge of desolace, west of the Kodo graveyard, a little NorthEast of Shadowprey village. 

https://classic.wowhead.com/npc=4688/Starving-Bonepaw#comments
https://classic.wowhead.com/npc=4688/Rabid-Bonepaw#comments

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/b8b5f68f98ff8771a6ec87ea6805ceced91c406a

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #12912, opened 2022-08-31
Bugs/445

DavuKnight:
## Changes Proposed:
- Update SmartAI causing 4690 to Move Randomnly instead of standing still 

## Issues Addressed:
- Closes 
https://github.com/chromiecraft/chromiecraft/issues/445
https://github.com/azerothcore/azerothcore-wotlk/issues/5366

## SOURCE:
None- Fixed based on the Bug report and that I dont remember any mob just standing still
## Tests Performed:
- Visited 4688 4689 and 4690 they all stagger around like drunken salors now. 

## How to Test the Changes:
1. Apply SQL script
2. .go xyz 1789.01	92.1613	2.94131
3. Ensure 4688 is moving randomly
4. .go xyz 1104.43	100.077	5.36425
5. Ensure 4689 is moving Randomly
6. .go xyz 1799.25	66.8346	3.06855
7. ensure 4690 is moving randomly


DavuKnight:
@Gultask thanks for the reply I Initially tried the MovementType approach your referenceing by changing it to 1 but it wouldnt take. Only thing different between the 2 packs that were working with Movement type 1 and the one that wasn't was that the one that wasn't working with Movement Type 1 had a Smart AI Script to apply a rabid bite and the other two did not.

Once I added it to the SmartAI it seemed to work well enough. The 10 made them move roughly the same way the other 2 packs were moving and seemed to be consistent with the other instances of the SmartAI doing the Random movement that I found in the SmartAI table.

Gultask:
I can think of 2 reasons for that. One of them is one I had with a recent PR of mine that dealt with the rare 'Araga'. It had a `creature_addon` entry that was assigning it a inexistent waypoint path. Removing it made it work.

The other problem could be due to changing the creature_template table instead of the creature table. I used to do it a lot :p

DavuKnight:
Good enough the Template thing might well have been it. I was using Keira for the first time instead of just using the database so really dont know what it was changing

DavuKnight:
Might take another stab at it later thanks for the feedback

Gultask:
No worries ^^ thanks for the PR, either way!

PR #12938, opened 2022-09-02
fix(DB/broadcast_text): Correct Field Marshal Snowfall's Light Leather text.

heyitsbench:
## Changes Proposed:
-  Swaps a couple variables in Field Marshal Snowfall's broadcast text for stating War Effort values.
Currently the dialogue looks like [this](https://user-images.githubusercontent.com/38561765/188220254-4e17e28e-f302-4a03-98ee-ccf65bb71b88.png) (as taken from the CC Ruins of AQ channel by @kerricsson). Specifically, Snowfall says "It looks like out of [number collected] light leather needed, we have [number quota]. This PR aims to swap those two variables to make it correct.

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

Common English language. Also [Wowpedia](https://wowpedia.fandom.com/wiki/Field_Marshal_Snowfall).

## Tests Performed:

- Untested.

## How to Test the Changes:

1. Don't know, can't figure out the War Effort on my local.


Nyeriah:
@heyitsbench thanks for the PR, however the dialogue is written by hand as originally those were gossips. 
broadcast_text table is never wrong, however, we are not using it for the war effort

heyitsbench:
So then where would the offending entry used for the effort be?

Nyeriah:
Should've been corrected here unless I did yet another baddie: https://github.com/Nyeriah/mod-war-effort/commit/d6dd43c92189deadd33efd51641324b584dafdac

heyitsbench:
Ah, so the war effort is being handled in a module rather than in the core itself, okay. Thank you!

PR #12985, opened 2022-09-07
fix(DB): Missing Ironforge Guard on Mountain

ZhengPeiRu21:
## Changes Proposed:
-  Add missing Ironforge Guard on top of Mountain
-  Values from Vmangos
- Thank you Bench for notice this small fun details ^_^

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/4027
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12943

## SOURCE:

Vmangos

## Tests Performed:

- Tested in game, guard is present and walks on mountain

## How to Test the Changes:

1. .go creature 190247
2. Check position and pathing is okay


heyitsbench:
https://youtu.be/e7MCOm1OxEs
This is what the path looks like in the Wrath pre-patch.

Gultask:
Climbs the mountain like a Skyrim horse 👍 

Very slight clipping at one point but it's totally due to the awkward angles. Seems to be similar to Bench's video in the spots that he appears.

heyitsbench:
> Seems to be similar to Bench's video in the spots that he appears.

Does the delay at the peak line up with the video? 30000 feels very long at a glance.

Gultask:
Ah, you're right. In the video it appears more like a 1-2s pause instead. I only looked at the path.

ZhengPeiRu21:
I have updated to reduce the wait time at the peak to 3 seconds, which seems closer to the video. If we are prefer then we can wait to merge until we have a sniff of waypoints to confirm all is exactly correct.

heyitsbench:
> If we are prefer then we can wait to merge until we have a sniff of waypoints to confirm all is exactly correct.

I've already got a sniff of the guard's pathing. Once WPP fixes its delay parsing, we can correct the value to be exact, but as of right now, I'd say 3000 is good.

Nefertumm (approved):

Issue #12116, opened 2022-06-19
Fishing skill not leveling properly


Stump876:
Orignial Chromiecraft report https://github.com/chromiecraft/chromiecraft/issues/3619 pretty much copy and pasted with some edits for GM commands.

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

Currently the way that fishing works in AC is that only successful "fish" catches can level up your skill, catching junk does not ever level up your skill, making it impossible to level up your fishing by catching junk in areas that are too high level for you.
![image](https://user-images.githubusercontent.com/68164076/174491802-02e0b671-5c8b-473e-af50-f268c57a0ffd.png)

### Expected Blizzlike Behaviour

In WotLK after patch 3.1.0, Fishing was significantly reworked and allowed you to catch 'junk' anywhere that you could fish, though you would only actually catch 'fish' if your skill was high enough for where you were fishing. So you could, for example, level fishing from 1-450 without ever leaving Dalaran, even though as it works in AC currently, you would never be able to level up by catching junk or coins in the fountain.

### Source

 [Patch 3.1.0](https://wowpedia.fandom.com/wiki/Patch_3.1.0)
`There is no longer a minimum skill requirement to fish in any zone. Every catch has the potential for fishing skill gains, but you are likely to catch worthless junk in areas that are too difficult for your skill.`

### Steps to reproduce the problem

1. .learn 7620
2. .additem 6256 and equip
3. .go xyz 2728.435791 3875.718262 0.506867
4. .gm off
5. fish as normal
6. Observe that you do not gain skillups.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/36f0a7757bde9573262f298aa1eaf6765af8295e

### Operating system

Ubuntu 20.04
Win10

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
[...]

PR #12996, opened 2022-09-09
fix(Core): Correct Post 3.1 Fishing Skill Leveling

ZhengPeiRu21:
## Changes Proposed:
-  Corrects fishing skill to have to chance to level even when fishing junk (correct post 3.1 behavior)
-  Also adds a hook that can be used by servers want to re-create the current pre 3.1 behavior.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12116
- Closes https://github.com/chromiecraft/chromiecraft/issues/3619

## SOURCE:

3.1 Patch Notes

## Tests Performed:

- Still needs testing

## How to Test the Changes:

1. Fish in high level zone with low level skill
2. See skill up when catching junk
3. Assure skill still can increase when catching fish as well!

To restore old behavior, a PlayerScript can be used. (I like the pre 3.1 behavior better, but that is an opinion ^_^):
```
    bool OnUpdateFishingSkill(Player* player, int32 /*skill*/, int32 /*zone_skill*/, int32 chance, int32 roll) override
    {
        if (chance < roll)
            return false;
        return true;
    }
```

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -1744,11 +1744,13 @@ void GameObject::Use(Unit* user)
 
                             LOG_DEBUG("entities.gameobject", "Fishing check (skill: {} zone min skill: {} chance {} roll: {}", skill, zone_skill, chance, roll);
 
+                            if (sScriptMgr->OnUpdateFishingSkill(player, skill, zone_skill, chance, roll))
+                            {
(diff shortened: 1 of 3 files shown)


Gultask:
Works 👍 
Fishing junk can increase skill level
At higher levels, fishing junk and actual fish both can contribute towards skill ups, and one has no advantage over the other.

Nefertumm (approved):

PR #13023, opened 2022-09-14
feat(core): OnFfaPvpStateUpdate Event

DavuKnight:
## Changes Proposed:
- Add the OnFfaPvpStateUpdate Event so Lua Scripts can know when a user enters or leaves Ffa Pvp.   

## Issues Addressed:

- Closes 

## Tests Performed:
- Builds Fine. There should be no Client Impact thats visible and testable. 

## How to Test the Changes:
55Honey has a direct need for this event in a current LuaScript and will Likely be the one to test it.


Nyeriah:
Hey, any reason why you’re checking if they have the flag before removing? It’s not a very common practice as it has no practical effect

Nyeriah (approved):

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -2118,6 +2128,7 @@ class ScriptMgr
     void OnCreatureUpdate(Creature* creature, uint32 diff);
     void OnCreatureAddWorld(Creature* creature);
     void OnCreatureRemoveWorld(Creature* creature);
+    void OnFfaPvpStateUpdate(Creature* creature, bool InPvp);
```
Nyeriah, line 2131: Should it be player instead of creature?

DavuKnight:
I needed to do the check for the event I didn’t want the Event to fire unless the Byte actually changed not just because the system Set it on or off when it was already on or off.  Sent from Mail for Windows From: SkjalfSent: Wednesday, September 14, 2022 3:33 AMTo: azerothcore/azerothcore-wotlkCc: DavuKnight; AuthorSubject: Re: [azerothcore/azerothcore-wotlk] OnFfaPvpStateUpdate Event (PR #13023) Hey, any reason why you’re checking if they have the flag before removing? It’s not a very common practice as it has no practical effect—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

Nyeriah:
Fine by me but the CI build is failing, please take a look at those errors

55Honey:
I don't understand why we fire the hook if a creature and a game object change their FfaPvP state. Do they have that state at all? Wouldn't it suffice to check just for players? Or I do read the code wrong maybe.

DavuKnight:
I did it everywhere I saw the UnitPVPStateFlags.UNIT_BYTE2_FLAG_FFA_PVP Byte being set. 

It should be Player and Creature. The Creature is to account for Pets. But I didnt dig into the why of someone trying to set FFA_PVP I just set the event up for any place the UNIT_BYTE2_FLAG_PVP was being set. The only change to the when/why or how its being set. Was that It now checks it before setting it. This is so the event only fires when the stat actually changes not whenever someone updates a state to the same state its already in. 

Looking at the call in GameObject thats a Spell handler I presume it ensures that your Per Enters FFA_PVP as soon as you summon it when necessary.

55Honey:
Tested with <https://github.com/azerothcore/mod-eluna/pull/63>
image

acidmanifesto:
Thank you for the PR

Nefertumm:
Hi Dav, we got a crash.
https://gist.github.com/Nefertumm/fea2a888b94dcc720f51688835cd4be3

DavuKnight:
Looking through the crash log Im not understanding the connection. Is there a specific Area of the log that assiciates the crash to this change? Or is it just a timing thing?

src/server/game/Entities/GameObject/GameObject.cpp
```diff
@@ -2086,8 +2086,14 @@ void GameObject::CastSpell(Unit* target, uint32 spellId)
         if (owner->HasUnitFlag(UNIT_FLAG_PLAYER_CONTROLLED))
             trigger->SetUnitFlag(UNIT_FLAG_PLAYER_CONTROLLED);
         if (owner->IsFFAPvP())
-            trigger->SetByteFlag(UNIT_FIELD_BYTES_2, 1, UNIT_BYTE2_FLAG_FFA_PVP);
+        {
+            if (!HasByteFlag(UNIT_FIELD_BYTES_2, 1, UNIT_BYTE2_FLAG_FFA_PVP))
+            {
+                sScriptMgr->OnFfaPvpStateUpdate(trigger, true);
+                trigger->SetByteFlag(UNIT_FIELD_BYTES_2, 1, UNIT_BYTE2_FLAG_FFA_PVP);
+            }
```
Nefertumm, line 2094: From the crashlog, this is related. I found it odd too :(

DavuKnight: Thread #19.2  is calling GameObject line 2092. 2092 is the event handler. Could it be something in an event payload in a lua script?
Should Event caLLs be wrapped in a try / catch / swallow to avoid crashing the server if the event overloads Blow out?

Nefertumm: I'm pretty sure there is no lua script running with that hook yet, although It's pretty odd indeed. It was a frost trap that made it crash, which is shown on the crashlog, but I don't have more information about it sadly :/ 
The event per-se it's okay, it doesn't need a try catch

DavuKnight: Seems to me the biggest change is the 
`    ASSERT(offset < 4);`
That HasByte Performs compared to the 

```
    if (offset > 3)
    {
        LOG_ERROR("entities.object", "Object::SetByteFlag: wrong offset {}", offset);
        return;
    }
```
That SetByteFlag Does

Anyone guess why HasByteFlag needs to validate that offset <4?

DavuKnight: https://github.com/azerothcore/mod-eluna/pull/63 went in a few hours back not sure how quick the server picks those up

55Honey: While the mod-Eluna PR is merged, there are no events registered yet in any Lua scripts.

PR #13141, opened 2022-09-26
fix(Core/TempleOfAhnQiraj): Bug trio

Nefertumm:
## Changes Proposed:
-  Codestyle.
-  Change the way to get each boss object
-  Don't let bosses dying auto attack.
-  Remove player damage req whenever they heal.
-  Do their final spell after actually dying and not before.
-  Trying to fix an odd case of they dying with all of them dropping loot and not doing their final spell.
-  Fix timers according to sniffs
-  Refactor their death mechanic to be more blizzlike.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/4087

## Tests Performed:

- Not tested.

## How to Test the Changes:

1. Engage bug trio, only one of them should drop loot, the dying creature should not auto attack.

### DIFF
--- a/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_bug_trio.cpp
+++ b/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_bug_trio.cpp
@@ -187,7 +187,7 @@ public:
 
     void UpdateAI(uint32 diff) override
     {
-        if (!UpdateVictim() || _isEating || !CheckInRoom())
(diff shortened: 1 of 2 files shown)


Gultask:
- [ ] I don't see Bloody Death being cast, but maybe I'm missing it. I'm not seeing either the sound effect or visual effect.
- [ ] Yauj and Vem still attack the player while they're eating
- [ ] They still turn to face the player while eating.

IntelligentQuantum (approved):

Issue #13145, opened 2022-09-26
[server] world server crash 0926


mpfans:
### Current Behaviour

crash log :

https://gist.github.com/mpfans/476f4fe9491916ff3b2db28bdf9b4d2d

### Expected Blizzlike Behaviour

none

### Steps to reproduce the problem

none

### AC rev. hash/commit

1f472bddd395073bf2a2d146ecb66489b36015fb

### Operating system

Ubuntu20.04

acidmanifesto:
Appears to be a proc issue or a spell issue.
```
[...]

Issue #6035, opened 2021-05-25
[Mage Talents] Master of Elements not restoring mana on critical strike


ghost:
#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [ ] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

Original report: https://github.com/chromiecraft/chromiecraft/issues/694

##### CURRENT BEHAVIOUR:

Succesful spell criticals do not always trigger Master of Elements (https://wowgaming.altervista.org/aowow/?spell=29076).
The following spells never trigger it:
Arcane Explosion - https://wowgaming.altervista.org/aowow/?spell=8439
Arcane Missiles -  https://wowgaming.altervista.org/aowow/?spell=8416
Blizzard - https://wowgaming.altervista.org/aowow/?spell=8427

And for other spells it doesn't always trigger:
https://i.imgur.com/cXyhNxH.png
https://i.imgur.com/l9yQJl0.png

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Every spell should be able to benefit from Master of Elements as of patch 3.0.2: https://wowwiki-archive.fandom.com/wiki/Patch_3.0.2

##### STEPS TO REPRODUCE THE PROBLEM:

1. Cast a spell
2. Land a critical
3. ???
4. No mana

##### EXTRA NOTES:

Master of Elements: https://wowgaming.altervista.org/aowow/?spell=29076

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/e50754d804987ec93792b6a4d113913ab018189b

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-server-auto-shutdown](https://github.com/azerothcore/mod-server-auto-shutdown)
[...]

PR #13154, opened 2022-09-27
fix(Core/Spells): Crash

Nefertumm:
## Changes Proposed:
-  Master of elements shouldn't proc from periodic spells.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13145

## Tests Performed:

- Not tested.

## How to Test the Changes:

1. ?
2.
3.

- [ ] I don't know if this is the right approach.

### DIFF
--- a/src/server/scripts/Spells/spell_mage.cpp
+++ b/src/server/scripts/Spells/spell_mage.cpp
@@ -825,6 +825,9 @@ class spell_mage_master_of_elements : public AuraScript
     {
         PreventDefaultAction();
 
+        if (ticksModifier == 0)
+            return;
+
         int32 mana = int32(_spellInfo->CalcPowerCost(GetTarget(), eventInfo.GetSchoolMask()) / ticksModifier);
         mana = CalculatePct(mana, aurEff->GetAmount());
 
--- a/src/server/scripts/Spells/spell_mage.cpp
+++ b/src/server/scripts/Spells/spell_mage.cpp
@@ -791,44 +791,20 @@ class spell_mage_master_of_elements : public AuraScript
     {
         _spellInfo = eventInfo.GetSpellInfo();
         if (!_spellInfo || !eventInfo.GetActor() || !eventInfo.GetActionTarget())
-        {
             return false;
-        }
 
-        bool selectCaster = false;
         // Triggered spells cost no mana so we need triggering spellInfo
         if (SpellInfo const* triggeredByAuraSpellInfo = eventInfo.GetTriggerAuraSpell())
-        {
             _spellInfo = triggeredByAuraSpellInfo;
-            selectCaster = true;
-        }
 
-        // If spell is periodic, mana amount is divided by tick number
-        if (eventInfo.GetTriggerAuraEffectIndex() >= EFFECT_0)
-        {
-            if (Unit* caster = GetCaster())
-            {
-                if (Unit* target = (selectCaster ? eventInfo.GetActor() : eventInfo.GetActionTarget()))
-                {
-                    if (AuraEffect const* aurEff = target->GetAuraEffect(_spellInfo->Id, eventInfo.GetTriggerAuraEffectIndex(), caster->GetGUID()))
-                    {
-                        ticksModifier = aurEff->GetTotalTicks();
-                    }
-                }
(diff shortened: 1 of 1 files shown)


mpfans:
@UltraNix  Can you check it..tyvm

Nefertumm:
I mean, it won't crash anymore but I want to know what the behavior should be in that situation.

Ovahlord:
pretty sure the proc shouldn't kick in from periodic spells at all. 

Proc flag 0x00010000, chance = 100, charges - 0
=================================================
  16 Deal Harmful Spell

only applies to direct damaging spells. so this whole aura based stuff can be removed alltogether.

src/server/scripts/Spells/spell_mage.cpp
```diff
@@ -825,6 +825,9 @@ class spell_mage_master_of_elements : public AuraScript
     {
         PreventDefaultAction();
 
+        if (ticksModifier == 0)
+            return;
+
```
acidmanifesto, line 830: Would it not be better to set it to default to 1 if it hits zero since zero makes it all crashy?
If we did a return wouldnt that cancel out of the whole spell entirely?
Since it is uint8, do we know if negative values makes it crash as well? Or if Negative Values are even possible?
```suggestion
if (ticksModifier == 0)
{
    ticksModifier = 1;
}
```

Nefertumm:
I'm waiting for confirmations about it on wotlk ptr.

Gultask:
Confirmed on WotLK Classic PTR. Periodic criticals from Glyph of Living Bomb do not proc Master of Elements.
Furthermore, when an AoE attack crits 2 different targets at once, only once proc of Master of Elements is seen.

Gultask:
Talent Burnout is interfering with Master of Elements, both can't proc at the same time.

Living Bomb periodical criticals can't proc Master of Elements.
Blizzard & the explosion from Living Bomb is behaving incorrectly with Master of Elements.

Since this is a crashfix I'd say it's best to merge ASAP and I'll open an issue about Master of Elements with more details.

Nyeriah (approved):

acidmanifesto:
> Talent Burnout is interfering with Master of Elements, both can't proc at the same time.
> 
> Living Bomb periodical criticals can't proc Master of Elements. Blizzard & the explosion from Living Bomb is behaving incorrectly with Master of Elements.
> 
> Since this is a crashfix I'd say it's best to merge ASAP and I'll open an issue about Master of Elements with more details.

right now im a bit confused. I will need some clarification. 
This fixes a crash and regressed specific spells?

Gultask:
You'd need to ask Nefertumm if he wants to merge it or not.

I don't actually know if the regression comes from this PR or not.

acidmanifesto:
@Nefertumm need to know if spell regression is else where or not before merging.

Nefertumm:
Idk, needs double check, but it breaks the mana return from aoe spells like blizzard (where it returns the whole 10% mana return instead of 10% / amount of ticks).

UltraNix:
@Nefertumm What about #6035? Arcane Missiles and Blizzard are periodic spells.

Nefertumm:
Yes, indeed, that's the reason I didn't merge this. Although right now is not working properly since every crit from Blizzard procs Master of Elements, and it should only proc once per cast. Feel free to supersede this PR with a better solution for the crash and the spell itself

UltraNix:
@Nefertumm 
`Although right now is not working properly since every crit from Blizzard procs Master of Elements, and it should only proc once per cast.` Source of it? Because then if it's once per cast, then Master of Elements should refund 10% of Blizzard cast, not divided by ticks number.

UltraNix:
https://wowpedia.fandom.com/wiki/Master_of_Elements
`For channeled spells that deal damage in ticks such as  [Arcane Missiles] and  [Blizzard], the mana returned is based on the tick that critically hit, and not the entire spell. For example, Arcane Missiles in its higher levels deals damage in 5 ticks: it will only return 15/30% of 1/5th of the cost of Arcane Missiles and not its total.` I see no info here about proc once per cast.

UltraNix:
Superseded by #13484

Issue #11183, opened 2022-03-28
[BWL] Razorgore the Untamed should not hit target that is Conflagrated 


temperrr:
### Current Behaviour

Razorgore the Untamed uses conflagrate and continues to hit the target during the effect

### Expected Blizzlike Behaviour

Razorgore the Untamed uses conflagrate on current target then during this affect hit the next target on top of his threat list, 

### Source
![image](https://user-images.githubusercontent.com/20357406/160346142-9146e1bb-562a-4096-bbe5-44a3b5ccfc98.png)
https://wowpedia.fandom.com/wiki/Razorgore_the_Untamed

### Steps to reproduce the problem

1. go to razorgore (.tele bwl)
2.  do first phase with 2 chars
3.  get conflagrated
4.  keep getting ganked

### AC rev. hash/commit

9e7d1e2a3af7

### Operating system

Manjaro x64

PR #13161, opened 2022-09-27
fix(Scripts/BlackwingLair) Razorgore should not reset with only one raid member alive that is conflagrated

SaxxonPike:
## Changes Proposed:
- Make it so that Razorgore's conflagrate functions like Drakkisath's conflagrate (which does this mechanic properly): threat is only temporarily reduced to zero
- Allow Razorgore to still attack conflagrated players if they are the only players left alive (as they will have 0 threat but still be eligible)

## Issues Addressed:
- This is a regression introduced with #11183 (although that partially addressed the issue, it was not entirely the correct fix)

## SOURCE:
- This shows what is supposed to happen when a solo player (only one on the threat table) is conflagrated: https://youtu.be/5FxU1dLJUs4?t=293 (starts 4:53)
- Notice how Razorgore continues attacking the conflagrated player because there's no one else on the threat table.

## Tests Performed:
- We built and ran AC, and entered BWL with 2 players to verify that targets switched appropriately during Conflagrate.
- We also ran it with 1 player to make sure the encounter did not reset due to Razorgore lacking eligible targets.

## How to Test the Changes:
Testing solo:
1. `.tele blackwinglair` and enter, alone, with no pets
2. Begin the encounter by killing the weak mobs on the left- but leave Razorgore loose
3. Observe that the fight does not reset when Razorgore conflags the player.
Testing intended:
1. `.tele blackwinglair` and enter with any number of friends
2. Begin the encounter by killing the weak mobs on the left- leave Razorgore loose
3. Observe that Razorgore's target changes when the tank is conflagrated.

- Not at this time.

### DIFF
--- a/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_razorgore.cpp
+++ b/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackwingLair/boss_razorgore.cpp
@@ -72,13 +72,13 @@ enum EVENTS
 class boss_razorgore : public CreatureScript
 {
 public:
-    boss_razorgore() : CreatureScript("boss_razorgore") {
-        _conflagrateThreat = 0.0f;
-    }
+    boss_razorgore() : CreatureScript("boss_razorgore") { }
 
     struct boss_razorgoreAI : public BossAI
     {
-        boss_razorgoreAI(Creature* creature) : BossAI(creature, DATA_RAZORGORE_THE_UNTAMED) { }
+        boss_razorgoreAI(Creature* creature) : BossAI(creature, DATA_RAZORGORE_THE_UNTAMED) {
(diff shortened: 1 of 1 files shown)


SaxxonPike:
I think this is as a result of the Threatmgr changes since our last merge, I will repair.

Gultask:
Sorry for the late reply. There is a lack of testers at the moment. Build is still failing for me.

Gultask:
Generally this is a safety measure to make it possible to solo. In WotLK videos I've never seen a boss with this mechanic that actually resets when using it. So I'd say this PR would indeed correct an issue.

Nefertumm:
Superseded by https://github.com/azerothcore/azerothcore-wotlk/pull/13782 as it fixes other bosses too. Thank you for your PR!

PR #13166, opened 2022-09-27
PlayerUpdates.cpp: Use blizz-like progressive skillup chance.

SaxxonPike:
## Changes Proposed:
- Make it so that the success chance of skilling up with a profession is on a linear slope rather than stepped.

## Issues Addressed:
- Recipes are color-coded to indicate roughly how successful the player will be in gaining a skill point from the craft. Currently, each color is represented directly by a percentage chance thoroughly: the default for all yellow and green skillups is 75% and 25% respectively. However, under the hood, it's actually a linear grade. This means just getting into yellow will come with a very high chance of success and the tail end of green will come with a very low chance of success. This is the way it works on both live and classic.

## SOURCE:
- In each of these videos, observe that "fresh" yellows have a better chance than yellows that are about to turn green.
- Alchemy 1-300 timelapse: https://www.youtube.com/watch?v=ywFRjzhv8WQ
- Engineering 1-300 timelapse: https://www.youtube.com/watch?v=-OXnztSm-3A

## Tests Performed:
- Tested using Leatherworking, First Aid, Tailoring
- Also tested with Runeforging, which is only really a profession in name

## How to Test the Changes:
1. First aid is probably the best candidate and where most players will craft the most yellow/green.
2. `.learn 3259` to get Apprentice First Aid.
3. `.additem 2589 200` to give yourself a bunch of Linen Cloth.
4. Craft using all of it. Notice how when the recipe first turns yellow it seems to go well, but at higher skills in yellow, it can take longer. The last point of green is most nasty.

- What happens with the `SkillChance.Green` and `SkillChance.Yellow` values in the config? They will go unused when this patch is merged in its current state.
- In the future, it may be nice to be able to customize the curve of the falloff. But for now, it should be linear.

### DIFF
--- a/src/server/game/Entities/Player/PlayerUpdates.cpp
+++ b/src/server/game/Entities/Player/PlayerUpdates.cpp
@@ -720,8 +720,7 @@ int ProgressiveSkillGainChance(int32 SkillValue, int32 LowChance, int32 HighChan
     return baseResult + (chanceDelta * progress * Coeff / cutoffDelta);
 }
 
-inline int SkillGainChance(uint32 SkillValue, uint32 GrayLevel,
-                           uint32 GreenLevel, uint32 YellowLevel)
+inline int SkillGainChance(uint32 SkillValue, uint32 GrayLevel, uint32 YellowLevel)
 {
     // Use a linear function for skillups instead of tiers.
     // This has the side effect of ignoring values for yellow/green.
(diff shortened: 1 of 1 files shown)


ZhengPeiRu21:
Here is another source confirm this is correct behavior: https://wowpedia.fandom.com/wiki/Profession

heyitsbench:
> Here is another source confirm this is correct behavior

[Same Wowpedia link from WotLK era.](https://wowpedia.fandom.com/wiki/Profession?oldid=2333782)

SaxxonPike:
This is failing the build due to the unused variable; separate commit to follow for recordkeeping sake.

acidmanifesto:
> This is failing the build due to the unused variable; separate commit to follow for recordkeeping sake.

Is there more commits to be added to this pr? At it's current state it seems a bit non-sensible and not working.

SaxxonPike:
> Is there more commits to be added to this pr? At it's current state it seems a bit non-sensible and not working.
Can you elaborate on "non-sensible" and "not working"?

We spent about two weeks to try and find issues or strangeness with this patch applied on our deployment, which was active and running during that time.

Is there anything about the algorithm we can clear up? I think the simplest way to explain is "a linear profession skill-up success rate from yellow to gray".

heyitsbench:
Gave this a test with linen bandages and elixirs of minor defense. For both, it took only one craft to skill up until I hit green, then the skill ups felt like they took exponentially more crafts with each level. Obviously could just be me since that's how randomness works, but that's what I experienced.

Maelthyrr:
I've spent some time checking the behavior on Woltk Classic and it's absolutely not linear. The behavior explain is for engineering.

- For red state, the player will always up to 1 level for every craft.
- For the yellow state, it's very complex. Globally, on the first yellow craft, the player will level up to 1 level every time. But after few crafts, the chance to level up will depend it will depend on the complexity of the craft. The more resources the craft requires, the more resources you have, the more likely you are to increase your skill level.
- Same behavior for the green state. It depends of the craft's complexity.

Issue #5094, opened 2021-04-01
Wetlands Mottled Raptor and Highland Raptor Loottables Incorrect


ghost:
Wetlands Mottled Raptor and Highland Raptor Loottables Incorrect

## CURRENT BEHAVIOUR:

The raptors currently drop a full array of blues. If you farm there for 20-30 mins you will gain 50-100 blue items.

## EXPECTED BLIZZLIKE BEHAVIOUR:

Each raptor only drops a couple blues listed on sources below. It should be less drops, and less items.

## SOURCE:

https://classic.wowhead.com/npc=1020/mottled-raptor

https://classic.wowhead.com/npc=1015/highland-raptor

## STEPS TO REPRODUCE THE PROBLEM:

1. Go to Whelgar's Excavation in Wetlands
2. Farm Raptors for 5-10 mins.
3. Observe many blues.

warcrafter16:
Hi if this problem still isn't fixed can I try working on it? Im new to OS and would like to try.

PR #13204, opened 2022-10-02
Fix(DB/reference_loot_template): Adjusted raptor drops

warcrafter16:
## Changes Proposed:
-  Changed the loot reference template for the Wetlands Mottled Raptors.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5094

## Tests Performed:

- Test in-game by defeating and looting raptors.

## How to Test the Changes:

1. Go to Whelgar's Excavation in Wetlands
2. Farm Raptors for 5-10 mins.
3. Observe less and correct blues being dropped.

- [ ] The issue with the Highland Raptor still exists. The Mottled and Highland raptors use the same formats but should be split apart since they drop different loot.


data/sql/base/db_world/reference_loot_template.sql
```diff
@@ -13953,7 +13953,7 @@ INSERT INTO `reference_loot_template` (`Entry`, `Item`, `Reference`, `Chance`, `
 	(45002, 3180, 0, 10, 0, 1, 0, 1, 1, 'Flecked Raptor Scale'),
 	(45002, 3685, 0, 60, 0, 1, 0, 1, 1, 'Raptor Egg'),
 	(45002, 5635, 0, 8, 0, 1, 0, 1, 1, 'Sharp Claw'),
-	(45002, 24065, 24065, 5, 0, 1, 1, 1, 1, '(ReferenceTable)'),
+	(45002, 24063, 24063, 3, 0, 1, 1, 1, 1, '(ReferenceTable)'),
```
heyitsbench, line 13956: Do not edit the base SQL files.

warcrafter16: Ok then what should i do if it appears the problem lies in the SQL files?

Gultask: We follow this guide: https://www.azerothcore.org/wiki/sql-standards :)
In this case it should be used a DELETE + INSERT query instead. i.e. 
```
DELETE FROM `reference_loot_template` WHERE `entry`=45002 AND `item` IN (24065, 24063);
```

Gultask (commented):
You also need to change it to remove both items 24063 AND 24065 like:

```
WHERE `entry` = 45002 AND `item` IN (24063, 24065);
```

or else, when this query is ran twice, it'll run into this error:
image

Gultask:
It works 👍 

Do you have a source for this change? Like, does it come from Mangos or TC, or Wowhead?

warcrafter16:
The source for the change comes from wowhead. The links are on the issue page but I will also post them here:

https://classic.wowhead.com/npc=1020/mottled-raptor

https://classic.wowhead.com/npc=1015/highland-raptor

acidmanifesto:
> The source for the change comes from wowhead. The links are on the issue page but I will also post them here:
> 
> https://classic.wowhead.com/npc=1020/mottled-raptor
> 
> https://classic.wowhead.com/npc=1015/highland-raptor

is there another reference you can use besides wowhead? they use a form of dirt data mining from a addon that also collects from pservers with custom loot tables.

warcrafter16:
Unfortunatley I just used the wowhead links. I am unfamiliar with WoW since I have never played it so im not sure of sites that would have good information on this and just used what was given at the time. I looked for the other sites mentioned earlier but could not find them and the other sites I found seemed to have similar ways of obtaining since they had the same data as wowhead.

Gultask:
There is a semi-frequent issue with mobs dropping items with higher levels than they are. I don't know if it's blizzlike or not.

acidmanifesto:
> There is a semi-frequent issue with mobs dropping items with higher levels than they are. I don't know if it's blizzlike or not.

Blizz seems to be doing alot of unschedule maintence if we are referencing wotlk classic. At this point of insanity, we would aim close to bliz simular then blizz like/exact.

Without continual growth and progress, such words as improvement, achievement, and success have no meaning.

If you feel like this adds to improvement then let me know. If you feel like you are able to fine tune it some then go for it. If not let me know if this is as far as you can go currently.

Gultask:
Personally I see an improvement here. Or, at least, I don't see anything wrong with this PR unless we can prove one thing over another.

acidmanifesto (approved):

PR #13216, opened 2022-10-02
fix(DB/NPC): Zenn Foulhoof

grimgravy:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12512

## SOURCE:

Wotlk Classic: https://www.youtube.com/watch?v=ZerX5j4xgDY
## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1664735105497287000.sql
+++ b/data/sql/updates/pending_db_world/rev_1664735105497287000.sql
@@ -1,0 +1,10 @@
+--
+DELETE FROM `waypoint_data` WHERE `id`=215000;
+INSERT INTO `waypoint_data` (`id`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `delay`, `move_type`, `action`, `action_chance`, `wpguid`) VALUES
+(215000, 1, 9921.88, 738.288, 1314.69, 2.06284, 40000, 0, 0, 100, 0),
+(215000, 2, 9922.85, 737.091, 1314.9, 5.33838, 40000, 0, 0, 100, 0);
+
+DELETE FROM `creature_template_addon` WHERE  `entry`=2150;
+
+UPDATE `creature` SET `MovementType`='2' WHERE  `guid`=46393;
+UPDATE `creature_addon` SET `path_id`='215000' WHERE  `guid`=46393;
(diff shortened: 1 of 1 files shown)


Gultask:
Are these waypoints sniffed? Last I knew about Zenn is that he has a random pattern but it's **very** slow. Like, he does some movement, stops for 30-40s then starts again. I think he was the first creature I sniffed in TBCC, but I lost the packet.

Also, in Acore we use GUIDx10 for waypoints instead of GUIDx100 :)

acidmanifesto (changes requested):
source of way points?

grimgravy:
is a few seconds link here: https://www.youtube.com/watch?v=ZerX5j4xgDY

acidmanifesto:
With no sniff and being told to reference a few seconds of a video with blur spots on it, we are unable to confirm if this is a retail video or not. If the npc is a random wandering distance it needs to be observed for a good while while being sniffed. Please provide verifiable retail video or a sniff, or reference and link to other verifiable database such as TC or mangos.

acidmanifesto (changes requested):
https://github.com/azerothcore/azerothcore-wotlk/pull/13216#issuecomment-1406535835

Gultask:
I believe that it is a scripted random movement, set to wander every once in a while and then stop. I can't prove it because I never bothered to sniff it again and I lost the packet I had cause I forgot to rename it :p

Maelthyrr:
If it's not a path, why not use "wander_distance"?

grimgravy:
> If it's not a path, why not use "wander_distance"?

not used for this type of movement

Gultask:
It's not waypoints either, I can prove it.

This is a special case of very delayed random WPs. I can't get the values for now because I deleted the sniff packet, though :P

grimgravy:
> @grimgravy what's the status of this?

I will try to come up with a better movement for the npc. I need time to build

Gultask:
This is legitimately an oddity. I've never really seen one of these before.

This NPC has a really slow wander. But it IS a wander

Nyeriah:
Hello,

This PR has not received any updates in an extended amount of time, therefore we are closing it for the time being as it seems to have been abandoned. If you wish to continue developing this, please leave a message and we will reopen it. Thanks for your contributions.

Regards

PR #13308, opened 2022-10-04
feat(docker): implemented dbimport

Yehonal:
## Changes Proposed:
- Implemented dbimport with docker
- deprecated db_assembler
- Fixed deno scripts and integrated them with the CI


apps/db_assembler/README.md
```diff
@@ -1,5 +1,7 @@
 ## Description
 
+**ATTENTION:** this tool is not supported anymore. It has been replaced by the **dbimport** tools integrated in AC server sources
+
```
FrancescoBorzi, line 4: shall we add also a warning when someone tries to use it? not necessarily in this PR

apps/docker/docker-cmd.ts
```diff
@@ -3,7 +3,7 @@ import * as ink from "https://deno.land/x/ink/mod.ts";
 import {
   Input,
   Select,
-} from "https://deno.land/x/cliffy@v0.18.2/prompt/mod.ts";
+} from "https://deno.land/x/cliffy@v0.25.2/prompt/mod.ts";
```
FrancescoBorzi, line 6: is there a way to centralise the list of dependencies and their versions? so one can update them all at once

FrancescoBorzi (approved):

PR #13346, opened 2022-10-08
fix(DB/creature_addon): Add Frenzy Aura to Vekniss Soldier

Gultask:
## Changes Proposed:
-  Adds aura 25801 to creature_template_addon at entry 15229

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3810

## SOURCE:

Retail sniffs. It's not a spell cast.

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1. .go c id 15229
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1665246218923522400.sql
+++ b/data/sql/updates/pending_db_world/rev_1665246218923522400.sql
@@ -1,0 +1,2 @@
+--
+UPDATE `creature_template_addon` SET `auras` = '25801' WHERE (`entry` = 15229);


data/sql/updates/pending_db_world/rev_1665246218923522400.sql
```diff
@@ -0,0 +1,3 @@
+--
+UPDATE `creature_template_addon` SET `auras` = '25801' WHERE (`entry` = 15229);
+UPDATE `creature_addon` SET `auras` = '25801' WHERE `guid` BETWEEN 87901 AND 87906;
```
Kitzunu, line 3: Assuming that the spawned creature guid is linked to the entry in creature_template you do not need to update creature_addon

Gultask: It didn't work if I only updated creature_template_addon, I tried.

Kitzunu: then there is some otherlying problem, as creature_template_anything should be the base for all creatures and creature_anything acts as the override.

Nefertumm (approved):

data/sql/updates/pending_db_world/rev_1665246218923522400.sql
Gultask: I'll try to see what's going on

Gultask: A question, though, wouldn't an empty auras field be an override too? What do I place there so that there is no auras?

Kitzunu: Never thought of that, I suppose that's never the case?

PR #13363, opened 2022-10-09
fix(Scripts/TempleOfAhnQiraj): Don't submerge Ouro if it's already en…

Nefertumm:
…raged

## Changes Proposed:
-  Add an enrage check for its third phase.

## Issues Addressed:

- Closes -

## Tests Performed:

- Not tested.

## How to Test the Changes:

1. Go to Ouro, engage and damage it until 20%
2. It should never submerge after that.


Gultask:
I saw no regressions 👍

Nyeriah (approved):

Issue #13409, opened 2022-10-14
[MSVC][std:c++latest] Azerothcore-wotlk failed to build with msvc due to error C2280 on Windows 


shizhuoer:
### Current Behaviour

Hi all,
Azerothcore-wotlk failed to build with msvc due to error C2280 on Windows. Could you please help look at this issue?

**Error info:**
 F:\gitP\azerothcore\azerothcore-wotlk\src\server\game\Instances\InstanceScript.cpp(344,1): error C2280: 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::basic_string(std::nullptr_t)': attempting to reference a deleted function [F:\gitP\azerothcore\azerothcore-wotlk\build_amd64\src\ser
       ver\game\game.vcxproj]

**log**
[build.log](https://github.com/azerothcore/azerothcore-wotlk/files/9782948/build.log)

**VS version: VS2022**

### Expected Blizzlike Behaviour

For Azerothcore_wotlk should be run pass on the windows.

### Source

No response

### Steps to reproduce the problem

1.git clone https://github.com/azerothcore/azerothcore-wotlk F:\gitP\azerothcore\azerothcore-wotlk
2.cd F:\gitP\azerothcore\azerothcore-wotlk
3 mkdir build_amd64 & cd build_amd64
4.set BOOST_ROOT=F:\tools\boost_1_77_0\x64
5.set MYSQL_ROOT=F:\tools\mysql_5_7_35\mysql-5.7.35-winx64
6.set OPENSSL_ROOT=F:\tools\OpenSSL\64
7.cmake -G "Visual Studio 17 2022" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DCMAKE_BUILD_TYPE=Release -DTOOLS=ON -DBUILD_TESTING=ON ..
8.set VSCMD_SKIP_SENDTELEMETRY=1 & "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat" -host_arch=amd64 -arch=amd64 & set _CL_= /std:c++latest
9.msbuild /m /p:Platform=x64 /p:Configuration=Release AzerothCore.sln /t:Rebuild

### Extra Notes

No response

### AC rev. hash/commit

None

### Operating system

Windows 10 x64

Kitzunu:
[...]

PR #13412, opened 2022-10-14
refactor(Core/Instance): make SetBossState void

Nefertumm:
## Changes Proposed:
-  Make it void instead of returning an std::string, it makes no sense and there are no uses for that currently on the core.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13409

## Tests Performed:

- Build on windows.

## How to Test the Changes:

1. Compile.


Kitzunu (approved):

Maelthyrr:
@Nefertumm Seems that break gcc compile.

Nefertumm:
?????

Kitzunu:
Why ?, no response after the comment. This breaks GCC

acidmanifesto:
Applying run-build label to ensure we have no issues.

Maelthyrr:
Tested in AQ40/20, Nax, and ICC and no crash, no compile errors, no worldserver errors

PR #13583, opened 2022-10-29
Fix compiling with gcc 12

tnixeu:
I noticed that compiling with gcc 12.1 fails.

## Changes Proposed:
-  remove `<T>`from a constructor
-  rename header guard because in conflicts with openssl header guard

## Tests Performed:
-  compile with gcc 12

## How to Test the Changes:

1. compile with gcc 12

I saw that in the openssl 3.0 PRs the header is also renamed. I was not sure if I should keep it in the PR or not. I kept it for now, because without it can not compile on gcc 12.

### DIFF
--- a/src/common/Cryptography/OpenSSLCrypto.h
+++ b/src/common/Cryptography/OpenSSLCrypto.h
@@ -15,8 +15,8 @@
  * with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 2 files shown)


Winfidonarleyan (approved):

FrancescoBorzi:
@Winfidonarleyan what's the status of this PR? ready to be merged? please don't forget to update the labels

PR #13704, opened 2022-11-05
fix(Scripts/Commands): Fix assertion error on mmaps path command

MikaMauger:
## Changes Proposed:
Fixes this error when using `.mmap path` with server running on debug mode (maybe that's just a specific combination of MSVC version and Windows).
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\include\optional(359) : Asserti
on failed: Cannot access value of empty optional
```

### DIFF
--- a/src/server/scripts/Commands/cs_mmaps.cpp
+++ b/src/server/scripts/Commands/cs_mmaps.cpp
@@ -61,7 +61,7 @@ public:
         return commandTable;
     }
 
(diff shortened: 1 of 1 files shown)


Kitzunu:
I'd like you to test on VS2022 as 19 isn't really supported anymore

Nefertumm (approved):

PR #13706, opened 2022-11-05
fix(Core/Unit): Exclude Sword Specialization and Hack and Slash from …

Nyeriah:
…the extra attack range check

- Exclude Hack and Slash (Rogue) and Sword Specialization (Warrior) from the extra attack (thrash) distance check
- This allows spells such as Whirlwind to proc it

How to test:
- .cheat power
- .cheat cooldown
- learn the sword spec talent (warrior)
- spam whirlwind, check you get white hits (easy to miss so pay attention)

Tested ingame

### DIFF
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -717,6 +717,12 @@ enum MeleeHitOutcome
     MELEE_HIT_GLANCING, MELEE_HIT_CRIT, MELEE_HIT_CRUSHING, MELEE_HIT_NORMAL
 };
 
+enum ExtraAttackSpells
+{
+    SPELL_SWORD_SPECIALIZATION   = 16459,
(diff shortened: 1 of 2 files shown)


Kitzunu (approved):

src/server/game/Entities/Unit/Unit.h
```diff
@@ -717,6 +717,12 @@ enum MeleeHitOutcome
     MELEE_HIT_GLANCING, MELEE_HIT_CRIT, MELEE_HIT_CRUSHING, MELEE_HIT_NORMAL
 };
 
+enum ExtraAttackSpells
```
Kitzunu, line 720: enum class? :)

Gultask:
It works. Tested both talents before and after. Using Whirlwind and Fan of Knives at the maximum possible range.

PR #13720, opened 2022-11-06
fix(Core/Loot): Fix quest items looting system

MikaMauger:
## Changes Proposed:
Fixes quest items distribution as described in this [issue](https://github.com/TrinityCore/TrinityCore/issues/22829). Problem applies to AC aswell.

Currently, quest items are seen by everyone, even in round robin or group mode. The first player who click on corpse get the item. This is frustrating for ranged players, and comes problematic for pickup groups (hello ninja looters).

The new rules are:
- in every mode, "non free for all" quest items are round robin
- if the round robin player can't loot quest item (he has not the quest or he has completed it), it will find the first available player in the group (can be changed to something else if someone got Blizz rules for this)
- "free for all" quest items won't change, everybody is able to see and loot it.

On a side effect, it will fix master loot by letting grey and white items round robin (green items is the minimum threshold for master loot).

## Issues Addressed:

I didn't find any issue on this in AC or CC.

## SOURCE:

> Quest loot is on a separate loot distribution and there are two common modes for this to occur. 1) In quests where the party is to collect one of a specific item from a specific mob (Collect the head of (mob name here)) then regardless of loot setting, the mob will indicate lootable by every party member with the quest (and without the object). Each member may loot the quest item off the corpse. The other loot on the corpse is subject to the party loot setting. 2) In other quests which require collecting several items of a type (harpy feathers for example), the quest items will drop on some mobs and be looted in the normal way, except that if the current looter can't collect the items (not on quest, already has full set) then the quest loot is FFA for those members who can still pick up the quest items. Other items on the corpse are governed by the party loot setting. On some quests the quest loot is always free-for-all and this may be the above mechanic gone haywire. Some quests where the party is collecting one item will use the named mob mechanism (1) (the [Shaman Voodoo Charm](https://wowwiki-archive.fandom.com/wiki/Shaman_Voodoo_Charm) from [The Sleeping Druid](https://wowwiki-archive.fandom.com/wiki/Quest:The_Sleeping_Druid) quest, for example, where all party members can get the charm off the same shaman.)

[Source](https://wowwiki-archive.fandom.com/wiki/Loot)

## Tests Performed:

A lot of combinations of group looting parameters, and quest with ffa and non ffa items.
Some tests around master loot and also case like [this](https://github.com/TrinityCore/TrinityCore/issues/24230).

## How to Test the Changes:

See tests performed. Test different combinations.

If you kill multiple monsters and aoe, you can have more quest items that you need when you start looting them (because calculation is done on monster kill). Your extra quest items won't go to the group. Again if someone knows the rules for this in order to be more "blizzlike", I would be happy to improve this PR.

### DIFF
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -404,7 +404,7 @@ LootItem::LootItem(LootStoreItem const& li)
 }
 
 // Basic checks for player/item compatibility - if false no chance to see the item in the loot
-bool LootItem::AllowedForPlayer(Player const *player, ObjectGuid source) const
+bool LootItem::AllowedForPlayer(Player const* player, ObjectGuid source) const
 {
     ItemTemplate const* pProto = sObjectMgr->GetItemTemplate(itemid);
     if (!pProto)
@@ -658,9 +658,9 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
     if (items.size() == MAX_NR_LOOT_ITEMS)
         return nullptr;
 
-    auto* ql = new QuestItemList();
+    QuestItemList* ql = new QuestItemList();
 
-    Player *lootOwner = (roundRobinPlayer) ? ObjectAccessor::FindPlayer(roundRobinPlayer) : player;
+    Player* lootOwner = (roundRobinPlayer) ? ObjectAccessor::FindPlayer(roundRobinPlayer) : player;
 
     for (uint8 i = 0; i < quest_items.size(); ++i)
     {
@@ -682,7 +682,8 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
         ql->push_back(QuestItem(i));
         ++unlootedCount;
 
-        if (!item.freeforall) {
+        if (!item.freeforall)
+        {
             item.is_blocked = true;
         }
 
@@ -701,7 +702,7 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
 
 QuestItemList* Loot::FillNonQuestNonFFAConditionalLoot(Player* player)
 {
-    auto* ql = new QuestItemList();
+    QuestItemList* ql = new QuestItemList();
 
     for (uint8 i = 0; i < items.size(); ++i)
     {
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -404,7 +404,7 @@ LootItem::LootItem(LootStoreItem const& li)
 }
 
 // Basic checks for player/item compatibility - if false no chance to see the item in the loot
-bool LootItem::AllowedForPlayer(Player const* player, bool isGivenByMasterLooter /*= false*/, bool allowQuestLoot /*= true*/, ObjectGuid source) const
+bool LootItem::AllowedForPlayer(Player const *player, ObjectGuid source) const
 {
     ItemTemplate const* pProto = sObjectMgr->GetItemTemplate(itemid);
     if (!pProto)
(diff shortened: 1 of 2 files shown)


src/server/game/Loot/LootMgr.cpp
```diff
@@ -404,7 +404,7 @@ LootItem::LootItem(LootStoreItem const& li)
 }
 
 // Basic checks for player/item compatibility - if false no chance to see the item in the loot
-bool LootItem::AllowedForPlayer(Player const* player, bool isGivenByMasterLooter /*= false*/, bool allowQuestLoot /*= true*/, ObjectGuid source) const
+bool LootItem::AllowedForPlayer(Player const *player, ObjectGuid source) const
```
Kitzunu, line 407:
```suggestion
bool LootItem::AllowedForPlayer(Player const* player, ObjectGuid source) const
```
always this way

src/server/game/Loot/LootMgr.cpp
```diff
@@ -658,28 +658,36 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
     if (items.size() == MAX_NR_LOOT_ITEMS)
         return nullptr;
 
-    QuestItemList* ql = new QuestItemList();
+    auto* ql = new QuestItemList();
+
+    Player *lootOwner = (roundRobinPlayer) ? ObjectAccessor::FindPlayer(roundRobinPlayer) : player;
```
Kitzunu, line 663: same here

src/server/game/Loot/LootMgr.cpp
```diff
@@ -658,28 +658,36 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
     if (items.size() == MAX_NR_LOOT_ITEMS)
         return nullptr;
 
-    QuestItemList* ql = new QuestItemList();
+    auto* ql = new QuestItemList();
```
Kitzunu, line 661: I'd prefer to refrain to use auto in this case, to easier know what this variable is pointing to.

src/server/game/Loot/LootMgr.cpp
```diff
@@ -658,28 +658,36 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
...
+        {
+            continue;
+        }
 
-            if (items.size() + ql->size() == MAX_NR_LOOT_ITEMS)
-                break;
+        ql->push_back(QuestItem(i));
+        ++unlootedCount;
+
+        if (!item.freeforall) {
```
Kitzunu, line 685: move bracket to newline

src/server/game/Loot/LootMgr.cpp
```diff
@@ -693,12 +701,13 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
 
 QuestItemList* Loot::FillNonQuestNonFFAConditionalLoot(Player* player)
 {
-    QuestItemList* ql = new QuestItemList();
+    auto* ql = new QuestItemList();
```
Kitzunu, line 704: same as above

src/server/game/Loot/LootMgr.h
```diff
@@ -176,8 +176,7 @@ struct LootItem
     LootItem() = default;
 
     // Basic checks for player/item compatibility - if false no chance to see the item in the loot
-    bool AllowedForPlayer(Player const* player, bool isGivenByMasterLooter = false, bool allowQuestLoot = true, ObjectGuid source = ObjectGuid::Empty) const;
-    bool AllowedForPlayer(Player const* player, ObjectGuid source) { return AllowedForPlayer(player, false, true, source); };
+    bool AllowedForPlayer(Player const *player, ObjectGuid source) const;
```
Kitzunu, line 179: same as above

src/server/game/Loot/LootMgr.cpp
Kitzunu: For more code standards you can read here https://www.azerothcore.org/wiki/cpp-code-standards

Kitzunu (changes requested):
Thanks for the contribution. I know it is still draft, but would still like to point out some things so that you don't need to change more stuff later :D

Valdifer:
Tested it in Zul'gurub with the quest 8201 and works, tested all loot modes

Nyeriah (approved):

Nyeriah:
Thanks for the PR :+1:

Nyeriah:
@MikaMauger Please take a look into this: https://github.com/chromiecraft/chromiecraft/issues/4542

Master looters can't see quest items to give them out anymore

MikaMauger:
I will.

Issue #13855, opened 2022-11-18
[server] world server crash 1118


mpfans:
### Current Behaviour

crash log :

https://gist.github.com/mpfans/7afa5976717ffd6b09843f949b0a224a

### Expected Blizzlike Behaviour

none

### Steps to reproduce the problem

none

### AC rev. hash/commit

2ae3f40847a6c439c6c1ca1c2361b52979e06f5c

### Operating system

Ubuntu20.04

kissingers:
the three are same problems cased by 368a363522ac6d2ec3410fd01d07b77ea462d8ec
the temp fix replace the SAI, look this link
https://github.com/azerothcore/azerothcore-wotlk/issues/13827

mpfans:
Temp fixed:

```UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 17058;```

Issue #13827, opened 2022-11-14
[Server] world carsh cased by patch "Dreghood Brute dont become friendly after the death of the taskmaster"


kissingers:
### Current Behaviour

Update
Now I can 100% replay the crash.

.go xyz -445.000153 4747.521 19.9338989
First kill the two creature 16937
Then kill the creature 17058
The SAI can't do action for creature 16937, because it is died. server crash.

![44a784422933f10ec4b59ff4246b8e66_202736262-5067a4d9-a899-41ce-a2ab-07ccb7480c96](https://user-images.githubusercontent.com/33601959/202738847-5b2ef7f5-a6bd-4179-a3be-5ccf2a3ab3a5.png)

Here is gdb.txt log, I guest it about some creature movement. I dont know how to collect log, only have gdb log below.

[gdb.zip](https://github.com/azerothcore/azerothcore-wotlk/files/10005258/gdb.zip)

update 1116 carsh log and dberror log
[gdb.txt](https://github.com/azerothcore/azerothcore-wotlk/files/10018846/gdb.txt)
[DBErrors.log](https://github.com/azerothcore/azerothcore-wotlk/files/10018855/DBErrors.log)

### Expected Blizzlike Behaviour

N/A

### Steps to reproduce the problem

N/A

### AC rev. hash/commit

AzerothCore rev. 5cf7d3d198f6+ 2022-11-13 18:41:10 -0300 (HEAD branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1n  15 Mar 2022 (library: OpenSSL 1.1.1n  15 Mar 2022)
Using Boost version: 1.74.0
Using MySQL version: 100515
Using CMake version: 3.18.4
Compiled on: Linux 5.15.64-1-pve
Worldserver listening connections on port 54345
Realmlist (Realm Id: 1) configured in port 54345
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /root/wow/bin/maps. Total size: 291014339 bytes
vmaps directory located in /root/wow/bin/vmaps. Total size: 658130553 bytes
mmaps directory located in /root/wow/bin/mmaps. Total size: 2320900760 bytes
Using zhCN DBC Locale as default. All available DBC locales: zhCN 
Using World DB: ACDB 335.8-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
[...]

PR #13862, opened 2022-11-18
fix(Core): Crash on random movement SAI

Nefertumm:
## Changes Proposed:
-  Add an alive check for the SAI's owner in case no targets were found.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13827
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13855

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

1. .go xyz -445.000153 4747.521 19.9338989
2. Kill both Dreghood brute and then kill the Illidari
3. No crash.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -1950,7 +1950,7 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                 }
             }
 
-            if (!foundTarget && me && IsCreature(me))
+            if (!foundTarget && me && IsCreature(me) && me->IsAlive())
             {
                 if (e.action.moveRandom.distance)
                     me->GetMotionMaster()->MoveRandom(float(e.action.moveRandom.distance));


UltraNix:
The problem is not core, but the SAI script.

kissingers:
> The problem is not core, but the SAI script.

but the cpp fix can prevent the same type SAI problem, it's more good than just fix SAI.

UltraNix:
But you can't prevent others doing wrong scripts everywhere in code.

kissingers:
> But you can't prevent others doing wrong scripts everywhere in code.

Yes it is not impossible to prevent all problems, but more reasonable to prevent such errors programmatically and strengthen the server robustness. Also there is no way saying that SAI cannot be used this way.

Nefertumm:
This current behavior on SAI is kinda bad tbh, even I thought about removing the whole part, since it doesn't make any sense to me. Also, if this was the intended behavior, it should be specifically written on that action explanation.

Nyeriah (approved):

PR #14010, opened 2022-12-01
fix(DB/loot): Improve Mazthoril recipe drop rates

neifion-00000000:
## Changes Proposed:
-  Increases the drop rate chance of 2 recipes, lowers the drop rate chance of 1 recipe to match wowhead's* values

## SOURCE:
*I used Wowhead values to save time, I also as always sniffed the loot manually and got every item to drop, usually multiple times, but it would take me at least 10hours to come up with better numbers and their numbers seemed "pretty close enough" for this and our numbers seemed wildly inaccurate

## Tests Performed:

- Untested but I don't see why it wouldn't work as is, its just a loot change and the SQL is written by HeidiSQL
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1669870365832502100.sql
+++ b/data/sql/updates/pending_db_world/rev_1669870365832502100.sql
@@ -1,5 +1,5 @@
 --
 -- Improve Mazthoril Recipe Drop Rates
-UPDATE `world`.`creature_loot_template` SET `Chance`=6 WHERE  `Entry`=7437 AND `Item`=14493 AND `Reference`=0 AND `GroupId`=0;
-UPDATE `world`.`creature_loot_template` SET `Chance`=2 WHERE  `Entry`=7437 AND `Item`=16054 AND `Reference`=0 AND `GroupId`=0;
-UPDATE `world`.`creature_loot_template` SET `Chance`=10 WHERE  `Entry`=7437 AND `Item`=13497 AND `Reference`=0 AND `GroupId`=0;
+UPDATE `creature_loot_template` SET `Chance`=6 WHERE  `Entry`=7437 AND `Item`=14493 AND `Reference`=0 AND `GroupId`=0;
+UPDATE `creature_loot_template` SET `Chance`=2 WHERE  `Entry`=7437 AND `Item`=16054 AND `Reference`=0 AND `GroupId`=0;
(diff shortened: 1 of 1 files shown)


neifion-00000000:
if the methodology here is good and approved there are a couple other ones I'd like to do as well that are smaller changes but I'm sure ours are wrong for 

while I don't like wowhead as a source obviously, after I farm a few of the drops I can usually at least tell if the wowhead number is closer to correct than our current one, and quite simply I'll get more done--as I'm currently farming out all the engineering patterns then tailoring patterns I can handle I should have a good idea to make informed improvements like this

data/sql/updates/pending_db_world/rev_1669870365832502100.sql
```diff
@@ -0,0 +1,5 @@
+--
+-- Improve Mazthoril Recipe Drop Rates
+UPDATE `creature_loot_template` SET `Chance`=6 WHERE  `Entry`=7437 AND `Item`=14493 AND `Reference`=0 AND `GroupId`=0;
+UPDATE `creature_loot_template` SET `Chance`=2 WHERE  `Entry`=7437 AND `Item`=16054 AND `Reference`=0 AND `GroupId`=0;
+UPDATE `creature_loot_template` SET `Chance`=10 WHERE  `Entry`=7437 AND `Item`=13497 AND `Reference`=0 AND `GroupId`=0;
```
ZhengPeiRu21, line 5:
```suggestion
UPDATE `creature_loot_template` SET `Chance`=6 WHERE `Entry`=7437 AND `Item`=14493 AND `Reference`=0 AND `GroupId`=0;
UPDATE `creature_loot_template` SET `Chance`=2 WHERE `Entry`=7437 AND `Item`=16054 AND `Reference`=0 AND `GroupId`=0;
UPDATE `creature_loot_template` SET `Chance`=10 WHERE `Entry`=7437 AND `Item`=13497 AND `Reference`=0 AND `GroupId`=0;
```
Spacing fix. ^_^

ZhengPeiRu21 (dismissed):
I think the change is good and using WoWHead numbers is okay in this case since it is an obvious improvement from what we have right now. I make one small suggestion to fix some spacing in your SQL.

data/sql/updates/pending_db_world/rev_1669870365832502100.sql
neifion-00000000: word I'll try to look out for that, it was done by HeidiSQL automatically, I'll see if I can mess with the default outputs tho for the future.

Gultask (approved):

PR #14043, opened 2022-12-03
fix(Core/Players): Fixed sending power regen update to nearby players.

### LINKED ISSUE #12694: Group & Target mana doesn't always update
### Current Behaviour

While in a group, mana value seen by other players won't update unless it's triggered by something else. That something else include things like health regenerating, getting flagged as afk, changing targets etc

Mana is supposed to update for other players as frequently as health does as long as it's not full

### Source

_No response_

1. Create two characters, having one of them be able to cast heal spells really helps
2. Create a group with these characters
3. Cast heals on yourself on one of them and wait for the mana to naturally regenerate (level 1 makes this easy)
4. Check the mana value seen by the other player, it shouldn't be full

### Extra Notes

To hopefully explain the issue a little better, this is what the second character sees before the first one uses up their mana:
image

This is what the second character sees after the first character has used up most of their mana **and** regenerated to full:
image

This is only a huge issue for me because it makes it impossible to track how much mana a healer has before a pull unless they change target frequently.

Updates to the player in other ways, such as being flagged as afk, triggers an update to the mana. I know this because I sat around seeing if mana ever updated and it only did so when I went afk.

I tested this with absolutely **no** addons to make sure it can't be that.

### AC rev. hash/commit

a22ba935ef6b03cfb94278f7825961716762308a

### Operating system

Debian 11 x64

### Custom changes or Modules

_No response_

UltraNix:
Fixes #12694

## Issues Addressed:

- Closes #12694

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

Create two characters, having one of them be able to cast heal spells really helps
Create a group with these characters
Cast heals on yourself on one of them and wait for the mana to naturally regenerate (level 1 makes this easy)
Check the mana value seen by the other player, it should be full

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -1919,6 +1919,7 @@ void Player::Regenerate(Powers power)
     addvalue += m_powerFraction[power];
     uint32 integerValue = uint32(std::fabs(addvalue));
 
+    bool forcedUpdate = false;
     if (addvalue < 0.0f)
     {
         if (curValue > integerValue)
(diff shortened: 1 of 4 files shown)


Samsequel:
Tested:

Mana updates in similar fashion for the party members as for the player themselves

acidmanifesto (approved):

PR #14116, opened 2022-12-10
fix(Core/LFG): Fixed showing dungeon access requirements only for lea…

### LINKED ISSUE #14070: [dungeon_access_requirements] leader_only does not work for LFG
### Current Behaviour

When an access requirement is set to leader_only, it is not enforced through LFG and the group must enter the instance manually

.

### Source

_No response_

`.quest add 10285` `.quest reward 10285` on one character
rest of the group does not complete quest
attempt to enter Opening the Dark Portal using LFD

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/721711d70f31ad46b615b630c379d61837d263f6

### Operating system

Windows 10

### Custom changes or Modules

_No response_

UltraNix:
…ders in LFG.

Fixes #14070

  

## Issues Addressed:

- Closes #14070

## Tests Performed:

- Tested ingame.

## How to Test the Changes:

`.quest add 10285` `.quest reward 10285` on one character
rest of the group does not complete quest
attempt to enter Opening the Dark Portal using LFD

### DIFF
--- a/src/server/game/DungeonFinding/LFGMgr.cpp
+++ b/src/server/game/DungeonFinding/LFGMgr.cpp
@@ -269,7 +269,7 @@ namespace lfg
             // Recalculate locked dungeons
             for (LfgPlayerDataContainer::const_iterator it = PlayersStore.begin(); it != PlayersStore.end(); ++it)
                 if (Player* player = ObjectAccessor::FindConnectedPlayer(it->first))
-                    InitializeLockedDungeons(player);
+                    InitializeLockedDungeons(player, nullptr);
         }
     }
 
@@ -388,11 +388,10 @@ namespace lfg
 
        @param[in]     player Player we need to initialize the lock status map
     */
-    void LFGMgr::InitializeLockedDungeons(Player* player, uint8 level /* = 0 */)
+    void LFGMgr::InitializeLockedDungeons(Player* player, Group const* group)
     {
         ObjectGuid guid = player->GetGUID();
-        if (!level)
-            level = player->getLevel();
+        uint8 level = player->getLevel();
         uint8 expansion = player->GetSession()->Expansion();
         LfgDungeonSet const& dungeons = GetDungeonsByRandom(0);
         LfgLockMap lock;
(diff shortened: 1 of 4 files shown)


heyitsbench:
I don't understand what the desired behavior is here. Party leader has completed quest 10285, partied with a character who has not completed that quest (character 2), and dungeon finder shows "The Black Morass" as locked, with the reasoning that character 2 hasn't completed the prerequisite quest, though looking at the original issue that looks to be unchanged.

Gultask:
In a group, entering the instance through the portal works fine if it has leader_only set to 1, but that field has no effect on RDF or LFG, the idea is to allow leader_only to affect RDF as well.

You can use https://github.com/azerothcore/mod-progression-system/blob/main/src/Bracket_61_64/sql/world/progression_61_64_dungeon_access_requirements_attunements.sql to test it. As it'll add more dungeons that use leader_only

heyitsbench:
So does the Black Morass not have this leader_only flag by default? If so, why was it given as a testing/repro step?

Gultask:
Black Morass is the only dungeon that has leader_only set to 1 by default.

heyitsbench:
> Black Morass is the only dungeon that has leader_only set to 1 by default afaik.

So then if this PR works as expected, both players should be able to queue as a party for Black Morass with only the leader having completed the quest, correct?

Gultask:
The Heroic version of Black Morass*

image

> So then if this PR works as expected, both players should be able to queue as a party for Black Morass with only the leader having completed the quest, correct?
Yes

heyitsbench:
> The Heroic version of Black Morass

Then in that case, RDF has an issue with both players lacking the Key of Time. Adding this key to the leader's inventory does not unlock the dungeon for both players to queue. Having the key in both players' inventories allows the players to queue if the leader has completed the quest, but not if the leader has not.

Gultask:
I think that's fine, if I'm understanding correctly. The item requirement in this case is not leader_only, only the quest is. Everyone needs to have the key to enter, while only the leader is required to complete the quest.

Which is obviously bizarre and I think the leader_only flag there is misplaced 😛

heyitsbench:
> I think that's fine, if I'm understanding correctly. The item requirement in this case is not leader_only, only the quest is. Everyone needs to have the key to enter, while only the leader is required to complete the quest.

Then in that case I'd say this PR is working as it should. Thanks ❤️

Nyeriah (approved):

PR #14130, opened 2022-12-11
fix(DB/HallsOfStone): Triggers in Tribunal of Ages

Si1ker:
Tested in live server with ~130 players
No negative script/side-effects found.

## Issues Addressed:

* Trigger attacking players - Fix trigger aggro (updates flags)
image

### DIFF
--- a/data/sql/updates/pending_db_world/update.sql
+++ b/data/sql/updates/pending_db_world/update.sql
@@ -1,0 +1,4 @@
+--
+UPDATE `creature_template` SET `flags_extra`=130 WHERE `entry`=31874;
+UPDATE `creature_template` SET `modelid1`=17200, `modelid2`=0 WHERE `entry` IN (31874,28235);
+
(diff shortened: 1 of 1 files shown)


Gultask:
Don't change modelid fields, those are sniffed. AC is missing probabilities fields. Until implemented create an SAI to change modelid on_respawn.
You can disable combat by adding `unit_flags`|768 as well.

Si1ker:
@Gultask check now.

Gultask:
Without any script the creature will definitely use the bad modelid every once in a while :S

Si1ker:
> Without any script the creature will definitely use the bad modelid every once in a while :S

@Gultask , what's the matter? addition of flag_extra 130 makes the trigger:
<html>
<body>


2 | CREATURE_FLAG_EXTRA_CIVILIAN | 0x00000002 | creature does not aggro (ignore faction/reputation hostility)
-- | -- | -- | --



</body>
</html>

<html>
<body>


128 | CREATURE_FLAG_EXTRA_TRIGGER | 0x00000080 | creature is trigger-NPC (invisible to players only)
-- | -- | -- | --



</body>
</html>

Previously field was set to 0 which caused the problem.

Gultask:
Ahh I see, sorry I'm not used to that flag.

Gultask (approved):

Issue #14209, opened 2022-12-13
A bunch of missing quest_template_locale entries.


heyitsbench:
### Current Behaviour

Currently, running a query like this:
```
SELECT `ID`, `locale` FROM quest_template_locale WHERE (`Title` = '' AND (SELECT `LogTitle` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`Details` = '' AND (SELECT `QuestDescription` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`Objectives` = '' AND (SELECT `LogDescription` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`EndText` = '' AND (SELECT `AreaDescription` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`CompletedText` = '' AND (SELECT `QuestCompletionLog` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`ObjectiveText1` = '' AND (SELECT `ObjectiveText1` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`ObjectiveText2` = '' AND (SELECT `ObjectiveText2` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`ObjectiveText3` = '' AND (SELECT `ObjectiveText3` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '') OR (`ObjectiveText4` = '' AND (SELECT `ObjectiveText4` FROM quest_template WHERE quest_template_locale.ID = quest_template.ID) != '');
```
results in 4,168 entries with missing localizations.

### Expected Blizzlike Behaviour

All of these entries should be filled.

### Source

Logical assumption.

### Steps to reproduce the problem

1. Launch the client with any localization of your choosing.
2. Add any of the quests this query results in.
3. Observe the missing localization.

### Extra Notes

My number may be wrong since I might have some entries I filled in based on my [recent PR](https://github.com/azerothcore/azerothcore-wotlk/pull/14202), but the point remains the same.

### AC rev. hash/commit

a131d2f7e8deb7c6a913aab758b681b2db1ba4be

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
[...]

PR #14210, opened 2022-12-13
fix(DB/quest_template_locale): Add missing zhCN localization to quest 349.

heyitsbench:
## Changes Proposed:
- Adds QuestDescription and LogDescription to missing zhCN quest_template_locale entry.

## Issues Addressed:

- Makes progress towards #14209 

## SOURCE:

Wrath Classic live sniffs.

## Tests Performed:

- Untested.

## How to Test the Changes:

1. Run the query.
2. Start a client with zhCN locale.
3. `.quest add 349`
4. Observe no missing localization.

- Sniffed every other quest entry that was "missing" localization from the original issue. No other quest was actually missing localization, they either don't exist on live servers or they don't have any text in the applicable fields.


mpfans:
This is consistent with the English quest, so this PR is not required

PR #14361, opened 2022-12-24
fix(Core/Unit): petrified lichen guard DoT effect

NinjaSoftworks:
## Changes Proposed:
-  Added handling of Spore Cloud (32642) aura from Petrified Lichen Guard
-  Now correctly procs Spore Cloud (32643) DoT instead of Lightning Shield (26364)

## Issues Addressed:

Previously unreported issue: Petrified Lichen Guard incorrectly proccing Lightning Shield instead of Spore Cloud (32643)
## SOURCE:

https://www.wowhead.com/wotlk/spell=32642/spore-cloud
spell.dbc

## Tests Performed:

- Built without errors
- Tested on Windows 10 x86_64
- Tested correct application of DoT effect
- Regression tested actual Lightning Shield spell to ensure this change did not break it

## How to Test the Changes:

1. Create/Select a character able to equip a shield
2. Give item 25828
3. Set Sporeggar rep to honored or above if not already set
4. Equip Petrified Lichen Guard and enter combat
5. Observe that each block correctly applies the DoT effect to the blocked enemy

- [ ] The Spore Cloud (32642) aura still overrides the actual Lightning Shield (324) aura, and vice versa, unsure if blizzlike.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -9278,8 +9278,13 @@ bool Unit::HandleProcTriggerSpell(Unit* victim, uint32 damage, AuraEffect* trigg
                 }
             case SPELLFAMILY_SHAMAN:
                 {
+                    // Spore Cloud (from Petrified Lichen Guard)
+                    if (auraSpellInfo->Id == 32642)
+                    {
+                        trigger_spell_id = sSpellMgr->GetSpellWithRank(32643, auraSpellInfo->GetRank());
+                    }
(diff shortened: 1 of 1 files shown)


NinjaSoftworks:
Upon further investigation, while this fix does solve part of the issue, it does so by introducing a special case where one is not needed. This spell should be using the default behaviors of Unit::HandleProcTriggerSpell, but is being interrupted by a false positive in a special case conditional for Lightning Shield. Will open a new PR with the complete fix once tested.

PR #14368, opened 2022-12-25
fix(Core/AI): Fix target selectors logic when accounting for tanks

Nyeriah:
## Changes Proposed:
-  Fix the logic so it skips tanks when it is supposed to
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1..go creature id 9018
2. notice she casts shadoword pain on tanks again
3.

- [ ]
- [ ]


Nyeriah:
@Maelthyrr Please take a look at this when you have the time, to see if this is intended or we should rename the variable.

Currently target selectors are broken and tanks are excluded by default, because despite **withTank** being set to true, tanks are never selected. So it seems to be like the logic was reversed. Have a look yourself.

55Honey (approved):

pklloveyou:
tc 
```

DefaultTargetSelector::DefaultTargetSelector(Unit const* unit, float dist, bool playerOnly, bool withTank, int32 aura)
    : _me(unit), _dist(dist), _playerOnly(playerOnly), _exception(!withTank ? unit->GetThreatManager().GetLastVictim() : nullptr), _aura(aura)
{
}

```

Nyeriah:
Not really concerned about how it looks like at TC when it's not working as intended.

pklloveyou:
I mean, TC is consistent with your PR
In addition, after using your PR, I tested two bosses and now they work normally

PR #14369, opened 2022-12-26
fix(MACOS/CI): configure os

locus313:
## Changes Proposed:
-  workaround for the current issue with brew python upgrade in github runners https://github.com/actions/runner-images/issues/6817

## Issues Addressed:

- Closes 

## Tests Performed:

- CI


heyitsbench:
Would it not be preferable to just delete that entire commented section since the workaround is no longer necessary whatsoever?

locus313:
> Would it not be preferable to just delete that entire commented section since the workaround is no longer necessary whatsoever?

Still currently in testing..... will clean up after testing is done

FrancescoBorzi (approved):

PR #14378, opened 2022-12-26
feat(Core/ObjectMgr): Add handling specific to heroic player level stats (DKs).

heyitsbench:
## Changes Proposed:
- Adds a couple different checks to if statements during loading of player level stats specific to death knights. This will allow the DB to not have over five-hundred entries it doesn't need, which this PR would also remove.

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

N/A

## Tests Performed:

- Tested with no level 1-54 stats for death knights, worldserver threw no warnings/errors, and I was able to create a death knight. Doing `.level -[number]` set all my base stats to 0.

## How to Test the Changes:

1. Change the configs (START_HEROIC_PLAYER_LEVEL and START_PLAYER_LEVEL) to random values.
2. Make sure that not only do those levels still work in-game if the stats are in the DB, but also that they throw warnings/errors if those stats are missing from the DB.

- [ ] I probably did this wrong, and this PR isn't *strictly* necessary anyway, but the fact that there were a bunch of fake stats that have to be in the DB or else it'll crash bothered me, hence this PR.

### DIFF
--- a/data/sql/updates/pending_db_world/remove-fake-dk-levelstats.sql
+++ b/data/sql/updates/pending_db_world/remove-fake-dk-levelstats.sql
@@ -1,0 +1,1 @@
+DELETE FROM `player_levelstats` WHERE `class` = 6 AND `level` < 55;
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -4100,16 +4100,16 @@ void ObjectMgr::LoadPlayerInfo()
                     continue;
 
                 // fatal error if no level 1 data
-                if (!info->levelInfo || info->levelInfo[0].stats[0] == 0)
+                if (!info->levelInfo || (info->levelInfo[sWorld->getIntConfig(CONFIG_START_PLAYER_LEVEL)].stats[0] == 0 && class_ != CLASS_DEATH_KNIGHT) || (info->levelInfo[sWorld->getIntConfig(CONFIG_START_HEROIC_PLAYER_LEVEL)].stats[0] == 0 && class_ == CLASS_DEATH_KNIGHT))
                 {
-                    LOG_ERROR("sql.sql", "Race {} Class {} Level 1 does not have stats data!", race, class_);
+                    LOG_ERROR("sql.sql", "Race {} Class {} initial level does not have stats data!", race, class_);
                     exit(1);
                 }
 
                 // fill level gaps
                 for (uint8 level = 1; level < sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL); ++level)
                 {
-                    if (info->levelInfo[level].stats[0] == 0)
+                    if ((info->levelInfo[level].stats[0] == 0 && class_ != CLASS_DEATH_KNIGHT) || (level >= sWorld->getIntConfig(CONFIG_START_HEROIC_PLAYER_LEVEL) && info->levelInfo[level].stats[0] == 0 && class_ == CLASS_DEATH_KNIGHT))
                     {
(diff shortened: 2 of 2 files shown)


acidmanifesto (approved):

Kitzunu (approved):

hahalmtdhl:
Without this data, a fatal error will occur and the console cannot be started normally

heyitsbench:
> Without this data, a fatal error will occur and the console cannot be started normally

If you only run the SQL component of this PR, then yes, the server will error during start-up, hence why there are changes to the loading of stats to account for the lost garbage data.

hahalmtdhl:
Tips: Class 6 Level 1 does not have health/mana data!
Normal after filling data

hahalmtdhl:
> > Without this data, a fatal error will occur and the console cannot be started normally
> 
> If you only run the SQL component of this PR, then yes, the server will error during start-up, hence why there are changes to the loading of stats to account for the lost garbage data.

Necessary inspection

Nyeriah:
There's a known issue arising from this that leads to an assert when custom server settings are loaded (DKs with level lower than 55). To address this you may fill the data using the warrior data as that was what it used before, in this custom scenario. Perhaps we should add a fallback at core level to address this.

heyitsbench:
I would add a warning to the conf pointing out the issue. [A known issue from what I know is that there's a crash when it asks for stats that don't exist from the DB](https://github.com/azerothcore/azerothcore-wotlk/issues/14808), which it would exit the program anyway, so an exit no matter what is intended from this PR if stats are missing. Workaround obviously is filling it with other stats, which isn't hard, but I'm definitely not doing it as I see no reason to include garbage data for custom use only, as the entire point of this PR was to *remove* that data.

Gultask:
If there's an option in the conf for it, it should be able to support it, no?

Maybe there could be a check like:

if (badData && class == dk)
{
    use warrior stats as before to not crash the server
}

?

heyitsbench:
Feel free to submit a PR (or revert this one) if you want to do it the way you want, glad these opinions got caught in review.

Nyeriah:
What's your problem anyway? It's not like things are immutable, it's not like people don't change their opinions and things that weren't thought about before couldn't get a revision some time later. It's pretty common in an opensource project, with many different authors and nobody to particularly claim ownership of what is written.

heyitsbench:
Then redo it to the ever-changing standards of the project.

Nyeriah:
Not sure where this is again coming from but okay, I am glad that we can keep moving forward, always improving what needs to be improved and rectifying what needs to be rectified. The goal is afterall to provide a solid emulator and to learn as much as possible with what we do. It's a real shame when we have to dabble over this instead of focusing on what really matters, and that is, writing code

htc16:
This PR was crashing my server. I managed to make a workaround, but now it's back to dropping, I'm reversing it directly to the source to see if it stops falling.
I'm limiting the maximum level of players so that we can enjoy all the game's content, however, less than level 55 after that the server breaks.

htc16:
Guys, my feedback on this PR:
I updated my December base for now this week, when I turned on the server the same did not start.
I made a gambiarra in the database the day before yesterday, and today, as amazing it seems bouched it all again and the server did not start.
Solution at the moment was me to reverse this PR directly on my source.
Note: I am using a maximum level below 55, which with this PR is not possible. But I'm doing this to force my players to enjoy all the contents before rising the level

heyitsbench:
@htc16 [Here's a solution that will work with this PR](https://gist.github.com/heyitsbench/21fa68994f0f20563148c164ee3dcc67). Just a simple query to run on your world DB.

htc16:
in reality there were 2 acts to work out.  The first one you reported and the second one was to revert to the source with this current pr we are talking about.

PR #14379, opened 2022-12-27
fix(DB/creature): Remove Bonelashers inside of a wall.

heyitsbench:
## Changes Proposed:
- Removes two [Bonelashers](https://www.wowhead.com/wotlk/npc=18470) from the walls of Auchindoun.

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

It's simply illogical!

## Tests Performed:

- Untested.

## How to Test the Changes:

1. Run the query.
2. `.go xyz -3381.7 4616.8 -22.4 4.3`
3. Turn on nameplates.
4. Observe a lack of Bonelashers inside walls.

### DIFF
--- a/data/sql/updates/pending_db_world/bonelasher-oob.sql
+++ b/data/sql/updates/pending_db_world/bonelasher-oob.sql
@@ -1,1 +1,1 @@
-DELETE FROM `creature` WHERE `guid` IN (66458, 66459);
+DELETE FROM `creature` WHERE `guid` IN (66458, 66459) AND `map` = 530 AND `spawntimesecs` = 300 AND `curhealth` = 5715;
--- a/data/sql/updates/pending_db_world/bonelasher-oob.sql
+++ b/data/sql/updates/pending_db_world/bonelasher-oob.sql
@@ -1,1 +1,1 @@
-DELETE FROM `creature` WHERE `guid` = 66458;
+DELETE FROM `creature` WHERE `guid` IN (66458, 66459);
--- a/data/sql/updates/pending_db_world/bonelasher-oob.sql
+++ b/data/sql/updates/pending_db_world/bonelasher-oob.sql
@@ -1,0 +1,1 @@
+DELETE FROM `creature` WHERE `guid` = 66458;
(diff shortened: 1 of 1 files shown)


heyitsbench:
There are more than one Bonelashers in the walls, gimme a bit.

heyitsbench:
The birds have been exterminated.

IntelligentQuantum (approved):

Samsequel:
@Gultask 

Should we remove these or are they just wrongly spawned (are they extra, or not :?)

Gultask:
Impossible to know what happened to create those but this isn't the first time, specially in Outland.
Remove them. If we go back to re-sniff the spawns it'd be easier to just re-create it rather than checking/speculating the individual spawns of a wandering creature.

data/sql/updates/pending_db_world/bonelasher-oob.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM `creature` WHERE `guid` IN (66458, 66459);
```
Gultask, line 1: Missing AND `id1`=XXXX

Gultask (changes requested):

data/sql/updates/pending_db_world/bonelasher-oob.sql
heyitsbench: Update [the wiki](https://www.azerothcore.org/wiki/sql-standards) with that info, please.

Gultask: No no I meant id1 😄 
There are hundreds of creatures with curhealth=5715, and thousands with map=530

```
DELETE FROM `creature` WHERE `guid` IN (66458, 66459) AND `id1`=18470;
```

heyitsbench: And there are most definitely > 2 creatures with an id1 of 18470, I really don't see the point of specifying further than a keyed column where duplicates aren't possible.

Gultask: It's been a longstanding practice for quite a while

image

heyitsbench: > It's been a longstanding practice for quite a while

I would not consider under two months to be "quite a while". But even still, it's not in the standard operating procedure on the wiki with regards to SQL queries, so I still wouldn't call it a standard practice.

Gultask: That's just an example I found. I've always been taught to add GUIDs and IDs together.

Gultask (approved):

data/sql/updates/pending_db_world/bonelasher-oob.sql
heyitsbench: Well for people like me, these standards need to be on the wiki. I can only scan through so many issues/PRs I have nothing to do with.

PR #14388, opened 2022-12-28
feat(Core/Player): delete character cache in Player::DeleteFromDB()

r-o-b-o-t-o:
## Changes Proposed:
- Remove the character from cache at the end of `Player::DeleteFromDB()`
- This ensures that the cache is always cleared when deleting a character. For example the `.character erase` command didn't clear the cache after deletion.

## Tests Performed:

Compiled and tested on Windows 11

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4240,7 +4240,7 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             return;
     }
 
-    const CharacterCacheEntry* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
+    CharacterCacheEntry const* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
     if (cache)
     {
         std::string name = cache->Name;
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -3943,6 +3943,8 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
 
     CharacterDatabasePreparedStatement* stmt = nullptr;
 
+    bool deleted = true;
+
     switch (charDelete_method)
     {
         // Completely remove from the database
@@ -4237,11 +4239,24 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             }
         default:
             LOG_ERROR("entities.player", "Player::DeleteFromDB: Unsupported delete method: {}.", charDelete_method);
+            deleted = false;
             return;
     }
 
+    if (deleted)
+    {
+        const CharacterCacheEntry* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
+        if (cache)
+        {
+            std::string name = cache->Name;
+            sCharacterCache->DeleteCharacterCacheEntry(playerGuid, name);
+        }
+    }
+
     if (updateRealmChars)
+    {
         sWorld->UpdateRealmCharCount(accountId);
+    }
 }
 
 /**
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4240,8 +4240,7 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             return;
     }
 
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -4237,11 +4239,24 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             }
         default:
             LOG_ERROR("entities.player", "Player::DeleteFromDB: Unsupported delete method: {}.", charDelete_method);
+            deleted = false;
             return;
     }
 
+    if (deleted)
+    {
+        const CharacterCacheEntry* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
```
Kitzunu, line 4248:
```suggestion
        CharacterCacheEntry const* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
```

src/server/game/Entities/Player/Player.cpp
```diff
@@ -4237,11 +4239,24 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
...
+        if (cache)
+        {
+            std::string name = cache->Name;
+            sCharacterCache->DeleteCharacterCacheEntry(playerGuid, name);
```
Kitzunu, line 4252:
```suggestion
            sCharacterCache->DeleteCharacterCacheEntry(playerGuid, cache->Name);
```

Why not this?

src/server/game/Entities/Player/Player.cpp
r-o-b-o-t-o: From what I read there is no difference between the two, is this the preferred codestyle for AC?

src/server/game/Entities/Player/Player.cpp
```diff
@@ -4237,11 +4239,24 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             }
         default:
             LOG_ERROR("entities.player", "Player::DeleteFromDB: Unsupported delete method: {}.", charDelete_method);
+            deleted = false;
```
Helias, line 4242: I do not think that this flag `deleted` is necessary, because if `deleted = false` the function will `return` so it will never execute the bottom code, right?

src/server/game/Entities/Player/Player.cpp
r-o-b-o-t-o: `DeleteCharacterCacheEntry` takes the string by reference, and erases the cache before using the `name`. So by the point you're using the `name` string it's invalid memory since it's freed just before. By creating a copy of the string before calling `DeleteCharacterCacheEntry` we're safe.

src/server/game/Entities/Player/Player.cpp
r-o-b-o-t-o: Good point, I think I read that too fast and mistook the `return;` for a `break;`

src/server/game/Entities/Player/Player.cpp
Kitzunu: Yes

src/server/game/Entities/Player/Player.cpp
Kitzunu: Cool, thanks for explaning

src/server/game/Entities/Player/Player.cpp
Kitzunu: You can find more on our codestyle here https://www.azerothcore.org/wiki/cpp-code-standards

r-o-b-o-t-o: Ah cheers, I looked for that page but couldn't find it 😄

Kitzunu (approved):

src/server/game/Entities/Player/Player.cpp
```diff
@@ -4240,8 +4240,17 @@ void Player::DeleteFromDB(ObjectGuid::LowType lowGuid, uint32 accountId, bool up
             return;
     }
 
+    CharacterCacheEntry const* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid);
+    if (cache)
```
Nyeriah, line 4244: You could wrap the variable inside the if

```cpp
if (CharacterCacheEntry const* cache = sCharacterCache->GetCharacterCacheByGuid(playerGuid))
{
       std::string name = cache->Name;
        sCharacterCache->DeleteCharacterCacheEntry(playerGuid, name);
}
```

Kitzunu (approved):

PR #14393, opened 2022-12-29
fix(Core/PlayerQuest): Get base skill value for quest prereqs.

heyitsbench:
## Changes Proposed:
- Replaces a `GetSkillValue` with `GetBaseSkillValue` when checking quest requisites for a player.

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

Wrath Classic PTR, same scenario as the test performed/suggested test.

## Tests Performed:

- Rudimentary test with [one quest](https://www.wowhead.com/wotlk/quest=6607), one [skill increasing consumable](https://www.wowhead.com/wotlk/item=6533), and one [skill increasing equippable item](https://www.wowhead.com/wotlk/item=12225).

## How to Test the Changes:

1. `.go xyz -4139.7 -4058.2 2.3 1.9 1`
2. `.setskill 356 125 300`
3. `.additem 6533 5`
4. `.additem 12225`
5. Equip the fishing pole and use the lure, observe that the quest (with a 225 skill prereq for fishing) is not available.


Samsequel:
Curious if like a racial from blood elfs or like Draenei should effect availability and if this PR would effect that.

heyitsbench:
> Curious if like a racial from blood elfs or like Draenei should effect availability and if this PR would effect that.

Earliest skill requirement for enchanting or jewelcrafting in our DB is jewelcrafting, 350. I don't think I'm going to get the blizzlike behavior for the racial bonuses.

Tested this PR with a draenei with jewelcrafting, set my skill level to 350/455 (thus, racial bonus included) and was *not* able to grab [a quest with 350 jewelcrafting required](https://www.wowhead.com/wotlk/quest=12918), which I'm going to *assume* is not blizzlike, thus this PR doesn't work in that case. 😦

Gultask:
I guess there isn't another function that excludes only temporary skill increases?

heyitsbench:
> I guess there isn't another function that excludes only temporary skill increases?

Looks like `GetBaseSkillValue` takes the permanent bonus from racial passives into account for its calculations. I've updated the PR accordingly, along with retested in both the original test scenario and the draenei jewelcrafting one, works as (I assume) it should.

Samsequel:
Tested; seems to work as described.

acidmanifesto (approved):

Issue #12260, opened 2022-07-01
[Fishing] Character remain stuck in fishing channeling after clicking the bobber prematurely


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3701

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

[Fishing](https://wowgaming.altervista.org/aowow/?spell=7738) keeps channelling after clicking the bobber before the fish get hooked, making character stuck in animation until he recasts fishing or other channelling spell.

**Note:** This does not affect cancelling fishing by moving.

https://www.youtube.com/watch?v=V-CLBW1u2uw

### Expected Blizzlike Behaviour

All animations and channelling should be naturally cancelled once the player stop fishing (for any reason)

### Steps to reproduce the problem

1. `.additem 6256`
2. `.learn 7620`
3. cast Fishing
4. click the bobber before a fish is hooked
5. see that the channeling is not cancelled

### Extra Notes

https://user-images.githubusercontent.com/100873791/176953966-f1728270-2c95-4852-8e30-569c6345fc81.mp4

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/6983903a89c556c0ca1239101cffc38098d6b081

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[...]

PR #14422, opened 2022-12-30
fix(Core/GameObject): Make fishing bobber finish a channeled spell rather than interrupt in all cases.

heyitsbench:
## Changes Proposed:
- Removes a scenario in which a gameobject is interacted with prematurely, where it then interrupts a spell rather than finishing it.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/3701
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12260

## SOURCE:

Wrath Classic PTR.

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. `.learn 7620`
2. `.additem 6256`
3. Go to a fishing spot.
4. Fish, click the bobber when it catches a fish, before it's caught a fish, let it time out, move after casting, un-equip the pole after casting, do whatever you want. The spell channel bar should *never* remain after you've cancelled it.

- Could not get `player->InterruptSpell` to actually disable the channel no matter what overloads I provided to it. Don't know if the function is just borked or if it doesn't apply to channeled spells, or what. The end visual result to the player is the same as when the spell is finished, i.e. when it actually has caught a fish on Classic, so the same behavior is likely fine here.
- ~~`tmpfish` is used in another gameobject interaction case, goobers, but I have no idea what that's for. I left it untouched there, but that's the reason I didn't gut the variable entirely.~~ I gutted the variable entirely.


heyitsbench:
Genuinely cannot tell what the `tmpfish` would do in the case of goober gobjects. Code never existed this way in TC or other projects I've looked at, so 🤷‍♀️. I'm just going to delete it. In the event someone finds out what the heck it's supposed to be doing, feel free to point it out.

heyitsbench:
Seems the modules CI build is failing due to an issue with [Eluna](https://github.com/azerothcore/mod-eluna), not this PR.

acidmanifesto (approved):

PR #14425, opened 2022-12-30
fix(Core/Auras): SPELL_AURA_MOD_SCALE should scale additively instead…

Nefertumm:
… of multiplicatively

## Changes Proposed:
-  Make AURA_MOD_SCALE additive instead of multiplicative.

## Issues Addressed:

- Closes -

## Tests Performed:

- Tested in-game with #14426

## How to Test the Changes:

1. Check pet scale, they should be normal
2. Load the PR linked above, mushrooms from the boss should be fairly visible compared without this PR.
3. Check others creatures failing with scale?
4.


Kitzunu (changes requested):

Gultask:
- Vekniss Borer: No change
- Spells such as 16711 and 8212 still stack, but doesn't look like it is multiplicative 
- On NPCs the Gor'Drek Ointment for quest 10488 works
- Cast spell 31691 (SHRINK) on self, then cast a bunch of 31698 (GROW) on self. It was fine.

Nefertumm:
zzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

Gultask:
Tested pets as well. They grow when leveling up and retain that size when dismissed.

Gultask (approved):

PR #14435, opened 2022-12-30
feat(Core/Mail): call CanSendMail() even if no items are sent

r-o-b-o-t-o:
## Changes Proposed:
- Currently, the `CanSendMail()` player hook is only called if the mail contains items, which doesn't make a lot of sense. This PR addresses that by running the hook if it hasn't been called from the items loop before.
- Adds a `SendMailResult` if the hook returns `false`, otherwise the client's mail UI bugs out.

## Tests Performed:

Compiled and ran on Windows 11, tested the hook with Eluna


src/server/game/Handlers/MailHandler.cpp
```diff
@@ -269,12 +269,19 @@ void WorldSession::HandleSendMail(WorldPacket& recvData)
...
         if (!sScriptMgr->CanSendMail(player, receiverGuid, mailbox, subject, body, money, COD, item))
         {
+            player->SendMailResult(0, MAIL_SEND, MAIL_ERR_INTERNAL_ERROR);
             return;
         }
 
         items[i] = item;
     }
 
+    if (items_count == 0 && !sScriptMgr->CanSendMail(player, receiverGuid, mailbox, subject, body, money, COD, nullptr))
```
Kitzunu, line 279:
```suggestion
    if (!items_count && !sScriptMgr->CanSendMail(player, receiverGuid, mailbox, subject, body, money, COD, nullptr))
```

acidmanifesto (approved):

Issue #14325, opened 2022-12-21
Corpse decay timers broken within the last three days


ChibaPet:
### Current Behaviour

I updated and deployed three days ago, on December 18th, and things were evidently fine. I just did so again today, AC> AzerothCore rev. 7f6788ce86a3 2022-12-20 16:09:17 +0000 (master branch) (Unix, RelWithDebInfo, Static), only plugin being transmog, and now corpses despawn immediately after being looted. I've only noted the issue with normal corpses, and haven't sought out other sorts to compare.

Relevant config, unchanged recently:

Corpse.Decay.NORMAL    = 300
Corpse.Decay.RARE      = 300
Corpse.Decay.ELITE     = 300
Corpse.Decay.RAREELITE = 300
Corpse.Decay.WORLDBOSS = 3600
Rate.Corpse.Decay.Looted = 1

I don't see any recent changes in src/server/game/Entities/Creature/Creature.cpp which makes me curious how it's behaving differently.

### Expected Blizzlike Behaviour

Obey Corpse.Decay.NORMAL * Rate.Corpse.Decay.Looted config

### Steps to reproduce the problem

Find and kill a normal mob. Loot it.

### AC rev. hash/commit

AzerothCore rev. 7f6788ce86a3 2022-12-20 16:09:17 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1n  15 Mar 2022 (library: OpenSSL 1.1.1n  15 Mar 2022)
Using Boost version: 1.74.0
Using MySQL version: 100518
Using CMake version: 3.18.4
Compiled on: Linux 5.10.0-20-amd64
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /opt/acore/data/maps. Total size: 291014951 bytes
vmaps directory located in /opt/acore/data/vmaps. Total size: 658130721 bytes
mmaps directory located in /opt/acore/data/mmaps. Total size: 2320900676 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 
Using World DB: ACDB 335.8-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
> List enable modules:
- mod-transmog

### Operating system

Debian 11

[...]

PR #14039, opened 2022-12-03
fix(Core/Config): Set default corpse decay for normal creatures to 5 …

UltraNix:
…minutes.

Fixes #11710

 

## Issues Addressed:

- Closes #11710

## Tests Performed:

- Not tested.

## How to Test the Changes:

Open your `worldserver.conf` and change `Corpse.Decay.NORMAL` to 300.
Go to the Barrens
Kill Adder, Sickly Gazelle, or Greater Plainstrider
Observe corpse decay time - corpse should be gone after 5 min.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #14474, opened 2023-01-02
Revert "fix(Core/Config): Set default corpse decay for normal creatures to 5 …"

Gultask:
Reverts azerothcore/azerothcore-wotlk#14039

Causing corpses to despawn instantly: https://github.com/azerothcore/azerothcore-wotlk/issues/14325

### DIFF
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -1691,13 +1691,13 @@ WorldBossLevelDiff = 3
 #    Corpse.Decay.RAREELITE
 #    Corpse.Decay.WORLDBOSS
 #        Description: Time (in seconds) until creature corpse will decay if not looted or skinned.
-#        Default:     300  - (5 Minutes, Corpse.Decay.NORMAL)
+#        Default:     60   - (1 Minute, Corpse.Decay.NORMAL)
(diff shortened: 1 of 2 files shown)


heyitsbench:
For the record, only the `Rate.Corpse.Decay.Looted` section should be reverted. Still not blizzlike, but more blizzlike with the `Corpse.Decay.NORMAL` set to 300.

Gultask:
We're testing it with different values and it still causes the issue, unfortunately.

heyitsbench:
> We're testing it with different values and it still causes the issue, unfortunately.

[Then your testing is wrong.](https://user-images.githubusercontent.com/38561765/210276277-1e1ecb19-a90e-4efa-aad7-33aa8084bcae.png)

Samsequel:
Tested it, and it instantly despawned after looting

Samsequel:
Revert fixed the issue

Samsequel:
image

Samsequel (approved):

heyitsbench:
Again, reverting just `Rate.Corpse.Decay.Looted` fixes the issue, you don't need to revert `Corpse.Decay.NORMAL` as well.

Gultask:
Me and Sam have no further knowledge of the issue. And we can't keep this PR and just expect people to know to change the config on their own servers.

UltraNix or yourself can open a PR to correct it later.

Issue #10986, opened 2022-03-12
Quest “A Change of Heart ” “Choose Your Path” series issues


fullofstarschen:
### Current Behaviour

“A Change of Heart ”  When you finish this quest,you can get a new ring instead the old one,but if you finish this quest in battle status,you can get the new one without lost the old one,so you can get two rings.The same  issue with “Choose Your Path” series quests.

### Expected Blizzlike Behaviour

When you finish the quest,you should lost the old one.

### Steps to reproduce the problem

1,Use a Mage and enter the Citadel,
2,Find NPC and get the quest,summon water elemental to attack the enemy far away,
3,Finish the quest when you are in battle status,
4,check the rings.   

### Extra Notes

I think this is because the ring can't be dismounted  in battle status.

### AC rev. hash/commit

5ed4db09e3c8

### Operating system

windows 10 x64

SKingOWN:
Confirmed!

PR #14487, opened 2023-01-03
fix(Core/Quests): Fixes quests requiring an item that cannot be unequipped in combat.

SKingOWN:
## Changes Proposed:
-  Fix quest "A Change of Heart"
-  Fix quest "Choose Your Path"

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10986

## Tests Performed:

- Builds on Ubuntu 20.04
- Builds on Windows 11 Pro with Visual Studio 2022
- Tested in game.

## How to Test the Changes:

1. First you need to reproduce the problem in the ISSUE ADDRESSED.
2. Then apply the fix and build the core.
3. .tele icecrownc
4. Enter to the instance Icecrown Citadel.
5. .mod repu 1156 exalted
6. .gm vis on
7. .gm off
8. Obtain and equip a ring from the quest "Choose Your Path".
9. Enter in combat.
10. Try to complete the quest "Choose Your Path".
11. And then yo will be removed from combat and the old ring will be removed from your inventory succesfully.

- [ ] To improve this solution, you must first check if the item required by the quest is equipped in a slot that cannot be unequipped in combat.
- [ ] I'm not sure if rewarding one quest takes you out of combat will affect any gameplay on another quest.

qg

### DIFF
--- a/src/server/game/Entities/Player/PlayerQuest.cpp
+++ b/src/server/game/Entities/Player/PlayerQuest.cpp
@@ -666,7 +666,20 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
     {
         if (sObjectMgr->GetItemTemplate(quest->RequiredItemId[i]))
         {
-            CombatStop();
+            if (IsInCombat())
+            {
+                for (int i = EQUIPMENT_SLOT_START; i < INVENTORY_SLOT_BAG_END; ++i)
+                {
+                    if (Item* pItem = GetItemByPos(INVENTORY_SLOT_BAG_0, i))
+                    {
+                        if (pItem->GetEntry() == quest->RequiredItemId[i])
+                        {
+                            CombatStop();
+                            break;
+                        }
+                    }
+                }
+            }
             DestroyItemCount(quest->RequiredItemId[i], quest->RequiredItemCount[i], true, true);
         }
     }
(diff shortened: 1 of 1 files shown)


Valdifer:
The thing i see doing that to forcing combatStop is that someone can abuse and exploit something in other quests, does the quest has an script inside scripts? i was watching instance_icecrown but nothing about the ring. 

Dunno if that could be fixed via item_scripts

SKingOWN:
> The thing i see doing that to forcing combatStop is that someone can abuse and exploit something in other quests, does the quest has an script inside scripts? i was looking instance_icecrown but nothing about the ring.
@Valdifer I also thought about that, if there was a way to make it so that you can't interact with these NPCs while in combat, then the CombatStop() would not be necessary; but i already tried that way and couldn't find solution.
Another thing that could be done to improve this is before calling CombatStop(), check if the item is equipped.
If any of you can help with that, it would be appreciated.

Valdifer:
@SKingOWN u can call player from that class doing this, could check using this:
if(!this->IsInCombat()) to check if player isnt in combat

SKingOWN:
@Valdifer Thanks to your suggestion I was able to find the necessary code to improve the fix. It now first checks if you're in combat, and if so then checks if the quest item is equipped, and then calls CombatStop() if both conditions are met.

SKingOWN:
Tested with the last commits and working perfectly.

acidmanifesto:
Is this handled blizz like? Also with this being a global change we will need testers on this to see what else may be affected.

Maelthyrr:
Close in favor of https://github.com/azerothcore/azerothcore-wotlk/pull/15092

PR #14519, opened 2023-01-06
fix(Scripts/HellfireRamparts): Fix Nazan getting stuck in middle air

Nefertumm:
## Changes Proposed:
-  Make the creature passive during the flypath.
-  It should perform a move land when it reaches the waypoint.
-  In all cases, set disable gravity and fly to off on death.

## Issues Addressed:

- Closes 

## Tests Performed:

-  Not tested.

## How to Test the Changes:

1. Go to Hellfire Ramparts, engage Vazruden and get it to 30% so Nazan descends.
2. Make sure it does the full path and it lands correctly (it shouldn't make flying movements when it is on ground)
3. ??


Maelthyrr:
Hi, tested 5 times in different configurations. The corpse will always fall and not be stuck in the air. So work fine :)

PR #14569, opened 2023-01-08
Fix module related cmake error

Foereaper:
## Changes Proposed:
-  Remove unnecessary regex check and use simple string replace instead

## Issues Addressed:
- The original code can cause errors depending on the absolute path to your source directory. This fixes that.

## SOURCE:

https://stackoverflow.com/questions/75045135/cmake-regex-replace-fails-to-compile

## Tests Performed:

- Tested and working

## How to Test the Changes:

1. Put your source directory in a subdirectory named C++, then try to generate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ FOREACH(subdir ${sub_DIRS})
         continue()
     endif()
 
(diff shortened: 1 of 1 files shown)


AnchyDev:
Tested this by cloning into `C:/AnchyDev/Projects/C++/azerothcore-wotlk/` and I can configure+build CMake without issues, compiling works without issues also.

acidmanifesto (approved):

Issue #12718, opened 2022-08-13
The Missing Diplomat: Private Hendel gets stuck on bridge and can't continue RP.


heyitsbench:
### Current Behaviour

Currently, upon subduing Private Hendel, if he is still on the dock and not on land, he can get stuck on the bridge and fail to continue his RP, prohibiting the quest credit from being given.

### Expected Blizzlike Behaviour

Private Hendel should get to whatever spot he needs to and credit should be given.

### Source

Common sense, also Temper.

### Steps to reproduce the problem

1. .quest add 1324
2. Fight Private Hendel
3. Try again until you observe the issue.

### Extra Notes

Possibly related to https://github.com/chromiecraft/chromiecraft/issues/3442.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/e6dd5d3e22d92dddc841079f9108275fd0359be1

### Operating system

Ubuntu 20.04

### Custom changes or Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
[...]

PR #14633, opened 2023-01-12
fix(DB/SAI): Set correct ReactState to Private Hendel on quest The Missing Diplomat

Gultask:
## Changes Proposed:
-  Changes ReactState to Passive instead of Aggressive
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12718
- Closes https://github.com/chromiecraft/chromiecraft/issues/3442

## Tests Performed:

- Tested ingame but could use more tests
- 

## How to Test the Changes:

1. `.quest add 1266`
2. `.go xyz -2897.153809 -3349.867432 31.299358 1 6.278478` don't move any further.
3. Complete quest 1266, take the next quest
4. Proceed with the quest, see if it plays normally

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1673526587449992500.sql
+++ b/data/sql/updates/pending_db_world/rev_1673526587449992500.sql
@@ -1,0 +1,2 @@
+--
+UPDATE `smart_scripts` SET `action_param6`=0 WHERE `entryorguid`=496601 AND `source_type`=9 AND `id`=3 AND `link`=0;


Gultask:
Doesn't work with Pets, I'll see if I can figure something out.

Gultask:
Ok. So the problem is **not** the SAI. It's definitely the movement around the pier.

PR #14707, opened 2023-01-18
Removal of unnecessary .dist file loading

Foereaper:
## Changes Proposed:
Removal of .dist file loading and checking. The .dist file has little to no value other than as a blueprint when the .conf file is what should be edited in all cases. The conf file should be the master, the core should be the fallback.

The core already has preset default values for any config option, and these hard coded default values is what should load in the case of an outdated .conf file, or if any options are missing.

If, for any reason, the config file is missing any values, the core will output the default config option and which config file is missing said values. The user is then instructed what to add to the config file. The core will not halt in these circumstances, and will fall back to the default values.

## Tests Performed:
- Compiled, ran and tested with missing config options. The core behaves as expected.

## How to Test the Changes:

1. Run the server without any .dist file present. The core should start as expected.
2. Remove any setting (other than critical settings like database connection info etc) from the .conf file
3. The core should still start, but output an error message in the console with what setting is missing from what file.

### DIFF
--- a/apps/ci/ci-gen-server-conf-files.sh
+++ b/apps/ci/ci-gen-server-conf-files.sh
@@ -34,6 +34,7 @@ if [[ $APP_NAME != "authserver" ]]; then
 		echo "MySQLExecutable = \"\""
 		echo "TempDir = \"\""
 		echo "SourceDirectory = \"\""
+		echo "Creatures.CustomIDs = \"\""
     } >> ./env/dist/$CONFIG_FOLDER/$APP_NAME.conf
 fi
 
--- a/apps/ci/ci-gen-server-conf-files.sh
+++ b/apps/ci/ci-gen-server-conf-files.sh
@@ -10,6 +10,6 @@ cp ./src/server/apps/$APP_NAME/$APP_NAME.conf.dist ./env/dist/$CONFIG_FOLDER/$AP
 sed -i "s/127.0.0.1;3306;acore;acore/localhost;3306;root;$MYSQL_ROOT_PASSWORD/" ./env/dist/$CONFIG_FOLDER/$APP_NAME.conf
 
 if [[ $APP_NAME == "worldserver" ]]; then
-    sed -i 's/DataDir = \"\"/DataDir = \".\/data"/' ./env/dist/$CONFIG_FOLDER/$APP_NAME.conf
+    sed -i 's/DataDir = \".\"/DataDir = \".\/data"/' ./env/dist/$CONFIG_FOLDER/$APP_NAME.conf
     git clone --depth=1 --branch=master --single-branch https://github.com/ac-data/ac-data.git ./env/dist/$BIN_FOLDER/data
 fi
--- a/apps/ci/ci-gen-server-conf-files.sh
+++ b/apps/ci/ci-gen-server-conf-files.sh
@@ -4,7 +4,7 @@ BIN_FOLDER=${3-"bin"}
 MYSQL_ROOT_PASSWORD=${4:-""}
 
 # copy dest files to conf files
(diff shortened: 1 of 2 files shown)


Yehonal:
I can't read the code now but the logic behind the PR makes sense and I like it. But there's a drawback here that must be taken in account: in this way we are not checking if the dist file has valid properties or not. If the intention is to have a single source of truth for the defaults, we can just keep the validation of the dists while do not use them for the defaults.

In any case, please make sure that the docker configurations/setup are still working after this

Foereaper:
The dist file is technically always copied from the source on build, so it should always be correct as a blueprint anyway. If anyone decides to do changes to said file then I wouldn't really worry much about that. The defaults set in the core should always be considered the truth, so.

In regards to docker, I reckon there needs to be an additional step to rename the .conf.dist file to .conf. We can do this as an additional build step (this would also fix the "failing" checks here on github), but the issue with this is if anyone has done any changes to said .conf file it would be overwritten. 

How are the config files currently handled in Docker, do you know? Do they always just use the base .dist files with base settings, unless the user creates a .conf themselves?

If so, then we can copy the .conf.dist to .conf if .conf does not exist. It would then be up to the end user to append any future configuration additions to the .conf file like normal.

Kitzunu:
I'd like to ask you to edit the https://www.azerothcore.org/wiki/server-setup accordingly too.

heyitsbench:
> I'd like to ask you to edit the https://www.azerothcore.org/wiki/server-setup accordingly too.

Is there even anything to edit there?

Kitzunu:
image

Foereaper:
> I'd like to ask you to edit the https://www.azerothcore.org/wiki/server-setup accordingly too.

The steps for copying and renaming the dist files would still be the same, unless we want the post build script to also output a proper .conf file if one does not already exist? I am indifferent, I have the cmake changes done to make the raw .conf files, I just have to add an existence check.

Kitzunu:
Tbh it doesnt matter, my only change would be to not say "copy" but just "rename" I guess

Foereaper:
> Tbh it doesnt matter, my only change would be to not say "copy" but just "rename" I guess

That I can agree with

Foereaper:
Looked at the Docker install, it already handles .conf files properly, so there's no need for any special handling here. 

The procedure will be the same as before for all other platforms, ie. copying/renaming the .dist file. I'm not going to push any special rename functionality for the post-build script, it's unnecessary and could break a few edge cases.

This PR should be good for now unless there's any specific feedback.

Do mind; This only sorts the default loading of .dist files for the stock config files. There is still handling left in the core for .dist files for modules. I would really recommend removing this as well, but that should be its own PR.

Kitzunu (approved):

Yehonal:
I was mostly referring to make sure that our .dist file is always functioning. Now that we do not read/validate it anymore, someone could apply wrong changes to that file via a PR. We might still want to test this file somehow in the pipeline if possible. Not a critical problem though

Foereaper:
This could be tested in the pipeline by using the dist files as template config files instead of crafting minimal config files (same sh script as edited a lot in this pr)

Foereaper:
Rewrote the conf generator script, it should now make a copy of the .dist and rename it .conf, this ensures that the full dist file is parsed. We then replace the relevant vars like SQL info and datadir.

Yehonal (approved):
GJ

Foereaper:
Is any further testing required? I have tested that it works as expected, and the automated builds show that it also works as expected. I reckon this is ready for a merge, it shouldn't have any backwards compatibility issues.

Kitzunu:
Appreciated

Issue #14033, opened 2022-12-03
Quest "Beer Basted Boar Ribs" not obtainable


BeefSteaker:
https://github.com/chromiecraft/chromiecraft/issues/4524
### What client do you play on?

enUS

### Faction

Alliance

### Content Phase:

1-19

### Current Behaviour

Quest https://wowgaming.altervista.org/aowow/?quest=384 is is not obtainable, npc does not interact

![WoWScrnShot_120222_225535](https://user-images.githubusercontent.com/93767462/205385595-33472e98-ceac-4561-9af9-b19da42d2b3e.jpg)

### Expected Blizzlike Behaviour

quest should be obtainable from level 5

### Source

lots of comments acknowledge its existence
https://www.wowhead.com/wotlk/quest=384/beer-basted-boar-ribs

### Steps to reproduce the problem

1. Create a dwarf/gnome
2. .level 4
3. .go c id 1267

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/11530ecb95a35f66a2c220781172c57e0749ac48

### Operating system

Windows 7

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #14034, opened 2022-12-03
fix(DB/quest_template): Beer Basted Boar Ribs (384) shouldn't require Cooking skill

Gultask:
## Changes Proposed:
-  Removes Skill requirement
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14033
- Closes https://github.com/chromiecraft/chromiecraft/issues/4524

## SOURCE:

WotLK Classic PTR

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1. Create a dwarf/gnome
2. .level 4
3. .go c id 1267

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #14711, opened 2023-01-19
Removed cooking requirement for Beer Basted Boar Ribs quest

sh3rp:
## Changes Proposed:
Remove Cooking requirement for quest

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14033
- Closes https://github.com/chromiecraft/chromiecraft/issues/4524

## SOURCE:
see issue

## Tests Performed:
- Created dwarf character, went to quest giver, verified quest giver is available without the cooking skill trained

## How to Test the Changes:
1. Create dwarf char
2. .levelup 4
3. .go c id 1267
4. Verify quest is available for pick up
5. Pick up quest from quest giver

### DIFF
--- a/data/sql/updates/pending_db_world/fix-14033.sql
+++ b/data/sql/updates/pending_db_world/fix-14033.sql
@@ -1,2 +1,3 @@
 UPDATE `quest_template` SET `QuestSortID` = 0 WHERE `ID` = 384;
 UPDATE `quest_template_addon` SET `RequiredSkillID` = 0, `RequiredSkillPoints` = 0 WHERE `ID` = 384;
+
--- a/data/sql/updates/pending_db_world/fix-14033.sql
+++ b/data/sql/updates/pending_db_world/fix-14033.sql
@@ -1,0 +1,2 @@
+UPDATE `quest_template` SET `QuestSortID` = 0 WHERE `ID` = 384;
+UPDATE `quest_template_addon` SET `RequiredSkillID` = 0, `RequiredSkillPoints` = 0 WHERE `ID` = 384;


heyitsbench:
In the `Issues Addressed:` section, please place `Closes` in front of the issue link, and if it's tied to a ChromieCraft issue, please try to include that one as well. For example:
```
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14033
- Closes https://github.com/chromiecraft/chromiecraft/issues/4524
```

Gultask:
We can't merge this, unfortunately.
QuestSortID are sniffed values. See: https://github.com/azerothcore/azerothcore-wotlk/pull/14034

It's probably an issue with the core.

sh3rp:
> We can't merge this, unfortunately. QuestSortID are sniffed values. See: #14034
> 
> It's probably an issue with the core.

I'm new here, so pardon my ignorance.

In the referenced PR, there is a single update to the `quest_template_addon` table which removes the skill requirement.  In this PR, I also removed the reference to the QuestSortID (which I believe is related to the quest DBC?) that references the cooking skill requirement.

Why can't that second part be executed?  What about "sniffed values" stops us from fixing them?  And I thought the quest_template modification was a Core modification?

heyitsbench:
> And I thought the quest_template modification was a Core modification?

Anything done in a SQL query is a change to the database.

> What about "sniffed values" stops us from fixing them?

Sniffed values are values taken *directly* from packets sent from Blizzard to the client (in this case, Wrath Classic iirc) so they are thus (for the vast, vast, vast majority) inherently correct.

Gultask:
Sniffed values are confirmed to be the true values used by Blizzard themselves. They are sent from the server directly to the client.

In this case:
`QuestSortID: -304` This is sniffed. We aren't supposed to change it since it's the value Blizzard themselves use and can be confirmed.

What I mean by "Core" is the cpp core of AzerothCore, not the DB itself.

sh3rp:
> Sniffed values are confirmed to be the true values used by Blizzard themselves. They are sent from the server directly to the client.
> 
> In this case: `QuestSortID: -304` This is sniffed. We aren't supposed to change it since it's the value Blizzard themselves use and can be confirmed.
> 
> What I mean by "Core" is the cpp core of AzerothCore, not the DB itself.
> 
> Edit: Almost got here before Bench😔

Got it.  I'll close this PR then and start digging around the Core.  This is a great opportunity to get familiar with the code base, I think.

Really appreciate your help @Gultask @heyitsbench.

Gultask:
Don't worry about it. And thanks for the contribution. It means a lot!❤️‍🔥

Issue #14728, opened 2023-01-21
Player should get Honorless target from hearth/portal/Gryphon


BeefSteaker:
https://github.com/chromiecraft/chromiecraft/issues/4844
### What client do you play on?

enGB

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

Player does not get honorless target from hearth/portal/Gryphon

https://user-images.githubusercontent.com/11332559/213868228-e8b27fc6-98d5-48fb-87c6-e2ca8ec861b2.mp4

https://user-images.githubusercontent.com/11332559/213868241-adfa136b-e0be-4297-8d8e-05c10c569793.mp4

### Expected Blizzlike Behaviour

Player should get honorless target from hearth/portal/Gryphon

https://user-images.githubusercontent.com/11332559/213868280-fc25e78f-d234-4854-b9d7-d8ba02630211.mp4

https://user-images.githubusercontent.com/11332559/213868255-06c76f80-85b7-41e2-b2cd-7cfb392b3c8e.mp4

### Source

[Patch 1.4.1 ](https://wowpedia.fandom.com/wiki/Patch_1.4.1)
> Honorless Target - Players who just land from a gryphon/wyvern, or step out of an instance will gain this buff for 30 seconds during which this target will give no contribution points if killed. This was originally implemented in the previous patch, but was kept invisible.

[wowhead comments](https://www.wowhead.com/wotlk/spell=2479/honorless-target#comments)
![image](https://user-images.githubusercontent.com/11332559/213868204-b6a09c6d-62e7-4b2c-8cdd-63480c302e3f.png)
![image](https://user-images.githubusercontent.com/11332559/213868213-299a10d2-b541-4730-8b8a-8f5658d89df8.png)

wrath classic

https://user-images.githubusercontent.com/11332559/213868280-fc25e78f-d234-4854-b9d7-d8ba02630211.mp4

https://user-images.githubusercontent.com/11332559/213868255-06c76f80-85b7-41e2-b2cd-7cfb392b3c8e.mp4

### Steps to reproduce the problem

.tele shattrath
use any portal

use any flight point

/use hearthstone

### Extra Notes

wrath classic seems to only have partial implementation as load screens do not trigger the buff. 
Probably as hardware is much better these days and load screens arent as big an issue.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b3e37226745a6e6de308bb7b00f0d2f9c596938e

[...]

PR #14731, opened 2023-01-21
fix(Core/Taxi): Correct cleanup for when the final destination is hit

Kitzunu:
* closes https://github.com/azerothcore/azerothcore-wotlk/issues/14728
* closes chromiecraft/chromiecraft#4844

## Changes Proposed:
-  Actually call the Cleanup CleanupAfterTaxiFlight() when the taxi is done
- Sneak in a correction for Honorless Target aura when teleporting or finishing taxi.

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Handlers/TaxiHandler.cpp
+++ b/src/server/game/Handlers/TaxiHandler.cpp
@@ -253,7 +253,7 @@ void WorldSession::HandleMoveSplineDoneOpcode(WorldPacket& recvData)
     {
         GetPlayer()->CleanupAfterTaxiFlight();
         GetPlayer()->SetFallInformation(GameTime::GetGameTime().count(), GetPlayer()->GetPositionZ());
-        GetPlayer()->CastSpell(GetPlayer(), 2479, true);
+        GetPlayer()->CastSpell(GetPlayer(), 2479 /*Honorless Target*/, true);
         return;
     }
 }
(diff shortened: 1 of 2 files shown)


src/server/game/Handlers/TaxiHandler.cpp
```diff
@@ -250,15 +251,11 @@ void WorldSession::HandleMoveSplineDoneOpcode(WorldPacket& recvData)
     // at this point only 1 node is expected (final destination)
     if (GetPlayer()->m_taxi.GetPath().size() != 1)
```
Kitzunu, line 252: this OpCode is triggered by vehicles too, which means this will always return true for vehicles...

src/server/game/Handlers/TaxiHandler.cpp
```diff
@@ -249,7 +249,7 @@ void WorldSession::HandleMoveSplineDoneOpcode(WorldPacket& recvData)
     }
 
     // at this point only 1 node is expected (final destination)
-    if (GetPlayer()->m_taxi.GetPath().size() != 1)
+    if (!GetPlayer()->GetVehicle() && GetPlayer()->m_taxi.GetPath().size() != 1)
```
Kitzunu, line 252: which doesn't help because it obviously triggers on leaving a vehicle too...

PR #14734, opened 2023-01-21
(apps/Db assembler) Re-make db_assembler script

pangolp:
## Changes Proposed:
-  Modify the current assembler system, since it is obsolete, and in reality, nobody uses it, because most of them enable the options in the emulator, so that it executes them for them. However, there is a small group of people, that sometimes, we prefer to correct the scripts manually, so it seemed to me a good idea, to generate a script, that allows only, to generate the files.

## Issues Addressed:

- Closes 

## Tests Performed:

- Windows 10

## How to Test the Changes:

1. Go to the folder `apps/db_assembler`
2. Execute the script.
3. Select menu options
4. The files should be generated within `env/dist/sql`

## Screenshots

menu

result


pangolp:
I am aware that this is a much less advanced version than the previous assembler, however, it would be interesting to have this alternative for those who want to continue uploading the scripts manually. Call me crazy, but I prefer to be watching the screen while that happens.

acidmanifesto (approved):
Nothing is wrong with having additional options.

Nyeriah:
1) The previous DB assembler relied on the file revision number in order to apply the files in the correct order. Is this addressed here?
2) Does the new assembler conflict with the auto populate routine? We also had issues with this and it would lead people to be stuck, again, previously, the import script had to take care of this but the import script had issues of its own

Kitzunu:
I have no issue with this.

But just to clarify, the DB Assambler will still be considered Deprecated and not a supported way to get SQL updates.

pangolp:
> 1. The previous DB assembler relied on the file revision number in order to apply the files in the correct order. Is this addressed here?
> 2. Does the new assembler conflict with the auto populate routine? We also had issues with this and it would lead people to be stuck, again, previously, the import script had to take care of this but the import script had issues of its own (it would choke on large SQL files and fail to apply)

1. This script, the only thing it does is to generate the files. It joins all the files that are in base/db_auth, and generates only 1, which is called auth_base.sql, so that you can execute them all together, instead of doing it 1 by 1. It doesn't even, apply them in the database, you have to do it manually.

2. No. It should not hinder the other system at all. Simply, it is an alternative to the previous system, more than anything, thought for those people, who want to stay in front of the monitor, watching how the file is uploaded, and in case there is a failure, maybe check it at the moment.

However, it has a disadvantage, and that is that the files must be well written. If, for example, someone forgets to leave a line break in updates/db_world, the file will be generated, but when uploading it, it will give an error. Because maybe the last line of a file is mixed with the next one. That is why it is important that the files always have a line break at the end. In addition, github by good practices, always asks for a line break at the end of your documents.

pangolp:
> I have no issue with this.
> 
> But just to clarify, the DB Assambler will still be considered Deprecated and not a supported way to get SQL updates.

I think it's very good. If you want to add a message that says: "we recommend the use of the automatic update system, added in the emulator", no problem. Just add that message and that's it. So that people know. However, those who want to not use that option, and prefer to do it manually, at their own risk, can have this alternative.

pangolp:
The `world_base`, I pass without problems. But the `world_updates` did not, so most likely... is that some file in the folder, is without a line break, as I mentioned before, so I am checking that.

world_base

world_update

Kitzunu:
Sounds good =)

pangolp:
```sql
Error occured at:2023-01-21 15:26:58
Line no.:146790
Error Code: 1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '/*------------------------------------------------------------------------------' at line 35484
```

I am looking for the line that caused the update to fail.
And then, we change the message, to warn that it is not the optimal solution.
That it is simply an alternative, for those who want to run their scripts manually.

Nyeriah:
> 2. No. It should not hinder the other system at all. Simply, it is an alternative to the previous system, more than anything, thought for those people, who want to stay in front of the monitor, watching how the file is uploaded, and in case there is a failure, maybe check it at the moment.

Just to be clear, the autoupdater already provides you with this. It shows every file as they are applied and throws you an error if there's any, halting the server afterwards. Furthermore, the error and what caused it is logged to the DBErrors file and can be consulted at any moment. It gives you the same output as a regular database manager application would.

I have to correct you on this as using this method will prevent you from using the autoupdater later on should you wish. The reason is, since there's no revision control, SQL files will be reapplied, and some of them simply can't (files changing DB structure for example). So it should be noted that those users will need to correct it themselves or wait until the next database squash in order to switch their update methods.

pangolp:
> > 2. No. It should not hinder the other system at all. Simply, it is an alternative to the previous system, more than anything, thought for those people, who want to stay in front of the monitor, watching how the file is uploaded, and in case there is a failure, maybe check it at the moment.
> 
> Just to be clear, the autoupdater already provides you with this. It shows every file as they are applied and throws you an error if there's any, halting the server afterwards. Furthermore, the error and what caused it is logged to the DBErrors file and can be consulted at any moment. It gives you the same output as a regular database manager application would.
> 
> I have to correct you on this as using this method will prevent you from using the autoupdater later on should you wish. The reason is, since there's no revision control, SQL files will be reapplied, and some of them simply can't (files changing DB structure for example). So it should be noted that those users will need to correct it themselves or wait until the next database squash in order to switch their update methods.

That is why it is an alternative. They are not compatible. The one who chooses this method, will always have to use this method. The one who uses the automatic emulator updater, will never use this tool.

pangolp:
The other moto, is to remove the db_assembler and only use the updates by the emulator automatically in a mandatory way. I am used to do it manually. So, I combine the SQLs and then run them.

Nyeriah:
Okay.

I have no objections to this, but unfortunately I'll second @Kitzunu when it comes to leaving the assembler as a non-recommended tool since its considerably worse than the autoupdater (both in user experience and maintenance, since it has no revision control and it may lead to issues).

pangolp:
> Okay.
> 
> I have no objections to this, but unfortunately I'll second @Kitzunu when it comes to leaving the assembler as a non-recommended tool since its considerably worse than the autoupdater (both in user experience and maintenance, since it has no revision control and it may lead to issues).

No problem. I will modify the menu later.
However, yesterday I tried it on Debian 11, and it worked.
Today I try it on Windows and it fails.

pangolp:
menu
@Kitzunu @Nyeriah I have added the warning message that was requested.
The script only combines the files.
For those who still want to add the files manually.

apps/db_assembler/db_assembler.sh
```diff
@@ -93,20 +93,22 @@ assembler()
 
 clear
 echo ""
-echo -e "\e[1;37m"
-echo "         Remember that the files are generated in the env/dist/sql folder."
-echo "         This tool is an alternative for those who prefer to execute SQL scripts manually."
+echo -e "\e[1;33m"
+echo "   We do not recommend using this script."
```
heyitsbench, line 97: I think it would be better to state that it's not the recommended method, this way it sounds like the tool just accepts that it's bad.
Instead of `We do not recommend using this script.`, I'd suggest: `This is not the recommended update method.`

pangolp: I accommodate it. Thanks for the suggestion.

pangolp: message

pangolp:
I just looked at the emulator's own assembler, and it doesn't even look like what I had seen. They've really done an amazing job with it. To such an extent, that I think this really wouldn't be that useful, however, if someone wants to merge the files, we could still give them the alternative to do so.

Yehonal:
Why do we need it? can't we just build the dbimport exec and run that one instead? it's a standalone tool that does exactly what the db_assembler did

pangolp:
> Why do we need it? can't we just build the dbimport exec and run that one instead? it's a standalone tool that does exactly what the db_assembler did

Yes, but the current db_assembler does not work. Moreover, they added a message that it is an obsolete tool. That's why I wanted to replace the script, for a simpler one, but that fulfills more or less the same function. I never understood very well the operation of the other db_assembler, and it seemed to me a good idea, since the current one is obsolete, to have an alternative to assemble the files.

pangolp:
Perhaps, it is more thought, in case you want to share the files to another person, and you don't want to pass file by file. Not having something to assemble them, you have to upload them to the DB, with the help of the emulator, and then make a backup. Actually, I don't have a concrete reason to have this tool. Something that justifies its use. It is simply an alternative, to assemble the files.

Yehonal:
What I mean is that now you have a tool called "dbimport" that is basically the db assembler but written in c++. So you do not have to run the entire server. You just need to compile the sources with the TOOLS enabled.
The db_assembler doesn't work anymore because it has been deprecated and can be even removed IMHO. 

I do not suggest to use SQL only because in this way the db_version table is not updated properly. I strongly suggest instead to run the dbimport and export the database with an SQL if you really need it

pangolp:
> What I mean is that now you have a tool called "dbimport" that is basically the db assembler but written in c++. So you do not have to run the entire server. You just need to compile the sources with the TOOLS enabled. The db_assembler doesn't work anymore because it has been deprecated and can be even removed IMHO.
> 
> I do not suggest to use SQL only because in this way the db_version table is not updated properly. I strongly suggest instead to run the dbimport and export the database with an SQL if you really need it

That's true, I tried it, and in fact, it doesn't even resemble what was there before. So you could basically delete this script and the previous one as well. The C++ tool, I tested it locally, because I could not upload the database, and it worked 10 points. However, in a server with Debian, the day before, I did not want to use it, because I did not know how it worked and as I saw in the Spanish chat, that the db_assembler was obsolete, I thought to change the script. But I agree, that you could directly remove the script and ready, so that people use the import of the emulator.

pangolp:
After using the emulator update, I realized that this really doesn't make sense. So it seems better to close it. In fact, this method should not be used, having the other tool, which is more complex, and which also keeps track of the SQL.

PR #14747, opened 2023-01-23
Fix configs in docker builds

michaeldelago:
## Changes Proposed
- This fixes an issue where the .conf files aren't properly generated for the docker build
  - With this issue, the dbimport service in the docker compose file would fail to start due to insufficient config. 
  - related to PR #14707
- create a ".user" file extension for custom configs in `env/docker/etc/`
- Layers the ".dockerdist" file and the ".user" file over the ".dist" file (in that order), and then copies the ".dist" file to the base ".conf" file
- Docker specific config changes go in the ".dockerdist" file and should still be committed to the repo
- User specific config changes go in ".user" file and should not be committed to the repo
- If the ".dist" file and the base ".conf" file are different, the ".conf" will be backed up before the ".dist" file is copied into it

## Issues Addressed:

- There isn't a specific issue, however this was reported and discussed in discord over the course of yesterday and today

## Tests Performed:

- Executing commands to build and start the game server
- after creating an account, I can log in and play without any issues. 

## How to Test the Changes:

1.  Execute commands
   
    ```bash
    $ ./acore.sh docker clean:build && ./acore.sh docker build:nocache && ./acore.sh docker start:app
    ```
2. Create account and log in as normal

- The largest issue that comes to mind is this is a breaking change for docker users (though, so was the change that caused this)
  - Docker users (and nobody else, of course) can no longer use the ".conf" for custom changes in docker setups. 
  - Their custom changes will have to be migrated to the ".user" file
  - Since the old ".conf" file is backed up, users should have minimal trouble copying the changes in the old ".conf" file to the new ".user" file

edit: link related PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,8 @@
 !/env/docker/data/.gitkeep
 !/env/docker/etc/
 /env/docker/etc/*
-!/env/docker/etc/authserver.conf.dockerdist
-!/env/docker/etc/worldserver.conf.dockerdist
-!/env/docker/etc/dbimport.conf.dockerdist
+!/env/docker/etc/*.conf.dockerdist
+!/env/docker/etc/*.conf.user
 !/env/docker/logs/.gitkeep
 /.env*
 /apps/joiner
--- a/.gitignore
+++ b/.gitignore
@@ -27,9 +27,6 @@
 /.env*
 /apps/joiner
 /deps/deno
-/data/sql/custom/*
-/src/server/scripts/Custom/*
-!/src/server/scripts/Custom/README.md
 
 /docker-compose.override.yml
 
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@
 !/env/docker/etc/
 /env/docker/etc/*
 !/env/docker/etc/*.conf.dockerdist
-!/env/docker/etc/*.conf.user
 !/env/docker/logs/.gitkeep
 /.env*
 /apps/joiner
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@
 /.env*
 /apps/joiner
 /deps/deno
+/data/sql/custom/*
 
 /docker-compose.override.yml
 
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@
 /apps/joiner
 /deps/deno
 /data/sql/custom/*
+/src/server/scripts/Custom/*
+!/src/server/scripts/Custom/custom_script_loader.cpp
+!/src/server/scripts/Custom/README.md
 
 /docker-compose.override.yml
 
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@
 /deps/deno
 /data/sql/custom/*
 /src/server/scripts/Custom/*
-!/src/server/scripts/Custom/custom_script_loader.cpp
 !/src/server/scripts/Custom/README.md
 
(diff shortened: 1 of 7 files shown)


michaeldelago:
In order to keep parity with non-docker setups, I've removed the ".user" file. It appears that backwards compatibility may have been missed with #14707; I'm sure it hasn't been a rare pattern for people to use `.conf` to only show their changes to `.dist`, and not touching dist at all.

for the sake of backwards compatibility, we could add something like this, after the `apps/compiler/includes/functions.sh:comp_compile` function merges in the `dockerdist` files:

```bash
# move all custom configs from conf onto a copy of conf.dist
cp "$base_conf.dist" "$base_conf.new"
conf_layer "$base_conf" "base_conf.new"
mv "$base_conf".new "$base_conf"
```

This could theoretically run over and over again, since it's only replacing non-default parameters from `.conf` to a copy of `.conf.dist`, and moving that file back to `.conf`

acidmanifesto:
Thank you for the PR.
i'm not 100% certain who is docker specialized outside of yehonal, so i requested a review from him for further evaulation.

Yehonal:
It seems too much complex for what we need, I would ask an opinion to @Foereaper as well since he's the author of that PR. Before diving into what happens here

michaeldelago:
@Yehonal to note on complexity

I think the complexity stems from the `conf_layer` function that this PR adds. While moderately more complex than other functions in the script, this function solves a few problems

- The docker-specific changes _need_ to be hardcoded somewhere, whether the script or in another config file
- We shouldn't hardcode docker-specific changes in the script
  - Hardcoding the config parameter changes in the script will make the script much longer
  - Hardcoding the config parameter changes in the script will make it less ergonomic to update them
- Docker-specific changes should be easy to find
  - leaving the `dockerdist` files in `env/docker/etc/` keeps this intuitive
- Reasoning and tracing how the configuration is modified during the build process should be easy

If you're fine with continuing this approach, do you have any suggestions on helping to clarify what the script is doing?

Yehonal:
Now I see what you're doing here. I suggest to add a bit more comments on your bash scripts since it's an advanced usage.

I've run the pipeline to test the scripts run on all the configurations. If so we can merge it.

I hope it's gonna address this issue as well: https://github.com/azerothcore/acore-docker/issues/12

Yehonal (approved):

Yehonal:
@michaeldelago after merging this PR we're now having an error with the ac-dev-server image: https://github.com/azerothcore/acore-docker/actions/runs/3992749270/jobs/6848864022

Looking at your code seems like this is caused by the fact you're not using `$confDir` anymore but the DOCKER_ETC_FOLDER instead.

We should create the `.conf` files under the `$confDir`. 

Can you explain this command? `base_conf="$(echo "$dockerdist" | rev | cut -f1 -d. --complement | rev)"`

michaeldelago:
@Yehonal `$base_conf` takes a dockerdist file and strips away the `dockerdist` from the filename, getting the "main" config file name. `cut -f1 -d. --complement` splits the a stream by `-d`, so in this case `.` and then `-f1` grabs the first field. `--complement` outputs the complement of that field, so it's everything _except_ that first field. 

```bash
$ dockerdist="./env/docker/etc/worldserver.conf.dockerdist"
$ base_conf="$(echo "$dockerdist" | rev | cut -f1 -d. --complement | rev)"
$ echo $base_conf
./env/docker/etc/worldserver.conf
```

PR #14757, opened 2023-01-25
feat(Core/AI): Move TaskScheduler to BossAI class

Nyeriah:
## Changes Proposed:
-  Move taskscheduler to the BossAI class
-  No longer requires importing the TaskScheduler header everywhere or instantiating it, simple use *scheduler*

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1. check if ayamiss/cthun still cast their abilities and such
2.
3.

- [ ] clean up every other boss file using the task scheduler
- [ ]

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
@@ -561,6 +561,7 @@ void BossAI::_Reset()
 
     me->ResetLootMode();
     events.Reset();
+    scheduler.CancelAll();
     summons.DespawnAll();
     if (instance)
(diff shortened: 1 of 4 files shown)


heyitsbench:
So, as my module uses TaskScheduler, would there be a different #include for that, and if so what would it be?

Nyeriah:
It only affects boss scripts, so if you are using a BossAI script, you no longer need to import TaskScheduler.h, it will be inherited from the BossAI class. Everything else remains the same e.g trash creatures and scripts of other nature

Kitzunu (approved):

Nefertumm (approved):
Highly appreciated!

PR #14423, opened 2022-12-30
fix(CORE): fix arm compile on macOS

motivewc:
## Changes Proposed:
-  Use ARM friendly replacements for AVX instructions, fixes argon2 compile.

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/3720

## Tests Performed:
- Built from source on macOS Ventura 13.2 on ARM Mac

## How to Test the Changes:
1. Build it on an M1 Mac 
2. ??
3. Profit

## Known Issues and TODO List:
No known issues, I have tested in-game.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #14761, opened 2023-01-26
feat(core) add macOS arm support

eyeofstorm:
## Changes Proposed:
!!! This PR based on PR #14423. !!! 
Thanks @motivewc for the original code.

-  Use ARM friendly replacements for AVX instructions, fixes argon2 compile.

## Tests Performed:

- Built from source on macOS Ventura 13.2 on M1 Mac.
- Built from source on Windows 11.

## How to Test the Changes:

Build and run.

No known issues.
Tested on M1 Mac and Windows 11.


acidmanifesto:
Adding run build label to see how this turns out.

acidmanifesto:
i am going to summon shin for this one as i am not a expert with anything with mac. but current workflow shows window passing which was not the case with the past closed pr.

Kitzunu (approved):

Kitzunu:
> i am going to summon shin for this one as i am not a expert with anything with mac. but current workflow shows window passing which was not the case with the past closed pr.

That is because this PR solves the issue by checking if _M_X64 is defined, which is part of MSVC

deps/argon2/argon2/blake2/blamka-round-opt.h
```diff
@@ -20,12 +20,17 @@
...
+#elif defined(__aarch64__) || defined(_M_ARM64)
+#include "sse2neon.h"
+#endif
+
 #if defined(__SSSE3__)
 #include <tmmintrin.h> /* for _mm_shuffle_epi8 and _mm_alignr_epi8 */
 #endif
 
-#if defined(__XOP__) && (defined(__GNUC__) || defined(__clang__))
+#if (defined(__x86_64__) || defined(_M_X64) || defined(_M_AMD64)) && defined(__XOP__) && (defined(__GNUC__) || defined(__clang__))
```
Kitzunu, line 33: Actually, why do you check the windows variable here?

Kitzunu:
CI passed and author tested it.
That is fine for me, as it is only adding a file in deps and doesnt change anything existing at all 👍

deps/argon2/argon2/blake2/blamka-round-opt.h
eyeofstorm: not only the windows. it check wether in x86_64 or not. linux will also hit this condition.

Kitzunu: yes but _M_X64 and _M_AMD64 are for MSVC and not linux, no? So why would you need to check for them here when we dont previously do that

eyeofstorm: Yes, you are right. x86intrin.h is for gcc/clang. MSVC uses intrin.h.
I will fix this. only __x86_64__ need to be check.

Kitzunu (approved):

Kitzunu:
Looks good to me

PR #14763, opened 2023-01-27
refactor[Core/reputation_spillover_template]: Reputation spillover table expansion

Alyst3r:
## Changes Proposed:
-  Update to `ObjectMgr::LoadReputationSpilloverTemplate()`, allowing two more factions to be added to `reputation_spillover_template` table. Allows more race-specific factions to have correct reputation gains from questing (like, for example Gilneas/Bilgewater Cartel that could be added in https://github.com/heyitsbench/mod-worgoblin without needs to do custom changes to ObjectMgr.cpp).

## Issues Addressed:
- None I know about.

## Tests Performed:
- Tested on Win10Home(22H2) but I assume it will work on any OS, it basically just builds upon pre-existing function. Tests done during playtesting of Faction.dbc bugfixes I've commited to Worgoblin module with backported Gilneas and Bilgewater Cartel from Cataclysm. As far as I can tell, everything was behaving correctly.

## How to Test the Changes:
1. Create new entry in `Factions.dbc` that has either alliance or horde as a parent faction (could just duplicate Stormwind/Orgrimmar with changed ID [and name to avoid confusion during playtesting]), then add said entry's ID to `reputation_spillover_template` in `acore_world` database under `faction5` (also fill `rate_5` and `rank_5` respectively): for alliance, for factions 47, 54, 69, 72 and 930, for horde factions 68, 76, 81, 530 and 911. Not 100% sure but `Factions.dbc` should probably be identical between server and client.
2. Create new character of any race that belongs to side you created new faction for and enter the world.
3. Do a couple of starting zone quests observing your reputation tab, if everything is done correctly, character should receive a fraction of quest reputation reward for every turn in. 

- Limits are still in place, if someone wanted to add more than 6 reputations, would still need to update core. I've made this PR mosty with Worgoblin module in mind, leaving a bit of leeway for additional race(s) (so, for example pandaren in the future).


heyitsbench:
Worth pointing out, Trinity master (Dragonflight) only has 5 faction/rate/rank columns, even including data for Gilneas/Bilgewater:
```sql
INSERT INTO `reputation_spillover_template` (`faction`, `faction1`, `rate_1`, `rank_1`, `faction2`, `rate_2`, `rank_2`, `faction3`, `rate_3`, `rank_3`, `faction4`, `rate_4`, `rank_4`, `faction5`, `rate_5`, `rank_5`) VALUES
(1133, 76, 0.25, 7, 530, 0.27, 7, 911, 0.25, 7, 81, 0.25, 7, 68, 0.25, 7),
(1134, 72, 0.25, 7, 47, 0.27, 7, 54, 0.25, 7, 930, 0.25, 7, 69, 0.25, 7);
```
I'm not sure this is the correct approach in terms of the general AC core, but even still, what is the reasoning for that 6th faction column?

Alyst3r:
Huh, interesting. I was using it to test pandaren factions. Tbh I have no way of checking how does it behave on retail currently, so I just assumed it would be something like this without digging through TC code. But I am pretty sure when I was playing back in Legion, questing in any alliance zone was also giving .25 rep to all other 6 reps, including Tushui Pandaren - maybe it is achieved a bit differently than this table?

acidmanifesto (changes requested):
image

Alyst3r:
Now that's embarassing, I did forget to commint this change for `SharedDefines.h`, should be resolved now.

acidmanifesto (approved):

data/sql/updates/pending_db_world/reputation_spillover_template-update.sql
```diff
@@ -0,0 +1,75 @@
...
+  `rate_4` FLOAT NOT NULL DEFAULT '0',
+  `rank_4` TINYINT UNSIGNED NOT NULL DEFAULT '0',
+  `faction5` SMALLINT UNSIGNED NOT NULL DEFAULT '0',
+  `rate_5` FLOAT NOT NULL DEFAULT '0',
+  `rank_5` TINYINT UNSIGNED NOT NULL DEFAULT '0',
+  `faction6` SMALLINT UNSIGNED NOT NULL DEFAULT '0',
+  `rate_6` FLOAT NOT NULL DEFAULT '0',
+  `rank_6` TINYINT UNSIGNED NOT NULL DEFAULT '0',
+  PRIMARY KEY (`faction`)
+) ENGINE=MYISAM DEFAULT CHARSET=utf8mb4 ROW_FORMAT=DYNAMIC COMMENT='Reputation spillover reputation gain';
```
Kitzunu, line 38: Why are you not altering the table instead of just recreating it?

Kitzunu (changes requested):

data/sql/updates/pending_db_world/reputation_spillover_template-update.sql
Alyst3r: Hmm, never thought about it, just went with the flow, should be resolved now ;].

Kitzunu:
I am generally against modifying the DB non-blizzlike for the sole purpose to allow a single module to do something. But at the same time, I might be the only one.

What happens when we have 100 different modules that want their niche changed into a table to make it work for them?

Nyeriah:
What is the status on this one?

Kitzunu (approved):

PR #14774, opened 2023-01-28
feat(Apps): Account creation script that isn't worldserver dependent

michaeldelago:
## Changes Proposed:
- Include a minimal account creation script 
- interacts with the database to create an account   
- Specify a gm level for the account being created
- Not dependent on the authserver or worldserver
- idempotent/declarative. Run it on every startup and the outcome won't change
- Faster out-of-the-box experience for new users and seasoned users alike
- minor shell (or PowerShell) scripting can heavily assist with creating managing many users as once

## Issues Addressed:
- There isn't a specific issue this addresses
- That said, I feel this is a feature that may help to improve the user experience of AzerothCore

## Tests Performed:
- generated many accounts, and logged in with them without any issues
- Created accounts and changed their password (by changing the `ACORE_PASSWORD` environment variable)
- After changing their password with AzerothCore console/gm command, running the script will reset the password to the desired value

## How to Test the Changes:
1. Script can be executed as follows:

```bash
$ MYSQL_HOST=mysql \
        ACORE_USERNAME="drekthar" \
        ACORE_PASSWORD="securepass22" \
        elixir account.exs
```

2. Log in with username "drekthar" and password "securepass22"
3. Run the script again.... and again
4. change the password, run the script
5. log in with new password

**Note** I consider these to be less of issues and more of things to consider about the script

- I wrote the script in `elixir`, which may be considered by some to be an _exotic_ language
  - In my opinion, the language itself is pretty simple and readable, but some things may not come immediately to some people
  - In order to help people execute the script, a Dockerfile is included for running the script without managing local dependencies
  - Additionally, there's a few examples within the `README.md` containing example commands for local, docker, and docker-compose based usage.
- I haven't tested this script on Windows or MacOS, because I do not own a Windows or MacOS machine. That said, elixir is a cross-platform, general purpose programming language, and I don't believe the code to be non-portable.
- This script will not delete users - that needs to be done manually. 


heyitsbench:
May I ask the appeal of a script like this?

michaeldelago:
> May I ask the appeal of a script like this? Primarily batch creation of accounts?

My response/usecase for this script is in the context of running the server through Docker, since that's what I use for running my server. 

At a very high level, the appeal is easier account management. The number itself isn't necessarily important; it's just a script that can be run for as many accounts as you may need. 

I personally put this in a similar boat as the dbimport program - a task that can be done by hand on startup, but it's a better user experience to be able to have it done automatically on startup. In the same vein as the dbimport, this is relatively idempotent and can be run with the same inputs and always it have the same result. 

Having to attach to the console and run a command to generate the initial account is clunky, and there's caveats that new users may find confusing and experienced users may find tedious.

Speaking from my personal anecdotal experience, many hosted applications allow for an administrator to create and manage users on startup through configuration parameters, as well. AzerothCore doesn't have any responsibility to behave like an enterprise software, but I appreciate the convenience of the feature.

Kitzunu:
I guess this is nice. My only question is if it should be added into this repo, or a stand-alone one. Like a module would be, but instead with the azerothcore-tools tag so it lands in https://www.azerothcore.org/catalogue#/home

acidmanifesto:
This in itself is not a far fetch thing to request or outside the realms of possibilities. Nearly every mmo out there has account creation handled to a arbiter service or world console of sort. Tera for instance has about 32 console windows, each handling a basic function of a mmo. Cities of heroes has about 15. At current wow emulation is about as extremely bare basic as bare basic can get. We put soo much responsibility and hate on the world server that compared to retail, we would be consider mad men for doing something insanely unorganized. Retail mmos would have nearly everything split up, map sercers, comms sercer, chat server, account management, account creation, account login, etc... Account creation when compared to other retail gaming as a service is its own thing.

acidmanifesto:
Marking as need feedback from other staffers. 
Imo this is a brave first step into transitioning into a attempt in splitting up stuff into its own services, although currently not its own service, i have seen zero attempts outside of cms for independent account creation currently

Maelthyrr:
I'm fine with the idea. Just two things:
- Is it possible to add a line to set gm level ? If you need always to use the command in worldserver to get the gm level, I don't see the benefit of this script.
- If you put this in master, it will be the same thing as docker. We need someone able to fix this if it's broken after an update.

Nyeriah:
There are issues with cached data server side if accounts are created by external tools

michaeldelago:
@Maelthyrr 
> Is it possible to add a line to set gm level ? 

Yeah, that's in there. Set GM Level with the `ACORE_GM_LEVEL` environment variable. If it's not set, it will default to 3. 

https://github.com/azerothcore/azerothcore-wotlk/pull/14774/files#diff-454c6c6e9999596f795529d758ad4b4c8d74796fe894dad0b09a30be0390a9acR45

https://github.com/azerothcore/azerothcore-wotlk/pull/14774/files#diff-454c6c6e9999596f795529d758ad4b4c8d74796fe894dad0b09a30be0390a9acR122-R134

> We need someone able to fix this if it's broken after an update.

I think this is a very apt concern. My point of view, in order to help mitigate this concern:

- The user isn't forced to use the script. Nothing is stopping someone from running `account create USER PASS` in the console. 
- While it's important to acknowledge that things break in unexpected ways, all of the ways I would expect this to break don't seem very likely
- The script is short (< 200 lines), and the language isn't complex. I'm confident that someone in the community would be able to fix it if they wanted it done.

Kitzunu (approved):

Kitzunu:
Thank you for your PR

Issue #11691, opened 2022-05-08
[Item] Unequipping Book of the Dead Despawns summoned minion.


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3480

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59
- [ ] 60

### Current Behaviour

Unequipping [[Book of the Dead]](https://www.wowhead.com/item=13353/book-of-the-dead#comments), which drops from Balnazzar in Stratholme, will despawn the skeleton. 

This is long known to not be the case. In classic even.

https://user-images.githubusercontent.com/100873791/167308301-8cd51e73-3591-4b21-b89c-0f650ef468be.mp4

### Expected Blizzlike Behaviour

Unequipping [[Book of the Dead]](https://www.wowhead.com/item=13353/book-of-the-dead#comments) will not despawn the summoned skeleton.

### Source

https://www.wowhead.com/item=13353/book-of-the-dead#comments:id=239392
![image](https://user-images.githubusercontent.com/32701428/167307061-523ce912-f099-41db-a083-d1cd5b877c02.png)

### Steps to reproduce the problem

Equip Book of the Dead.
Click it tosummon the Skeleton.
Unequip Book of the Dead.
Notice how the skeleton despawns.

### Extra Notes

Book of the Dead (Item ID: 13353)
[Summon Skeleton](https://wowgaming.altervista.org/aowow/?spell=17490#used-by-item) Spell ID: 17490
[Summoned Skeleton](https://wowgaming.altervista.org/aowow/?npc=11200) NPC ID: 11200

**Edit:** The same applies to the Ancient Cornerstone Grimoire from Onyxia. ([source for how it should work](https://www.wowhead.com/item=17067/ancient-cornerstone-grimoire#comments:id=740847))
Item ID: 17067
Spell ID: 17490
NPC ID: 11200

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/283ce18bb7b961264eae983a401aaacb89725273

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
[...]

PR #14839, opened 2023-02-01
fix(CORE/Player): Book of the Dead unequip fix

NinjaPleezAC:
## Changes Proposed:
-  Adds the Summon Skeleton effect of Book of the Dead to unequip exceptions list
-  Allows player to unequip Book of the Dead and have the summoned skeleton remain for its duration

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11691
- Closes https://github.com/chromiecraft/chromiecraft/issues/3480
- Also fixes Ancient Cornerstone Grimoire

## SOURCE:

Additional source material in the above issue links.
https://www.wowhead.com/item=13353/book-of-the-dead#comments:id=239392

## Tests Performed:

- Used active ability of Book of the Dead, swapped to a different offhand equip, confirmed summoned skeleton remained active.

## How to Test the Changes:

1. Create/Select character level 58+
2. .additem 17067
3. Equip Book of the Dead, wait until active ability is ready
4. Use active ability
5. Swap to any other offhand item or otherwise unequip Book of  the Dead
6. Observe summoned skeleton remains active
7. Observe summoned skeleton correctly disappears after its full 1minute duration

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -7008,8 +7008,8 @@ void Player::ApplyItemEquipSpell(Item* item, bool apply, bool form_change)
         // Spells that should stay on the caster after removing the item.
         constexpr std::array<int32, 2> spellExceptions =
         {
-            /*Electromagnetic Gigaflux Reactivator*/ 11826,
-            /*Book of the Dead - Summon Skeleton*/ 17490
+            11826,  //Electromagnetic Gigaflux Reactivator
+            17490   //Book of the Dead - Summon Skeleton
         };
         const auto found = std::find(std::begin(spellExceptions), std::end(spellExceptions), spellData.SpellId);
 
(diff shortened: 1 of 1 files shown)


NinjaPleezAC:
I'd like to note I'm not a huge fan of where this exception list is stored, and would be presenting a solution that moves it to data or at least outside of this method if I were more familiar with the codebase.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -7006,7 +7006,11 @@ void Player::ApplyItemEquipSpell(Item* item, bool apply, bool form_change)
             continue;
 
         // Spells that should stay on the caster after removing the item.
-        constexpr std::array<int32, 1> spellExceptions = { /*Electromagnetic Gigaflux Reactivator*/ 11826 };
+        constexpr std::array<int32, 2> spellExceptions =
+        {
+            /*Electromagnetic Gigaflux Reactivator*/ 11826,
+            /*Book of the Dead - Summon Skeleton*/ 17490
```
Kitzunu, line 7012: trivial, but the comment would be preferred to be after the spellID

NinjaPleezAC: Noticed that, wasn't sure whether to go with existing formatting or what. Fixed now.

Maelthyrr:
Tested and work fine. Thanks :)

Kitzunu (approved):

Issue #14912, opened 2023-02-09
The Tauren Chieftains do not seem to play in the World's End Tavern.


heyitsbench:
### Current Behaviour

There is no band playing in the World's End Tavern.

### Expected Blizzlike Behaviour

The Tauren Chieftains band should be playing in the World's End Tavern in Shattrath every so often, complete with music and effects. Shattrath Saul should also introduce the band ten-ish minutes before they play, walking around the lower city all the while.

### Source

Wrath Classic

### Steps to reproduce the problem

1. `.tele shattrath`
2. Go to the tavern.
3. Wait.
4. Observe no band.

### Extra Notes

The timing mentioned on multiple websites is every three hours every day, but this is not entirely true. I'm not sure what the timing actually is, it does seem to be every three hours with some exceptions. In my quick look on Classic, they played at six P.M. server/local time, but not at nine P.M. I don't know why, but there are definitely some exceptions or nuance to the timings.

I did sniff this, recording can be found [here](https://www.youtube.com/watch?v=G1VmGCCOhGY).

### AC rev. hash/commit

e7e119283f776746f34f409480c567e87f078483

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
All times noted are both local/server time (Eastern Standard Time).

Spotted on Wrath Classic at:
6 PM, Wednesday (2/8/23)
2 PM, Thursday (2/9/23)

heyitsbench:
I did double check just in case, this performance has no impact on the band placed in Silvermoon City, so there can be two copies of the band in two separate places.

heyitsbench:
It seems to me that the concerts played by L70ETC are all the same regardless of location. The BlizzCon concert seems to only be triggered hourly, the one in Shattrath's tavern happens every four hours (from 2AM), and the one in BRD triggers daily at 8PM and can be manually triggered by a character with a Direbrew remote. I have sniffed the latter two (obviously no way I can sniff the former), and I have listed out what I've observed the creatures doing during the performance [here](https://gist.github.com/heyitsbench/47df255b86d580c4b3def178437f1ae1). Some creatures (the controllers and the goblin creatures/drumset in BRD) are permanently spawned, the band members themselves are only spawned when the show begins.

PR #14919, opened 2023-02-09
fix(DB/Event): Implement L70ETC Concert

Kitzunu:
* closes https://github.com/azerothcore/azerothcore-wotlk/issues/14912

Co-Authored-By: Dr-J <1444120+dr-j@users.noreply.github.com>

## Changes Proposed:
-  Implement L70ETC Concert
-  Implement L70ETC Announcement

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

 Tested 3 years ago 👀 

1. .event start 81 / 82
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1675978583572906600.sql
+++ b/data/sql/updates/pending_db_world/rev_1675978583572906600.sql
@@ -1,11 +1,5 @@
 --
 
--- Delete old records of the event
-SET @OLDEVENT := 32;
-
-DELETE FROM `game_event` WHERE `eventEntry` = @OLDEVENT;
-DELETE FROM `game_event_creature` WHERE  `eventEntry` = @OLDEVENT;
-
 -- Implement L70ETC Concert
 SET @CGUID := 124091; -- Needs 13 consecutive guids
 SET @ANNOUNCEVENT := 81; -- Event for Announces from Shattrath Saul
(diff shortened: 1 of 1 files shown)


Gultask:
@heyitsbench You saw it live, maybe you're the best one to test this out? 👀

acidmanifesto:
Im not seeing any of the time emotes in this with each of the band members. They have special emotes that sync with the music.

heyitsbench:
Well right off the bat, I can guarantee this isn't blizzlike, as this would line up with what Wowpedia says, which is wrong. That being the occurrence every three hours. As observed on Wrath Classic, this is definitely incorrect. As for the actual scripting, I don't know that I'll be all too good at spotting special effects like that, especially since the performance I saw didn't have them entirely lined up with the song regardless. 🤷‍♀️

heyitsbench:
Alright, so in my observations the duration between ETC concerts seems to be four hours rather than Wowpedia's suggested three, same as Perry Gatner. I attached a recording of the performance to the original issue if anyone else wants to give it a test with a reference, but this is inherently an improvement from what currently (doesn't) exists. I'll try to give it a test when I have some spare time to compare the two, but it already exists in another core, so it can't be that bad. 😛

Maelthyrr:
Is it in WIP or finish ?

Kitzunu:
No time to fix it atm

PR #14914, opened 2023-02-09
Fix adduser/addgroup commands in dockerfile

michaeldelago:
## Changes Proposed:
-  adduser and addgroup commands in Dockerfile have `acore` hardcoded

## Issues Addressed:

- No issue found, but this was discussed in discord 

## Tests Performed:

- I haven't performed any tests

## How to Test the Changes:

1. Run command that should fail in previous setup
    ```bash
    $ docker compose --profile app build --build-arg DOCKER_USER=foobar
    ```

## Known Issues and TODO List:

- I need to test this change. I can't because I'm working right now. 

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

Issue #14644, opened 2023-01-13
Feature: Minimal Docker/Compose Infrastructure


michaeldelago:
### Describe your feature request or suggestion in detail

Hey all!

I've been looking through the [Docker build](https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/docker/Dockerfile) and the [docker-compose.yaml](https://github.com/azerothcore/azerothcore-wotlk/blob/master/docker-compose.yml), and it seems like the default Docker builds are geared around development rather than deployment. I think using containers for development is fantastic, but I also think the approach recommended in the [Docker setup](https://www.azerothcore.org/wiki/install-with-docker) guide should create an environment that's understandable to people who may not be familiar with AzerothCore, but are familiar with running containerized services.

Container builds (in the standard recommended method, `./acore.sh docker build`) result in a few quite large containers:

```bash
$ docker image ls --filter "reference=acore/*:master"
REPOSITORY                         TAG       IMAGE ID       CREATED      SIZE
acore/ac-wotlk-worldserver-local   master    ef7d43e00b0d   5 days ago   1.58GB
acore/ac-wotlk-authserver-local    master    ea120a17973c   5 days ago   1.58GB
acore/ac-wotlk-dev-server          master    bc00ce91a1c3   5 days ago   3.17GB
```

It's fair to note that container builds are layered, but I personally think that layering is a "nice to have" rather than something that should be used as a main point; some deployment environments (Namely those where the authserver and the worldserver could end up on separate servers) won't be able to benefit from layering. 

Going through the `Dockerfile` itself (as well as the `docker-compose.yaml`), a few things stand out to me that may be causing bloating the containers:

- All of `./data/sql` is included in the container build 
  - Increases container size by ~700MB
  - `/data/sql` is only used for a one time initialization job
- a Deno binary is included in the container build
  - Increases container size by ~100MB 
  - All build/run operations for the container can be performed without using the acore.sh dashboard
- The `servicebase` stage includes a large amount of [unneeded packages](https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/docker/Dockerfile#L126)
  - Increases container size by ~600MB
  - It appears that `gdb` is installed, along with the `-dev` versions of many of the library dependencies

Other things I noticed in the Docker build that could be removed to reduce complexity:

- simplify the compilation steps
  - many short shell scripts with includes that `source` other short shell scripts with their own includes, making it difficult at times to trace what exactly is happening
[...]

Issue #2486, opened 2019-12-03
doubt


elementalwow:
I need to know if there is how to put c ++ Script in the core Azerothcore, because I tried always gives errors.

Viste:
Yes you can! but you don't help you self! A good tone and good help to yourself would be to provide additional information:
What script?
Where you put script?
How you do it ?
What error you get?

FrancescoBorzi:
Please read: http://www.azerothcore.org/wiki/How-to-ask-for-help

PR #14990, opened 2023-02-13
feat(Docker): Clean up build and runtime

michaeldelago:
I can be contacted on discord as mynameismeat\#2486. Please feel free to reach out/tag me in a channel if you have any questions, comments, concerns.

## Goals
- Smaller docker containers
  - Less extraneous tooling
  - Merge various build targets into single target per component
  - Utilize multi-stage builds to ensure contents of containers are only runtime dependencies
- remove "prod" container distinction
- Enable users to get started with minimal effort
- wrap the worldserver in a way that safely runs `server shutdown` when the container is shutdown
- Allow for creation/management of an "admin account"
- Maintain compatibility with documentation and `acore-docker` within reason.
- Use docker's built-in caching for ccache management
- Remove the need (and the ability :p) from the user to access the console.

## Changes Proposed:
-  This is a large PR with many files changed.
- Files changed and description of changes involved
    - `apps/docker/Dockerfile`
        - Remove "prod" distinction
        - utilize multi-stage container builds
        - Split the `ac-dev-server` out into its' own dockerfile
    - `docker-compose.yml`
        - Remove "prod" distinction
    - `apps/docker/docker-cmd.ts`
        - Remove "prod" distinction
    - `apps/docker/client-data.sh`
        - Script to install client data
        - function extracted out of `apps/installer/includes/functions.sh`
    - `apps/docker/entrypoint.sh`
        - Starts azerothcore components
        - starts worldserver wrapper if desired
            - The wrapper script can create a user on startup and runs `server shutdown 5s 0` in the case of a SIGTERM
        - templates out config files
            - There was a few ways I wanted to do this, but to avoid spooky action at a distance it just templates the `.conf` in place.
    - `env/docker/etc/*.dockerdist`
        - Template out MySQL connection
        - disable logging to a file. Docker handles logging, logging to a file as well doesn't really do anything for us
    - `deps/boost/CMakeLists.txt`
        - Allow boost to be statically linked if `DOCKER` env var is defined.
        - Boost libraries (even non `-dev` versions) took a large amount of space in the final libraries
    - `apps/compiler/includes/functions.sh`
        - Revise for params that can include spaces
    - `apps/bash_shared_includes.sh`
        - Allow Deno to be option with `SKIP_INSTALL_DENO` env var

## Issues Addressed:

- I don't think it counts if I created the issue, but #14644
- There's a wrapper that creates an admin user, similar outcome to #14774 but not exactly the same
- Can replace #14914

## Tests Performed:

- Able to build, login, and play without issues. Modules seem to work without issues. 

## How to Test the Changes:

0. Destroy previous tooling with `docker compose --profile app --profile prod down -v --remove-orphans`
1. Start AzerothCore via command `docker compose --profile app up --build`
2. Observe new account being created in docker logs
3. Login with creds `admin` `admin` and confirm functionality
4. Run ctrl+c to stop the server. Quickly, check in game message logs showing the shutdown
5. Install a module (`git clone https://github.com/azerothcore/mod-transmog --depth 1 modules/mod-transmog`)
6. Rebuild with the command from `1`, it should be similar in speed to builds with ccache
7. Confirm module works

- the mysql client needs to be added to the runtime container, and it adds more space is needed imo. It'd be nice if we can ditch it entirely since `libmysqlclient` is needed anyway
- wrapping the worldserver isn't ideal. It'd be fantastic if we could 
   - create a user off of worldserver environment variables natively
   - Safely shutdown the server on SIGTERM/SIGINT
- Console access isn't necessarily available to the user. I personally don't think this is an issue, but I can see how it'd get in the way. Might be worth setting up SOAP or RA access by default.
- A health check would be great. Since I'm already using `expect` for wrapping the worldserver, it might not be too much of a stretch for it to access telnet and run `server info` or something


Yehonal:
I'm gonna review this soon, please wait for it

michaeldelago:
Ran into a small issue with config management. will need to work out but otherwise this should be fine for review.

Yehonal:
Why has the prod step been removed from the Dockerfile?

I've sent you a friend request on Discord to speed up the information gathering about this PR.

michaeldelago:
I've taken the process for dev, local, and prod images and refactored both build stages into a single setup.

Having all of the components of azerothcore, and then the dev, local, and prod images for each just increases confusion.

Yehonal:
> I've taken the process for dev, local, and prod images and refactored both build stages into a single setup.
> 
> Having all of the components of azerothcore, and then the dev, local, and prod images for each just increases confusion. 

The idea behind was to make sure that the images remains small and without extra files/libs when not needed.
Prod for instance, the image should not include all the libraries to compile

michaeldelago:
In the new Dockerfile, that's what happening in the `build` stage:

https://github.com/azerothcore/azerothcore-wotlk/blob/97382f05548f40f8f1cf16b7c5ae1b76dcbbe912/apps/docker/Dockerfile#L53-L113

If the operator wants to re-compile, they can run another `docker compose --profile up -d --build` (or `./acore.sh docker build && ./acore.sh docker start:d`) regardless of if they're in a dev or prod setup.

> Prod for instance, the image should not include all the libraries to compile

The current images don't abide by this. For both a "prod" or "dev" setup in the current docker builds with the `local` images, the `servicebase` stage installs source libraries as well as the entire `/data` directory, making the resulting images roughly 1.6GB. The non-local images copy over the right programs, so add on ~800MB for the worldserver and much less for everything else.

The images created by the dockerfile total to these sizes:

```bash
$ docker image ls | grep acore | awk '{print $1, $7}' | column -t
acore/ac-wotlk-worldserver  826MB
acore/ac-wotlk-db-import    197MB
acore/ac-wotlk-authserver   205MB
acore/ac-wotlk-client-data  93.9MB
```

55Honey:
@Yehonal friendly reminder to review this when you can.

apps/docker/client-data.sh
Yehonal, line 1: I do not like to have duplicated code that must be maintained eventually. Can't we put this code somewhere in a shared file and use it for both the installer and the docker? 

I'm also not really sure where this is used, I can't see anything calling this script. What is this for?

apps/docker/dev-server.Dockerfile
```diff
@@ -0,0 +1,146 @@
...
+
+# Do not use acore dashboard to install
+# since it's not cacheable by docker
+RUN apt-get update && apt-get install -y gdb gdbserver git dos2unix lsb-core sudo curl unzip \
+  make cmake clang libmysqlclient-dev \
+  libboost-system1.7*-dev libboost-filesystem1.7*-dev libboost-program-options1.7*-dev libboost-iostreams1.7*-dev \
+  build-essential libtool cmake-data openssl libgoogle-perftools-dev google-perftools \
+  libssl-dev libmysql++-dev libreadline6-dev zlib1g-dev libbz2-dev mysql-client \
+  libncurses5-dev ccache \
+  && rm -rf /var/lib/apt/lists/*
```
Yehonal, line 35: Wouldn't be smarter to have a single docker file to share the part of setting at least the OS configuration for the building part?

Since this is essentially the same between the dev server and the build process for the production images.

I'm always trying to avoid having lot of stuff to maintain, especially if we can't use the acore.sh here that would have been our single-source-of-truth for the dependency installation

apps/docker/docker-cmd.ts
```diff
@@ -38,11 +38,7 @@ shellCommandFactory(
 shellCommandFactory(
   "build",
   "Build the authserver and worldserver",
-  [
-    "docker compose --profile local --profile dev --profile dev-build build --parallel",
-    "docker image prune -f",
-    "docker compose run --rm --no-deps ac-dev-build bash apps/docker/docker-build-dev.sh",
-  ],
+  ["docker compose --profile app build"],
```
Yehonal, line 41: `--parallel`  and the `docker image prune -f` can still be beneficial, ain't they?

apps/docker/entrypoint.sh
```diff
@@ -0,0 +1,62 @@
...
+  cat << EOF 
+#==============================================================================#
+# You are using SAFE_WORLDSERVER mode.
+#
+# Be aware:
+#   - You CANNOT use the worldserver console.
+#   - All commands must be performed over SOAP or telnet.
+#
+# Please see [ https://www.azerothcore.org/wiki/remote-access ] for further info.
+#==============================================================================#
```
Yehonal, line 52: If we do not allow to use the console, these guides are not going to be valid anymore:

Acore project: https://github.com/azerothcore/acore-docker#access-the-worlserver-console-and-create-an-account

Access the worldserver: https://www.azerothcore.org/wiki/install-with-docker

Would it be possible to offer this as an option for those that really need to have the docker images in a safe production environment?

deps/boost/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ if(WIN32)
   set(Boost_USE_STATIC_RUNTIME     OFF)
 endif()
 
+if(DEFINED ENV{DOCKER})
+  set(Boost_USE_STATIC_LIBS ON)
+endif()
```
Yehonal, line 28: Can you please explain the reason for it?

Yehonal (changes requested):
It looks a great PR and I would like to invite you to keep maintaining the docker layer from now on, however, I would like first to solve the following points before merging it:

1. avoid duplicated code that also means more complexity to maintain everything once we upgrade dependencies, client data version etc.

2. We should possibly give the possibility to access the console by attaching to the container or change the guides everywhere on AC wiki and acore-docker README

3. I'm still not really sure what is the benefit of having 2 different docker files

apps/docker/client-data.sh
michaeldelago: This was called as the `CMD` for the docker image that handled client-data. I've since made the client-data container use the code from the current location `apps/installer/includes/functions.sh` with some of my changes added (though functionality hasn't changed)

apps/bash_shared/includes.sh
```diff
@@ -15,7 +15,7 @@ source "$AC_PATH_SHARED/common.sh"
 
 source "$AC_PATH_SHARED/deno.sh"
 
-denoInstall
+[ -z "$SKIP_INSTALL_DENO" ] && denoInstall
```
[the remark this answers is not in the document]
michaeldelago: This would prevent the install of deno if the `SKIP_INSTALL_DENO`.

In more recent commits, it's been removed.

apps/docker/dev-server.Dockerfile
michaeldelago: With my latest commits, I've removed the dev-server, as I have a hard time figuring out where it fits in.

On my machine rebuilds are always less than a minute with the way the docker build stages are set up. If code changes, docker ONLY re-compiles the programs and adds it into the proper container

1. skeleton
  a. build
  b. runtime
    i. authserver # These 3 get binaries from the build step
    ii. worldserver
    iii. db-import
  c. client-data

apps/docker/entrypoint.sh
michaeldelago: Yeah, I'm conflicted on it. Personally, I think that the console in the context of docker can counteract the benefits of docker, but as it stands the console is an essential component.

Updating documentation is an easy fix, however. 

I can make a small script that works like a console so there's transparent functionality to using the repl. Python's stdlib has a [Cmd module](https://docs.python.org/3/library/cmd.html) that'd probably make this pretty easy.

deps/boost/CMakeLists.txt
michaeldelago: Statically compiling boost into the binaries rather than installing them in the runtime container saves a good amount of disk space

apps/docker/docker-cmd.ts
michaeldelago: `--parallel` isn't un-helpful, but the build step only happens in one container and then the artifacts get copied to multiple containers, making that build step (which also takes the longest) a bottleneck. 

I'm pretty sure `docker compose build` and `docker compose up --build` actually are parallel by default? don't quote me on that :D

`docker image prune -f` can affect other containers/projects on the computer that I wouldn't want to mess with

apps/docker/dev-server.Dockerfile
Yehonal: dev server is used for the vscode devcontainer setup, where you can work within a docker container directly

Yehonal:
I'm sorry but this PR can't be merged if the ac-dev-server is removed, this is a feature documented here:

https://www.azerothcore.org/wiki/install-with-docker#how-to-use-the-dev-container

apps/docker/entrypoint.sh
FrancescoBorzi: > Would it be possible to offer this as an option for those that really need to have the docker images in a safe production environment?

yeah that would be nice, I'd prefer to keep this possibility

Yehonal: @michaeldelago we have a `acore-docker` project (https://github.com/azerothcore/acore-docker)  that is intended to be a simple docker-compose, that uses the images deployed by the azerothcore-wotlk repo, to spin up the azerothcore without knowing anything else than how to install docker. 

The goal is to keep the testing of azerothcore simple and stupid for those people that want to give AC a go with zero-knowledge or even for bug triagers etc.
These people do not want to follow complex steps to setup a telnet connection and just want to attach their terminal to the container console. 

That project is cloned 30 times per day and is pretty essential for us so we need to support this ability keeping the documentation untouched or slightly changed without increase the complexity.

apps/docker/docker-cmd.ts
Yehonal: Apparently it isn't by default: https://ardalis.com/speed-up-docker-compose-with-parallel-builds/

You can measure the time with that option or without

apps/docker/client-data.sh
Yehonal: I see that you have merged the code back into the functions.sh, I guess we can resolve this comment

apps/docker/dev-server.Dockerfile
Yehonal: I wouldn't even have a separated db if possible, there's not really a reason for that I guess

apps/docker/docker-cmd.ts
michaeldelago: Going off of the documentation from docker compose, there is "unlimited" parallelism by default

https://github.com/docker/compose/blob/00f72cb553b6d35b6f3e18830ea8f2f5f785faac/docs/reference/compose.md?plain=1#L47

https://docs.docker.com/engine/reference/commandline/compose/#options

apps/docker/entrypoint.sh
michaeldelago: To clarify, the behavior that we're talking about is optional and opt-in. The images that are generated work in the way the documentation describes, though there is an extra mode for running the worldserver that has some properties that some people may find to be useful.

The lines referenced that shows a brief warning when using `SAFE_WORLDSERVER` mode. `SAFE_WORLDSERVER` mode controls the worldserver through an `expect` script, as well as creates an admin account for the server. If the container receives a SIGTERM, this expect script will issue a `server shutdown 5s 0` command to the worldserver. 

`SAFE_WORLDSERVER` mode is only activated if the operator has set the `SAFE_WORLDSERVER` environment variable, seen [in the entrypoint](https://github.com/michaeldelago/azerothcore-wotlk/blob/docker-multi-stage/apps/docker/entrypoint.sh#L10) and [the compose file](https://github.com/michaeldelago/azerothcore-wotlk/blob/docker-multi-stage/docker-compose.yml). Without `SAFE_WORLDSERVER`, the worldserver will operate in the exact same way that many people are used to.

I will note that I did just remove the default setting of "true" in the compose file, though that's doesn't change the operations for https://github.com/azerothcore/acore-docker

michaeldelago:
My largest reservation on the code for devcontainer is that I don't like the code specifically concerned with development to be so closely integrated with something I expect to run in production. If I want to edit something in the compose file for the server I run, I have to scroll through what is essentially noise to reach the parts I'm looking for.

My understanding is that it's relatively common convention to have the .devcontainer directory contain its' own compose file and Dockerfile. Can we reach a compromise where the dev server's docker setup lives in .devcontainer, and the repo's main compose file and dockerfile are for running and building the application in a non-dev environment?

Yehonal:
> My largest reservation on the code for devcontainer is that I don't like the code specifically concerned with development to be so closely integrated with something I expect to run in production. If I want to edit something in the compose file for the server I run, I have to scroll through what is essentially noise to reach the parts I'm looking for.
> 
> My understanding is that it's relatively common convention to have the .devcontainer directory contain its' own compose file and Dockerfile. Can we reach a compromise where the dev server's docker setup lives in .devcontainer, and the repo's main compose file and dockerfile are for running and building the application in a non-dev environment?


The reasons why I would keep everything in one single file are the following:

1. Easier to maintain - you update the "skeleton" stage once. We should avoid adding too many places where to maintain the OS configurations and keep it centralized. It's ok if we can't use the acore.sh dashboard for caching purposes (although we might find a solution for it as well), but at least let's try not to add even more code duplication

2. Easier to test the compilation in our CI and use the acore-dashboard to build and run the dev container without adding complexity. I personally use the dev container both with vscode and intellij (https://www.jetbrains.com/help/idea/docker.html). If we put the dockerfile into the vscode folder then we remove the possibility to use the same dockerfile/docker-compose container for other IDEs/dev environments. 

To solve the "Dockerfile mess" I would just suggest to put everything regarding the dev part at the bottom...e.g.:

- skeleton
    - build
    - runtime
       - authserver # These 3 get binaries from the build step
       - worldserver
       -  db-import
    - client-data
    - dev
    
Where dev inherit from the skeleton + it adds what is missing to have a complete dev environment

Same thing for the docker-compose, where the dev container can be specified at the end and it just reuses the production db. I do not see the need for a different database, and if someone needs that, it's always possible to create the `docker-compose.override.yml` file

apps/docker/entrypoint.sh
Yehonal: ah perfect then, if you need to enable the SAFE_WORLDSERVER then we can resolve this conversation. Can you also prepare the PR to update our wiki please?

Yehonal:
Any news here? Do you have ETA? Need further info/help from me?

michaeldelago:
Sorry, I've been poking it here and there but haven't gotten a commit it until just now.

> Easier to maintain - you update the "skeleton" stage once. We should avoid adding too many places where to maintain the OS configurations and keep it centralized. It's ok if we can't use the acore.sh dashboard for caching purposes (although we might find a solution for it as well), but at least let's try not to add even more code duplication 

I'm sorry to say that I disagree. I don't think that having a main Dockerfile for the actual components and a separate Dockerfile given to contributors if they'd like to use devcontainer is too much of an extra burden on maintaining them. If the devcontainer Dockerfile was more complex it'd be a harder choice for me, but as it stands I prefer to have the mental separation of the Dockerfiles over strict de-deduplication. 

> If we put the dockerfile into the vscode folder then we remove the possibility to use the same dockerfile/docker-compose container for other IDEs/dev environments.

This is incorrect. I can run `docker compose -f .devcontainer/docker-compose.yml run ac-dev-server bash` for the same experience. I don't use VSCode or Jetbrains IDE's (typically stick to vim) but I imagine as long as the user can configure a compose file they should be good to go regardless of environment. 

> Easier to test the compilation in our CI and use the acore-dashboard to build and run the dev container without adding complexity. 

I haven't messed with it in a while but if I remember correctly, I was able to cut out a non-insignificant portion of the GHA workflows, but I can definitely revisit those and make sure they're not doing anything crazy.

> the dev container can be specified at the end and it just reuses the production db. I do not see the need for a different database

My point of view for this is that it's not that there's a different database, it's that either docker-compose.yml files have the services (just mysql, in this case) to host the game server.

apps/docker/entrypoint.sh
michaeldelago: Yeah, I can do that

FrancescoBorzi:
@Yehonal can you please check this again?

michaeldelago:
Closing as I don't see a high chance of this getting merged. 

I may revisit this in the future, though it'd probably be much better as a collection of smaller PR's rather than one big one.

Issue #6639, opened 2021-06-29
Wetlands Dragonmaw Bonewarder missing Skeleton Minion


Samsequel:
**CURRENT BEHAVIOUR:**

Sometimes the Dragonmaw Bonewarder has a skeleton, but usually they despawn during fight or after aggro reset. They won't return

**EXPECTED BLIZZLIKE BEHAVIOUR:**

At all times (unless killed during combat) a skeleton should be summoned by them

**SOURCE:**

https://wowpedia.fandom.com/wiki/Dragonmaw_Bonewarder

**STEPS TO REPRODUCE THE PROBLEM:**

1. .go creature 9738
2.  Observe that he has no minion, or when he has, he will despawn during combat.
3.  Observe that a lot of other Bonewarder's do not have a minion with them.

**EXTRA NOTES:**

**AC HASH/COMMIT:**
 
b2aa4f3f94b9

**OPERATING SYSTEM:**

Windows 10 
**MODULES:**

-
**CUSTOMIZATIONS:**

-

PR #15082, opened 2023-02-17
fix(DB/Creature): Wetlands Dragonmaw Bonewarder missing Skeleton Minion

avarishd:
## Changes Proposed:
- Dragonmaw Bonewarders to respawn Skeleton on Reset (After Combat, On Respawn or Spawn) rather than only on respawn.
"At all times (unless killed during combat) a skeleton should be summoned by them"
- Dragonmaw Bonewarder's pet (Skeleton) to force despawn after being dead for 1s. (if the body is not despawned OR there is an active pet. when Bonewarder resets, he will **NOT** summon a new one).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6639

## Tests Performed:

- tested in-game

## How to Test the Changes:

Dragonmaw Bonewarder & Skeleton
.go c id 1057
Observe that they have minions and that they're not despawning OOC, kill their skeletons and then reset the Warder (after waiting the 1s for the skeleton's corpse to despawn) to see if they spawn a new one.

- [ ][ In classic, Skeletons were despawning when their owner was killed](https://youtu.be/9udzwb7BeJ0?t=1398), which presumably was not the case for original release. (nothing in SummonProperties dbc)
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1676624876337239900.sql
+++ b/data/sql/updates/pending_db_world/rev_1676624876337239900.sql
@@ -1,12 +1,3 @@
--- Infected Kodo Beast - add missing ability, Stampede.
-DELETE FROM `creature_template_spell` WHERE `CreatureID`=25596;
-INSERT INTO `creature_template_spell` (`CreatureID`, `Index`, `Spell`, `VerifiedBuild`) VALUES
-(25596, 0, 45876, 12340),
-(25596, 1, 45877, 12340);
-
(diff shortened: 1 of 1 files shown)


Nyeriah:
Thanks for the PR, however, please keep 1 fix per PR unless they're strictly related. This can't be merged as is.

Nyeriah (changes requested):
Issues that are not related to each other need their own PR

Nyeriah (approved):

Issue #10827, opened 2022-02-26
Sha'ni Proudtusk not despawning


ishidinho:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [X] 50-59
- [ ] 60

### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/3106

Quest [Dreadmaul Rock](https://classic.wowhead.com/quest=3821/dreadmaul-rock), done in Burning Steppes, requires us to click on the remains of a troll – on top of an altar– which will proc a ghost next to them, completing the quest and offering us a new one.
However, this NPC doesn't seem to de-spawn after the event takes place and, upon clicking on the corpse, we get to see "our" ghost plus all the previous ones summoned by other players, stacking by by the tens.
![ 2022-02-19 at 19 02 31 2](https://user-images.githubusercontent.com/77953030/154838221-04441b55-1456-4ca1-a989-53812570d169.png)

Tester:

The issue is confirmed.

![WoWScrnShot_022622_095253](https://user-images.githubusercontent.com/62797536/155836948-ae23d25d-0bfe-4dbe-98e2-ce0617690e6d.jpg)

### Expected Blizzlike Behaviour

Clicking on Sha'ni remains should proc ONE ghost, that should eventually despawn after is visible to us, so other players don't get it as well.

### Steps to reproduce the problem

Original report:
Have a Horde character at level 48 or more
Get quest 3821
Locate altar at coords 80,45 (aprox) on Burning Steppes
Click on Sha'ni remains
Observe npcs stacking, where it should be just one

Tester:
1. `.quest add 3821`
2. `.go gameobject 4523`
3. Click Sha'ni Proudtusk's Remains

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/f82c60526b9955c697b1ee804a8e7613f62c1568

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
[...]

PR #15074, opened 2023-02-17
fix(DB/Quest): Kill'em With Sleep Deprivation/Look At the Size of It 

Gultask:
Authored-By: Dr-J [1444120+dr-j@users.noreply.github.com](mailto:1444120+dr-j@users.noreply.github.com)

## Changes Proposed:
-  Ports the above quest events from TC
-  

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/6773
- Closes https://github.com/chromiecraft/chromiecraft/issues/1104

## SOURCE:

TC

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #15094, opened 2023-02-18
fix(DB/Conditions): Sha'ni Proudtusk infinite spawning. 

avarishd:
## Changes Proposed:
- (AND) Add condition to not spawn NPC if there is already an NPC within 10 yards.
- (AND) Add QUEST_STATE (Rewarded and Complete) condition that allows to spawn NPC after completing 3821

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10827
- Closes https://github.com/chromiecraft/chromiecraft/issues/3106

## Tests Performed:

- tested

## How to Test the Changes:

.q a 3821
.go gameobject id 160445
Spam click it while having the quest in your log AND after turning it in.
Notice that there will always be 1 NPC, and no more.

- [x] Unknown if you should be able to summon her AFTER you complete 3821. In case you don't accept the quest that she starts and she despawns, resulting in you having to wait for another person to summon her.
- [x] If you talk with the npc, someone can potentially spam click to summon/despawn to disrupt you.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1676673725150035900.sql
+++ b/data/sql/updates/pending_db_world/rev_1676673725150035900.sql
@@ -1,12 +1,5 @@
--- Sha'ni Proudtusk - On summon, despawn others.
-UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 9136;
-
-DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 9136);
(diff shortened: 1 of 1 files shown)


Gultask:
Instead of despawning you can set up a condition to only summon Sha'ni if there isn't one already nearby since the gob uses SmartAI

Here's an example and where I learned that from: https://github.com/azerothcore/azerothcore-wotlk/pull/15074/files#diff-85c028f95a71a196147203afc49222f96e2f09a877d87707e4b80798428d111dR115

It'd solve your issue in the TODO

Nyeriah:
Could also use this to make the condition simpler 
image

avarishd:
This PR is actually "old" and in these couple of days I've learned quite a bunch about conditions.

avarishd:
@Gultask

Gultask (approved):

Maelthyrr:
Hi, thanks for the PR. Tested with two char:

- [X] Spam clic will spawn only one npc
- [X] If another player use the gob, it doesn't spawn a new npc
- [X] When the npc despawn, the player can respawn her by clicking again on the gob
- [X] If a player doesn't have the quest, he can't summon the npc

Issue #14944, opened 2023-02-11
boss "Kirtonos the Herald"   sommoned out of instance_scholomance. never show up.


jerrytt:
### Current Behaviour

Quest - http://old.wowhead.com/quest=5384
Item to summon him - http://old.wowhead.com/item=13523
BOSS: https://www.wowhead.com/npc=10506/kirtonos-the-herald

src/server/scripts/EasternKingdoms/Scholomance/instance_scholomance.cpp
Position KirtonosSpawn = Position(315.028, 70.5385, 102.15, 0.385971);

on my server this Position is out of  instance_scholomance . 
and after give the task item , boss never show. 

.go xyz  315.028 70.5385 102.15 289
this cmd will teleport out of instance .

### Expected Blizzlike Behaviour

boss should show

### Steps to reproduce the problem

.go xyz  315.028 70.5385 102.15 289
.additem 13523

click the fire. boss never show up.

### AC rev. hash/commit

AzerothCore rev. 0ed7813e84f5+ 2023-02-05 00:05:20 +0100 (master branch) (Unix, RelWithDebInfo, Static) (worldserver-daemon) ready...
Calendar deletion of old events.
Starting up Auction House Listing thread...
AC> .server debug
AC> AzerothCore rev. 0ed7813e84f5+ 2023-02-05 00:05:20 +0100 (master branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 3.0.7 1 Nov 2022 (library: OpenSSL 3.0.7 1 Nov 2022)
Using Boost version: 1.81.0
Using MySQL version: 100904
Using CMake version: 3.25.2
Compiled on: Linux 
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in ./data/maps. Total size: 291014951 bytes
vmaps directory located in ./data/vmaps. Total size: 658130721 bytes
mmaps directory located in ./data/mmaps. Total size: 2192910844 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 
Using World DB: ACDB 335.9-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0

### Operating system

archlinux

patou01:
Can confirm the issue on latest.

[...]

PR #15131, opened 2023-02-19
Fix (scripts/Scholomance) - Kirtonos the Herald event now starts correctly again

patou01:
The root cause of the issue is that when a creature is summoned by an instance, the ``IsSummonedBy()`` method is not called anymore. 

Additionally, I noticed that he's swimming through the air, so I checked how it's done for Onyxia and tried to copy from there.

## Changes Proposed:
-  Use `DoAction` on the summoned creature (Kirtonos). Reuse the `IN_PROGRESS` define because it doens't hurt..
- Add NO_MOVE_FLAGS_UPDATE, I copied the query from Keira3, I don't recall how they're supposed to be made if it should OR or actually use absolute values.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14944

## Tests Performed:

- Ran it on my own computer, the boss seems to start moving and not swim in the air in humanoid form.
-  I did notice earlier on that he doesn't always start, so maybe we want to have a few tries to be sure.

## How to Test the Changes:

1. .teleport scholomance
2. Enter the dungeon
3. .go xyz 325.122009 93.6436000 101.637001
4. .add 13523
5. right click the brazier
6. Kirtonos should spawn and start moving (flying around) and then come to the place where you are. He should behave normally after that, no swimming in the air.


src/server/scripts/EasternKingdoms/Scholomance/boss_kirtonos_the_herald.cpp
```diff
@@ -128,6 +118,25 @@ class boss_kirtonos_the_herald : public CreatureScript
             }
         }
 
+        void SetData(uint32 type, uint32 data) override
```
Nyeriah, line 121: In that case, you could also use DoAction()

src/server/scripts/EasternKingdoms/Scholomance/instance_scholomance.cpp
```diff
@@ -131,7 +131,11 @@ class instance_scholomance : public InstanceMapScript
                             // summon kirtonos and close door
                             if (_kirtonosState == NOT_STARTED)
                             {
-                                instance->SummonCreature(NPC_KIRTONOS, KirtonosSpawn);
+                                TempSummon* sum = instance->SummonCreature(NPC_KIRTONOS, KirtonosSpawn);
+                                if(sum)
+                                {
+                                    sum->AI()->SetData(DATA_KIRTONOS_THE_HERALD, IN_PROGRESS); // Set the status of the boss to start the event
+                                }
```
Nyeriah, line 138:
```suggestion
                                if (Creature* kirtonos = instance->SummonCreature(NPC_KIRTONOS, KirtonosSpawn))
                                {
                                    kirtonos ->AI()->SetData(DATA_KIRTONOS_THE_HERALD, IN_PROGRESS); // Set the status of the boss to start the event
                                }
```

src/server/scripts/EasternKingdoms/Scholomance/boss_kirtonos_the_herald.cpp
patou01: Done, didn't know it existed.

Nyeriah (approved):

Nyeriah:
Thanks for the PR 👍

Issue #11975, opened 2022-06-07
Game breaking Mind Control/Charm bugs


T1ti:
### Current Behaviour

Most mind control spells have no cooldown and GCD, only spell data is used.
You can mind control creatures and **spam their spells without any cooldown**, which is completely broken, you can solo dungeons easily with mind controlled creatures etc by spaming powerful spells.

### Expected Blizzlike Behaviour

**Mind controlled creatures spells sould have the same cooldowns as the creature has**

### Source

Some classic videos
https://youtu.be/1yQ6xu1c4wc?t=39
https://www.youtube.com/watch?v=P9ZPv8k6tXc
https://youtu.be/4RgAkOqNqAo?t=28 (With gnomish MC cap, it's 100% classic footage because of the clock and HD shadows)

Finding videos of it in PVE on youtube is extremly hard but it's very common knowldge they have cooldowns, or they could be abused easily, anyone can test on classic

### Steps to reproduce the problem

1: Use any mind control/charm effect on a creature
2: test if their abilities are spamable (abilities that should have cooldowns like a shield bash), you can specifically try the scarshield sppellbinder in the videos above, it's an extremly popular mob to mind control for the buff

### Extra Notes

Probably the same issue with the warlock's Enslave demon and other Charming effects.

This is an extremly tough problem to solve as AC/TC chose to use a global scripting system for creatures spells (smart AI), currently the creature_tempalte_spells structure is extremly basic,n it's just creature id x spellid.

Solutions would be : 

- Either get that data from the smart ai somehow by filtering their spell cast actions
- Easier solution : Import the vmangos system and data, which is an extended version of the current creature_spell_template table with additional columns for cooldowns and such (they also use this data for normal creature spells system)

![image](https://user-images.githubusercontent.com/40864460/172289437-b5f7de22-4e62-41b7-b038-ef5783a09c6e.png)

Also note that on classic the tooltip updates with the cooldown after cast
![image](https://user-images.githubusercontent.com/40864460/172290042-6cc38b68-5785-464a-8b82-1a69f1096312.png)

![image](https://user-images.githubusercontent.com/40864460/172290016-11a19eab-7d81-43ef-9e81-4eb8c9c1a3e1.png)

### AC rev. hash/commit

70f3891352744ede66b12c3e1b2e52d3d5794564

### Operating system

windows 10 x64

thenewblock:
I can confirm. This is extremely game breaking and needs to be fixed ASAP.

[...]

PR #15143, opened 2023-02-20
fix(Core/Cooldown): Implement spell cooldown overrides to address cha…

Nyeriah:
…rmed creature spell cooldowns

## Changes Proposed:
-  Implement spell_cooldown_overrides to allow adding cooldown overrides with ease, for creature charmed cooldowns
-  Add the cooldown overrides to the table to fix charmed creatures not having cooldowns

## Issues Addressed:

- Updates https://github.com/azerothcore/azerothcore-wotlk/issues/11975#issuecomment-1436315862

## Tests Performed:

- 
image

- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1676897566022575900.sql
+++ b/data/sql/updates/pending_db_world/rev_1676897566022575900.sql
@@ -1,0 +1,9 @@
+--
+CREATE TABLE IF NOT EXISTS `spell_cooldown_overrides` (
+  `Id` INT UNSIGNED NOT NULL,
+  `RecoveryTime` INT UNSIGNED NOT NULL DEFAULT '0',
+  `CategoryRecoveryTime` INT UNSIGNED NOT NULL DEFAULT '0',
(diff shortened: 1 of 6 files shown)


Gultask (approved):

data/sql/updates/pending_db_world/rev_1676897566022575900.sql
```diff
@@ -0,0 +1,9 @@
+--
+CREATE TABLE IF NOT EXISTS `spell_cooldown_overrides` (
```
Kitzunu, line 2: DROP TABLE IF EXISTS `spell_cooldown_overrides`

Nyeriah: For the record, it doesn't matter because the table is empty so this SQL is safe to be ran twice, the conditional will prevent errors

Issue #12500, opened 2022-07-23
Serpentbloom Snakes 


banjanja13:
### Current Behaviour

When looting Serpentbloom a stationary Serpentbloom Snake spawns

### Expected Blizzlike Behaviour

This Serpentbloom Snake should move about wildly and erratically

### Source

https://classic.wowhead.com/npc=3680/serpentbloom-snake#comments
https://user-images.githubusercontent.com/50233983/179848313-3193f0ff-5f57-4a59-b593-970ee3d98599.png
https://youtu.be/YSXFkzfF8rg?t=799

### Steps to reproduce the problem

1. Take the quest
2. Go to wailing caverns and pick a serpentbloom

### Extra Notes

Original issue: https://github.com/chromiecraft/chromiecraft/issues/3834#issue-1310056500
My reproduction of the problem: https://youtu.be/JCa6lJIjLWk
My sources to back up the correct behavior: https://www.youtube.com/watch?v=e0HWQWEa2Ek

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/35dfcf588f62339c44ff99a3e7e85f0b983ed923

### Operating system

Ubuntu 20.04

### Custom changes or Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[...]

PR #15151, opened 2023-02-21
fix(DB/SAI): Serpentbloom Snakes no movement.

avarishd:
## Changes Proposed:
- On spawn add random movement

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/12500
- Progresses https://github.com/chromiecraft/chromiecraft/issues/3834#issue-1310056500

## Tests Performed:

- tested

## How to Test the Changes:

.q a 962
.go gameobject id 13891
Click and look for the snek

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1676986377096927800.sql
+++ b/data/sql/updates/pending_db_world/rev_1676986377096927800.sql
@@ -1,0 +1,5 @@
+-- Wailing Caverns - Serpentbloom Snake
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 3680;
+DELETE FROM `smart_scripts` WHERE `entryorguid` = 3680;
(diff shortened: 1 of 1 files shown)


Gultask:
In the source video they had very different movement, closer to waypoints instead of random in a radius. This PR would serve for now, but wouldn't really close the issue, could you update it so it does not happen automatically? I use 'Progresses' myself.

avarishd:
> In the source video they had very different movement, closer to waypoints instead of random in a radius. This PR would serve for now, but wouldn't really close the issue, could you update it so it does not happen automatically? I use 'Progresses' myself.

I have the feeling that this WP running was added in 4.1.0's "rework", but I cannot find any videos from before cataclysm.

Maelthyrr:
Tested and work fine. @Gultask do you approve?

Gultask (approved):

Gultask:
Oh yeah, totally. We can go investigate further later. Until then this is certainly an improvement.

Issue #15157, opened 2023-02-22
Flying Carpet is not available from any trainer in TBC


v-mstrs:
### Current Behaviour

Flying Carpet is not available from any trainer in Outlands. Tested at 70 and with Riding 225

### Expected Blizzlike Behaviour

Flying Carpet should be available from all trainers in Outlands (Miralisse, Weaver Aoa, Dalinna, Hama and the bookshelves in Shattrath City) 

### Source

https://www.wowhead.com/wotlk/spell=60969/flying-carpet

### Steps to reproduce the problem

Get 300 tailoring then look at recipes available 

### AC rev. hash/commit

bdc70d8219c0

### Operating system

Linux Mint 21.1

PR #15161, opened 2023-02-22
fix(DB/Trainer) Add Flying Carpet to Tailoring trainers

noisiver:
## Changes Proposed:
-  Add Flying Carpet to the tailoring trainers that were missing it

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15157

## SOURCE:

https://www.wowhead.com/wotlk/spell=60969/flying-carpet#taught-by-npc

## Tests Performed:

- Checked the listed creatures to make sure they have the spell available in their list

## How to Test the Changes:

1. Check the creatures listed in the SQL and make sure they have the spell available to train

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1677089599224449400.sql
+++ b/data/sql/updates/pending_db_world/rev_1677089599224449400.sql
@@ -1,0 +1,10 @@
+--
+DELETE FROM `npc_trainer` WHERE `SpellID`=-201041 AND `ID` IN (18749, 18772, 26964, 33613, 33636, 33684);
+INSERT INTO `npc_trainer` (`ID`, `SpellID`) VALUES
+-- Flying Carpet
+(18749, -201041), -- Dalinna <Master Tailoring Trainer>
+(18772, -201041), -- Hama <Master Tailoring Trainer>
+(26964, -201041), -- Alexandra McQueen <Grand Master Tailoring Trainer>
+(33613, -201041), -- Tailoring
+(33636, -201041), -- Miralisse <Master Tailoring Trainer>
+(33684, -201041); -- Weaver Aoa <Master Tailoring Trainer>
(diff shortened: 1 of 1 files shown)


Gultask:
Correct me if I'm wrong, but wouldn't this also add 67 other spells to those trainers?

image

noisiver:
> Correct me if I'm wrong, but wouldn't this also add 67 other spells to those trainers?
> 
> image

Oh, right. I would have to confirm all those before adding them. Sigh. I don't know if adding the spell straight up is a viable option but I'll look at it later.

Gultask:
It could very well be the references that are wrong, and you'd only need to add that spell to a reference they all share

noisiver:
I'll take a look at it as soon as I can. Thank you for reminding me about this since I didn't even consider it.

noisiver:
Took me long enough but I updated it to only add Flying Carpet to said trainers.

Maelthyrr:
Hi, Thanks for the PR. Look to work well.

Gultask:
Wouldn't it be better to remove it from reference 201041 and add it to reference 201040 instead?

Spell 60969 is currently in reference 201041
image

noisiver:
> 

It depends on if that reference is used by all trainers required and also no other trainers. I did look on wowhead, the wotlk classic database, and it does have the list of trainers on it so I guess I could look into it.

Gultask:
I think everyone's here, no?
ID 26964 uses reference 201040
image

noisiver:
I was confusing myself but there is in fact one trainer that **should** have Flying Carpet but doesn't have that reference. What should I do about that?

image
image

The first image is all trainers that has that reference and the second is all that should have Flying Carpet. [Meline Duskpath](https://www.wowhead.com/wotlk/npc=33596/meline-duskpath) is the one that should have it but doesn't have that reference.

noisiver:
For now I've added it to the reference. Hopefully you can give me some advice on how to handle the last trainer. Perhaps it's missing the reference.

data/sql/updates/pending_db_world/rev_1677089599224449400.sql
```diff
@@ -0,0 +1,4 @@
+--
+DELETE FROM `npc_trainer` WHERE `SpellID`=60969;
```
Gultask, line 2: Usually it's better to add 2 fields when deleting. Just to be a bit safer
```suggestion
DELETE FROM `npc_trainer` WHERE `SpellID`=60969 AND `ID` IN (201040, 201041);
```

Nyeriah (approved):

PR #15201, opened 2023-02-25
fix(Instance/Script): Fix Factory Door in Deadmines being closable

Johaine:
## Changes Proposed:
-  The boss (Rhahk'Zor) activates the door in his room instead of only changing gameobject state
-  The lever behind the door then won't be able to close the door again when players click on it
- Change Instance script

## Issues Addressed:

- Closes [10340](https://github.com/azerothcore/azerothcore-wotlk/issues/10340)

## SOURCE:

In this WoW Classic video you see the door open as soon as the boss is killed:
https://youtu.be/71UPdm30VrI?t=448

Shortly after, the player clicks the door lever and nothing happens:
https://youtu.be/71UPdm30VrI?t=471

This is also the only logic way a group cannot be softlocked in the instance. If the lever would close _and_ open the door, the following could happen:
1) All players go through the door, a player closes the door behind the group.
2) The group progresses through the instance and wipes at some point
3) Everyone has to run in and stands in front of the closed door. Even though the lever would be able to open it again, nobody can reach it, locking any further progress
(This would be even more annoying because you might wipe at the last boss and would fail the instance much later.)
The only way to prevent that is to make the door static and players can't close it to begin with.

## Tests Performed:

- Killed the boss, go through the now open door. Click the lever and watch nothing happen
- Restarted server, entered DM again and confirmed that the correct state was preserved

## How to Test the Changes:

1. Enter Deadmines, progress to first boss and defeat it (Door must open)
2. Click on the lever in the door way (Door must not close)
3. Restart server
4. After server restart and entering the saved DM instance, the door must still be open and the lever must not close the door when clicked

While researching this issue, I found more issues with DM (will create separate issues and PRs later)

- [ ] Fix DM Instance script (currently, after server restart other doors could be closed by players)
- [ ] Cleanup all DM door scripts (most are scripted twice with only one way of activation required)
- [ ] Possibly overhaul the core logic for saving and restoring Gameobjects in instances

### DIFF
--- a/data/sql/updates/pending_db_world/DM_FactoryDoor_Fix.sql
+++ b/data/sql/updates/pending_db_world/DM_FactoryDoor_Fix.sql
@@ -1,1 +1,1 @@
-UPDATE `acore_world`.`smart_scripts` SET `action_type`=9, `comment`='Rhahk\'Zor - On Just Died - Activate Gameobject' WHERE  `entryorguid`=644 AND `source_type`=0 AND `id`=2 AND `link`=3;
+UPDATE `smart_scripts` SET `action_type`=9, `comment`='Rhahk\'Zor - On Just Died - Activate Gameobject' WHERE `entryorguid`=644 AND `source_type`=0 AND `id`=2 AND `link`=3;
--- a/data/sql/updates/pending_db_world/DM_FactoryDoor_Fix.sql
+++ b/data/sql/updates/pending_db_world/DM_FactoryDoor_Fix.sql
@@ -1,0 +1,1 @@
+UPDATE `acore_world`.`smart_scripts` SET `action_type`=9, `comment`='Rhahk\'Zor - On Just Died - Activate Gameobject' WHERE  `entryorguid`=644 AND `source_type`=0 AND `id`=2 AND `link`=3;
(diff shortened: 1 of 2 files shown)


data/sql/updates/pending_db_world/DM_FactoryDoor_Fix.sql
```diff
@@ -0,0 +1 @@
+UPDATE `acore_world`.`smart_scripts` SET `action_type`=9, `comment`='Rhahk\'Zor - On Just Died - Activate Gameobject' WHERE  `entryorguid`=644 AND `source_type`=0 AND `id`=2 AND `link`=3;
```
Gultask, line 1:
```suggestion
UPDATE `smart_scripts` SET `action_type`=9, `comment`='Rhahk\'Zor - On Just Died - Activate Gameobject' WHERE `entryorguid`=644 AND `source_type`=0 AND `id`=2 AND `link`=3;
```
Delete the `acore_world`. part as some people use different names for their databases :)

Johaine: Done, thanks. Will remember that for the future. :)

Gultask: The DB part is fine. The cpp part I'll leave to the others :P

Nyeriah (approved):

Nyeriah (approved):

Issue #13908, opened 2022-11-22
[Nagrand] The Ring of Blood: The Warmaul Champion


Santaria94:
### Current Behaviour

1. Seems to only charge to the current target.
2. Not immune to CC and disarm.

### Expected Blizzlike Behaviour

1. https://www.wowhead.com/tbc/quest=9973/the-ring-of-blood-the-warmaul-champion#comments
>By 6934 (35,600 – 13·116·150) on 2006/12/10 (Patch 2.0.1)		
This boss will Mortal Strike, **randomly charge group members**, and hits extremely hard.

>By moosandra on 2007/03/09 (Patch 2.0.10)
He also charges people at random (little damage) and goes right back to the tank. This also has a stun effect when he charges you.

https://youtu.be/dY-KJEcABGY?t=372
https://youtu.be/Zv3WboMWC5Y?t=417

2. Found only mention of fear and disarm, but most likely should be immune to all types of CC
https://www.wowhead.com/npc=18402/warmaul-champion
>By 861345 (1,522 – 2·14) on 2011/12/25 (Patch 4.3.0)		
...and he's unaffected by psychic scream and psychic horror doesn't disarm him (from what I can tell).

### Source

https://www.wowhead.com/tbc/quest=9973/
https://www.wowhead.com/npc=18402/warmaul-champion
https://youtu.be/dY-KJEcABGY?t=372
https://youtu.be/Zv3WboMWC5Y?t=417

### Steps to reproduce the problem

.go cr 66480
.quest add 9973

### Extra Notes

The rest of the guys from this quest line are also not immune to CC. However, comments about these bosses being impossible or very difficult to kill solo hint that all bosses should be immune to all types of CC.

https://www.wowhead.com/tbc/npc=18398
https://www.wowhead.com/tbc/npc=18399
https://www.wowhead.com/tbc/npc=18400
https://www.wowhead.com/tbc/npc=18401
https://www.wowhead.com/tbc/npc=18402

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/afebbb816cb7fe84fbec0c46beb532b8b5215dbe

### Operating system

Windows 10 x64

Issue #13907, opened 2022-11-22
[Nagrand] The Ring of Blood: Skra'gath


Santaria94:
### Current Behaviour

1. Doesn't gain magic reduction after taking the first magic damage.
2. Does not use Piercing Shadow. Also probably not immune to interruption.
3. Doesn't use Dark Shriek.
4. Doesn't remove threat after using Shadow Burst.

### Expected Blizzlike Behaviour

1. After taking the first magic damage, must reduce damage by 75% to this magic school. Has the right abilities for it.
https://www.wowhead.com/tbc/spell=34331/damage-reduction-arcane
https://www.wowhead.com/tbc/spell=34333/damage-reduction-fire
https://www.wowhead.com/tbc/spell=34334/damage-reduction-frost
https://www.wowhead.com/tbc/spell=34336/damage-reduction-holy
https://www.wowhead.com/tbc/spell=34335/damage-reduction-nature
https://www.wowhead.com/tbc/spell=34338/damage-reduction-shadow

https://www.wowhead.com/tbc/npc=18401/
>By DarkFerret (1,654 – 1·23) on 2007/04/26 (Patch 2.0.12)		
Resistant to the first school of magic used against him.

The Death Knight deals Frost Damage and Skra'gath gets buff.
https://youtu.be/ivPcGMH-mbg?t=20

2. 0:27 - uses "Piercing Shadow" with interrupt immunity
https://youtu.be/ivPcGMH-mbg?t=25

3. I learned this ability in the game. The visual effect is like a thunderclap. In the video you can see:
https://youtu.be/ivPcGMH-mbg?t=33 - 0:35
https://youtu.be/ivPcGMH-mbg?t=49 - 0:50
https://youtu.be/ivPcGMH-mbg?t=60 - 1:01

4. This is stated in the ability tooltip itself, also a comment from wowhead:

https://www.wowhead.com/quest=9972/the-ring-of-blood-skragath#comments
>By 117913 (2,830 – 1·5·29) on 2009/03/24 (Patch 3.0.9)
[...]

Issue #13904, opened 2022-11-21
[Nagrand] The Ring of Blood: The Blue Brothers


Santaria94:
### Current Behaviour

1. You can only pull one boss, and the second will not join the fight.
2. If you kill one boss and let the second disappear, then when you restart the quest, it will be enough to kill only one boss to complete.
3. Bosses are not immune to CC and Poison.
4. Bosses have "Eviscerate". Similar to rogues, this should be their strongest ability, but currently it's their weakest ability, dealing a ridiculous 200-300 damage to clothes.

### Expected Blizzlike Behaviour

1. Both bosses must be in the fight. https://youtu.be/Of47zUauuKY?t=32
2. Both bosses must be defeated on the same try.
3. Bosses should be immune to at least sheep, but most likely to all CC. They also seem to be immune to poisons. 
Pay attention to the video. The rogue never poisoned them in the whole fight: https://youtu.be/Of47zUauuKY?t=32
https://www.wowhead.com/quest=9967
>By 9387 (2,005 – 11·19) on 2006/12/20 (Patch 2.0.1)		
These mobs are immune to sheep, so don't even think about it.
4. I haven't found exact damage values for "Eviscerate", but I'm sure it should be much stronger.

### Source

https://www.wowhead.com/quest=9967
https://youtu.be/Of47zUauuKY?t=32

### Steps to reproduce the problem

.go cr 66480
.quest add 9967

### Extra Notes

The rest of the guys from this quest line are also not immune to CC. However, comments about these bosses being impossible or very difficult to kill solo hint that all bosses should be immune to all types of CC.

https://www.wowhead.com/tbc/npc=18398
https://www.wowhead.com/tbc/npc=18399
https://www.wowhead.com/tbc/npc=18400
https://www.wowhead.com/tbc/npc=18401
https://www.wowhead.com/tbc/npc=18402

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/afebbb816cb7fe84fbec0c46beb532b8b5215dbe

### Operating system

Windows 10 x64

Issue #13903, opened 2022-11-21
[Nagrand] The Ring of Blood: Brokentoe


Santaria94:
### Current Behaviour

1. Not immune to CC abilities.
2. Emote "Brokentoe prepares to charge" is missing.
3. If the player is at max melee range, the ability "Hoof Stomp" does not hit them. I'm guessing this is due to the boss's frame being too big.

### Expected Blizzlike Behaviour

1. Guess I should be immune to all types of CC
https://www.wowhead.com/npc=18398
Rogue's comment
>By 24919 (616 – 1·6) on 2007/06/17 (Patch 2.1.1)		
Quite a tough NPC, he has about 60k hp, hits for 1500 on leather, **immune to stuns, disorients and poisons** and he does a war stomp which stuns the person he's fighting.
This warlock doesn't directly state that Brokentoe is not immune to fear, but the strategy he suggested hints at it.
>8inchunbuffed on 2009/05/12 (Patch 3.1.1)
I own a demonology warlock...practically can solo ANYTHING haha health funnel and drain health, plus some DoT and your fine
Also don't see comments from solo druids using roots

2. Emote "Brokentoe prepares to charge" - 0:22
https://youtu.be/nGskGcryyUM?t=21

3. Also take a look at the boss frame in the video. His hooves are on the edge of the frame.

### Source

https://www.wowhead.com/npc=18398
https://www.wowhead.com/tbc/quest=9962
https://youtu.be/nGskGcryyUM?t=21

### Steps to reproduce the problem

.go cr 66480

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/afebbb816cb7fe84fbec0c46beb532b8b5215dbe

### Operating system

Windows 10 x64

Gultask:
I assume the rogue is talking about CC poisons as otherwise I believe that'd mean nature immunity. Looking at videos from shamans this isn't the case.

The model info is all correct. Both CombatReach and BoundingRadius so I don't have any other ideas.

PR #15205, opened 2023-02-26
fix(DB/SAI): Rewrite Ring of Blood

Gultask:
## Changes Proposed:
- Use sniffed data as much as possible

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15190
- Closes https://github.com/chromiecraft/chromiecraft/issues/5095
- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/13903
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13904
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13907
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13908
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15176
- Closes https://github.com/chromiecraft/chromiecraft/issues/4952

## Tests Performed:

- Tested Ingame
- 

## How to Test the Changes:

1. `.go c id 18471`
2.
3.

- [x] Mogor
- [x] Immunity flags and combat tweaks

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1677368120858755900.sql
+++ b/data/sql/updates/pending_db_world/rev_1677368120858755900.sql
@@ -251,16 +251,19 @@ DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 18069);
 INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
 (18069, 0, 0, 0, 9, 0, 100, 0, 0, 30, 6000, 8000, 0, 11, 16033, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 'Mogor - Within 0-30 Range - Cast \'Chain Lightning\''),
 (18069, 0, 1, 0, 9, 0, 100, 0, 0, 20, 8000, 18000, 0, 11, 39529, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 'Mogor - Within 0-20 Range - Cast \'Flame Shock\''),
-(18069, 0, 2, 0, 2, 1, 100, 1, 0, 60, 0, 0, 0, 11, 15982, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Mogor - Between 0-60% Health - Cast \'Healing Wave\' (Phase 1) (No Repeat)'),
-(18069, 0, 3, 0, 2, 2, 100, 0, 0, 50, 15000, 15000, 0, 11, 15982, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Mogor - Between 0-50% Health - Cast \'Healing Wave\' (Phase 2)'),
-(18069, 0, 4, 0, 0, 0, 100, 0, 3400, 9400, 13200, 26400, 0, 11, 18975, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Mogor - In Combat - Cast \'Summon Ice Totem\''),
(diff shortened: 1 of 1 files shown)


Gultask:
Done. Ready for review. Also requesting a test to see if it does not break for someone else.

avarishd:
Going by listed issues.
General
- [x] Can't start multiple encounters.
- [x] Boss(es) despawn correctly after not being engaged within 30s. Followed with a mocking line from Gurg. (With a quest fail)
- [x] Gurg yells when %n beats the encounter.
- [x] Correct group/player gets rewarded, even if someone else kills the current boss. (even if tapped)
- [x] Mogor talks according to the current quest progress.
- [ ] unknown interaction: All bosses except twins aggro on spawn to the invoker, no matter how far away you are.
- [ ] Sometimes Gurg quest status isn't being updated until a relog (very, very rare, idk if related to PR or not)

Brokentoe
- [x] has emote
- [x] Immune to CC (tested HoJ && Fear)
- [ ] Hoof Stomp ability can still be avoided by being at max melee range.
(But this can only be done if he is snared, which you no longer can do)

Twins
- [x] Both pull at the same time
- [x] If you kill 1, the other doesn't despawn. (in combat)
- [ ] unknown interaction: You can kill 1, run away and reset and then kill the other one before it despawns. (You get credit)
- [x] You can kill 1, run away, let the other one despawn. -> You get quest failed (when the other one despawns)
- [ ] Immune to poisons
- [x] Casts Sinnister Strike && Eviscerate

Rokdar
no issues listed, worked fine

Skra'gath
- [x] Gains damage reduction against the first element (tested fire/shadow) that hits it.
- [x] Uses Piercing Shadow and is immune to interupts.
- [x] unknown interaction: Uses Dark Shriek, but no matter if the victim is casting or not. (It's a kick that deals damage)
- [x] All threat is removed after Shadow Burst.

Warmaul Champion
- [x] Immune to disarms
- [ ] Charge does happen, but it's very, very rare and it looks like minimum distance is like 15-20y, instead of just few (from https://www.youtube.com/watch?v=dY-KJEcABGY&t=372s)

Mogor
- [x] Starting his quest, roleplaying from him and Gurg works
- [x] Walks into the middle, following his WP and is Immune (friendly) until he reaches his destination.
- [x] Ice totem is very frequent.
- [x] Phase 2 always happens (repeated 5 times)
- [x] Instantly despawns when evaded, instantly respawns after being defeated.

Gultask:
> Sometimes Gurg quest status (personal, other characters see it fine) isn't being updated until a relog (very, very rare, idk if related to PR or not)

If it comes back on a relog, I really have no idea :P 
After Mogor there's a 5s delay before the questgiver flag comes back, otherwise it's all in the Actionlist, if quest credit is given then the flag should be re-added, always

> unknown interaction: Edge case, not always happening -> Bosses (except brokentoe) aggro on spawn to the invoker, no matter how far he is.

This "not always happening" is what worries me, because it's intended behaviour :D 
It happens with .gm on, maybe something else as well? Target is stored by the questgiver and then sent to the bosses, so maybe there should be a bigger delay between the target being sent and the aggro
Aggro is 2.4s after spawn, target should actually be sent immediately on spawn, but it's not possible in SAI, afaik. I know this because they have targets in their talk packets.

> Hoof Stomp ability can still be avoided by being at max melee range.

I think we may have had a similar problem with the dog in MC, unknown to me. That's why it won't close the issue.

> unknown interaction: You can kill 1, run away and reset and then kill the other one before it despawns. (You get credit)

I didn't think about testing this one because it's fairly hard to, but my gut says that this behaviour wouldn't be too far off from blizz standards. The amount of times I've soft-locked instances by testing them too far is kinda crazy. Battle of Mount Hyjal can hard lock if you go too far and hit triggers too early.
The only solution I can think of would be despawn on evade, but that wouldn't be blizzlike, I've already made that compromise with Mogor 😩

> Immune to poisons

I'll have to check videos again but I think it means immunity to CC poisons, otherwise I think the immunity would be to nature spells in general?

>  Charge does happen, but it's very, very rare

I used Cmangos data for this one, with 8-25y min range. The repeat is around 20s so maybe that needs adjusting. I'll see the videos.

Gultask:
Oh yeah and big thanks for such a detailed test! 😄

avarishd:
Regarding to the "immune to poisons", I think https://github.com/azerothcore/azerothcore-wotlk/issues/13904 is wrong, and in the source the rogue has blade twisting talent, it only shows immune when he sinister strikes (and kidney, kek). Also the shaman in the video is blasting so yeah, definitely not immune to poisons/nature, just cc.

acidmanifesto (approved):

Gultask:
> unknown interaction: You can kill 1, run away and reset and then kill the other one before it despawns. (You get credit)

You actually can in Retail!

> Charge does happen, but it's very, very rare

Videos are kinda inconclusive. Mostly 1-2 charges before he dies, sometimes not happening at all after pull: https://youtu.be/JoBR2BbSTPE?t=606
I'll keep Cmangos for now.

PR #15213, opened 2023-02-26
fix(Core/Hook): Call OnUnitDeath for all unit types.

AnchyDev:
## Changes Proposed:
Allow all unit deaths to call the OnUnitDeath(victim, killer) hook.

## Issues Addressed:
-

## Tests Performed:
- Tested OnUnitDeath for player death and it works.

## How to Test the Changes:

1. Test any module with this hook.

- You should have nullptr validation inside of this hook for your modules because there is none prior.
- Some modules using this hook may need to be updated if they assume it is still only relevant for creatures.


Nyeriah (approved):

Nyeriah:
Scripts running this hook might need aditional checks to prevent crashes

PR #15222, opened 2023-02-27
fix(scripts/boss_reliquary_of_souls): Combat status

Si1ker:
Tested - doesn't break other stuff

* if you want only the ultimate/godlike fix, close this and work on it

## Issues Addressed:

- Closes ##15217


acidmanifesto (approved):
Simple and i see no other attempts to fix this exceedingly specific issue. Kept within a single void of a boss script. I do not see any possible way for this to break anything in the core.

avarishd:
- [x] Tested, boss instantly despawns when there are no alive players. Which allows you to enter the instance after a wipe.

- [x] Bonus: Resets even if there is a DI-ed player.

PR #15255, opened 2023-03-03
fix(Core/Spells): Draenei Call of Air 4/4 Susurrus buff duration.

avarishd:
Not exactly a "blizzlike", but from watching the source this seems better solution than the painful 1minute and 30seconds of flying...
If deemed that is better to reduce the speed, edit the "Closes" to "Progress", or let me know.
## Changes Proposed:
- Reduce buff duration from 1,5m to 28s, which is the exact flying time with current speed.
## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13158
- Closes https://github.com/chromiecraft/chromiecraft/issues/4113

## SOURCE:

in the issue
## Tests Performed:

- tested

## How to Test the Changes:

.q add 9554
.go quest starter 9554


Maelthyrr:
Hi, tested and look fine. The quest work always, and the timing also.

Maelthyrr (approved):

PR #15291, opened 2023-03-06
fix(db/creature_template): Remove creature 138

neifion-00000000:
## Changes Proposed:
-  Removes creature 138
-  Just look at this thing, what's it doing?  be gone.
 https://wowgaming.altervista.org/aowow/?npc=138

## SOURCE:

This doesn't exist anywhere, nor should it.  If you can prove me wrong I'd be very interested to see it.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1678062827982837200.sql
+++ b/data/sql/updates/pending_db_world/rev_1678062827982837200.sql
@@ -1,0 +1,3 @@
+--
+-- This entry should not exist and appears to serve no purpose
+DELETE FROM `creature_template` WHERE `entry`=138;


Gultask (approved):

Nyeriah:
If it's a custom/private server leftover it's okay to delete it but if it's a sniffed remain/QA creature it shouldn't be deleted

neifion-00000000:
to be clear I'm not advocating for removing any mobs that show up in any legit databases  or can be pinged from the blizzard servers, just ones that aren't and don't (like this case)

I have passed up dozens of those oddities before finding that guy

Gultask:
```
ServerToClient: SMSG_QUERY_CREATURE_RESPONSE (0x2914) Length: 5 ConnIdx: 1 Time: 08/18/2022 09:48:02.477 Number: 1953
Entry: 138
no response
```

heyitsbench:
> no response

There's likely many things that would return no response that still belong in the DB.

neifion-00000000:
> > no response
> 
> There's likely many things that would return no response that still belong in the DB.

this was from the brute force packet, which is how the creature_template was populated

while it doesn't 100% confirm it wasn't in wotlk originally from a logical standpoint (it not being in classic doesn't mean it wasn't in wotlk), but it does from a contextual standpoint because all of the other misc entries do exist (only this mob and other confirmed custom mobs are missing, so the chances of it not being the same are slim).

Issue #6490, opened 2021-06-21
Feature: use std::chrono


Kitzunu:
![image](https://user-images.githubusercontent.com/24550914/137370596-d28934bc-8833-40eb-b314-960bb0c0d9d3.png)

We should start replacing events where we use time like "x * IN_MILLISECONDS" to std::chrono instead

Example: https://github.com/azerothcore/azerothcore-wotlk/pull/6489

Winfidonarleyan:
yes

Eronox:
> > so I dont have to swap them when I see them in code?
> 
> I don't know what you mean?

The origin intent is to change the current way of defining times to use the chrono system instead, like so: 
https://github.com/azerothcore/azerothcore-wotlk/pull/8498/files

Your PR seems to be making this unneccessary, that instead todays way will be just changed to chrono, instead of manually changing each code part one stumbles upon to chrono.

At least this is what I think he meant 😄

Winfidonarleyan:
About this, i think if we convert everything to std::chrono, the arguments that some functions require will become clearer and there will be fewer mistakes from developers

Nyeriah:
Yeah, it's much more readable and way more pratical to use. Looking forward to this.

Kitzunu:
I think this issue serves no purpose anymore.

It is well known that std::chrono is preferred at this point

PR #15304, opened 2023-03-07
refactor(Core): Some new conversions to std::chrono

Maelthyrr:
## Changes Proposed:
-  Convert some methods to std::chrono (Addwaypoint, SendCriteriaUpdate, SetAttackTimer, SetDuration, and SetMaxDuration)

## Issues Addressed:

- Part of https://github.com/azerothcore/azerothcore-wotlk/issues/6490

## Tests Performed:

- Not now
- 

## How to Test the Changes:

1.
2.
3.

- [ ] Fix mod-progress
- [ ]


src/server/scripts/Events/brewfest.cpp
```diff
@@ -2038,7 +2038,7 @@ class spell_direbrew_disarm : public AuraScript
         if (Aura* aura = GetTarget()->GetAura(SPELL_DIREBREW_DISARM_GROW))
         {
             aura->SetStackAmount(aura->GetStackAmount() + 1);
-            aura->SetDuration(aura->GetDuration() - 1500);
+            aura->SetDuration(Milliseconds(aura->GetDuration()) - 15s);
```
balleny, line 2041: @Maelthyrr this doesnt seem right - could case a crash if negative

Maelthyrr: You're right. But does some aura durations can be negative ?

balleny: from an ingame perspective: no
code-wise: doubt.

maybe, it makes sense in the broader scope, to add some logic checks to functions (i think we have had crash issues with a random scheduler that had to select between a lower value and an higher value, but accidentally, the higher value was smaller than the lower value) and overloaded functions that prevent negative duration

PR #15309, opened 2023-03-08
fix(DB/Quest): Improvements for request texts

ohkasdasf:
## Changes Proposed:
-  This fixes a few wrong entries in the quest_request_items.sql table.

## Issues Addressed:

- Closes an unreported issue.

## Tests Performed:

- No tests performed.

## How to Test the Changes:

1. Accept respective quest.
2. Talk to the quest giver NPC for the request items text.

### DIFF
--- a/data/sql/updates/pending_db_world/qri01.sql
+++ b/data/sql/updates/pending_db_world/qri01.sql
@@ -1,0 +1,10 @@
+DELETE FROM `quest_request_items` WHERE `ID` IN (124, 191, 192, 257, 258, 289, 457, 511);
+INSERT INTO `quest_request_items` (`ID`, `EmoteOnComplete`, `EmoteOnIncomplete`, `CompletionText`, `VerifiedBuild`) VALUES
+(124, 1, 0, 'My horses still seem skittish, and I still hear that cursed wolf cry at night. Please, $N, do something about those Gnolls!', 12340),
+(191, 1, 0, 'Go kill 10 panthers! Show us what you\'ve got!', 12340),
+(192, 1, 0, 'What are you doing here, $n! You should be crawling through the brush, trying to kill 10 Shadowmaw Panthers. I thought you fancied yourself a big game $c?', 12340),
+(257, 1, 1, 'No luck? Don\'t feel too bad, $ner...$B$BNot everyone can be me.', 12340),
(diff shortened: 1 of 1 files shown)


heyitsbench:
The VerifiedBuild column is meant to be the game build the data was collected from. If you're getting these values from sniffs in Classic/Retail, you could fill the value out from there (i.e. 48340 for current WotLK Classic), but if it's not coming straight from a sniff or in-game but instead a source like Wowhead, you'd enter a 0 (or whatever AC staff decide a non-verified entry) in the field.

Maelthyrr:
Yes, can you link the sources of these entries?

ohkasdasf:
> The VerifiedBuild column is meant to be the game build the data was collected from. If you're getting these values from sniffs in Classic/Retail, you could fill the value out from there (i.e. 48340 for current WotLK Classic), but if it's not coming straight from a sniff or in-game but instead a source like Wowhead, you'd enter a 0 (or whatever AC staff decide a non-verified entry) in the field.

That I knew for other locale files, however I was surprised since the english main file had always a build number of 12340 which I perceived to be the equivalent of 0 so i copied it. I will then change all those values to 0.

> Yes, can you link the sources of these entries?

Sorry for not linking all the sources, this happened a bit as a by-product regarding locale fixing. I checked these entries manually by double checking sources such as wowhead/classic playthroughs and old databases. If you check those edited entries with the original ones you can notice that the current text is either a text from another quest or not related to the matter of the quest. Some have minor errors which could also be neglected.

https://www.wowhead.com/wotlk/quest=124/a-baying-of-gnolls
https://www.wowhead.com/wotlk/quest=191/panther-mastery
https://www.wowhead.com/wotlk/quest=192/panther-mastery
https://www.wowhead.com/wotlk/quest=257/a-hunters-boast (the NPC of this quest chain addresses the player by the wrong name reqularly which is accomplished as follows: $ner, $nama, $nath, $nah; original values have here fully written out names)
https://www.wowhead.com/wotlk/quest=258/a-hunters-challenge (see above)
https://www.wowhead.com/wotlk/quest=289/the-cursed-crew (missing gender equivalent)
https://www.wowhead.com/wotlk/quest=457/the-balance-of-nature (number differs from objective text)
https://www.wowhead.com/wotlk/quest=511/encrypted-letter (gender should be addressed instead of name)

PR #15637, opened 2023-03-28
fix(Scripts/Molten Core): Correct Baron Geddon Inferno damage values

ZhengPeiRu21:
## Changes Proposed:
-  Fix Bardon Geddon Inferno Damage
-  Source: Vmangos (https://github.com/vmangos/core/pull/1853)

## Issues Addressed:

- Closes unreported issue

## SOURCE:

Vmangos (https://github.com/vmangos/core/pull/1853)
See linked page for additional sources and info

## Tests Performed:

- Compiles, not yet tested in game

## How to Test the Changes:

1. Engage Baron Geddon, confirm damage is correct for each tick

### DIFF
--- a/src/server/scripts/EasternKingdoms/BlackrockMountain/MoltenCore/boss_baron_geddon.cpp
+++ b/src/server/scripts/EasternKingdoms/BlackrockMountain/MoltenCore/boss_baron_geddon.cpp
@@ -174,14 +174,13 @@ public:
                         break;
                     case 4:
                     case 5:
-                        multiplier = 3;
-                        break;
-                    case 6:
-                    case 7:
                         multiplier = 4;
                         break;
-                    case 8:
-                        multiplier = 5;
+                    case 6:
+                        multiplier = 6;
+                        break;
+                    case 7:
+                        multiplier = 10;
                         break;
                 }
 
--- a/src/server/scripts/EasternKingdoms/BlackrockMountain/MoltenCore/boss_baron_geddon.cpp
+++ b/src/server/scripts/EasternKingdoms/BlackrockMountain/MoltenCore/boss_baron_geddon.cpp
@@ -168,18 +168,18 @@ public:
                 int32 multiplier = 1;
                 switch (aurEff->GetTickNumber())
                 {
-                    case 2:
                     case 3:
+                    case 4:
                         multiplier = 2;
                         break;
-                    case 4:
                     case 5:
+                    case 6:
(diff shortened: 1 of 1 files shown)


elthehablo:
In my experience of local testing, The 500 tick only hits once. According to the original PR on vmangos it should be 500, 500, 1000, 1000, 2000, 2000, 3000, 5000 (8 total). My tests are as follows on AC:

**notes before changes:**

500 first tick
1000 second tick
1000 third tick
1500 fourth tick
1500 fifth tick
2000 sixth tick
2000 seventh tick
2500 eighth tick

**with your PR it is:**

500 first tick
1000 second tick
1000 third tick
2000 fourth tick
2000 fifth tick
3000 sixth tick
5000 seventh tick
500 eighth tick

so even though the values seem correct now. the first tick seems to go as the last tick... at least for me. so this needs some more work and commits to get working correctly. at least if my tests are correct. my suspicions are in the way you have handled your switch statement here. tick probably starts at 1 and should go to 8 as there seem to be 8 ticks. where you did case 7 which has multiplier set to 10, I think it should still be case 8. I think you making it case 7 so at the 7th tick made it so that for the last tick it reverts back to multiplier = 1.

src/server/scripts/EasternKingdoms/BlackrockMountain/MoltenCore/boss_baron_geddon.cpp
```diff
@@ -174,14 +174,13 @@ class spell_geddon_inferno : public SpellScriptLoader
...
                     case 5:
-                        multiplier = 3;
+                        multiplier = 4;
                         break;
                     case 6:
-                    case 7:
-                        multiplier = 4;
+                        multiplier = 6;
                         break;
-                    case 8:
```
elthehablo, line 183: make this one still case 8 instead case 7 what you did. because tick 8 has to do 5000 damage

elthehablo (commented):
ideas to improve on new script are as follows

ZhengPeiRu21:
I have just pushed a commit that should count the ticks differently. Would you please mind retesting when you have time? Thanks!

elthehablo:
Ticks are now correctly 500, 500, 1000, 1000, 2000, 2000, 3000, 5000 as intended :)

Should be good to merge now!

elthehablo (approved):
works as intended now!

Nyeriah (approved):

Issue #8530, opened 2021-10-15
Various NPCs Incorrectly Dropping Blood of the Mountain


Edgardiez7:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [x] 50-59

<html>
<body>
<h3 style="box-sizing: border-box; margin-top: 24px; margin-bottom: 16px; font-size: 1.25em; font-weight: 600; line-height: 1.25; color: rgb(201, 209, 217); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Current Behaviour</h3><p style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(201, 209, 217); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="https://wowgaming.altervista.org/aowow/?item=11382" rel="nofollow" style="box-sizing: border-box; background-color: transparent; color: var(--color-accent-fg); text-decoration: none;">Blood of the Mountain</a></p>

name | entry | chance
-- | -- | --
Firegut Brute | 7035 | 0,02
War Reaver | 7039 | 0,02
[...]

PR #15768, opened 2023-04-01
fix(DB/Loot): Blood of the Mountain dropping from wrong mobs.

avarishd:
## Changes Proposed:
- Remove Blood of the Mountain from all creatures, except Molten Destroyer and change to 13%

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/2090
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8530

## SOURCE:

https://www.wowhead.com/wotlk/item=11382/blood-of-the-mountain#dropped-by
## Tests Performed:

- sql

## How to Test the Changes:

```sql
SELECT * FROM creature_loot_template WHERE item=11382
```


Gultask:
We don't use Wowhead percentages for a myriad of reasons. Only if you know *exactly* what you're doing.
I'd keep the original value.

avarishd:
It is for sure over 10%, played vanilla recycled myself. Hmmm... 
Original value is 7.1259

Gultask (approved):

Issue #7819, opened 2021-09-11
[Bloodmyst Isle] Razormaw Event is custom


IntelligentQuantum:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [X] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour

Original post: https://github.com/chromiecraft/chromiecraft/issues/1692

[Razormaw](https://wowgaming.altervista.org/aowow/?npc=17592), objective of the [quest of the same name](https://wowgaming.altervista.org/aowow/?quest=9689) hits for 150-200 damage quite quickly on appropriately leveled mail wearers, as well as occasionally casting a low level swipe.

![WoWScrnShot_090921_074501](https://user-images.githubusercontent.com/10586372/132862486-916265d7-afce-401b-b065-90989e67ded1.jpg)

### Expected Blizzlike Behaviour

[TBC Wowhead comments](https://tbc.wowhead.com/npc=17592/razormaw#comments) say that the melee swings should deal about 50 damage on an appropriately leveled Shaman. Many other comments mention the dragon being easily soloable by most classes, whereas my Paladin with a dual boxed healer couldn't survive with heals being spammed.

Swipe should also not be one of Razormaw's abilities (as per the Wowhead link above).

It's also worth noting that when I summoned the dragon, it appeared almost instantly and aggroed, where as on retail she is supposed to circle around before landing. I noticed on the local [DB](https://wowgaming.altervista.org/aowow/?npc=17592) that there is pathing scripted, but it seems it's not working right in game.

### Source

[This video](https://youtu.be/lo7EU2qcFYY?t=125) from Classicl TBC shows the fight with an appropriately leveled Hunter's pet taking 30 damage or so per hit.

### Steps to reproduce the problem

1. Accept [Razormaw](https://wowgaming.altervista.org/aowow/?quest=9689)
2. Summon [Razormaw](https://wowgaming.altervista.org/aowow/?npc=17592) at the Ever Burning Pyre in Bloodmyst Isle

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/595bb6adccbabc714469f3935541978283b8bdfb

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[...]

PR #15785, opened 2023-04-02
fix(DB/Creature): fix damage modifier of Razormaw

elthehablo:
from 7.5 to 2.5. probably still a bit high. subject to change

## Changes Proposed:
-  damagemodifier from 7.5 to 2.5 (might be 2 or 1.5) but that would make it really easy

## Issues Addressed:

- Progress https://github.com/chromiecraft/chromiecraft/issues/5300
- Progress https://github.com/azerothcore/azerothcore-wotlk/issues/7819

## SOURCE:

https://youtu.be/9010SfVJPN0 but this shows damage to a pet. so ideally you should test this as a hunter with a pet at the same level. a cat pet, that is

## Tests Performed:

- tested myself as a paladin with appropriate gear at level 19. I take around 60ish damage each melee hit now, which is doable if it's a 2 man quest

## How to Test the Changes:

1. `.quest add 9689`
2. .tele bloodmyst
3. move to quest location
4. summon boss
5. see how much damage you take on melee, preferably as a hunter with a cat pet

- [ the boss flies over the ground. it should walk, as seen in the video linked above ]

### DIFF
--- a/data/sql/updates/pending_db_world/razormawdamageupdate.sql
+++ b/data/sql/updates/pending_db_world/razormawdamageupdate.sql
@@ -1,0 +1,3 @@
+-- damage modifier edit of Razormaw from 7.5 to 2.5
+
+UPDATE `creature_template` SET `DamageModifier` = 2.5 WHERE `entry` = 17592 AND `modelid1` = 17256;
--- a/data/sql/updates/pending_db_world/razormawdamageupdate.sql
+++ b/data/sql/updates/pending_db_world/razormawdamageupdate.sql
@@ -1,3 +1,3 @@
--- damage modifier edit of Razormaw from 7.5 to 2.5
+-- damage modifier edit of Razormaw from 7.5 to 1.36 based on mangos values
 
-UPDATE `creature_template` SET `DamageModifier` = 2.5 WHERE `entry` = 17592 AND `modelid1` = 17256;
+UPDATE `creature_template` SET `DamageModifier` = 1.36 WHERE `entry` = 17592 AND `modelid1` = 17256;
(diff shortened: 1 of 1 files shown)


Kitzunu:
Can this be sniffed?

Gultask:
Sort of. Min and Max damages are sniffed (40-47 from with Multiplier 1.27 TBCmangos)
You need to calculate with unit_stats to figure out the Modifier

acidmanifesto:
> Can this be sniffed?

It can be but im confused with the OP's subject to change comment.

elthehablo:
my value of 2.5 is an estimate and should probably be fine-tuned is what I meant

data/sql/updates/pending_db_world/razormawdamageupdate.sql
```diff
@@ -0,0 +1,3 @@
+-- damage modifier edit of Razormaw from 7.5 to 1.36 based on mangos values
+
+UPDATE `creature_template` SET `DamageModifier` = 1.36 WHERE `entry` = 17592 AND `modelid1` = 17256;
```
Gultask, line 3: Modelids are sniffed. Until probabilities get implemented we need to use scripting to change models instead, but this looks like leftover, though :p
```suggestion
UPDATE `creature_template` SET `DamageModifier` = 1.36 WHERE `entry` = 17592;
```

Gultask (changes requested):

Gultask:
I think I know the issue with the hovering. Let me see if we have sniffs for the waypoints

data/sql/updates/pending_db_world/razormawdamageupdate.sql
heyitsbench: The model isn't getting changed here though, just being used in tandem with the entry for the WHERE section.

Gultask: Oh right, I didn't notice it :c 
It's not needed in this case

Gultask (approved):

Issue #15780, opened 2023-04-02
(DB) Ferra boss add in DM:W doesn't have correct loot


elthehablo:
edits/link by elthehablo

original ChromieCraft issue: https://github.com/chromiecraft/chromiecraft/issues/5321

### What client do you play on?
enUS

### Faction
Both

### Content Phase:
50-59

### Current Behaviour

Missing rares

Edit by elthehablo: looks like it requires reference id 24016 and id 6000 (traveler's backpack and rares) to be added. this is solely based on wowhead data, which has proven to not always be correct. direct data from retail may be needed, but obviously that is annoying to test

### Expected Blizzlike Behaviour

More loot

Edit by elthehablo: specifically, looks like it's missing green drops, traveler's backpack, etc. interestingly enough, wowhead doesn't show rare drops, but I think that if it's possible to drop traveler's backpack, it can also drop vanilla rares. for AC this will be the fact of the matter, because traveler's backpack is together with rare drops in the reference loot table.

### Source

https://www.wowhead.com/wotlk/npc=14308/ferra#drops;mode:normal

Edit by elthehablo: link above does show that there is probably something missing, but more proof would be nice

### Steps to reproduce the problem

1. Check Keira

Edit by Elthehablo: or any other database editing tool ofc :p can also be tested in game, but yes the db only shows 2 drops rn

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/d030260dd4b7e0aae88f609222877bd2aabdafe0

### Operating system

Linux Mint 21.1 x84_64

### Custom changes or Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[...]

PR #15786, opened 2023-04-02
fix(DB/Creature): Ferra Boss now has its correct loot

Dr-Arayashiki:
## Changes Proposed:
-  add missing loot to npc 14308 (Ferra)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15780

## SOURCE:

https://www.wowhead.com/wotlk/npc=14308/ferra
## Tests Performed:

- tested in game

## How to Test the Changes:

1. open the database
2. I added the missing loot to the boss
3. add npc 14308 and see

- [ ]
- [ ]


heyitsbench:
Couple things: 1) you still have the Grunda mount fix SQL in your branch. 2) You should be attaching the issues to close in your issue.

Dr-Arayashiki:
> Couple things: 1) you still have the Grunda mount fix SQL in your branch. 2) You should be attaching the issues to close in your issue.

I'm still being beaten to move here in git rsrs

Dr-Arayashiki:
> Couple things: 1) you still have the Grunda mount fix SQL in your branch. 2) You should be attaching the issues to close in your issue.

now it worked ok

neifion-00000000:
from what I can gather (these tables are not easy to work off) gear is ~5.25%, bags are ~0.125%, gems are ~1.5% with bias toward aquamarines, the lockbox is ~1%, and the leather is a bad report/skinning loot

The gear should come off a reference table, though which will need to be figured out (one with similar loot as shown tho).   There are two open PRs for this same mob as well, but neither of them are in good enough shape to ship at the moment, I'll be copy/pasting this into both tho.

Dr-Arayashiki:
> from what I can gather (these tables are not easy to work off) gear is ~5.25%, bags are ~0.125%, gems are ~1.5% with bias toward aquamarines, the lockbox is ~1%, and the leather is a bad report/skinning loot
> 
> The gear should come off a reference table, though which will need to be figured out (one with similar loot as shown tho). There are two open PRs for this same mob as well, but neither of them are in good enough shape to ship at the moment, I'll be copy/pasting this into both tho.

yes, but before that only two items were falling, I wanted to at least supply a little of what was missing, I thought it would at least solve part of the problem ^^

neifion-00000000:
yeah but there is a correct way to do this and this isn't it.  Among other things this will now likely drop multiple greens at a time (instead of a green occasionally) and drops leather.  The green loot needs to come off a reference table, the leather needs removed, and the rates need adjusted downwards a lot.

Dr-Arayashiki:
> yeah but there is a correct way to do this and this isn't it. Among other things this will now likely drop multiple greens at a time (instead of a green occasionally) and drops leather. The green loot needs to come off a reference table, the leather needs removed entirely as it's skinning data, and the rates need adjusted downwards a lot (the rates I posted would be acceptable/close).

I see, I'll try to work on the loot reference and put the drop rates shown in wowhead :)

Gultask:
@Dr-Arayashiki 
I'd suggest that you first need to study a bit about references and the loot system AC has: https://www.azerothcore.org/wiki/loot_template

There are many issues with this approach, not to mention that loot issues are notoriously hard to deal with.

Gultask:
Using the method you've shown here would lead to massive bloating of the loot tables. That's why we use references.

For example:
We have 100 NPCs dropping the same 100 items. That's 10000 rows. But if we use references instead it'll be only 201, since the reference will hold the 100 items, and each NPC will only require 1 row to link it to the reference.

Another thing are the chances.

In your current PR, one kill could result in several green drops. In fact, the chance of **not** dropping any loot is close to 1-4% and, consequently, the chance of seeing at least 1 green item is >90%, which is... improbable.

Gultask:
I'd recommend looking at loot tables that already exist. Have you used https://github.com/azerothcore/Keira3 ? It's pretty easy to navigate. And I highly recommend to a beginner.

PR #15808, opened 2023-04-03
fix(DB/SAI): Use SAI for Spitelash Naga in Azshara

LeFroid:
Taken from TrinityCore

These NPCs were previously using ScriptName 'npc_spitelashes' which did not seem to have any scripting for combat. After switching to SAI the behavior is as expected.

## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Gultask (changes requested):
3 Things:

1. Always add column names in INSERTS

2. Follow the SQL standards: 1 line per row inserted. Example:
```
INSERT INTO `table_1` VALUES
(1000, ...),
(2000, ...),
(3000, ...);
```

3. There's a Spitelash Script in zone_azshara.cpp that you'll need to delete if it won't be used anymore

LeFroid:
Thanks for reviewing, I believe I just addressed your points.

Nyeriah (approved):

PR #14495, opened 2023-01-04
fix(DB/creature_template): sniff walk/run values

neifion-00000000:
## Changes Proposed:
- Updates over 16k creatures, about 1/3 (Affected rows: 6,143) needing fixed values for walk and/or run speeds: 

## SOURCE:

Made possible by sniff tools including:
ymir, WPP, malcrom's tool
and sniffs provided to the Sniff-Collective
almost 1000 sniffs were scraped for this PR

for reference purposes if anyone continues this other than me, we are currently up to date on these values using sniffs collected up to 2023-1-03

## Tests Performed:

- Largely untested other than a quick glance, if issues are found they should not be related to values here as barring some anomaly with the parser the values here should be 100% correct

## Known Issues and TODO List:

- I suspect my SQL is not up to standards, I am open to suggestions.
- Yes I'm aware 10,000 lines are extra, such is the price to getting 6,143 good lines so fast  

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #15813, opened 2023-04-04
fix(DB/Creature): Use sniffed speeds for Heroic versions of TBC creatures

Gultask:
## Changes Proposed:
-  Speed was heavily polluted in heroics
-  Most creatures don't change speeds between difficulties
- Most sniffs are from heroics, so if they do change speeds it won't negatively impact heroics

## Issues Addressed:

- Closes 

## SOURCE:

Neifion's Classic sniffs

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


heyitsbench:
Shouldn't this have already been handled with #14495? Is it just that more data is available for these creatures?

Gultask:
If you go to the heroic maps ingame all entries that are sniffed will be of the normal versions. So the heroic versions never got fixed in the first place.

In all the instances where we compared speeds NormalxHeroic they were always the same. I think there might be some exceptions (Raging Flames from Netherbinder Sepethrea, for example)

elthehablo:
How about the voidwalkers from Vorpil? Or were those correct? I'm too lazy to check the ids in here

Gultask:
Without this PR the Heroic versions travel at 1.14 speed, the same as the majority of creature run speeds, which is slightly faster than the player

elthehablo:
Really? They moved at blazing speeds when I was testing Vorpil on heroic. Might be patched now tho. This was before I made my PR, like a week before official SL release

Nyeriah (approved):

PR #15819, opened 2023-04-04
fix(DB/Loot) Use Condition for Quest Drop "Two Halves Become One"

LeFroid:
In quest Two Halves Become One, the Jabbering Ghoul drops an item (good luck other half charm) which is used to create the quest reward item (good luck charm). Because it is an indirect quest item drop, we must add a condition instead of relying on QuestRequired from creature_loot_template.

## Changes Proposed:
-  Use conditions table to reward good luck other-half-charm only if player has the requisite quest (5051)

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.  In game, enter command `.q add 5051` to add quest
2. Go to Jabbering ghoul, `.go c id 10801` 
3. Kill npc and confirm you can loot good luck other-half-charm (item id 12722)

- [ ]
- [ ]


Gultask:
I don't get it, the quest works normally for me without the PR. Why the change?

LeFroid:
I thought the quest item wasn’t dropping when I last tried. Let me try again with a clean DB. This PR is unnecessary if the item is already dropping of course.

Gultask:
No worries

LeFroid:
Looks like this isn't needed afterall, thanks

Issue #14594, opened 2023-01-09
[Quest] A Spirit Guide - Ancestral Spirit Wolf aura


BeefSteaker:
https://github.com/chromiecraft/chromiecraft/issues/4727
### What client do you play on?

enGB

### Faction

Horde

### Content Phase:

61-64

### Current Behaviour

Ancestral Wolf Spirit is summoned as a companion
"Ancestral Spirit Wolf Buff Timer" buff is visible

![image](https://user-images.githubusercontent.com/11332559/211364436-8dde0397-f339-4bbc-b162-4996877aeea1.png)

When the Ancestral Spirit Wolf is summoned at Krun Spinebreaker's corpse, the spirit wolf does not generate the "[Guided by the Spirits](https://www.wowhead.com/tbc/spell=29938/guided-by-the-spirits)" aura as soon as it spawns, but instead at the very end of the escort, leading to a confusing and buggy appearance.

### Expected Blizzlike Behaviour

When the Ancestral Spirit Wolf is summoned, it should begin to generate the "[Guided by the Spirits](https://www.wowhead.com/tbc/spell=29938/guided-by-the-spirits)" aura as soon as it spawns.

### Source

https://www.wowhead.com/tbc/quest=9410/a-spirit-guide

Video showing the aura being applied correctly:
https://www.youtube.com/watch?v=ovCoEXxznPc&t=110s

### Steps to reproduce the problem

1. Accept the quest "[A Spirit Guide](https://www.wowhead.com/tbc/quest=9410/a-spirit-guide)"
2. Go to Krun Spinebreaker
3. Summon the Ancestral Spirit Wolf
4. Should receive aura

### Extra Notes

The aura itself works and the quest is completeable, but the aura is not generated immediately when the spirit wolf spawns. This leads to the quest appearing bugged and incompleteable. I only discovered that the quest in fact works by accident while I was writing this report and noticed that, while AFK at the Mag'har NPC, I had become friendly for a moment.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/c05c45c51d3d414ab3bec40e3e4bc2b5c10a977e

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[...]

PR #15892, opened 2023-04-09
fix(DB/Quest): Rework 'A Spirit Guide' (9410)

LeFroid:
Fixes behavior of ancestral spirit wolf

## Changes Proposed:
-  Add waypoint to spirit wolf
-  Player can now follow wolf to Mag'har camp, instead of the wolf following the player wherever they go

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14594

## Tests Performed:

- Tested quest to completion

## How to Test the Changes:

1. .q add 9410
2. .go c id 17062
3. Use item given in step 1 to summon spirit wolf (item 23669)
4. Follow wolf to Gorkan Bloodfist (16845)

- The aura sets Mag'har faction to friendly when within range of the spirit wolf, but if the player never leaves the radius of the wolf from the moment it's summoned, the faction change is not applied. That is to say, the player has to leave the range of the aura, and walk back into range for it to apply properly. I am not sure why this is.
- The NPC Earthcaller Ryga has a scripted event, not necessary for quest completion, which is not working. When the wolf finishes the waypoint, it triggers the smart script by setting the data field 0 of Ryga (id 17123) to value 1 (uses combination of `SMART_EVENT_DATA_SET` and `SMART_EVENT_SET_DATA`). Even when manually running `.npc set data 0 1` which should trigger the same, nothing happens. Perhaps this is a larger issue with smart script type `SMART_EVENT_DATA_SET`
- After testing this change a second time, the wolf was despawned right before reaching the final waypoint (did not happen on first test). This makes the change **unacceptable** at the moment.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1680983241816980712.sql
+++ b/data/sql/updates/pending_db_world/rev_1680983241816980712.sql
@@ -1,0 +1,134 @@
+-- Ancestral Spirit Wolf SAI (Source: Sniffs & https://www.youtube.com/watch?v=xFF8DOSeMcg)
+SET @ID := 17077;
+UPDATE `creature_template` SET `speed_walk` = 1.6, `speed_run` = 1, `unit_flags` = 768, `AIName` = "SmartAI", `ScriptName` = "" WHERE `entry` = @ID;
+DELETE FROM `smart_scripts` WHERE `entryorguid` = @ID AND `source_type` = 0;
+DELETE FROM `smart_scripts` WHERE `entryorguid` BETWEEN @ID*100+0 AND @ID*100+3 AND `source_type` = 9;
+INSERT INTO `smart_scripts` (`entryorguid`,`source_type`,`id`,`link`,`event_type`,`event_phase_mask`,`event_chance`,`event_flags`,`event_param1`,`event_param2`,`event_param3`,`event_param4`,`event_param5`,`action_type`,`action_param1`,`action_param2`,`action_param3`,`action_param4`,`action_param5`,`action_param6`,`target_type`,`target_param1`,`target_param2`,`target_param3`,`target_param4`,`target_x`,`target_y`,`target_z`,`target_o`,`comment`) VALUES
+(@ID,0,0,0,54,0,100,0,0,0,0,0,0,80,@ID*100+0,0,0,0,0,0,1,0,0,0,0,0,0,0,0,"Ancestral Spirit Wolf - On Just Summoned - Run Script"),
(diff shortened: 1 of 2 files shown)


Gultask:
What's the source of the waypoints?

Gultask (commented):
Use single quotes instead of double quotes

data/sql/updates/pending_db_world/rev_1680983241816980712.sql
```diff
@@ -0,0 +1,134 @@
+-- Ancestral Spirit Wolf SAI (Source: Sniffs & https://www.youtube.com/watch?v=xFF8DOSeMcg)
+SET @ID := 17077;
+UPDATE `creature_template` SET `speed_walk` = 1.6, `speed_run` = 1, `unit_flags` = 768, `AIName` = "SmartAI", `ScriptName` = "" WHERE `entry` = @ID;
```
Gultask, line 3: Speeds are already sniffed: https://github.com/neifion-00000000/azerothcore-wotlk/blob/76e9f511a1d3f08f3fe9d224c7ae488c39930a5c/data/sql/updates/pending_db_world/rev_1672823650132911600.sql
```suggestion
UPDATE `creature_template` SET `unit_flags` = 768, `AIName` = "SmartAI", `ScriptName` = "" WHERE `entry` = @ID;
```

Gultask:
Thanks for the PR!
Is this a cherry-pick?

acidmanifesto:
> Thanks for the PR! Is this a cherry-pick?

https://github.com/TrinityCore/TrinityCore/commit/c905af818c81d41d5e0b025de97c0cdbd1b417cb#diff-236b08d19a9b383633264786da628ee1210eafdee55addce2e318ba526e48a7c

LeFroid:
Thanks @acidmanifesto 
I will make your proposed changes @Gultask . In the meantime, I tested this change a second time and found the spirit wolf to despawn right before reaching the final waypoint. This made the quest incompletable. I will have to look further into root cause

pklloveyou:
image
I fixed them through hacker code

data/sql/updates/pending_db_world/rev_1680983241816980712.sql
LeFroid: Updated accordingly, but still facing the issue where the script is behaving inconsistently. Can we add a WIP label to the PR? I don't know how or don't have permission to do so.

LeFroid:
Closing this PR

PR #15901, opened 2023-04-10
update En'kilah Hatchling unit_flags 

cTrinity:
add 256 Unable to fight or communicate with players

 update En'kilah Hatchling(id = 25388,25390) unit_flags add 256 (Unable to fight or communicate with players)


Gultask (changes requested):
Alright, 3 things 😄

1. Do not delete or insert creature_template entries, only modify them. Using bitmask operations in this case:

```
UPDATE (...) SET `unit_flags`=`unit_flags`|FLAG|FLAG WHERE (...);
```

2. unit_flags are sniffed, so you need a source for the change. Either another core such as TC would be good too. Otherwise we can check sniffs to be certain

3. Remember to name your PRs correctly according to the standards in our wiki. In this case I'll re-name it for you.

In any case, thanks for the PR!

cTrinity:
Thank you for reminding me. I'm not very familiar with the rules

heyitsbench:
I also ask that you stick to and properly fill out the PR template.

PR #15931, opened 2023-04-13
fix(DB):  reset Scourgelord Tyrannus

### LINKED ISSUE #15824: [Pit of Saron]: Scourgelord Tyrannus doesn't restart when you die
### Current Behaviour

Scourgelord Tyrannus doesn't restart when you die

should reset the event correctly when you die against Scourgelord Tyrannus

### Source

_No response_

1 . go Pit of Saron
2. kill the first two bosses
3. go to Scourgelord Tyrannus start combat and die

### Extra Notes

only tested in 5 player heroic mode, but in normal mode for sure there must be the same problem

### AC rev. hash/commit

f10928fa5cd845278b75f5c93c972a4b71216045

### Operating system

Ubuntu 20.04

### Custom changes or Modules

_No response_

Tony931023:
authored-by: @pklloveyou  
## Changes Proposed:
-  reset Scourgelord Tyrannus
-  

## Issues Addressed:

- Closes #15824

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1681146435796877700.sql
+++ b/data/sql/updates/pending_db_world/rev_1681146435796877700.sql
@@ -1,0 +1,2 @@
+-- Reset Scourgelord Tyrannus 
+UPDATE `creature_template` SET `flags_extra` = `flags_extra` &~ 2147483648 WHERE `entry` IN (36658);


heyitsbench:
CC: @Nyeriah

Tony931023:
@heyitsbench 
to give you the credits of the arrangement is it?
authored-by: @pklloveyou

heyitsbench:
> to give you the credits of the arrangement is it?

Don't know how it's done on Github web, as I do it with Github Desktop.

Kitzunu:
`Co-authored-by: username <email@email.com>`

Kitzunu:
Email can be found if you go to a previous commit and add `.patch` to the end of the URL

Nyeriah:
This is probably wrong as it doesn't fix the issue but remove his despawn on evade mechanic, he likely should despawn on evade. Needs to be checked on retail

pklloveyou:
Yes, it's just a temporary solution

pklloveyou:
I think we should make the ice dragon disappear as well

FrancescoBorzi:
what's the status of this PR?

GastonCK:
Tested, now it resets after a wipe :D

Before (no reset)
WoWScrnShot_063023_115650

After (reseted)
WoWScrnShot_063023_120828

GastonCK:
@Nyeriah can we merge this?

Nyeriah:
> @Nyeriah can we merge this?

No since it's not how its intended to work but a workaround to an issue this PR does not address, you are free to use this temporarily if you need to though.

pangolp:
I have the suspicion that if we remove the flag, what we are going to do is make the npc stop being immune at all times, even during the dialogue. So I think that actually, we should try something else, more related to the same boss script. I'll see what I can find out.

Nyeriah:
Closing it for the time being considering that at this time the info we have is that he should despawn on evade, therefore this is not the correct way to fix this

Issue #15952, opened 2023-04-14
Spirit based health regen is half of what it should be


Tereneckla:
### Current Behaviour

Ticks for the value that is displayed on the character sheet when you hover over spirit

### Expected Blizzlike Behaviour

Ticks for twice that value as ticks happen every 2 seconds and the sheet shows the regen per second

### Source

2 seconds / 1 second = 2

### Steps to reproduce the problem

target self
.damage 1000
note down the value restored and the time between ticks
compare to value displayed by hovering over the spirit stat

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/539801377fb96f2c3377e051fa9264315501fd1e

### Operating system

Windows 10

PR #15955, opened 2023-04-14
fix (Core/Player) Double health regen from spirit

Tereneckla:
## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15952

## SOURCE:

Script is from PaperDollFrame
PaperDollFrame shows Health Regen / second
Health Regen ticks are every 2 seconds

## Tests Performed:

- tested ingame

## How to Test the Changes:

target self
.damage 1000
note down the value restored and the time between ticks
compare to value displayed by hovering over the spirit stat

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -5138,7 +5138,7 @@ float Player::OCTRegenHPPerSpirit()
     if (baseSpirit > 50)
         baseSpirit = 50;
     float moreSpirit = spirit - baseSpirit;
-    float regen = baseSpirit * baseRatio->ratio + moreSpirit * moreRatio->ratio;
+    float regen = (baseSpirit * baseRatio->ratio + moreSpirit * moreRatio->ratio) * 2;
     return regen;
 }
 


Kitzunu:
Where is this from? How can it be confirmed?

Tereneckla:
As described in the PR:
The paperdoll formula shows Health Regen / second
This formula was copied for the server, but Health Regen ticks happen every 2 seconds

To 100% confirm someone should log into wotlk classic

Kitzunu:
I will rephrase then, what is a paperdoll formula 😅

Tereneckla:
It's the one used by the client to show the transformed value of various stats
For spirit for example
image

As per the comment https://github.com/azerothcore/azerothcore-wotlk/pull/15955/files#diff-baadebd8cd1117ca48225f316a5ab3fd5fd55b20963394d302341147183db067R5135 in the Life Regen formula it is based on that

Currently every tick of life regen is as high as the paperdoll says, but paperdoll is per second and ticks happen every 2 seconds

Kitzunu:
Got it, thanks for explaining

Kitzunu (approved):

Issue #13839, opened 2022-11-15
Creature_formations cannot handle overlapping formations


Gultask:
### Current Behaviour

It's currently not possible to use creature_formations to link the same member to 2 different leaders, making some Blizzlike mechanics hard to implement.

If both leaderGUID and memberGUID are made primary keys, making the query is possible but one formation will work while the other will not.

### Expected Blizzlike Behaviour

In the image below, all of the packs in red are linked together. And, if the boss is pulled before all of the packs are dealt with, they will aggro as well.

![image](https://user-images.githubusercontent.com/100873791/201969296-275fa093-08f5-437e-a78b-54ab16489299.png)

### Source

Retail

### Steps to reproduce the problem

<details><summary>SQL Query</summary>
ALTER TABLE `creature_formations`
	DROP PRIMARY KEY,
	ADD PRIMARY KEY (`memberGUID`, `leaderGUID`) USING BTREE;

-- Static Formations
DELETE FROM `creature_formations` WHERE `leaderGUID` IN (83304,83306,83309,83311,83314,83316,83332,83335,83338,83340,83347,86396,91124,91134,91139,91151,91154,91162,91163,91164,91167,91174,91179,91182,91186,91188,91197,91204,91208,91212,91218,91223,91226,91231,91238,91242) AND `memberGUID` IN (83304,83305,83306,83307,83309,83310,83311,83312,83313,83314,83315,83316,83317,83318,83319,83320,83321,83332,83333,83335,83336,83338,83339,83340,83341,83342,83343,83347,83348,86396,91121,91122,91124,91125,91126,91134,91135,91136,91137,91138,91139,91140,91141,91142,91143,91151,91152,91153,91154,91155,91156,91162,91163,91164,91165,91166,91167,91168,91169,91170,91174,91175,91176,91179,91180,91182,91183,91184,91186,91187,91188,91189,91190,91194,91197,91198,91199,91200,91201,91202,91203,91204,91205,91206,91208,91209,91210,91212,91213,91214,91218,91219,91220,91221,91222,91223,91224,91225,91226,91227,91228,91229,91231,91232,91233,91238,91239,91242,91243,91244) AND `groupAI` IN (1, 3);
INSERT INTO `creature_formations` (`leaderGUID`, `memberGUID`, `dist`, `angle`, `groupAI`) VALUES
(83304, 83304, 0, 0, 3),
(83304, 83305, 0, 0, 3),
[...]

PR #15994, opened 2023-04-18
fix(DB/ShadowLabyrinth): Blackheart the Inciter's room link

Si1ker:
## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5434

## Tests Performed:

- Tested
image


Gultask:
Those member guids are already in use to form formations around the arena.
creature_formations can't handle overlapping formations: https://github.com/azerothcore/azerothcore-wotlk/issues/13839
It's an ongoing problem. It needs to be handled by script or by reworking the system

Si1ker:
what's this thrash? xd

I noticed they were already in DB but served no purpose, they were in a 1:1 relation which has no effect afaik

Gultask:
> what's this thrash? xd

dist is set to 100, should be 0

> I noticed they were already in DB but served no purpose, they were in a 1:1 relation which has no effect afaik.

The room is divided into several groups of 4 and 6-mob packs that are linked together and pull together. Without it they rely on assist calls, which shouldn't be the case. All instance mobs are the same in this regard.
It's really a limitation of the core that formations can't overlap. At the very least it **is** possible to set the boss to Call for Assist on pull, which is cleaner, imo.

Si1ker:
> > what's this thrash? xd
> 
> dist is set to 100, should be 0
> 
> > I noticed they were already in DB but served no purpose, they were in a 1:1 relation which has no effect afaik.
> 
> The room is divided into several groups of 4 and 6-mob packs that are linked together and pull together. Without it they rely on assist calls, which shouldn't be the case. All instance mobs are the same in this regard. It's really a limitation of the core that formations can't overlap. At the very least it **is** possible to set the boss to Call for Assist on pull, which is cleaner, imo.

edit: spoiler - not trying to argue, just expressing my point xd

Just changed dist, and now is passing the build.

In the other hand, two things for why I belive is ok as it is here.

1- Before this change, they had a creature_formation with themselves (no group pool, literally 1:1 relation), is not even configured at a pack level to work as you described. Resume: They're not linked in packs even by default.

2- Call for help on aggro prevents any possible scenario of trying to get just 1 from the packs, I just tested.
You mentioned "They rely on assist calls, which shouldn't be the case" - this just gives me more doubts ...
a- based on what should not be the case?
b- Are there literally 0 dungeons groups that relly only in CallForHelp on aggro and every single of them are configured with creature_formations? (we both know the answer..)

Gultask:
> Before this change, they had a creature_formation with themselves (no group pool, literally 1:1 relation), is not even configured at a pack level to work as you described. Resume: They're not linked in packs even by default.

image

> based on what should not be the case?

My own experience, mostly. In Blizz CallForHelp is a variable, not static like in AC. There are many unwanted pulls in Botanica for that reason.

My thoughts were that Blackheart was easy to implement a callforhelp function on aggro because the furthest NPC in the room is about 90y away, which is far enough that we won't have unwanted pulls

Si1ker:
> > Before this change, they had a creature_formation with themselves (no group pool, literally 1:1 relation), is not even configured at a pack level to work as you described. Resume: They're not linked in packs even by default.
> 
> image
> 
> > based on what should not be the case?
> 
> My own experience, mostly. In Blizz CallForHelp is a variable, not static like in AC. There are many unwanted pulls in Botanica for that reason.
> 
> My thoughts were that Blackheart was easy to implement a callforhelp function on aggro because the furthest NPC in the room is about 90y away, which is far enough that we won't have unwanted pulls

@Gultask ok, seems I missed the others connected to others, will do cpp + some improvements/ref.
check now, nothing weird getting pulled outside room.

Gultask:
Awesome, best of both worlds 😄

Nyeriah:
cc @Gultask

Gultask (approved):

Issue #16127, opened 2023-04-29
(The Blood Furnace) lever is not work, can not continue the next content


WildBlizzard:
### Current Behaviour

Using the "lever" afterwards, the prisoner door has no subsequent activation work, unable to trigger combat and thus unable to normally proceed with subsequent dungeon content.
![image](https://user-images.githubusercontent.com/32586354/235309477-f1bd0954-7ea0-4272-b09e-a94cffa1e72c.png)

### Expected Blizzlike Behaviour

After clicking the lever, the surrounding prisoner doors automatically open, fight and defeat the prisoner inside, at this time the central door opens, the boss enters and triggers the boss battle.

### Source

https://www.youtube.com/watch?v=dhSjk2-4zLA
6:18 - 7:50

### Steps to reproduce the problem

go to the instance (The Blood Furnace) , after first boss, you will see that issue.

### Extra Notes

This is the first time I have raised an issue, please notify me promptly if there is anything wrong, thank you for the trouble!

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/pull/15026

### Operating system

CentOS7.9

Nyeriah:
That is not a valid AC revision, also make sure to update your core before reporting new issues.

WildBlizzard:
> That is not a valid AC revision, also make sure to update your core before reporting new issues.

Sorry, I don't know if the above should add my AC version, here to supplement it, my AC code version is 00e42645b8b625e04de141b4aa9aabdd114dda08

LeFroid:
I closed https://github.com/azerothcore/azerothcore-wotlk/issues/15873 because @Gultask was not able to reproduce the issue, but I've been facing the same problem

I made several changes to `boss_broggok.cpp` to try fixing it, eventually this version worked for me in both heroic and normal versions
```cpp
enum eEnums
{
    SAY_AGGRO               = 0,

    SPELL_SLIME_SPRAY       = 30913,
    SPELL_POISON_CLOUD      = 30916,
    SPELL_POISON_BOLT       = 30917,
    SPELL_POISON            = 30914,
    SPELL_SUMMON_INCOMBAT_TRIGGER = 26837,
};

class boss_broggok : public CreatureScript
{
    public:
        boss_broggok() : CreatureScript("boss_broggok") { }

        struct boss_broggokAI : public BossAI
        {
            boss_broggokAI(Creature* creature) : BossAI(creature, DATA_BROGGOK) { }

[...]

PR #16143, opened 2023-04-30
fix(Scripts/TheBloodFurnace) Broggok lever trigger

LeFroid:
## Changes Proposed:
-  Update go_broggok_lever to use GameObjectAI
-  Update method to obtain pointer to Broggok (was previously returning nullptr in my local)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16127 

## SOURCE:

See my comment here: https://github.com/azerothcore/azerothcore-wotlk/issues/16127#issuecomment-1528831063 
The root of the issue seems deeper than this particular script, and apparently affecting a subset of users. From my testing I found that a certain ScriptMgr hook was not being invoked at all when it was supposed to, in `GameObject::Use` (that being the gameobject hook `OnGossipHello`)

## Tests Performed:

- 
- 

## How to Test the Changes:

1. .tele TheBloodFurnace
2. Skip ahead to broggok room
3. Pull lever and ensure the event begins normally

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Outland/HellfireCitadel/BloodFurnace/instance_blood_furnace.cpp
+++ b/src/server/scripts/Outland/HellfireCitadel/BloodFurnace/instance_blood_furnace.cpp
@@ -39,6 +39,16 @@ ObjectData const creatureData[] =
     { NPC_KELIDAN, DATA_KELIDAN }
 };
 
+Creature *GetBroggok(InstanceScript *instance)
+{
+    Creature* broggok = instance->GetCreature(DATA_BROGGOK);
+    if (!broggok)
+    {
+        broggok = instance->GetCreature(NPC_BROGGOK);
+    }
+    return broggok;
+}
+
 class instance_blood_furnace : public InstanceMapScript
 {
 public:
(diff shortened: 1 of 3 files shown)


src/server/scripts/Outland/HellfireCitadel/BloodFurnace/instance_blood_furnace.cpp
```diff
@@ -39,6 +39,16 @@ ObjectData const creatureData[] =
     { NPC_KELIDAN, DATA_KELIDAN }
 };
 
+Creature *GetBroggok(InstanceScript *instance)
```
Nyeriah, line 42: Functions in C++ added outside the class scope are added as global symbols, meaning this will be available universally in the entire project

Nyeriah:
Thanks for the PR, but if the issue is a non functional data storage, this PR strays away from what is necessary to achieve it

LeFroid:
There were two issues I observed:

1. The GameObjectScript hook OnGossipHello was not being invoked in this dungeon, even when registered properly
2. The call to `instance->GetCreature(DATA_BROGGOK)` was returning `nullptr`

src/server/scripts/Outland/HellfireCitadel/BloodFurnace/instance_blood_furnace.cpp
LeFroid: Do you suggest making the change inline in both places where the function is invoked?

Nyeriah:
I just tried this again in both heroic and normal and the level is still working for me

LeFroid:
Closing as the user reporting the issue said it's been fixed in 590dcca82b9c6e5d5f49fb5f954b084a88070f84

PR #16279, opened 2023-05-15
refactor(scripts/Outland): World Boss model update

Si1ker:
tittle


elthehablo:
kara next? pog!

elthehablo:
> Isn't EventMap good? Why must we become TaskScheduler ?

It is more elegant imo, and more intuitive. But it was probably chosen because it is more efficient (don’t know for sure though)

Nyeriah (approved):

Issue #15910, opened 2023-04-10
[Hunter] Snake Trap


Tereneckla:
https://github.com/chromiecraft/chromiecraft/issues/5357
### What client do you play on?

enUS

### Faction

Both

### Content Phase:

Generic

### Current Behaviour
Damage against a target with 4641 Armor as a level 70 hunter
![image](https://user-images.githubusercontent.com/50233983/230974118-84a91721-ff66-4719-9e8d-9b1a4c7c3a19.png)
![image](https://user-images.githubusercontent.com/50233983/230974160-e31c86c1-ad96-47e9-bad8-cddef8cb3e6f.png)

They follow the hunter as sort of guardians
They count as pets and get damage bonuses from the beast mastery talents

### Expected Blizzlike Behaviour

These values probably are for level 80
Vipers hit for ~38-53 damage
Venomous Snakes hit for ~16-24 damage

Overall I'd say Vipers should deal ~1/5th of the current damage and Venomous Snakes 1/10th

They guard the original trap placement area when there is nothing to attack around
beast mastery talents have no influence

### Source

Looks like a lvl 73 hunter:
Snakes hits are ~12-14 and ~25-30
https://youtu.be/ULZYl0cVMlc
https://wowpedia.fandom.com/wiki/Snake_Trap?oldid=2341652

You can see them head back to the trap placement area here
https://youtu.be/ULZYl0cVMlc?t=200
https://youtu.be/ULZYl0cVMlc?t=214

On wotlk classic at level 70:
![image](https://user-images.githubusercontent.com/50233983/230973816-fb0d9f54-74d7-42c8-8dc2-41f18c7ef2d3.png)
![image](https://user-images.githubusercontent.com/50233983/230972802-6505ec79-5868-428f-9cf6-e94d8db5fdde.png)
![image](https://user-images.githubusercontent.com/50233983/230972807-d9961631-4bc3-493a-9292-322dbf64b9ca.png)
![image](https://user-images.githubusercontent.com/50233983/230972819-ff13cb66-cfec-485e-8e8b-5ab8fe2c18d2.png)

With 4641 Armor at level 70 that base damage means;
Venomous Snake 9-11 -> 21-26 damage per hit
Viper: ~20 -> ~47 damage per hit

### Steps to reproduce the problem

.cast 34600

### Extra Notes

https://wowgaming.altervista.org/aowow/?npc=19833
https://wowgaming.altervista.org/aowow/?npc=19921
[...]

PR #16057, opened 2023-04-22
fix(Scripts/Pet): hunter snake trap creatures base damage

Blaiseum:
## Changes Proposed:
-  Set base damage for hunter trap's snakes, accordingly with the report.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15910

## SOURCE:

* https://wowpedia.fandom.com/wiki/Snake_Trap?oldid=2341652
> [...]
> The Vipers hit for 38-53. 4 of these appear.
> 
> The Venomous Snakes hit for 16-24. 4 of these appear. 

* https://github.com/azerothcore/azerothcore-wotlk/issues/15910

## Tests Performed:

- Tested on modified server, with hunter (lev 80).
- Inserted/enabled additional logs on core (i.e., `entities.unit`); the base damage is congruent with the expected report.

## How to Test the Changes:

1. Set `snake trap` on enemy.
2. Check logs/damage before and after the patch.

PR #16326, opened 2023-05-19
fix(Core/Spells): Hunter snake trap damage

Nefertumm:
## Changes Proposed:
-  Remove random multipliers for pets that doesn't have a petlevelstats entry.
-  Add a formula to calculate the damage (it's not perfect but for now it works.)
-  Refactor some code.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15910

## SOURCE:

https://github.com/azerothcore/azerothcore-wotlk/issues/15910#issuecomment-1553915910
With the data provided, I generated a linear regression model for each formula.
Models for Venomous Snake: 
Min damage:
image

Max damage:
image

As we can see the models are significant BUT we need to check the residuals to see if they are normal...
image
As we can see the data for the maximum damage is not quite normal, but we can leave it as it is for now (this would explain some behaviour later)

Model plots (blue line is the model for the min damage, red line is the model for the max damage):
image
image

So now we can calculate the formula for both, min and max damage, being:
dmg_min = 0.51 * lvl - 24.7 +- 4.5  (where 4.5 is the error or margin, whatever you want to call it)
dmg_max = 0.57 * lvl - 23.8 +- 4.6 (same as above)

Conclusion: The models generated could not be the best ones for this dmg formula, since we can see on the plots that the data presents a step-like function instead of linear. 

Models for Viper:
Min damage:
image

Max damage:
image

This time we get, again, significant models for each one of them, with the addition that their residuals fulfil the normality test.
image

Model plots: (same as above)
image
image

With the models, we can calculate the formula again, being:
dmg_min = 1.2 * lvl - 56.8 +- 8.7
dmg_max = 1.37 * lvl - 59.8 +- 9.4

src code I made for calculations: [link](https://drive.google.com/file/d/1jGSlew9JGIc44LlAL_vlzXdsXtgiZB9i/view?usp=sharing)

## Tests Performed:

-  Tested in-game.

## How to Test the Changes:

1. Create a hunter, cast snake trap near an enemy.

- [ ] Maybe the stat that were present on pet_hunter were right? 🤔 
- [ ] Check if those stat multiplier doesn't fuck everything else.

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1348,6 +1348,14 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
                             SetBaseWeaponDamage(BASE_ATTACK, MAXDAMAGE, float(petlevel + (petlevel / 4)));
                             break;
                         }
+                    case NPC_VENOMOUS_SNAKE:
+                        SetBaseWeaponDamage(BASE_ATTACK, MINDAMAGE, float((petlevel / 2) - 25));
+                        SetBaseWeaponDamage(BASE_ATTACK, MAXDAMAGE, float((petlevel / 2) - 18));
+                        break;
+                    case NPC_VIPER:
+                        SetBaseWeaponDamage(BASE_ATTACK, MINDAMAGE, float(petlevel / 2 - 10));
+                        SetBaseWeaponDamage(BASE_ATTACK, MAXDAMAGE, float(petlevel / 2));
+                        break;
                     case NPC_GENERIC_IMP:
                     case NPC_GENERIC_VOIDWALKER:
                         {
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -1349,12 +1349,12 @@ bool Guardian::InitStatsForLevel(uint8 petlevel)
                             break;
                         }
                     case NPC_VENOMOUS_SNAKE:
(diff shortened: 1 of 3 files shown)


heyitsbench:
Supersedes https://github.com/azerothcore/azerothcore-wotlk/pull/16057, I assume?

Nefertumm:
Indeed.

Gultask (approved):

Nefertumm:
Update: Models are calculated, waiting for feedback. I will update the code once I'm sure formulas given by the models are okay to use.

sassouiabd:
Good Job !

heyitsbench:
> Can this be applied to priest pet or its a different subject ?

Different issue.

Nefertumm:
An update on this matter. The model would work but with some issues, as to calculate more precise models we would need to gather more data, that way the models would be more robust than now. So we actually have 2 options, use the models as they are right now, or use the current applied formula which is a interpolation formula between the level 70 dmg and the level 80 dmg.

heyitsbench:
> as to calculate more precise models we would need to gather more data

Speaking to this, I will point out that I can spend a bit sometime soon on the Classic PTR and can flesh out the data points for levels 70/80 at least (since template characters). Obviously not the thirteen levels we'd like it to be, but do you think that'd be enough to refine the model at all?

Nefertumm:
It could but it might be precise for 70/80 and not for the levels in between, but anyway, it will improve the model for sure.
My recommendation is leave the formula as the interpolation for now, and do a follow up soon.

PR #16352, opened 2023-05-22
fix(Scripts/UBRS): Update UBRS/Pyroguard Emberseer script to fix issues and increase accuracy

kjack9:
## Changes Proposed:
- Remove EVENT_PLAYER_CHECK which was causing triggering the fight to work only sporadically and does not exist in the WotLK Classic version
- Break out pre-enter-combat phase to better match actual timing
- Pre-enter-combat sequence starts 2s after the adds are engaged (instead of 32s) to better match actual timing
- timing now matches DBM and multiple videos
- disable fire shield hack during pre-enter-combat to keep the fight from resetting
- remove faster add despawn which doesn't appear to be in the WotLK Classic version
- remove adds doing EVENT_ENCAGE every 10-20s, which was unnecessary and caused them to glitch move around the player

## Issues Addressed:

- None

## SOURCE:

Used these two videos as reference:
- https://youtu.be/roNyUYbM0JE?t=507 (Season of Mastery, good to watch timing of emotes and boss actions)
- https://www.youtube.com/watch?v=H0F02urENz0&t=283s (fight activated by only one player summoning)
- https://youtu.be/1tGdVHw9HME?t=227 (final version on live before rework, shows single player summoning and same timing)

## Tests Performed:

- start/restart the fight multiple times, checking timing with stopwatch app

## How to Test the Changes:

1. `.go gameobject 87838` (Blackrock Alter)
2. channel ritual for 5 seconds
3. 5 seconds later the fight should start and proceed as normal
4. note that the boss starts growing immediately, and becomes attackable 65 seconds later

- would be great to figure out why channeling the Blackrock Alter cancels after 5s instead of continuing (Uldaman altar for the final boss does the same thing)

### DIFF
--- a/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_pyroguard_emberseer.cpp
+++ b/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_pyroguard_emberseer.cpp
@@ -101,7 +101,7 @@ public:
             switch (data)
             {
                 case 1:
-                    events.ScheduleEvent(EVENT_PRE_FIGHT_1, 5s);
+                    events.ScheduleEvent(EVENT_PRE_FIGHT_1, 2s);
                     instance->SetBossState(DATA_PYROGAURD_EMBERSEER, IN_PROGRESS);
                     break;
                 case 2:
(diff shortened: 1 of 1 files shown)


Nyeriah:
In the video you posted as source he describes that it used to require 5 players and it had been updated recently to allow single player runs

kjack9:
> In the video you posted as source he describes that it used to require 5 players and it had been updated recently to allow single player runs

According to reports, he was updated some time before patch 3.0.3. So this patch is consistent with 3.3.5a. 

https://www.wowhead.com/wotlk/npc=9816/pyroguard-emberseer#comments:id=573149

The previous AC functionality (all players in the dungeon are supposed to be summoning) didn't work right AND didn't reflect how the summon ever worked.

kjack9:
Any outstanding concerns or questions about this change?

Nyeriah (approved):

Issue #16296, opened 2023-05-17
(Creature): Uvuros has no immunities.


heyitsbench:
### Current Behaviour

[Uvuros](https://wowgaming.altervista.org/aowow/?npc=21102) has no immunities.

### Expected Blizzlike Behaviour

He should be immune to roots, stuns, fear, and horror. Definitely susceptible to bleeds though.

### Source

Wrath Classic

### Steps to reproduce the problem

1. `.go c id 21102`
2. Cast any spell with the listed mechanics.
3. Observe his susceptibility.

### Extra Notes

Originally reported at https://github.com/chromiecraft/chromiecraft/issues/5582.

### AC rev. hash/commit

97dff23ba1edaaf29b410e33962d565bf7028df0

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

PR #16380, opened 2023-05-27
fix(DB/ShadowmoonValley): add immunities to Uvuros

alaxandir:
## Changes Proposed:
- Add immunities to Uvuros
- Immune to Fear, Root, Slow, Spell Interrupt, Silence, Snare, Freeze, Horror

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16296

## SOURCE:
- According to user comments on WowHead from original Wrath, this mob has certain immunities which have been added. https://www.wowhead.com/wotlk/npc=21102/uvuros
- In addition the PR requested these immunities be added for Bliz-like expectation

## Tests Performed:
- Tested Mage and Warrior in game
- Immune to appropriate abilities such as frost nova root, fear, counter spell, the slow effect of frost bolt.

## How to Test the Changes:
`go c id 21102`
Cast any spell with the listed mechanics.
Observe his susceptibility.

- [ ] Noticed that Fel-Reaver also has no immunities except for Bleed, unsure if this is blizzard like. I admittedly don't know how to test this to observe expected behaviours short of playing on the official classic wrath server, which I don't have.

Discord #AiurJordan9442


heyitsbench:
On Wrath Classic, he is susceptible to Cone of Cold's slow effect.

heyitsbench:
You need to add `.sql` to the end of the filename by the way.

alaxandir:
Sorry, my first PR learning git. Thanks for the help.

heyitsbench:
> Sorry, my first PR learning git. Thanks for the help.

No problem, thanks for the PR! ❤️

elthehablo:
might want to change your title. for db fixes we generally would do something as follows in this case:

`fix(DB/ShadowmoonValley): add immunities to Uvuros`

for future reference, immunities are always really hard to prove. sometimes it's possible to crossreference with other dbs (TC for instance) to see what they are doing and take values from there, but those are also not always correct

elthehablo:
aditionally for sql the convention is to make them look like this:
```sql
01 --
02 MY SQL STATEMENT WHERE `id` = 0;
03 
```

heyitsbench:
> aditionally for sql the convention is to make them look like this:

It's fine how it is. 🙂

Nyeriah:
The bitwise operator should be used when updating bitmask fields

```sql 
UPDATE `creature_template` SET `mechanic_immune_mask` = `mechanic_immune_mask` |XXX|YYY WHERE `entry` = XXXX;
```

elthehablo:
> > aditionally for sql the convention is to make them look like this:
> 
> It's fine how it is. 🙂

it is fine is that it will work the same. it is less readable tho

heyitsbench:
> it is fine is that it will work the same. it is less readable tho

Adding a line of just dashes has no impact on readability IMO. If you mean the bitmask operator, I do agree.

elthehablo:
> > it is fine is that it will work the same. it is less readable tho
> 
> Adding a line of just dashes has no impact on readability IMO. If you mean the bitmask operator, I do agree.

the latter :)

heyitsbench:
> btw is he poly immune?

He is on Classic.

alaxandir:
requested changes made, added polymorph immunity

Nyeriah (approved):

Nyeriah:
Thanks for the PR 👍

elthehablo:
Good work on first PR!

PR #16396, opened 2023-05-29
fix(Scripts/ShatteredHalls): hack fix to prevent pets from pulling the non-attackable peons in Nethekurse room

elthehablo:
## Changes Proposed:
-  intro now starts on lockpick
-  

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5692

## SOURCE:

See issue

## Tests Performed:

- tested the immunity part
- the lockpick part has to be thoroughly tested

## How to Test the Changes:

1.
2.
3.

- [x] make a better way to solve this problem also removing the non-attackable hack fix that was created in the first place
- [x] make it so door can be picked and the intro starts as well


elthehablo:
tested and works but https://github.com/chromiecraft/chromiecraft/issues/5692#issuecomment-1567150574 is still an issue present with this hack, albeit a minor one

Nyeriah (approved):

Issue #16237, opened 2023-05-12
[Mechanar] Mechanar Tinkerer low hp explosion does not deal damage


balleny:
### Current Behaviour

Mechanar Tinkerer charge at low hp but dont explode - they just die

### Expected Blizzlike Behaviour

after charging a target, they should cast the explosion (1 sec casttime) and die by this ability.

### Source

https://wowgaming.altervista.org/aowow/?npc=19716

https://www.icy-veins.com/tbc-classic/the-mechanar-dungeon-guide

These deal high damage to all nearby enemies constantly. Spread out to avoid the high AoE damage; they will also charge to a random enemy when low on health.

https://youtu.be/O59tAY_ni0w?t=194
https://youtu.be/5x9UY0pQCSw?t=418

### Steps to reproduce the problem

.

### Extra Notes

https://wowgaming.altervista.org/aowow/?npc=19716
https://wowgaming.altervista.org/aowow/?spell=35058

Chromiecraft report
https://github.com/chromiecraft/chromiecraft/issues/5270

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/5e9b88a9a9ddf510ee2e7a95861988a1fb3e54ca

### Operating system

Ubuntu 20.04

### Custom changes or Modules

[mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
[mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
[mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
[mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
[mod-eluna](https://github.com/azerothcore/mod-eluna)
[mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
[mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[mod-progression-system](https://github.com/azerothcore/mod-progression-system)
[mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #16411, opened 2023-05-31
fix(Core/SmartScripts): Allow SAI_ACTION_DIE to accept params

Nefertumm:
## Changes Proposed:
-  Modify the SAI_ACTION to accept a timer as param1, this allows us to execute the action after x ms.
-  Modify Mechanar Tinkerer SAI to fix their Nether Explosion cast.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5270
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16237

## Tests Performed:

- Tested in-game.

## How to Test the Changes:

1. Go to The Mechanar and lower the hp of one Mechanar Tinkerer
2. It should charge and immediately start casting Nether Explosion.
3. After the cast, you should take damage and the creature should die immediately.

- [x] Update wiki x.x


elthehablo:
Tested by me as well! Works well. Any concerns about existing creatures using the same action type? Probably not, because they all have that parameter unused which should not affect it, right?

Nyeriah (approved):

PR #16446, opened 2023-06-02
fix(DB/Creature): Archaedas despawns his adds after dying

kjack9:
## Changes Proposed:
-  adds additional SAI actions to despawn the various add types in the room upon death

## Issues Addressed:

- adds continue to attack players after Arcahedas dies

## SOURCE:

https://youtu.be/MSlqelUouNQ?t=142 - shows the adds despawning after boss death

## Tests Performed:

- tested in-game

## How to Test the Changes:

1. `.go creature 33537`
2. Kill Archaedas
3. Observe adds despawning

- this is an improvement but doesn't yet match Blizzlike. Looks like the spell that is supposed to take out all the adds isn't working. I don't have the skills to fix that yet, but I can at least get us closer and save players a run back if they defeat the boss and then get killed by adds.


Kitzunu:
in this case your title should be `fix(DB/Creature)`

We use the "base" class, such as Creature/Instance/GameObject/Spell etc. in the second part, never table names 👌

Nyeriah (approved):

Kitzunu:
Actually my bad, I was wrong.
I have learned that when a PR only touches smart scripts and nothing else creature related it should be DB/SAI

Sorry for that 😅

Issue #16433, opened 2023-06-01
(Creature) Defias Cutpurse not doing spells


therelus:
### Current Behaviour

Defias Cutpurse https://wowclassicdb.com/wotlk/npc/94 not doing spells, only auto attack

### Expected Blizzlike Behaviour

This NPC must do spells

### Steps to reproduce the problem

1.Found them
2.Fight with them
3.Watch if they are using their spells

### AC rev. hash/commit

AzerothCore rev. https://github.com/azerothcore/azerothcore-wotlk/commit/96a5224d927fde51860619abfe32e2fa5b3f14fa+ 2023-04-22 07:25:31 +0700 (npcbots_3.3.5 branch) (Win64, Release, Static) (worldserver-daemon)

### Operating system

x64

elthehablo:
Spell is assigned in `smart_scripts`. If it's not working it might be a problem with the spell itself. Have you tried turning your back to the mob?

elthehablo:
I assume the problem here lies in the fact that the ability requires energy. So changing the castflags should fix this

therelus:
Yea i turned back but nothing, but I´m using a revision of AzerothCore rev. https://github.com/azerothcore/azerothcore-wotlk/commit/96a5224d927fde51860619abfe32e2fa5b3f14fa+ 2023-04-22, maybe they fixed it already :P

elthehablo:
> Yea i turned back but nothing, but I´m using a revision of AzerothCore rev. https://github.com/azerothcore/azerothcore-wotlk/commit/96a5224d927fde51860619abfe32e2fa5b3f14fa+ 2023-04-22, maybe they fixed it already :P

Check referenced pull request

elthehablo:
What is your source on this?

elthehablo:
I will close this issue as the spell should not be present on wotlk. The report was correct that the spell was not working so I will close as completed and make a PR to remove the SAI from Defias Cutpurse

PR #16471, opened 2023-06-04
fix(DB/ElwynnForest): allow Defias Cutpurse to use their Backstab ability

elthehablo:
## Changes Proposed:
- add 2 castflag to allow them to use their backstab ability which usually costs 60 energy, which they don't have

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16433

## SOURCE:

Ability is in db. Mob cannot cast it because of energy requirements 

## Tests Performed:

- tested ingame. works
- 

## How to Test the Changes:

1. Go to Elwynn Forest towards the north where the Defias Cutpurse camps are
2. Turn your back
3. Should get backstabbed

It could technically be possible they use a different backstab version. But that would require sniffing. According to wowhead they do seem to use a backstab that uses energy.


Gultask:
There is an event BEHIND_TARGET which would work better for Backstab

elthehablo:
> There is an event BEHIND_TARGET which would work better for Backstab

Yeah. Didn't know that existed, but that is more elegant than firing the spell and just having failing conditions most of the time

elthehablo:
Unfortunately this event doesn't seem to make them trigger the cast. I will revert it back for now. As the condition is within the spell to be behind the target, and this is not circumvented by any castFlag, this works. It is not elegant, but it does work.

Vyperhand:
> Unfortunately this event doesn't seem to make them trigger the cast. I will revert it back for now. As the condition is within the spell to be behind the target, and this is not circumvented by any castFlag, this works. It is not elegant, but it does work.

Idea/brainstorming:  Is there a known working rogue NPC elsewhere with backstab that code could be cannibalized from to optimize it and make it consistent with similar mobs?

Gultask:
> Idea/brainstorming: Is there a known working rogue NPC elsewhere with backstab that code could be cannibalized from to optimize it and make it consistent with similar mobs?

Yeah we can query from BEHIND_TARGET event. Dunno why this one didn't work

Gultask:
@elthehablo missing file extension

elthehablo:
> @elthehablo missing file extension

Lmao yeah. Wtf was I on

heyitsbench:
> Lmao yeah. Wtf was I on

Clearly not .SQL. 🙂

elthehablo:
> > Lmao yeah. Wtf was I on
> 
> Clearly not .SQL. 🙂

image

Gultask:
Actually, no, there is an issue with this. Triggered flags don't take range into question. So this will make them cast from infinite range. This is likely the wrong spell ID.

elthehablo:
Could you investigate please?

heyitsbench:
Would it be possible to find a video of this spell being cast by this NPC? Tried to get a sniff of it on Classic but they would not cast it no matter what I did.

elthehablo:
Very likely then that this ability may not exist? It should be cast quite regularly if you face your back towards them

Vyperhand:
> Would it be possible to find a video of this spell being cast by this NPC? Tried to get a sniff of it on Classic but they would not cast it no matter what I did.

Can't find a video yet, but one of the top comments on wowhead references it.
https://www.wowhead.com/wotlk/npc=94/defias-cutpurse#comments

elthehablo:
> > Would it be possible to find a video of this spell being cast by this NPC? Tried to get a sniff of it on Classic but they would not cast it no matter what I did.
> 
> Can't find a video yet, but one of the top comments on wowhead references it. https://www.wowhead.com/wotlk/npc=94/defias-cutpurse#comments
> 
> with the age of that comment, this could be one of the things that didn't get re-solved in Classic, perhaps? Or otherwise perhaps a victim of the starting zone nerfs?
> 
> might check vanilla classic to see if the behavior is any different for a sniff.

Please do. But yeah could be a vanilla thing that didn't get added to classic.

heyitsbench:
Checked on the Classic Era PTR, the backstab ID is correct (53) and happens very often and very easily. Still does not happen on Wrath Classic however, I suspect removed with starting zone nerfs.

elthehablo:
I vote keep it. Merge merge merge. Zugzug.

Vyperhand:
> Checked on the Classic Era PTR, the backstab ID is correct (53) and happens very often and very easily. Still does not happen on Wrath Classic however, I suspect removed with starting zone nerfs.

That sounds pretty darn deliberate to me, if it's present in vanilla classic and gone by wrath classic.  I can't imagine they un-solved it moving classic forward.

ZhengPeiRu21:
Based on Bench's comparison with Classic, I agree that this skill probably was removed as part of the starting zone nerfs, similar to the Defias Pillager's fireball spell being changed to a weaker one, so not having it is correct WotLK behavior. (Although it could be something else, like an unintentional bug by Blizzard at the time or just making it much less common.)

If you don't mind, though, I will include your fix in Individual Progression as part of "un-nerfing" the Vanilla world.

elthehablo:
> Based on Bench's comparison with Classic, I agree that this skill probably was removed as part of the starting zone nerfs, similar to the Defias Pillager's fireball spell being changed to a weaker one, so not having it is correct WotLK behavior. (Although it could be something else, like an unintentional bug by Blizzard at the time or just making it much less common.)
> 
> If you don't mind, though, I will include your fix in Individual Progression as part of "un-nerfing" the Vanilla world.

By all means!

elthehablo:
Closing this PR then as well as the issue.

Issue #16325, opened 2023-05-19
(Holidays): Holiday dates must be filled for the next however many years.


heyitsbench:
### Current Behaviour

We have run out of Darkmoon Faires.

### Expected Blizzlike Behaviour

We need more Darkmoon Faires.

### Source

Assumption based on the established eternal presence of the Darkmoon Faire.

### Steps to reproduce the problem

1. Open the calendar.
2. Observe the Darkmoon Faire is no more.

### Extra Notes

We just need to add more dates for events in `holiday_dates`. Shouldn't be too difficult.

### AC rev. hash/commit

be48e3c2493b420c62c293897e321c0bf7c0483f

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

Kitzunu:
Is it just out of the calender? Because it shouldnt stop as the end time is 2030-01-01 06:00:00

heyitsbench:
> Is it just out of the calender? Because it shouldnt stop as the end time is 2030-01-01 06:00:00

Yes, it only affects the in-game calendar, the events still play out in the world as expected.

lemondsMs:
confirm

Darkmoon is displayed on the calendar

kerricsson:
Good day, players at CC discovered this bug in game and started to talk about it in sever discord (because now it's time for DMF).

Synful-Syn:
We should never have to manually feed by hand the upcoming recurrent world events such as Darkmoon Faire. It should have been automated in the first place. Otherwise, we'll just push the very same problem later.

heyitsbench:
Whichever solution comes first should be acceptable.

Gultask:
Could anyone see if this works?
https://gist.github.com/Gultask/c06f6dbb543594618164ffc2241c31ec

Uses the Classic schedule convention

Nyeriah:
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/47818697/a682602f-38db-471d-9ac1-075bf4e37058)

Doesn't seem to do anything

AFAIK that's one of AC's tables created with the idea of moving client data to the database, I am not sure if it's even implemented tbh

Gultask:
I definitely managed to switch dates around for *existing* entries, but new ones don't appear for me

heyitsbench:
[...]

PR #16474, opened 2023-06-04
fix(DB/Holidays): Add the next 75 Darkmoon Faires

Gultask:
## Changes Proposed:
-  
-  

## Issues Addressed:

- Progress https://github.com/azerothcore/azerothcore-wotlk/issues/16325

## SOURCE:

Follows the WoW Classic convention: Every 1st Sunday
Retail seems to not follow a consistent rule from what we've seen

## Tests Performed:

- Tested ingame
- 

## How to Test the Changes:

1. If you need to: https://www.nirsoft.net/utils/run_as_date.html
2. But there's already one running with this convention up to next sunday
3.

- [ ]
- [ ]


Gultask:
Problem is that it also deletes all entries from before the current date

Nyeriah (approved):

Nyeriah:
Works for me 👍

PR #16488, opened 2023-06-06
fix(DB/Worldboss): Lower Doom Lord Kazzak's damage

Nefertumm:
## Changes Proposed:
-  Lower DamageModifier from 90 to 35.

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5708

## SOURCE:

TC db

## Tests Performed:

- None.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1686083132284151700.sql
+++ b/data/sql/updates/pending_db_world/rev_1686083132284151700.sql
@@ -1,2 +1,2 @@
 --
-UPDATE `creature_template` SET `DamageModifier` = 35 WHERE `entry` = 18728;
+UPDATE `creature_template` SET `DamageModifier` = 65 WHERE `entry` = 18728;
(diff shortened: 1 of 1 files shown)


Gultask:
35 is the default used for every single boss before. We should re-calculate based on mindmg/maxdmg fields

Gultask (changes requested):
comment

data/sql/updates/pending_db_world/rev_1686083132284151700.sql
```diff
@@ -0,0 +1,2 @@
+--
+UPDATE `creature_template` SET `DamageModifier` = 35 WHERE `entry` = 18728;
```
Gultask, line 2:
```suggestion
UPDATE `creature_template` SET `DamageModifier` = 65 WHERE `entry` = 18728;
```

Gultask (approved):

PR #16506, opened 2023-06-09
chore(Core/MiscHandler): SendAreaTriggerMessage with integer parameter

pangolp:
## Changes Proposed:
-  In order to translate many of the emulator modules, it is necessary, in this case, to enable the SendAreaTriggerMessage method to accept an integer value as an argument, so that it can be sent from acore_string

## Issues Addressed:

- Closes

## Tests Performed:

- Windows 10
- In-game

## How to Test the Changes:

1. Actually this PR, is related to another one that I have in the mod-individual-xp, where previously, the messages were set inside the same module, but now they are read from the DB. The question is that in order to publish this PR, first I need this to be approved. https://github.com/azerothcore/mod-individual-xp/pull/35

### DIFF
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -718,6 +718,27 @@ void WorldSession::SendAreaTriggerMessage(const char* Text, ...)
     SendPacket(&data);
 }
 
+void WorldSession::SendAreaTriggerMessage(uint32 entry, ...)
+{
+    char const* format = GetAcoreString(entry);
+    if (format)
+    {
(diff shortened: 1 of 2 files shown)


pangolp:
@Nyeriah If you can check this PR, it would be interesting, to be able to send messages that are not only text in a SendAreaTriggerMessage, so we can use the acore_string messages for example. I want to use it to translate some community modules, like individual-xp for example. I have a stopped PR, because I need this change. Thanks.

pangolp:
Sorry for the mention, I hope not to bother you with it.

Nyeriah (approved):

PR #16509, opened 2023-06-10
chore(Core/ScriptMgr):Add Hook OnBeforeFillQuestLootItem

pangolp:
## Changes Proposed:

- The idea is to add a hook, which is used in a module, to be able to change the condition of the item and to allow all the members of the party to obtain the same item from the quest.
- In other words, if you do a quest, which consists of getting 12 items from an npc, if there are 5 people in the party, they should kill at least 60 npc, as long as the drop is at 100%. Because if the drop is lower, you should kill many more.
- Thanks to this, it is achieved that people do not abandon missions, which involve having to get a lot of items from an npc.

## Issues Addressed:

- Closes 

## Tests Performed:

- Windows 10
- In-game

## How to Test the Changes:

1. This change by itself does not work. It has to be accompanied by a module, which I will create once this code is working. Unless, of course, I have to make changes to it. I have tested it, without the module and with the module, and in both cases, the behavior was as expected.
2. Without the module, the behavior is normal. The group eliminates an npc, and at the time of obtaining the quest item (in case it falls), only one of them will be able to obtain it.
3. With the module, however, by modifying the behavior of the variable, all members of the group are able to put the quest item, making the quest less tedious to perform.

## Related module

https://github.com/pangolp/mod-quest-loot-party

## Screenshots

WoWScrnShot_061123_103825

WoWScrnShot_061123_103840

### DIFF
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -679,6 +679,8 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
     {
         LootItem& item = quest_items[i];
 
+        sScriptMgr->OnBeforeLootItem(player, item);
+
         // Quest item is not free for all and is already assigned to another player
         // or player doesn't need it
         if (item.is_blocked || !item.AllowedForPlayer(player, sourceWorldObjectGUID))
--- a/src/server/game/Loot/LootMgr.cpp
+++ b/src/server/game/Loot/LootMgr.cpp
@@ -679,7 +679,7 @@ QuestItemList* Loot::FillQuestLoot(Player* player)
     {
         LootItem& item = quest_items[i];
 
-        sScriptMgr->OnBeforeLootItem(player, item);
+        sScriptMgr->OnBeforeFillQuestLootItem(player, item);
 
         // Quest item is not free for all and is already assigned to another player
         // or player doesn't need it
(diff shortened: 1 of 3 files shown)


AnchyDev:
I have tested this using breakpoints and it seems to work fine, it's pretty straight forward.
However I would make some adjustments to the hook naming as it doesn't really reflect what it's doing.
I would probably name it `OnBeforeFillQuestLootItem` or something similar.

pangolp:
> I have tested this using breakpoints and it seems to work fine, it's pretty straight forward.
> However I would make some adjustments to the hook naming as it doesn't really reflect what it's doing.
> I would probably name it `OnBeforeFillQuestLootItem` or something similar.

You say to change the name? Of course it can be done.

AnchyDev:
> > I have tested this using breakpoints and it seems to work fine, it's pretty straight forward.
> > However I would make some adjustments to the hook naming as it doesn't really reflect what it's doing.
> > I would probably name it `OnBeforeFillQuestLootItem` or something similar.
> 
> You say to change the name? Of course it can be done.

Correct, the hook is not actually fired when looting but actually when the loot items (quest items specifically) are filled.
So it makes more sense to use a different name.

pangolp:
> Correct, the hook is not actually fired when looting but actually when the loot items (quest items specifically) are filled.
> So it makes more sense to use a different name.

I'll update it now.

pangolp:
@AzerothCoreDev Please, if someone can review it, I could really use this change in order to finish the module. Thank you.

pangolp:
Thanks @Nyeriah .

PR #16580, opened 2023-06-17
fix(scripts/Steamvault): Switch Mekgineer Steamrigger's enrage to Heroic mode only

Si1ker:
## Changes Proposed:
- Change (5min) enrage timer to Heroic only (see source)

(The difficulty increment in heroic for summons is that besides de HP checks, there's an additional summon every 15s-24s)

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5790

## SOURCE:

https://wowwiki-archive.fandom.com/wiki/Mekgineer_Steamrigger
&
image
&
image

## Tests Performed:

- Tested

### DIFF
--- a/src/server/scripts/Outland/CoilfangReservoir/SteamVault/boss_mekgineer_steamrigger.cpp
+++ b/src/server/scripts/Outland/CoilfangReservoir/SteamVault/boss_mekgineer_steamrigger.cpp
@@ -30,8 +30,6 @@ enum MekgineerSteamrigger
     SPELL_SAW_BLADE             = 31486,
     SPELL_ELECTRIFIED_NET       = 35107,
     SPELL_ENRAGE                = 26662,
-    SPELL_REPAIR_N              = 31532,
-    SPELL_REPAIR_H              = 37936,
 
     SPELL_SUMMON_MECHANICS_1    = 31528,
     SPELL_SUMMON_MECHANICS_2    = 31529,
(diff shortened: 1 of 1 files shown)


Gultask:
Adds are summoned on a timer in Heroic, we saw it in retail

Si1ker:
@Gultask in this timer? (15600ms - 25400ms)

Gultask:
yeah that's what I found. I based it off the timestamps between each cast of the summon spells.
I thought it was weird too, almost fixed it as well shortly before Heroics came out because I had forgotten about it

Si1ker:
> yeah that's what I found. I based it off the timestamps between each cast of the summon spells. I thought it was weird too, almost fixed it as well shortly before Heroics came out because I had forgotten about it

Judging by common sense, all timers in Heroic would be on that timer.
It seems impossible in Heroic it only has 1 summon on that timer, this makes it way easier than normal.
Sounds about right?

Si1ker:
`Gnome Repairers: A few times during the fight at 75%, 50% and 25% health, he will call upon 3-4 gnomes to heal him. These gnomes have 5k hp and deal 200-500 melee damage on hit. These are non-elite mobs and should receive all focus from DPS, to prevent boss from healing. The healing is a channeled spell which break on damage, and after it has been broken the gnome will start it again after a few seconds, so keep the DPS up.`
.
.
Heroic Mode
Steamrigger in heroic mode is almost identical to normal mode, **except that an extra gnome appears** every 7 seconds or so. So DPS need to burn it down in addition to the normal 3 spawn that comes when Steamrigger announces "Tune em' up boys!". The tank can Thunderclap to interrupt heals temporarily, but the gnomes quickly switch back to healing so it only buys 3-4 seconds at most. Druid tanks can use swipe to prevent healing until dps can pick up the gnomes. Priests should Fear gnomes that are beating on clothies when possible.

Si1ker:
When I read "heroic" and "adds" definitely something seems off with that 1 add per timer ..  so the approach proposed here have much more sense.

image

Gultask:
https://youtu.be/n4v6l2oyahQ?t=1772
Check the video, only around 3 mechanics spawn during the fight

Si1ker:
> https://youtu.be/n4v6l2oyahQ?t=1772 Check the video, only around 3 mechanics spawn during the fight

Weird, but undeniable. 🤔
Also confirmed it [in this vid as well](https://youtu.be/4yGmX_YrMck?t=1424)

I don't think there's a video for the enrage timer of 5min happening only in Heroic so we rely on wowheadComments/Guides that mention that's the behavior? changing...

balleny:
from https://wowwiki-archive.fandom.com/wiki/Mekgineer_Steamrigger

> Steamrigger in [heroic mode](https://wowwiki-archive.fandom.com/wiki/Heroic_mode) is almost identical to normal mode, except that an extra gnome appears every 7 seconds or so.

boss should summon adds every 25% AND time based adds on top.

Nyeriah:
Best if we check it on retail to be sure if we haven’t already

Si1ker:
> > https://youtu.be/n4v6l2oyahQ?t=1772 Check the video, only around 3 mechanics spawn during the fight
> 
> Weird, but undeniable. 🤔 Also confirmed it [in this vid as well](https://youtu.be/4yGmX_YrMck?t=1424)
> 
> I don't think there's a video for the enrage timer of 5min happening only in Heroic so we rely on wowheadComments/Guides that mention that's the behavior? changing...

@Gultask  ?

Gultask:
5 minutos
image

balleny:
looking into videos
https://www.youtube.com/watch?v=CS1QdwH1aKE
and similar on youtube, there is indeed only 1 add spawning every now and then. no percentage based spawns

Issue #16621, opened 2023-06-25
[Quest]: Replacement Phial


The-GhostRider:
### Current Behaviour

After such a long time of being out of the game, I have returned to playing. So, I found myself with some quests in the "Quest log," and one of them was specifically the Quest: ["Replacement Phial."](https://wowgaming.altervista.org/aowow/?quest=3375) However, I realized that it was an Alliance quest, so I wondered, "Why do I have an Alliance quest in my Quest Log if I am Horde?

![Screenshot_1](https://github.com/chromiecraft/chromiecraft/assets/1884642/6923cbfc-3eec-4b30-b77f-a0b31938859b)

### Expected Blizzlike Behaviour

The Quest should not be taken by Horde Characters

### Source

https://obvious.ofcourse

### Steps to reproduce the problem

I have no idea how I got the quest, I was probably in Uldaman in a group with some Alliance and they shared the quest with me
**(I assume)**

![image](https://github.com/chromiecraft/chromiecraft/assets/1884642/c7f3813a-3d1c-40f5-8b84-353a3a94009d)

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/fb7164be05ff507a0e25200c9e267b87186e2850

### Operating system

Win10 (64)

PR #16622, opened 2023-06-25
Fix(DB/Quest): Replacement Phial

The-GhostRider:
## Changes Proposed:
- Fix Quest: [Replacement Phial](https://wowgaming.altervista.org/aowow/?quest=3375)
- I've set the quest to be taken only by the Alliance faction, (as it should be).

## Issues Addressed:

- Closes: https://github.com/chromiecraft/chromiecraft/issues/5834
- Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/16621

## Tests Performed:

- Tested in game and working
- Tested on Windows 10 (64)

image

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1687699164334187200.sql
+++ b/data/sql/updates/pending_db_world/rev_1687699164334187200.sql
@@ -1,0 +1,2 @@
+-- Fix Quest Replacement Phial Only available to Alliance --
+UPDATE `quest_template` SET `AllowableRaces` = 1101 WHERE (`ID` = 3375);


heyitsbench:
This does *not* line up with a sniff of this quest. The `AllowableRaces` for quest `3375` according to a 3.4.0 47168 sniff is `1791`.

The-GhostRider:
> This does _not_ line up with a sniff of this quest. The for quest according to a 3.4.0 47168 sniff is .`AllowableRaces``3375``1791`

Well, then you have misinterpreted/"sniffed" wrong it....

It is not possible for the Horde to deliver the **Quest:** [Replacement Phial](https://wowgaming.altervista.org/aowow/?quest=3375)
image

to **this NPC:** [Talvash del Kissel](https://www.wowhead.com/npc=6826/talvash-del-kissel)
**in Ironforge:**
image

**For several reasons:**

- The guards would kill him, and he would never reach his destination.
- The NPC belongs to the Alliance, so the Horde would perceive him as a hostile entity and wouldn't be able to initiate the dialogue window to complete the quest.
- Logically speaking, the mere idea of the Horde infiltrating Alliance territory or vice versa to complete a quest is preposterous.

heyitsbench:
I'm not saying that it is logical for a Horde player to be able to turn in or even accept the quest, I am only pointing out that you are proposing to change the `AllowableRaces` value for this quest to something that directly contradicts a sniff.

The-GhostRider:
> I'm not saying that it is logical for a Horde player to be able to turn in or even accept the quest, I am only pointing out that you are proposing to change the `AllowableRaces` value for this quest to something that directly contradicts a sniff.

I got it, but then what are we even doing here!?
I don't understand your point.
Should we just leave the quest bugged as it is because the Sniff is saying something that makes no sense whatsoever?

Honestly...
Sometimes, certain comments, I just don't understand them at all...
Maybe it's my mistake... I don't know...

heyitsbench:
> what are we even doing here!?

I am sorry to sound condescending, but hear me out.

The goal of AC is to emulate Blizz (and fix a few blizzlike bugs here and there). One of the best ways to do this is to take data directly from their servers where possible, sniffs being one avenue to accomplish that. The `quest_template` table is meant *explicitly* to hold onto that verified data (that's what the point of any table with a `VerifiedBuild` column is for). Unlike certain things like spells (in select circumstances), there are very few situations where changing the values in that table are anything but a hackfix.

That being said, this issue is directly a result of cross-faction grouping. If I had my druthers, I would prefer some sort of check when cross-faction grouping is enabled to prevent these sorts of quests from being shared. Something I very much don't agree with is verified data being modified to avoid a situation that is the product of a custom change. Maybe the reviewers/maintainers disagree with me, but as one of the contributors who champions sniffs and the data they provide, I would prefer this issue to stay a possibility rather than modify known blizzlike data.

The-GhostRider:
> > what are we even doing here!?
> 
> I am sorry to sound condescending, but hear me out.
> 
> The goal of AC is to emulate Blizz (and fix a few blizzlike bugs here and there). One of the best ways to do this is to take data directly from their servers where possible, sniffs being one avenue to accomplish that. The `quest_template` table is meant _explicitly_ to hold onto that verified data (that's what the point of any table with a `VerifiedBuild` column is for). Unlike certain things like spells (in select circumstances), there are very few situations where changing the values in that table are anything but a hackfix.
> 
> That being said, this issue is directly a result of cross-faction grouping. If I had my druthers, I would prefer some sort of check when cross-faction grouping is enabled to prevent these sorts of quests from being shared. Something I very much don't agree with is verified data being modified to avoid a situation that is the product of a custom change. Maybe the reviewers/maintainers disagree with me, but as one of the contributors who champions sniffs and the data they provide, I would prefer this issue to stay a possibility rather than modify known blizzlike data.

**I still don't understand what you're talking about..**

image

image

image

image

Nyeriah:
There are tools to extract retail data, they call the process “sniffing”. Quest template uses sniffed data, so an alternative solution must be found to this

Kitzunu:
Wowhead is not an accurate source to rely on. Sniffed data is not wrong. You can read here what data should be sniffed https://www.azerothcore.org/wiki/sniffing-and-parsing

The-GhostRider:
> Wowhead is not an accurate source to rely on. Sniffed data is not wrong. You can read here what data should be sniffed https://www.azerothcore.org/wiki/sniffing-and-parsing

Hello Kitzunu, thanks for joining the conversation.

Well then...
Regardless of whether Wowhead is reliable or not, I start from the assumption that:

- _It is 100% certain that a Horde player cannot go to an Alliance player's home to complete a quest, and vice versa._

- _It is 100% certain that (regardless of whether the quest was marked or not as Alliance), it has always been an Alliance quest._
- _It is 100% certain that Blizzard in WoW has made mistakes/oversights that they sometimes corrected in subsequent patches of that expansion, or that they never corrected at all, and eventually even corrected in later expansions due to laziness or some other internal mismanagement they had as a company._

So, my point is:

- _If Blizzard has displayed a significant amount of incompetence, why should we also fall into that category?_
- _If we already know that this is and always will be an Alliance quest, and Blizzard has eventually recognized the error and marked it as "Alliance," where does it say that we cannot "improve" on a mistake they made earlier? (especially since they later corrected it themselves because they acknowledged the mistake).
I don't believe that the Azeroth Core project was created to be the "Twin Project" of "Trinity Core" or to be the "Twin Project" of "Blizzard's Failures." I believe that the essence/reason behind the creation of the Azeroth Core project was to "do better" and not "repeat the SAME ignorant mistakes made by others," whether those "others" are "Blizzard" or "TrinityCore" or anyone else._

**This is what I think and the principle/ideal that I believe in.**

Nyeriah:
Hey, I will try to clarify this:

1) It seems clear to me that the quest should not be available to Horde players and it doesn't seem that is being questioned here
2) I don't believe we disagree that something should be done about this and I don't think anyone has said it shouldn't be fixed

We fundamentally disagree on **how** it should be done: fields that come from sniffs won't be changed without sniffs to back them up. There should be alternative ways to address this, like perhaps a special flag.

The-GhostRider:
> Hey, I will try to clarify this:
> 
> 1. It seems clear to me that the quest should not be available to Horde players and it doesn't seem that is being questioned here
> 2. I don't believe we disagree that something should be done about this and I don't think anyone has said it shouldn't be fixed
> 
> We fundamentally disagree on **how** it should be done: fields that come from sniffs won't be changed without sniffs to back them up. There should be alternative ways to address this, like perhaps a special flag.
> 
> Thanks for your continued support and interest in collaborating with our project

I went a little back in time for you..

https://web.archive.org/web/20131114010003/https://www.wowhead.com/quest%3D3375

image

Kitzunu:
Okay sure. You have basically sent the same screenshot 5 times. That doesnt change the fact that the way you fix the issue is wrong.

Nyeriah is right, this should be fixed, but in an alternative way that alignes with the sniffed values.

Nyeriah:
Nobody is arguing this, as you’ve been told by multiple people. @heyitsbench already confirmed the values we have are correct so there needs to be another way to address this

I appreciate your interest in helping but I’ll be closing this now as we don’t seem to be going anywhere and this can’t be accepted as is.

elthehablo:
> > Okay sure. You have basically sent the same screenshot 5 times. That doesnt change the fact that the way you fix the issue is wrong.
> > 
> > Nyeriah is right, this should be fixed, but in an alternative way that alignes with the sniffed values.
> 
> No Kitzunu, it's not the same Screenshot.
> 
> The Screenshots above referred to the Wowhead (now 2023).
> While this last Screenshot refers to Wowhead of 2013 when there was still the Wowhead of WOTLK (the first WOTLK).
> So in this screenshot you can see the real DATA from 10 years ago..
> is the purpose of this site: https://web.archive.org (To show past events of the Web that are no longer visible)..

Bruv you gotta understand wowhead does not give accurate enough information to understand exactly how these quests work on retail servers. Not in 2009, not in 2013 and not in 2023. That is why people gather sniffs

heyitsbench:
> heyitsbench already confirmed the values we have are correct

For the record, unless something is wrong with my DB, AC has an `AllowableRaces` value of 0 for quest 3375, which is still inaccurate.

The-GhostRider:
> > heyitsbench already confirmed the values we have are correct
> 
> For the record, unless something is wrong with my DB, AC has an `AllowableRaces` value of 0 for quest 3375, which is still inaccurate.

And what would be the "Accurate" number according to the Sniffs you made?

The-GhostRider:
> > And what would be the "Accurate" number according to the Sniffs you made?
> 
> ```
> ServerToClient: SMSG_QUERY_QUEST_INFO_RESPONSE (0x2A96) Length: 1232
> Quest ID: 3375
> AllowableRaces: 1791
> ```

Interesting...

image

elthehablo:
Once more, as debated above. The system that makes sure that horde players won't receive this quest is handled somewhere else in retail. But, apparently these flags are set like these on retail so this is also how we should do this, and then find a way sometime to implement the same features blizz uses to not allow the quest to end up in Horde quest logs

Gultask:
That quest still exists in Retail, right? 10.1?
If someone pings me during the weekend I can try looking at the error message and such

elthehablo:
Once more, nobody is trying to say your work is shit or anything like that. People just want it to be fixed the way it works on retail. AC strives to be like retail wotlk

elthehablo:
> That quest still exists in Retail, right? 10.1?
> If someone pings me during the weekend I can try looking at the error message and such

Late ping

elthehablo:
@Gultask once more just to make sure happyge

Gultask:
image

Gultask:
I thought it was TBC

image

PR #16635, opened 2023-06-27
fix: correct the parameter order

zskamljic:
## Changes Proposed:
-  Update parameter order to match naming in docs

## Issues Addressed:
- No issue created, noticed while using

## SOURCE:
The function is specified as follows:
```c++
OnUpdateGatheringSkill(Player* /*player*/, uint32 /*skill_id*/, uint32 /*current*/, uint32 /*gray*/, uint32 /*green*/, uint32 /*yellow*/, uint32& /*gain*/)
```
However the use site used:
```c++
script->OnUpdateGatheringSkill(player, skillId, gray, green, yellow, currentLevel, gain);
```
Instead of:
```c++
script->OnUpdateGatheringSkill(player, skillId, currentLevel, gray, green, yellow, gain);
```

## Tests Performed:
Created a module as following:
```c++
class CraftingGainScript : public PlayerScript {
public:
    CraftingGainScript() : PlayerScript("CraftingGainScript_PlayerScript") {
    }

    void
    OnUpdateGatheringSkill(Player *player, uint32 skill, uint32 skill_level, uint32 gray, uint32 green, uint32 yellow,
                           uint32 &gain) override {
        if (skill_level <= yellow) {
            gain *= 2;
        }
    }

    void OnUpdateCraftingSkill(Player *player, const SkillLineAbilityEntry *entry, uint32 skill_level,
                               uint32 &gain) override {
        if (skill_level <= entry->TrivialSkillLineRankLow) {
            gain *= 2;
        }
    }
};
```
Expected result: crafting orange difficulty would grant 2 skillups instead of one, but 1 for other levels (if triggered)

## How to Test the Changes:
See script above. Try gathering plants with orange difficulty, observe that 2 skill levels are awarded.

### DIFF
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -1009,7 +1009,7 @@ void ScriptMgr::OnGetMaxSkillValue(Player* player, uint32 skill, int32& result, 
 void ScriptMgr::OnUpdateGatheringSkill(Player *player, uint32 skillId, uint32 currentLevel, uint32 gray, uint32 green, uint32 yellow, uint32 &gain) {
     ExecuteScript<PlayerScript>([&](PlayerScript* script)
     {
-        script->OnUpdateGatheringSkill(player, skillId, gray, green, yellow, currentLevel, gain);
+        script->OnUpdateGatheringSkill(player, skillId, currentLevel, gray, green, yellow, gain);
     });
 }
 


heyitsbench:
May want to update the docs rather than the hook itself.

zskamljic:
I've added the hook about a week ago and the docs+function parameters were in their intended places, but the parameter order was not. I thought it would be preferable to have related parameters grouped at the start, especially since the hook for crafting skills works the same.

Kitzunu (approved):

Issue #3244, opened 2020-08-02
Escape from Durnholde - Thrall should stop resurrecting


BarbzYHOOL:
##### CURRENT BEHAVIOUR:

During this quest, you must escort Thrall, but if he dies he is resurrected. But normally, after a while he shouldn't get resurrected.

When i'm testing, he is dying like 10 times and still resurrected :P

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Read the comments here https://www.wowhead.com/quest=29599/escape-from-durnholde#comments he should stop being rezzed after an unknown amount of times. We could arbitrarily decide it's 3 or scour the web for info like here https://wowwiki.fandom.com/wiki/Escape_from_Durnholde_Keep

##### STEPS TO REPRODUCE THE PROBLEM:

1. Do this quest in old hillsbrad dunrholde https://wow.gamepedia.com/Taretha's_Diversion or if you done it before, take the explosives at the start of the dungeon and then do like this guy https://www.youtube.com/watch?v=YsdxCxqlZ_E&feature=youtu.be
2. When you release Thrall, he will attack mobs and if he dies he will be resurrected everytime

##### AC HASH/COMMIT:

0572166a

pangolp:
Does anyone know if this is still buggy or if it has been fixed?

Johaine:
There is a limit in the script and it has been there for 7 years if I see that right. But from a first look it might be bugged. I'll test it and fix it if its broken.

Currently the limit is at 19 allowed deaths. After the 20th death Thrall shouldn't respawn.

pangolp:
> There is a limit in the script and it has been there for 7 years if I see that right. But from a first look it might be bugged. I'll test it and fix it if its broken.
> 
> Currently the limit is at 19 allowed deaths. After the 20th death Thrall shouldn't respawn.

I didn't know there was a limit to the number of attempts. That fact was unknown to me.

PR #16636, opened 2023-06-27
fix(Scripts/EscapeFromDurnholde): Fix escort mission getting stuck

Johaine:
Fix escort event possibly getting stuck by preventing DATA_ESCORT_PROGRESS from being decreased

When all players leave the instance (e.g. corpse walk after wipe) and the first player re-entered the old hillsbrad, CleanupInstance() is called

CleanupInstance() schedules EVENT_FINAL_BARRELS_FLAME which unconditionally sets DATA_ESCORT_PROGRESS, possibly rolling back the instance progress and removing the gossip option from Thrall

Trivia: The way it was previously handled, you could unstuck the instance by leaving _twice_ and re-entering. After the second CleanupInstance() Lieutenant Drake would respawn and could be killed again (exploit alert!).
This would set the escort progress to the correct value again and Thrall's gossip option would reappear. This explains the second comment on the Chromiecraft issue

## Changes Proposed:
-  Enforce only increasing DATA_ESCORT_PROGRESS, preventing any kind of rollback on the instance progress

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5803

## Tests Performed:

- Multiple Runs in the instance, killing Thrall at different point of the escort, leaving the instance at different points in time etc

## How to Test the Changes:

1. Go to Old Hillsbrad Foothills
2. Place 5 bombs in barrels and kill first boss
3. Confirm that Thrall can be talked to and has a gossip option to start the escort
4. Leave the instance, restart the server or something similar
5. Talk to respawned Thrall again and confirm that the gossip option is still available and escort can be restarted / continued

- A few improvements to the instance are planned but I want to get the fix out. Currently planned:
- Fix Captain Skarloc's horse attacking players
- Fix the armorer RP event
- Check (and possibly fix) the limit for failing the escort (issue #3244)
- Fix Thrall being level 67 on heroic, it is level 71 on official servers

### DIFF
--- a/src/server/scripts/Kalimdor/CavernsOfTime/EscapeFromDurnholdeKeep/instance_old_hillsbrad.cpp
+++ b/src/server/scripts/Kalimdor/CavernsOfTime/EscapeFromDurnholdeKeep/instance_old_hillsbrad.cpp
@@ -66,7 +66,7 @@ public:
 
         void OnPlayerEnter(Player* player) override
         {
-            if (instance->GetPlayersCountExceptGMs() == 1)
+            if (instance->GetPlayersCountExceptGMs() <= 1)
                 CleanupInstance();
 
             EnsureGridLoaded();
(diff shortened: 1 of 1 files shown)


Maria-sequel:
Can confirm it works, even during escort he will just be back at the start.

PR #16657, opened 2023-07-01
fix(core/spell): Stop `Blink` from the falling player to move upward by 1y and stop moving backward when facing a wall

Synful-Syn:
## Changes Proposed:
- Using Blink while falling at a high height will no longer warp the player by 1y upward.
- Added a logic that prevents the player from moving backward by using `Blink`.

## Issues Addressed:

- Closes nothing, I've discovered it while failing to correct something else in Blink.

## Tests Performed:

- 
- 

## How to Test the Changes:

Learn Blink: `.learn 1953`

Use Blink without having to wait for the mana or the cooldowns.
`.cheat cooldown`
`.mod mana 239432432`

Make a `.gps` macro so you may see your Z coordinate while testing the falling from a high altitude.

-

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -1587,8 +1587,8 @@ void Spell::SelectImplicitCasterDestTargets(SpellEffIndex effIndex, SpellImplici
                             //LOG_ERROR("spells", "(collision) destZ rewrited in prevZ");
 
                             // Don't make the player move backward from the xy adjustments by collisions.
-                            if ((DELTA_X > 0 && startx > destx) || (DELTA_X < 0 && startx < destx)
-                            ||  (DELTA_Y > 0 && starty > desty) || (DELTA_Y < 0 && starty < desty))
+                            if ((DELTA_X > 0 && startx > destx) || (DELTA_X < 0 && startx < destx) ||
+                                (DELTA_Y > 0 && starty > desty) || (DELTA_Y < 0 && starty < desty))
                             {
                                 destx = startx;
                                 desty = starty;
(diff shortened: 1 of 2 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -1586,6 +1586,15 @@ void Spell::SelectImplicitCasterDestTargets(SpellEffIndex effIndex, SpellImplici
                                 destz = prevZ;
                             //LOG_ERROR("spells", "(collision) destZ rewrited in prevZ");
 
+                            // Don't make the player move backward from the xy adjustments by collisions.
+                            if ((DELTA_X > 0 && startx > destx) || (DELTA_X < 0 && startx < destx)
+                            ||  (DELTA_Y > 0 && starty > desty) || (DELTA_Y < 0 && starty < desty))
```
Kitzunu, line 1591:
```suggestion
                            if ((DELTA_X > 0 && startx > destx) || (DELTA_X < 0 && startx < destx)
                                ||  (DELTA_Y > 0 && starty > desty) || (DELTA_Y < 0 && starty < desty))
```

Synful-Syn: I made the spacing this way so they would be perfectly alligned for readability purposes.
image

With the change you are offering, there would be a double space right here.
image

I do prefer not changing it but if we must, I'd remove the double space or making it a single hard-to-read line.
image

What do you prefer? Making it a single line or removing the double space?

Kitzunu: Remove the double space, I had missed that initially

Kitzunu: What you could do is moeve the || to the previous line for readability

Synful-Syn: I did not think of that, I love it. :D

Nefertumm (approved):

GastonCK:
Tested
https://www.youtube.com/watch?v=tukMsHaJV50

PR #16662, opened 2023-07-02
feat(CORE):Add getter for completed achievement map in AchievementManager

Tralenor:
## Changes Proposed:
-  Add a Function on Achievement Manager to get read-only access to the map which contains all completed achievements. This allows for easier access to the Information of a players completed achievements. 

## Tests Performed:

- Does it build without errors? yes!
- Did you test in-game?  yes, with a custom module where I listened to the OnAchiCompleted Hook an then accessed the map over:  auto completedAchievements = player->GetAchievementMgr()->GetCompletedAchievements();
- On which OS did you test?  Ubuntu and Debian
- Describe any other tests performed --> none.

## How to Test the Changes:

No functionality change out of the box, just a "Getter"-Fucntion to allow modules easier access to Information about a Players achieved Achievements - thus there is nothing no test out of the box. 
An example on how this could be used:     
```c++
void OnAchiComplete(Player *player, AchievementEntry const *achievement) override {
        auto completedAchievements = player->GetAchievementMgr()->GetCompletedAchievements();

        auto achievementsPoints = std::accumulate(completedAchievements.cbegin(), completedAchievements.cend(), 0,
                                                  [](int sum, auto element) {
                                                      return sum + sAchievementStore.LookupEntry(element.first)->points;
                                                  });

        std::stringstream message{};
        message << "Achievement gained, your new total is: " << achievementsPoints;

        player->GetSession()->SendAreaTriggerMessage("%s", message.str().c_str());
    }
```

### DIFF
--- a/src/server/game/Achievements/AchievementMgr.cpp
+++ b/src/server/game/Achievements/AchievementMgr.cpp
@@ -525,6 +525,7 @@ void AchievementMgr::ResetAchievementCriteria(AchievementCriteriaCondition condi
     for (AchievementCriteriaEntryList::const_iterator i = achievementCriteriaList->begin(); i != achievementCriteriaList->end(); ++i)
     {
         AchievementCriteriaEntry const* achievementCriteria = (*i);
+
         AchievementEntry const* achievement = sAchievementStore.LookupEntry(achievementCriteria->referredAchievement);
         if (!achievement)
             continue;
@@ -2392,6 +2393,10 @@ bool AchievementMgr::CanUpdateCriteria(AchievementCriteriaEntry const* criteria,
         return false;
 
     return true;
+}
+
+const CompletedAchievementMap& AchievementMgr::getCompletedAchievementMap() {
+    return _completedAchievements;
 }
 
 AchievementGlobalMgr* AchievementGlobalMgr::instance()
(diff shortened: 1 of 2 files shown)


src/server/game/Achievements/AchievementMgr.cpp
```diff
@@ -2394,6 +2394,10 @@ bool AchievementMgr::CanUpdateCriteria(AchievementCriteriaEntry const* criteria,
     return true;
 }
 
+const CompletedAchievementMap& AchievementMgr::GetCompletedAchievements() {
```
Kitzunu, line 2397:
```suggestion
CompletedAchievementMap const& AchievementMgr::GetCompletedAchievements()
{
```

You can read more about the C++ standards we use here: https://www.azerothcore.org/wiki/cpp-code-standards

Kitzunu (changes requested):

src/server/game/Achievements/AchievementMgr.h
```diff
@@ -300,6 +300,7 @@ class AchievementMgr
 
     void RemoveCriteriaProgress(AchievementCriteriaEntry const* entry);
     CriteriaProgress* GetCriteriaProgress(AchievementCriteriaEntry const* entry);
+    const CompletedAchievementMap& GetCompletedAchievements();
```
Kitzunu, line 303: and here :D

Tralenor: ouch - gonna fix it, 1sec

src/server/game/Achievements/AchievementMgr.cpp
Tralenor: thank you, is there somekind of linter I could use to enforce them? (or at least parts of it?)

src/server/game/Achievements/AchievementMgr.h
Tralenor: should be fixed now.

Kitzunu (approved):

Issue #9550, opened 2021-12-07
Horde's Soothing Turtle Bisque quest has incorrect POI


Shineslippers:
### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [X] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour

Original report: https://github.com/chromiecraft/chromiecraft/issues/2490
The quest "Soothing Turtle Bisque" has wrong POIs on the map.

![WoWScrnShot_120621_145024](https://user-images.githubusercontent.com/62797536/144857703-bf1d495b-802c-409a-ada0-329c57a77505.jpg)

Tester:

The issue is confirmed. Horde's [Soothing Turtle Bisque](https://wowgaming.altervista.org/aowow/?quest=7321) quest has incorrect POI

![WoWScrnShot_120721_132210](https://user-images.githubusercontent.com/85256266/145011607-dbd5ba12-6555-483c-9562-31a2f2a9d03b.jpg)

### Expected Blizzlike Behaviour

Correct POIs could be sniffed from Allince's [Soothing Turtle Bisque](https://wowgaming.altervista.org/aowow/?quest=555)

### Steps to reproduce the problem

.q a7321
show quest map

### Extra Notes

X

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/fc4b5a93a0318610e14af3a3bb50a1719f5ee765

### Operating system

Windows 10

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
[...]

PR #16710, opened 2023-07-09
fix(DB/questpoi): 7321 - Soothing Turtle Bisque - correct POI for Horde

wowquester:
This fixes POI area for 7321 which shares it with the 555 (alliance version) which has the correct ones. Please beware of my MySQL skill, although it fixes the problem, I encourage you to review this code and if necessarry, make it more compact. The basic idea is just to duplicate entries from 555 to 7321 - this results in correct behavior of the quest poi on the world map.
PS:(I am VERY new to the world of fixing things... please be nice and if anything, try to explain it a as would to a small child.)

## Changes Proposed:
-  Corrected POI behavior for Horde side (Q7321)
- 

## Issues Addressed:

- [Closes](https://github.com/azerothcore/azerothcore-wotlk/issues/9550)

## SOURCE:

It's the same cooking quest - shares location, task and creatures, Aliance verions works as expected, see https://github.com/azerothcore/azerothcore-wotlk/issues/9550

## Tests Performed:

Horde
WoWScrnShot_070923_104341
Ali
WoWScrnShot_070923_104346
Horde only
WoWScrnShot_070923_104358
Previous behavior (the misplaced is the original state)
https://user-images.githubusercontent.com/85256266/145011607-dbd5ba12-6555-483c-9562-31a2f2a9d03b.jpg
- 

## How to Test the Changes:

1. .go tax 13
2. .q add 7321 (Horde)
3. .q add 555 (Alliance)
4. Compare them

- [ ]
- [ ]


Kitzunu:
This needs to be sniffed

elthehablo:
Bench already explained to you that this is a "hack-fix", and that goes against the AC principle especially for low priority issues like this. So someone needs to sniff the POIs and then make this PR.

wowquester:
> Bench already explained to you that this is a "hack-fix", and that goes against the AC principle especially for low priority issues like this. So someone needs to sniff the POIs and then make this PR.

Ah, sorry about that. I thought that hack-fix means it's undoable...

elthehablo:
> > Bench already explained to you that this is a "hack-fix", and that goes against the AC principle especially for low priority issues like this. So someone needs to sniff the POIs and then make this PR.
> 
> Ah, sorry about that. I thought that hack-fix means it's undoable... 

Hack-fix means that the fix does relieve the problem somewhat, but it is not the way it should work. The way it should work usually being the way it is implemented on classic servers. The end-goal of AC is to be very close to that. Thanks for the contribution though and keep it up :)

Gultask:
> This needs to be sniffed

Impossible for now. I'm in the mind now that it's ok to guess/approximate as long as there's an issue clearly stating that this needs to be sniffed later. I'll wait for some more opinions, though.

Kitzunu (approved):

Issue #16674, opened 2023-07-05
Worldserver crash when entering northern Zul'Drak DEBUG only


bewarellamas:
### Current Behaviour

Worldserver is crashing when entering the northern area of Zul'Drak. Tested entering from the south and crashes when going into the Altar of Mam'toth area. Also tested by teleporting into Gundrak instance and exiting via the portal and it crashes there as well. Happens on multiple accounts with clients on two different devices. This is on a freshly compiled client using source pulled today (July 5th) with no mods installed. It has been happening before on previous versions, but I was finally able to get a meaningful crash log (gist link below).

https://gist.github.com/bewarellamas/bed6d06cec4ba14756bd6a4a1f0189e3

### Expected Blizzlike Behaviour

It shouldn't crash.

### Steps to reproduce the problem

Enter northern Gun'drak via Altar of Mam'toth area, or by exiting Gundrak.

### Extra Notes

I thought it could be a map data file issue so I replaced my data files using the link from the AC discord to make sure they were good. AC Data V16

### AC rev. hash/commit

893a9beeacc4

### Operating system

Windows 10 x64

### Custom changes or Modules

None. Fresh compile from source on July 5th.

elthehablo:
Looks like some smart scripts over there are the issue

elthehablo:
Just gotta check the DB for some weird stuff somewhere in Zul'drak. Some SAI entry is probably not filled in correctly and throwing this error

![Screenshot_20230706_180557_GitHub.jpg](https://github.com/azerothcore/azerothcore-wotlk/assets/83884799/1b893b6c-ca37-4301-967c-89e16e09fc3a)

![Screenshot_20230706_180846_GitHub.jpg](https://github.com/azerothcore/azerothcore-wotlk/assets/83884799/81030a0a-0182-4e64-b126-fb9f0e1a22a9)

![Screenshot_20230706_180820_GitHub.jpg](https://github.com/azerothcore/azerothcore-wotlk/assets/83884799/44436b6b-be7b-46f9-87f9-5d7fb316e5b5)

elthehablo:
Stack overflow so might be trying to process an action that no longer exists or something

bewarellamas:
Random note, I am doing this on a test server with a fresh database so that point is known with testing.

[...]

PR #16739, opened 2023-07-11
fix(Core/SmartAI) : use explicit stack DS for ProcessAction instead of recursion

Dantsz:
## Changes Proposed:
-  Add a deque to SmartScript object to store data about linked events instead of calling ```SmartScript::ProcessEvent``` recursively
-  ```SmartScript::ProcessEventsFor``` uses the deque to process events

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16674

## SOURCE:

Before:

https://github.com/azerothcore/azerothcore-wotlk/assets/19265585/f7484f17-a606-44f3-a6fa-ad77d5322ace

After:

No longer crashes

## Tests Performed:

- Went to Gundrak and the server did not crash
- Killed Tormar Frostgut in garm and game did not crash
- Went to Shattered Halls to see if mobs execute the scripts correctly and it was ok

## How to Test the Changes:

Run in Debug, MSVC (I have 19.37.32705.0 installed)

Scenario 1:

1. Fly to Gundrak 
2. Should no longer crash
3.

Scenario 2(video):

1. addItem 41431
2. go to end of garm in Storm Peaks
3. place dynamite
4. Kill mob
5. Script should run without crashing

- [Might affect performance]
- [ I cannot test every script to check if anything is affected]


Kitzunu:
@Winfidonarleyan @Nefertumm Can one of you review this please? :)

sudlud:
@walkline you got any opinion on this?

src/server/game/AI/SmartScripts/SmartScript.h
```diff
@@ -294,23 +296,25 @@ class SmartScript
...
                 {
-                    return (*i);
+                    return std::ref(*i);
                 }
             }
         }
-        SmartScriptHolder s;
-        return s;
+        return std::nullopt;
     }
```
walkline, line 313: Lol, I was wondering why it was implemented that way. 
I thought maybe it was because the object’s lifetime is short and we need to work with copies of the objects. 
But then I noticed that we're not clearing the mEvents vector 😅
So, good catch!

walkline (commented):
Have a small nitpick, but overall, it looks good to me!

sudlud:
@Dantsz can you please address the latest review :)

Dantsz:
> @Dantsz can you please address the latest review :)

Will try to find time to do so

sudlud:
Thanks for the update!

So how can we go about testing this PR - to my understanding this basically effects all SmartAI handling? Or just SmartAI with linked events?

Can we expect a performance gain here or is this more of a refactoring into a "cleaner" handling?

Dantsz:
> So how can we go about testing this PR - to my understanding this basically effects all SmartAI handling? 

It does affect all scripts, I don't think script length would be relevant outside of the long scripts which crash, so testing it on smaller scripts should be fine.

> Can we expect a performance gain here or is this more of a refactoring into a "cleaner" handling?

Performance impact should be small and probably is platform dependent, the memory footprint could be larger, it's not a cleaner handling, as the links should be processed in the same order and with the same parameters, it's just moving memory usage from the stack (recursion) to the heap so we don't have a stack overflow when the links get big.

sudlud (approved):

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -145,7 +145,16 @@ void SmartScript::ProcessEventsFor(SMART_EVENT e, Unit* unit, uint32 var0, uint3
             ConditionSourceInfo info = ConditionSourceInfo(unit, GetBaseObject(), me ? me->GetVictim() : nullptr);
 
             if (sConditionMgr->IsObjectMeetToConditions(info, conds))
-                ProcessEvent(*i, unit, var0, var1, bvar, spell, gob);
+            {
+                ASSERT(executionStack.empty());
```
Helias, line 149: I am always a bit unsure about `ASSERTs` instead of a LOG_ERROR(), because these make crashes the server

sudlud:
Tested.
I checked a few creatures with SmartAI with linked events which worked as expected. Also couldn't reproduce the crash in the linked issue anymore.

@Dantsz can you please address the latest review by @Helias, @walkline can you also please take look at this again

src/server/game/AI/SmartScripts/SmartScript.cpp
Dantsz: In case `executionStack.empty()` is not true it means something messed with the stack and the behavior is going to be different than the original. 

This shouldn't happen and if somehow is it's a flaw that should be fixed. I believe logging it would suggest the intention that it's expected and will be handled, which is not the case.

Issue #15867, opened 2023-04-07
[Gruul's Lair] High King Maulgar - Phase 2: Intimidating Shout


elthehablo:
originally reported by balleny at: https://github.com/chromiecraft/chromiecraft/issues/4873
also reported by NinesCC at: https://github.com/chromiecraft/chromiecraft/issues/5347

### What client do you play on?

enUS

### Faction

Both

### Content Phase:

70

### Current Behaviour

Maulgar does not cast Intimidating Shout once he is lower than 50% health (Phase 2)
https://youtu.be/7EYlPgMsrVI?t=7

### Expected Blizzlike Behaviour

After the King gets down to 50% and enters Phase 2, he should be casting Intimidating Shout every now and then, fearing the MT and melee and making it necessary to have a OT to pick it up when this happens.

### Source

_**Phase 2**

At 50% High King Maulgar enters Phase 2. He becomes big and red and drops his hammer reducing his melee damage by a good amount so that the OTs can survive a few seconds without heal after a taunt. In this phase he will charge random party members. **He will also cast Intimidating Shout on the main tank which will incapacitate him for 4 seconds and fear everyone around him.** To deal with intimidating shout, have a second tank taunt Maulgar after all the other adds are down (the MT should immediately taunt him back) and then run out of range of the fear and wait._
https://wowpedia.fandom.com/wiki/High_King_Maulgar

Original TBC 2007 footage
https://youtu.be/ZcgMzrI70FE?t=249

### Steps to reproduce the problem

1. .teleport gruul
2. kill the King Maulgar adds
3. bring the Maulgar to 50% so he drops the hammer and turn red
4. Wait for the King to cast Intimidating Shout

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/37f0b7095f949293c13f1d913f8917407ffe3b1b

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-BreakingNewsOverride](https://github.com/AnchyDev/BreakingNewsOverride)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[...]

PR #16743, opened 2023-07-12
fix(Scripts/GruulLair): Fix Maulgar not casting intimidating shout

Nyeriah:
## Changes Proposed:
-  fix target type as its not an aoe spell
-  

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5347
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15867

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Outland/GruulsLair/boss_high_king_maulgar.cpp
+++ b/src/server/scripts/Outland/GruulsLair/boss_high_king_maulgar.cpp
@@ -85,7 +85,7 @@ struct boss_high_king_maulgar : public BossAI
                 context.Repeat(35s);
             }).Schedule(0ms, [this](TaskContext context)
             {
-                DoCastSelf(SPELL_ROAR);
+                DoCastVictim(SPELL_ROAR);
                 context.Repeat(20600ms, 29100ms);
             });
         });


Nefertumm (approved):

balleny:
assuming this fix is on the PTR, it does not work.
looking at the code, it seems that the issue is more related to the nesting of the ability, which prevents it from triggering at all.
(assuming that it is supposed to trigger at 50%?!)

Nyeriah:
> assuming this fix is on the PTR, it does not work.

It's not.

elthehablo:
> assuming this fix is on the PTR, it does not work.
> looking at the code, it seems that the issue is more related to the nesting of the ability, which prevents it from triggering at all.
> (assuming that it is supposed to trigger at 50%?!)

ScheduleHealthCheck event and everything that is inside the event that you see after the `[&]` fires at the health values shown in the first parameter. It works.

elthehablo:
> > assuming this fix is on the PTR, it does not work.
> > looking at the code, it seems that the issue is more related to the nesting of the ability, which prevents it from triggering at all.
> > (assuming that it is supposed to trigger at 50%?!)
> 
> ScheduleHealthCheckEvent and everything that is inside the event that you see after the `[&]` fires at the health values shown in the first parameter. It works.

PR #16750, opened 2023-07-13
fix(Core/AI): Implement OnSpellCast, OnSpellFailed, OnSpellStart hooks from TC

Gultask:
Cherry-Pick: https://github.com/TrinityCore/TrinityCore/commit/050f55b62fded4cc73b76e42d56d98ea950cc26a

## Changes Proposed:
-  
-  

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Nyeriah (approved):

Nyeriah:
But why ;(

Gultask:
Crashed and never did what I wanted it to :(
I'm thinking of any uses for it as well and I couldn't think of any. Best one would be OnSpellFailed since some spells are re-cast on fail (Cleave re-casts on a miss). And if I don't use it at all it'll just be bloat, methinks

Issue #16458, opened 2023-06-04
Quests "Big Bone Worm" and "Fumping", spawns disappear in 1 minute


Stump876:
Original CC report https://github.com/chromiecraft/chromiecraft/issues/5514 confirmed on my local

Related to https://github.com/azerothcore/azerothcore-wotlk/issues/13144

### What client do you play on?

enUS

### Faction

Both

### Content Phase:

65-69

### Current Behaviour

Anything summoned by the Fumper- Big worms, small worms, sand gnomes- will despawn in one minute after summon, dead or alive, looted or not. This is more noticable during the big bone worm quest because it takes longer to kill.
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/68164076/979cabf7-9dc2-4b26-a963-6bfaf4eaf737)
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/68164076/9368ab69-6622-45da-b28b-6ef7210408e1)

### Expected Blizzlike Behaviour

It should behave like any other mob during combat and not despawn unless it goes out of combat and resets (player dies or runs out of range)

### Source

https://youtu.be/L6PdXZNUjpI?t=92 notice the time battle lasts over 1 minute and mob does not despawn

### Steps to reproduce the problem

1. `.go c 77628`
2. `.quest add 10930`
3. Kill a kodo use the fumper
4. Begin timer with whatever spawns
5. Auto attack for 1 minute
6. watch it vanish
7. `.go c 1977362`
8. `.quest add 10929`
9. Use fumper
10. Begin timer with whatever spawns
11. Auto attack for 1 minute
12. Watch it vanish.

### Extra Notes

SIde note: As many times as I tested this, I never once spawned sand gnomes during the big bone worm quest. Maybe it was just me?

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/ce6ca597b54e64f328cacddae5f003cdc41a343d

### Operating system

Windows 11

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-BreakingNewsOverride](https://github.com/AnchyDev/BreakingNewsOverride)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
[...]

PR #16759, opened 2023-07-15
fix(DB/SmartScript): Fix Sand Sifter and Hai'shulud despawning after 1 minute

Samsequel:
## Changes Proposed:
-  checks ooc time instead

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16458
- closes https://github.com/chromiecraft/chromiecraft/issues/5514

## Tests Performed:

- tested

## How to Test the Changes:

1. see linked issues
 
### DIFF
--- a/data/sql/updates/pending_db_world/fumper.sql
+++ b/data/sql/updates/pending_db_world/fumper.sql
@@ -1,6 +1,4 @@
 --
 UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` IN (22039, 22482); 
 DELETE FROM `smart_scripts` WHERE `entryorguid` IN (22038, 22482) AND (`source_type` = 0) AND (`id` IN (12));
-INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/fumper.sql
```diff
@@ -0,0 +1,6 @@
+--
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` IN (22039, 22482); 
+DELETE FROM `smart_scripts` WHERE (`entryorguid` IN (22038, 22482) AND (`source_type` = 0) AND (`id` IN (12));
```
Gultask, line 3: extra parentheses
```suggestion
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (22038, 22482) AND (`source_type` = 0) AND (`id` IN (12));
```

Gultask:
Hmm these are spells they're summoned from.
w/ DurationIndex 4 which is 2mins

If you leave them be won't they despawn by themselves? Without any SAI rows, that is

Samsequel:
> Hmm these are spells they're summoned from. w/ DurationIndex 4 which is 2mins
> 
> If you leave them be won't they despawn by themselves? Without any SAI rows, that is

you are right, they despawn after 2 minutes now

Gultask:
In Combat, too?

Samsequel:
> In Combat, too?

yep

Gultask:
There isn't a way unless we hack it worse than it is now

Samsequel:
> There isn't a way unless we hack it worse than it is now
> 
> Imo keep it 2mins

so remove the whole smartaction instead?

Gultask:
ye

data/sql/updates/pending_db_world/fumper.sql
```diff
@@ -0,0 +1,4 @@
+--
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` IN (22039, 22482); 
+DELETE FROM `smart_scripts` WHERE `entryorguid` IN (22038, 22482) AND (`source_type` = 0) AND (`id` IN (12));
+
```
Gultask, line 4:
```suggestion
--
DELETE FROM `smart_scripts` WHERE `entryorguid` IN (22038, 22482) AND (`source_type` = 0) AND (`id` IN (12));
```

Gultask (approved):

PR #16775, opened 2023-07-17
fix(DB/Spell): Darkmoon Card: Heroism should only trigger on melee autoattacks

Samsequel:
## Changes Proposed:
-  set flag

## Issues Addressed:

- Closes no issues

## SOURCE:

https://wowwiki-archive.fandom.com/wiki/Darkmoon_Card:_Heroism
## Tests Performed:

- tested

### DIFF
--- a/data/sql/updates/pending_db_world/heroism.sql
+++ b/data/sql/updates/pending_db_world/heroism.sql
@@ -1,1 +1,1 @@
-update `spell_proc_event`set`procFlags` = '4' where `entry` = 23689;
+UPDATE `spell_proc_event`SET`procFlags` = 4 WHERE `entry` = 23689;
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/heroism.sql
```diff
@@ -0,0 +1 @@
+UPDATE `spell_proc_event`SET`procFlags` = 4 WHERE `entry` = 23689;
```
Nefertumm, line 1:
```suggestion
UPDATE `spell_proc_event` SET `procFlags` = 4 WHERE `entry` = 23689;
```

Samsequel: oooooooooooooops

Issue #11026, opened 2022-03-15
 Darkmoon Card: Blue Dragon healing spellcast behavior 


Contrubius:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [X] 60
### Current Behaviour
![Blue Dragon](https://user-images.githubusercontent.com/100415626/158389705-4a3ce536-de93-4e2b-9054-416a65aafed2.png)

Darkmoon card blue dragon currently has a chance to proc on Glyph of Holy Light aoe heals, as well as Healing Over TIme spells and other Glyph heals, which is not Blizzlike, I go into more detail in TRIAGER NOTES. I suspect this also occurs on multi-target heals (I.e. Wild Growth/Prayer of Healing/Circle of Healing giving 5 chances to proc).

### Expected Blizzlike Behaviour

Darkmoon Card: Blue Dragon should only proc on the original spell cast and not on the effects which should not be considered as spellcasts.

### Source
https://www.wowhead.com/item=46051/meteorite-crystal Original WOTLK comments.

### Steps to reproduce the problem

Cast 100 heals as Holy Paladin with Holy Light + HL Glyph on 5 targets and see trinket proc way too often. Same for healing over time spells which should be 1 spellcast only.
### Extra Notes

Bug found by Hyperbearr.

### Triager notes:
I would like to add that I do think that bug reporter is wrong in this case, Beacon + target heal is supposed to each give a chance of 2% proc. 

![Meteorite Crystal](https://user-images.githubusercontent.com/100415626/158389769-76adaeb3-7141-47a9-9852-51f1727b7432.png)

I back up my argument with trinket Meteorite Crystal, which works on the same principle, or is also "Triggered by a spellcast", except that on each spellcast healer or caster get a buff for each spell, stacking up to 20, 1 for each spellcast. What it would do for Beaconed + healed target is double the buff stack amounts from +1 to +2, counting Beaconed target on each heal.  That behavior indeed got changed but with the first patch of Cataclysm, making this behavior Blizzlike throughout WOTLK. 

![meteorite](https://user-images.githubusercontent.com/100415626/158386522-86d02c76-0870-4140-a4b1-bc811cafaf23.jpg)

Upon testing from the comments from my test what is working as intended:

[...]

PR #16779, opened 2023-07-17
fix(DB/Spell): Darkmoon Card: Blue Dragon should only trigger on spellcast

Samsequel:
## Changes Proposed:
-  only proc on spellcast

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11026
- Closes https://github.com/chromiecraft/chromiecraft/issues/3004

## SOURCE:

Neifion testing (see issue attached)

## Tests Performed:

- tested but I would like someone else to check too

## How to Test the Changes:

1. equip blue dragon
2. cast spells and see if no difference between aoe heal and st heal (like holy nova)
3.

### DIFF
--- a/data/sql/updates/pending_db_world/darkmoonbluedragon.sql
+++ b/data/sql/updates/pending_db_world/darkmoonbluedragon.sql
@@ -1,1 +1,2 @@
+--
 update `spell_proc_event` set `procPhase` = '1' where `entry` = 23688;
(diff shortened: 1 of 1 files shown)


neifion-00000000:
have sniffed data on this as proof, but TLDR is it's well beyond the margin of error for it to proc every time, and I also got lucky and proved there's no ICD in the same go so this PR should presumably make this correct

Nefertumm (approved):

PR #16806, opened 2023-07-22
feat(Core/Config): add "include" statement to config parser

michaeldelago:
Please feel free to mention me (@mynamismeat) in discord. I prefer communication in a channel of the AC discord rather than as a DM.  

## Changes Proposed:
- Add an "include" statement to the configuration parser
  - This allows users to logically group configuration parameters with each other, based on the file they're in. 
  - the `include` statement works with either files or directories
  - `include` files are parsed _after_ the base file has completed parsing
  - included files are in-line with the root file. This means that a parameter after an include will supersede the include, and if the include is after the parameter, it will supersede the root file.
  - Cyclical `include` statements will cause a failure
- Configuration parser has been rewritten to use regex rather than reformatting the lines of the file. 
  - This can make it easier to expand the config file format in the future
- Empty files do not halt execution of the worldserver

## Tests Performed:

- It builds without issues
- It starts without issues
- It appears to parse all debug options without issues

## How to Test the Changes:

1. run an initial build with `./acore.sh docker build`
2. start the container with `docker compose --profile app up ac-worldserver`. hit CTRL+C after everything seems satisfactory
3. change the worldserver config (`env/docker/etc/worldserver.conf`). I adding this line to the bottom:

    ```ini
    include /azerothcore/env/dist/etc/worldserver.conf.dockerdist
    ```

4. Run the build and start the new container
5. Test the worldserver config with a directory. First, run this command:

    ```console
    $ mkdir env/docker/etc/worldserver.d
    $ mv env/docker/etc/worldserver.conf.dockerdist env/docker/etc/worldserver.d/dockerdist.conf
    ```

    Change the include statement at the bottom of the worldserver config:

    ```ini
    include /azerothcore/env/dist/etc/worldserver.d/
    ```

6. Again, build and start the container to ensure that the worldserver starts
7. To test a cyclical include, add `include /azerothcore/env/dist/etc/worldserver.conf` to the end of `env/docker/etc/worldserver.d/dockerdist.conf`. The worldserver should fail.

- With this change, I'd like to deprecate [the conf layering](https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/compiler/includes/functions.sh#L189-L220) that happens as a part of the compiler shell scripts
  - The whole section can be reduced to `echo "include /path/to/dockerdist" >> /path/to/main/conf`
  - Overall, I think a more featured configuration system should be favored over extra steps that happen right after compile time. 
- I'd also like to implement configuration sourcing from environment variables. I intended to do that with this PR, though I wasn't happy in the way it turned out so I removed it
  - it's probably better to keep this as small as possible, anyways.
- I removed erroring on duplicate parameters because I don't think that this is an issue. If this isn't shared by other contributors, an alternative I have in mind is  to prevent duplicate parameters in the same file, and then emitting an `info` log when a parameter gets overwritten between one file and another.  

### DIFF
--- a/src/server/apps/authserver/authserver.conf.dist
+++ b/src/server/apps/authserver/authserver.conf.dist
@@ -31,6 +31,19 @@
 ###################################################################################################
 
 ###################################################################################################
+# INCLUDING OTHER CONFIGURATION FILES
+#
+#    include ...
+#    includeDir ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will override conflicting configs occurring before it. Configs
+#                    after an include will override conflicting configs defined in the include
+#
+# include /etc/azerothcore/authserver.conf
+# includeDir /etc/azerothcore/common.d/
+###################################################################################################
+
+###################################################################################################
 # AUTH SERVER SETTINGS
 #
 #    LogsDir
@@ -431,14 +444,3 @@ Logger.root=4,Console Auth
 
 #
 ###################################################################################################
-#
-#    include ...
-#       Description: Include another configuration file. The paramaters from
-#                    that file will overwrite any conflicting parameters in this
-#                    file.
-#
-#include /etc/azerothcore/authserver.conf
-#
-## To include a directory (and all files under it), make sure there's a slash at the end of the path
-#
-#include /etc/azerothcore/common.d/
--- a/src/server/apps/authserver/authserver.conf.dist
+++ b/src/server/apps/authserver/authserver.conf.dist
@@ -431,3 +431,14 @@ Logger.root=4,Console Auth
 
 #
 ###################################################################################################
+#
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/authserver.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -4153,3 +4153,14 @@ Metric.OverallStatusInterval = 1
 
 #
 ###################################################################################################
+#
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/worldserver.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -53,6 +53,19 @@
 ###################################################################################################
 
 ###################################################################################################
+# INCLUDING OTHER CONFIGURATION FILES
+#
+#    include ...
+#    includeDir ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will override conflicting configs occurring before it. Configs
+#                    after an include will override conflicting configs defined in the include
+#
+# include /etc/azerothcore/worldserver.conf
+# includeDir /etc/azerothcore/common.d/
+###################################################################################################
+
+###################################################################################################
 # CONNECTIONS AND DIRECTORIES
 #
 #    RealmID
@@ -4153,14 +4166,3 @@ Metric.OverallStatusInterval = 1
 
 #
 ###################################################################################################
-#
-#    include ...
-#       Description: Include another configuration file. The paramaters from
-#                    that file will overwrite any conflicting parameters in this
-#                    file.
-#
-#include /etc/azerothcore/worldserver.conf
-#
-## To include a directory (and all files under it), make sure there's a slash at the end of the path
-#
-#include /etc/azerothcore/common.d/
--- a/src/tools/dbimport/dbimport.conf.dist
+++ b/src/tools/dbimport/dbimport.conf.dist
@@ -29,6 +29,19 @@
 ###################################################################################################
 
 ###################################################################################################
+# INCLUDING OTHER CONFIGURATION FILES
+#
+#    include ...
+#    includeDir ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will override conflicting configs occurring before it. Configs
+#                    after an include will override conflicting configs defined in the include
+#
+# include /etc/azerothcore/dbimport.conf
+# includeDir /etc/azerothcore/common.d/
+###################################################################################################
+
+###################################################################################################
 # DB IMPORT CONFIG
 #
 #    LogsDir
@@ -294,15 +307,6 @@ Appender.DBImport=2,5,0,DBImport.log,w
 #
 
 Logger.root=4,Console DBImport
+
+#
 ###################################################################################################
-#
-#    include ...
-#       Description: Include another configuration file. The paramaters from
-#                    that file will overwrite any conflicting parameters in this
-#                    file.
-#
-#include /etc/azerothcore/dbimport.conf
-#
-## To include a directory (and all files under it), make sure there's a slash at the end of the path
-#
-#include /etc/azerothcore/common.d/
--- a/src/tools/dbimport/dbimport.conf.dist
+++ b/src/tools/dbimport/dbimport.conf.dist
@@ -295,3 +295,14 @@ Appender.DBImport=2,5,0,DBImport.log,w
 
 Logger.root=4,Console DBImport
 ###################################################################################################
+#
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/dbimport.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
--- a/apps/compiler/includes/functions.sh
+++ b/apps/compiler/includes/functions.sh
@@ -160,8 +160,10 @@ function comp_compile() {
           # if dist file doesn't exist, skip this iteration
           [ ! -f "$distPath" ] && continue
 
-          # replace params in foo.conf.dist with params in foo.conf.dockerdist
-          conf_layer "$dockerdist" "$distPath" " # Copied from dockerdist"
+          # override configs in configs in the root file with those needed for
+          # docker
+          echo >> "$distPath"
+          echo "include \"$(basename "$dockerdist")\"" >> "$distPath"
 
           # Copy modified dist file to $confDir/$filename
           # Don't overwrite foo.conf if it already exists.
@@ -184,37 +186,4 @@ function comp_build() {
 function comp_all() {
   comp_clean
   comp_build
-}
-
-# conf_layer FILENAME FILENAME
-# Layer the configuration parameters from the first argument onto the second argument
-function conf_layer() {
-  LAYER="$1"
-  BASE="$2"
-  COMMENT="$3"
-
-  # Loop over all defined params in conf file
-  grep -E "^[a-zA-Z\.0-9]+\s*=.*$" "$LAYER" \
-    | while read -r param
-      do
-        # remove spaces from param
-        # foo       = bar becomes foo=bar
-        NOSPACE="$(tr -d '[:space:]' <<< "$param")"
-
-        # split into key and value
-        KEY="$(cut -f1 -d= <<< "$NOSPACE")"
-        VAL="$(cut -f2 -d= <<< "$NOSPACE")"
-        # if key in base and val not in line
-        if grep -qE "^$KEY" "$BASE" && ! grep -qE "^$KEY.*=.*$VAL" "$BASE"; then
-          # Replace line
-          # Prevent issues with shell quoting 
-          sed -i \
-            's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' \
-            "$BASE"
-        else
-          # insert line
-          echo "$KEY = $VAL$COMMENT" >> "$BASE"
-        fi
-      done
-  echo "Layered $LAYER onto $BASE"
 }
--- a/src/common/Configuration/Config.cpp
+++ b/src/common/Configuration/Config.cpp
@@ -164,7 +164,7 @@ namespace
                 if (match.size() == 2)
                 {
                     std::string includeFile = std::ssub_match(match[1]).str();
-                    // Save include statesments to be parsed after all the keys
+                    // Save include statements to be parsed after all the keys
                     // are added for this file.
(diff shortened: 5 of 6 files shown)


Kitzunu:
I don't get this.
Changing anything in the conf file will overwrite the original config value taken from the core

Kitzunu:
And duplicate values for the same config should definetly generate an error

michaeldelago:
> I don't get this. Changing anything in the conf file will overwrite the original config value taken from the core

I'm glad that there's default config options throughout the codebase, but I don't think that's the same thing as manually specifying configuration parameters. I prefer to be as specific as possible when specifying configs, even if that means that the default value is being left in.

Being able to layer configs with an `include` statement makes it much easier to compose a configuration. 

A general example I can give for this is running multiple worldservers that all share some config options, but they also each have specific configs of their own. Being able to have a `worldserver.conf` that can `include` those common parameters, but also the specific config file containing options for that server could make managing all of them a bit easier.

```ini
# worldserver.conf, at the end of the file
include /etc/azerothcore/common-worldserver.conf
# mounted through docker to have a generic name
include /etc/azerothcore/custom-worldserver.conf # custom world conf
```
An include directive is a common pattern. Some examples from popular programs:

- [Apache](https://httpd.apache.org/docs/2.4/mod/core.html#include)
- [git](https://git-scm.com/docs/git-config#_includes)
- [redis](https://redis.io/docs/management/config-file/) (no anchors for an easy link but it's the first main section)
- [mysql](https://dev.mysql.com/doc/refman/8.0/en/option-files.html#option-file-inclusions)


> I just want to point out that we dont read the .dist file. All default values are defined in the core. The .conf file could just contain 1 config to override that specific one and it would work

That's correct, the servers don't read the dist file. [Documentation](https://www.azerothcore.org/wiki/server-setup#creating-the-config-files) says to just copy the `.conf.dist` files into the `.conf` files. the `.conf.dockerdist` files is a different story, however. As a part of the [docker build process](https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/compiler/includes/functions.sh#L154), these files are merged/layers overtop the `.conf.dist` file, and as long as the `.conf` file doesnt exist, the `.conf.dist` file is copied into the `.conf` file.

walkline:
> included files are "layered" on top of whichever file contained the include statement. This means that any parameters in the base are overwritten by the included file.

This line doesn't make sense to me as well. When I see any include command, I expect the file that has the include keyword (root config) to have higher priority than an included file.

E.g.

common.conf:
```
MyOptionA=0
```

root.conf:
```
include common.conf

MyOptionA=1
```

In your case, MyOptionA would be equal to 0, but I would expect 1.

However, I would understand this case:

root.conf:

```
MyOptionA=1

include common.conf
```

MyOptionA is equal to 0 since include comes after the declaration of the variable. 
But still, it can create some confusion and problems with understanding what value you will get if you have a bunch of configs.

By the way support of environment variables implemented here - https://github.com/azerothcore/azerothcore-wotlk/pull/16817

michaeldelago:
@walkline This makes sense to me. I was thinking of it in only placing includes at the bottom, but I agree that it makes more sense to include the statements inline.

Duplicate values skipping in the same file has been added back, as well

Kitzunu:
Okay I will dum it down to see if I understand it.

Basically what you are doing is opening up the possibility to read several conf files. So you could have

worldserverServerConf.conf - Read by Worldserver 1
 - includes worldserverDatabaseConnection.conf
worldserverCustomConf.conf - Read by Worldserver 2
 - includes worldserverDatabaseConnection.conf

worldserverDatabaseConnection.conf - Read by WS 1 & 2
Which could hold options like CharacterDatabaseInfo

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -4153,3 +4153,14 @@ Metric.OverallStatusInterval = 1
...
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/worldserver.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
```
Kitzunu, line 4166: I would like this to follow the style used by conf files, and I would like it to be placed under AllowLoggingIPAddressesInDatabase option.

This is what I propse it to look like, but I am open to suggestions

```conf
#
#    ConfigurationInclude.file
#    ConfigurationInclude.dir
#        Description: The path to other configuration files to be loaded.
#                     Several files are seperated with a comma.
#                     A directory needs to end with a forward slash (/)
#        Example:     "C:/acore/conf/config1.conf"
#                     "C:/acore/conf/"
#        Default:     "" - (Disabled)
#

ConfigurationInclude.file = ""
ConfigurationInclude.dir  = ""
```

src/tools/dbimport/dbimport.conf.dist
```diff
@@ -295,3 +295,14 @@ Appender.DBImport=2,5,0,DBImport.log,w
...
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/dbimport.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
```
Kitzunu, line 308: I would like it placed under TempDir

src/server/apps/authserver/authserver.conf.dist
```diff
@@ -431,3 +431,14 @@ Logger.root=4,Console Auth
...
+#    include ...
+#       Description: Include another configuration file. The paramaters from
+#                    that file will overwrite any conflicting parameters in this
+#                    file.
+#
+#include /etc/azerothcore/authserver.conf
+#
+## To include a directory (and all files under it), make sure there's a slash at the end of the path
+#
+#include /etc/azerothcore/common.d/
```
Kitzunu, line 444: I would like it placed  under AllowLoggingIPAddressesInDatabase

michaeldelago:
@Kitzunu 

yeah. Perhaps a simpler example:

someone stores their worldserver/authserver config in git to keep track of changes, but they don't want to put their database passwords into git. At the end of the config in git, there's this line

```ini
include secrets.conf
```

During deployment, they grab the configuration file containing the 3 `DatabaseInfo` lines that have the passwords from a more secure storage, saving it as `secrets.conf`

with the `include` statement, it's much easier for the server admin to add these secrets in. Without it, they'd have to edit the config in some way that is likely going to be much less reliable.

src/server/apps/worldserver/worldserver.conf.dist
michaeldelago: I like the separate statement for specifying to include a directory. Makes the whole thing more clear. 

I prefer keeping the syntax for specifying an include different from the syntax for declaring a config parameter value. 

I'm thinking that it should be OK to include multiple files, such as 

```ini
include foo.conf
include bar.conf
```

The `include` statement is a property of syntax rather than a new configuration parameter and because of that, I think we should treat it like a property of the config files syntax. Having the `include` be declared the same way as a standard parameter but follow different rules may be confusing

```ini
# This is not OK, only the first parameter is read
DataDir = "/foo"
DataDir = "/bar"
# This is perfectly fine, both files get included
ConfigurationInclude.file = "foo.conf"
ConfigurationInclude.file = "bar.conf"
```

---

As far as placement in the config file for the documentation, I'm thinking after this section near the top of the file (where the syntax is described) makes the most sense

https://github.com/azerothcore/azerothcore-wotlk/blob/a4444a8b9d9d36d87b0f2a35f1a14ee2d9bd12da/src/server/apps/worldserver/worldserver.conf.dist#L40-L53

Kitzunu: Okay I get your point of having it seperate. Then I want it at the top of the file. But I still want it to follow the config style at the base.

My suggestion would be ... "bar.conf;foo.conf"

michaeldelago: so just to be clear, something could look like this? It's an extreme example that we (hopefully) won't ever see, but I'm just trying to make sure it looks how we want it too

```ini

# Multiple includes can be specified with a semicolon between then, or 
# separate include statements
include "/path/to/conf1.conf;/path/to/conf2.conf"
include "/path/to/conf3.conf"

# if foo or baz parameter occurs in the above config files, they will 
# be overwritten
foo = bar
baz = blah

# if foo or baz is in the below config files ,
# they will overwrite the foo or baz that happens above

# Directories can be included with the includeDir statement, and they 
# support semicolon as a separator as well
includeDir "/path/to/confDir"
include "/path/to/conf4.conf"
```

Kitzunu: Yep

Kitzunu: Looks good.

What happens if you have the same config option in 2 different files you include?

f.ex if I were to have WorldDatabaseInfo in conf1, conf2, worldserver.conf

michaeldelago: It's wherever it occurs "last". A way to think about `include` is that it functions as if you put all the parameters from the file you're referencing into the "root" file (worldserver.conf in this case). That said, duplication rules _in the same file_ still apply.


---

The final configuration value (and the one that is used) is `conf2`

<details>

<summary>worldserver.conf</summary>

```ini
include path/to/conf1.conf

WorldDatabaseInfo = "root"

include path/to/conf2.conf
```

</details>

<details>

<summary>path/to/conf1.conf</summary>

```ini
WorldDatabaseInfo = "conf1"
```

</details>

<details>

<summary>path/to/conf2.conf</summary>

```ini
WorldDatabaseInfo = "conf2"
```

</details>


This config "expands" out into something like this:

<details>

<summary>Config Expansion</summary>

```ini
WorldDatabaseInfo = "conf1"

WorldDatabaseInfo = "root"

WorldDatabaseInfo = "conf2"
```

</details>

---

The final configuration value is `prod`

<details>

<summary>worldserver.conf</summary>

```ini
WorldDatabaseInfo = "root"

include path/to/local.conf

include path/to/prod.conf
```

</details>

<details>

<summary>path/to/local.conf</summary>

```ini
WorldDatabaseInfo = "local"
```

</details>

<details>

<summary>path/to/prod.conf</summary>

```ini
WorldDatabaseInfo = "prod"
```

</details>

the `prod.conf` takes presedence, since it's where the last value comes from

<details>

<summary>Config Expansion</summary>

```ini
WorldDatabaseInfo = "root"

WorldDatabaseInfo = "local"

WorldDatabaseInfo = "prod"
```

</details>

---

The final configuration value is `conf2`

<details>

<summary>worldserver.conf</summary>

```ini
# worldserver.conf
WorldDatabaseInfo = "root"

include path/to/conf1.conf

include path/to/conf2.conf
```

</details>

<details>

<summary>path/to/conf1.conf</summary>

```ini
# test/to/conf1.conf
WorldDatabaseInfo = "conf1"
```

</details>

<details>

<summary>path/to/conf2.conf</summary>

```ini
# test/to/conf2.conf
WorldDatabaseInfo = "conf2"
```

</details>

michaeldelago: Also, most of the examples above are probably larger than we'd realistically see "in the wild". The main reason I wrote this PR is to help reduce the complexity/magic in the docker image build process:

<details>

<summary>Shell functions used in dashboard after compilation</summary>

https://github.com/azerothcore/azerothcore-wotlk/blob/026fe7c9627c01d10c674c6de5936ad2c1d0c94f/apps/compiler/includes/functions.sh#L189-L220

https://github.com/azerothcore/azerothcore-wotlk/blob/026fe7c9627c01d10c674c6de5936ad2c1d0c94f/apps/compiler/includes/functions.sh#L152-L169

</details>

This code controls how the files at `env/docker/etc/*.dockerdist` work. 

What happens is that after the docker build compiles, it will take all of the `dockerdist` files and do a find-and-replace so the proper values get set up in the `worldserver.conf`, `authserver.conf` that get created to start playing on the docker infra immediately. 

This works, but I think it's harder to maintain, there's more that can go wrong, and I think it's too much "extra" being done by the build process rather than doing this in code. I have a goal to greatly simplify the docker/docker-compose build process (see https://github.com/azerothcore/azerothcore-wotlk/pull/14990), and I think the changes in the this PR are needed as a part of that. 

With this PR, that ugly (I can call it ugly, I wrote it :p) function can be be reduced to this:

```bash
echo "include \"$(basename "dockerdist")\"" >> "$confDir/$filename"
```

Since this change is the underlying goal of the PR, I've included this change to help make the use of the feature clear and so I don't have to make another PR as soon as this one is merged, if it's merged.

ben-of-codecraft:
@michaeldelago since you are in there making changes to compiler/function.sh 

one issue in that file is for Mac ARM builds:
```shell
OSOPTIONS=""


    echo "Platform: $OSTYPE"
    case "$OSTYPE" in
      darwin*)
        OSOPTIONS=" -DMYSQL_ADD_INCLUDE_PATH=/usr/local/include -DMYSQL_LIBRARY=/usr/local/lib/libmysqlclient.dylib -DREADLINE_INCLUDE_DIR=/usr/local/opt/readline/include -DREADLINE_LIBRARY=/usr/local/opt/readline/lib/libreadline.dylib ->
        ;;
      msys*)
        OSOPTIONS=" -DMYSQL_INCLUDE_DIR=C:\tools\mysql\current\include -DMYSQL_LIBRARY=C:\tools\mysql\current\lib\mysqlclient.lib "
        ;;
    esac

``` 

/usr/local/* for the OSOPTIONS are not going to work, it was broken for me: 

I think this is safe to set the right libraries paths: 
```sh
echo "Platform: $OSTYPE"
case "$OSTYPE" in
  darwin*)
    ARCHTYPE=$(uname -m)
    if command -v brew &> /dev/null && [ "$ARCHTYPE" = "arm64" ]; then
      OSLIBPATH="/opt/homebrew"
    else 
      OSLIBPATH="/usr/local"
    fi
    OSOPTIONS=" -DMYSQL_ADD_INCLUDE_PATH=$OSLIBPATH/include -DMYSQL_LIBRARY=$OSLIBPATH/lib/libmysqlclient.dylib -DREADLINE_INCLUDE_DIR=$OSLIBPATH/opt/readline/include -DREADLINE_LIBRARY=$OSLIBPATH/opt/readline/lib/libreadline.dylib -DOPENSSL_INCLUDE_DIR=$OSLIBPATH/opt/openssl@1.1/include -DOPENSSL_SSL_LIBRARIES=$OSLIBPATH/opt/openssl@1.1/lib/libssl.dylib -DOPENSSL_CRYPTO_LIBRARIES=$OSLIBPATH/opt/openssl@1.1/lib/libcrypto.dylib "
    ;;
  msys*)
    OSOPTIONS=" -DMYSQL_INCLUDE_DIR=C:\tools\mysql\current\include -DMYSQL_LIBRARY=C:\tools\mysql\current\lib\mysqlclient.lib "
    ;;
esac
```

michaeldelago:
Closing this as the desired end result has been superseded by #17040

PR #16817, opened 2023-07-25
feat(Core/Config): Implement config override with env vars

walkline:
## Changes Proposed:
Implement overriding of configuration from the .conf file with environment variables. Environment variables keys are autogenerated based on the keys defined in .conf file.

Usage example:
```
$ export AC_DATA_DIR=/usr
$ AC_WORLD_SERVER_PORT=8080 ./worldserver
```

On windows:
```
> $env:AC_REALM_ID = '2'; .\worldserver
```

## Issues Addressed:

- Closes 

## Tests Performed:

- Run the server with an overridden config. 

### DIFF
--- a/src/common/Configuration/Config.cpp
+++ b/src/common/Configuration/Config.cpp
@@ -21,6 +21,7 @@
 #include "StringFormat.h"
 #include "Tokenize.h"
 #include "Util.h"
+#include <cstdlib>
 #include <fstream>
 #include <mutex>
 #include <unordered_map>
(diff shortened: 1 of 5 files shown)


Kitzunu:
and why isnt the conf file good enough?

walkline:
The conf file is fine, but there are some cases when environment variables would be more preferable. One such case is a cloud world with containerization. In that world, environment variables are a standard way to pass configuration to the containers.

Also, maybe we can simplify our docker setup by utilizing environment variables.

One more use case is clustering. With clustering, in most cases, you need to run several worldservers with 99% identical conf file. Using environment variables, you can simply use the same conf file and just override the things that are required (like port values).

Kitzunu:
Okay, now we're getting somewhere :D 

About docker, I will simply put it as nobody active within the AC devs knows anything docker.

> Environment variables keys are autogenerated based on the keys defined in .conf file.

What do you mean by this? Will the env vars automatically get filled with AC vars? 
Because I do not think we want to do that, as the fallback should always be the base conf values that are defined in the core if a config option is missing.

And for read order I guess it will go

1. Env var
2. conf file
3. base conf from core

walkline:
This PR doesn’t modify any env vars (you need to set env vars somewhere externally); it only reads them. 

The code that you mentioned generates env var names so it can attempt to read them. If it finds an env var with the corresponding key, it uses the env var value; if not, it takes the value from the config. 

So basically it follows the same order as you described.

Kitzunu:
Okay then I dont have any issues with this pr 👍

Kitzunu:
So GM.InGMList.Level would become AC_GM_IN_GML_IST_LEVEL if I understand it correctly.
Trying to understand everything so I can document it later

walkline:
> TEST_F(ConfigEnvTest, ValueWithSeveralUpperlLaters)
{
    EXPECT_EQ(sConfigMgr->GetOption<int>("GM.InGMList.Level", 1), 50);
    setenv("AC_GM_IN_GMLIST_LEVEL", "42", 1);
    EXPECT_EQ(sConfigMgr->OverrideWithEnvVariablesIfAny().empty(), false);
    EXPECT_EQ(sConfigMgr->GetOption<int>("GM.InGMList.Level", 0), 42);
}

@Kitzunu I wrote this code a long time ago and forgot how it works. I expected `GM.InGMList.Level` to be equal to `AC_GM_IN_GM_LIST_LEVEL`, but I decided to check this with a test. As a result, we were wrong, it should be `AC_GM_IN_GMLIST_LEVEL` 😁

walkline:
Any additional actions required to merge it?

Kitzunu:
I will just test it and review the code and then it should be good to go I reckon

Kitzunu:
at first I didnt think it worked, but I figured out I made a user variable instead of a system variable. with sys var it works on Windows

Kitzunu (approved):

Kitzunu:
Well I just run windows, and that's all I gonna test :D

Issue #16274, opened 2023-05-15
[Karazhan] Spectral Performer Spotlight ability


v-mstrs:
https://github.com/chromiecraft/chromiecraft/issues/5393

### Current Behaviour

Only works for mobs, it should also affect players. Range is too big, should be shown as debuff when standing in it

### Expected Blizzlike Behaviour

Spotlight will place a beam of light on the ground, increasing damage dealt for anyone inside, enemies or players. (aura as debuff)

### Source

https://www.wowhead.com/tbc/guide/trash-karazhan-strategy-burning-crusade-classic
https://youtu.be/Iv6zSyhmLGc?t=675

### Steps to reproduce the problem

1. .teleport Karazhan , enter
2. .go xyz -10917.5 -1745.3 90.4 532

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/00e42645b8b6

### Operating system

Linux Mint 21.1

elthehablo:
Reported again at: https://github.com/chromiecraft/chromiecraft/issues/5618

v-mstrs:
`INSERT INTO spell_linked_spell VALUES (29683, 32214, 0, "Spotilght should affect everyone");`
 from https://github.com/SunstriderEmu/BugTracker/commit/8c4235be673999ce424c911472d74d50cf7edf33
 You now do get the 20% dmg buff (shown as debuff) but it stacks still
 
 enemies get 50% that comes from [29683](https://www.wowhead.com/tbc/spell=29683/spotlight) ?? also stacks
![2023-06-19_15-36](https://github.com/azerothcore/azerothcore-wotlk/assets/104088833/308a91d4-77ff-4bf1-aa65-d2db54bfb2aa)

PR #16819, opened 2023-07-26
fix(DB/spell_linked_spell) Add 20% buff to Spotlight

v-mstrs:
https://github.com/SunstriderEmu/BugTracker/commit/8c4235be673999ce424c911472d74d50cf7edf33
## Changes Proposed:
-  Add 20% buff to Spotlight

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16274
- Closes https://github.com/chromiecraft/chromiecraft/issues/5393
- Closes https://github.com/chromiecraft/chromiecraft/issues/5618

## Tests Performed:

- Works

## How to Test the Changes:

1. enter karazhan
2. .go xyz -10917.5 -1745.3 90.4 532

- (currently) Spectral Performer cast this randomly between 1 - 8 seconds and repeat this every 20-35 seconds
so sometimes it can be casted on pull making this ability redundant. A solution could be to increase from 1 to 3 seconds(?)
- Pretty sure enemies SHOULD get 50% buff

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1690342063161377379.sql
+++ b/data/sql/updates/pending_db_world/rev_1690342063161377379.sql
@@ -1,0 +1,2 @@
+--
+INSERT INTO spell_linked_spell VALUES (29683, 32214, 0, "Spotlight 20% player buff");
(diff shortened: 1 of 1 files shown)


elthehablo:
Please make a more descriptive title, and the TODO list has check boxes, in markdown they are styled as follows `- [ ]` you can put an X in them to mark them as completed like so: `- [x]`

elthehablo (changes requested):
Insert statements ALWAYS need a delete statement in there too, even if there was previously no value present. Also you are missing column names. Take a look at https://www.azerothcore.org/wiki/sql-standards

elthehablo:
tested, works. enemies seem to get 50% players get 20%. so if that is how it is supposed to be, it does seem to work

elthehablo:
Nice work. Should be good to go now

elthehablo:
Although fix TODO maybe

PR #16825, opened 2023-07-26
fix(Core/Position): Normalize Position.GetRelativeAngle

TuriansNotBad:
## Changes Proposed:
-  Normalize angle returned by both Position.GetRelativeAngle functions.
-  Currently GetRelativeAngle can return angles values which will later cause issues in ChaseMovementGenerator.PositionOkay. For example if it returns a relative angle -4.9 it is then passed onto ChaseAngle.IsAngleOkay function. For desired RelativeAngle value of 3.2 we will get a diff value of 8.1 which is larger than 2pi and will make the 2nd argument to std::min negative, returning said argument. Given that Tolerance for most sane applications should be positive this will cause the function to sometimes return incorrect result.

https://github.com/azerothcore/azerothcore-wotlk/assets/54774532/7e15fe71-a06c-4084-8d11-b25f627a19b2

From PetAI.cpp, PetAI::DoAttack:
```cpp
float angle = combatRange == 0.f && target->GetTypeId() != TYPEID_PLAYER && !target->IsPet() ? float(M_PI) : 0.f;
float tolerance = combatRange == 0.f ? float(M_PI_4) : float(M_PI * 2);
me->GetMotionMaster()->MoveChase(target, ChaseRange(0.f, combatRange), ChaseAngle(angle, tolerance));
```
From this, melee pets should stand right behind the target with pi/4 tolerance. combatRange is set based on if its gonna cast ranged spells, for melees should be 0.

I tested this in Naxxramas with warlock's Felhunter pet. Isolate any single enemy with lots of health, ensure you have a massive threat lead on your pet and start slowly going around in circles.
In my case I moved until my pet was somewhere about 1.5 radians absolute angle to the target. (target:GetAngle(pet) ~ 1.5 radians) and target's orientation was ~5.9 radians. After this I changed the direction of my rotation and as you can first see at roughly 10 seconds in the video and again at the end pet does not move behind it despite actual relative angle diff being approx double the tolerance (pi/4). For some combinations of absolute angle, orientation, and chase angle it works as expected, for others not at all. Also note how small of an angle change it reacts to at the beginnig of the video.

It should not be possible to replicate this with this PR.

There really aren't any other times where ChaseAngle is used that I found.
GetRelativeAngle is used only in a couple of other places but they were passing angle to cos/sin so were never affected.

tldr; this same funciton is normalized in trinity core if you don't trust my math/footage.
https://github.com/TrinityCore/TrinityCore/blob/f9033a5dbd559fde3030a21d83d664983d95f39f/src/server/game/Entities/Object/Position.h#L139C5-L139C5.

## Issues Addressed:

- Closes 

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Object/Position.h
+++ b/src/server/game/Entities/Object/Position.h
@@ -196,10 +196,10 @@ struct Position
 
     float GetRelativeAngle(const Position* pos) const
     {
-        return GetAngle(pos) - m_orientation;
+        return NormalizeOrientation(GetAngle(pos) - m_orientation);
     }
 
-    [[nodiscard]] float GetRelativeAngle(float x, float y) const { return GetAngle(x, y) - m_orientation; }
+    [[nodiscard]] float GetRelativeAngle(float x, float y) const { return NormalizeOrientation(GetAngle(x, y) - m_orientation); }
     [[nodiscard]] float ToAbsoluteAngle(float relAngle) const { return NormalizeOrientation(relAngle + m_orientation); }
 
     void GetSinCos(float x, float y, float& vsin, float& vcos) const;


Nefertumm:
Any certain case to test?

TuriansNotBad:
@Nefertumm added more info

Nefertumm:
Thank you for the details!

Nefertumm (approved):

Kitzunu:
@Nefertumm did you test it?

Nefertumm:
No

PR #16828, opened 2023-07-27
Fix XP disable NPC script

Foereaper:
## Changes Proposed:
- Fixes the script to use the xp toggle cost config option properly.

## Issues Addressed:

- Previously this config variable did not work, as the cost defined in the db would override this config option.

## Tests Performed:

- It builds, logic should be fine. Feel free to test ingame.

## How to Test the Changes:

- Set the config option to something else than 10g, it should now reflect properly ingame.

### DIFF
--- a/src/server/scripts/World/npcs_special.cpp
+++ b/src/server/scripts/World/npcs_special.cpp
@@ -2033,8 +2033,19 @@ public:
 
     bool OnGossipSelect(Player* player, Creature* /*creature*/, uint32 /*sender*/, uint32 action) override
     {
+        auto toggleXpCost = sWorld->getIntConfig(CONFIG_TOGGLE_XP_COST);
+        
         ClearGossipMenuFor(player);
-
+        
+        if (!player->HasEnoughMoney(toggleXpCost))
+        {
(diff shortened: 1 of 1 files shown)


avarishd:
The amount of gold is now properly shown in the box, but you've removed the logic for handling the gold.

Foereaper:
Does the core not handle this automatically when a gossip option has moneyBox defined? It definitely does in PlayerGossip, so I assumed it was handled the same throughout the core.

avarishd:
Afaik, there is no logic at all for [ScriptedGossip](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/AI/ScriptedAI/ScriptedGossip.cpp), thus usage of action(s), however there is for [PlayerGossip](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Player/PlayerGossip.cpp#L287).

Foereaper:
I restored the money logic, although a little differently to prevent the amount of nesting in the original function.

avarishd:
All works now, although I'm unsure about using auto to assign data type. Have to wait for the actual reviewers feedback on that.

Kitzunu (approved):

Issue #16705, opened 2023-07-08
 [Hillsbrad Foothills] NPC High Executor Darthalia (1645) - should stand still 


wowquester:
### Current Behaviour

High Executor Darthalia walks in a strange way, short distances, makes "hard" to talk to her for a longer time.

https://github.com/azerothcore/azerothcore-wotlk/assets/138876396/0dbe4c0e-2491-40a1-8932-afb6ba097274

### Expected Blizzlike Behaviour

She should stay still.

### Source

https://www.youtube.com/watch?v=6qaFdQsL4aM&t=390s (i know it's small, but you can see it in the bg she doesn't move at all)

### Steps to reproduce the problem

.go xyz -36.3284 -927.111572 54.467083
See for yourself

### AC rev. hash/commit

bda9b404038e

### Operating system

Win11x64

### Custom changes or Modules

None

wowquester:
Thanks!

PR #16829, opened 2023-07-27
fix(Creature): High Executor Darthalia movement

ZhengPeiRu21:
## Changes Proposed:
-  Remove movement from High Executor Darthalia

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16705

## SOURCE:

See original issue

## Tests Performed:

- NPC no longer moves

## How to Test the Changes:

1. .go xyz -36.3284 -927.111572 54.467083

- I did not confirm with Classic that this is correct behavior for the NPC, but the linked video seems correct. Someone could check if they would like to be more confident this is correct behavior.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1690478105252488100.sql
+++ b/data/sql/updates/pending_db_world/rev_1690478105252488100.sql
@@ -1,0 +1,3 @@
+--
+UPDATE `creature` SET `wander_distance` = 0, `MovementType` = 0 WHERE `guid` = 15542;
+


Gultask (approved):

heyitsbench:
This fix is unfortunately incorrect, Darthalia does have wander movement on Wrath Classic.

ZhengPeiRu21:
> This fix is unfortunately incorrect, Darthalia does have wander movement on Wrath Classic.

Thank you Bench for checking. I will revert the change.

elthehablo:
> The issue here seems to be that talking to her doesn't pause the wander movement (?) as it does with waypoints

Might need some core changes then? Add boolean for interrupt movement on gossip or something? Probably already exists but idk

PR #16832, opened 2023-07-27
feat(Core): Add clustering support

### LINKED ISSUE #11072: Feature: clustering

Related of https://github.com/azerothcore/azerothcore-wotlk/issues/1141

Got idea from Fall1n who has done it on SunwellCore in his private core
https://www.reddit.com/r/wowservers/comments/8u1s8i/spell_attro_cu_fake_delay_does_it_work_only_on/e1c2ol6/
https://bitbucket.org/lukaasm/hellground-wowemu/commits/d09c115242ee9f57da7eed11e8c764f7faeb2f70

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/106149193-feature-clustering?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

walkline:
## Changes Proposed:
This is a work in progress, but the aim of this PR is to add support for clustering.

In short, this feature allows players to be split between several world servers in the same realm. However, this is only the tip of the iceberg. More details about the benefits of this approach and some architecture explanations are described here - https://github.com/azerothcore/azerothcore-wotlk/discussions/16748.

It's a work in progress (WIP), but it already provides testable functionality. The tiny version of the demo is available here - https://youtu.be/TCYXdrRMqT4. In this demo, I started 2 worldservers, one specifically for one instance map, and the second one for the rest of the available maps. You will see how players switch between the worldservers.

Also, there is a demo that covers more features - https://youtu.be/4C7487KpzjM.

This PR integrates with the https://github.com/walkline/ToCloud9 project, and most of the logic that solves distributed system problems is implemented in that repository.

## How to Test the Changes:

You can test it with docker-compose instructions describer in the [README file](https://github.com/walkline/ToCloud9#readme).
For without docker and Windows + WSL users use [this document](https://github.com/walkline/ToCloud9/blob/master/doc/RunNonDockerWinWSLAzerothCore.md).
For without docker on the rest of the platforms documentation would be added later.

## Features and TODO List:

- [x] Worldservers and other services discovery. 
- [x] Services communication with NATS and gRPC.
- [x] Redirect players from one worldserver to another on map change.
- [x] Worldserver crash recovery (players would be redirected to the another available worldserver).
- [x] Automatic load balancing maps between worldservers.
- [x] Shared pool of Items and Players GUID. 
- [x] Shared pool of Instances GUID.
- [x] Who opcode in cluster support. 
- [x] Whispers in cluster support.
- [x] Guild related opcodes in cluster support (except cases below).
- [ ] Guild creation in cluster support.
- [ ] Guild bank in cluster support.
- [x] Mail in cluster support.
- [x] Auction hause in cluster support.
- [x] Friends list in cluster support.
- [x] Global channels support.
 90% Parties (and Raids) in cluster support.
- [x] Sync instance data.
- [x] Battlegrounds in cluster support.
- [x] Battlegrounds cross-realm support.
- [ ] Arenas in cluster support.
- [ ] LFG in cluster support.
- [x] Sync transports between worldservers.
- [x] Save logs to a file feature.
- [x] Create Helm chart for the whole solution.
- [ ] Some other stuff...

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11072

## Help Appreciated
1. I’m looking for contributors. You can contribute to this project - https://github.com/walkline/ToCloud9, or to this AzerothCore branch - https://github.com/walkline/azerothcore-wotlk/tree/cluster-mode. Yeah, maybe Go language is not well known in this community, but it is an easy language to learn. Additionally, ToCloud9's microservice architecture allows you to use other languages like C++, C#, Python, Rust or any other similar language. For example, you could try implementing the auction house service in a different language. You can count on my assistance.
2. Provide feedback about the project and the general idea here or in the [Discord channel](https://discord.gg/QxfBD9uGbN).
3. Press the Star button in [this repository](https://github.com/walkline/ToCloud9). By doing so, you increase the project's visibility, making it more likely to attract contributors.

## Summary by CodeRabbit

* **New Features**
  * Added cluster mode support for distributed worldserver deployments with cross-realm capabilities.
  * Introduced dynamic map assignment and player redirection mechanisms for load balancing.
  * Enhanced battleground start and player management for multi-server environments.
  * Added realm-aware GUID generation for characters and items in cross-realm scenarios.

* **Configuration**
  * New cluster settings configuration section with `Cluster.Enabled`, `Cluster.AvailableMaps`, and `Cluster.IsCrossrealm` options.


Winfidonarleyan:
Wow, very good

Kitzunu:
Cool cool. But I will recommend you doing it in smaller PRs, big prs have a tendency of being overlooked from reviews

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
```diff
@@ -0,0 +1,255 @@
...
+#include "TC9GrpcHandler.h"
+#include "ObjectAccessor.h"
+#include "Item.h"
+#include "Bag.h"
+#include "Player.h"
+
+GetPlayerItemsByGuidsResponse ToCloud9GrpcHandler::GetPlayerItemsByGuids(uint64 playerGuid, uint64* items, int itemsLen)
+{
+    Player *player = ObjectAccessor::FindPlayer(ObjectGuid(playerGuid));
+    if (player == nullptr)
```
Kitzunu, line 27: if (!player)

walkline:
> Cool cool. But I will recommend you doing it in smaller PRs, big prs have a tendency of being overlooked from reviews

@Kitzunu I would love to go that way. But this means that those small changes need to be merged to start working on another one. And in that case, it would mean that cluster mode changes would be merged, and some functionality would not work since the feature is not 100% complete. Or it's not a problem?

Kitzunu:
I just thought if you could implement the base stuff in one PR, and then like all guild related cluster stuff in one PR etc.

But this is fine too, just speaking from experience :D

walkline:
> I just thought if you could implement the base stuff in one PR, and then like all guild related cluster stuff in one PR etc.
> 
> But this is fine too, just speaking from experience :D

@Kitzunu Then this PR can be considered as the base one. Can I remove the WIP prefix in that case?

By the way, this PR would be even smaller once https://github.com/azerothcore/azerothcore-wotlk/pull/16817 gets merged.

Just to clarify, most of the work is done outside of AzerothCore. For example guild and mails are handled by ToCloud9 components - [guilds](https://github.com/walkline/ToCloud9/blob/master/apps/game-load-balancer/session/guild.go), [mails](https://github.com/walkline/ToCloud9/blob/master/apps/game-load-balancer/session/mail.go).

Kitzunu:
got it

TheSCREWEDSoftware:
@walkline what's the current state of this PR?

walkline:
I'm still working on it. In upcoming months the update would include battlegrounds support with cross-realm capability.

src/server/game/Entities/Object/ObjectGuid.h
```diff
@@ -18,14 +18,15 @@
...
-#include "Define.h"
 #include <deque>
 #include <functional>
 #include <list>
 #include <set>
 #include <unordered_set>
 #include <vector>
+#include "ByteBuffer.h"
+#include "Define.h"
+#include "TC9Sidecar.h"
```
sudlud, line 29: Thanks for the update regarding the includes! to my understanding the library includes should still come after the project specific includes like it was before here. So the order is partner-header-file, project includes, external library includes

src/server/game/Server/Protocol/Opcodes.h
```diff
@@ -1338,7 +1338,9 @@ enum Opcodes : uint16
     SMSG_COMMENTATOR_SKIRMISH_QUEUE_RESULT1         = 0x51C,
     SMSG_COMMENTATOR_SKIRMISH_QUEUE_RESULT2         = 0x51D,
     SMSG_MULTIPLE_MOVES                             = 0x51E, // uncompressed version of SMSG_COMPRESSED_MOVES
-    NUM_MSG_TYPES                                   = 0x51F
+    TC9_CMSG_PREPARE_FOR_REDIRECT                   = 0x51F,
+    TC9_SMSG_READY_FOR_REDIRECT                     = 0x520,
```
sudlud, line 1342: Can we be sure that these are not in use by the client for a different purpose? Or if this list is really complete already as of now

src/server/game/Battlegrounds/Battleground.cpp
```diff
@@ -272,6 +273,7 @@ void Battleground::Update(uint32 diff)
         if (!GetInvitedCount(TEAM_HORDE) && !GetInvitedCount(TEAM_ALLIANCE))
         {
             m_SetDeleteThis = true;
+            SetStatus(STATUS_WAIT_LEAVE);
```
sudlud, line 276: Is this a general fix (so probably separate pr maybe) or just relevant when for the new cluster handling?

src/server/game/Entities/Item/Item.cpp
```diff
@@ -381,7 +381,7 @@ void Item::SaveToDB(CharacterDatabaseTransaction trans)
                 if ((uState == ITEM_CHANGED) && IsWrapped())
                 {
                     stmt = CharacterDatabase.GetPreparedStatement(CHAR_UPD_GIFT_OWNER);
-                    stmt->SetData(0, GetOwnerGUID().GetCounter());
+                    stmt->SetData(0, GetOwnerGUID().GetRawValue());
```
sudlud, line 384: Why is this necessary?

src/server/game/Battlegrounds/Battleground.cpp
walkline: It's relevant for clustering. Since we're sending battleground status to the matchmaking service with clustering, it's important to notify the matchmaking service in this case as well. For non-cluster setup, this change has no impact.

src/server/game/Server/Protocol/Opcodes.h
walkline: These are new opcodes that current wow 3.3.5 clients CAN'T generate. These packets are generated by the ToCloud9 gateways. I'm not planning to add new opcodes, quite the opposite, actually. I plan to get rid of these packets and move them to gRPC protocol. But that's something for the future.

walkline:
With the latest changes, the friends list and channels have been implemented. The list of remaining items to be implemented is not that big anymore.

walkline:
I will address the merge conflicts and some code review comments a little later.

There are also some updates: it now supports Windows, and we have the auction house implementation.

src/server/game/Entities/Player/PlayerStorage.cpp
```diff
@@ -5018,9 +5018,9 @@ bool Player::LoadFromDB(ObjectGuid playerGuid, CharacterDatabaseQueryHolder cons
         return false;
     }
 
-    ObjectGuid::LowType guid = playerGuid.GetCounter();
+    uint64 guid = playerGuid.GetRawValue();
 
-    Object::_Create(guid, 0, HighGuid::Player);
+    Object::_Create(playerGuid);
```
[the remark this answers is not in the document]
walkline: You are missing the cross-realm cases, where we have a cross-realm MySQL reverse proxy that routes queries to realm databases based on the player's low GUID. In non-cross-realm cases, we shouldn't have any issues, as it would be hard to overflow the player's counter, and the high GUID is always 0 for players.

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -81,146 +81,146 @@ bool LoginQueryHolder::Initialize()
...
     res &= SetPreparedQuery(PLAYER_LOGIN_QUERY_LOAD_CHARACTER_SETTINGS, stmt);
 
     stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_CHAR_PETS);
-    stmt->SetData(0, lowGuid);
+    stmt->SetData(0, rawGUID);
     res &= SetPreparedQuery(PLAYER_LOGIN_QUERY_LOAD_PET_SLOTS, stmt);
 
     stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_CHAR_ACHIEVEMENT_OFFLINE_UPDATES);
-    stmt->SetData(0, lowGuid);
+    stmt->SetData(0, rawGUID);
```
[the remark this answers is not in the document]
walkline: You are missing the cross-realm cases, where we have a cross-realm MySQL reverse proxy that routes queries to realm databases based on the player's low GUID. In non-cross-realm cases, we shouldn't have any issues, as it would be hard to overflow the player's counter, and the high GUID is always 0 for players.

src/server/game/Server/WorldSocket.cpp
```diff
@@ -600,105 +601,107 @@ void WorldSocket::HandleAuthSessionCallback(std::shared_ptr<ClientAuthSession> a
...
 
-        auto* stmt = LoginDatabase.GetPreparedStatement(LOGIN_UPD_MUTE_TIME_LOGIN);
-        stmt->SetData(0, account.MuteTime);
-        stmt->SetData(1, account.Id);
+        // Update the last_ip in the database as it was successful for login
+        stmt = LoginDatabase.GetPreparedStatement(LOGIN_UPD_LAST_IP);
+        stmt->SetData(0, address);
+        stmt->SetData(1, authSession->Account);
         LoginDatabase.Execute(stmt);
     }
```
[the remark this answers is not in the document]
walkline: As intended. In cluster mode there will be a gateway service in front of this server.

src/server/game/Battlegrounds/Battleground.cpp
```diff
@@ -1089,6 +1091,11 @@ void Battleground::RemovePlayerAtLeave(Player* player)
     // if the player was a match participant
     if (participant)
     {
+        if (sToCloud9Sidecar->ClusterModeEnabled())
+            sToCloud9Sidecar->OnPlayerLeftBattleground(player->GetGUID().GetCounter(),
+                                                       player->GetGUID().GetRealmID(),
+                                                       GetInstanceID());
```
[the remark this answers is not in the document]
walkline: You are missing the cross-realm cases, where we have a cross-realm MySQL reverse proxy that routes queries to realm databases based on the player's low GUID. In non-cross-realm cases, we shouldn't have any issues, as it would be hard to overflow the player's counter, and the high GUID is always 0 for players.

walkline:
Hopefully, all the feedback has been addressed. Let me know if anything is missing.

src/server/game/Maps/Map.cpp
```diff
@@ -1672,7 +1674,7 @@ void Map::SendInitTransports(Player* player)
     // Hack to send out transports
     UpdateData transData;
     for (TransportsContainer::const_iterator itr = _transports.begin(); itr != _transports.end(); ++itr)
-        if (*itr != player->GetTransport())
+        if (*itr != player->GetTransport() && player->InSamePhase(*itr))
```
sudlud: `InSamePhase` here isn't cluster-gated. On a default server the transport stays at `PHASEMASK_NORMAL` (the `SetPhaseMask(2, ...)` in `CreateMoTrans` is behind `Cluster.Enabled`), so any player under a quest phase (Ebon Hold, Wrathgate, Quel'Danas, the Northrend hubs) gets no create-block for boats or zeppelins.

`SendInitTransports` only runs from `AddPlayerToMap` and on login, never on phase change, so they stay invisible at least until the player changes map. Was this meant to be inside the cluster branch?

src/server/game/Handlers/ChannelHandler.cpp
```diff
@@ -38,6 +39,26 @@ void WorldSession::HandleJoinChannel(WorldPacket& recvPacket)
...
         if (!zone || !GetPlayer()->CanJoinConstantChannelInZone(channel, zone))
             return;
+
+        if (ChatChannelsEntry const* channel = sChatChannelsStore.LookupEntry(channelId))
+        {
+            auto const locale = GetSessionDbcLocale();
+            std::string const& zoneName = zone->area_name[locale];
+            char const* nameExt = nullptr;
+            if (channel->flags & CHANNEL_DBC_FLAG_CITY_ONLY)
+                nameExt = sObjectMgr->GetAcoreStringForDBCLocale(LANG_CHANNEL_CITY).c_str();
```
sudlud: `GetAcoreStringForDBCLocale` returns `std::string` by value, so `nameExt` dangles before `snprintf` reads it. `PlayerUpdates.cpp:551-554` does the same thing correctly by keeping the string in a local.

This block also isn't cluster-gated: it replaces the client-sent `channelName` for every constant channel. On a locale mismatch the player joins a different `Channel` than they asked for, and `HandleLeaveChannel` still uses the client name so it can't find it again. If the pattern string is empty, `channelName` ends up empty and the join is dropped with no feedback.

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -699,7 +700,20 @@ void WorldSession::HandlePlayerLoginOpcode(WorldPacket& recvData)
     ObjectGuid playerGuid;
     recvData >> playerGuid;
 
-    if (!IsLegitCharacterForAccount(playerGuid))
+    if (PlayerLoading() || GetPlayer() != nullptr || !playerGuid.IsPlayer())
```
sudlud: This block is unreachable. Line 687 already returns when `RealmLoginEnabled` is off, and line 693 already kicks on `PlayerLoading() || GetPlayer()`, so the only way in is `!playerGuid.IsPlayer()`, and the inner config check is then always false. Drop it.

src/server/game/Battlefield/Zones/BattlefieldWG.cpp
```diff
@@ -243,6 +243,20 @@ bool BattlefieldWG::SetupBattlefield()
 
 bool BattlefieldWG::Update(uint32 diff)
 {
+    // TODO: solve this merge conflict.
```
sudlud: Commented-out block with an unresolved merge TODO. `m_saveTimer` doesn't exist in this class and master's `Update` is just the base call, so nothing is being restored. Please remove.

src/server/game/Maps/TransportMgr.cpp
```diff
@@ -412,6 +415,10 @@ MotionTransport* TransportMgr::CreateTransport(uint32 entry, ObjectGuid::LowType
     if (map && map->IsDungeon())
         trans->m_zoneScript = map->ToInstanceMap()->GetInstanceScript();
 
+    // Merge assumption that it was removed.
```
sudlud: Master has no `setActive` call here either, so this comment documents a removal that never happened. Please remove.

src/server/game/Server/WorldSession.cpp
```diff
@@ -1563,21 +1572,69 @@ void WorldSession::InitializeSessionCallback(CharacterDatabaseQueryHolder const&
...
+    player->SaveToDB(trans, false, true);
+    AddTransactionCallback(CharacterDatabase.AsyncCommitTransaction(trans)).AfterComplete([this](bool success)
+    {
+        WorldPacket data(TC9_SMSG_READY_FOR_REDIRECT, 1);
+        data << uint8(!success); // 0 - Success, 1 - Failed.
+        SendPacket(&data);
+
+        if (!success)
+        {
+            LOG_ERROR("network", "Failed to save player, AccountId = %d", GetAccountId());
```
sudlud: `%d` in an fmt logger: the account id is dropped and the literal `%d` is printed. Use `{}`. Same on the `LOG_DEBUG` below.

src/server/game/Entities/Object/ObjectGuid.h
```diff
@@ -20,6 +20,7 @@
 
 #include "ByteBuffer.h"
 #include "Define.h"
+#include "TC9Sidecar.h"
```
sudlud: This pulls `<future>` and the unscoped macros `MAX_MAP_ID` / `DEFAULT_NON_CROSSREALM_REALM_ID` into nearly every TU. Could the cluster branch move out of the inline `Generate()` so this header stays clean? `MAX_MAP_ID` as a bare macro is also a collision risk for modules.

src/server/game/Entities/Object/ObjectGuid.h
```diff
@@ -283,7 +285,7 @@ class ObjectGuidGeneratorBase
     ObjectGuidGeneratorBase(ObjectGuid::LowType start = 1) : _nextGuid(start) { }
 
     virtual void Set(ObjectGuid::LowType val) { _nextGuid = val; }
-    virtual ObjectGuid::LowType Generate() = 0;
+    virtual ObjectGuid::LowType Generate(uint16 realmId = DEFAULT_NON_CROSSREALM_REALM_ID) = 0;
```
sudlud, line 290: Adding a defaulted parameter to a pure virtual breaks any out-of-tree module overriding `Generate()`. Worth flagging for module authors.

src/server/game/TC9Sidecar/TC9Sidecar.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+bool ToCloud9Sidecar::IsMapAssigned(uint32 mapId)
+{
+    if (mapId >= MAX_MAP_ID)
+        return false;
+
+    return _assignedMapsByID[mapId];
+}
+
+uint32 ToCloud9Sidecar::GenerateCharacterGuid(uint16 realmId)
```
sudlud, line 151: `uint32(...)` truncates the sidecar's `uint64`, and `Generate()` returns `LowType` (uint32), so bits 32-47 are always zero for locally generated GUIDs.

`GetRealmID()` reads exactly those bits, so it returns 0 for every character created here, and 0 for items unconditionally since it's `IsPlayer()`-gated. `Item::CreateItem` and `Battleground::RemovePlayerAtLeave` both consume it. Is character/item creation expected to be gateway-side only?

deps/libsidecar/include/libsidecar/tc9_api.h
```diff
@@ -0,0 +1,353 @@
+#ifndef TC9_API_H
```
sudlud: These duplicate the flat headers with incompatible signatures (`int TC9InitLib(..., bool, const char*)` vs `void TC9InitLib(..., uint8_t, char*)`; `TC9GetNextAvailableCharacterGuid(uint32_t)` vs `(int)`; `TC9StartBattlegroundHandler` vs `BattlegroundStartHandler`). Nothing includes them and the stub implements only the flat set. Remove?

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -454,6 +457,122 @@ void InstanceSaveMgr::LoadCharacterBinds()
...
+            uint32 mapId = fields[1].Get<uint16>();
+            time_t resettime = time_t(fields[2].Get<uint32>());
+            uint8 difficulty = fields[3].Get<uint8>();
+            uint32 completedEncounters = fields[4].Get<uint32>();
+            std::string instanceData = fields[5].Get<std::string>();
+
+            time_t extendedResetTime = 0;
+
+            MapEntry const* entry = sMapStore.LookupEntry(mapId);
+            if (entry->IsRaid() || difficulty > DUNGEON_DIFFICULTY_NORMAL)
```
sudlud: No null check. `AddInstanceSave` guards this, and a stale `instance` row naming a map absent from the DBC crashes here. This path also skips the difficulty validation and `sMapMgr->RegisterInstanceId` that `LoadInstanceSaves` does.

src/server/game/Instances/InstanceSaveMgr.cpp
```diff
@@ -454,6 +457,122 @@ void InstanceSaveMgr::LoadCharacterBinds()
     lock_instLists = false;
 }
 
+void InstanceSaveMgr::LoadInstanceSavesAndBindsForMapIDs(std::vector<uint32> mapIDs, InstanceSaveHashMap &instanceSaveStorage, PlayerBindStorage &localPlayerBindStorage)
```
sudlud: This runs on a `std::async` worker via `OnMapsReassigned`. `GetExtendedResetTimeFor` reads `m_resetExtendedTimeByMapDifficulty` while the world thread writes it through `_ResetOrWarnAll`, and the function also issues sync DB queries and calls `save->RemovePlayer` off-thread. Any reason not to load on the world thread, or snapshot what it needs first?

src/server/game/TC9Sidecar/TC9GroupHooks.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "TC9GroupHooks.h"
+#include "Group.h"
+#include "GroupMgr.h"
+#include "Log.h"
+
+void ToCloud9GroupHooks::OnGroupCreated(EventObjectGroup *group)
+{
+    LOG_INFO("server", "Group created. ID: {}; Leader: {}.", group->guid, group->leader);
+
+    Group* g = new Group();
```
sudlud, line 32: `new Group()` skips what `Group::Create` sets: `m_leaderName`, `m_looterGuid`, and `_initRaidSubGroupsCounter()`, so `m_subGroupsCounts` stays null even for `GROUPTYPE_RAID` and subgroup capacity goes unenforced. `AddMemberWithGuid` also leaves `member.name` empty.

`AddGroup` only writes `GroupStore[counter]`; `RegisterGroupId` is what marks the id used in `_groupIds`, so `GenerateGroupId` can later hand out the same id and `Group::Create` overwrites this entry.

Neither this nor `AddMemberWithGuid` dedupes, so a duplicate or replayed event leaks a Group or duplicates a MemberSlot.

`EventObjectGroup::looterGuid` is never read anywhere.

src/server/apps/worldserver/Main.cpp
```diff
@@ -397,11 +399,15 @@ int main(int argc, char** argv)
         cliThread.reset(new std::thread(CliThread), &ShutdownCLIThread);
     }
 
+    sToCloud9Sidecar->Init(worldPort, realm.Id.Realm);
```
sudlud, line 403: `Init()` runs after `SetInitialWorldSettings()`, so `ClusterModeEnabled()` is false during all map, transport and instance-save loading. `StartDB()` and `MotionTransport::CreateMoTrans` already work around this by reading the config directly. Could `Init()` move before world init so the accessor is usable everywhere?

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
```diff
@@ -0,0 +1,354 @@
...
+    Player *player = ObjectAccessor::FindPlayer(ObjectGuid(playerGuid));
+    if (!player)
+    {
+        GetPlayerItemsByGuidsResponse resp;
+        resp.errorCode = PlayerItemErrorCodePlayerNotFound;
+        return resp;
+    }
+
+    int itemsFound = 0;
+    std::unique_ptr<Item* []> foundItems(new Item * [itemsLen]);
```
sudlud, line 45: `itemsLen` is an unchecked signed `int` from the caller: negative throws `bad_array_new_length`, and the `items[i]` reads are unbounded. Behind the gateway so not player-reachable, but worth a guard.

src/server/game/Chat/Chat.cpp
```diff
@@ -270,18 +270,19 @@ bool ChatHandler::ParseCommands(std::string_view text)
     return _ParseCommands(text.substr(1));
 }
 
-std::size_t ChatHandler::BuildChatPacket(WorldPacket& data, ChatMsg chatType, Language language, ObjectGuid senderGUID, ObjectGuid receiverGUID, std::string_view message, uint8 chatTag,
-                                    std::string const& senderName /*= ""*/, std::string const& receiverName /*= ""*/,
-                                    uint32 achievementId /*= 0*/, bool gmMessage /*= false*/, std::string const& channelName /*= ""*/)
-{
+std::size_t ChatHandler::BuildChatPacket(WorldPacket &data, ChatMsg chatType, Language language, ObjectGuid senderGUID,
```
sudlud: Unrelated reformat: K&R braces and `WorldPacket &data` both go against the codestyle (Allman, `Type&`). Could this hunk be dropped?

deps/libsidecar/stub/libsidecar.c
```diff
@@ -0,0 +1,138 @@
...
+void TC9InitLib(uint16_t port, uint32_t realmID, uint8_t isCrossRealm, char* availableMaps, uint32_t** assignedMaps, int* assignedMapsSize) { panicWithTC9Unavailable("TC9InitLib"); }
+
+// TC9GracefulShutdown gracefully stops all running services.
+//
+void TC9GracefulShutdown() { panicWithTC9Unavailable("TC9GracefulShutdown"); }
+
+// TC9ReadyToAcceptPlayersFromMaps notifies servers registry that this server
+// loaded maps related data and ready to accept players from those maps.
+//
+void TC9ReadyToAcceptPlayersFromMaps(uint32_t* maps, int mapsLen) { panicWithTC9Unavailable("TC9GracefulShutdown"); }
```
sudlud: `TC9ReadyToAcceptPlayersFromMaps` panics with `"TC9GracefulShutdown"`, and `TC9SetMonitoringDataCollectorHandler` with `"TC9SetOnMapsReassignedHook"`.

src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -30,8 +31,27 @@
 #include "Spell.h"
 #include "Vehicle.h"
 #include "WorldModel.h"
+#include <chrono>
 
-MotionTransport::MotionTransport() : Transport(), _transportInfo(nullptr), _isMoving(true), _pendingStop(false), _triggeredArrivalEvent(false), _triggeredDepartureEvent(false), _passengersLoaded(false), _delayedTeleport(false)
+namespace
+{
+    // Any date would work. Lets use this one - August 10, 2023, 00:00:00.
+    std::time_t startTimestamp = 1689139200;
```
sudlud: `1689139200` is 2023-07-12 05:20 UTC, not August 10 midnight. Also `tarnsportStartDate` typo, and both globals could be `const`.

src/server/game/Entities/Item/Item.cpp
```diff
@@ -1096,29 +1096,35 @@ Item* Item::CreateItem(uint32 item, uint32 count, Player const* player, bool clo
         return nullptr;                                        //don't create item at zero count
 
     ItemTemplate const* pProto = sObjectMgr->GetItemTemplate(item);
-    if (pProto)
+    if (!pProto)
     {
-        if (count > pProto->GetMaxStackSize())
-            count = pProto->GetMaxStackSize();
+        ABORT();
```
sudlud, line 1101: `ABORT()` is `[[noreturn]]`, so this `return nullptr` is unreachable.

deps/libsidecar/include/player-money-api.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum PlayerMoneyErrorCode {
+    PlayerMoneyErrorCodeNoError        = 0,
+    PlayerMoneyErrorCodeNoHandler      = 1,
+    PlayerMoneyErrorCodePlayerNotFound = 2,
+    PlayerMoneyErrorCodeToMuchMoney    = 3,
```
sudlud: `ToMuchMoney` -> `TooMuchMoney`. ABI-visible, so it needs the Go side too.

src/server/game/TC9Sidecar/TC9GroupHooks.h
```diff
@@ -0,0 +1,30 @@
+//
+//  TC9GroupHooks.h
+//  game
+//
+//  Created by Anton Popovichenko on 26/08/2023.
```
sudlud: Missing the AGPL header the other files under `TC9Sidecar/` carry. Same for `TC9GroupHooks.cpp`.

deps/libsidecar/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+              true)
+
+        set_target_properties(libsidecar PROPERTIES
+            IMPORTED_LOCATION_NOCONFIG
+              "${CMAKE_INSTALL_RPATH}/libsidecar.${LIBSIDECAR_EXTENSION}"
+            IMPORTED_SONAME_NOCONFIG
+              "libsidecar.${LIBSIDECAR_EXTENSION}")
+    endif()
+
+endif()
```
sudlud: No newline at end of file. Same in `stub/battleground-api.c`.

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -4893,3 +4895,50 @@ Respawn.ForceCompatibilityMode = 0
...
+#                 1 - (Enabled)
+
+Cluster.IsCrossrealm=0
+
+#
+###################################################################################################
+
+###################################################################################################
+#                                                                                                 #
+# CLUSTER SETTINGS END                                                                            #
```
sudlud, line 4948: No newline at end of file.

sudlud (commented):
I tried to give this a thorough review with Claude.

Heads up on CI: both clang-20 jobs fail on `ChannelHandler.cpp:49` with `-Wdangling-assignment-gsl` (the `nameExt` comment above). The other red jobs are fail-fast cancellations of that one, not separate problems. Older compilers accept the line, which is why it only shows up on the ubuntu-26.04 runners.

Cluster mode disabling digest verification, IP/country lock, ban check and the security-level gate is understood as intentional with a gateway in front, but the `Cluster.Enabled` config block should say that plainly. As written, an operator exposing the worldserver port has no way to know from the config that ban enforcement is off.

Kitzunu's 2023 point still applies: the Chat.cpp reformat and the channel-name rewrite are independent of clustering and would land much faster as their own small PRs. The transport phase filter, the dead login block and the two commented-out blocks look like leftovers of this branch rather than intended changes, and can simply be dropped here.

sudlud:
@coderabbitai review

src/server/game/Server/WorldSession.cpp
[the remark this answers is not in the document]
sudlud: Fixed in 9ebe483 — both calls now use `{}`.

src/server/game/TC9Sidecar/TC9GroupHooks.h
```diff
@@ -0,0 +1,30 @@
...
+//
+
```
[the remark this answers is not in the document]
sudlud: Fixed in c774235 — replaced the banner with the standard AGPL header on both the `.cpp` and `.h`.

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -81,145 +81,146 @@ bool LoginQueryHolder::Initialize()
     SetSize(MAX_PLAYER_LOGIN_QUERY);
 
     bool res = true;
-    ObjectGuid::LowType lowGuid = m_guid.GetCounter();
+    uint64 rawGUID = m_guid.GetRawValue();
 
     CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_CHARACTER);
-    stmt->SetData(0, lowGuid);
+    stmt->SetData(0, rawGUID);
     res &= SetPreparedQuery(PLAYER_LOGIN_QUERY_LOAD_FROM, stmt);
```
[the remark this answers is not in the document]
sudlud: For a normal player GUID the high bits are zero, so `GetRawValue()` equals the counter and binds losslessly to the 32-bit column. The raw GUID is intentional for cross-realm, where the MySQL reverse proxy routes on it — the same point walkline made on the earlier CodeRabbit finding, which was withdrawn.

src/server/game/Entities/Player/PlayerStorage.cpp
```diff
@@ -4977,7 +4977,7 @@ void Player::SetHomebind(WorldLocation const& loc, uint32 areaId)
     stmt->SetData (2, m_homebindX);
     stmt->SetData (3, m_homebindY);
     stmt->SetData (4, m_homebindZ);
-    stmt->SetData(5, GetGUID().GetCounter());
+    stmt->SetData(5, GetGUID().GetRawValue());
     CharacterDatabase.Execute(stmt);
```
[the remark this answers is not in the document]
sudlud: For a normal player GUID the high bits are zero, so `GetRawValue()` equals the counter and binds losslessly to the 32-bit column. The raw GUID is intentional for cross-realm, where the MySQL reverse proxy routes on it — the same point walkline made on the earlier CodeRabbit finding, which was withdrawn.

src/server/game/Entities/Player/Player.cpp
```diff
@@ -3783,7 +3783,7 @@ void Player::_LoadSpellCooldowns(PreparedQueryResult result)
 void Player::_SaveSpellCooldowns(CharacterDatabaseTransaction trans, bool logout)
 {
     CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_CHAR_SPELL_COOLDOWN);
-    stmt->SetData(0, GetGUID().GetCounter());
+    stmt->SetData(0, GetGUID().GetRawValue());
     trans->Append(stmt);
```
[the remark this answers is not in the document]
sudlud: For a normal player GUID the high bits are zero, so `GetRawValue()` equals the counter and binds losslessly to the 32-bit column. The raw GUID is intentional for cross-realm, where the MySQL reverse proxy routes on it — the same point walkline made on the earlier CodeRabbit finding, which was withdrawn.

src/server/game/Entities/Item/Item.cpp
```diff
@@ -348,7 +348,7 @@ void Item::SaveToDB(CharacterDatabaseTransaction trans)
                 uint8 index = 0;
                 CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(uState == ITEM_NEW ? CHAR_REP_ITEM_INSTANCE : CHAR_UPD_ITEM_INSTANCE);
                 stmt->SetData(  index, GetEntry());
-                stmt->SetData(++index, GetOwnerGUID().GetCounter());
+                stmt->SetData(++index, GetOwnerGUID().GetRawValue());
                 stmt->SetData(++index, GetGuidValue(ITEM_FIELD_CREATOR).GetCounter());
                 stmt->SetData(++index, GetGuidValue(ITEM_FIELD_GIFTCREATOR).GetCounter());
```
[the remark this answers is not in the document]
sudlud: For a normal player GUID the high bits are zero, so `GetRawValue()` equals the counter and binds losslessly to the 32-bit column. The raw GUID is intentional for cross-realm, where the MySQL reverse proxy routes on it — the same point walkline made on the earlier CodeRabbit finding, which was withdrawn.

src/server/game/TC9Sidecar/AsyncTask.h
```diff
@@ -0,0 +1,79 @@
...
+
+    void ExecuteAsync()
+    {
+        // Capture the function by value so a moved AsyncTask does not leave
+        // the in-flight async holding a dangling this pointer.
+        AsyncFunction fn = asyncFunc;
+        asyncTask = std::async(std::launch::async, [fn = std::move(fn)]() mutable
+        {
+            return fn();
+        });
```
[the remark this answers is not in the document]
sudlud: This compiles: `std::async` returns `future<T>`, which converts to `shared_future<T>` via its converting move constructor, so the assignment is well-formed. The PR already builds on the gcc/clang-18 jobs, confirming it.

src/server/game/TC9Sidecar/AsyncTask.h
```diff
@@ -0,0 +1,79 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _ASYNC_TASK_H
+#define _ASYNC_TASK_H
+
+#include <functional>
+#include <future>
+
```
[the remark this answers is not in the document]
sudlud: Fixed in 53eb19f — `<chrono>` is now included directly.

src/server/game/Handlers/ChatHandler.cpp
```diff
@@ -713,6 +713,18 @@ namespace Acore
 
 void WorldSession::HandleTextEmoteOpcode(WorldPacket& recvData)
 {
+    uint32 text_emote;
+    recvData >> text_emote;
+
+    constexpr uint32 readyEmote = 126;
+
+    // Handle confirmation redirect when the player types "/read" after the new node becomes available for redirection.
```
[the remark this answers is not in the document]
sudlud: Fixed in bd73e58 — the comment now says /ready.

src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -30,8 +31,27 @@
...
+
+    // Calculates time of the next departure cycle.
+    std::chrono::system_clock::time_point calculateNextDepartureTime(int oneIterationInterval) {
+        std::chrono::system_clock::time_point currentTime = std::chrono::system_clock::now();
+        std::chrono::milliseconds interval(oneIterationInterval);
+        std::chrono::milliseconds timeSinceStart = std::chrono::duration_cast<std::chrono::milliseconds>(currentTime - tarnsportStartDate);
+        int64 intervalsPassed = timeSinceStart.count() / oneIterationInterval;
+        std::chrono::system_clock::time_point nextDeparture = tarnsportStartDate + (interval * (intervalsPassed + 1));
+        return nextDeparture;
+    }
```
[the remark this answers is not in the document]
sudlud: Fixed in 3a47064 (Allman braces); the typo was corrected earlier in 251b5ab.

deps/libsidecar/include/player-money-api.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+typedef enum PlayerMoneyErrorCode {
+    PlayerMoneyErrorCodeNoError        = 0,
+    PlayerMoneyErrorCodeNoHandler      = 1,
+    PlayerMoneyErrorCodePlayerNotFound = 2,
+    PlayerMoneyErrorCodeToMuchMoney    = 3,
```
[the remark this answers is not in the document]
sudlud: Fixed in d1d11b8 — renamed to TooMuchMoney across both headers and its use; enum value 3 unchanged.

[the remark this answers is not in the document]
sudlud: Heads-up: these libsidecar headers are vendored from ToCloud9, and the same typo is in the source there (`game-server/libsidecar/player-money-api.h` and `libsidecar-cpp/include/player-money-api.h`). Worth fixing upstream too, otherwise a re-vendor would revert this and leave `TC9GrpcHandler.cpp` referencing a name that no longer exists.

src/server/game/Maps/Map.cpp
sudlud: Fixed in cfbe214 — the `InSamePhase` filter now applies only in cluster mode, so stock servers keep showing transports to phased players.

src/server/game/Handlers/ChannelHandler.cpp
sudlud: Fixed in 5cdcf26 — bound the city string to a local (fixes the dangling read / clang-20 build error) and gated the whole rewrite behind cluster mode.

src/server/game/Handlers/CharacterHandler.cpp
sudlud: Fixed in adc6582 — removed the unreachable block.

src/server/game/Battlefield/Zones/BattlefieldWG.cpp
sudlud: Fixed in 889d142 — removed the dead block.

src/server/game/Maps/TransportMgr.cpp
sudlud: Fixed in c9a06ae — removed the stale comment.

src/server/game/Entities/Object/ObjectGuid.h
sudlud: Fixed in 06beaa0 — moved the cluster branch out-of-line to ObjectGuid.cpp and dropped the TC9Sidecar.h include from ObjectGuid.h; `DEFAULT_NON_CROSSREALM_REALM_ID` is now a constexpr there.

deps/libsidecar/include/libsidecar/tc9_api.h
sudlud: Fixed in 92c8d1a — removed the unused tc9_* header tree.

src/server/game/Instances/InstanceSaveMgr.cpp
sudlud: Fixed in 505f4fa — the worker now only runs the DB reads and returns plain rows; all manager-state work (InstanceSave construction, reset-time lookups, bind wiring) moved to the world-thread callback.

src/server/game/TC9Sidecar/TC9GroupHooks.cpp
sudlud: Fixed in 7200c59 and eedcd9 — set the missing fields (leader name, looter, raid subgroup counter), dedup on event redelivery, and register the group id.

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
sudlud: Fixed in c5861d6 — guard `itemsLen <= 0`.

src/server/game/Chat/Chat.cpp
sudlud: Fixed in ee9136a — reverted to the upstream formatting.

deps/libsidecar/include/player-money-api.h
sudlud: Fixed in d1d11b8 — renamed to `TooMuchMoney` across the headers and its use.

src/server/game/Server/WorldSession.cpp
```diff
@@ -1563,21 +1572,69 @@ void WorldSession::InitializeSessionCallback(CharacterDatabaseQueryHolder const&
...
+
+        LOG_DEBUG("network", "Saved, AccountId = %d", GetAccountId());
+
+        Player* player = GetPlayer();
+        if (!player)
+            return;
+
+        player->m_Events.AddEventAtOffset([this](){
+            KickPlayer("HandlePrepareForRedirect client redirected");
+        }, 100ms);
```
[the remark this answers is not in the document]
sudlud: Fixed in 65e2178 — the redirect now records a `_redirecting` flag honored inside `LogoutPlayer`, so the friend-offline broadcast and OnPlayerLogout are suppressed during a node handoff.

src/server/game/Battlefield/Zones/BattlefieldWG.cpp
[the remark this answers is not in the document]
sudlud: Fixed in 889d142 — removed the dead block.

src/server/game/Entities/Object/ObjectGuid.h
```diff
@@ -20,6 +20,7 @@
 
 #include "ByteBuffer.h"
 #include "Define.h"
+#include "TC9Sidecar.h"
 #include <deque>
```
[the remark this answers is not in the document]
sudlud: Fixed in 06beaa0 — ObjectGuid.h no longer includes TC9Sidecar.h.

src/server/game/TC9Sidecar/TC9Sidecar.cpp
sudlud: Checked the ToCloud9 source: the guid provider (`game-server/libsidecar/guids/mgr.go`, `CrossrealmMgr.Next` -> `provider.Next()`) returns a plain per-realm counter from an allocated diapason, with no realm bits encoded — realmID only selects the pool. So the `uint32()` cast is lossless and `GetRealmID()` returning 0 for locally-created objects is by design; realm bits are injected at the gateway on login, not at local creation. Looks intended rather than a truncation bug. Leaving open to confirm that gateway detail.

src/server/game/Instances/InstanceScript.cpp
```diff
@@ -41,6 +42,9 @@ BossBoundaryData::~BossBoundaryData()
 
 void InstanceScript::SaveToDB()
 {
+    if (sToCloud9Sidecar->ClusterModeEnabled() && !sToCloud9Sidecar->IsMapAssigned(instance->GetEntry()->MapID))
```
sudlud, line 45: This early return opens a window during map handoff: `OnMapsReassigned` unassigns the map immediately, but players stay up to a minute until the redirect kick, and any encounter progress in that window is silently dropped while the new owner loads the old data. Could the unassign be deferred until the map is drained, or saves kept until the last player left?

sudlud:
Heads-up on a ToCloud9 issue that affects the checked cross-realm BG support: the mysql reverse proxy only routes prepared statements. Plain-text queries go through `TC9Proxy.HandleQuery`, which sends them to `ConnByRealm(0)`, i.e. an arbitrary connection. `Player::_SaveSpellCooldowns` writes its batch INSERT as plain text inside the save transaction, so for a visiting cross-realm player those rows land in the wrong realm DB and outside the routed transaction. Same applies to any other raw-string write. Probably needs plain-query routing in the proxy (or converting such writes to prepared statements).

src/server/apps/worldserver/Main.cpp
walkline: It can't, since registering would mark worldserver as ready to accept players, and we don't want that to happen while worldserver is loading.

src/server/game/TC9Sidecar/TC9Sidecar.cpp
walkline: Yeah, it's intended. We are generating counters only, realm id excluded here.

src/server/game/Entities/Object/ObjectGuid.h
walkline: I doubt any module will override the GUID generator.

sudlud:
A few open issues in the redirect flow:
- `HandleTC9PrepareForRedirect` is registered PROCESS_THREADUNSAFE but is invoked from map worker threads via `Map::UpdatePlayersRedirectKickEvent` and the /ready path (CMSG_TEXT_EMOTE is THREADSAFE). With MapUpdate.Threads > 1, two full `SaveToDB` runs can execute concurrently off the world thread. Related: sidecar item GUID generation (thread unsafe per guids.go) is reachable from worker threads through `Item::CreateItem`.
- `Player::SaveToDB` defers into DELAYED_SAVE_PLAYER while `IsBeingTeleportedFar()`, which is exactly the redirect-on-map-change case, so the READY ack can follow an empty transaction and the new node may load a stale snapshot.
- `TC9_CMSG_PREPARE_FOR_REDIRECT` is STATUS_AUTHED and client-reachable, so a modified client can spam full save transactions.

sudlud:
Same root as the plain-query issue above: the proxy's `charGuidColumnToTableMap` is missing `character_stats`, `character_gifts`, `item_refund_instance`, `character_declinedname` and `character_achievement_offline_updates`, but the core binds `GetRawValue()` for statements on those tables. On a crossrealm node the SELECTs match nothing and the INSERTs overflow the 32-bit columns, which aborts the save transaction in strict mode. The cluster instance-save loader also uses plain-text `Query`, which the proxy cannot route.

sudlud:
Found while cross-checking ToCloud9: `game-server/libsidecar/events-group.go:91-94` advances the member array pointer cumulatively (by `i` relative to the already-advanced pointer), so `OnGroupCreated` delivers garbage member GUIDs for groups with 3 or more members. Worth fixing upstream; on the AC side `AddMemberWithGuid` then materializes phantom members from character-cache misses.

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
```diff
@@ -0,0 +1,374 @@
...
+        item.guid = pItem->GetGUID().GetRawValue();
+        item.entry = pItem->GetEntry();
+        item.owner = playerGuid;
+        item.bagSlot = pItem->GetBagSlot();
+        item.slot = pItem->GetSlot();
+        item.isTradable = pItem->CanBeTraded(true);
+        item.count = pItem->GetCount();
+        item.flags = pItem->GetUInt32Value(ITEM_FIELD_FLAGS);
+        item.durability = pItem->GetUInt32Value(ITEM_FIELD_DURABILITY);
+        item.randomPropertyID = pItem->GetItemRandomPropertyId();
```
sudlud, line 73: `PlayerItem.randomPropertyID` is `int8_t` in the contract (here and in ToCloud9), but `GetItemRandomPropertyId()` is int32 with suffix IDs negative in the thousands, so any "of the ..." item fetched cross-node gets a corrupted enchant id. `durability` as `uint8_t` truncates above 255 too. Needs a contract change on both sides.

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (!bg)
+        return BattlegroundErrorBattlegroundNotFound;
+
+    for (int i = 0; i < request->alliancePlayersToAddSize; i++)
+    {
+        Player *player = ObjectAccessor::FindPlayer(ObjectGuid(request->alliancePlayersToAdd[i]));
+        if (player)
+        {
+            player->SetEntryPoint();
+            player->SetBattlegroundId(bg->GetInstanceID(), bg->GetBgTypeID(), 1, true, bgTypeId == BATTLEGROUND_RB, player->GetTeamId(true));
```
sudlud, line 323: queueSlot is hardcoded to 1 here and below; for players whose actual slot is 0 the later status packets address the wrong client queue frame.

src/server/game/Instances/InstanceScript.cpp
walkline: Making the transition completely smooth is very problematic.
Here’s a concrete example of why: Imagine a raid with 19 players fighting a boss. One player disconnects (internet drop, real-life interruption, etc.).
Because the number of worldservers has changed, the system triggers a map redistribution and reassigns this instance to a different worldserver.
Shortly after the redistribution, the disconnected player logs back in.
For him we have to load the instance in the state it had at the moment of redistribution (including the still-alive boss and all encounter data).
Meanwhile, the other 19 players continue the fight, kill the boss, loot it, and then get moved to the new worldserver.
When they arrive, the boss is still alive on that worldserver - because the instance was loaded with the older state for the first player who joined.
At least they keep the loot they already received.
To make this smoother we would ideally wait until all players have transitioned, then reload the instance and reconcile the state differences.
However, the core probably doesn’t support reloading an already-loaded instance, and even if it did, the experience would still be jarring for the player who arrived first (the boss would appear and then suddenly disappear).
Given all of that, I would leave the current behaviour as-is.
This situation should be rare. Players are notified about the upcoming transition, so they can choose to wait or transition manually with /ready. It’s inconvenient, but it shouldn’t be the end of the world.

src/server/game/Server/WorldSession.cpp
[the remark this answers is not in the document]
sudlud: Reverted in dbfeb1d45d. Both flag placements turned out wrong: set synchronously it can go stale on an abandoned handoff, set in the deferred kick it misses the gateway closing the socket right after READY (interceptors.go:249), which tears the session down before the kick fires. Suppressing the logout side effects probably needs to be part of the redirect handshake itself rather than session state guessed on the worldserver.

sudlud (approved):

walkline:
> HandleTC9PrepareForRedirect is registered PROCESS_THREADUNSAFE but is invoked from map worker threads via Map::UpdatePlayersRedirectKickEvent and the /ready path (CMSG_TEXT_EMOTE is THREADSAFE). With MapUpdate.Threads > 1, two full SaveToDB runs can execute concurrently off the world thread.

I marked it as PROCESS_THREADUNSAFE, probably being overly defensive. It can likely be marked as thread-safe, since I don’t see any unsafe operations there. I also don’t see any situations where SaveToDB could be executed concurrently for the same player.

> Related: sidecar item GUID generation (thread unsafe per guids.go) is reachable from worker threads through Item::CreateItem.

Item GUID generation should be safe in C++ implementation of libsidecar (successor of Go version). But original AC has a race condition there 😅


> Player::SaveToDB defers into DELAYED_SAVE_PLAYER while IsBeingTeleportedFar(), which is exactly the redirect-on-map-change case, so the READY ack can follow an empty transaction and the new node may load a stale snapshot.
TC9_CMSG_PREPARE_FOR_REDIRECT is STATUS_AUTHED and client-reachable, so a modified client can spam full save transactions.

These are valid points. We should probably create issues in the TC9 repository so they aren’t lost.

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
walkline: Valid point

src/server/game/TC9Sidecar/TC9GrpcHandler.cpp
walkline: I think this is mostly a binding action, queue packets are handled by the gateway and matchmaking services. So I would consider it a minor thing.

walkline:
> Heads-up on a ToCloud9 issue that affects the checked cross-realm BG support: the mysql reverse proxy only routes prepared statements. Plain-text queries go through TC9Proxy.HandleQuery, which sends them to ConnByRealm(0), i.e. an arbitrary connection. Player::_SaveSpellCooldowns writes its batch INSERT as plain text inside the save transaction, so for a visiting cross-realm player those rows land in the wrong realm DB and outside the routed transaction. Same applies to any other raw-string write. Probably needs plain-query routing in the proxy (or converting such writes to prepared statements).

> Same root as the plain-query issue above: the proxy's charGuidColumnToTableMap is missing character_stats, character_gifts, item_refund_instance, character_declinedname and character_achievement_offline_updates, but the core binds GetRawValue() for statements on those tables. On a crossrealm node the SELECTs match nothing and the INSERTs overflow the 32-bit columns, which aborts the save transaction in strict mode. The cluster instance-save loader also uses plain-text Query, which the proxy cannot route.

Yeah, mysql reverse proxy for cross-realm is still experimental. We need to carefully review all possible queries before using it anywhere except for testing.

Kitzunu:
Will merge on Monday for the 3 year celebration

Kitzunu:
3 year development anniversary today 🥳 
Well done! Happy it is merged :)

sudlud:
🎉🎉🎉🎉🎉🎉

Issue #16801, opened 2023-07-20
In the icecrown citadel, the 1000G that should be paid for recovering the ring have not been deducted.


dayn9t:
### Current Behaviour

Creature: 38316, 'Ormus the Penitent', 'Death Knight Armor'
Item: 50402,  'Ashen Band of Endless Vengeance'
Gossip menu options: 
- 10996, 6, 0, 'I appear to have lost my ring.'
- 10995, 0, 0, 'It won\'t.', 0, 1, 1, 0, 0, 0, 10000000, 'A contribution is required.'

azerothcore-wotlk/src/server/game/Entities/Player/PlayerGossip.cpp
```
251    GossipMenuItem const* item = gossipMenu.GetItem(gossipListId);
252    if (!item)  // FIXME: Menu item not found
253        return;
...
392    ModifyMoney(-cost); // Unreachable
```

### Expected Blizzlike Behaviour

...

### Steps to reproduce the problem

...

### AC rev. hash/commit

commit 30197f40d687eae97a07d6730b6933c63f1a3983 (HEAD -> master, origin/master, origin/HEAD)
Author: AzerothCoreBot <azerothcorebot@gmail.com>
Date:   Wed Jul 19 05:57:13 2023 +0000

### Operating system

Ubuntu 22.04

Kitzunu:
He shouldnt provide such service anyway. Probably leftover stuff from sunwell...

Kitzunu:
I stand corrected.

If the ring has been deleted the option shows and you can recover your ring.
In today's retail it is however free to recover it

Kitzunu:
I cannot reproduce on 60eba10e46ea

PR #16851, opened 2023-07-29
fix(DB/Creature): Ormus the Penitent

Kitzunu:
* closes https://github.com/azerothcore/azerothcore-wotlk/issues/16801

## Changes Proposed:
-  From all the sources I can gather Ormus the Penitent should not offer a way to get the ring back if it is discarded.
-  

## Issues Addressed:

- Closes 

## SOURCE:

* https://gamefaqs.gamespot.com/boards/534914-world-of-warcraft/64078529
    *  > You can switch the ring, but there isn't an option to restore it if it's deleted.
    * > You can use the Item Recovery thing on Battle.net. It tracks stuff you vendor/destroy.
## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1690642821596651700.sql
+++ b/data/sql/updates/pending_db_world/rev_1690642821596651700.sql
@@ -1,0 +1,4 @@
+--
+DELETE FROM `gossip_menu_option` WHERE `MenuID` = 10996 AND `OptionID` = 6;
+DELETE FROM `smart_scripts` WHERE `entryorguid` = 38316;
+UPDATE `creature_template` SET `AIName` = '';


Gultask:
The text **does** exist in broadcast_text, ID 38227
ID 38226 is also used by the same NPC

Gultask:
image

Kitzunu:
Sure, but could it have been left over from cataclysm? Or earlier versions? Because I can't find any claim for it ever existing

Gultask:
These rings might still exist. It's best to check in Retail if recovery is possible through this method.
I never trust comments anymore :(

Kitzunu:
I can check retail

heyitsbench:
> Sure, but could it have been left over from cataclysm? Or earlier versions?

Doesn't mean a whole heck of a lot, but its ID is before other Wrath texts. (~40,400 for example seems to be from the Zalazane's fall event)

Kitzunu:
I stand corrected.

If the ring has been deleted the option shows and you can recover your ring.
In today's retail it is however free to recover it

PR #12707, opened 2022-08-12
fix(DB/Quest): Add Aqual Quintessence questline steps.

heyitsbench:
## Changes Proposed:
-  Adds the questline to acquire the Aqual Quintessence used for dousing fires in Molten Core to Duke Hydraxis with the exception of starting the quest [Poisoned Water](https://www.wowhead.com/wotlk/quest=6804/poisoned-water).

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

Wrath Classic Beta:
- Took a template character to Hydraxis, he offered the [Stormers and Rumblers](https://www.wowhead.com/wotlk/quest=6805/stormers-and-rumblers) quest only. This is what should be the only option for any player unless they're going through a progressive server.
- Took a character from live Burning Crusade Classic that picked up the Poisoned Water quest and Stormers and Rumblers quest, copied them to Wrath Classic Beta, completed the two quests and observed that the next quest in the chain, [Eye of the Emberseer,](https://www.wowhead.com/wotlk/quest=6821/eye-of-the-emberseer) was available to accept.

## Tests Performed:

- Untested.

## How to Test the Changes:

1. Run the in-game commands `.quest add 6804`, `.quest add 6805`, `.quest complete 6804`, `.quest complete 6805`.
2. Go to Duke Hydraxis and continue his quest chain to its final quest, [A Hero's Reward](https://www.wowhead.com/wotlk/quest=7486/a-heros-reward).

## Known Issues and TODO List:

Needs to get tested.
It's unknown how deep the quest chain goes and what parts are still left over from the original chain in Wrath, but given that the UBRS quest is available to pick up in my own testing, as well as the [Wowpedia article](https://wowpedia.fandom.com/wiki/Aqual_Quintessence) pointing out that the quest chain was completable and the achievement was able to be earned, I myself believe this is blizzlike behavior.
This query is taken straight from the [progression system module.](https://github.com/azerothcore/mod-progression-system/blob/main/src/Bracket_60_1_2/sql/world/progression_60_1_2_MC_poisoned_water.sql) If this is incorrect or needs tweaking, please let me know.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

[...]

PR #16938, opened 2023-08-07
fix(DB/Text): creature_text cleanup

wdraenor:
## Changes Proposed:
-  Deletes 2 unnecessary rows of text that aren't used anywhere in the game, and was removed in patch 3.2.2 
-  

## Issues Addressed:

- Closes 

## SOURCE:

https://wowpedia.fandom.com/wiki/Celebrating_Good_Times

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1691367589804532144.sql
+++ b/data/sql/updates/pending_db_world/rev_1691367589804532144.sql
@@ -1,2 +1,1 @@
-DELETE FROM `creature_text` WHERE `TEXT`='Behold the might of the Alliance! The dread lady, Onyxia, hangs from the arches! Let the rallying cry of the dragon slayer lift your spirits!' AND `CreatureID` = 14394;
-DELETE FROM `creature_text` WHERE `TEXT`='Citizens and allies of Stormwind, on this day, history has been made. $n has laid waste to that which had attempted to usurp the rule of the kingdom. Gather round and join me in honoring our heroes.' AND `CreatureID` = 14394;
+DELETE FROM `creature_text` WHERE (`BroadcastTextId`, `CreatureID`) IN ((9496, 14394), (9495, 14394));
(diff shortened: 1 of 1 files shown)


heyitsbench:
Is there a source that with 3.2.2 those that still had the head could no longer see this RP event? If there isn't, I disagree with this change wholeheartedly.

wdraenor:
> Is there a source that with 3.2.2 those that still had the head could no longer see this RP event? If there isn't, I disagree with this change wholeheartedly.

I'm not sure of a source for that, I was able to add the prerequisite quest "Victory for the Alliance" to myself in-game with the head of Onyxia but I wasn't able to turn in the quest, using GM commands to give myself the "Celebrating Good Times" quest also wouldn't let me turn in the quest to Major Mattingly, so you're never able to see this text in-game.

heyitsbench:
> I was able to add the prerequisite quest "Victory for the Alliance" to myself in-game with the head of Onyxia (given to me by GM commands) but I wasn't able to turn in the quest, using GM commands to give myself the "Celebrating Good Times" quest also wouldn't let me turn in the quest to Major Mattingly, so you're never able to see this text in-game.

To be honest, I would sooner call *that* the issue here.
 
> The prerequisite quest "Victory for the Alliance" was also removed in patch 3.2.2 [Wowpedia link]

I wouldn't trust a site like Wowpedia for that sort of thing, they tend to say the quest is removed just if it's inaccessible to new players. The quest technically still exists even in retail today.

I am of the opinion that these quests and their turn-ins should still exist in 3.3.5 if a player from an older patch happened to hold onto the items/quests, as Blizz does have a [bit of a track record with that sort of thing](https://github.com/azerothcore/azerothcore-wotlk/pull/12707). It's obviously not a great source given the circumstances, but the quests and their RP do still exist on Wrath Classic today despite the revised Ony raid also being released. Of course as you've noticed, there is no source proving or disproving this, and I would prefer the quests be corrected to have working turn-ins with base AC, but in the absence of evidence, I definitely don't think stripping out what used to exist is a good idea.

wdraenor:
I see what you mean, you've raised some valid arguments which makes me think this is not a good change, so I will be closing this.

PR #16939, opened 2023-08-07
fix(DB/Text): npc_text clean up

wdraenor:
## Changes Proposed:
-  Removes text from Cataclysm:
- Thrall - "I believe Aggra is almost here, and the other aspects have assembled.$B$BAre you ready to begin, $N?" (BroadcastTextID0 = 52965, ID = 18268)
-  Kilrok Gorehammer - "Yes. We regrouped east of the pass and prepared a counter-attack, but the next day, the skies reddened and the land was cleaved in two, as easily as a child might split open a cactus apple. See how the devastation stretches from horizon to horizon?$B$BOur Alliance foes in their makeshift bunkers were as surprised and disorganized as we were by the disaster. It would've been the perfect time to strike! But Warlord Gar'dul insisted we pull back to the south to refortify.$B$B<Kilrok spits on the ground.>$B$BNot all of us followed. I stood toe-to-toe with demons in the last war, $N. I'm not about to turn tail and let the Alliance plant roots on my soil." (BroadcastTextID0 = 37057, ID = 15137)

## Issues Addressed:

- Closes 

## SOURCE:

https://wowpedia.fandom.com/wiki/The_Nordrassil_Summit#Notes
https://www.wowhead.com/quest=29326/the-nordrassil-summit (added in patch 4.2.0)

https://wowpedia.fandom.com/wiki/Kilrok_Gorehammer#Quotes
https://www.wowhead.com/npc=37154/kilrok-gorehammer (added in patch 4.0.3)

## Tests Performed:

- 
- 

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1691385815761723590.sql
+++ b/data/sql/updates/pending_db_world/rev_1691385815761723590.sql
@@ -1,0 +1,1 @@
+DELETE FROM `npc_text` WHERE (`BroadcastTextID0`, `ID`) IN ((52965, 18268), (37057, 15137));


wdraenor:
After going through npc_text some more, there is a **ton** of texts from Cataclysm. Are these texts being kept for a reason?

PR #16946, opened 2023-08-07
Fix docker compile issues

ben-of-codecraft:
## Changes Proposed:
-  Adds Modules mount to world server for modules that reference the directory, that were broken in docker builds. 
-  Changes the inline sed that would cause a permissions issue and show up as a db-import failure. 

## Issues Addressed:

- Closes 

Build fails with this message out of the box 
```bash
azerothcore-wotlk-with-npcbots-ac-db-import-1 exited with code 1
service "ac-db-import" didn't complete successfully: exit 1
error: Uncaught (in promise) Error: Failed with error: 1, however,
          it's not related to this Deno script directly. An error occurred within
          the script called by the command itself
          throw new Error(`Failed with error: ${status.code}, however,
```

## Tests Performed:

-  It was building with errors, and now it does not
- I have tested on Mac Apple chip Docker Desktop and 2 Ubuntu 20 installs

## How to Test the Changes:

1. Run ./acore.sh docker build 
2. Run ./acore.sh docker start:app
3. 

- [ ]
- [ ]


apps/compiler/includes/functions.sh
```diff
@@ -208,9 +208,8 @@ function conf_layer() {
         if grep -qE "^$KEY" "$BASE" && ! grep -qE "^$KEY.*=.*$VAL" "$BASE"; then
           # Replace line
           # Prevent issues with shell quoting 
-          sed -i \
-            's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' \
-            "$BASE"
+          # Updated replace due to permissions creating config files across platforms.
```
michaeldelago, line 211: comment documenting the change is not needed

apps/compiler/includes/functions.sh
```diff
@@ -208,9 +208,8 @@ function conf_layer() {
         if grep -qE "^$KEY" "$BASE" && ! grep -qE "^$KEY.*=.*$VAL" "$BASE"; then
           # Replace line
           # Prevent issues with shell quoting 
-          sed -i \
-            's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' \
-            "$BASE"
+          # Updated replace due to permissions creating config files across platforms. 
+          sed 's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' "$BASE" > tmp; cat tmp > "$BASE"; rm tmp
```
michaeldelago, line 212: `mktemp` creates a temporary filename that's guaranteed to be unique.

`cat` and redirection works, but they aren't needed here. `mv` makes more sense.

```suggestion
          TEMP="$(mktemp)"
          sed 's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' "$BASE" > "$TEMP" 
          mv "$TEMP" "$BASE"
```

michaeldelago:
What error were you getting to cause this? It'll be helpful to update the PR's body with that for better context on what the issue may be. 

I'm not sure if `sed -i` is the problem here, more so that the container is running as root for the compilation. from a clean build, what happens when you set the `DOCKER_USER` environment variable to `acore`

note that some places in `docker-compose.yml` have the default set to root https://github.com/azerothcore/azerothcore-wotlk/blob/a05ee01ec416845b0a54747172ce24f2685eb07c/docker-compose.yml#L83 whereas in other places it's not specified (like the [ac-db-import](https://github.com/azerothcore/azerothcore-wotlk/blob/a05ee01ec416845b0a54747172ce24f2685eb07c/docker-compose.yml#L154) service) or it defaults to `acore`: https://github.com/azerothcore/azerothcore-wotlk/blob/a05ee01ec416845b0a54747172ce24f2685eb07c/docker-compose.yml#L14

I think this is a bug, and should probably be fixed. Root shouldn't really be the default anywhere.

ben-of-codecraft:
The only error that I was getting was the following: 
```bash
azerothcore-wotlk-with-npcbots-ac-db-import-1 exited with code 1
service "ac-db-import" didn't complete successfully: exit 1
error: Uncaught (in promise) Error: Failed with error: 1, however,
          it's not related to this Deno script directly. An error occurred within
          the script called by the command itself
          throw new Error(`Failed with error: ${status.code}, however,
```
This was on a new Mac. 

Appreciate the suggested changes and can make those.

ben-of-codecraft:
Tested setting DOCKER_USER acore environment variable and updating docker conf, neiither created a clean build.

ben-of-codecraft:
This is the error when you set user to _acore_

```bash
-- Configuring incomplete, errors occurred!
See also "/azerothcore/var/build/obj/CMakeFiles/CMakeOutput.log".
See also "/azerothcore/var/build/obj/CMakeFiles/CMakeError.log".
CMake Error: Cannot open file for write: /azerothcore/var/build/obj/CMakeCache.txt.tmp
CMake Error: : System Error: Permission denied
CMake Error: Unable to open cache file for save. /azerothcore/var/build/obj/CMakeCache.txt
CMake Error: : System Error: Permission denied
```

michaeldelago:
I'm pretty sure the "easy" fix (for both of these error, actually) is to run `sudo chown -R $(id -u): ./` in the azerothcore-wotlk root. This is assuming that the `DOCKER_USER_ID` env var is the same as the output of `id -u` (if not, you can set that in [here](https://github.com/azerothcore/azerothcore-wotlk/blob/master/conf/dist/env.docker#L25)). You may be able to actually just delete `./var/docker`, which maps to `./var/build/obj`; that's just build cache. 

Long term, I'm hoping to get https://github.com/azerothcore/azerothcore-wotlk/pull/16806 merged, which offers a more elegant interface to solve the same problem that you're encountering. Instead of messing around with shell (while i love it, notoriously wonky), why not just let config files reference each other?

apps/compiler/includes/functions.sh
```diff
@@ -208,9 +208,9 @@ function conf_layer() {
         if grep -qE "^$KEY" "$BASE" && ! grep -qE "^$KEY.*=.*$VAL" "$BASE"; then
           # Replace line
           # Prevent issues with shell quoting 
-          sed -i \
-            's,^'"$KEY"'.*,'"$KEY = $VAL$COMMENT"',g' \
-            "$BASE"
+         TEMP="$(mktemp)"
```
michaeldelago, line 211: nitpick, `TEMP` is missing a space in its indentation. I probably left a bad "suggestion" block in the last comment

```suggestion
          TEMP="$(mktemp)"
```

ben-of-codecraft:
I agree with what you are saying about includes of configs vs. rewriting them like they are today.  I am not certain how long it will be before the other PR hits. Today, if you git clone the repo down and follow the instructions for docker install on Mac, it will not build, and throw the error above.  This one/three liner at least gets it working out of the box for the moment.

gruberchris:
I don't think this issue is unique to only macOS docker builds. I just tried building on Ubuntu linux and I get the following:
image

gruberchris:
I also tried the alternative suggestions @michaeldelago mentioned here but those don't seem to resolve the issue for me. 
image

michaeldelago:
> I also tried the alternative suggestions @michaeldelago mentioned here but those don't seem to resolve the issue for me.

@gruberchris That error happened depending on the coreutils you're using. You can change that to `sudo chown -R $(id -u):$(id -g) ./`

The dbimport fails like that if there isn't a config. Can you confirm that you've got the `dbimport.conf` in the right location? it should be in `env/dist/etc/dbimport.conf`. If it's not there, there's probably the "dist" file right next to it, and you can just rename/copy that file into the proper `env/dist/etc/dbimport.conf`.

ben-of-codecraft:
@michaeldelago @gruberchris what would happen to me is it would create sed-2d2dsfs - files in the conf on start, then failed.  Once swapped in the change, it is fine.  

I installed the branch that removes deno and it also addresses this permission error.

ben-of-codecraft:
I am closing there has been a full rebuild of how docker has been done, and in that build this becomes unnecessary. At least it is here now for those who need a quick fix that do not go through the process of updating to a full new decompose.

PR #17022, opened 2023-08-16
fix(DB/TirisfalGlades): rework Agamand Family Crypt creature positions to make sure creatures don't aggro weirdly anymore

elthehablo:
## Changes Proposed:
This PR proposes changes to:

adds new creature positions
adds waypoints for 1 rotting ancestors

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5699

## SOURCE:

sniffs by [Bench](github.com/heyitsbench) from wotlk classic

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ] spawns are probably pooled and this means that this PR will likely be overhauled in the future by bigger project that aims to make the starting zones better with help of sniffs


elthehablo:
can anyone tell me why the action is failing? too tired to see it rn

Gultask:
EOL conversion probs
image

elthehablo:
> EOL conversion probs image

I think I did that... but something prob got messed up in the process

Gultask:
Ok so I gotta ask some questions for this one :p 
How bad was it before? How much of an improvement is this?
On first glance, I can see that wander_distance was left at a default of 5, is that true for all creatures? Currently it isn't all 5
We also have some flags_extra that could help, in particular 33554432 will prevent the creature from calling for assistance on aggro

elthehablo:
> Ok so I gotta ask some questions for this one :p How bad was it before? How much of an improvement is this? On first glance, I can see that wander_distance was left at a default of 5, is that true for all creatures? Currently it isn't all 5 We also have some flags_extra that could help, in particular 33554432 will prevent the creature from calling for assistance on aggro

I do understand that. I think they do call for assistance. No definitive proof for that, but this is known to also be an issue on live. My theory is that on CC mobs are just cluttered too close to each other so they pull everything. With my tests on this PR it did get a lot better. Positions are all sniffed btw. Is wander distance sniffable? I didn't look very hard but couldn't find it in this parse :p

elthehablo:
I mean how bad was it before/how good is it now is why I want people to test this btw :)

data/sql/updates/pending_db_world/agamand.sql
```diff
@@ -0,0 +1,66 @@
+-- 
+DELETE FROM `creature` WHERE `id1` IN (1530, 1534, 1658);
```
Kitzunu, line 2: GUID is pri key. Delete GUID.

Kitzunu (changes requested):

Kitzunu:
> I mean how bad was it before/how good is it now is why I want people to test this btw :)

> Tested in-game by the author.

You should have the answer to the question though 🤔

data/sql/updates/pending_db_world/agamand.sql
elthehablo: In this instance I want all of them removed and replaced. Is this not allowed in this case? I mean I can put all the guids between brackets but I am also lazy

Kitzunu: You need to remove GUID, as it is the only primary key in this table. Otherwise, there is a risk for duplicate entries.

elthehablo:
> > I mean how bad was it before/how good is it now is why I want people to test this btw :)
> 
> > Tested in-game by the author.
> 
> You should have the answer to the question though 🤔

I'll do some more testing and report back

elthehablo:
So here is a recording from CC PTR:

https://www.youtube.com/watch?v=M65k7nsiAmA&ab_channel=EltheHablo

My local AC with this PR:

https://www.youtube.com/watch?v=3I7fz_eeP7E&ab_channel=EltheHablo

I think most mobs do wander on wotlk classic, except for the one that has wps. iirc wander behaviour is not completely understood on wotlk classic (could be wrong here). so I don't know how we can definitively prove which mobs should wander and how far. the main point of this PR is that you don't get shrekt the moment you pull the first mob downstairs. this is where I and many more have died

Kitzunu (approved):

Issue #16239, opened 2023-05-12
[PvP] Getting flagged through AoE abilities


balleny:
### Current Behaviour

Using AOE abilities without being flagged for pvp, will flag you for pvp if the abilities hit an enemy player who is flagged for pvp.

### Expected Blizzlike Behaviour

without being flagged for pvp, aoe abilities shouldnt affect players of the opposite faction that are flagged, and not flag the player using the abilities.

### Source

https://wowwiki-archive.fandom.com/wiki/Bluewall?oldid=2089274

Also note it is not possible to become flagged through use of AOE attacks around players or NPCs, the attack will simply have no effect and will not flag you. Bear this in mind if you are fleeing from NPC guards and you try and AOE fear/snare/etc to escape as it will do nothing. This behaviour also prevents griefing attacks where flagged players could stalk a player using AOE to farm mobs.

### Steps to reproduce the problem

2 chars, opposite faction
flag char a for pvp
use aoe abilities like arcane explosion, volley, consecration, thunderclap, starfall, explosive trap in proximity on char b

same applies for "jumping" abilities like chain lightning, avengers shield, living bomb, seed of corruption

### Extra Notes

this is a critical bug IMHO, because it is on the one hand abused ingame, and also bannable.

not 100% sure about abilities like whirlwind, divine storm, swipe (bear). I would assume those should NOT have an effect except if the player is being targeted.

(additional source, use a translator tool of your choice: https://redmine.rising-gods.de/issues/12904)

Chromiecraft report:
https://github.com/chromiecraft/chromiecraft/issues/5253

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/5e9b88a9a9ddf510ee2e7a95861988a1fb3e54ca

### Operating system

Ubuntu 20.04

### Custom changes or Modules

[mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
[mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
[mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
[mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
[mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
[...]

PR #17064, opened 2023-08-23
fix(Core/Unit): Added bluewall AoE check

AGandrup:
## Changes Proposed:
- AoE spells/abilities done by non-flagged players now has no effect on pvp flagged players.

## Issues Addressed:

- None.

## SOURCE:

"_it is not possible to become flagged through use of AOE attacks around players or NPCs, the attack will simply have no effect and will not flag you._" written 21 October 2008, https://wowpedia.fandom.com/wiki/Bluewall?oldid=1639893

"_Consecration and other similar spells can no longer be used by non-PvP flagged players to damage PvP flagged enemies._", Patch 1.7.0, 13 September 2005, https://wowpedia.fandom.com/wiki/Patch_1.7.0

"_In WoW aoe effects does not hit pvp flagged players unless you're pvp flagged as well._" written 27 December 2010, https://forums-archive.anarchy-online.com/showthread.php?585603-Turn-PVP-on-off&s=1c6663365aa80dc8fb5d4b159fc2e78f&p=5916665&viewfull=1#post5916665

## Tests Performed:

- Builds
- Tested in-game, works as described above.

## How to Test the Changes:

1. Create two accounts
2. On one account create a Horde character, on the other an Alliance character.
3. Meet up in game.
4. Have one flagged for pvp and the other not flagged.
5. On the non-flagged character cast whatever AoE spell on the flagged character - notice it has no effect and does not trigger pvp flag.
6. Try whatever direct attack/spell - notice that it will hit and trigger pvp flag.

- None so far.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13957,7 +13957,7 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
     // additional checks - only PvP case
     if (playerAffectingAttacker && playerAffectingTarget)
     {
-        if (!IsPvP() && bySpell && bySpell->IsAffectingArea())
+        if (!IsPvP() && bySpell && bySpell->IsAffectingArea() && !bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK))
             return false;
 
         if (target->IsPvP())
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13957,7 +13957,7 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
     // additional checks - only PvP case
     if (playerAffectingAttacker && playerAffectingTarget)
     {
-        if (!IsPvP() && bySpell->IsAffectingArea())
+        if (!IsPvP() && bySpell && bySpell->IsAffectingArea())
             return false;
 
         if (target->IsPvP())
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13957,6 +13957,9 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
     // additional checks - only PvP case
     if (playerAffectingAttacker && playerAffectingTarget)
     {
+        if (!IsPvP() && bySpell->IsAffectingArea())
+            return false;
+
         if (target->IsPvP())
             return true;
 


AGandrup:
I fell over this spell attribute: ``SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK`` and thought maybe it would be more correct to check with ``bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK)`` rather than ``bySpell->IsAffectingArea()``. However I tried it but it does not seem to work and I am thinking its either because it does not mean what I think it means or some (if not all) AoE spells, which is supposed to have this attribute, is missing it and needs to be updated somehow. 🤔

Perhaps someone can clarify.

Cinnamohn:
> I fell over this spell attribute: `SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK` and thought maybe it would be more correct to check with `bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK)` rather than `bySpell->IsAffectingArea()`. However I tried it but it does not seem to work and I am thinking its either because it does not mean what I think it means or some (if not all) AoE spells, which is supposed to have this attribute, is missing it and needs to be updated somehow. 🤔
> 
> Perhaps someone can clarify.

Based on the wording of `SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK` and a few of the random spells I found flagged with it, I believe it's meant to ignore this new check. So the code would look more like this:

```
if (!IsPvP() && bySpell && bySpell->IsAffectingArea() && !bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK))
    return false;
```

AGandrup:
> > I fell over this spell attribute: `SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK` and thought maybe it would be more correct to check with `bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK)` rather than `bySpell->IsAffectingArea()`. However I tried it but it does not seem to work and I am thinking its either because it does not mean what I think it means or some (if not all) AoE spells, which is supposed to have this attribute, is missing it and needs to be updated somehow. 🤔
> > Perhaps someone can clarify.
> 
> Based on the wording of `SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK` and a few of the random spells I found flagged with it, I believe it's meant to ignore this new check. So the code would look more like this:
> 
> ```
> if (!IsPvP() && bySpell && bySpell->IsAffectingArea() && !bySpell->HasAttribute(SPELL_ATTR5_IGNORE_AREA_EFFECT_PVP_CHECK))
>     return false;
> ```

Yes I think you are right. I also found that in the the Cataclysm Preservation Project (TrinityCore Cata branch) they use this check as well, though the other way around: https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/blob/58302441a20e79f1c3affcd49c74efa89bbc3a79/src/server/game/Entities/Object/Object.cpp#L3114

Nyeriah (approved):

balleny:
relevant issue
https://github.com/azerothcore/azerothcore-wotlk/issues/16239

Issue #17109, opened 2023-08-30
Env vars for configuration do not work for modules


michaeldelago:
### Current Behaviour

In using some modules, I was not able to configure them with env vars.

Specifically, it was [SoloCraft](https://github.com/azerothcore/mod-solocraft) though this issue has been reported in Discord regarding [ah-bot](https://github.com/azerothcore/mod-ah-bot)

- SoloCraft has a (probably?) bug where Blackwing Lair has a level set to "40"
    - In actuality, Blackwing lair is a level 60 raid that requires completion of a level [55-60 dungeon (UBRS)](https://github.com/azerothcore/mod-solocraft/blob/master/src/Solocraft.cpp#L103) to access it
- Accessing BWL at level 60 with SoloCraft doesn't work as it seems like it should, since SoloCraft doesn't scale enemies if you're 10 levels out of range. 

### Expected Blizzlike Behaviour

Environment variables as configuration should work for Modules

### Source

N/A: This is a feature of the core itself

### Steps to reproduce the problem

1. `git clone git@github.com:azerothcore/azerothcore-wotlk`
2. `cd azerothcore-wotlk/modules && git clone git@github.com:azerothcore/mod-solocraft && cd ../`
3. Compile the server as normal. 
    1. I would use docker but deployment method isn't relevant here.
    2. `./acore.sh docker build`
4. Throw in a configuration for the module to the environment
    1. For a "traditional" deployment, this can be done by setting an environment variable as you normally would:
    ```console
    $ AC_SOLOCRAFT_BLACKWING_LAIR_LEVEL=60 env/dist/bin/worldserver
    ```
    2. With docker-compose, you can create an override:
    
    ```yaml
    # docker-compose.override.yml
    services:
      ac-worldserver:
        environment:
          AC_SOLOCRAFT_BLACKWING_LAIR_LEVEL: 60
    ```

5. It does work if the parameter is removed (or commented out) from the modules config files (in this case, `env/dist/etc/modules/Solocraft.conf` and `env/dist/etc/modules/Solocraft.conf.dist`)

### Extra Notes

Solocraft has inconsistencies in its' configuration schema. In some places, it has the first letter capitalized `Solocraft` (such as [`Solocraft.BlackwingLair.Level`](https://github.com/azerothcore/mod-solocraft/blob/master/src/Solocraft.cpp#L113)) whereas other places use PascalCase, like `SoloCraft`

### AC rev. hash/commit

[...]

PR #17114, opened 2023-08-31
refactor(Core): Source config more aggressively from env, rephrase a few of the messages

michaeldelago:
## Changes Proposed:
This PR proposes changes to:

### Description

- Change the `GetValueDefault` function in `Config.cpp` to always choose environment variable over which value is in a config file
    - This is typically how environment var configs work in many programs; the environment var always take precedence over the config file
    - The removes the need to execute an `OverrideWithEnvVariablesIfAny()` step
- This fixes env vars for modules since previously, modules would be loaded after the `OverrideWithEnvVariablesIfAny()` function
- I also changed some of the output messages related to loading configurations from env vars. They kind of made it seem like a config using an environment var was recovering from an error rather than just a method of loading config values
- Additionally, the noted message would log the value of the environment variable, which is (quite) bad if you're using environment variables for database configs. Since it was removed, this is fixed.
    - `"Missing property {} in config file {}, recovered with environment '{}' value."` 
- Log lines for specific situations as of this pr:
    - Config missing from both file and environment: 

    ```
    Config: Missing property Solocraft.Enable in config file /home/mike/builds/azerothcore-wotlk/env/dist/etc/worldserver.conf, add "Solocraft.Enable = 1" to this file or define 'AC_SOLOCRAFT_ENABLE' as an environment variable.
    ```

    - Config found in environment, and it's upsert the stored config value: 

    ```
    Config: Found config value 'Solocraft.BlackwingLair.Level' from environment variable 'AC_SOLOCRAFT_BLACKWING_LAIR_LEVEL'.
    ```

- Provides an easy method to get an environment variable equivalent of a parameter name. comment out the parameter name (so the default isnt in the conf), and see what the output for the "missing property" error is

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17109

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. `git clone git@github.com:azerothcore/azerothcore-wotlk`
2. `git clone git@github.com:azerothcore/mod-solocraft azerothcore-wotlk/modules/mod-solocraft`
3. Compile the server as normal.
4. Throw in a configuration for the module to the environment
    1. For a "traditional" deployment, this can be done by setting an environment variable as you normally would:
    ```console
    $ AC_SOLOCRAFT_BLACKWING_LAIR_LEVEL=60 env/dist/bin/worldserver
    ```
    2. With docker-compose, you can create an override:
    
    ```yaml
    # docker-compose.override.yml
    services:
      ac-worldserver:
        environment:
          AC_SOLOCRAFT_BLACKWING_LAIR_LEVEL: 60
    ```

    2. The server will log that the config is being pulled in from the env var

5. Log in, create a character. Observe that the config is working. in this case, run these commands after creating your character and logging in

```console
.levelup 59
.tele bwl
```

6. Notice that the instance is working properly.

7. removed (or comment out) the relevant parameter from the modules config files and remove the env var, however you have it set. The worldserver should log an error if the config is missing and describe what to do
    1. in this case, `env/dist/etc/modules/Solocraft.conf` and `env/dist/etc/modules/Solocraft.conf.dist`

None that I know of.

### DIFF
--- a/src/common/Configuration/Config.cpp
+++ b/src/common/Configuration/Config.cpp
@@ -283,7 +283,7 @@ namespace
         return result;
     }
 
-    std::string getEnvVarName(std::string const& configName)
+    std::string GetEnvVarName(std::string const& configName)
     {
         return "AC_" + IniKeyToEnvVarKey(configName);
     }
--- a/src/common/Configuration/Config.cpp
+++ b/src/common/Configuration/Config.cpp
@@ -119,7 +119,7 @@ namespace
             auto const& itr = fileConfigs.find(confOption);
             if (itr != fileConfigs.end())
             {
-                PrintError(file, "> Config::LoadFile: Dublicate key name '{}' in config file '{}'", confOption, file);
+                PrintError(file, "> Config::LoadFile: Duplicate key name '{}' in config file '{}'", confOption, file);
                 return true;
             }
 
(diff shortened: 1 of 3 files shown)


src/common/Configuration/Config.cpp
```diff
@@ -356,28 +356,30 @@ template<class T>
 T ConfigMgr::GetValueDefault(std::string const& name, T const& def, bool showLogs /*= true*/) const
 {
     std::string strValue;
+
     auto const& itr = _configOptions.find(name);
-    if (itr == _configOptions.end())
+    bool notFound = itr == _configOptions.end();
+    Optional<std::string> envVar = EnvVarForIniKey(name);
```
walkline, line 362: I have a concern with this approach. In this approach, every function call to retrieve a certain config value would result in a syscall to retrieve an environment variable. And since not all config values are cached, with these changes it could potentially impact performance. Personally, I would prefer to load and cache all environment variables during startup.

michaeldelago: Yeah, this was in the back of my mind while I was working on this, but I was worried it'd be a premature optimization. 

It can't hurt though, I'll probably be able to make this change evening.

michaeldelago: @walkline Added a cache of env vars

src/common/Configuration/Config.cpp
```diff
@@ -282,9 +283,14 @@ namespace
         return result;
     }
 
+    std::string getEnvVarName(std::string const& configName)
```
Kitzunu, line 286:
```suggestion
    std::string GetEnvVarName(std::string const& configName)
```

src/common/Configuration/Config.cpp
```diff
@@ -329,6 +335,29 @@ bool ConfigMgr::Reload()
     return true;
 }
 
+// Check the _envVarCache if the env var is there
+// if not, check the env for the value
+Optional<std::string> getEnvFromCache(std::string const& configName, std::string const& envVarName)
```
Kitzunu, line 340:
```suggestion
Optional<std::string> GetEnvFromCache(std::string const& configName, std::string const& envVarName)
```

Kitzunu (approved):

Issue #17129, opened 2023-09-01
[Rescue OOX-17/TN!] impossible to complete in group


Pyromanihack:
Original issue: https://github.com/chromiecraft/chromiecraft/issues/4913

### What client do you play on?

enGB - enUS

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

while doing this quest in group, only who started the quest get the reward

### Expected Blizzlike Behaviour

everyone in the group should complete the quest

### Source

It is logical, since the whole party has accepted the quest

### Steps to reproduce the problem

1. Join in a party
2. `.additem 8623` with each party member and start the quest Find OOX-17/TN!
3. `.go zone 60.21 64.67` with each party member and complete the quest Find OOX-17/TN!
4. Accept the next quest Rescue OOX-17/TN! with one player (a window will appear to the other party members, you need to click ok)
5. Do the escort with all the party
6. At the end of the quest, only the first player will complete the quest

### AC rev. hash/commit

6cd3b2a32cd9

### Operating system

Ubuntu 20.04 - Windows 10

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #17139, opened 2023-09-01
fix(DB/Tanaris): [Rescue OOX-17/TN!] completes for all party members now

elthehablo:
## Changes Proposed:
This PR proposes changes to:

Store target list and target list as `target_type` seem to not work correctly in some cases. I would suggest changing to `target_type` 16 for all of these. The reason this might've not been chosen the first time around is to perhaps prevent griefing, when you throw someone out of the party? But this is highly debatable to be an issue and would be a minor issue imo.  This `target_type` (16), to me, seems like the best solution.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17129
- Closes https://github.com/chromiecraft/chromiecraft/issues/4913

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

Do this on multiple characters (up to 5)
1. `.additem 8623`
2. `.list creature 7784`
3. `.go creature <homingchickenGUID>`
4. turn in the quest from the item on all characters but don't accept escort yet
5. when all first quests are turned in on all characters, accept escort
6. complete escort and witness that everybody gets the credit

- [ ] fix waypoints with sniffs, as the chicken clips through a hill in the beginning
- [ ] fix other escorts in azeroth with the same issues


Gultask:
Does it work? Doesn't make sense to me since UPDATE_OOC doesn't have an invoker. Maybe it's calling for the last invoker stored, which would be the party, but still, I'd imagine it'd be overwritten by something. Usually Store Targetlist worked for me.

Oh well, if it works it works.

elthehablo:
It worked for me. More testing welcome ofc

Gultask:
I'll do one myself if I can remember, or if you're interested, one thing that caught my eye is On Death -> Fail Quest.
Invoker should be the killer, but it's an escort quest so the killer won't be players.

elthehablo:
> I'll do one myself if I can remember, or if you're interested, one thing that caught my eye is On Death -> Fail Quest. Invoker should be the killer, but it's an escort quest so the killer won't be players.

Best to keep that as 12 then and re-add the store data

elthehablo:
I don't really find it an elegant way tbh. I wish there was a better way

Kitzunu:
stastus

Nyeriah:
statusd

Kitzunu:
Status

Nyeriah:
Status

YungElf:
Tested.

Tested on 2 characters, accepted Escort quest within a second or so of each other, got chicken to safety and both received credit for the quest. As stated by elthehablo, waypoints need to be sniffed as the chicken clips in the beginning of the quest. Also unsure if this is incorrect, but the quest is NOT shareable in its current state. 
Regardless, this PR fixes the issue listed. 

Thanks!

MikaMauger:
There are a lot of escort quests which won't complete for all party members. There was a [fix for this in TrinityCore](https://github.com/TrinityCore/TrinityCore/issues/24706), and they chose to implement it using this trick (target_type=16).

Issue #17201, opened 2023-09-07
Adamantite Shell Machine has a 1.5 second GCD


Pyromanihack:
Original issue: https://github.com/chromiecraft/chromiecraft/issues/5149

### What client do you play on?

enGB - enUS

### Faction

Alliance

### Content Phase:

70

### Current Behaviour

Currently, Adamantite Shell Maker boxes when used, (5 charges of 200 shells) cost a GCD. Adamantite Arrow Maker boxes work correctly and you can spam 200 arrows out of the 5 charge box with no GCD.

### Expected Blizzlike Behaviour

Both boxes should be spam click-able to receive the ammunition with no GCD.

### Source

From wowhead (GCD set to zero)
Spell ID 45171
![shell](https://github.com/azerothcore/azerothcore-wotlk/assets/82369562/4350630f-eb7b-4453-8dfb-eb60124f8c5e)

### Steps to reproduce the problem

1. `.additem 34504` Adamantite Shell Machine
2. `.additem 20475` Adamantite Arrow Maker 
3. Notice one takes a gcd and the other does not.

### AC rev. hash/commit

dd647da913ef

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
[...]

PR #17202, opened 2023-09-07
feat(DB/Item_template): Set Adamantite Shell Machine GCD to zero

Pyromanihack:
## Changes Proposed:
This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17201
- Closes https://github.com/chromiecraft/chromiecraft/issues/5149

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. `.additem 34504` Adamantite Shell Machine
2. `.additem 20475` Adamantite Arrow Maker
3. Now Adamantite Shell Machine has not a GCD (like Arrow Maker)

- [ ]
- [ ]

### DIFF
--- a/Adamantite Shell.sql
+++ b/Adamantite Shell.sql
@@ -1,0 +1,3 @@
+
+
+UPDATE `item_template` SET `spellcooldown_1` = 0 WHERE (`entry` = 34504); -- Set the cooldown of Adamantite Shell Machine to zero


heyitsbench:
Just to confirm, you yourself tested this on live servers?

Pyromanihack:
> Just to confirm, you yourself tested this issue on live servers?

First I observed the bug on my local server, and then I applied the changes and re-tested on local

heyitsbench:
> First I observed the bug on my local server, and then I applied the changes and re-tested on local

My concern is no source for the bug in the original issue(s).

Pyromanihack:
> > First I observed the bug on my local server, and then I applied the changes and re-tested on local
> 
> My concern is no source for the bug in the original issue(s).

I'm updating the sources

samsam1993:
This seems to be a core-related issue. The GCD does not go away with this SQL modification.

The-GhostRider:
> This seems to be a core-related issue. The GCD does not go away with this SQL modification.

It works..
Delete Cache..

The-GhostRider (approved):
it's good.
For me ti can be merged..

Kitzunu:
Did you test this on live servers? (I know the template says so I just want to make sure)

Not saying it should have a GCD, but perhaps it should have an ICD

Because just to throw a wrench in this change

image

The-GhostRider:
> Did you test this on live servers? (I know the template says so I just want to make sure)
> 
> Not saying it should have a GCD, but perhaps it should have an ICD
> 
> The spell has no GCD in DBC data, so there is nothing wrong with the spell. It is the item that has the cooldown so the proof from wowhead about the spell is useless.
> 
> Because just to throw a wrench in this change
> 
> ![image]

"This is strange, though, because there is a senseless discrepancy on Wowhead...
HERE: [Adamantite Shell Machine](https://www.wowhead.com/wotlk/item=34504/adamantite-shell-machine)
We have an ICD (Internal Cooldown) on the item of 1500ms
image

While HERE: [Adamantite Arrow Maker](https://www.wowhead.com/wotlk/item=20475/adamantite-arrow-maker)
There is a discrepancy with its counterpart...
image

Now, these two items/spells essentially do the same thing... except one gives you "Arrows" and the other gives you "Ammo."

Why does one have an ICD and the other doesn't? (I wonder).

- Is it a Blizzard mistake? (an oversight)
- Is it right for both to have an ICD?
- Is it right for both not to have an ICD?

What is the difference?"

Give an explanation to this @Kitzunu

Kitzunu:
I don't know, this is why I ask it to be properly investigated. 

Sometimes I need to play the devil's advocate when things are just presented out of a hunch or from wowhead. I don't want random things being pushed into the core to later just be reverted because it was just wrong.

But this could be a classic case of Blizzard being Blizzard. Sometimes that is the only explanation to their behavior.
It is also the reason as to why we should never trust wowhead. But Bench will check it out on live servers.

The-GhostRider:
> I don't know, this is why I ask it to be properly investigated.
> 
> Sometimes I need to play the devil's advocate when things are just presented out of a hunch or from wowhead. I don't want random things being pushed into the core to later just be reverted because it was just wrong.
> 
> But this could be a classic case of Blizzard being Blizzard. Sometimes that is the only explanation to their behavior. It is also the reason as to why we should never trust wowhead. But Bench will check it out on live servers.

Instead, you know, maybe I've found an explanation for this discrepancy...

Do you know what it might be (in my opinion)?
If you look at the item descriptions, the two items are nearly identical!
They only differ in two things...

The name of the item
and the "Warning" that's present in only one of them...
Is it possible that Blizzard added those 1500ms of ICD as a form of "Role Play" to fit the context of the role-playing game, and since the item "Warns" that it "Contains Explosives" and should be handled with care... they inserted that CD for that reason...

So it's just a nonsensical role-play element... 🤣🤣🤣

Pyromanihack:
Interesting point of view, Ghostrider

At this point, I'd say I can close this PR because it's blizzlike

PR #17240, opened 2023-09-13
fix: cppcheck failure step

Helias:
## Changes Proposed:
This PR make the pipeline fails when cppcheck returns an error

## Issues Addressed:

- cppcheck should make fail the pipeline when it returns an error

## Tests Performed:

a proper pipeline success should be enough

## How to Test the Changes:

1. if at least once cppcheck make fail the pipeline it should be enough


Kitzunu:
Question: Why do we have this cpp check and codefactor?

Helias:
> Question: Why do we have this cpp check and codefactor?

cpp check to increase the code quality, codefactor I guess for the same reason.

cpp check will prevent us to leave some possible null-pointer around the project without any check or uninitialized variables and other checks

codefactor I think is common used and it does not need to any explanation

Any tool for static check I think is a good plus to have in the pipeline

Kitzunu:
Yea my question was moreso why both

Kitzunu (approved):

PR #17302, opened 2023-09-18
chore(docs): Update Code of Conduct to version 2.1

Kitzunu:
https://www.contributor-covenant.org/version/2/1/code_of_conduct/

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -47,8 +47,7 @@ or harmful.
 
 Community leaders have the right and responsibility to remove, edit, or reject
 comments, commits, code, wiki edits, issues, and other contributions that are
-not aligned to this Code of Conduct, and will communicate reasons for moderation
-decisions if considered neccessary.
+not aligned to this Code of Conduct.
 
 ## Scope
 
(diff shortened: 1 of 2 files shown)


.github/CODE_OF_CONDUCT.md
```diff
@@ -1,76 +1,134 @@
...
-* The use of sexualized language or imagery and unwelcome sexual attention or
- advances
-* Trolling, insulting/derogatory comments, and personal or political attacks
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
 * Public or private harassment
-* Publishing others' private information, such as a physical or electronic
- address, without explicit permission
+* Publishing others' private information, such as a physical or email address,
```
55Honey, line 35: I think electronic address planned to refer to any doxxable data. At least it should. We should reflect that. Maybe add
`identities on other platforms`
to the list.

Kitzunu: These were intended changes Contributor Covenant has made between revisions of the file. But I can definitely reword it

Kitzunu: I will leave this as-is unless you think otherwise. I interpret it as it's just examples due to the `such as`. `Publishing others' private information` should cover it all imo

PR #17364, opened 2023-09-26
fix(Scripts/Magtheridon): make the timers of Quake and Blast Nova independent from each other

elthehablo:
## Changes Proposed:
This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/5906

## SOURCE:

Sniff used for blast nova timer, video evidence for quake timer using the TBC classic world first video (as I couldn't find it in sniffs)

[This](https://youtu.be/PVL-m0CvfI0?si=ooxbB1uoytDh6t5I&t=144) video clearly shows that blast nova and quake are fully independent. So this is the most likely solution

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] some issues regarding the beams, the videos show they seem to stay up for a bit longer after clicking
- [ ] some of the other issues reported

### DIFF
--- a/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
+++ b/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
@@ -185,6 +185,7 @@ struct boss_magtheridon : public BossAI
         }).Schedule(55650ms, [this](TaskContext context)
         {
             DoCastSelf(SPELL_BLAST_NOVA);
+            scheduler.DelayAll(9999ms);
 
             _interruptScheduler.Schedule(50ms, GROUP_INTERRUPT_CHECK, [this](TaskContext context)
             {
--- a/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
+++ b/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
@@ -185,7 +185,7 @@ struct boss_magtheridon : public BossAI
         }).Schedule(55650ms, [this](TaskContext context)
         {
             DoCastSelf(SPELL_BLAST_NOVA);
-            scheduler.DelayAll(9999ms);
+            scheduler.DelayAll(10s);
 
             _interruptScheduler.Schedule(50ms, GROUP_INTERRUPT_CHECK, [this](TaskContext context)
             {
--- a/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
+++ b/src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
@@ -166,7 +166,7 @@ struct boss_magtheridon : public BossAI
         {
             me->CastCustomSpell(SPELL_BLAZE, SPELLVALUE_MAX_TARGETS, 1);
             context.Repeat(11s, 39s);
-        }).Schedule(40s, [this](TaskContext context)
+        }).Schedule(28300ms, [this](TaskContext context)
         {
             DoCastSelf(SPELL_QUAKE);
             _castingQuake = true;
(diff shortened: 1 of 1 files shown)


balleny:
commonly observed is that quake happens before blast nova for the first 3 to 4 iterations, then positions are switched.
is this achieved with the timers in place?

see
https://www.youtube.com/watch?v=PVL-m0CvfI0
https://www.youtube.com/watch?v=QpgdAHJANcA

elthehablo:
> commonly observed is that quake happens before blast nova for the first 3 to 4 iterations, then positions are switched. is this achieved with the timers in place?
> 
> see https://www.youtube.com/watch?v=PVL-m0CvfI0 https://www.youtube.com/watch?v=QpgdAHJANcA

yes this is achieved you can do this by simply comparing the timers:

sequence:
first quake at: 28300ms

first blast nova at: 55650ms

second quake at 28300 + (avg(56300ms, 64300ms)) which is about 1.5mins

second blast nova at: around 2ish mins

max timers don't exist in classic wow, so there is only so much we can do here. now the way we have scripted it, there dependency is completely gone, although it might still seem dependent some of the time, which is likely the reason it got coded the way it was. if you look at the videos though, this new way seems to be the correct way

elthehablo:
> I was mostly asking if it was tested because there is also the DelayAll in, which might be not reflected in the simple addition of timers. Also, testing should be conducted to ensure that the boss doesnt start to use quake while already channeling blastnova - given that the earthquake wasnt properly displayed in the combatlog yet either.

ye I will add a delay for that

elthehablo:
> problem here is that the delay persists even when interrupted...

but that is quite minute and maybe even blizzlike

elthehablo:
After having let him have at me for around 7 minutes or so I can pretty confidently say that this seems to be working as intended. You should also not be able to get a quake now whilst he's casting blast nova as that would be catastrophic

balleny:
> After having let him have at me for around 7 minutes or so I can pretty confidently say that this seems to be working as intended. You should also not be able to get a quake now whilst he's casting blast nova as that would be catastrophic

i mostly raised the concern as some magtheridon mythic mode (on Chrmie or if tuned on other servers) could result in the fight lasting 5 mins or longer for the second phase alone

elthehablo:
> > After having let him have at me for around 7 minutes or so I can pretty confidently say that this seems to be working as intended. You should also not be able to get a quake now whilst he's casting blast nova as that would be catastrophic
> 
> i mostly raised the concern as some magtheridon mythic mode (on Chrmie or if tuned on other servers) could result in the fight lasting 5 mins or longer for the second phase alone

unless I am not mistaken delay postpones abilities, so they will still be queued, but just be on halt for x seconds

src/server/scripts/Outland/HellfireCitadel/MagtheridonsLair/boss_magtheridon.cpp
```diff
@@ -175,29 +175,33 @@ struct boss_magtheridon : public BossAI
...
                 _castingQuake = false;
                 me->SetReactState(REACT_AGGRESSIVE);
                 me->GetMotionMaster()->MoveChase(me->GetVictim());
-                DoCastSelf(SPELL_BLAST_NOVA);
+            });
+            context.Repeat(56300ms, 64300ms);
+        }).Schedule(55650ms, [this](TaskContext context)
+        {
+            DoCastSelf(SPELL_BLAST_NOVA);
+            scheduler.DelayAll(9999ms);
```
Kitzunu, line 188: was this sniffed? because it looks very weird

elthehablo: Just 1ms below 10s. Following what I've seen happen on Gruul. That's the channel time of blast wave

elthehablo: > was this sniffed? because it looks very weird

But I can make it 10s if you want?

Nyeriah (approved):

PR #17387, opened 2023-09-29
fix(core/scripting) Calculate percent-based damage before `ModifyPeriodicDamageAurasTick` hook

kjack9:
## Changes Proposed:
This PR proposes changes to:

This change moves the calculation of damage auras based on a percent of the target's health to before the `ModifyPeriodicDamageAurasTick` hook. Since the hook has no knowledge of whether the `damage` value is absolute or percent-based, this will allow modules to alter the final absolute damage amount as would be expected.

## SOURCE:

## Tests Performed:

This PR has been:

Tested using an upcoming version of AutoBalance. I can succesfully modify the damage of [Giant Insect Swarm (70475)](https://www.wowhead.com/wotlk/spell=70475/giant-insect-swarm).

```
AutoBalance_UnitScript::ModifyPeriodicDamageAurasTick(): BEFORE: Putricide's Trap -1516 Warriortest (Giant Insect Swarm - 70475)
AutoBalance_UnitScript::_Modify_Damage_Healing(): Using the source creature's damage multiplier: 0.7223805
AutoBalance_UnitScript::_Modify_Damage_Healing(): Returning modified damage: -1516 * 0.7223805 = -1095.1289
AutoBalance_UnitScript::ModifyPeriodicDamageAurasTick(): AFTER: Putricide's Trap -1095 Warriortest (Giant Insect Swarm - 70475)
```

image

## How to Test the Changes:

Use a module to hook `ModifyPeriodicDamageAurasTick` and modify the `damage` parameter by any amount.


kjack9:
Sorry for the duplicate PRs, feel free to delete the other (broken) one. Git and I aren't on speaking terms right now.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -6329,8 +6336,6 @@ void AuraEffect::HandlePeriodicDamageAurasTick(Unit* target, Unit* caster) const
             // 5..8 ticks have normal tick damage
         }
     }
-    else // xinef: ceil obtained value, it may happen that 10 ticks for 10% damage may not kill owner
```
kjack9, line 6332: I know I'm replacing an `else` with a conditional, but the only two spell aura types that will hit this function anyways are `SPELL_AURA_PERIODIC_DAMAGE_PERCENT` and `SPELL_AURA_PERIODIC_DAMAGE`.

https://github.com/azerothcore/azerothcore-wotlk/blob/ce9a9d9345ad79dfd262dd1ebe2c98956deaa66c/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L1131-L1134

kjack9:
Happy to discuss this here or on Discord if anyone has questions about this fix. I will say that percent-based damage will be broken for scaling (as it is already) in current and upcoming AutoBalance releases if this remains unmerged. Let me know how I can change it or help move it forward.

Nyeriah:
I am actually concerned that this will affect preexisting scripts adversely and they will no longer be able to reach their intended behavior. e.g percentage based damage is usually fixed and we usually don't want to further modify it, but with this, spells will no longer deal the damage they were intended to as the value will be modified inside the hook

kjack9:
> I am actually concerned that this will affect preexisting scripts adversely and they will no longer be able to reach their intended behavior. e.g percentage based damage is usually fixed and we usually don't want to further modify it, but with this, spells will no longer deal the damage they were intended to as the value will be modified inside the hook

Yep, totally reasonable concern. I still think it should be changed. Let me justify why.

If all a module is looking at is the `damage` value, there is no indication of whether that value represents an absolute damage or a percent of character health. I would argue that based on the name (`damage`), it would be reasonable to assume that it's an absolute damage amount. And, once implemented, it would sure seem to work fine using that assumption... until the one-off DoT that ticks for a percentage. If your module increases "5" damage by what you think is a `modest` 500 damage, you just killed the player with a hit for 505% of their max health.

Assuming you even know that percent-based damage is handled differently, the only way to check to see what kind of damage is incoming is to peak at the `spellInfo` that has been passed into the hook. You'd have to know to look at `GetAuraType()`, then access the target's max health, then calculate the actual damage done yourself.  That seems at best unintuitive to me.

Finally, the `spellInfo` parameter was only introduced [10 months ago](https://github.com/azerothcore/azerothcore-wotlk/commit/a73d4173dfab04de9f58a9768c97f31caf5f81ea). Any module that hasn't been updated since this commit inherently suffers from this bug since it wasn't possible for them to even check this condition if they wanted to.

Here are some modules in GitHub that were either created before `spellInfo` was a thing, or do not use `spellInfo` to check the Aura type ([search link](https://github.com/search?q=ModifyPeriodicDamageAurasTick+-path%3A**%2FScriptMgr.*+-path%3A**%2FUnitScript.*+-path%3A**%2FSpellAuraEffects.*+-path%3A%22**%2FSpellAuras.*%22+language%3AC%2B%2B&type=code&p=1)):

- [mod-individual-progression](https://github.com/ZhengPeiRu21/mod-individual-progression/blob/f5bfc527294bf0bbfd9f88bd6e355e29a3705ad9/src/IndividualProgressionPlayer.cpp#L666)
- [HardMode](https://github.com/AnchyDev/HardMode/blob/fb11f64696032a611e2be45c07cad9f8e7443efd/src/HardModeHooks/HardModeHooksUnit.cpp#L135)
- [mod-zone-difficulty](https://github.com/azerothcore/mod-zone-difficulty/blob/c0529e0c6777564fbfbb934229cf8411557241e3/src/mod_zone_difficulty_scripts.cpp#L230) - actually detects if it's a `SPELL_AURA_PERIODIC_DAMAGE_PERCENT` aura, but then assumes (like I did) that the damage value is the absolute damage, not the percentage.
- [mod-ForgedInstance](https://github.com/hatersgit/hateforge/blob/1d209e04b0669dc1004574d6a277807d164e356c/modules/mod-ForgedInstance/src/ForgedInstance_scripts.cpp#L140)
- [mod-quickbalance](https://github.com/hallgaeuer/mod-quickbalance/blob/b6cc79e2ec4a9d87d52169519fb30263bfaf6db8/src/QuickBalance.cpp#L238) - AutoBalance alternative

Notably, I could find NO examples of a module that uses this hook and properly accounts for the percent damage condition. `mod-zone-difficulty` gets half credit for even recognizing that it's a possibility at all.

In conclusion, I think you're going to silently fix a lot more modules with this change than you could possibly break. 

Thank you for coming to my TED talk.

kjack9:
> percentage based damage is usually fixed and we usually don't want to further modify it, but with this, spells will no longer deal the damage they were intended to as the value will be modified inside the hook

Just to clarify on this point, today these percentages are being modified anyways but in a way that is not transparent to the mod author. I can see an argument for not having a percent-based aura trigger this hook AT ALL, and maybe that's what we should be talking about?

Nyeriah (approved):

PR #17407, opened 2023-10-03
fix(Core/Player) Stop small durability loss from still occuring when DurabilityLoss options in worldserver.conf are set to 0

hallgaeuer:
## Changes Proposed:
This PR proposes changes to:

A small durability loss on e.g. death still occurs even when all the "DurabilityLoss" options in worldserver.conf are set to 0.
This happens because the calculation for the actual durability points lost always "rounds up" to 1 point at the minimum.

## Issues Addressed:

- Closes 

## SOURCE:

No source, because this is a non-blizzlike server configuration.

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Set all "DurabilityLoss" options in worldserver.conf to 0 and restart your server
2. Get killed by an enemy of your choice (don't use the .die command, which doesn't seem to incur durablity losses?)
3. Observe that no durability is lost on equipped items

I have also verified that durability is still correctly lost if configured so in worldserver.conf.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4651,7 +4651,7 @@ void Player::DurabilityLossAll(double percent, bool inventory)
 
 void Player::DurabilityLoss(Item* item, double percent)
 {
-    if(!item)
+    if(!item || percent == 0.0)
         return;
 
     uint32 pMaxDurability = item ->GetUInt32Value(ITEM_FIELD_MAXDURABILITY);


heyitsbench:
Would it not be preferable to have the if statement on L4664 check for a case of zero percent?
```cpp
if (pDurabilityLoss && pDurabilityLoss < 1)
```

hallgaeuer:
> Would it not be preferable to have the if statement on L4664 check for a case of zero percent?
> 
> ```c++
> if (pDurabilityLoss && pDurabilityLoss < 1)
> ```

I think that could cause side effects. `pDurabilityLoss` is an int value of the actual durability points lost. 
The condition seems to ensure that there's always at least one point lost, even when the configured percentage lost is very small. 

With your proposed change small durability loss percentages wouldn't lose any durability points anymore I think. 
Which is why I still prefer the early return if the percentage is actually 0.

At least that's my understanding of it. Feel free to correct me if I'm wrong :)

Nefertumm:
I don't like this, using comparision operators between double type (float, etc) is bad.

hallgaeuer:
It's not always "bad". In fact it depends on the use case. Comparing floating points to zero in this way is fine, if the value is not the result of some calculation (which it isn't in this case, it's just a config). 
A quick google seach "c++ comparing float to zero" do reveal plenty of results saying it's fine.
But you're welcome to propose another way.

Nefertumm:
Ah, then it's fine. I thought this was indeed a result of some calculation, that's why I said that. If that's just the config value, then we're good.

Nefertumm (approved):

PR #17416, opened 2023-10-04
fix: worldport command now teleports players to the expected coordinates

natrist:
## Changes Proposed:
This PR proposes changes to:

## Issues Addressed:

1. The worldport command handler's net message reading logic is wrong and the player is not teleported to the requested coordinates
2. Only online players can be teleported

## SOURCE:

## Tests Performed:
This PR has been:

## How to Test the Changes:
1. Log into the game world
2. Execute the worldport command (Usage: worldport \<continentID\> [x y z] [facing])

- [ ] The movecharacter command (NYI) could probably be handled by the worldport command handler later on, as they share the same logic

### DIFF
--- a/src/server/game/Handlers/MiscHandler.cpp
+++ b/src/server/game/Handlers/MiscHandler.cpp
@@ -1091,34 +1091,34 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
     SendPacket(&data);
 }
 
-void WorldSession::HandleWorldTeleportOpcode(WorldPacket& recv_data)
+void WorldSession::HandleWorldTeleport(WorldPacket& msg)
 {
-    uint32 time;
-    uint32 mapid;
-    float PositionX;
-    float PositionY;
-    float PositionZ;
-    float Orientation;
-
-    recv_data >> time;                                      // time in m.sec.
-    recv_data >> mapid;
-    recv_data >> PositionX;
-    recv_data >> PositionY;
-    recv_data >> PositionZ;
-    recv_data >> Orientation;                               // o (3.141593 = 180 degrees)
-
-    if (GetPlayer()->IsInFlight())
-    {
-        LOG_DEBUG("network", "Player '{}' ({}) in flight, ignore worldport command.", GetPlayer()->GetName(), GetPlayer()->GetGUID().ToString());
+    if (!IsGMAccount()) {
+        SendNotification(LANG_PERMISSION_DENIED);
         return;
     }
 
-    LOG_DEBUG("network", "CMSG_WORLD_TELEPORT: Player = {}, Time = {}, map = {}, x = {}, y = {}, z = {}, o = {}", GetPlayer()->GetName(), time, mapid, PositionX, PositionY, PositionZ, Orientation);
+    // READ THE MESSAGE DATA
+    auto eventTime      = msg.read<uint32>();
+    auto continentID    = msg.read<uint32>();
+    auto player         = msg.read<ObjectGuid>();
+    auto position       = msg.read<G3D::Vector3>();
+    auto facing         = msg.read<float>();
 
-    if (AccountMgr::IsAdminAccount(GetSecurity()))
-        GetPlayer()->TeleportTo(mapid, PositionX, PositionY, PositionZ, Orientation);
-    else
-        SendNotification(LANG_PERMISSION_DENIED);
+    // LOOK FOR A PLAYER OBJECT IN WORLD
+    // THAT MATCHES THE PROVIDED GUID
+    // OR USE THE ACTIVE PLAYER
+    Player* playerPtr = player ? ObjectAccessor::FindPlayer(player) : GetPlayer();
+    if (!playerPtr) {
+        return;
+    }
+
+    LOG_GM(GetAccountId(),
+           "Player {} sent command: worldport {} {} {} {} {}",
+           playerPtr->GetName(), continentID, position.x, position.y, position.z, facing);
+
+    // TODO: WorldTeleport(eventTime, continentID, playerPtr, position, facing);
+    playerPtr->TeleportTo(continentID, position.x, position.y, position.z, facing, TELE_TO_GM_MODE);
 }
 
 void WorldSession::HandleWhoisOpcode(WorldPacket& recv_data)
--- a/src/server/game/Server/Protocol/Opcodes.cpp
+++ b/src/server/game/Server/Protocol/Opcodes.cpp
@@ -136,7 +136,7 @@ void OpcodeTable::Initialize()
     /*0x005*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_QUERY_OBJECT_POSITION,                              STATUS_NEVER);
     /*0x006*/ DEFINE_HANDLER(CMSG_QUERY_OBJECT_ROTATION,                                            STATUS_NEVER,      PROCESS_INPLACE,        &WorldSession::Handle_NULL                              );
     /*0x007*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_QUERY_OBJECT_ROTATION,                              STATUS_NEVER);
-    /*0x008*/ DEFINE_HANDLER(CMSG_WORLD_TELEPORT,                                                   STATUS_LOGGEDIN,   PROCESS_THREADUNSAFE,   &WorldSession::HandleWorldTeleportOpcode                );
+    /*0x008*/ DEFINE_HANDLER(CMSG_WORLD_TELEPORT, STATUS_LOGGEDIN, PROCESS_THREADUNSAFE, &WorldSession::HandleWorldTeleport);
     /*0x009*/ DEFINE_HANDLER(CMSG_TELEPORT_TO_UNIT,                                                 STATUS_LOGGEDIN,   PROCESS_INPLACE,        &WorldSession::Handle_NULL                              );
     /*0x00A*/ DEFINE_HANDLER(CMSG_ZONE_MAP,                                                         STATUS_NEVER,      PROCESS_INPLACE,        &WorldSession::Handle_NULL                              );
     /*0x00B*/ DEFINE_SERVER_OPCODE_HANDLER(SMSG_ZONE_MAP,                                           STATUS_NEVER);
--- a/src/server/game/Server/WorldSession.h
+++ b/src/server/game/Server/WorldSession.h
@@ -915,7 +915,7 @@ public:                                                 // opcodes handlers
     void HandleReportPvPAFK(WorldPacket& recvData);
 
     void HandleWardenDataOpcode(WorldPacket& recvData);
-    void HandleWorldTeleportOpcode(WorldPacket& recvData);
+    void HandleWorldTeleport(WorldPacket& msg);
     void HandleMinimapPingOpcode(WorldPacket& recvData);
     void HandleRandomRollOpcode(WorldPackets::Misc::RandomRollClient& packet);
     void HandleFarSightOpcode(WorldPacket& recvData);


heyitsbench:
Gave this a test. It seems the usage is actually `worldport [MapID] [X] [Y] [Z] [O]`. Providing `603 1897.76 115.94 342.38 0`, for example, took me to Ulduar's Descent into Madness (Vezax's room). Interestingly, whereas `.go xyz` fails to teleport me for not being in a raid group, `worldport` does not fail. Without providing any of those parameters and just the `worldport` command, it took me to the gates of Stormwind regardless of race/class. All looks well. 👍

Curious on the source/use case for `worldport`ing offline players though.

natrist:
> Gave this a test. It seems the usage is actually `worldport [MapID] [X] [Y] [Z] [O]`. Providing `603 1897.76 115.94 342.38 0`, for example, took me to Ulduar's Descent into Madness (Vezax's room). Interestingly, whereas `.go xyz` fails to teleport me for not being in a raid group, `worldport` does not fail. Without providing any of those parameters and just the `worldport` command, it took me to the gates of Stormwind regardless of race/class. All looks well. 👍
> 
> 
> 
> Curious on the source/use case for `worldport`ing offline players though.

Yes, that was an oversight on my part. The command indeed takes a mapID as first argument, before the rest of the options. You can also enter a SafeLoc name instead, which is why when you enter nothing it just defaults to the Stormwind GY, as that's probably the first entry in the SafeLoc file.

The worldport command itself does not allow you to teleport other players but the client still sends a null GUID, which is likely because either the GM tool uses the packet for teleporting online players or the handler is also used for the movecharacter command, which does the same thing but on the selected character on the character selection screen.

heyitsbench:
> The worldport command itself does not allow you to teleport other players but the client still sends a null GUID, which is likely because either the GM tool uses the packet for teleporting online players or the handler is also used for the movecharacter command, which does the same thing but on the selected character on the character selection screen.

Ah, I misread the second issue the PR addresses in your description then, I apologize. In that case, I have nothing to complain about. 😄

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1117,36 +1117,56 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
...
-    uint32 time;
-    uint32 mapid;
-    float PositionX;
-    float PositionY;
-    float PositionZ;
-    float Orientation;
+    uint32_t    continentID;
+    uint32_t    requestTimeMs;
+    float       facing;
+    uint64_t    player;
```
Kitzunu, line 1125: remove _t

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1117,36 +1117,56 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
...
+        if (!pPlayer)
+        {
+            QueryResult result = CharacterDatabase.Query("SELECT guid FROM characters WHERE guid={}", player);
+            if (result->GetRowCount())
+            {
+                uint32_t zoneID = MapMgr::instance()->GetZoneId(0, continentID, position.m_positionX, position.m_positionY, position.m_positionZ);
+
+                CharacterDatabase.Execute(
+                    "UPDATE characters SET map={}, position_x={}, position_y={}, position_z={}, orientation={}, zone={} WHERE guid={}",
+                    continentID, position.m_positionX, position.m_positionY, position.m_positionZ, facing, zoneID, player
```
Kitzunu, line 1161: use TeleportTo() instead

natrist: I'm explicitly porting the player offline. Want me to split the logic into a new function?

natrist:
I cleaned up the code and split up the movecharacter command logic into code that I'll submit for another pull request, to keep things tidy.

Here's a video of the movecharacter and worldport commands on WowGM:
https://www.youtube.com/watch?v=r233qne9s6A

Development on the movecharacter command is not yet completed on the client side but the supporting function calls are all implemented.

heyitsbench:
No idea if you already know or if this already supports it, but it does seem the original `worldport` command did work even with commas separating the coordinates. E.g. `worldport 1 16223.8, 16259, 13.2022, 88.6758` (macro from older, I believe 3.0, PTR client)

natrist:
> No idea if you already know or if this already supports it, but it does seem the original `worldport` command did work even with commas separating the coordinates. E.g. `worldport 1 16223.8, 16259, 13.2022, 88.6758` (macro from older, I believe 3.0, PTR client)

It is not supported but I suspect it could be a side effect of the function converting the string tokens to floating point numbers.

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1117,36 +1117,56 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
     SendPacket(&data);
 }
 
-void WorldSession::HandleWorldTeleportOpcode(WorldPacket& recv_data)
+void WorldSession::HandleWorldTeleport(WorldPacket& message)
```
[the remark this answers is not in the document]
Kitzunu: iirc it is just a naming convension that all opcodes have it in the name

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1117,36 +1117,36 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
...
+    Player* playerPtr = player ? ObjectAccessor::FindPlayer((ObjectGuid)player) : GetPlayer();
+
+    if (!playerPtr)
     {
-        LOG_DEBUG("network", "Player '{}' ({}) in flight, ignore worldport command.", GetPlayer()->GetName(), GetPlayer()->GetGUID().ToString());
+        SendNotification("Player not found");
         return;
     }
 
-    LOG_DEBUG("network", "CMSG_WORLD_TELEPORT: Player = {}, Time = {}, map = {}, x = {}, y = {}, z = {}, o = {}", GetPlayer()->GetName(), time, mapid, PositionX, PositionY, PositionZ, Orientation);
```
Kitzunu, line 1144: dont get rid of this, debugs in opcodes are good

src/server/game/Handlers/MiscHandler.cpp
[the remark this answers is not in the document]
natrist: So you want me to put it back as it was, right?

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1117,36 +1117,36 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
...
+    message >> position.m_positionZ;
+    message >> position.m_orientation;
 
-    if (GetPlayer()->IsInFlight())
+    Player* playerPtr = player ? ObjectAccessor::FindPlayer((ObjectGuid)player) : GetPlayer();
+
+    if (!playerPtr)
     {
-        LOG_DEBUG("network", "Player '{}' ({}) in flight, ignore worldport command.", GetPlayer()->GetName(), GetPlayer()->GetGUID().ToString());
+        SendNotification("Player not found");
```
Kitzunu, line 1145:
```suggestion
        SendNotification("WorldTeleport: Player not found");
```

src/server/game/Handlers/MiscHandler.cpp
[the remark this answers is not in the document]
Kitzunu: ye

Kitzunu:
We can get this merged as soon as you figure out the reviews above...

natrist:
> We can get this merged as soon as you figure out the reviews above...

I'm waiting for my other PR that removes redundant network logging calls to get merged. As soon as that's settled I will take another look at the requested changes and reviews.

Kitzunu:
Aight, can you ping me in it and I can take a look

heyitsbench:
> I'm waiting for my other PR that removes redundant network logging calls to get merged.

Your time has finally arrived.

src/server/game/Handlers/MiscHandler.cpp
natrist: I am sending the Player not found notification as that is what the server would normally send on Blizzard realms.

heyitsbench: Is this a debug server-only type string or is this getting sent to players? Because if it's getting sent to players, I feel like there should be some localize-able way to send it.

natrist: Unfortunately it seems like those kinds of things were not localized by Blizzard.

src/server/game/Handlers/MiscHandler.cpp
natrist: I don't see the point but I don't care, so it's done 😂

src/server/game/Handlers/MiscHandler.cpp
```diff
@@ -1090,34 +1090,34 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
...
-    recv_data >> Orientation;                               // o (3.141593 = 180 degrees)
+    //  IF THE PLAYER GUID IS NULL IT MEANS
+    //  THE USER ISSUING THE COMMAND IS THE TARGET
+    Player* playerPtr = player ? ObjectAccessor::FindPlayer((ObjectGuid)player) : GetPlayer();
 
-    if (GetPlayer()->IsInFlight())
+    if (!playerPtr)
     {
-        LOG_DEBUG("network", "Player '{}' ({}) in flight, ignore worldport command.", GetPlayer()->GetName(), GetPlayer()->GetGUID().ToString());
+        SendNotification("Player not found");   //  DO NOT TRANSLATE
```
Nyeriah, line 1112: What's special about this string specifically that it should not be translated?

Kitzunu: Also, you should prefix it with "WORLDPORT: ...."

src/server/game/Handlers/MiscHandler.cpp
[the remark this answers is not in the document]
Kitzunu: Dont know why recv_data was ignored

src/server/game/Handlers/MiscHandler.cpp
natrist: In modern internal builds, the client has an extra message handler that displays the console message _Player not found_.
The message does not seem to exist as of the LK alpha/beta builds and no handler can be found for it in patch 3.3.5a.

It is safe to assume the functionality was not present in patch 3.3.5a.
However, we can imagine that Blizzard would still display the error message, as many commands could in theory use it. Add to that the fact that Blizzard sends notifications when cheats are toggled on and off and the theory starts to make sense.

src/server/game/Handlers/MiscHandler.cpp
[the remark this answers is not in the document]
natrist: I'm confused. Do you want me to rename the function parameter as well?

src/server/game/Handlers/MiscHandler.cpp
Nyeriah: I appreciate the attention to detail but mimicking their internal builds is not something we aim for when it comes to something that adds no value to the project, like a single error string not localised

Kitzunu: We try to mimic gameplay as much as possible. But admin/gm features etc sometimes can be fun to try and mimic, but errors strings like these are no point in mimicing

sudlud:
Closing this for now as it has been abandoned for quite some time.

Please feel free to re-open and update the PR again.

Issue #8608, opened 2021-10-18
Triple truesilver node at Desolace


jamadaha:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [X] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1736

A trio of nodes spawn quite next to each other.
![Screenshot from 2021-09-14 22-45-07](https://user-images.githubusercontent.com/84640316/133487722-fcee8f7d-670a-4447-8b0a-58b9285f15ba.png)

Tester: confirmed.
It is the ores, as seen in the provided picture, 9256, 9257 and 9258

### Expected Blizzlike Behaviour

I think these should be tied that only one of the three should be up at a given moment.

Testers note:
These should either be moved, or in a pool. Currently they are not in any kind of pool, which is probably incorrect, but not really new knowledge.

### Steps to reproduce the problem
1. .go o 9256

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/7e2e6f8ee85f82f57a3c6e0f6f645b4c42292e7a

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
[...]

PR #17444, opened 2023-10-07
fix(DB/Gameobject): Pool triple Truesilver Vein in Desolace

avarishd:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8608
- Closes https://github.com/chromiecraft/chromiecraft/issues/1736

## SOURCE:

## Tests Performed:

This PR has been:
- [] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

.go ga 9256,9257,9258
Only one should be spawned at a time

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1696691443957086100.sql
+++ b/data/sql/updates/pending_db_world/rev_1696691443957086100.sql
@@ -1,9 +1,32 @@
 --
-DELETE FROM `pool_template` WHERE `entry`=11699;
-INSERT INTO `pool_template` (`entry`, `max_limit`, `description`) VALUES (11699, 1, 'Truesilver Vein - Desolace');
+DELETE FROM `gameobject` WHERE `guid` IN (9266,9268,9270,9272,9273,9274);
+INSERT INTO `gameobject` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `position_x`, `position_y`, `position_z`, `orientation`, `rotation0`, `rotation1`, `rotation2`, `rotation3`, `spawntimesecs`, `animprogress`, `state`, `ScriptName`, `VerifiedBuild`, `Comment`) VALUES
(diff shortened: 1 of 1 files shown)


neifion-00000000:
so the way these work is actually more complicated than this, but this would be an improvement over existing so might want to be considered regardless, or updated to share with a mithril node at a low %.

TLDR is TS/rare veins overlap other mining veins and any genuine sniffed location of a TS vein location is also that of another mine (in this specific case prolly mithril, but it can place hold with other mines as well)

avarishd:
Possibly iron too, "judging" by wowhead. (Yes, I know)

avarishd:
Something like that? @neifion-00000000

neifion-00000000:
this will work I think, but you could potentially do it all as a single pool with chance if you'd like for this specific case as you are only trying to make 1 of those spawn at once right?  I do believe this is the correct method anyway, all evidence I have supports that its much more likely things are all on their own individuals pools rather than large pools with multiple spawns, basically all but proven

prolly don't need to go back and change it but so you know, the reason they do pool of pool is to allow multiple picks of pooled objects without the overlap

avarishd:
That was useful information. Thanks!
But this will cause to sometimes (most of the time) to not have anything spawned, which I guess would be the intended behaviour?

One question though, how/where do you come up with the chances?

neifion-00000000:
> That was useful information. Thanks! `Pool Id 11699 has no equal chance pooled entites defined and explicit chance sum is not 100. The pool will not be spawned.`
> 
> One question though, how/where do you come up with the chances?

They were your values I just split them in 3 cuz there's 3 locations?  I didn't check the math tho so I might not have added up to 100 properly and been 99.99 or something.

avarishd:
[The error was caused because they weren't equally chanced.](https://www.azerothcore.org/wiki/pool_gameobject#chance)

Nyeriah (approved):

PR #17460, opened 2023-10-08
ResurrectPlayer add 'OnBeforePlayerResurrect' hook

### LINKED ISSUE #17370: Feature: Add core hooks for altering how death/resurrection works.

Add hooks to change or alter how death works.
As hardcore servers are becoming more popular it would be great to have the option to keep someone as a ghost perm.
The hook could potentially change how the resurrect button works and make that customisable where it may not change the text but the fuction would be different, example would be that pressing the button would just teleport the character back to the graveyard.

If the hooks are present in the core then we can change how it works with modules hopefully.
I'm not technically minded in this area but hoping that someone could see some value in this feature!

### Additional context

_No response_

### LINKED ISSUE #17388: Feature: Add onBeforePlayerResurrect call in the ResurrectPlayer method

In Player:: ResurrectPlayer, there is currently a method call of sScriptMgr ->OnPlayerResurrect (this, restore_percent, applySickness), but this method is called after the resurrection takes effect, and has no ability to interfere with whether the player role can be resurrected. I think it is necessary to add an onBeforePlayerResurrect method before processing the resurrection logic, and decide whether to execute resurrection according to the method return value, This can easily implement features such as hard core mode

_No response_

### Additional context

_No response_

netweaver:
## Changes Proposed:

This PR proposes changes to:

Adds the PlayerScript OnBeforePlayerResurrect() hook at the begining of  Player::ResurrectPlayer() function, triggered before player resurrection proceed.  Resurrection will not be applied when It returns false. Thus, module can control whether a player can be resurrected or not  by implement the hook handler.

The original intention was to easily imlement a hardcore mode similar to the official Steel Soul through the Eluna module. Therefore, after this PR merge, I will submit  "PLAYER_EVENT_BEFORE_RESURRECT" PR to the Eluna module (which has already been tested but, you know, due to code dependencies).

Notice： This implementation is not perfect (although it does not affect the actual prevention of resurrection). I found that there are different pre-processing steps when players are resurrected under different circumstances, but they are scattered in different processing logic. For example, when performing a "Corpse Run" resurrection, the tombstone disappears first (surprisingly, not after resurrection). Some situations have already been identified, and they may potentially cause issues. I will list them in the known issues section below.

I haven't traced every instance of code calling for player character resurrection, but I believe that placing this hook in various processing logics may not be the best choice. As part of the proposed solution, it includes these two aspects:
1. Making Player::ResurrectPlayer() return a Boolean value.
2. Moving some of the pre-processing steps after the resurrection logic and determining whether to perform additional processing based on its return value.

addtional: fix 'IsValidBoolScript' to call all registered scripts , for example , both module A and  module  B implement playerScript, may handler the same hook ,  only one module will be triggered before.

## Issues Addressed:

- Closes #17370
- Closes #17388 

## Tests Performed:

This PR has been:

## How to Test the Changes:
I tested it using Lua scripts (through the Eluna module), but it should also be relatively straightforward to test by writing a script class for implementing playerScript.

- When resurrecting via the "Corpse Run" method and preventing the resurrection, both the corpse and the tombstone disappear. This prevents the possibility of resurrecting via this method when conditions allow it later.
- When a Death Knight dies in their starting area and is resurrected by the "Valkyr", if the resurrection is prevented, the character will not be in the ghost state (appearing as if resurrected but not actually). However, the character becomes stuck and unable to move. Using the "unstuck" button in the Help interface returns the character to the graveyard in ghost form.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -4385,6 +4385,19 @@ void Player::BuildPlayerRepop()
 
 void Player::ResurrectPlayer(float restore_percent, bool applySickness)
 {
+    LOG_DEBUG("entities.player", "Player::ResurrectPlayer: enter Resurrecting player {} ({})", GetName(), GetGUID().ToString());
+
+    //call the beforePlayerResurrect script first
+    //if false is returned, no resurrection will occur
+    if (!sScriptMgr->OnBeforePlayerResurrect(this, restore_percent, applySickness))
+    {
+        LOG_DEBUG("entities.player", "Player::ResurrectPlayer: OnBeforePlayerResurrect returned false for player {} ({})", GetName(), GetGUID().ToString());
+        return;
+    }
+
+    LOG_DEBUG("entities.player", "Player::ResurrectPlayer: begin Resurrecting player {} ({})", GetName(), GetGUID().ToString());
+        
+
     WorldPacket data(SMSG_DEATH_RELEASE_LOC, 4 * 4);        // remove spirit healer position
     data << uint32(-1);
     data << float(0);
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -1487,6 +1487,26 @@ void ScriptMgr::OnSetServerSideVisibilityDetect(Player* player, ServerSideVisibi
 //    });
 //}
 
+bool ScriptMgr::OnBeforePlayerResurrect(Player* player, float restore_percent, bool applySickness)
+{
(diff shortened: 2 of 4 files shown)


Winfidonarleyan:
https://github.com/WarheadCore/WarheadBand/commit/d0f8f1ec4380254bdf0b12623ce3e04079e685a4 for example

netweaver:
why codestyle check shows 'Remove whitespace at the end of the lines above' 
I checked those files ,there are no whitespaces exist

netweaver:
I am working on those know issues,  undering testing now

Kitzunu:
Status

Nyeriah:
It seems like this PR has been abandoned. Please request a reopen if you wish to continue working on this

Issue #17439, opened 2023-10-07
The CPU usage rate of the worldserver process is 105%


moncat2005:
### Current Behaviour

The CPU usage of the worldserver process compiled by the acore version submitted before September 28, 2023 is approximately 30%.
The CPU usage of the worldserver process compiled by the acore version submitted after September 29, 2023 has increased to 105%.
It may be related to the introduction of grid functionality.
The configuration parameters of worldserverconf have also been typeset and sorted between these two progressive versions.
minworldupdatetime parameter in the configuration file has been reset to the same 10
In addition, the three parameters of visibility.distance remain consistent with the previous version.

### Expected Blizzlike Behaviour

Resolve abnormal CPU usage issues

### Steps to reproduce the problem

At present, it has been found that the version rolled back to September 28th has returned to normal.

### AC rev. hash/commit

AzerothCore rev. a2c6af4c5683+ 2023-10-01 07:26:31 +0000 (master branch)

### Operating system

Ubuntu 18.04

### Custom changes or Modules

Eluna module

Kitzunu:
@AGandrup Could you take a look, I think it all is surrounding the visibility change. I can't pin point the actual issue myself atm... and if we don't get it fixed within the next few days we need to revert it for the time being

kissingers:
Yes , https://github.com/azerothcore/azerothcore-wotlk/pull/15919  this only fix a very little problem, but case the server 300% CPU usage

kissingers:
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/33601959/5612b07f-6def-4cad-9af6-3f116f0f0e9a)
first is update the https://github.com/azerothcore/azerothcore-wotlk/pull/15919
second is npcbots with 200 npc and without update the patch
third is clear server without update the patch

avarishd:
That explains why AC has been so hard to run.

mpfans:
Encountered the same issue after updating this PR https://github.com/azerothcore/azerothcore-wotlk/pull/15919

heyitsbench:
@mpfans, would you be able to test #17480?

mpfans:
> @mpfans, would you be able to test #17480?

Tested without improvement

AGandrup:
> > @mpfans, would you be able to test #17480?
> 
> Tested without improvement

That sounds very odd to me if you dont see any improvement at all, because that PR stops gameobjects as well as transport from being set to active. What tests have you done?

[...]

Issue #16668, opened 2023-07-04
[server]world server crash 0630


mpfans:
### Current Behaviour

crash log:

https://gist.github.com/mpfans/85034978f5cefb4aa77afa18bac4c077

### Expected Blizzlike Behaviour

none

### Steps to reproduce the problem

none

### AC rev. hash/commit

6b846df6a36e4ce72e9c26313d14aae70bad33e5

### Operating system

Ubuntu20.04

### Custom changes or Modules

mod-anticheat

PR #17480, opened 2023-10-10
fix(Core/Grid): Address bugs and performance issues introduced by visibility notifier implementation

AGandrup:
- Corrected std::chrono from seconds to milliseconds
- Got rid of leftover code that caused objects to not show up on time

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


heyitsbench:
Based on your updates in the linked PR thread, this should resolve (at least partially) the erroneously increased CPU usage as well as poorly timed gobject updates. Does this PR at all touch the freezing issue mentioned as well in that thread that you know of?

AGandrup:
> Based on your updates in the linked PR thread, this should resolve (at least partially) the erroneously increased CPU usage as well as poorly timed gobject updates. Does this PR at all touch the freezing issue mentioned as well in that thread that you know of?

I am not really sure if it had any impact on the CPU usage - something tells me it didn't, unfortunately.
In terms of the freeze issue, I haven't experience that at all yet, so I am still not sure what that could be. If you know any specific cases where these freezes usually occur, please let me know.

Nyeriah:
Freezes usually happen in the morning, so some event

Nyeriah:
We suspect it’s the transport boats event, when it arrives

AGandrup:
Status: Good news everyone, I have managed to cut down the CPU usage to something that looks a bit more reasonable.
 
image

God knows if I have broken something else in the process, but I guess time will tell. I need to clean up and double check the changes I have made, so I wont be able to push anything before sometime tomorrow, so stay tuned.
If this fixes the freezing issue or not I have no idea - still haven't experienced any, so I can't tell.
@heyitsbench @Nyeriah @Kitzunu

iipamonkey:
Regarding high CPU utilization, in bug #17439 @moncat2005 reported that rolling back to the September 28th build eliminates the high CPU utilization. Shouldn't that help narrow down the issue quite a bit?

src/server/game/Maps/Map.cpp
```diff
@@ -1015,8 +1007,6 @@ template<>
 void Map::RemoveFromMap(MotionTransport* obj, bool remove)
 {
     obj->RemoveFromWorld();
-    if (obj->isActiveObject())
-        RemoveFromActive(obj);
```
AGandrup, line 1019: See comment above.

AGandrup:
The issue seems to be related to gameobjects being set to "active". The last commit removes that logic. TC does not seem to have that logic either.

In my search to find the cause of these issues, I have found other differences between AC and TC related to visibility notifier, which I feel needs to be addressed as well as it could obviously also have a say in performance and what not. But the rabbit hole goes deep, so I'm gonna need some more time.

AGandrup:
Status: CPU usage is reaching a point that feels pretty good, imo - see picture below. However, I noticed that when I go to Wintergrasp, it tend to ramp up pretty quickly and is really slow to ramp down again, which is not really that satisfying. I assume it has something to do when buildings, walls and stuff are spawned/added to the map. So that is what I am looking into atm.

image

pklloveyou:
image
It's really low, unbelievably low
One difference is that I changed std::chrono::milliseconds to uint32

AGandrup:
> image It's really low, unbelievably low One difference is that I changed std::chrono::milliseconds to uint32 Actually, I feel that time_t would be better

time_t is what it originally was, but I was told to use std::chrono instead. If it makes sense I can go back to using time_t.

Kitzunu:
If chrono really is that resource heavy, then definetly go back to time_t

AGandrup:
> If chrono really is that resource heavy, then definetly go back to time_t

It is not - at least not according to my test. However there is a chance that I haven't been using it correctly. When switching to chrono I had to invoke .count() on some things and apparently this returns microseconds (?) Untill I figure this out, im just going to revert back to time_t.

I have found something else which drastically lowers cpu usage, and funny enough it comes down to a single method call which for some reason has been outcommented in AC. Im gonna push it later and we can talk about it. That and also transport being set to active in AC causes trouble.

wsntrl:
You are mixing **chrono::duration** and **time_t** types, which is a very bad idea by itself.
Only **chrono::seconds.count** can be directly converted to **time_t**, **chrono::milliseconds.count** must be divided by **1000**.

AGandrup:
Status: Just as I thought I couldn't get CPU usage any lower. lol wtf.

image

I don't know what CPU usage was before, but this seem pretty decent if I may say so myself. Still testing a lot, for bugs and what not so perhaps this is too good to be true, idk yet 😬

AGandrup:
> You are mixing **chrono::duration** and **time_t** types, which is a very bad idea by itself. Only **chrono::seconds.count** can be directly converted to **time_t**, **chrono::milliseconds.count** must be divided by **1000**.

Sorry, missed your message. But gotcha, knew I was missing something. Just gonna stick to time_t for now though.

Nefertumm:
Btw https://gist.github.com/Nefertumm/22705ebb1d9cdff40d942abf6f7c8446 this is one of the crash we had once the commit was up on our server (we reverted it locally after a good couple of them). If you could take a look, great.

AGandrup:
Alright, I haven't made a status in a while so its about time.
This has kinda evolved into something more than just implementing visibility notifier - not because I want to but because I had to 😝 As I've said before, the rabbit hole goes deep and I've been trying hard not to get too much out of scope with this PR, though theres a few places where I couldn't help myself. Its just after the first PR which went horrible with all the issues, I realized I couldn't cut any corners.
Therefore you might see changes with this latest push and think to yourself "what does that have to do with anything" but trust me it does in one way or another. The spaghetti code is real. But if you are still sceptical, feel free to ask ofc.

Here is some highlights of what have been done:
- Removed/reduced unnecessary usage of setActive method which caused grids to never unload which further meant that creatures and objects etc. kept updating.
- Removed unnecessary calls to LoadGrid and EnsureGridCreated method etc. Added checks around some of them to make sure they aren't being called if grid is already loaded. An example, transports no longer calls LoadGrid  - players will make sure of that.
- Removed unnecessary calls to PlayerRelocation, CreatureRelocation, GameObjectRelocation and DynamicObjectRelocation methods. In relation to this, I had to add new logic to handle movement behaviour on transports as well how passengers are loaded and unloaded. And also new logic to handle movement for creatures (see cherry pick below). 
- Implemented missing SetFarVisible method (Get-method was already there)
- Added (and removed) checks in some places to help performance - probably easier if I leave a comment in code.
- A lot of refactoring and other stuff I can remember right now 😛
- Cherry-picked https://github.com/TrinityCore/TrinityCore/pull/28884
- Cherry-picked https://github.com/TrinityCore/TrinityCore/pull/24049

Something to take note of: It is very important not to just blindly set creatures, objects or whatever to active. This applies to both scripts in core as well as smart scripts in database. A rule of thumb is to NEVER set anything active on continents unless it is absolutely necessary for example for some event which has to keep running regardless of nearby players or for npcs such as fel reaver that roams the land. If something is set to active make sure to remove it from being active again at some point if possible either by setActive(false) or despawn. Otherwise grids with those objects in them will never unload and will keep updating which obviously will affect performance. For instances and bgs it does not matter as much.

AGandrup:
Anyway, if this doesn't fix the performance issues I honestly don't know what will.

Please test it out and let me know what you think.

@Nyeriah @Kitzunu @heyitsbench tagging you guys for visibility (no pun intended)

Fireevil:
Thank you for your work, I will also help with the test

AGandrup:
> Btw https://gist.github.com/Nefertumm/22705ebb1d9cdff40d942abf6f7c8446 this is one of the crash we had once the commit was up on our server (we reverted it locally after a good couple of them). If you could take a look, great.

Sorry, missed this comment as well. Will look into it tomorrow but could you try with the new changes?

AGandrup:
> Thank you for your work, I will also help with the test

Sweet, thank you! I will go to bed now tho - its 5 am here 🥴

sudlud:
With all the issues related to this PR right now I'm wondering
- what is the expected benefit from this implementation?

AGandrup:
> Thanks for the update! With all the issues earlier related to this PR right now I'm trying to understand: what is the expected benefit from this implementation?

To begin with (with the first PR) it was really only at matter of fixing an issue where NPCs (typically large ones like Fel Reaver) as well as gameobjects, would only be visible at a very short distance. After looking into it, I learned that AC had their own visibility notifier system implemented with very limited visibility range and no way of adjusting any of it in config file. So after having discussed it on Discord it was concluded that it could be worth implementing TC's system - not because it would be necessarily better, but rather just to be more aligned with TC, and also we knew that TC didnt have any problems with visibility so rather than trying to fix AC's current system, we figured it would be just as easy to swap it out with TC's 😛 That is really the story and motivation behind the first PR. 

Now, me being super navie and tunnel visioned on getting visibility to work, I pretty much thought (hoped) I could just search the places with AC's current visibility logic and swap it with TC's. But in order to do that I also had to touch on other parts of the grid system (hence why for example grid clean up was also added). And to be fair, it worked but I had not give performance a single thought because everything seemed to run smoothly as it used to. Well guess what, it didnt. Turns out that you dont just mess around with the grid system and expect it to work, so here we are.

So now I have tried to do my best fixing those issues with performance and tried not to tunnel visioned on that, meaning whenever I have touched something, say transports, I have made sure to test that over and over before applying any changes. This is also why changes haven't been coming in chunks of small commits. That said, I do not promise that none of this doesnt introduce new bugs, so that is why I need someone to review and test it.

sudlud:
Thanks for the insight!

Nyeriah:
@AGandrup Even if there's only a slight improvement from the current situation we can merge it and continue it from here. Unfortunatelly I won't be able to test it in an environment that grants us any relevant feedback

AGandrup:
> @AGandrup Even if there's only a slight improvement from the current situation we can merge it and continue it from here. Unfortunatelly I won't be able to test it in an environment that grants us any relevant feedback

Sure, if you have enough trust in it, feel free to do so. However I would love if someone with at least 50 players, preferably more, could test it out first. And I also welcome anyone to check creature movement behaviour as well as transport and movement on transport, to see if they notice any difference. Transports include zeppelin, boats and also elevators.
And obviously visibility and loading of all sorts of objects.

AGandrup:
@pklloveyou sorry for the tag, but any chance you could test again as well? 😬 🙏

Nyeriah:
Please update the branch, the CI is stuck

Nyeriah (approved):

Kitzunu (approved):

AGandrup:
Been looking throug the crash reports:
https://gist.github.com/Nefertumm/22705ebb1d9cdff40d942abf6f7c8446
https://gist.github.com/Nyeriah/84748d422ed2b4fca35b4e746b04a10d
And I think maybe it has something to do with i_visibleNow being a vector which can contain duplicates. Originally it was a set but was changed to a vector at some point- why? I do not know. We only ever loop over it, so I dont see a reason for it to be a vector. I went back in commit history but it has been like that since the first AC commit, so ye idk. Reverted it back to a set so we at least know it only contain unique elements.

Kitzunu:
Probably some shit from sunwell :D

AGandrup:
> Probably some shit from sunwell :D

Ye I guess lol. But i'm not entirely sure that it is what is causing the crash because you could ask why it hasn't been an issue before. I haven't experienced any crashes at all myself though, so chances are also it has been fixed through the many other changes. But I guess we'll just have to wait and see. Either way, I think using set is the proper way to go.

AGandrup:
Its possible that this PR together with the first PR, fixes this crash from 3 months ago. https://github.com/azerothcore/azerothcore-wotlk/issues/16668 It seems to be related to unloading/cleanup of grids

Kitzunu:
Maybe 🤷 Not seen a crash report since it was opened so might be local changes causing it too

AGandrup:
Fair, I just noticed it

PR #17487, opened 2023-10-11
feat(DB): Add column for comment (spell_cooldown_overrides)

avarishd:
Merge after https://github.com/azerothcore/azerothcore-wotlk/pull/17486 please.

Question - How do we deal with this .sql, since it cannot be rerun?


Kitzunu:
I do not see the point with a comment field just to specify the spell name. Spells are identified by their spellid, this is not needed

avarishd:
Spells are, where they are used at, not.

Kitzunu:
I still don't see the point.

What makes it more needed in this table versus spell_proc_event or spell_custom_attr or any other spell table?

Look, if people want it I won't be a bitch about it. I just want reasons and arguments when implementing stuff

avarishd:
Think of comment columns as an optional QoL, they are there if you need them, can always hide them if you don't.
Can wait for more feedback, not a big deal if it doesn't go through.

IntelligentQuantum:
But I think it's good because you don't need to go to the game to check

Nefertumm:
I'm not against this one as you could default it to null if you don't use it.

Kitzunu (approved):

PR #17398, opened 2023-10-01
fix (DB/SAI): Stratholme/Baron Rivendare (10440) encounter fixes

kjack9:
## Changes Proposed:
This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [X] Database (SAI, creatures, etc).

During the Baron Rivendare (10440) encounter, six Mindless Skeletons (11197) are spawned from the bone piles around the room. In the current version of AC, often 1 or more of the skeletons do not appear properly and fall through the floor and/or wall. This causes Baron to sacrifice them and heal himself, increasing the difficulty of the encounter significantly.

This change moves those spawn points slightly forward and on the flat ground in front of the bone piles. This appears to give them skeletons the proper footing to not fall through the floor anymore. Interestingly, the skeletons still appear to visually come from the top of the bone piles, so I'm not sure if the actual spawn point is offset from the one in `spell_target_position` or if there's something else at play. Regardless, this DB change resolves the issue.

Additionally, I added a SAI trigger to despawn the summoned adds when Baron Rivendare is reset. This resolves some awkward silence between the Baron and his sacrificial followers if all players die.

## Issues Addressed:

- Closes #12257
- Closes https://github.com/chromiecraft/chromiecraft/issues/3709

## SOURCE:

The changes have been validated through:
- [X] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [X] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1. `.go xyz 4032.49 -3357.65 115.06 329 1.52`
2. Engage Baron Rivendare (10440)
[...]

Issue #12257, opened 2022-07-01
[Stratholme] Baron Rivendare spawns Mindless Skeletons underground


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3709

### What client do you play on?

enUS

### Faction

- [ ] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [X] 60

### Current Behaviour

Some of the [Mindless Skeleton](https://wowgaming.altervista.org/aowow/?npc=11197) spawned during the Baron 
Rivendare encounter appear below the ground.

### Expected Blizzlike Behaviour

Mobs should spawn above ground so the group can deal with them.

### Source

.

### Steps to reproduce the problem

1. .go c id 10440
2. wait 20s (11s initially) for the Raise Dead spell to be cast
3. sometimes one or more of the 6 summoned skeletons will spawn underground

### Extra Notes

![image](https://user-images.githubusercontent.com/61170874/176596411-808d3a41-3db9-430b-be8e-5f7f8932d5fe.png)

Baron Rivendare is scripted in SAI. The spell used for the summoning is 17475

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/e92428940539d178e08012db6e2681690d2308aa

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
[...]

PR #17506, opened 2023-10-14
Fix (core/SpellEffects): Don't randomize summoned guardian position if target is in the DB

kjack9:
## Changes Proposed:

This PR proposes changes to:

Adds an extra condition to the position randomization function for summoned guardian. This condition ensures that if the target for the summon is from the DB (`spell_target_position`) that it will appear exactly at that position every time.

## Issues Addressed:

- Needed for #17398
- Related to #12257
- Related to https://github.com/chromiecraft/chromiecraft/issues/3709

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

Before applying the PR:
1. `.go xyz 4032.49 -3357.65 115.06 329 1.52`
2. Engage Baron Rivendare (10440)
3. Observe that the first wave of skeletons appear exactly at the spell effect on each bone pile
4. Observe that subsequent waves appear in a radius around this point, sometimes causing skeletons to be summoned outside the room (through the wall)
5. Observe that the corpses from the skeletons are in different positions each time (because they are being summoned into random positions around the target position)

image

After applying the PR:
1. `.go xyz 4032.49 -3357.65 115.06 329 1.52`
2. Engage Baron Rivendare (10440)
3. Observe that all waves of skeletons appear exactly at the spell effect on each bone pile, in the same spot each time. None of them disappear into the wall.
4. Observe that the corpses from the skeletons overlap near-perfectly because they are being summoned to the exact target position each time

image

None

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -6061,8 +6061,7 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const* 
         // target position is specified in the DB AND the effect has no or zero radius
         if ((totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET) ||
             (GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB &&
-                (!GetSpellInfo()->Effects[i].HasRadius() || GetSpellInfo()->Effects[i].RadiusEntry->RadiusMax == 0))
-           )
+            (!GetSpellInfo()->Effects[i].HasRadius() || GetSpellInfo()->Effects[i].RadiusEntry->RadiusMax == 0)))
         {
             pos = *destTarget;
         }
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -6058,11 +6058,11 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const* 
 
         // xinef: do not use precalculated position for effect summon pet in this function
         // it means it was cast by NPC and should have its position overridden
-        // unless the destination was specified in a DB entry
+        // unless the target is a position specified in the DB
         if
         (
             (totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET) ||
-            (GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB)
+            (!GetSpellInfo()->Effects[i].HasRadius() && GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB)
         )
         {
             pos = *destTarget;
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -6057,13 +6057,12 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const* 
         Position pos;
 
         // xinef: do not use precalculated position for effect summon pet in this function
-        // it means it was cast by NPC and should have its position overridden
-        // unless the target is a position specified in the DB
-        if
-        (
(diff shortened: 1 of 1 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -6058,7 +6058,12 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const*
 
         // xinef: do not use precalculated position for effect summon pet in this function
         // it means it was cast by NPC and should have its position overridden
-        if (totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET)
+        // unless the target is a position specified in the DB
+        if
+        (
+            (totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET) ||
+            (!GetSpellInfo()->Effects[i].HasRadius() && GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB)
+        )
```
Nefertumm, line 6066:
```suggestion
        if ((totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET) ||
            (!GetSpellInfo()->Effects[i].HasRadius() && GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB))
```

kjack9: I updated the format but added a new condition for a zero-radius spells. Trying to walk the line between compact and unreadable. I still prefer the original format (I think breaking it out is way easier to read), but let me know what you think of this hybrid approach.

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -6057,8 +6057,12 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const*
...
 
         // xinef: do not use precalculated position for effect summon pet in this function
-        // it means it was cast by NPC and should have its position overridden
-        if (totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET)
+        // it means it was cast by NPC and should have its position overridden unless the
+        // target position is specified in the DB AND the effect has no or zero radius
+        if ((totalNumGuardians == 1 && GetSpellInfo()->Effects[i].Effect != SPELL_EFFECT_SUMMON_PET) ||
+            (GetSpellInfo()->Effects[i].TargetA.GetTarget() == TARGET_DEST_DB &&
+                (!GetSpellInfo()->Effects[i].HasRadius() || GetSpellInfo()->Effects[i].RadiusEntry->RadiusMax == 0))
+           )
```
Kitzunu, line 6065:
```suggestion
            (!GetSpellInfo()->Effects[i].HasRadius() || GetSpellInfo()->Effects[i].RadiusEntry->RadiusMax == 0)))
```

kjack9:
Any questions or concerns remaining on this PR before it's merged? Should fix Baron and any other fight where the summoning spells are targeted in the DB. Let me know if there are any other changes you'd like to see.

elthehablo (approved):

elthehablo:
approved. but am thinking whether this is fully backwards compatible. I cannot really imagine it would break something though

kjack9:
> approved. but am thinking whether this is fully backwards compatible. I cannot really imagine it would break something though

I did think about that some. I think the worst case scenario is that some event was relying on summoned guardians being spread out when they're created if they're using the same spawn location over and over again. The only effect from that would be visual - they'd show up on top of each other. And if they REALLY want that behavior, the right move is to change the radius of the spell, not to rely on buggy behavior.

If it does show up in the issue list, punt it my way and I'll own getting it fixed.

Thanks for reviewing.

elthehablo:
> > approved. but am thinking whether this is fully backwards compatible. I cannot really imagine it would break something though
> 
> I did think about that some. I think the worst case scenario is that some event was relying on summoned guardians being spread out when they're created if they're using the same spawn location over and over again. The only effect from that would be visual - they'd show up on top of each other. And if they REALLY want that behavior, the right move is to change the radius of the spell, not to rely on buggy behavior.
> 
> If it does show up in the issue list, punt it my way and I'll own getting it fixed.
> 
> Thanks for reviewing.

Well put :)

PR #17532, opened 2023-10-17
fix(scripts/TempleOfAhnQiraj): Remove spell "Twin Empathy" (1177) that is not present on live

kjack9:
## Changes Proposed:

This PR proposes changes to:

This PR removes the spell "Twin Empathy" (1177) from the Twin Emperors encounter and the `spell_dbc` database.

The spell "Twin Empathy" was used [in WoW Classic](https://www.wowhead.com/classic/spell=1177/twin-empathy) to cause damage to one emperor to equally affect the other (informed conjecture). At some point between its original implementation and WotLK, [the spell was removed from the game](https://www.wowhead.com/wotlk/spell=1177). It no longer appears in the game's DBCs (confirmed with WoW Spell Editor). The only reason it exists in AC today is because of its entry in `spell_dbc` (shoutout to @AnchyDev).

I took a sniff on retail for this encounter and confirmed that no spell is being cast between the two emperors when health decreases.

```
ServerToClient: SMSG_UPDATE_OBJECT (0x27BF) Length: 238 ConnIdx: 1 Time: 10/17/2023 01:59:24.047 Number: 1051
NumObjUpdates: 5
MapID: 531 (531)
HasRemovedObjects: False
Data size: 227
[...]
[1] UpdateType: Values
[1] Object Guid: Full: 0xREDACTED Creature/0 R3020/S29775 Map: 531 Entry: 15275 Low: 3021144
[1] EntryID: 15275
[...]
[1] Health: 809095
```
Time passes in which there are no spells cast between the emperors.
```
ServerToClient: SMSG_UPDATE_OBJECT (0x27BF) Length: 429 ConnIdx: 1 Time: 10/17/2023 01:59:25.374 Number: 1145
NumObjUpdates: 2
MapID: 531 (531)
HasRemovedObjects: False
Data size: 418
[...]
[1] UpdateType: Values
[1] Object Guid: Full: 0xREDACTED Creature/0 R3020/S29775 Map: 531 Entry: 15275 Low: 3021144
[1] Health: 807315
```

This would seem to indicate that the health sync is being done entirely server-side without using any spell. This PR replicates that functionality in AC.

I find no evidence in AC's source files that any other encounter is using this spell.

((Side note: I found this issue while researching why AutoBalance was scaling the Twin Empathy spell. This PR gets closer to Blizzlike while having the happy side effect of fixing the issue with AutoBalance as well.))

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. `.go creature 144285`
2. Engage the Emperors
3. Target one of the Emperors and do `.damage 100000`
4. Note that damage decreases by 100k on both bosses
5. Do `.damage 10000000`
6. Note that both emperors die as expected

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1697524019251340056.sql
+++ b/data/sql/updates/pending_db_world/rev_1697524019251340056.sql
@@ -1,0 +1,2 @@
+-- Remove unneeded spell "Twin Empathy" (1177)
+DELETE FROM `spell_dbc` WHERE `Id`=1177;
--- a/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_twinemperors.cpp
+++ b/src/server/scripts/Kalimdor/TempleOfAhnQiraj/boss_twinemperors.cpp
@@ -25,7 +25,6 @@
 enum Spells
 {
     // Both
-    SPELL_TWIN_EMPATHY            = 1177,
     SPELL_TWIN_TELEPORT_1         = 800,
     SPELL_TWIN_TELEPORT_VISUAL    = 26638,
     SPELL_HEAL_BROTHER            = 7393,
@@ -114,7 +113,7 @@ struct boss_twinemperorsAI : public BossAI
             {
                 float dmgPct = damage / (float)me->GetMaxHealth();
                 int32 actualDmg = dmgPct * twin->GetMaxHealth();
-                twin->CastCustomSpell(twin, SPELL_TWIN_EMPATHY, &actualDmg, nullptr, nullptr, true);
+                twin->SetHealth(twin->GetHealth() - actualDmg);
             }
         }
     }


Nyeriah:
Server side spells don’t show in sniffs however

Nyeriah:
I have two issues with this PR:

1) spell_dbc is read only, so we don't delete or insert new rows in there. The only times its legit for us to delete things, is when we identify its pollution from azerothcore former contributions/sunwell. However, it is not the case here, since I've just checked a clean trinitycore install and the spell is in the DB, so it is a legit server side spell

2) As aforementioned, server side spells don't show in sniffs, so the fact you didn't see it in your parse doesn't add much to your claim

kjack9:
Awesome, I'm learning a lot here! I made an assumption that self-casts for a creature would show up in a sniff, but if that is not the case then I can't conclusively prove that there's not a spell involved. There certainly WAS a spell in vanilla since I can find it in the Vanilla DBCs, but I don't see it in BC or WotLK DBCs.

I understand that `spell_dbc` is (almost always) read-only, and I know you mentioned using a clean TrinityCore install as evidence of it being a legitimate server-side spell. Is that a design principal of AC or is there something about the TrinityCore DB that makes it a good source of information about spells that wouldn't necessarily be seen by the client?

I assume that adding the SPELL_AURA_SHARE_DAMAGE_PCT aura type to the `spell_dbc` entry is out of the question? The spell in there now is super barebones and doesn't carry any Aura type that I can tell. 

If we can't update `spell_dbc` with the appropriate AuraType, is blacklisting this spell by ID from  AutoBalance the only way I can filter it out?

Thanks for reviewing and for your explanations - helps me understand how to make better suggestions in the future. :+1:

Nefertumm:
For now, I'd blacklist the spell in the autobalance mod, as whenever I scripted it, I did it according to sniffs and CMangos (for serverside spells)

kjack9:
@Nyeriah I'd still be interested in the answers to my questions if you have time, but I do understand that this PR won't be accepted as-is. I'm going to close it.

Nyeriah:
DBC data *can* be wrong at times, but it's almost never the case with things like spell effects etc. It is usually a value that is off

> Is that a design principal of AC or is there something about the TrinityCore DB that makes it a good source of information about spells that wouldn't necessarily be seen by the client?

It makes it more reliable than ours because ours has private server (Sunwell) pollution unfortunately, that we are getting rid off as we go, as well as some other changes from the past that shouldn't have been made

PR #17621, opened 2023-10-30
(Core/Hooks) Add `OnAfterDatabaseLoadCreatureTemplates` hook

kjack9:
## Changes Proposed:

This PR proposes changes to:

Adds the `OnAfterDatabaseLoadCreatureTemplates` hook to `DatabaseScript`. This hook fires just after the creature template database (`ObjectMgr::_creatureTemplateStore`) has been loaded into `ObjectMgr::_creatureTemplateStoreFast`. This allows module authors to edit individual creature templates after they're loaded from the database but before they are applied to the creature. A variety of SQL-less in-memory DB edits can be accomplished with this, including changes that are not otherwise possible in code.

Additionally, I have fixed a hidden bug where doing a `.reload creature_template xxx` would not update `ObjectMgr::_creatureTemplateStoreFast`.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. checkout the code and compile
2. `.go xyz -11703.262695 -1965.512085 58.327675 309 4.811652`
3. target Zealot Zath (11348)
4. issue the `.npc info` command
5. observe that the immunity mask matches Zath's DB value: `604059227`

Download this (simple, WIP) module. Recompile and restart:
https://github.com/kjack9/mod-smallcraft/archive/ad50cdb1fbf1d13e73ded8e5f65d0801ad50c2fa.zip

1. `.go xyz -11703.262695 -1965.512085 58.327675 309 4.811652`
2. target Zealot Zath (11348)
3. issue the `.npc info` command
4. observe that the immunity mask matches the [value set in the module](https://github.com/kjack9/mod-smallcraft/blob/ad50cdb1fbf1d13e73ded8e5f65d0801ad50c2fa/src/Smallcraft.cpp#L40)
5. issue the command `.reload creature_template 11348`
6. issue the `.npc info` command
7. observe that the immunity mask still matches the [value set in the module](https://github.com/kjack9/mod-smallcraft/blob/ad50cdb1fbf1d13e73ded8e5f65d0801ad50c2fa/src/Smallcraft.cpp#L40)

N/A

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -602,10 +602,6 @@ void ObjectMgr::LoadCreatureTemplates()
     _creatureTemplateStore.rehash(result->GetRowCount());
     _creatureTemplateStoreFast.clear();
 
-    // resize the fast cache to the highest creature id
-    uint32 highestCreatureId = result->Fetch()[0].Get<uint32>();
-    _creatureTemplateStoreFast.resize(highestCreatureId + 1, nullptr);
-
     uint32 count = 0;
     do
     {
@@ -641,6 +637,12 @@ void ObjectMgr::LoadCreatureTemplate(Field* fields, bool triggerHook)
     uint32 entry = fields[0].Get<uint32>();
 
     CreatureTemplate& creatureTemplate = _creatureTemplateStore[entry];
+
+    // enlarge the fast cache as necessary
+    if (_creatureTemplateStoreFast.size() < entry + 1)
+    {
+        _creatureTemplateStoreFast.resize(entry + 1, nullptr);
+    }
 
     // load a pointer to this creatureTemplate into the fast cache
     _creatureTemplateStoreFast[entry] = &creatureTemplate;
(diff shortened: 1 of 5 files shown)


src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -752,6 +739,26 @@ void ObjectMgr::LoadCreatureTemplate(Field* fields)
     creatureTemplate.ScriptID              = GetScriptId(fields[70].Get<std::string>());
 }
 
+/**
+* @brief (Re)loads the fast Creature Template cache from the main cache. Does not reload from the DB.
+*/
+void ObjectMgr::LoadCreatureTemplatesFastCache()
```
kjack9, line 745: I strongly considered making a second overload of this function that takes a single EntryId and only updates that. But since you'd have to iterate the entire _creatureTemplateStore to find it, I figured it's simpler and easier to just refresh the entire thing. The only times this should get called anyways is on server start and when the `.reload creature_template xxx` command is used, so performance impact should be virtually non-existant.

Nyeriah:
How thread safe is this? Could someone tell? E.g creatures being spawned/added to world during the reload process

kjack9:
> How thread safe is this? Could someone tell? E.g creatures being spawned/added to world during the reload process

For the main functionality during initial load, it's being done in the same function call as the actual loading of the templates so I wouldn't expect a problem there. But threading is well outside my skillset at this point. 

As for the reload, would adding a function overload that only changes the affected entry alleviate any potential threading issues? If so I can get that put in this evening.

kjack9:
Had a quick chat with Nyeriah and I'm going to alter this to not reload the entire fast cache. Testing a solution now.

kjack9:
New approach - I've refactored the fast cache into the `LoadCreatureTemplate` function and given a way for outside classes to trigger the new hook. This should load the fast cache immediately as the DB is pulled from, and should remove any chance of reloaded creatures being out of date in the fast cache. There also shouldn't be any creatures affected during the reload other than the one being reloaded.

Let me know if this solution is a better fit.

pklloveyou:
Can I add hooks to the server without sharing modules?

Nyeriah (approved):

Issue #17577, opened 2023-10-25
worldserver crash when a player create a character with wrong race and class


Vitumoc:
### Current Behaviour

I use a latest version of AzerothCore, here is my server info:
AzerothCore rev.  2217563fb1dc 2023-10-20 12:50:59 +0000 (master branch) (Unix, RelWithDebInfo, Static)

when a user use a modified client, and try to create a Orc Mage(Normally this is not possible), the worldserver will crash, an this is the end of the logfile.

![image](https://github.com/azerothcore/azerothcore-wotlk/assets/31403302/adeffb1d-04ed-4e77-b3fc-0eac1b660255)

### Expected Blizzlike Behaviour

Character creation failed
user be kick out

### Steps to reproduce the problem

1. downlaod the patch
link：https://pan.baidu.com/s/1SR559ws1YtXxHnrGI3qUNw 
code：l5kl

2. copy the path to client
3. login an create a orc mage

### AC rev. hash/commit

AzerothCore rev. 2217563fb1dc 2023-10-20 12:50:59 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 24. Characters in world: 22.
Connection peak: 33.
Server uptime: 1 day(s) 30 minute(s) 16 second(s)
Update time diff: 9ms, average: 8ms.
Using SSL version: OpenSSL 3.0.2 15 Mar 2022 (library: OpenSSL 3.0.2 15 Mar 2022)
Using Boost version: 1.74.0
Using CMake version: 3.22.1
Using MySQL version: 100612
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 5.15.0-87-generic
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /home/vc/azeroth-data/maps. Total size: 291014951 bytes
vmaps directory located in /home/vc/azeroth-data/vmaps. Total size: 658130721 bytes
mmaps directory located in /home/vc/azeroth-data/mmaps. Total size: 2192910844 bytes
Default DBC locale: enUS.
All available DBC locales: enUS 
Using World DB: ACDB 335.10-dev
Latest LoginDatabase update: 2023_04_24_00.sql
Latest CharacterDatabase update: trasmorg.sql
Latest WorldDatabase update: trasm_world_texts.sql
[...]

PR #17625, opened 2023-11-01
fix(Core/Character):fix worldserver with autobalance mod crash when character creation fails

Vitumoc:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17577
- Closes https://github.com/azerothcore/mod-autobalance/issues/166

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:


kjack9:
I don't think I love this solution to the problem - doesn't this try to make a new map with ID 0 (Eastern Kingdoms, I believe)?

Instead, might I recommend touching the GetMap() function directly?

https://github.com/azerothcore/azerothcore-wotlk/blob/52d37a927dc921785fcac35bd8830221299495c7/src/server/game/Entities/Object/Object.h#L526C41-L526C47

Or it might be better to have the `OnPlayerLeaveCombat` hook not be called at all for a player who was never actually created?

kjack9:
I will see if I can find a good way to test, but I like your latest solution a lot better. 👍

Vitumoc:
When a player was not actually created, no cleanup is required. player is just a `Player*`.

https://github.com/Vitumoc/azerothcore-wotlk/blob/747aa8efbe93a8949f4d908e90024174290ee3f7/src/server/game/Handlers/CharacterHandler.cpp#L538-L559

`MotionMaster` will be automatically deleted.No need to use clean function.
when the player is created successfully, the AT_LOGIN_FIRST flag must be set. On the contrary, when the player creation fails, m_atLoginFlags will be AT_LOGIN_NONE.

I will do some testing on this change later.

kjack9 (approved):
This solution makes a lot of sense to me and prevents the firing of hooks/functions on characters that never actually entered the world.

Nyeriah:
Please fix code style (no php style braces, braces go to new line)

Nyeriah:
Also AtLoginFlags is a bitmask, so it needs bitwise evaluation (the check will fail if for some reason they have any other flag during their first login in addition to the first login flag)

Nyeriah (approved):

PR #17627, opened 2023-11-01
Correct default metric interval

AGandrup:
## Changes Proposed:

- Change the default metric interval to 1 second instead of 10 seconds.
- Added some further explanation to the description.
## Issues Addressed:

- Not being able to gather any data in Grafana due to batch of data being too large to be sent.

## Tests Performed:

I no longer get an error in worldserver.exe console, and I get data in Grafana as expected.

## How to Test the Changes:

1) Follow this guide on how to set up monitoring: https://www.azerothcore.org/wiki/monitoring-azerothcore-with-grafana
2) See that you get data as expected.

### DIFF
--- a/src/common/Metric/Metric.cpp
+++ b/src/common/Metric/Metric.cpp
@@ -71,7 +71,7 @@ void Metric::LoadFromConfigs()
 {
     bool previousValue = _enabled;
     _enabled = sConfigMgr->GetOption<bool>("Metric.Enable", false);
-    _updateInterval = sConfigMgr->GetOption<int32>("Metric.Interval", 10);
+    _updateInterval = sConfigMgr->GetOption<int32>("Metric.Interval", 1);
 
     if (_updateInterval < 1)
     {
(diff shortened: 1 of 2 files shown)


heyitsbench:
Would you be able to test the minimum point where data will start to get rejected and include that in the conf warning? Maybe irrelevant point, but it'd be nice to be sure.

AGandrup:
> Would you be able to test the minimum point where data will start to get rejected and include that in the conf warning? Maybe irrelevant point, but it'd be nice to be sure.

Sure, I can try later. But I don't know if I can get the exact batch size of when it starts to get rejected if thats what you want. Probably only a rough estimate. Normally one should also be able to increase or remove this limit in the influxdb.conf by adjusting the `max-body-size` setting, but for some reason it does not seem to work (at least not for me). I did a little bit of research and from what I can tell its a bug with some older versions of influx and I guess that includes v1.8.10 which I am using.

Kitzunu (approved):

PR #17637, opened 2023-11-03
Feat (Core) Add `OnAfterSpecSlotChanged` hook

kjack9:
## Changes Proposed:

This PR proposes changes to:

Add a `OnAfterSpecSlotChanged` hook that fires just after the player changes their spec slot using the dual spec feature. This is to be used for an upcoming public module I am working on.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -155,7 +155,7 @@ void ScriptMgr::OnPlayerTalentsReset(Player* player, bool noCost)
 
 void ScriptMgr::OnAfterSpecSlotChanged(Player* player, uint8 newSlot)
 {
-    ExecuteScript<PlayerScript>([&](PlayerScript* script)
+    ExecuteScript<PlayerScript>([=](PlayerScript* script)
     {
         script->OnAfterSpecSlotChanged(player, newSlot);
     });
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -153,6 +153,15 @@ void ScriptMgr::OnPlayerTalentsReset(Player* player, bool noCost)
     });
 }
 
+
+void ScriptMgr::OnAfterSpecSlotChanged(Player* player, uint8 newSlot)
+{
+    ExecuteScript<PlayerScript>([&](PlayerScript* script)
+    {
+        script->OnAfterSpecSlotChanged(player, newSlot);
+    });
+}
+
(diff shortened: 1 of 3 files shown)


AnchyDev:
Calling the hook `OnAfterActivateSpec` might make more sense here since it describes exactly what it's doing. That's just my opinion however, it's up to you if you agree. 😄

kjack9:
I don't care which one it gets called, whatever gets the PR merged. If changing it will get me the okay to have it merged, I'll be happy to.

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -153,6 +153,14 @@ void ScriptMgr::OnPlayerTalentsReset(Player* player, bool noCost)
     });
 }
 
+void ScriptMgr::OnAfterSpecSlotChanged(Player* player, uint8 newSlot)
+{
+    ExecuteScript<PlayerScript>([&](PlayerScript* script)
```
Winfidonarleyan, line 158:
```suggestion
    ExecuteScript<PlayerScript>([=](PlayerScript* script)
```

kjack9: I'm fine to make this change, but I do not understand it. I am mirroring the hooks similar to this one, and they are all using `&`. Why is this hook different?

AnchyDev: The difference is you capture a copy or a reference, `&` is by reference, `=` is by copy.
So with the change, it means you cannot affect the original player object because you are accessing a copy of that player object. This effectively makes it read-only since none of the writes will affect the original object.

AnchyDev (approved):
Looks good to me.

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
Winfidonarleyan: > The difference is you capture a copy or a reference, `&` is by reference, `=` is by copy. So with the change, it means you cannot affect the original player object because you are accessing a copy of that player object. This effectively makes it read-only since none of the writes will affect the original object.
> 
> Whether or not this is intended is up to you to dispute with Winfidonarleyan 😄

Only copy `ptr` player, object player not copy

Winfidonarleyan (approved):

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
Winfidonarleyan: https://en.cppreference.com/w/cpp/language/lambda `Lambda capture`

Winfidonarleyan:
Thanks for PR

Issue #16676, opened 2023-07-05
Creature level scaling spells do not seem to scale damage properly.


heyitsbench:
### Current Behaviour

Spells with the [SPELL_ATTR0_SCALES_WITH_CREATURE_LEVEL](https://www.wowhead.com/wotlk/spells?filter=58;1;0) attribute do not seem to scale the damage value correctly based on caster level.

For this example, I will refer to the [Bristleback Shaman](https://wowgaming.altervista.org/aowow/?npc=2953) (level four in this particular example) and their [Lightning Bolt](https://wowgaming.altervista.org/aowow/?spell=9532) ability. Against a level one Tauren warrior (with `.cheat god` on), [this is what the combat log looks like](https://github.com/azerothcore/azerothcore-wotlk/assets/38561765/e36dcbe7-9df9-47f3-9050-b9fd3a642c63).

### Expected Blizzlike Behaviour

Unsure at this point.

As a part of the prior example, on Classic in the same conditions, [this is what the combat log looks like](https://github.com/azerothcore/azerothcore-wotlk/assets/38561765/04340de5-60d5-4e7b-94d8-2fc8fb42790e).

### Source

Wrath Classic

### Steps to reproduce the problem

1. Create a Tauren warrior.
2. `.go c id 2953`
3. Let it cast [Lightning Bolt](https://wowgaming.altervista.org/aowow/?spell=9532) at you.
4. Observe its damage.

### Extra Notes

There is currently a [VMaNGOS PR](https://github.com/vmangos/core/pull/2007) open targeting this same issue. Could possibly cherry-pick if we can verify on Wrath Classic that the formula is the same.

Unsurprisingly, this issue may have been reported on the CC issue tracker for various individual casters/spells. Those will have to be triaged later since there's probably a whole bundle bunch of em. As well, individual casters/spells could have been reported on the AC repo. Those will have to be identified as well.

### AC rev. hash/commit

893a9beeacc4554f9ac85755b40520a9b8e6d411

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

Gultask:
https://github.com/chromiecraft/chromiecraft/issues/5818
https://github.com/chromiecraft/chromiecraft/issues/5820

PR #17649, opened 2023-11-07
fix(Core/Spells): Correct spell scaling formula.

heyitsbench:
If you want something done, you gotta do it yourself.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16676
- Likely closes various other AC/CC issues.

## SOURCE:

- [X] Live research (checked on live servers, i.e. Classic WotLK)
- [X] The changes promoted by this pull request come partially or entirely from another project (cherry-pick), in this case [VMaNGOS](https://github.com/vmangos/core/commit/9cfcf9eb6fe55fdc082738b8012a4ed275328235)/[balakethelock](https://github.com/balakethelock).

## Tests Performed:

This PR:
- [X] Has been tested in-game by the author.
- [X] Requires further testing and may have edge cases to be tested.

## How to Test the Changes:

1. Find spells that are cast that [have the spell scaling attribute](https://wowgaming.altervista.org/aowow/?spells&filter=cr=58;crs=1;crv=0).
2. Find scenarios in the world in which the spell is cast upon you as a player.
3. Test that the spell damage seems appropriate.

It's possible, likely even, that there are discrepancies between this solution and values observed on live servers. However, from what testing I have done (obviously not every creature in the world), they are not at all far off and are much closer with this solution than current values.


Kitzunu:
See! I told you "you do it". He listened 🥰😝

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -502,10 +502,13 @@ int32 SpellEffectInfo::CalcValue(Unit const* caster, int32 const* bp, Unit const
 
             if (canEffectScale)
             {
-                GtNPCManaCostScalerEntry const* spellScaler = sGtNPCManaCostScalerStore.LookupEntry(_spellInfo->SpellLevel - 1);
-                GtNPCManaCostScalerEntry const* casterScaler = sGtNPCManaCostScalerStore.LookupEntry(caster->GetLevel() - 1);
-                if (spellScaler && casterScaler)
-                    value *= casterScaler->ratio / spellScaler->ratio;
+                CreatureTemplate const* cInfo = caster->ToCreature()->GetCreatureTemplate();
+
```
Kitzunu, line 506: Is there 100% no risk of this ever being null?

src/server/game/Spells/SpellInfo.cpp
```diff
@@ -502,10 +502,13 @@ int32 SpellEffectInfo::CalcValue(Unit const* caster, int32 const* bp, Unit const
...
+                CreatureBaseStats const* pCBS = sObjectMgr->GetCreatureBaseStats(caster->GetLevel(), caster->getClass());
+                float CBSPowerCreature = pCBS->BaseDamage[cInfo->expansion];
+                CreatureBaseStats const* spellCBS = sObjectMgr->GetCreatureBaseStats(_spellInfo->SpellLevel, caster->getClass());
+                float CBSPowerSpell = spellCBS->BaseDamage[cInfo->expansion];
+                value *= CBSPowerCreature / CBSPowerSpell;
```
Kitzunu, line 511: same for these

src/server/game/Spells/SpellInfo.cpp
heyitsbench: > Is there 100% no risk of this ever being null?

Caster is evaluated on L450, and, correct me if I'm wrong, I don't see either method being at risk of causing a null given caster exists.

src/server/game/Spells/SpellInfo.cpp
heyitsbench: Can't guarantee nothing in this block can be at risk of a null, but for the most part they're either A) checked for previously (caster example), or used similarly previously (_spellInfo's SpellLevel for instance).

src/server/game/Spells/SpellInfo.cpp
elthehablo: Yeah this doesn't look like it could be null to me. Maybe if caster somehow can't be cast to Creature

Kitzunu: I did not know caster was already checked. Then no, the only way it would null is of there is no ct, which is highly unlikely anyway

Kitzunu (approved):

elthehablo:
Yeah looks good to me also

elthehablo (approved):

Nyeriah:
Doesn’t look like the crash we talked about on Disxord was addressed

src/server/game/Spells/SpellInfo.cpp
Nyeriah: It’s not granted to be a creature, and in that case, it will crash

elthehablo:
Ah shit. Let's revert this then?

heyitsbench:
As I said on Discord, I tested your proposed scenario of a player learning a spell with the attribute through GM commands and casting it, and experienced no crash, therefore I didn't feel it needed addressing.

pklloveyou:
It feels like the code is prone to crashing.

heyitsbench:
> It feels like the code is prone to crashing.

Is this a hypothetical or do you have a crashlog?

pklloveyou:
> > It feels like the code is prone to crashing.
> 
> Is this a hypothetical or do you have a crashlog?

hypothetical

heyitsbench:
If it's actually crashing, I'll submit the PR myself to correct it and add the check that caster is creature, but again, in my testing, I could not get it to crash no matter what I did.

pklloveyou:
> If it's actually crashing, I'll submit the PR myself to correct it and add the check that caster is creature, but again, in my testing, I could not get it to crash no matter what I did.

Okay.

elthehablo:
Doesn't hurt to try and catch a potential null in the script right? Why not make it now?

heyitsbench:
> Doesn't hurt to try and catch a potential null in the script right?

Null is already checked for.

elthehablo:
> > Doesn't hurt to try and catch a potential null in the script right?
> 
> Null is already checked for.

I mean on the tocaster cast

heyitsbench:
> I mean on the tocreature cast

I have been told the only situation caster (in this case) can apply to as a unit is creatures or players. I can't think of a situation where caster being a creature would cause the ToCreature cast to fail, so that's off the table leaving players. The if statement on L461 also removes players from the pool through IsControlledByPlayer, as well as player controlled creatures for good measure, so there shouldn't be a situation where a player unit makes it to that line. From my understanding, ToCreature should not fail here.

As I've said, I've tested this and it doesn't crash. I did say the PR should probably see testing by others, but obviously not something I can control. If it is crashing somehow, throw a crashlog and I'll correct it, but I see no reason to throw if statements everywhere if they're not needed.

Nyeriah:
>IsControlledByPlayer()

Just checks if they're a charmed mob, if it's not a charmed unit, then it code will execute, it doesn~t account wether the unit we are handling is a player or not

You said you tested it and its not crashing so cool we can move on

PR #17712, opened 2023-11-15
fix(Core/Creature): `SelectLevel` hooks return `IsDead() = True` during respawn

kjack9:
## Changes Proposed:

This PR proposes changes to:

When a `Respawn()` occurs, `SelectLevel()` is currently called before the creature's `DeathState` is updated to `DeathState::JustRespawned`. This has the strange effect of calling the `OnBeforeCreatureSelectLevel()` and `Creature_SelectLevel()` hooks when the `creature->IsDead()` function returns `True`. It also makes it very difficult for a module to determine what is a spawned-dead creature vs a respawning creature.

This PR creates a new `DeathState::Respawning` that is set when the respawn process has started but the actions associated with `DeathState::JustRespawned` aren't due to happen yet. 

This solution should be fairly low-risk since it only affects the `DeathState` during `SelectLevel()` and some smaller functions, and I don't see anything in there that relies on `DeathState`. The hooks would be affected, but they should see improved accuracy with `IsDead()`.

Another option I considered was moving `DeathState::JustRespawned()` to before `SelectLevel()`. While it seemed to work okay, I am concerned about some of the function calls in `setDeathState` when the state is `DeathState::JustRespawned` and how they'll react if `SelectLevel()` hasn't been called again.

Alternatively, we could just remove the `SelectLevel()` call entirely for respawning, since that function was presumably called when the creature was created the first time.

Interested in hearing what everyone thinks.

## Issues Addressed:

- Closes https://github.com/azerothcore/mod-autobalance/issues/172

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:


Kitzunu (changes requested):
Everything about this feels weird.

1. DeathState::Respawning does nothing here. And if it were to be used it should be added into the switch at Creature.cpp L#610
2. If the module hook is the issue, then I would highly prefer you change where and how the hook is called instead of changing core functionality to fix a module.

kjack9:
Good discussion, thank you for replying.

The issue in the core is that the `SelectLevel()` hooks see different states for `IsDead()` depending on whether a creature is being respawned or not. When the creature is first created, the hooks see `IsDead() = false` but if the creature is respawned they see `IsDead() = true`, even if the creature never died (and was force respawned). As a module author, I cannot tell if this creature is actually dead or not until the first `OnAllCreatureUpdate()`. That seems like a core issue to me since the hooks are reporting inconsistent information to modules.

I think the hooks are in the right place - before and after stat generation. I'm not sure where they could be moved that they wouldn't be affected by this anymore.

What did you think of my other proposed solutions? I only picked this one because it had the lowest chance of breaking something else. Do we really need to run `SelectLevel()` and re-apply stats from the creature template during a respawn event? The creature already has its stats determined and applied, does it dying alter it in a way that those stats need to be recalculated? Any changes made by a module would be overwritten by this call, meaning that the same changes need to be made again. If a module/script really wanted that behavior they could call `SelectLevel()` themselves.

kjack9:
I am starting to test with that `SelectLevel()` removed entirely - I am not sure it needs to be there. I'd like to collect some cases where creatures might have their stats/characteristics changed during combat, and on respawn need to be reset to their template values.

I'll keep a running list of things I've tested here:
- Raging Colossus (19188) - OK

kjack9:
What triggered this in the initial issue was a `Respawn(force=True)` being called here:

https://github.com/azerothcore/azerothcore-wotlk/blob/bdc970b68be55102fadd71387d3cc3385f887daf/src/server/scripts/Outland/TempestKeep/Eye/boss_kaelthas.cpp#L200

It's the forced respawn of an alive creature that causes this. Would you prefer I find a way to keep the hooks in `SelectLevel` from firing if the creature was respawned while still alive? I understand that `Respawn(true)` is manipulating the death state to get its magic to happen.

I'm open to suggestion here, but from the module side I don't know how to handle this differently if the hook is giving me incorrect information. Let me know what you all think.

Kitzunu:
But for isDead() to be true the creature would have DeathState::Dead or DeathState::Corpse which I don't see Respawn() ever setting unless the creature was not alive at first

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2041,11 +2041,12 @@ void Creature::Respawn(bool force)
...
 
         LOG_DEBUG("entities.unit", "Respawning creature {} (SpawnId: {}, {})", GetName(), GetSpawnId(), GetGUID().ToString());
+        setDeathState(DeathState::Respawning);
         m_respawnTime = 0;
         ResetPickPocketLootTime();
         loot.clear();
-        SelectLevel();
 
+        SelectLevel();
         setDeathState(DeathState::JustRespawned);
```
Kitzunu, line 2050: and this would just be a hackfix to be fair

kjack9:
> But for isDead() to be true the creature would have DeathState::Dead or DeathState::Corpse which I don't see Respawn() ever setting unless the creature was not alive at first

This happens during a forced respawn, even if the creature is alive.
https://github.com/azerothcore/azerothcore-wotlk/blob/e90d7a2f9246107e066eda45175c933c5ad0d1e5/src/server/game/Entities/Creature/Creature.cpp#L2011-L2014

Regardless, SelectLevel (IMO) shouldn't see different `IsDead()` values depending on whether the creature is being spawned for the first time or is being respawned. That's the root of the issue here.

> and this would just be a hackfix to be fair

Which alternative would you like me to pursuit?
- don't call `SelectLevel()` again when respawning creatures since stats were determined upon initial creation
- move [`setDeathState(DeathState::JustRespawned);`](https://github.com/azerothcore/azerothcore-wotlk/blob/e90d7a2f9246107e066eda45175c933c5ad0d1e5/src/server/game/Entities/Creature/Creature.cpp#L2049) to just _before_ the `SelectLevel()` call
- create an `isRespawning` bool that is passed down to the hooks so they can tell if `IsDead()=true` actually means `Respawning`. This is just a different implementation of the current solution.

kjack9:
Wanted to follow up here. I still think that having a death state between "Dead/Corpse" and "JustRespawned" solves this problem nicely with zero side effects. I am also willing to pursue any of the other options I've laid out. Let me know which direction to run.

kjack9:
My next step forward is to do a core patch for AutoBalance to resolve this issue. The patch would be the changes in this PR. I don't want to do that, especially for a module that is as widely-used as AutoBalance, but I don't see any other path forward at this point. I'd much rather fix it in a way that is more acceptable to the core team, but I don't know what that looks like. Let me know if a core patch is the way I should go here.

Kitzunu:
Honestly, I dont know, maybe @Nyeriah got a suggestion?

kjack9:
> Is this an issue anywhere else other than Kael's adds? Because his script is bad and needs to be rewritten to begin with. Just so we don't change something in the core in order to workaround a bad script that needs to be worked either way

Yes, anywhere a force respawn is done on a creature that is alive. I'll see if I can find another replicatable scenario that shows the issue.

sudlud:
Closing this for now as it has been abandoned for quite some time.

Please feel free to re-open and update the PR again.

PR #17714, opened 2023-11-15
fix a Mac build issue introduced by commit #17643

natrist:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

The Mac build fails with the following 2 error messages:
`src/common/Utilities/StringFormat.h:75:53: error: no type named 'locale' in namespace 'std'`
`src/common/Utilities/StringFormat.h:75:72: error: no member named 'locale' in namespace 'std'`

## Tests Performed:
- Built on macOS Sonoma 14.1.1 (23B81)

### DIFF
--- a/src/common/Utilities/StringFormat.cpp
+++ b/src/common/Utilities/StringFormat.cpp
@@ -17,7 +17,6 @@
 
 #include "StringFormat.h"
 #include "Define.h"
-#include <locale>
 
 template<class Str>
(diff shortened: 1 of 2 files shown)


dedmen:
I ran into same issue when building on Linux with clang 15.
And applied the same fix

Nyeriah (approved):

PR #17716, opened 2023-11-15
fix(Core/Creature): Reduce FAMILY_ASSISTANCE_RADIUS in start area

TWC002:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- https://github.com/chromiecraft/chromiecraft/issues/5316

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. .go zonexy 41.6 25.6 14
2. Try to pull one orc
3.

- [ ]
- [ ]


Kitzunu (changes requested):
I will block this PR on some counts.

1. Do not remove and ignore the PR template.
2. Aggro radius and/or help radius is sniffed. (IIRC)
3. The use of hardcoded values.
4. That this is a hackfix.

Kitzunu:
This should be investigated further before we settle with this fix.

Perhaps @heyitsbench has some insight into it?

Saniok88:
Actually a reported that quest only as an example, but the issue is global, not just with this quest alone.
The only reason i mentioned that quest, because its easy to reproduce, but there exist other cases, again, trying to solo Skull Rock (cave in north-east of Durotar) at level 10-12 will also prove that. 

I dont say, that only that quest needs changes, i say, that global variable (FAMILY_ASSISTANCE_RADIUS) need to be tweaked, it should NOT be 50-100% higher, than blizzlike value in any zone, not just in Durotar.

Saniok88:
This commit should be declined (such hacks is a bad practice, globally radius of 7 should be fine anyway), that radius should be changed in config file from 10 to 7 (and also in the CPP file, where that config file is read). Again, currently it is too high, should be 7, not 10.

The config file is:
azerothcore-wotlk-master\src\server\apps\worldserver\worldserver.conf.dist
Line 2796:
`CreatureFamilyAssistanceRadius = 10`
value "10" should be changed to "7", to become both: Blizzlike and not feel as inaccurate, as it currently is (for players, who actually play the game, especially solo, and in areas with higher mob density).

Again, CPP file too:
azerothcore-wotlk-master\src\server\game\World\World.cpp
Line 1034:
`    _float_configs[CONFIG_CREATURE_FAMILY_ASSISTANCE_RADIUS]      = sConfigMgr->GetOption<float>("CreatureFamilyAssistanceRadius", 10.0f);`
There "10.0f" should be replaced with "7.0f"

Saniok88:
> I will block this PR on some counts.
> 
>     1. Do not remove and ignore the PR template.
> 
>     2. Aggro radius and/or help radius is sniffed. (IIRC)
> 
>     3. The use of hardcoded values.
> 
>     4. That this is a hackfix.

regarding hackfix i replied above, and regarding sniffing, believe me, im not a developer, but i got huge experience playing solo, current inaccurate radius is way too obvious for experienced players. So you either believe to those, who have experience and good understanding, or infinitely wait for sniffing by developers/programmers, who have quite low experience and dont know how it should be. I would prefer to have this issue finally fixed for good (the way i suggested above). If you still want to actually test it personally, rather than believing to those, who did it already 100 times, then try:
1) to pull carefully orcs near Fizzle (in Durotar). Ofcourse you should not hurry and make sure, that you dont pull, when patrol comes (if there are some, or if some mobs wander). You will see, that when you expect to pull a single orc, nearby orc from too far will also come, so 2+ orcs pulled, when not intended.

NOTE:
(i know, there exist Voidwalker pets with some of those hostile orcs, and they hit hard with their short cooldown instant-cast spells (i think it is not Blizzlike, but im not sure and currently it is not our topic), so soloing a single orc with his Voidwalker pets can be too hard at level 12, you can use any level, even level 70 for that purpose, because the Family_Assistance does not depend on player level).

2) Also try to pull single orcs in Skull Rock (Cave in Durotar, east from Orgrimmar's entrance), inside the cave go rightmost. You will get double-orc pull between the place, where player stands in the beginning of the video (literally at first second of the video) and the place, where his target orc stays (so between these 2 points, you will get double-orc pull because of way too high Family_Assistance).
https://www.youtube.com/watch?v=4pmZojbH8SY

Again, These are just two examples, but everywhere else in the game world, even in Hellfire Peninsula in Zeth Gor same problem can be seen very obviously too. So the problem is in every zone, there just needs to be config fix.

If you think, that most of the players on ChromieCraft (or any other wow of.realm or p.server) are: adult, accountable, not lazy, attentive, mindful, know about Git Hub, have account created, understand how to describe the bug, khow how to write bug reports, actually care at all to report anything, on so on ... then you are wrong, this is not how real world works, there exist some amount of obvious bugs, that literally yell at players "hey, you imagine, how developers are blind to to fix so obvious bug for so long" (something, that could be noticed and fixed from week one, on AC exists for years). There are 5 such bugs, that i personally noticed and will mention it here now:
1. Family-Assistance radius (discussed here)
2. Corpses (normal, so non-elite) despawn in 60 !!! seconds, should be 300 (i know, there was attempt to fix it in 2023, but eventually it was rolled back, because of bugged "Rate.Corpse.Decay.Looted", but who prohibits you from keeping "CONFIG_CORPSE_DECAY_NORMAL" being 300 seconds ?? only the fact, that 99.9% of players and testers either dont care or notice, and the remaining 0.01% of players/testers just gave-up hopes to see it ever fixed or accepted. To fix it, "Rate.Corpse.Decay.Looted" can be ignored (tho, its code is probably bugged in CPP implementation, it's value does not matter much for gameplay, anyway). What really matters is only the "Corpse.Decay.NORMAL" - should be 300, not 60 (it is also set in cpp file too)
3. Navigation Meshes need fixing (i have no idea how this term is called in WoW, if not same), but on Azeroth creatures sometime run through wall, for example in Hillsbrad Foothills inside that big building with Magistrate Burnside low hp humanoids when running away - they run through walls (from inside the building). Same may apply for buildings in Silverpine Forest, Western Plaguelands, and probably everywhere else. This issue together with 1st heavily make life way harder for hardcore solo players, if dont blame, them, that they dont understand technical reasons behind it, and how it works, but i can blame you, that you dont care and dont fix it.
4. Casters "forgetting" that they can and should walk (run to their targets), and instead they stay in place, even between casts (for example when spell is on cd or was interrupted by melee "kick"). I believe, "it may be intentionally coded like that, but for experienced players it feels like it is a bug (and it never will evaluated as "non-bug" it is definitely a bug, and makes pulling harder, which affects gameplay negatively, so this bug is moderately annoying.
5. Level up hp/mana increased amount (yellow colored text). It is just a cosmetic thing, and does not ruin gameplay for players (unlike other mentioned bugs)

Your testers ignore above 4-5 issues for years, and you still dont believe me, that Family-Assistance-Radius needs even more "sniffing" ? Geez !

Current Family-Assistance-Radius probably will not ever be reported by anyone, but only because it's existence only proves, that either developers and testers are blind, or they care at all, so why anyone would report something, that is literally obvious? (it feels, like a waste of time for a potentional tester/reporter). Because if developers would ever care (or were not blind), they would fix it long time ago, because these issues are obvious and hugely impact gameplay for solo players.

X-eme:
> 1. .go zonexy 41.6 25.6 14
> 2. Try to pull one orc

I pulled one orc 💚

TheSCREWEDSoftware:
What's the current state of this PR? @TWC002

PR #17772, opened 2023-11-18
feat (core/cs): Make `.go xyz` ignore non-numeric data to improve copy-paste usability

kjack9:
## Changes Proposed:

This PR proposes changes to:

Rewrites the command handler function for `.go xyz` to handle flexible inputs. The new implementation will use regex to extract float and integer values from the provided arguments and will discard any other characters.

Since the parsing doesn't care about any of the surrounding text, many different versions of coordinates are supported by this implementation. The only requirement is that the numbers are in the right order (`x y [z [mapId [orientation]]]`) and have at least one delimiter character (not a number) between each number.

The following all work as expected:
`.go xyz 4073.1 -3536.1 123.5 329 2.8` - previously-required format
`.go xyz 4073.2-3536.2,123.5` - any non-number works as a delimiter
`.go xyz Position: X: 4073.3 Y: -3536.3 Z: 123.5` - sniff parse format
`.go xyz Position: X: 4073.4 Y: -3536.4 Z: 123.5 329 O: 2.8` - sniff parse format with `mapId` inserted
`.go xyz (4073.5f,-3536.5f,123.5f` - C++ format, doesn't have to be complete
`.go xyz Position(4073.6f, -3536.6f, 123.5f, 329 2.8f)` - C++ position with `mapId` inserted
`.go xyz 4073.7-3536.7f123.5?329` - any non-number works as a delimiter
`.go xyz This IsAStupidTextBut4073.8ItWorks Fine-3536.8BecauseOf123.5RegExParsing329`

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

`.tel strath` -- puts you in the correct map for the commands that don't specify one

Try each of these cases one at a time. Each one should move you diagonally slightly so you can tell it worked. Some will have orientation and others won't.

`.go xyz 4073.1 -3536.1 123.5 329 0.1`
`.go xyz 4073.2-3536.2,123.5`
`.go xyz Position: X: 4073.3 Y: -3536.3 Z: 123.5`
`.go xyz Position: X: 4073.4 Y: -3536.4 Z: 123.5 329 O: 0.1`
`.go xyz (4073.5f,-3536.5f,123.5f`
`.go xyz Position(4073.6f, -3536.6f, 123.5f, 329 0.1f)`
`.go xyz 4073.7-3536.7f123.5?329`
`.go xyz This IsAStupidTextBut4073.8ItWorks Fine-3536.8BecauseOf123.5RegExParsing329`

Feel free to bring in your own test cases. As long as the numbers are in the correct order (`x, y, [z [mapId [orientation]]]`) with at least one non-number character between each number it should work fine.

The ordering of parameters is weird when copy-pasting. I'd like to find a way to make the mapId be "whatever the last argument is" while still giving the option to omit it entirely. I considered making is so that `x y z o` all need to be in float format (`123.0` rather than `123`), so then I can just find the last number that isn't in that format, but I want feedback before I implement that.

### DIFF
--- a/src/server/scripts/Commands/cs_go.cpp
+++ b/src/server/scripts/Commands/cs_go.cpp
@@ -307,46 +307,45 @@ public:
             return false;
         }
 
-        // split args by space
-        std::vector<std::wstring> delimitedValues;
-        boost::split(delimitedValues, wInputCoords, boost::is_any_of(L" "));
-
-        if (delimitedValues.size() < 2)
+        // extract float and integer values from the input
+        std::vector<float> locationValues;
+        std::wregex floatRegex(L"(-?\\d+(?:\\.\\d+)?)");
+        std::wsregex_iterator floatRegexIterator(wInputCoords.begin(), wInputCoords.end(), floatRegex);
+        std::wsregex_iterator end;
+        while (floatRegexIterator != end)
         {
-            return false;
-        }
+            std::wsmatch match = *floatRegexIterator;
+            std::wstring matchStr = match.str();
 
-        // create a new vector of floats that only contains args that look like floats (or integer for mapId)
-        std::vector<float> floatValues;
-        for (const std::wstring& value : delimitedValues)
-        {
-            std::wsmatch match;
-            if (std::regex_match(value, match, std::wregex(L"(^\\-?[0-9]*[\\.]?[0-9]*$)")))
+            // try to convert the match to a float
+            try
             {
-                if (std::stof(value)) // ensures that the leading `.` doesn't get put into the vector
-                {
-                    floatValues.push_back(std::stof(value));
-                }
+                locationValues.push_back(std::stof(matchStr));
             }
+            // if the match is not a float, do not add it to the vector
+            catch (std::invalid_argument const&){}
+
+            ++floatRegexIterator;
         }
 
-        if (floatValues.size() < 2)
+        // X and Y are required
+        if (locationValues.size() < 2)
         {
             return false;
         }
 
         Player* player = handler->GetSession()->GetPlayer();
 
-        uint32 mapId = floatValues.size() >= 4 ? uint32(floatValues[3]) : player->GetMapId();
+        uint32 mapId = locationValues.size() >= 4 ? uint32(locationValues[3]) : player->GetMapId();
         Map const* map = sMapMgr->CreateBaseMap(mapId);
 
-        float x = floatValues[0];
-        float y = floatValues[1];
-        float z = floatValues.size() >= 3 ? floatValues[2] : std::max(map->GetHeight(x, y, MAX_HEIGHT), map->GetWaterLevel(x, y));
-        // map ID (floatValues[3]) already handled above
-        float o = floatValues.size() >= 5 ? floatValues[4] : 0.0f;
+        float x = locationValues[0];
+        float y = locationValues[1];
+        float z = locationValues.size() >= 3 ? locationValues[2] : std::max(map->GetHeight(x, y, MAX_HEIGHT), map->GetWaterLevel(x, y));
+        // map ID (locationValues[3]) already handled above
+        float o = locationValues.size() >= 5 ? locationValues[4] : 0.0f;
 
-        if (!MapMgr::IsValidMapCoord(mapId, x, y))
+        if (!MapMgr::IsValidMapCoord(mapId, x, y, z, o))
         {
             handler->PSendSysMessage(LANG_INVALID_TARGET_COORD, x, y, mapId);
             handler->SetSentErrorMessage(true);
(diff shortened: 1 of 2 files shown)


sudlud:
Looking forward to this 🙃

src/server/scripts/Commands/cs_go.cpp
```diff
@@ -288,34 +291,68 @@ class go_commandscript : public CommandScript
...
-    static bool HandleGoXYZCommand(ChatHandler* handler, float x, float y, Optional<float> z, Optional<uint32> id, Optional<float> o)
+    /**
+     * @brief Teleports the GM to the specified world coordinates, optionally specifying map ID and orientation.
+     *
+     * @param handler The ChatHandler that is handling the command.
+     * @param inputCoords The coordinates to teleport to in format "x y z [mapId [orientation]]".
+     * @return true The command was successful.
+     * @return false The command was unsuccessful (show error or syntax)
+     */
+    static bool HandleGoXYZCommand(ChatHandler* handler, Tail inputCoords)
```
Kitzunu, line 302:
```suggestion
    static bool HandleGoXYZCommand(ChatHandler* handler, Tail args)
```

src/server/scripts/Commands/cs_go.cpp
```diff
@@ -288,34 +291,68 @@ class go_commandscript : public CommandScript
...
-            if (!MapMgr::IsValidMapCoord(mapId, x, y))
-            {
-                handler->PSendSysMessage(LANG_INVALID_TARGET_COORD, x, y, mapId);
-                handler->SetSentErrorMessage(true);
-                return false;
-            }
-            Map const* map = sMapMgr->CreateBaseMap(mapId);
-            z = std::max(map->GetHeight(x, y, MAX_HEIGHT), map->GetWaterLevel(x, y));
+            return false;
+        }
```
Kitzunu, line 335: There should be an error message if not x and y is provided.

Kitzunu (changes requested):

src/server/scripts/Commands/cs_go.cpp
kjack9: The current solution causes the syntax help to be shown. What error would you like me to show instead?

src/server/scripts/Commands/cs_go.cpp
kjack9: Fair enough, I'll change this.

kjack9:
Wiki update: https://github.com/azerothcore/wiki/pull/935

src/server/scripts/Commands/cs_go.cpp
Kitzunu: handler->PSendSysMessage(LANG_INVALID_TARGET_COORD, x, y, mapId);
            handler->SetSentErrorMessage(true);
            return false;

kjack9: I'm so sorry, I don't understand.

At the point in the function that you're suggesting this, I don't have `x`, `y`, or `mapId` yet. In fact, we know that the user hasn't provided at least 2 values, so `x` or `y` are going to be null. Emitting `LANG_INVALID_TARGET_COORD` here also means that it's impossible to get access to the syntax help.

I _do_ check that the coordinates provided are valid further down, and if not I send the `LANG_INVALID_TARGET_COORD` message with the `x`, `y`, and `mapId` values I was able to parse out.
https://github.com/kjack9/azerothcore-wotlk/blob/ca1435384192fc05331b32da82dea69be486d755/src/server/scripts/Commands/cs_go.cpp#L348-L353

Sorry to be missing this one. Let me know what you're wanting in this situation.

heyitsbench: Most existing commands in my experience do stick the syntax in your face as an error message when improper syntax is supplied, so personally at least I agree with kjack's current implementation.

kjack9:
It's not a blocker to this PR merging, but I'd be interested in comments on this potential todo item:

> The ordering of parameters is weird when copy-pasting. I'd like to find a way to make the mapId be "whatever the last argument is" while still giving the option to omit it entirely. I considered making is so that x y z o all need to be in float format (123.0 rather than 123), so then I can just find the last number that isn't in that format, but I want feedback before I implement that.

src/server/scripts/Commands/cs_go.cpp
Kitzunu: this would give the syntax too. And keep the same behavior as before.

Kitzunu: never mind my brain is not functioning correctly

Kitzunu (approved):

Kitzunu:
> x y z o all need to be in float format

No, this will just make the command bad to use.

kjack9:
Thanks to all the reviewers! Anything else on this PR before it's good to go?

Issue #16670, opened 2023-07-04
Ulduar Steel Council automatically resets some time after death


zxnhrdz:
### Current Behaviour

steelbreaker
runemaster_molgeim
stormcaller_brundir

After all bosses are killed, the Steel Council will automatically reset within a certain period of time

### Expected Blizzlike Behaviour

It takes a week before it resets

### Steps to reproduce the problem

Pass all the bosses, then return to the Steel Council. You will find that it has been reset!

### AC rev. hash/commit

[21cab04]

### Operating system

Ubuntu 20.04

### Custom changes or Modules

Eluna

zxnhrdz:
Enter the raid and keep beating XT-002 Dekonstruktor, then enter the Odur antechamber. There are three bosses in the antechamber, if you only kill boss_assembly_of_iron alone, it will not reset. But if you kill Kologarn or Auriaya first, and then kill assembly_of_iron, boss_assembly_of_iron will automatically reset in an hour or so.

Please focus on testing it! Thanks!

kjack9:
I may take a look at this soon, but I see where the problem is coming from.

When the creatures evade, they are using the `DespawnOnEvade()` function triggered by their CREATURE_FLAG_EXTRA_HARD_RESET creature flag.

https://github.com/kjack9/azerothcore-wotlk/blob/7d3f237f71b2149c9d9758524150a6c27aea2de9/src/server/game/AI/CreatureAI.cpp#L225-L231

Since a respawn delay isn't specified, it defaults to 20 seconds.

https://github.com/kjack9/azerothcore-wotlk/blob/7d3f237f71b2149c9d9758524150a6c27aea2de9/src/server/game/Entities/Creature/Creature.h#L277-L278

`DespawnOnEvade()` calls `DespawnOrUnsummon()`, which immediately despawns the creature but gives it a 20s respawn timer, presumably to let it re-appear. It's notable that this isn't working right - the bosses immediately respawn when I reset the encounter.

Finally, since there is no mechanism to restore the craeture's respawn timers from 20s, they are stuck with the new respawn timer.

I'm not 100% sure what the "right" fix is for this - you almost need to restore the original respawn timer that's been saved somehow. I can do that pretty simply in some CustomData, but that certainly doesn't feel like a proper solution.

I'll sleep on it and see if I can figure out anything tomorrow.

kjack9:
Also possible that the manual respawn that's happening in the boss script is causing this too.

[...]

PR #17805, opened 2023-11-20
fix(DB/Creature): Lady Vashj should despawn on evade

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1700477079190660500.sql
+++ b/data/sql/updates/pending_db_world/rev_1700477079190660500.sql
@@ -1,0 +1,2 @@
+--
+UPDATE `creature_template` SET `flags_extra` = `flags_extra` |2147483648 where `entry` = 21212;


55Honey (approved):

elthehablo (approved):

Aniquilo:
DespawnOnEvade will make the boss exploitable, actually becouse the respawnTimer is set to 20s, for ref multiple issues already addressed:

https://github.com/azerothcore/azerothcore-wotlk/issues/16670

Nyeriah:
Can you elaborate on how will it make it "exploitable" ?

Aniquilo:
> Can you elaborate on how will it make it "exploitable" ?

Of course:

To exploit this boss you have to wait until the next day or realm reset for it to happen, basically what happens is that the DespawnOnEvade method causes the respawn time to be overwritten and not use the value that is used in the database in the creature table, normally bosses have a 7 day respawn time, all a player has to do is wait for the realm reset or 1 day to pass and come back the next day and the boss will be alive again.

That's why it's so dangerous to use this flag, because regardless of the bosses' respawn time, the boss will be alive again.

That's the reason i attached the issue of Iron Assembly, the comments made by other users, this issue is also present on all the bosses that are currently using this flag.

Nyeriah:
Dead bosses can't evade, so what you are describing here doesn't make any sense. Maybe it does for assembly of iron because there are three bosses and there's a different context.

Aniquilo:
> Dead bosses can't evade, so what you are describing here doesn't make any sense. Maybe it does for assembly of iron because there are three bosses and there's a different context.

Dead bosses can't evade, but they have a respawn timer which is modified by that flag, just .npc info and you will see what i mean with this, it affects all the bosses that use this flag, if the players actually wipe in the boss, it will despawn on evade, and the respawn timer will be modified by the new timer, sorry if i didn't explain it properly, but basically the boss needs to reset and DespawnOnEvade for this to happen.

PR #17993, opened 2023-12-11
fix(Scripts/SSC): make The Lurker Below spout rotate faster again

elthehablo:
## Changes Proposed:

This PR proposes changes to:

Rotation speed lines up with retail sniffs

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/6448

## SOURCE:

## Tests Performed:

This PR has been:

untested

## How to Test the Changes:

1.
2.
3.

### DIFF
--- a/src/server/scripts/Outland/CoilfangReservoir/SerpentShrine/boss_lurker_below.cpp
+++ b/src/server/scripts/Outland/CoilfangReservoir/SerpentShrine/boss_lurker_below.cpp
@@ -254,7 +254,7 @@ class spell_lurker_below_spout : public AuraScript
 
     void CalcPeriodic(AuraEffect const* /*aurEff*/, bool& /*isPeriodic*/, int32& amplitude)
     {
-        amplitude = 1000;
+        amplitude = 200;
     }
 
     void HandleEffectApply(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)


heyitsbench:
Both speed and duration now correct, or does the duration still need correction?

Nyeriah:
The duration is correct, as for the speed it needs more research. Spell procs every 200ms in sniffs

elthehablo:
So 0.1f per tick, (2pi/0.1) so we need 63 rotations to do a full circle. 16 s a tick every 200ms so that is 16*5 = 80 ticks, so 80/63 = 126% of a rotation

elthehablo:
250ms would do just shy of a full rotation

heyitsbench:
When you say needs more research, does that refer to the amount of rotation in terms of degrees?

elthehablo:
Maybe. Hard to tell. Rotation looks smooth on retail but they might have a different implementation of rotation in their core

elthehablo:
As expected he does a little over a circle, which lines up to the videos I saw. Should be good to merge

Nefertumm (approved):

balleny:
turns way to quickly now. takes about 12 seconds for a full circle instead of 16

balleny:
also, he is set to have 3 seconds intro into Spout and 16 seconds rotation.
but, there is also a whirlwind which takes place after 18 seconds (https://github.com/azerothcore/azerothcore-wotlk/blob/d266bf3d15849eb315aab48247035a54a0c51a72/src/server/scripts/Outland/CoilfangReservoir/SerpentShrine/boss_lurker_below.cpp#L138)
so he whirlwinds before finishing Spout.

going by DBM, the overall event should take 22 seconds, hence 19 seconds for the full spout animation (which could equate to 16 or 18 seconds for 360 degree rotation plus the remainer for the additional rotation observed.

PR #18010, opened 2023-12-13
chore: remove Unit::AddAura() func overloads and simplify signature

natrist:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Simplify the Unit API and reduce redundancy in code

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- The anti-cheat module will need to be updated to reflect the API changes

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -4232,7 +4232,7 @@ void Unit::ProcessTerrainStatusUpdate()
     if (curLiquid != _lastLiquid)
     {
         if (_lastLiquid && _lastLiquid->SpellId)
-            RemoveAurasDueToSpell(_lastLiquid->SpellId);
+            RemoveAura(_lastLiquid->SpellId);
 
         // Set _lastLiquid before casting liquid spell to avoid infinite loops
         _lastLiquid = curLiquid;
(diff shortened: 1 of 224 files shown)


pklloveyou:
I don't recommend doing this.
After PR
    `A adds a SPELL for B` Become `B adds a SPELL for B`.

src/server/scripts/Commands/cs_gm.cpp
```diff
@@ -210,7 +210,7 @@ class gm_commandscript : public CommandScript
         }
         else
         {
-            _player->AddAura(VISUAL_AURA, _player);
```
Kitzunu, line 213: But in places like this, when we add the aura to the same "caster" I think this change is good

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -6100,11 +6100,11 @@ void Spell::SummonGuardian(uint32 i, uint32 entry, SummonPropertiesEntry const*
         {
             // Inferno, Warlock summon spell
             case 1122:
-                caster->AddAura(61191, summon);
```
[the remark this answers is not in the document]
heyitsbench: This is also my concern, for the record.

sudlud:
https://github.com/TrinityCore/TrinityCore/commit/8e9d2cdf01929f513e37eccbfdea952aa04e78f6

Just leaving this here as reference where this was originally implemented.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -19113,50 +19113,43 @@ void Unit::GetPartyMembers(std::list<Unit*>& TagUnitMap)
...
-    if (Aura* aura = Aura::TryRefreshStackOrCreate(spellInfo, effMask, target, this))
-    {
-        aura->ApplyForTargets();
-        return aura;
+        if (IsImmunedToSpellEffect(spellInfo, i))
+            effectMask &= ~(1 << i);
     }
-    return nullptr;
+
+    Aura* auraObject = Aura::TryRefreshStackOrCreate(spellInfo, effectMask, this, this);
```
dedmen, line 19140: This looks incorrect.

Ref https://github.com/azerothcore/azerothcore-wotlk/blob/06525cdc1e79cc041583af40b9f923adb7996d64/src/server/game/Spells/Auras/SpellAuras.cpp#L326

Previously "caster" was the object casting the aura. Now its the same as the one receiving it.


The caster is relevant here: 
https://github.com/azerothcore/azerothcore-wotlk/blob/06525cdc1e79cc041583af40b9f923adb7996d64/src/server/game/Entities/Unit/Unit.cpp#L4305
And here:
https://github.com/azerothcore/azerothcore-wotlk/blob/06525cdc1e79cc041583af40b9f923adb7996d64/src/server/game/Spells/Auras/SpellAuras.cpp#L409



If the aura is not singlestack
https://github.com/azerothcore/azerothcore-wotlk/blob/06525cdc1e79cc041583af40b9f923adb7996d64/src/server/game/Entities/Unit/Unit.cpp#L4324
Then it does matter who the caster is, that information is lost in this refactor.

I haven't  checked what the caster info stored in Aura class is used for. But this alrady looks wrong.

natrist: Thanks for the feedback.
I should have tagged my PR as a draft. I am well aware of the issues but the bigger problem is none of this code is elegant, performant or concise and the 'owner' in this context is completely irrelevant.

Nyeriah:
Marking as abandoned since this PR hasn't received any updates in a long while and is still a draft. If there's interest in continuing this PR please leave a message and we will reopen it

Issue #17929, opened 2023-12-03
Bleeding Hollow Archers again


The-GhostRider:
### Current Behaviour

Mostly stationary, only moving to get closer to use their bow and arrow or backing of when their target moves inside of them
Melees when their target is in melee with them, but does not initiate melee movement

### Expected Blizzlike Behaviour

Behaves like 99% of ranged NPCs should
Ranged attacks when the target is outside of ~10yd
Melee attacks when the target is inside of ~10yd, including melee movement like following their target if the target moves away slowly or backing off when stepping inside of them
If the target moves away fast enough or they switch target to someone outside of that range they switch to ranged again

### Source

https://youtu.be/oOCVydMtJ2U?t=56

### Steps to reproduce the problem

.gm on
.go c 202645
Clean a little to test properly
Attack from range the Archers and look at their behaviour
Move into melee and then back out again

### Extra Notes

https://wowgaming.altervista.org/aowow/?npc=17270
As extra notes, I recommend viewing the comments from the original report on CC

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b3e37226745a6e6de308bb7b00f0d2f9c596938e

### Operating system

Ubuntu 20.04

PR #17557, opened 2023-10-22
Improve chase movement

AGandrup:
## Changes Proposed:

Improve creature chase movement so it doesn't appear "laggy" and so creatures does not always stop immediately at max melee range.
See current behaviour here:
https://youtu.be/3f27EWJBPr4

New behaviour:
https://youtu.be/4dKaqg_jPjs

(Note: Videos still uploading/processing so quality might still be bad).

## Issues Addressed:

- Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/17543
- Closes: https://github.com/azerothcore/azerothcore-wotlk/issues/13630

## SOURCE:

The logic to predict chase destination is derived from [ChaseMovementGenerator.cpp](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/blob/master/src/server/game/Movement/MovementGenerators/ChaseMovementGenerator.cpp) so credits to @Ovahlord of [The Cataclysm Preservation Project](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore)

## Tests Performed:

This PR has been:
- [ ] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

Issue #12108, opened 2022-06-19
 Rooted npcs should ignore out of reach targets for current highest threat melee target selection #3634 


Contrubius:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [X] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

Melee NPCs stand while rooted and wait until the root is over to reach the highest threat target if out of range.

[1. Druid rooted nearest melee threat target ignored ](https://streamable.com/avmuot)
[2. Mage Frost Nova Root highest threat target ignored ](https://streamable.com/jaikdm)
### Expected Blizzlike Behaviour

n = rooted npc
h = current threat holder
p = lower on threatlist

n ------------ h
p

n should attack p, because it can't reach h
after the root ends it should go and attack h again (unless someone overthreated by the regular 130% 110% rules relative to h)

### Source
![Rooted target](https://user-images.githubusercontent.com/100415626/174473896-cf4c0ae5-b79d-496d-97e3-a06feafc262d.jpg)
Another WoWWiki guide source.

![image](https://user-images.githubusercontent.com/50233983/174457737-5261b787-491b-489c-b8c4-12f8bda4a6cf.png)
https://www.wowhead.com/spell=122/frost-nova#comments:id=36313

![image](https://user-images.githubusercontent.com/50233983/174457729-ba2bbdbd-f91a-4f1c-84e7-ad69ade1f554.png)
https://wowwiki-archive.fandom.com/wiki/Root

### Steps to reproduce the problem

Assemble 2 controllable units (2 players, player and pet, whatever) and an npc
build some threat on the npc
root the npcs
move the unit highest in threat away from the npc

### Extra Notes

Originally reported: https://github.com/chromiecraft/chromiecraft/issues/3634

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/36f0a7757bde9573262f298aa1eaf6765af8295e

### Operating system
Windows 10.

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[...]

PR #18026, opened 2023-12-17
Fix combat movement

AGandrup:
@heyitsbench This one is for you 😛
First of all, this PR aims to address some underlying issues with combat movement for ranged creatures that became a lot more noticable after implementing [improvements to chase movements](https://github.com/azerothcore/azerothcore-wotlk/pull/17557).

But as always I went a bit out of scope and decided to fix other issues with combat movement and spellcaster behaviour, that I found along the way 😛

Heres some before and after videos to show some of the things that this PR fix:
Before: https://youtu.be/vwS1tz2gk3Y
After: https://youtu.be/1PQGNUMc6A0

## Changes Proposed:

This PR proposes changes to:
- Fix ranged creatures not chasing to melee range when unable to cast spells
- Fix ranged creatures sometimes not moving
- Fix ranged creatures not chasing in melee range
- Removed code related to another attempt at fixing combat movement.
- Removed SMART_ACTION_SET_CASTER_COMBAT_DIST and updated smarts scripts accordingly.
- Fix issue where creatures would always cast their spells regardless of having event chance set below 100% in SAI.
- Fix issue with SAI linked actions being executed even if the spell required to execute it had failed.
- Some more juicy stuff  from cherry picks related to SAI, that I can't remember right now... 😛 

Cherry picks (All credits to mr. @jackpoz and @joshwhedon):
- https://github.com/TrinityCore/TrinityCore/commit/7fb74326205acbc5c0fa6b2aa9731d891e788d64
- https://github.com/TrinityCore/TrinityCore/commit/63a6e1e0480df659e773a3df5e1acc2027ebcf15
- https://github.com/TrinityCore/TrinityCore/commit/ca25e8d0199730c0976ebc37317e9407aceccc34
- https://github.com/TrinityCore/TrinityCore/commit/96b289cadbf304e427ddbd281d1a6f592af612a7

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17929

## Tests Performed:

This PR has been:
- Compiles
- Tested in game

## How to Test the Changes:

- ...

- TODO: Bug hunt. Knowing AC, this will probably end up breaking something else, god knows where 🙃 
- TODO: Fix SAI for ranged creatures in DB - Most of them probably needs to be reworked entirely. Should use SMARTCAST_COMBAT_MOVE flag.

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -56,7 +56,8 @@ SmartScript::SmartScript()
     mTemplate = SMARTAI_TEMPLATE_BASIC;
     mScriptType = SMART_SCRIPT_TYPE_CREATURE;
     isProcessingTimedActionList = false;
-
+    mCurrentPriority = 0;
+    mEventSortingRequired = false;
     _allowPhaseReset = true;
 }
 
@@ -78,6 +79,12 @@ void SmartScript::OnReset()
         {
             InitTimer((*i));
             (*i).runOnce = false;
+        }
+
+        if ((*i).priority != SmartScriptHolder::DEFAULT_PRIORITY)
+        {
+            (*i).priority = SmartScriptHolder::DEFAULT_PRIORITY;
+            mEventSortingRequired = true;
         }
     }
     ProcessEventsFor(SMART_EVENT_RESET);
@@ -106,14 +113,18 @@ void SmartScript::ProcessEventsFor(SMART_EVENT e, Unit* unit, uint32 var0, uint3
 
 void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, uint32 var1, bool bvar, SpellInfo const* spell, GameObject* gob)
 {
+    e.runOnce = true;//used for repeat check
+
     //calc random
-    if (e.event.event_chance < 100 && e.event.event_chance)
+    if (e.event.event_chance < 100 && e.event.event_chance && !(e.event.event_flags & SMART_EVENT_FLAG_TEMP_IGNORE_CHANCE_ROLL))
     {
         uint32 rnd = urand(1, 100);
         if (e.event.event_chance <= rnd)
             return;
     }
-    e.runOnce = true;//used for repeat check
+
+    // Remove SMART_EVENT_FLAG_TEMP_IGNORE_CHANCE_ROLL flag after processing roll chances as it's not needed anymore
+    e.event.event_flags &= ~SMART_EVENT_FLAG_TEMP_IGNORE_CHANCE_ROLL;
 
     if (unit)
         mLastInvoker = unit->GetGUID();
@@ -577,6 +588,8 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
             if (e.action.cast.targetsLimit)
                 Acore::Containers::RandomResize(targets, e.action.cast.targetsLimit);
 
+            bool failedSpellCast = false, successfulSpellCast = false;
+
             for (WorldObject* target : targets)
             {
                 // may be nullptr
@@ -609,6 +622,23 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                     if (e.action.cast.castFlags & SMARTCAST_INTERRUPT_PREVIOUS)
                         me->InterruptNonMeleeSpells(false);
 
+                    SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(e.action.cast.spell);
+                    float distanceToTarget = me->GetDistance(target->ToUnit());
+                    float spellMaxRange = me->GetSpellMaxRangeForTarget(target->ToUnit(), spellInfo);
+                    float spellMinRange = me->GetSpellMinRangeForTarget(target->ToUnit(), spellInfo);
+
+                    // Let us not try to cast spell if we know it is going to fail anyway. Stick to chasing and continue.
+                    if (distanceToTarget > spellMaxRange) {
+                        failedSpellCast = true;
+                        CAST_AI(SmartAI, me->AI())->SetCombatMove(true, spellMaxRange - 0.1f); // -0.1f to fix rounding
+                        continue;
+                    }
+                    else if (distanceToTarget < spellMinRange) {
+                        failedSpellCast = true;
+                        CAST_AI(SmartAI, me->AI())->SetCombatMove(true);
+                        continue;
+                    }
+
                     TriggerCastFlags triggerFlags = TRIGGERED_NONE;
                     if (e.action.cast.castFlags & SMARTCAST_TRIGGERED)
                     {
@@ -624,16 +654,32 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                     if (e.action.cast.castFlags & SMARTCAST_COMBAT_MOVE)
                     {
                         // If cast flag SMARTCAST_COMBAT_MOVE is set combat movement will not be allowed unless target is outside spell range, out of mana, or LOS.
-                        CAST_AI(SmartAI, me->AI())->SetCombatMove(spellCastFailed);
+                        if (result == SPELL_FAILED_OUT_OF_RANGE)
+                            // if we are just out of range, we only chase until we are back in spell range.
+                            CAST_AI(SmartAI, me->AI())->SetCombatMove(true, spellMaxRange - 0.1f); // -0.1f to fix rounding
+                        else
+                            // if spell fail for any other reason, we chase to melee range.
+                            CAST_AI(SmartAI, me->AI())->SetCombatMove(spellCastFailed);
                     }
 
                     if (spellCastFailed)
-                        RecalcTimer(e, 500, 500);
+                        failedSpellCast = true;
+                    else
+                        successfulSpellCast = true;
 
                     LOG_DEBUG("scripts.ai", "SmartScript::ProcessAction:: SMART_ACTION_CAST: Unit {} casts spell {} on target {} with castflags {}",
                               me->GetGUID().ToString(), e.action.cast.spell, target->GetGUID().ToString(), e.action.cast.castFlags);
                 }
             }
+
+            // If there is at least 1 failed cast and no successful casts at all, retry again on next loop
+            if (failedSpellCast && !successfulSpellCast)
+            {
+                RetryLater(e, true);
+                // Don't execute linked events
+                return;
+            }
+
             break;
         }
         case SMART_ACTION_SELF_CAST:
@@ -4538,7 +4584,8 @@ void SmartScript::UpdateTimer(SmartScriptHolder& e, uint32 const diff)
             {
                 if (me && me->HasUnitState(UNIT_STATE_CASTING))
                 {
-                    e.timer = 1200;
+                    //e.timer = 1200;
+                    RaisePriority(e);
                     return;
                 }
             }
@@ -4596,6 +4643,18 @@ void SmartScript::UpdateTimer(SmartScriptHolder& e, uint32 const diff)
                     break;
                 }
         }
+
+        if (e.priority != SmartScriptHolder::DEFAULT_PRIORITY)
+        {
+            // Reset priority to default one only if the event hasn't been rescheduled again to next loop
+            if (e.timer > 1)
+            {
+                // Re-sort events if this was moved to the top of the queue
+                mEventSortingRequired = true;
+                // Reset priority to default one
+                e.priority = SmartScriptHolder::DEFAULT_PRIORITY;
+            }
+        }
     }
     else
         e.timer -= diff;
@@ -4623,6 +4682,12 @@ void SmartScript::OnUpdate(uint32 const diff)
         return;
 
     InstallEvents();//before UpdateTimers
+
+    if (mEventSortingRequired)
+    {
+        SortEvents(mEvents);
+        mEventSortingRequired = false;
+    }
 
     for (SmartAIEventList::iterator i = mEvents.begin(); i != mEvents.end(); ++i)
         UpdateTimer(*i, diff);
@@ -4677,6 +4742,33 @@ void SmartScript::OnUpdate(uint32 const diff)
         }
         else mTextTimer -= diff;
     }
+}
+
+void SmartScript::SortEvents(SmartAIEventList& events)
+{
+    std::sort(events.begin(), events.end());
+}
+
+void SmartScript::RaisePriority(SmartScriptHolder& e)
+{
+    e.timer = 1;
+    // Change priority only if it's set to default, otherwise keep the current order of events
+    if (e.priority == SmartScriptHolder::DEFAULT_PRIORITY)
+    {
+        e.priority = mCurrentPriority++;
+        mEventSortingRequired = true;
+    }
+}
+
+void SmartScript::RetryLater(SmartScriptHolder& e, bool ignoreChanceRoll)
+{
+    RaisePriority(e);
+
+    // This allows to retry the action later without rolling again the chance roll (which might fail and end up not executing the action)
+    if (ignoreChanceRoll)
+        e.event.event_flags |= SMART_EVENT_FLAG_TEMP_IGNORE_CHANCE_ROLL;
+
+    e.runOnce = false;
 }
 
 void SmartScript::FillScript(SmartAIEventList e, WorldObject* obj, AreaTrigger const* at)
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -57,14 +57,6 @@ SmartScript::SmartScript()
     mScriptType = SMART_SCRIPT_TYPE_CREATURE;
     isProcessingTimedActionList = false;
 
-    // Xinef: Fix Combat Movement
-    mActualCombatDist = 0;
-    mMaxCombatDist = 0;
-
-    smartCasterActualDist = 0.0f;
-    smartCasterMaxDist = 0.0f;
-    smartCasterPowerType = POWER_MANA;
-
     _allowPhaseReset = true;
 }
 
@@ -91,10 +83,6 @@ void SmartScript::OnReset()
     ProcessEventsFor(SMART_EVENT_RESET);
     mLastInvoker.Clear();
     mCounterList.clear();
-
-    // Xinef: Fix Combat Movement
-    RestoreMaxCombatDist();
-    RestoreCasterMaxDist();
 }
 
 void SmartScript::ProcessEventsFor(SMART_EVENT e, Unit* unit, uint32 var0, uint32 var1, bool bvar, SpellInfo const* spell, GameObject* gob)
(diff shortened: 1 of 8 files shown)


elthehablo:
Awesome! Let me run some tests

Gultask:
Hey I haven't checked out yet but I've gathered some info some time ago from Retail

- When in melee range, they engage in melee and keep up with players if they're within ~10y
- When silenced, start melee attacks
- When player moves out of casting range they move in range, closer than maximum cast range (spell is 40y, they move to 30y)
- When player is rooted, they'll move out of melee range (~10y)
- When player is incapacitated, they're temporarily removed from the threat table of all while under the effects.
- When in melee range, SheatheState is changed
- These movement updates happen in 1200ms intervals. Possibly every movement update does so. This is most easily seen in formations. Members update their follow position every 1200ms (3 packet cycles)

By eye, it seems that the majority of proximity-based delays are between 1 or 2 standard timers.
So if the timer is 3s, when nearby it'll be either 6s or 9s
Of course, AI Updates in Blizz are every 1200ms so I'm approximating.
Interrupts do not influence timers ime

AGandrup:
Hey, good to know. Will probably have to do a few adjustments then. For example right now I just made casters move to their max spell range if they are out of range. They probably have to move to maxSpellRange-chaseRange, but thats just me guessing.

Didnt check with the root mechanic. Thinking about it I do remember ranged creatures behave like that, but unsure if its true for all of them - if not it is probably something that must be handled in SAI?

About the 1200ms, I was wondering where that number came from. AC already use this number but i kinda thought it was some hackfix related to the previous attempt at fixing combat movement, lol. I did change it in one place when I cherry picked one of the commits from TC, so will have to change it back.

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -649,11 +648,38 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
...
-                    me->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlags);
+                    SpellCastResult result = me->CastSpell(target->ToUnit(), e.action.cast.spell, triggerFlags);
+                    bool spellCastFailed = (result != SPELL_CAST_OK && result != SPELL_FAILED_SPELL_IN_PROGRESS);
+
+                    if (e.action.cast.castFlags & SMARTCAST_COMBAT_MOVE)
+                    {
+                        // If cast flag SMARTCAST_COMBAT_MOVE is set combat movement will not be allowed unless target is outside spell range, out of mana, or LOS.
+                        if (result == SPELL_FAILED_OUT_OF_RANGE)
+                            // if we are just out of range, we only chase until we are back in spell range.
+                            CAST_AI(SmartAI, me->AI())->SetCombatMove(true, spellMaxRange - 0.1f); // -0.1f to fix rounding
```
Gultask, line 659: I'd suggest ~0.75*spellMaxRange

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -4576,7 +4584,8 @@ void SmartScript::UpdateTimer(SmartScriptHolder& e, uint32 const diff)
             {
                 if (me && me->HasUnitState(UNIT_STATE_CASTING))
                 {
-                    e.timer = 1200;
+                    //e.timer = 1200;
+                    RaisePriority(e);
```
Gultask, line 4588: the 1200 timer was to emulate Blizz 😭

AGandrup: Got it. I didnt delete it because i wasnt sure what it was 😝

Tereneckla:
There also is this issue with combat targeting if you want to do more https://github.com/azerothcore/azerothcore-wotlk/issues/12108 ;)

Gultask:
> Didnt check with the root mechanic. Thinking about it I do remember ranged creatures behave like that, but unsure if its true for all of them - if not it is probably something that must be handled in SAI?

Don't have to if you don't want to :p
it's just some details we caught, can fix later in another PR too, your choice

> About the 1200ms, I was wondering where that number came from. AC already use this number but i kinda thought it was some hackfix related to the previous attempt at fixing combat movement, lol. I did change it in one place when I cherry picked one of the commits from TC, so will have to change it back.

Ya Blizz updates AI every 1200ms, you can see it in sniffs, spells are only cast in 1200ms intervals, the shortest amount of time between casts is also 1200ms, the only exception are On Aggro spells which do exist as well and are instantaneous.
I think it's 3 packet cycles the proper term?

src/server/game/AI/SmartScripts/SmartScript.cpp
elthehablo: Agreed. Otherwise cheesing is quite easy when you can just do 1 step to outrange every time

Gultask: In my experience the chase movement for 40y spell was 30y so 🙈

AGandrup: I agree they should go further than max spell range, but 0.75 is a magic number imo. Is there any way we can find the correct chase range 🤔 
I also need to make sure then range doesnt become shorter than min spell range. Bad example: if spell max range is is 40yd and min is 35yd we cannot chase down to 30yd because then spell would become unusable. The solution here is simple but I wonder if there are more edge cases to be aware of - there probably is 😛

AGandrup: Yes, we can simply use std::max(spellMaxRange*0.75, spellMinRange), but its back to the question if this is correct behaviour. Im fine with it personally.

elthehablo: I personally have no idea. Might be able to do some testing tonight. What Gultask said made sense to me though

elthehablo: Good on you for wanting proof though. I agree with that notion

AGandrup:
> When player is rooted, they'll move out of melee range (~10y)

@Gultask I am looking a bit into this but I have some questions, perhaps you know the answer:
When moving back, do they run to a random location in any direction or do they just do a 180? If memory serves me right, they do a 180 but I haven't played in ages - it could also be that they dont do 180 but rather just move backwards like a player holding the 'S' key, however I don't recall ever seeing that, at least not in vanilla, maybe in wotlk?
Do they always run equally far? I mean, if they always run to minimum spell range, the answer is no, but perhaps there is a fixed range?
Also how far are they allowed to run? You said ~10yd yourself, but what if min spell range is larger than that? Should they just not run at all?
Is it only the creature casting the root effect that run or is it every ranged creature within melee range?
Should the creature always move? If creature tries to move back but it is invalid position (for example standing up against a wall) should it look for another direction to move or should it just not move at all then?

Sorry that was a lot of questions, there's probably more to come 😛

AGandrup:
> There also is this issue with combat targeting if you want to do more #12108 ;)

I'll see what I can do but wont promise anything 👀

Gultask:
> When moving back, do they run to a random location in any direction or do they just do a 180?

180 and move away. I've never seen a backpedal myself

> Do they always run equally far?

Now this I'm unsure but to me it *looks* like it. It's hard to be sure because distance calculation is a bit iffy on retail, since you need to use addons and it's never perfect. As for spell range the majority of it consists of those ranged attacks like bows and guns. I think most have a 8y minimum range or something along those lines. By eye I think that 10y was the distance I observed them moving away as well. Far enough that they can fire but not close enough for melee players to hit them.

> Is it only the creature casting the root effect that run or is it every ranged creature within melee range?

Every ranged creature that has the rooted player as its victim. If it's not attacking the rooted player the creatures ignore them.

> Should the creature always move? If creature tries to move back but it is invalid position (for example standing up against a wall) should it look for another direction to move or should it just not move at all then?

Good question. I probably have to test this somewhere.

> Are there (other) cases where creatures don't move? For example if it gets silenced right after using root.

Dunno. Silenced creatures engage in melee, but simple interrupts w/o silence I can't quite recall.

AGandrup:
> Every ranged creature that has the rooted player as its victim. If it's not attacking the rooted player the creatures ignore them.

Okay, how soon does this happen? Once the root is applied or on next attempt at spell cast?
So a or b:
A) player is rooted, all ranged creatures within melee range immediately stop what they are doing and move out of range right away
B) player is rooted, ranged creature checks if player is rooted and run if true, then cast spell

AGandrup:
> Dunno. Silenced creatures engage in melee, but simple interrupts w/o silence I can't quite recall.

Hmm, I have to make check if creatures next spell can be silenced.

elthehablo:
> > Every ranged creature that has the rooted player as its victim. If it's not attacking the rooted player the creatures ignore them.
> 
> Okay, how soon does this happen? Once the root is applied or on next attempt at spell cast? So a or b: A) player is rooted, all ranged creatures within melee range immediately stop what they are doing and move out of range right away B) player is rooted, ranged creature checks if player is rooted and run if true, then cast spell

Yep the latter. This should happen with creatures that throw stuff or cast bows and have like a net effect.

We're talking, at least originally, about the reverse effect. When a creature gets rooted, they should attack whomever is in melee range. This currently doesn't happen on AC. But tbh this is far out of the scope of this PR.

elthehablo:
If you could take a look at that as well, that would be lovely though :p

elthehablo:
Maybe when rooted, wipe threat temporarily (save threat list for when root breaks) fire a scheduler that checks for hostile players in melee range and tries to put them on a new threat list

AGandrup:
> Maybe when rooted, wipe threat temporarily (save threat list for when root breaks) fire a scheduler that checks for hostile players in melee range and tries to put them on a new threat list

Hmm, ye could be one approach. My initial thought would just be to create a new player list searcher, which looks for all players within melee range and then returns the player with highest threat.

elthehablo:
Do we think someone is going to do detailed tested on classic for the range? I highly doubt it personally. What is the status of this beyond that?

elthehablo:
Because afk you can't really easily determine range on the classic ptr

heyitsbench:
> Do we think someone is going to do detailed tested on classic for the range?

Not without creature IDs to test with.

> Because afaik you can't really easily determine range on the classic ptr

Could always sniff it.

elthehablo:
> > Do we think someone is going to do detailed tested on classic for the range?
> 
> Not without creature IDs to test with.
> 
> > Because afaik you can't really easily determine range on the classic ptr
> 
> Could always sniff it.

Yeah you are right. Lemme try something

elthehablo:
Sorry wasn't able to find anything myself

elthehablo:
Anyone that can compare this with retail? Accurately? I tried some stuff but casting distance is hard to measure in sniffs imo

elthehablo:
The behaviour seems as follows:

- casters should always try to move to around 75% of max cast range
- this means that two things can happen when a player moves whilst the creature is casting:
    1. when the player moves between 80%-100% of max cast range, the creature will finish its cast and then move to 75% again
    2. when the player moves to +100% of max cast range, the creature will fail the cast and move to 75% again
    3. as we can see from 1. there seems to be a 5y (guesstimate) where the creature doesn't move
Sources:
soonTM

If you could put this behaviour or your spin on it in the PR, we can merge this. Nice work on this, again, and this will mean you supplied AC with a feature that was really sought after for a long time.

AGandrup:
Perhaps you can use Grim Reach or Destructive Reach talents to get +3-6yd on 30yard destruction or affliction spells. That you can maybe use to get a more exact way to measure when he starts to move when already within his max spell range

heyitsbench:
> If you run just slightly outside of that range so he starts chasing, can you check if he runs in range of your 30yard spells?

Somewhat inconsistent it seems, but more often than not he does not. He stops closer to ~35y away. Sometimes he does land at ~25y or so.

AGandrup:
On my local changes I just went with the lazy approach of substracting 5 yard (melee range) from spellMaxRange, because I decided if spellMaxRange 5 yard or shorter, and we are outside that range, we should chase with 0yd chase range, which is the default.

Anyway, the functionality is all there more or less, so its just a matter of fine tuning it to be as blizzlike as possible.

I've also fixed the combat movement for casters when target is rooted, I just haven't pushed it yet because of Christmas stuff and all that 😛

AGandrup:
> > If you run just slightly outside of that range so he starts chasing, can you check if he runs in range of your 30yard spells?
> 
> Somewhat inconsistent it seems, but more often than not he does not. He stops closer to ~35y away. Sometimes he does land at ~25y or so.

I see. So it can't be a fixed value - or it can, but then there is some logic missing. Idk, we can go with 0.75 for all I care. Its not like players are going to notice anyway. But if we can guess the logic behind it, I would be happy to implement it.

heyitsbench:
> So it can't be a fixed value - or it can, but then there is some logic missing.

I don't know, my initial suspicion was that it's a problem of timing, but it doesn't seem to be that. It definitely just seems to be a randomly selected distance, sometimes close to ~15y (all by eye by the way, not exact distances), other times ~35y, ~25y, etc etc. *Maybe* it's related to the geometry surrounding the path of glory, but I really can't discern a pattern.

AGandrup:
> > So it can't be a fixed value - or it can, but then there is some logic missing.
> 
> I don't know, my initial suspicion was that it's a problem of timing, but it doesn't seem to be that. It definitely just seems to be a randomly selected distance, sometimes close to ~15y (all by eye by the way, not exact distances), other times ~35y, ~25y, etc etc. _Maybe_ it's related to the geometry surrounding the path of glory, but I really can't discern a pattern.

Hmm. Sounds like it could be related to your speed (and how far you moved (distance) between server tick), in order to try keep up with you. So for example if you run with mount i.e. 100% in a straight line, I am guessing he will chase closer to you, than if you ran on foot. 

But ye, just guesswork.

elthehablo:
Yeah or for all we know they added some randomness to "distance to walk from spellmaxrange to target where to start casting again" :p anyway enjoy your Christmas lads. Nice work on adding the rooted part too

elthehablo:
Yeah or for all we know they added some randomness to "distance to walk from spellmaxrange to target where to start casting again" :p anyway enjoy your Christmas lads. Nice work on adding the rooted part too

AGandrup:
Aight I've pushed my changes now so you can try for yourself. Here's a demo as well: https://youtu.be/MSrddL7gskM

elthehablo (approved):

boom8866:
The harpoon objid: 192175 used in boss 26693 battle cannot be activated because no harpoon has been dropped

Nyeriah:
Hello @boom8866 !

Thanks for your contribution. Could you give us a little more information to help us address your issue? What makes you believe this code has led to the issues you are experiencing?

boom8866:
@Nyeriah  My player encountered a problem in the game where SAI's target casting spells were sometimes ineffective

pklloveyou:
Not only that, but also many errors occurred after PR

boom8866:
@AGandrup There are many strange problems! It's really hard to point out

AGandrup:
> @AGandrup There are many strange problems! It's really hard to point out

Well can't really do anything with that information. Surely you can be a bit more specific? Like what are you experincing? What npc etc?
Perhaps you can record it?

boom8866:
The problem I currently know is that when multiple players trigger the target spell condition of SAI in battle, SAI's casting of spells on the target will be ineffective，Then I tested alone and didn't find any problems,...So it's strange

pklloveyou:
> > Not only that, but also many errors occurred after PR
> 
> Such as?

quest id 13190

AGandrup:
> The problem I currently know is that when multiple players trigger the target spell condition of SAI in battle, SAI's casting of spells on the target will be ineffective，Then I tested alone and didn't find any problems,...So it's strange

What npc are we talking about? And what does ineffective mean? Does it fail to cast spell or does it cast but nothing happens? Does it happen no matter what or only under some circumstances, for example its target is rooted.

Btw, as stated in the PR, some SAI's may need to be reworked.

elthehablo:
> > The problem I currently know is that when multiple players trigger the target spell condition of SAI in battle, SAI's casting of spells on the target will be ineffective，Then I tested alone and didn't find any problems,...So it's strange
> 
> What npc are we talking about? And what does ineffective mean? Does it fail to cast spell or does it cast but nothing happens? Does it happen no matter what or only under some circumstances, for example its target is rooted.
> 
> Btw, as stated in the PR, some SAI's may need to be reworked.

If it is just the latter there is no need for us to think about remedying this specific PR. We can just fix the few SAI events that this potentially breaks. We will need to make a list of things that are broken then.

If people that discovered problems can make issues and link to here that would be splendid

AGandrup:
Ye. Either way I need more information if I should look into this. I am a bit busy these days so I don't have much time to sit and debug let alone fix the problem.

elthehablo:
Dw it is not your resonsibility alone

elthehablo:
> > > Not only that, but also many errors occurred after PR
> > 
> > 
> > Such as?
> 
> quest id 13190

You're correct in saying this doesn't seem to be working anymore. Yet, are you sure this was caused by this PR? And if so, why do you think so?

pklloveyou:
> > > > Not only that, but also many errors occurred after PR
> > > 
> > > 
> > > Such as?
> > 
> > 
> > quest id 13190
> 
> You're correct in saying this doesn't seem to be working anymore. Yet, are you sure this was caused by this PR? And if so, why do you think so?

After delete PR, it works

pklloveyou:
> > > > > Not only that, but also many errors occurred after PR
> > > > 
> > > > 
> > > > Such as?
> > > 
> > > 
> > > quest id 13190
> > 
> > 
> > You're correct in saying this doesn't seem to be working anymore. Yet, are you sure this was caused by this PR? And if so, why do you think so?
> 
> After delete PR, it works

image

AGandrup:
Meh, probably just an edge case I forgot to check for. I'll fix it when I get time.

elthehablo:
Seems odd to me. It's a friendly dummy creature casting on TARGET_INVOKER_PARTY

skelUA:
@AGandrup  this PR Broke this quest 
https://github.com/azerothcore/azerothcore-wotlk/issues/18181
 , some problem in LOS

sudlud:
@AGandrup 
I think I can also confirm an issue with SmartAI creature targetting after this PR

testing on d0eae390b

I added the following simple SmartAI script which casts a spell on a target creature once after a timeout of 420ms:

```sql
UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 29802;

DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 29802);
INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
(29802, 0, 0, 0, 1, 0, 100, 0, 420, 420, 0, 0, 0, 0, 11, 55281, 2, 0, 0, 0, 0, 9, 29807, 0, 42, 0, 0, 0, 0, 0, 'Cosmetic Toy Plane - Out of Combat - Cast \'Toy Plane Cable\'');
```

This spell connects the toy plane 29802 with the hook (creature_template id 29807) on the ceiling in the dalaran toy store:
.go c id 29478

Observe theres no "cable" connecting the plane with the hook on the ceiling.

Then i reverted the commits
9a80b7e1f
8f127f9e2

And it works as expected

Tereneckla:
> If a hunter's pet attack a ranged creature,they will rotate together very fast~WTF

That happened with permanently rooted creatures before this PR already

kissingers:
This problem might have some relation with the PR
https://github.com/azerothcore/azerothcore-wotlk/issues/18234

kissingers:
might here break the next SAI link action
image

elthehablo:
> might here break the next SAI link action
> image
> 

Okay that is idd probably why but it shouldn't break stuff when the spells properly execute. Then in most cases something else is wrong. Buy yes we can try removing the link and see if it works

elthehablo:
Overall this PR did break some things. But so far I have not really seen things that can't be relatively easily remedied

PR #18163, opened 2024-01-13
feat(Core/Hook): New hook OnBeforePlayerLogout

NathanHandley:
## Changes Proposed:
Adding a new hook "OnBeforePlayerLogout" which fires _before_ the logout logic is executed for a player inside WorldSession::LogoutPlayer.  This is to support a mod I made which allows a character to change their class, which you can find [here](https://github.com/NathanHandley/mod-multi-class).

There is logic to perform before the logout save event, which made utilizing the existing OnPlayerLogout non-viable (visual item changes on class switch were not being pushed and registered on the client).  Since the save hook does not have the "logout" bool available, I was left with a choice to add this new hook or update/overload that hook.  It seems that a OnBeforeLogout could be useful for more hypothetical use cases, so I elected with this approach.  Also, this allows it to work in the situation that save on logout isn't enabled for some reason.

## Issues Addressed:
none

## SOURCE:
n/a

## Tests Performed:
- Tested on both an Ubuntu 22.04 and Windows 10 Pro server

## How to Test the Changes:
The mod [here](https://github.com/NathanHandley/mod-multi-class) uses this hook.  Install that mod, and switch your class while wearing gear.

none

### DIFF
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -361,7 +361,7 @@ void ScriptMgr::OnPlayerLoadFromDB(Player* player)
     });
 }
 
-void ScriptMgr::OnPlayerBeforeLogout(Player* player)
+void ScriptMgr::OnBeforePlayerLogout(Player* player)
 {
     ExecuteScript<PlayerScript>([&](PlayerScript* script)
         {
--- a/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
@@ -361,11 +361,11 @@ void ScriptMgr::OnPlayerLoadFromDB(Player* player)
     });
 }
 
-void ScriptMgr::OnPlayerPreLogout(Player* player)
+void ScriptMgr::OnPlayerBeforeLogout(Player* player)
 {
     ExecuteScript<PlayerScript>([&](PlayerScript* script)
         {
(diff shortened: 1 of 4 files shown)


src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -361,6 +361,14 @@ void ScriptMgr::OnPlayerLoadFromDB(Player* player)
     });
 }
 
+void ScriptMgr::OnPlayerPreLogout(Player* player)
```
Kitzunu, line 364:
```suggestion
void ScriptMgr::OnPlayerBeforeLogout(Player* player)
```

NathanHandley: Changed per review

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -361,6 +361,14 @@ void ScriptMgr::OnPlayerLoadFromDB(Player* player)
     });
 }
 
+void ScriptMgr::OnPlayerBeforeLogout(Player* player)
```
Kitzunu, line 364:
```suggestion
void ScriptMgr::OnBeforePlayerLogout(Player* player)
```

My bad! Doing several things at the same time.

NathanHandley: No biggie.  It's an easy change to make.  Thanks for reviewing.

Kitzunu (approved):

Issue #11690, opened 2022-05-08
Core/DBUpdater: Skipping over Squash commits result in lost SQL files.


Kitzunu:
### Current Behaviour

If you update your core from commit A -> C skipping the stop at B where the squash is you will lose the DB updates in between. Which will cause continued bugs or error in startup if there was structure changes in between.

### Expected Blizzlike Behaviour

Updating A -> C should still import the unimported archived updates that were before the squash.

### Steps to reproduce the problem

Be on commit A, squash, skip to a newer commit.

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/f742604aeed129cb7a542b5c21fc8038a5fc7940

### Operating system

W11

Kitzunu:
I think the issue might be the directory structure we have in /archive/. The Updater **always** runs non-hashed archived updates, in the `$/data/sql/archive/db_name` directory. However we dump our files in folders called f.ex 9.x which the updater can't find, thus these files are not run.

So if we dump the archived changes into the db_name dir, then I believe that should solve this issue

PR #18197, opened 2024-01-20
feat(DB): Release ACDB 10.0.0

Kitzunu:
# NEW SQUASH METHOD!

- closes https://github.com/azerothcore/azerothcore-wotlk/issues/11690

> [!CAUTION]
> These steps are only for project maintainers who intend to update base files.

We ONLY squash into BASE files. We never move files.
ARCHIVE dir is UNUSED.
All update files ALWAYS exist in the updates dir.

as the `updates` table in base files always will contain the entries from the updates dir they will never be run again on a clean setup. 

How to do the squash.

> [!IMPORTANT]
> A squash needs to be done on a clean database. Drop all tables in Auth, Characters and World.

1. Update the acore.json file. Increment version by one
2. Create a new file in the updates/db_world/ dir, the file should be incremented containing

> [!NOTE]
> It should NOT be placed as a pending update and should only be done on world db. This is because we want the updated DB version to follow into the base files

```sql
UPDATE `version` SET `db_version`='ACDB 335.11-dev', `cache_id`=11 LIMIT 1;
```
> [!NOTE]
> Remember to increment the db_version and cache_id the same as acore.json

4. Drop all your databases, and run Worldserver to populate a clean database.
5. Export the databases using i.e HeidiSQL

> [!IMPORTANT]
> Set the following values
> Tables -> DROP + CREATE
> Data -> Delete + insert (truncate existing data)
> Max INSERT size -> 1024
> This is so that no unexpected issues occur.

6. Move the exported table files into the base directory to update the existing files.
7. Make a PR


Nyeriah (approved):

Kitzunu:
found an issue

Kitzunu:
By doing this https://github.com/azerothcore/azerothcore-wotlk/pull/18198 should hopefully fix our issues

Issue #18113, opened 2024-01-05
From Docker... AzerothCore does not support MySQL versions below 5.7 or MariaDB versions below 10.5.


wbrione:
### Current Behaviour

The title speaks for itself... In addition, my database is MariaDB 10.6

### Expected Blizzlike Behaviour

Well... Be able to use the MariaDB 10.6 database according to the documented compatibility matrix ;-)

### Source

Well, I tried to change the Dockerfile, adding the packages required to use the MariaDB database, but I probably did the update in the wrong place ;-)

### Steps to reproduce the problem

After compiling the images, the error already occurs in the first attempt to access the database (DBImport.log)...

...

AzerothCore rev. 33ef56b00210+ 2024-01-05 09:48:50 +0000 (master branch) (Unix, RelWithDebInfo, Static) (dbimport)
<Ctrl-C> to stop.

(...)

     AzerothCore 3.3.5a  -  www.azerothcore.org

> Using configuration file:       /azerothcore/env/dist/etc/dbimport.conf
> Using SSL version:              OpenSSL 3.0.2 15 Mar 2022 (library: OpenSSL 3.0.2 15 Mar 2022)
> Using Boost version:            1.74.0
 
Configuration field WorldDatabaseInfo was overridden with environment variable.
Configuration field CharacterDatabaseInfo was overridden with environment variable.
Configuration field LoginDatabaseInfo was overridden with environment variable.
Configuration field LogsDir was overridden with environment variable.
Opening DatabasePool 'acore_auth'. Asynchronous connections: 1, synchronous connections: 1.
MySQL client library: 8.0.35
MySQL server ver: 5.5.5-10.6.16-MariaDB-1:10.6.16+maria~ubu2004 
Connected to MySQL database at mariadb
AzerothCore does not support MySQL versions below 5.7 or MariaDB versions below 10.5.

Found server version: 50505. Server compiled with: 80035.
DatabasePool Login NOT opened. There were errors opening the MySQL connections. Check your log file for specific errors

###########################

From my database container...

root@a21acac0df43:/# mysql -V
mysql  Ver 15.1 Distrib 10.6.16-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

###########################

So... If MariaDB 10.5+ is support... Why docker image has not have the requirements already installed (e.g.: libmariadb-dev-compat, mariadb-client, libmariadbclient-dev, and so) ???

### Extra Notes

[...]

PR #18201, opened 2024-01-20
Fix(core): Support MariaDB 10.x when using mysql client libraries

michaeldelago:
## Changes Proposed:

This PR proposes changes to:

Change the way the MySQL/MariaDB's version is checked to account for the `5.5.5` prefix that mariadb `10.x` uses

## Issues Addressed:

- Related to issue https://github.com/azerothcore/azerothcore-wotlk/issues/18113 

## SOURCE:

error appears when setting the database container to `mariadb:10.6`

```
Opening DatabasePool 'acore_auth'. Asynchronous connections: 1, synchronous connections: 1.
Database "acore_auth" does not exist
Do you want to create it? [yes (default) / no]:
AzerothCore does not support MySQL versions below 5.7 or MariaDB versions below 10.5.

Found server version: 50505. Server compiled with: 80035.
DatabasePool Login NOT opened. There were errors opening the MySQL connections. Check your log file for specific errors
...
MySQL server ver: 5.5.5-10.6.16-MariaDB-1:10.6.16+maria~ubu2004 
```

Notice on the last line that it sees the Mariadb Version with the `5.5.5-` prefix. Newer versions of mariadb don't actually do this - this was changed in MariaDB 11.0

## Tests Performed:

This PR has been tested with MariaDB versions:

- 10.5
- 10.6
- 10.11
- 11.2

## How to Test the Changes:

1. checkout master
2. create `docker-compose.override.yaml` file with contents:

```yaml
services:
  ac-database:
    image: mariadb:10.6
```

3. run `docker compose up --build`. The db import will fail due to the version check not working properly
4. checkout this branch
5. run `docker compose up --build`. The db import will fail for a different reason

- MariaDB is actually broken for a different reason - the collation of the `db_world.item_dbc` table isn't compatible. Issue incoming for that

### DIFF
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -383,7 +383,7 @@ void DatabaseWorkerPool<T>::KeepAlive()
 *
 * Adapted from stackoverflow response
 * https://stackoverflow.com/a/2941508
-* 
+*
 * @param mysqlVersion The output from GetServerInfo()/mysql_get_server_info()
 * @return Returns true if the Server version is incompatible
 */
@@ -391,7 +391,7 @@ bool DatabaseIncompatibleVersion(std::string const mysqlVersion)
 {
     // anon func to turn a version string into an array of uint8
     // "1.2.3" => [1, 2, 3]
-    auto parse = [](const std::string& input)
+    auto parse = [](std::string const& input)
     {
         std::vector<uint8> result;
         std::istringstream parser(input);
--- a/src/server/database/Database/DatabaseWorkerPool.cpp
+++ b/src/server/database/Database/DatabaseWorkerPool.cpp
@@ -371,17 +371,22 @@ void DatabaseWorkerPool<T>::KeepAlive()
         Enqueue(new PingOperation);
 }
 
-// Returns true if the version string given is incompatible
-//
-// Intended to be used with mysql_get_server_info()'s output as the source
-//
-// DatabaseIncompatibleVersion("8.0.35") => false
-// DatabaseIncompatibleVersion("5.6.6") => true
-// DatabaseIncompatibleVersion("5.5.5-10.5.5-MariaDB") => false
-// DatabaseIncompatibleVersion("5.5.5-10.4.0-MariaDB") => true
-//
-// Adapted from stackoverflow response
-// https://stackoverflow.com/a/2941508
+/**
+* @brief Returns true if the version string given is incompatible
+*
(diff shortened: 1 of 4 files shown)


src/server/database/Database/DatabaseWorkerPool.cpp
```diff
@@ -377,6 +379,52 @@ void DatabaseWorkerPool<T>::KeepAlive()
...
+    std::string minVersion = "5.7.0";
+
+    // If the version string contains "MariaDB", use that
+    if (mysqlVersion.find("MariaDB") != std::string::npos)
+    {
+        // All MariaDB 10.X versions have a prefix of 5.5.5 from the
+        // mysql_get_server_info() function
+        if (mysqlVersion.rfind("5.5.5-", 0) == 0)
+            offset = 6;
+        minVersion = "10.5.0";
```
Kitzunu, line 419: Instead of hardcoding minVersion here, can we move the definition to a header file so they are all stored together. Come the day we change requirements it will be a pain in the ass to find all the definition.

With that, also move the definition below to a headerfile alongside the defined above

https://github.com/azerothcore/azerothcore-wotlk/blob/7f7f18e48906959460171e05733baa736a730048/src/server/database/Database/DatabaseWorkerPool.cpp#L42-L48

src/server/database/Database/DatabaseWorkerPool.cpp
```diff
@@ -377,6 +379,52 @@ void DatabaseWorkerPool<T>::KeepAlive()
...
+// Intended to be used with mysql_get_server_info()'s output as the source
+//
+// DatabaseIncompatibleVersion("8.0.35") => false
+// DatabaseIncompatibleVersion("5.6.6") => true
+// DatabaseIncompatibleVersion("5.5.5-10.5.5-MariaDB") => false
+// DatabaseIncompatibleVersion("5.5.5-10.4.0-MariaDB") => true
+//
+// Adapted from link
+// https://stackoverflow.com/a/2941508
+bool DatabaseIncompatibleVersion(std::string const mysqlVersion) {
```
Kitzunu, line 393:
```suggestion
bool DatabaseIncompatibleVersion(std::string const mysqlVersion)
{
```

src/server/database/Database/DatabaseWorkerPool.cpp
michaeldelago: Yeah, that makes sense. I replaced the `MIN_MYSQL_SERVER_VERSION` since the spot I'm editing is the only place it's used

Kitzunu (approved):

PR #18203, opened 2024-01-20
fix(scripts/commands) prevent crash on set motd

MellianStudios:
## Changes Proposed:

This PR proposes changes to:

## SOURCE:

## Tests Performed:
This PR has been:

## How to Test the Changes:
".server set motd something something" currently crashes server because it understands first word as realm id

### DIFF
--- a/src/server/scripts/Commands/cs_server.cpp
+++ b/src/server/scripts/Commands/cs_server.cpp
@@ -514,10 +514,14 @@ public:
     }
 
     // Define the 'Message of the day' for the realm
-    static bool HandleServerSetMotdCommand(ChatHandler* handler, int32 realmId, Tail motd)
+    static bool HandleServerSetMotdCommand(ChatHandler* handler, Optional<int32> realmId, Tail motd)
     {
         std::wstring wMotd   = std::wstring();
         std::string  strMotd = std::string();
+
+        if (!realmId) {
+            realmId = static_cast<int32>(realm.Id.Realm);
+        }
 
         if (motd.empty())
         {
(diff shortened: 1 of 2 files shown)


src/server/scripts/Commands/cs_server.cpp
```diff
@@ -524,6 +524,14 @@ class server_commandscript : public CommandScript
             return false;
         }
 
+        try {
+            std::stoi(realmId);
+        } catch (...) {
+            LOG_ERROR("misc", "Invalid realm ID {} in command .server set motd $realmId $MOTD", realmId);
+
+            return false;
+        }
```
Kitzunu, line 533: I dont like this. I would rather do this
```cpp
static bool HandleServerSetMotdCommand(ChatHandler* handler, int32 realmId, Tail motd)
{
    if (!realmId)
        realmId = -1

Comment (Kitzunu) [reply src/server/scripts/Commands/cs_server.cpp:533]: See that I change realmId from a string to int

File: src/server/scripts/Commands/cs_server.cpp [hunk 2]
```diff
@@ -514,16 +514,11 @@ class server_commandscript : public CommandScript
     }
 
     // Define the 'Message of the day' for the realm
-    static bool HandleServerSetMotdCommand(ChatHandler* handler, std::string realmId, Tail motd)
+    static bool HandleServerSetMotdCommand(ChatHandler* handler, int32 realmId, Tail motd)
     {
         std::wstring wMotd   = std::wstring();
         std::string  strMotd = std::string();
```
Kitzunu, line 521:
```suggestion
    static bool HandleServerSetMotdCommand(ChatHandler* handler, Optional<int32> realmId, Tail motd)
    {
        std::wstring wMotd   = std::wstring();
        std::string  strMotd = std::string();

        if (!realmId)
            realmId = -1;
```

Kitzunu: I don't think it is necessary to force realmId here. If none is provided we update for all realms

Kitzunu (approved):

PR #18222, opened 2024-01-24
fix(Core/Quests): Create quest money reward config and adjust max level bonus reward function.

heyitsbench:
This PR impacts very little. Aside from some quests that have dynamically changing RewardBonusMoney values (e.g. class quests such as [Klaven's Tower](https://wowgaming.altervista.org/aowow/?quest=2359)), this will not affect an AC install with default config values at all, and in those cases it'll only be apparent to those looking at quest query results. For users that have modified their configs (especially those with reduced maximum player levels), this will impact the rewarded gold from quests completed at max level.
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- May resolve https://github.com/chromiecraft/chromiecraft/issues/6534, could not reproduce on current master with outlined test steps below.
- Likely resolves https://github.com/chromiecraft/chromiecraft/issues/6114
- Likely resolves https://github.com/azerothcore/mod-progression-system/issues/352

## SOURCE:

## Tests Performed:

This PR has been:
  - Tested [Banish More Demons](https://wowgaming.altervista.org/aowow/?quest=11051) on a level seventy character with the MaxPlayerLevel set to seventy and the RewardBonusMoney set to 0.5. Rewarded money was 8g19s50c ((4g40s40c * RewardQuestMoney(1)) + ((12,650 * 6 = 7g59s) * RewardBonusMoney(0.5) = 3g79s50c))

## How to Test the Changes:

1. Adjust added config, max level config, as well as existing rate configs (i.e., RewardBonusMoney).
2. Ensure that the following is true:
2a. When completing a quest below max level, the rewarded money is multiplied by RewardQuestMoney.
2b. When completing a quest at max level, the rewarded money is the sum of both the typical rewarded XP (multiplied by RewardQuestMoney) as well as a bonus reward (the product of the XP otherwise gained and six copper).


src/server/game/Quests/QuestDef.cpp
```diff
@@ -252,15 +252,19 @@ int32 Quest::GetRewOrReqMoney(uint8 playerLevel) const
...
 
 uint32 Quest::GetRewMoneyMaxLevel() const
 {
+    uint32 rewMoney = 0;
+
     if (HasFlag(QUEST_FLAGS_NO_MONEY_FROM_XP))
-        return 0;
+        return rewMoney;
+
+    rewMoney = (XPValue(sWorld->getIntConfig(CONFIG_MAX_PLAYER_LEVEL)) * (6 * COPPER));
```
heyitsbench, line 265: This (the copper multiplication) was done for readability, can be removed if need be.

src/server/game/Quests/QuestDef.cpp
```diff
@@ -252,15 +252,19 @@ int32 Quest::GetRewOrReqMoney(uint8 playerLevel) const
...
 
-    return static_cast<int32>(RewardBonusMoney * sWorld->getRate(RATE_REWARD_BONUS_MONEY));
+    return static_cast<int32>(rewMoney * sWorld->getRate(RATE_REWARD_BONUS_MONEY));
```
Kitzunu, line 267: I think this doesn't make sense. `RewardBonusMoney * ConfigRate` should be correct as the Bonus Money comes from quest_template.RewardBonusMoney (RewardBonusMoney var). By changing this you render this DB field useless

src/server/game/Quests/QuestDef.cpp
```diff
@@ -252,15 +252,19 @@ int32 Quest::GetRewOrReqMoney(uint8 playerLevel) const
         }
     }
 
-    return static_cast<int32>(rewardedMoney * sWorld->getRate(RATE_REWARD_BONUS_MONEY));
+    return static_cast<int32>(rewardedMoney * sWorld->getRate(RATE_REWARD_QUEST_MONEY));
```
Kitzunu, line 255: this is good.

src/server/game/Quests/QuestDef.cpp
Kitzunu: So either it is a fixed rate and then RewardBonusMoney field should be deleted.
Or there are some cases where a quest has a different bonus reward, to which changes need to be made to take RewardBonusMoney into account.

Also I would like to know where the 6 comes from

heyitsbench: > By changing this you render this DB field useless

Yes. The RwardBonusMoney field for quest_template entries not only change every time max level moves on Blizz servers, but for some quests (such as Klaven's Tower as mentioned in the PR description), it changes with your player level. It doesn't matter since afaik the reward value doesn't get used until you are at max level, but the field itself isn't fixed on Blizz. I can add a hard-coded check for if the max player level config is 80, and only *then* use the RewardBonusMoney value, but that still wouldn't account for dynamic reward values, plus hard-coding sucks just to use the same value as the calculation yields. Plus, according to [an old version of the XP Wowpedia article](https://wowpedia.fandom.com/wiki/Experience_point?oldid=2148651), depending on the expansion you owned and what your own max level was, the bonus would be different.

Trust me, it hurts to make a brute-forced value useless, but in this case it's wrong to keep using it.

> Also I would like to know where the 6 comes from

The formula for the reward has been known for [a long time](https://wowpedia.fandom.com/wiki/Quest?oldid=1035002), and it has been correct for every single quest I've looked at. Apparently it changed post-Wrath, but luckily we're not dealing with that.

Kitzunu: 1. Okay then we delete that DB column. Otherwise, people will think it still is used. You will also need to delete the reference in the conf.
2. Aight. Can you add a comment with the link to wowpedia so nobody goes and changes it in the future

heyitsbench: > Okay then we delete that DB column. [...] You will also need to delete the reference in the conf. [...] Can you add a comment with the link to wowpedia [...]

Done.

heyitsbench:
Tested several quests with default config values as well, all seems to be working as expected.

elthehablo (approved):

StraysFromPath:
Don't forget to updata wiki documentation.

https://www.azerothcore.org/wiki/quest_template

heyitsbench:
[Yes, I know.](https://github.com/azerothcore/wiki/pull/954)

Issue #17656, opened 2023-11-08
Container ac-db-import   service "ac-db-import" didn't complete successfully: exit 1 


sw586:
### Current Behaviour

![微信截图_20231108090403](https://github.com/azerothcore/azerothcore-wotlk/assets/24788222/a04a8cf8-9c3a-4fc4-96ce-f4db87a16977)
![微信截图_20231108090443](https://github.com/azerothcore/azerothcore-wotlk/assets/24788222/c17b5a4d-4fcc-462c-846d-8426dd633ec0)
![微信截图_20231108090630](https://github.com/azerothcore/azerothcore-wotlk/assets/24788222/c53b4ec7-a3fb-41c1-a756-2af1edb900ce)
Each step is based on the operation: https://www.azerothcore.org/wiki/install-with-docker
Executing: docker compose up -d --build
Error reported later:
 Container ac-db-import         service "ac-db-import" didn't complete successfully: exit 1           

### Expected Blizzlike Behaviour

Docker version 24.0.7, build afdd53b
Docker Compose version v2.21.0
CPU: ARM

### Steps to reproduce the problem

https://www.azerothcore.org/wiki/install-with-docker
git clone https://github.com/azerothcore/azerothcore-wotlk.git --depth 1
cd azerothcore-wotlk
docker compose up -d --build

### AC rev. hash/commit

 Container ac-db-import         service "ac-db-import" didn't complete successfully: exit 1           

### Operating system

Ubuntu 20.04 ARM

Daeymon:
I'm also unable to deploy a fresh build from this repo with `docker compose up -d --build` (no modules, fresh download of repo). Mine ends in:
`service "ac-db-import" didn't completed successfully: exit 126`

I'm still able to successfully deploy from the "acore-docker" repo using `docker compose up -d` ([after incorporating pull request 34](https://github.com/azerothcore/acore-docker/pull/34)).

To get some more detail, I did run it without the "-d" option and it gave me the error as:
`ac-db-import         | /bin/bash: /entrypoint.sh: Permission denied`

Kitzunu:
@michaeldelago

StewartPolsky:
I am having the same issue

michaeldelago:
@Daeymon Can you share the full output when you run `docker compose up ac-db-import`? If it's many lines (more than 75), can you share the relevant error output as well as a few lines above and below?

@sw586 Can you perform a similar process as above? 

[...]

PR #18238, opened 2024-01-29
fix(Docker): Check write access for config dir

michaeldelago:
## Changes Proposed:

This makes it so the startup script (known as the entrypoint) for the docker containers will ensure that it owns the config and filelog directories on startup. 

This is important since there's the `dist` config files that get created during build, and these need to be transferred to the runtime config directory at runtime if they don't exist already.

Technically speaking, we could remove the bind mount (actually comment it out in the `docker-compose.yaml` file) so that the reader can see there's an option to do so. The reasoning for this is to nudge the user in the direction of using environment variables for config rather than files. On the other hand, this breaks compatibility for other users and we don't really need to do that for something so small.

## Issues Addressed:

- https://github.com/azerothcore/azerothcore-wotlk/issues/17656
    - It's common for users to clone the repo as root, which makes the config directories owned to root.
    - This is problematic because the servers for azerothcore run as non-root by default
    - As noted, the startup process creates files in the config dir, and a non-root user can't do that in a directory where they don't have write access
    - All in all this results in a somewhat confusing output on exit. It's a somewhat common thing to support.

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. run command `sudo chown -R 0:0 env/dist` so its all owned to root
2. run `docker compose up --build ac-db-import` to build and run the db import. it should fail, with a big warning message at the top describing the error and possible resolution
3. run `sudo chown -R 1000:1000 env/dist` so it's owned to the default acore user
4. run `docker compose up --build ac-db-import`, it should build and start like normal

### DIFF
--- a/apps/docker/entrypoint.sh
+++ b/apps/docker/entrypoint.sh
@@ -1,14 +1,46 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+CONF_DIR="${CONF_DIR:-/azerothcore/env/dist/etc}"
+LOGS_DIR="${LOGS_DIR:-/azerothcore/env/dist/logs}"
+
+if ! touch "$CONF_DIR/.write-test" || ! touch "$LOGS_DIR/.write-test"; then
+    cat <<EOF
+===== WARNING =====
+The current user doesn't have write permissions for
+the configuration dir ($CONF_DIR) or logs dir ($LOGS_DIR).
+It's likely that services will fail due to this.
+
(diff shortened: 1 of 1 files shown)


walkline (approved):

sudlud:
@michaeldelago what is the state of this PR? :)

michaeldelago:
@sudlud This is good to merge, pending review.

sudlud (approved):

PR #18243, opened 2024-01-31
feat(Core/Unit): New helper IsClass and script hook OnPlayerIsClass

NathanHandley:
## Changes Proposed:
This PR proposes changes to:

Added a new hook "OnPlayerIsClass" inside a new helper interception function "Player::IsClass" (which has a virtual base Unit::IsClass) and replaced most comparison uses of getClass().

This hook allows script implementation of many previously class restricted behaviors (some requiring dbc changes).  For some examples that can be implemented for all or select classes:
- Roll on and use armor of different classes (i.e. plate on a Mage, librams on a priest)
- Use class-restricted Taxis (such as the Druid one from Moonglade)
- Use class abilities on different classes (pick pocket on non-rogue, rune abilities on a warrior)
- Apply stat modifiers for feral forms on non-druids
- Allow enabling of class-restricted quests (such as the Sea Lion quest line)
- Enable class trainers on different classes
- Allow completely new classes composed of other class unique capabilities. 
- And lots more!  As you can make "any explicit behavior restricted by class A can be done via script with class B"

This is one major steps towards solving the server core custom code fragmentation across code bases where custom classes are implemented, and does so without fundamentally changing the logic built up over the years in this code base.  Two other hurdles remain: pets and powers.  Pets still needs a PR, but a PR has been opened to address powers separately [here](https://github.com/azerothcore/azerothcore-wotlk/pull/18293).

## Issues Addressed:
none

## SOURCE:
n/a

## Tests Performed:
This PR has been:

## How to Test the Changes:
To test the change impact to core/stock: just run the core and perform basic behaviors.  More obvious behaviors to look at would be ensuring that stats look like they should on each class, and that abilities like overpower and pick pocket operate as expected.

To test this change as a hooked script, implement the PlayerScript hook OnPlayerIsClass.  An example of this implementation can be found in this mod branch I'm working on here: https://github.com/NathanHandley/mod-multi-class/blob/EternalWrath/src/MultiClass_PlayerScript.cpp (line 37). 

**Reviewers:** This PR was built to be as focused as possible so while it does touch a lot of files, it's almost only straight getClass() to IsClass comparator swaps.  In a few spots the switch statements had to get broken into if trees in order for the hook to be useful.  The only true code tree 'restructures' was in zone_moonglade.cpp around the two druid flight NPCs so that quest and taxi contexts could be attributed correctly.

These are not issues, but rather TODO items outstanding that will come in later PRs:
- This doesn't fully avail cross-class pets via a mod.  It will take some logic changes and one or two new hooks to address that, which will come about in a future PR.
- This doesn't address all cross-class powers (i.e. rage on a Mage).  For that, this PR is required: https://github.com/azerothcore/azerothcore-wotlk/pull/18293


src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -1038,6 +1038,19 @@ bool ScriptMgr::CanRepopAtGraveyard(Player* player)
...
 }
 
+Optional<bool> ScriptMgr::IsClass(Player const* player, Classes unitClass, ClassContext context)
+{
+    if (ScriptRegistry<PlayerScript>::ScriptPointerList.empty())
+        return {};
+    for (auto const& [scriptID, script] : ScriptRegistry<PlayerScript>::ScriptPointerList)
+    {
+        Optional<bool> scriptResult = script->IsClass(player, unitClass, context);
+        if (scriptResult != std::nullopt)
```
Winfidonarleyan, line 1048:
```suggestion
        if (scriptResult)
```

NathanHandley: Changed.

NathanHandley:
Updated title to show that it's a helper as well as a script hook

Nyeriah (approved):

PR #18278, opened 2024-02-06
feat(Handlers/LootHandler.cpp) Moving invocation of the OnAfterCreatu…

pangolp:
## Changes Proposed:

This PR proposes changes to:
- [x] Hook

## Issues Addressed:

- Closes

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. It does not require many tests, because the only thing you do is move the place where it is invoked. It is a hook that is used in the area loot module, which is poorly located, or rather, as a result of the new changes, it needs to be called in another location.


heyitsbench:
This makes no sense, you're moving a hook next to another hook.

pangolp:
I didn't say anything, cancel the pull request, use the hook that already exists, and change the code in the loot aoe, don't worry.

PR #18295, opened 2024-02-09
feat(fix/build) mysql_stmt_bind_param deprecated in MYSQL 8.3

pangolp:
## Changes Proposed:

This PR proposes changes to:

- Currently, the library is no longer supported on macOS, so the tests were failing. That way, we can continue to preserve both methods, and for now support all operating systems, since now the tests also work on macOS.

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

- Apparently the library in macOS is outdated. Which causes the tests, when doing a merge, to fail in said operating system.
- I don't know if there are other bugs, I hope the tests can be corrected, and in that way, know if there are more of them.

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -228,7 +228,7 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
     uint32 _s = getMSTime();
 
 #ifdef __APPLE__
-    if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND))
+    if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND, m_mStmt->GetNumParameters(), nullptr))
 #else
     if (mysql_stmt_bind_param(msql_STMT, msql_BIND))
 #endif
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -227,7 +227,7 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
 
     uint32 _s = getMSTime();
 
-#ifdef __APPLE__
+#if MYSQL_VERSION_ID >= 80300
     if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND, m_mStmt->GetParameterCount(), nullptr))
 #else
     if (mysql_stmt_bind_param(msql_STMT, msql_BIND))
@@ -279,7 +279,7 @@ bool MySQLConnection::_Query(PreparedStatementBase* stmt, MySQLPreparedStatement
 
     uint32 _s = getMSTime();
 
-#ifdef __APPLE__
+#if MYSQL_VERSION_ID >= 80300
     if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND, m_mStmt->GetParameterCount(), nullptr))
 #else
     if (mysql_stmt_bind_param(msql_STMT, msql_BIND))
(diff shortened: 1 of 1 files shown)


src/server/database/Database/MySQLConnection.cpp
```diff
@@ -227,7 +227,11 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
 
     uint32 _s = getMSTime();
 
+#ifdef __APPLE__
```
Kitzunu, line 230: no need for this. just use mysql_stmt_bind_named_param().

mysql_stmt_bind_param() is deprecated and will be removed in future versions of MySQL

Kitzunu (changes requested):

src/server/database/Database/MySQLConnection.cpp
pangolp: The reason for the pull request was not to replace the method, but to solve the test problem in macOS. In Windows, I do not have access to that method, and if I change it, then I am obliged to update the library, which I will do at some point, but I prefer to leave it that way for now. Once we are sure that we cannot support previous versions (which still work), we can make that change that you say. But for now, the goal is to get it to pass testing on macOS only, not to upgrade MySQL.

pangolp: For the moment, I prefer to leave that condition and solve only the testing problem. I think that later, as you say, we can have that method used directly, but at least I, for example in Windows, do not have that method. I'm not sure what version of MySQL I'm using, but the one I have at the moment gives me an error, it says it can't find it. I think it is 8.0 and that method was added later. Better let's do things calmly, and in this particular case, let's fix the compilation problem on macOS, then we check the rest.

pangolp: I think mysql_stmt_bind_named_param is currently in 8.3, but in 8.0, for example, it is not. What happens to operating systems or distributions that cannot access that function? We cannot leave them without support. I prefer, therefore, to leave it currently as it is, and then when things become a little more standardized, then if I make the definitive change. Unless you, who are more experienced than me, have more information or think I'm wrong. Please, if so, let me know, and please provide me with information so I can read about it. Thank you. @Kitzunu

Kitzunu: Okay but what if someone builds Mac with MySQL 8.0.

Add a MySQL version check instead of a mac check :)

pangolp: > Okay but what if someone builds Mac with MySQL 8.0.
> 
> Add a MySQL version check instead of a mac check :)

I like this alternative better.

pangolp: I'm not sure, but I think that this method is added in 8.3, so if you have 8.3 or higher, use the new one and if not, use the old one.

Kitzunu: yea so something like

#if MYSQL_VER >= 803000u

or whatever the correct variable would be

pangolp: Thanks.

Kitzunu (approved):

0xCiBeR:
This PR broke builds on Ubuntu 22.04 & MariaDB, see #18307

pangolp:
The problem is neither Ubuntu nor Debian, the problem is that MariaDB still uses some constants after it was separated, but this topic is already being addressed in another pull request. https://github.com/azerothcore/azerothcore-wotlk/pull/18310

Issue #18307, opened 2024-02-11
build error on raspberry pi 4b after #18295


sandbox404:
### Current Behaviour

[ 15%] Building CXX object src/server/database/CMakeFiles/database.dir/Database/MySQLConnection.cpp.o
/home/pi/azerothcore/src/server/database/Database/MySQLConnection.cpp:231:9: fatal error: use of undeclared identifier 'mysql_stmt_bind_named_param'
    if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND, m_mStmt->GetParameterCount(), nullptr))
        ^
[ 15%] Linking CXX executable map_extractor
1 error generated.
make[2]: *** [src/server/database/CMakeFiles/database.dir/build.make:189: src/server/database/CMakeFiles/database.dir/Database/MySQLConnection.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1052: src/server/database/CMakeFiles/database.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 15%] Built target map_extractor
[ 15%] Linking CXX executable mmaps_generator
[ 15%] Built target mmaps_generator
make: *** [Makefile:136: all] Error 2

### Expected Blizzlike Behaviour

build success

### Steps to reproduce the problem

1. sudo apt update && sudo apt-get install git cmake clang default-libmysqlclient-dev libssl-dev libbz2-dev libreadline-dev libncurses-dev mariadb-server libboost-all-dev
2. git clone https://github.com/azerothcore/azerothcore-wotlk.git --branch master --single-branch azerothcore
3. mkdir azerothcore/build
4. cd azerothcore/build
5. cmake ../ -DCMAKE_INSTALL_PREFIX=$HOME/azeroth-server/ -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DWITH_WARNINGS=1 -DTOOLS_BUILD=all -DSCRIPTS=static -DMODULES=static -DCMAKE_BUILD_TYPE=Release
6. make -j3

### AC rev. hash/commit

6b1dd7e19ef52c2760d65849a31a29af70ad9795

### Operating system

Debian 12 arm64

0xCiBeR:
Same error on latest pull 1d8c4056b3b8671fef2e356d940cef493de5f6d5 for ubuntu 22.04

pangolp:
[...]

PR #18310, opened 2024-02-11
feat(fix/build) ignore mariadb in the conditional

pangolp:
## Changes Proposed:

This PR proposes changes to:

- In the latest versions of MySQL, a method was changed, which gave errors in macOS. At first, the verification was going to be on that operating system, but in the end, it was done on the MySQL version, to try to continue supporting previous versions. However, it omitted the fact that MariaDB also continues to use this constant for the version and that was proving errors in some operating systems, such as Debian 12.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18307

## SOURCE:

## Tests Performed:

This PR has been:
- [x] Debian 12
- [x] Windows 10

## How to Test the Changes:

1. Download the changes and compile.
2. The only thing that does the pull request is this and previous, and check the version of MySQL to know what method you should use.

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -227,7 +227,7 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
 
     uint32 _s = getMSTime();
 
-#if !defined(MARIADB_VERSION_ID) && MYSQL_VERSION_ID >= 80300
+#if !defined(MARIADB_VERSION_ID) && (MYSQL_VERSION_ID >= 80300)
     if (mysql_stmt_bind_named_param(msql_STMT, msql_BIND, m_mStmt->GetParameterCount(), nullptr))
 #else
     if (mysql_stmt_bind_param(msql_STMT, msql_BIND))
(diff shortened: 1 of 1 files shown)


src/server/database/Database/MySQLConnection.cpp
```diff
@@ -227,7 +227,7 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
 
     uint32 _s = getMSTime();
 
-#if MYSQL_VERSION_ID >= 80300
+#if !defined(MARIADB_VERSION_ID) && MYSQL_VERSION_ID >= 80300
```
sudlud, line 230: Do you want to add (...) to the second condition, too? 🙂

pangolp: > Do you want to add parantheses (...) to the second condition? 🙂

Of course.

pangolp:
```c++
#if (!defined(MARIADB_VERSION_ID) && (MYSQL_VERSION_ID >= 80300))
```
Shouldn't it be this way?

elthehablo (approved):

PR #17596, opened 2023-10-27
Fix (Core) Resolve issue with dialog options randomly not working

kjack9:
## Changes Proposed:

This PR proposes changes to:
-  [X] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

https://github.com/azerothcore/azerothcore-wotlk/issues/15989#issuecomment-1782229624

## Issues Addressed:

- Closes #15989
- Closes https://github.com/chromiecraft/chromiecraft/issues/3997

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.
- [X] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1. `.go creature 116913`
2. Moving left to right breaks the dialog before this change

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #18331, opened 2024-02-15
fix(Core/GameObject): Remove unneeded ClearGossipMenuFor call

r4d1sh:
## Changes Proposed:

This PR essentially repeats this - #17596
But for GameObjects

Removed another unnecessary ClearGossipMenuFor call which randomly broke gossip menus.

For example, if you are talking with some NPC, and at that moment you receive/lose some item or gold, your client will send CMSG_QUESTGIVER_STATUS_MULTIPLE_QUERY. If at the same time there are some GameObjects near you that can be questgiver, then a chain of method calls will lead to ClearGossipMenuFor and your gossip menu with NPC will suddenly breaks. You won't even understand why this could happen.

This PR proposes changes to:

## Tests Performed:

This PR has been:

### DIFF
--- a/src/server/game/Scripting/ScriptDefines/GameObjectScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/GameObjectScript.cpp
@@ -128,7 +128,6 @@ uint32 ScriptMgr::GetDialogStatus(Player* player, GameObject* go)
     ASSERT(go);
 
     auto tempScript = ScriptRegistry<GameObjectScript>::GetScriptById(go->GetScriptId());
-    ClearGossipMenuFor(player);
     return tempScript ? tempScript->GetDialogStatus(player, go) : DIALOG_STATUS_SCRIPTED_NO_STATUS;
 }
 


kissingers:
I find when  LUA use GossipMenu , if somebody change the money or when in group and changed subgroup, the Menu will not continue to be clicked. Is this can fix the problem?

r4d1sh:
> I find when LUA use GossipMenu , if somebody change the money or when in group and changed subgroup, the Menu will not continue to be clicked. Is this can fix the problem?

Yes. Change subgroup / money amount cause the client to send CMSG_QUESTGIVER_STATUS_MULTIPLE_QUERY and this problem happens.

Theoretically, this could be a problem in your Lua script. But I am 95% sure that this is exactly the problem that I described, and this PR will fix it. You can check.

elthehablo (approved):

kissingers:
Good fix, all LUA GossipMenu looks fine when player change money or subgroup!

PR #18390, opened 2024-02-24
fix(Core/Spells): Fixed work of sobering spells and other improvements for drunk system

r4d1sh:
The first commit related to `SPELL_EFFECT_INEBRIATE` (100) effect. The problem is that there are not only inebriation but also sobering spells (with basepoint < 0). The current implementation uses an unsigned type, ignoring sobering spells. This leads to a uint8 underflow, resulting in the spell's behavior being completely unexpected: it can both inebriate and sober you up.

Example:

1. `.gobject add temp 185914`
2. Use it!
3. If you were completely sober before, you will now get maximum inebriation. Although coffee was supposed to sober you up by 20%. This happened because of uint8 underflow.

There is another problem: there are spells with an inebriate effect with basepoint values so large or small that it inevitably leads to uint8 overflow/underflow. For example, `29690` or `46874`. As intended, `29690` should make you extremely drunk, but in reality, the behavior can be unpredictable each time. If someone keeps casting this spell on you again and again, you will likely notice that sometimes you get drunk and sometimes you sober up.

The first commit completely fixes this issues.

The second commit is related to the aura effect `SPELL_AURA_MOD_FAKE_INEBRIATE` (304). The main problem is that the handler of this effect ignores the value of player's actual inebriation. In any case, there was a lot of redundant (duplicate) code here. Now the logic for updating invisibility detect has been moved to a separate method, and all known issues have been fixed.

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -4474,17 +4474,25 @@ void Spell::EffectInebriate(SpellEffIndex /*effIndex*/)
     }
 
     uint8 currentDrunk = player->GetDrunkValue();
-    uint8 drunkMod = damage;
-    if (currentDrunk + drunkMod > 100)
-    {
+    int32 drunkMod = damage;
+
+    if (drunkMod == 0)
+        return;
+
+    // drunkMod may contain values ​​that are guaranteed to cause uint8 overflow/underflow (examples: 29690, 46874)
+    // In addition, we would not want currentDrunk to become more than 100.
+    // So before adding the values, let's check that everything is fine.
+    if (drunkMod > 0 && drunkMod > static_cast<int32>(100 - currentDrunk))
         currentDrunk = 100;
-        if (rand_chance() < 25.0f)
-            player->CastSpell(player, 67468, false);    // Drunken Vomit
-    }
+    else if (drunkMod < 0 && drunkMod < static_cast<int32>(0 - currentDrunk))
+        currentDrunk = 0;
     else
-        currentDrunk += drunkMod;
+        currentDrunk += drunkMod; // Due to previous checks we can be sure that currentDrunk will not go beyond [0-100] range.
 
     player->SetDrunkValue(currentDrunk, m_CastItem ? m_CastItem->GetEntry() : 0);
+
+    if (currentDrunk == 100 && roll_chance_i(25))
+        player->CastSpell(player, 67468, false);    // Drunken Vomit
 }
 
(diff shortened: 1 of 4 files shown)


src/server/game/Spells/SpellEffects.cpp
```diff
@@ -4474,17 +4474,25 @@ void Spell::EffectInebriate(SpellEffIndex /*effIndex*/)
...
-    {
+    int32 drunkMod = damage;
+
+    if (drunkMod == 0)
+        return;
+
+    // drunkMod may contain values ​​that are guaranteed to cause uint8 overflow/underflow (examples: 29690, 46874)
+    // In addition, we would not want currentDrunk to become more than 100.
+    // So before adding the values, let's check that everything is fine.
+    if (drunkMod > 0 && drunkMod > static_cast<int32>(100 - currentDrunk))
```
sudlud, line 4485: I would prefer braces (...) around the conditions for readability here, also on the following if (...) conditions.
Would you agree? 🙂

sudlud:
Thanks for the PR! Just did a quick test.

**before PR**
`.gobject add temp 185914`
drinking the coffee
makes you drunk and vomit

spell "Sober Up - DEBUG"
`.cast 46874`
makes you drunk and vomit

I can't really test spell `29690` somehow, always says "bad target"

---

**after PR**
`.gobject add temp 185914`
be sober
drinking the coffee
nothing happens

drink cheep beer
`.additem 19222 42`
slowly increases drunkiness until vomiting

drinking coffee from `185914` slowly makes you more sober again with each coffee

spell "Sober Up - DEBUG"
`.cast 46874`
instantly makes you sober again

---

So everything looks good so far 🙂

Can you please also provide steps to test the changes to `SPELL_AURA_MOD_FAKE_INEBRIATE`?

src/server/game/Spells/SpellEffects.cpp
r4d1sh: > I would prefer braces (...) around the conditions for readability here, also on the following if (...) conditions. Would you agree? 🙂

I don't think that extra braces will improve code readability in this case. I prefer to explicitly use optional braces just to avoid confusion with priority of `&&` and `||` operators. But that's a different case.

sudlud: Alright, I'd still prefer them but it's not my call as I'm not maintainer.

sudlud (approved):

sudlud:
Tested, looks good to me 🙃

Nyeriah (approved):

PR #18519, opened 2024-03-10
chore. Avoid failures in the tests of custom items

pangolp:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## How to Test the Changes:

1. We have to ignore the ids of the items in the dbc that are not blizzlike, because if not, the tests in the modules fail, since the validation is executed, it does not find the item in the original dbc, and generates the error, causing do not continue the test.

```
Run source ./apps/ci/ci-error-check.sh
Checking Startup Errors

The Errors.log file contains startup errors:

Item (Entry: [7](https://github.com/azerothcore/mod-item-level-up/actions/runs/8221525862/job/22481986264#step:11:8)0[10](https://github.com/azerothcore/mod-item-level-up/actions/runs/8221525862/job/22481986264#step:11:11)01) does not exist in item.dbc! (not correct id?).

Please solve the startup errors listed above!
Error: Process completed with exit code 1.
 ```

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -2873,7 +2873,7 @@ void ObjectMgr::LoadItemTemplates()
 
         if (!dbcitem)
         {
-            LOG_ERROR("sql.sql", "Item (Entry: {}) does not exist in item.dbc! (not correct id?).", entry);
+            LOG_DEBUG("sql.sql", "Item (Entry: {}) does not exist in item.dbc! (not correct id?).", entry);
             continue;
         }
 
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -2871,6 +2871,9 @@ void ObjectMgr::LoadItemTemplates()
         // Checks
         ItemEntry const* dbcitem = sItemStore.LookupEntry(entry);
 
+        if (!dbcitem)
+            continue;
+
         if (dbcitem)
         {
             if (enforceDBCAttributes)
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -2872,51 +2872,49 @@ void ObjectMgr::LoadItemTemplates()
         ItemEntry const* dbcitem = sItemStore.LookupEntry(entry);
 
         if (!dbcitem)
-            continue;
-
-        if (dbcitem)
         {
-            if (enforceDBCAttributes)
+            LOG_ERROR("sql.sql", "Item (Entry: {}) does not exist in item.dbc! (not correct id?).", entry);
+            continue;
+        }
+
+        if (enforceDBCAttributes)
+        {
+            if (itemTemplate.Class != dbcitem->ClassID)
             {
-                if (itemTemplate.Class != dbcitem->ClassID)
-                {
-                    LOG_ERROR("sql.sql", "Item (Entry: {}) has wrong Class value ({}), must be ({}).", entry, itemTemplate.Class, dbcitem->ClassID);
-                    itemTemplate.Class = dbcitem->ClassID;
-                }
-                if (itemTemplate.SubClass != dbcitem->SubclassID)
-                {
-                    LOG_ERROR("sql.sql", "Item (Entry: {}) has wrong Subclass value ({}) for class {}, must be ({}).", entry, itemTemplate.SubClass, itemTemplate.Class, dbcitem->SubclassID);
-                    itemTemplate.SubClass = dbcitem->SubclassID;
-                }
-                if (itemTemplate.SoundOverrideSubclass != dbcitem->SoundOverrideSubclassID)
-                {
-                    LOG_ERROR("sql.sql", "Item (Entry: {}) does not have a correct SoundOverrideSubclass ({}), must be {}.", entry, itemTemplate.SoundOverrideSubclass, dbcitem->SoundOverrideSubclassID);
-                    itemTemplate.SoundOverrideSubclass = dbcitem->SoundOverrideSubclassID;
-                }
-                if (itemTemplate.Material != dbcitem->Material)
-                {
-                    LOG_ERROR("sql.sql", "Item (Entry: {}) does not have a correct material ({}), must be {}.", entry, itemTemplate.Material, dbcitem->Material);
-                    itemTemplate.Material = dbcitem->Material;
-                }
-                if (itemTemplate.InventoryType != dbcitem->InventoryType)
-                {
-                    LOG_ERROR("sql.sql", "Item (Entry: {}) has wrong InventoryType value ({}), must be {}.", entry, itemTemplate.InventoryType, dbcitem->InventoryType);
-                    itemTemplate.InventoryType = dbcitem->InventoryType;
-                }
-                if (itemTemplate.DisplayInfoID != dbcitem->DisplayInfoID)
-                {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pangolp** [comment]
https://github.com/azerothcore/mod-item-level-up/pull/8

**Comment by heyitsbench** [comment]
Maybe a debug log message in such a case?

**Comment by pangolp** [comment]
> Maybe a debug log message in such a case?

I don't understand what you're trying to tell me. What we are trying to avoid is that the tests are executed on the entries of the ids of the custom items, because they will never be found within the dbc, because they do not exist for the emulator.

But maybe I misunderstood you, so if you can give me more detail, maybe I can respond.

**Comment by heyitsbench** [comment]
I'm thinking replace the log on L2916/2919 with a LOG_DEBUG instead of LOG_ERROR to say it's not enforcing attributes on the item, that way there's still communication to the user without closing the process.

**Comment by pangolp** [comment]
> I'm thinking replace the log on L2916/2919 with a LOG_DEBUG instead of LOG_ERROR to say it's not enforcing attributes on the item, that way there's still communication to the user without closing the process.

Well, I understand. We can make that change if necessary too.

**Comment by pangolp** [comment]
> I'm thinking replace the log on L2916/2919 with a LOG_DEBUG instead of LOG_ERROR to say it's not enforcing attributes on the item, that way there's still communication to the user without closing the process.

I think that the issue of `LOG_ERROR` is intended specifically so that if it is generated, it will not pass the tests. That's fine, for blizzlike items. But when you have a module that adds a custom item, it will never find it in the item.dbc, so we have to try to prevent the error message from appearing. Initially, I was going to make the ID tomorrow or equal to the last largest, but there are many intermediate numbers within the table, which also do not exist and would clearly have the same problem.

```sql
SELECT t.`entry` + 1
FROM `item_template` t
WHERE NOT EXISTS (
    SELECT * 
    FROM `item_template` t2
    WHERE t2.`entry` = t.`entry` + 1
)
LIMIT 1
```

For example, 160. And if someone used that id, it would also possibly give them an error, because the item.dbc with id 160 does not exist either.

```
 entry  name                                
------  ------------------------------------
   150  Deprecated Tauren Apprentice Pants  
   151  Deprecated Tauren Apprentice Shirt  
   152  OLDPrimitive Leather Belt           
   153  Primitive Kilt                      
   154  Primitive Mantle                    
   155  Deprecated Tauren Recruit's Belt    
   156  Deprecated Tauren Recruit's Pants   
   157  Deprecated Tauren Recruit's Shirt   
   159  Refreshing Spring Water             
```

```sql
SELECT `entry`, `name` FROM `item_template` WHERE `entry` BETWEEN 150 AND 170;
```

heyitsbench:
Well if this PR is merged as it is, then the existing LOG_ERROR will never be hit, since it's explicitly targeting the situation you're talking about; an item template not in DBC, so it should either A) be removed, or B) be replaced with a LOG_DEBUG.

pangolp:
> Well if this PR is merged as it is, then the existing LOG_ERROR will never be hit, since it's explicitly targeting the situation you're talking about; an item template not in DBC, so it should either A) be removed, or B) be replaced with a LOG_DEBUG.

It seems to me that the LOG_DEBUG would be incorrect. Maybe I failed in the logic I wanted to apply, but the error has to be generated, as long as we are talking about a blizzlike item. We have to ignore custom items. But it may be that I have failed in the logic I applied. I had several ideas in mind, until this one uploaded. But if it can be changed, it can be changed.

pangolp:
I don't think anyone creates an item above the latest blizzlike. The available free entries are not used. It is always done starting from the last one. So perhaps, my first logic of ignoring those that are superior to the last blizzlike, could be correct.

```c++
if (dbcitem)
    ....
else if (entry <= 56806)
    LOG_ERROR("sql.sql", "Item (Entry: {}) does not exist in item.dbc! (not correct id?).", entry);
```

heyitsbench:
My grievance right now is this:
```cpp
if (!dbcitem) // not in DBC
    continue;

if (dbcitem) // the only other option
{
    // do stuff
}
else // this will never be taken since no dbcitem has already been evaluated and continued out of
    LOG_ERROR // so this log error will never be hit
```
I'm personally not a fan of magic, arbitrary numbers, but that's not my call.

pangolp:
> ```c++
> if (!dbcitem) // not in DBC
>     continue;
> 
> if (dbcitem) // the only other option
> {
>     // do stuff
> }
> else // this will never be taken since no dbcitem has already been evaluated and continued out of
>     LOG_ERROR // so this log error will never be hit
> ```

It happens that the modules run the same tests as the core. Therefore, we have to avoid those messages, because if not, the modules will never pass the tests. If not, we couldn't have modules that add new items. We would have to make the items that already exist and edit them in any case. But anyway, it would give an error, because when you compare the values, it may not pass the tests either if the class or some other value is modified. Unless you take an element exactly the same as the original. But if it is custom, it is okay to add a new value.

Anyway, I thought of the logic very quickly, I may have made a mistake.
Therefore, if I have to make changes, I do them, no problem.

pangolp:
> Does this mean the core refuses to load if we add new items in the item table ?

The emulator does not. The tests. The emulator will let you, but since an Errors.log will be generated, it will not pass the tests, because the item does not exist in a blizzlike way, it is custom.

pangolp:
image

By using this change, errors are no longer generated in the file, and the tests run correctly.
But, I don't know if it's the best way to do it.
I just know that it works. If there is another one, I modify it.

Elmsroth:
@pangolp I think the error is uesefull so it will force you to update the Item.dbc with your entry to be correct and not to forget to distribute yout MPQ patch to your players.

heyitsbench:
> We have to ignore the personalized ones.

Why? The correct way to add a custom item is through both template and DBC.

pangolp:
> > We have to ignore the personalized ones.
> 
> Why? The correct way to add a custom item is through both template and DBC.

Github tests will be carried out on a blizzlike emulator, not a custom one. They are tests designed for the emulator, for the modules.

pangolp:
Maybe I'm wrong but we don't have to stop people from adding new content. What we have to avoid is that if they edit any blizzlike component, the tests notify us, so we can review them manually and discuss them. Control the dbc that are not blizzlike, they generate errors like this one that the module gives because it wants to add an element that does not exist.

pangolp:
Am I clear with my argument or is there something that may not be understood? What I want to avoid with this change is that an error occurs due to a custom id, because if not, all the modules that we have with custom items will fail, since it will search for the item in the DBC and will not find it. find. Now, different, is if I grab item 12345, and edit it. This is where we need the evidence, because that item does exist in the blizzlike DBC. If someone edits something of it, the error must appear, so that we can see why they decided to modify it. But it doesn't make sense that it fails with custom items. The tests have to be done for blizzlike items. If I am wrong, tell me and we will look for another way, but we are going to have problems with all the items that are added and we would have to approve the pull requests of the modules, without the need for them to pass the tests.

Nyeriah:
Put the log message inside the IF and make it a INFO logger or DEBUG

pangolp:
> Put the log message inside the IF and make it a INFO logger or DEBUG

I don't understand what he wants to tell me, and the translator doesn't help me.

pangolp:
I'm making some changes, which I'm going to upload now, but basically, if I understand correctly, the problem, the issue would be, do we have to report the error, but continue anyway? Sorry if I misunderstood you, but between what @Nyeriah mentioned and some other examples I saw of other .dbc checks, they made me think that.

pangolp:
I hope I interpreted what they told me correctly and if not, well, we'll keep trying.

Nyeriah (approved):

pangolp:
Thanks @Nyeriah

PR #18579, opened 2024-03-18
chore. Add hook OnSetArenaMinPlayersPerTeam

pangolp:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author. (I'm still doing tests)

## How to Test the Changes:

1. The idea of the change is to be able to add the minimum number of players so that the arena begins. It is intended for the `mod-arena-3v3-solo-queue` module but other modules may also use it at some point if they need it.

### DIFF
--- a/src/server/game/Battlegrounds/BattlegroundMgr.cpp
+++ b/src/server/game/Battlegrounds/BattlegroundMgr.cpp
@@ -404,8 +404,11 @@ Battleground* BattlegroundMgr::CreateNewBattleground(BattlegroundTypeId original
     if (bg->isArena())
     {
         uint32 maxPlayersPerTeam = ArenaTeam::GetReqPlayersForType(arenaType) / 2;
+        uint32 minPlayersFerTeam = ArenaTeam::GetReqPlayersForType(arenaType);
         sScriptMgr->OnSetArenaMaxPlayersPerTeam(arenaType, maxPlayersPerTeam);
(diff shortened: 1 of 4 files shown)


Elmsroth:
Nice feature @pangolp <3

pangolp:
> Nice feature @pangolp <3

We have to test, if it does not affect the current scoring system. Because I created it for the [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue) module thinking that it could be used, but I don't know if I really need it. However, since we have `OnSetArenaMaxPlayersPerTeam`, it doesn't hurt that we also have `OnSetArenaMinPlayersPerTeam`. But you have to make sure that this modification does not affect the default behavior of the arenas.

sudlud:
I think this PR needs an update now to comply with the latest updates to the scripitng system.

similar to how it is done in #18760.

pangolp:
> I think this PR needs an update now to comply with the latest updates to the scripting system.
> 
> similar to how it is done in #18760.

Possibly, in the event of a merger, changes would have to be made. The issue is to know if it is really worth to merge it, because this hook was created for the 3v3 module, but I think that in the end it does not need it. If the module does not require it, and there is no other module that uses it, most likely, it will end up closing this pull request.

PR #18628, opened 2024-03-29
feat(Core/Optimization): Reduce CPU consumption by allocating sendBuffer only when it is needed.

walkline:
We have a network thread that makes approximately 1k invocations per second of this function for every socket:
https://github.com/azerothcore/azerothcore-wotlk/blob/73340b94e3dbcc67da8e52ea5ddea06298c2c6d4/src/server/game/Server/WorldSocket.cpp#L161-L166
 
Most of the time, `_bufferQueue` remains empty, and we should do nothing inside this function. 
However, for each call, we allocate a 4KB buffer even when it is unnecessary. 
As a result, a significant portion of CPU time is dedicated to the allocation and deallocation of this buffer.

Below, you can observe the reduction in CPU consumption (with one player online) after implementing this change.
Before:
Screenshot 2024-03-29 at 16 23 54
After:
Screenshot 2024-03-29 at 16 30 58

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Server/WorldSocket.cpp
+++ b/src/server/game/Server/WorldSocket.cpp
@@ -161,42 +161,45 @@ void WorldSocket::CheckIpCallback(PreparedQueryResult result)
 bool WorldSocket::Update()
 {
     EncryptableAndCompressiblePacket* queued;
-    MessageBuffer buffer(_sendBufferSize);
-    while (_bufferQueue.Dequeue(queued))
+    if (_bufferQueue.Dequeue(queued)) 
     {
-        queued->CompressIfNeeded();
-
-        ServerPktHeader header(queued->size() + 2, queued->GetOpcode());
-        if (queued->NeedsEncryption())
-            _authCrypt.EncryptSend(header.header, header.getHeaderLength());
-
-        if (buffer.GetRemainingSpace() < queued->size() + header.getHeaderLength())
+        // Allocate buffer only when it's needed but not on every Update() call.
+        MessageBuffer buffer(_sendBufferSize);
+        do 
         {
+            queued->CompressIfNeeded();
+            ServerPktHeader header(queued->size() + 2, queued->GetOpcode());
+            if (queued->NeedsEncryption())
+                _authCrypt.EncryptSend(header.header, header.getHeaderLength());
+
+            if (buffer.GetRemainingSpace() < queued->size() + header.getHeaderLength())
+            {
+                QueuePacket(std::move(buffer));
+                buffer.Resize(_sendBufferSize);
+            }
+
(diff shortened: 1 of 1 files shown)


elthehablo (approved):

kissingers:
I don't know how to write with C++, but the logic is understand. 
I'm wondering if it's possible to just keep a 4k buff for each session when it's initialised, and write it in every time it's refreshed when it's needed, instead of determining if it needs to be reallocated every time? This would further reduce CPU and make memory usage more stable.

walkline:
I'm not a C++ expert either, and feel free to correct me, but I think with the suggested approach, there's a chance that the buffer would be allocated on the heap. However, with the current implementation, I believe it should be allocated on the stack. And I would prefer stack over heap in this case.

kissingers:
> I'm not a C++ expert either, and feel free to correct me, but I think with the suggested approach, there's a chance that the buffer would be allocated on the heap. However, with the current implementation, I believe it should be allocated on the stack. And I would prefer stack over heap in this case.

I can probably know what you mean, if use heap, means that the probability not so easy hits CPU cache, and frequent transfers in and out of memory to the cache may slower?

src/server/game/Server/WorldSocket.cpp
```diff
@@ -161,43 +161,46 @@ void WorldSocket::CheckIpCallback(PreparedQueryResult result)
...
+            ServerPktHeader header(queued->size() + 2, queued->GetOpcode());
+            if (queued->NeedsEncryption())
+                _authCrypt.EncryptSend(header.header, header.getHeaderLength());
 
-        ServerPktHeader header(queued->size() + 2, queued->GetOpcode());
-        if (queued->NeedsEncryption())
-            _authCrypt.EncryptSend(header.header, header.getHeaderLength());
+            if (buffer.GetRemainingSpace() < queued->size() + header.getHeaderLength())
+            {
+                QueuePacket(std::move(buffer));
```
blah900, line 177: First time reviewing code here but out of curiosity, is there a need to queue an empty buffer?

IIUC right now the buffer doesn't have any information and we're queueing an empty packet?

doowahree, line 177: First time reviewing code here but out of curiosity, is there a need to queue an empty buffer?

IIUC right now the buffer doesn't have any information and we're queueing an empty packet?

This part of the code seems like a no-op.

doowahree: Ahh this is done in a do loop. False alarm!

pklloveyou:
On my server for 1 days and no players have reported bugs

Issue #5985, opened 2021-05-21
Rewrite combat and threat system to be mutual from TC 


Kitzunu:
- https://github.com/TrinityCore/TrinityCore/commit/6113b9dec2e0081c36c36c2974a28ed0693bc601
    - https://github.com/azerothcore/azerothcore-wotlk/pull/15026
- https://github.com/TrinityCore/TrinityCore/commit/e2a1ccd118d129b96e09ff1a15ed0adb1d4a3897
- https://github.com/TrinityCore/TrinityCore/commit/2739a5c5f5d5a359eee4eef1a6486ddcaf24faaf
- https://github.com/TrinityCore/TrinityCore/commit/532ab1c7f8653d1a2e48aa1f1f8a9ba1041d4bb7
- https://github.com/TrinityCore/TrinityCore/commit/9f9507e6a1fd50a5ce643a4096c1712700244a61

Maybe I missed a commit, but this is what I found easily

FrancescoBorzi:
@Kitzunu benefits of such system ?

Kitzunu:
a not fucked up system where mobs aggro the player when your pet has higher aggro than you

Kitzunu:
and a lot of other good stuff which can be found in the commit messages =)

Winfidonarleyan:
Nice nice, very need this

Winfidonarleyan:
https://github.com/WarheadCore/Warhead/commit/74af880217fc567c9949d7557408718af889402f

SoHigh495:
Bounty?

Kitzunu:
What about it

SoHigh495:
How much do you think is needed to get it done? Druid still is very buggy.

Winfidonarleyan:
300 minimum

Winfidonarleyan:
1. https://github.com/TrinityCore/TrinityCore/commit/e2a1ccd118d129b96e09ff1a15ed0adb1d4a3897

sschepens:
What's the status of this issue? I can see that an attempt was made about a year ago but seems to have been closed because of licensing issues, do we still have those licensing issues?

PR #18630, opened 2024-03-30
fix(Scripts/BossAI): add optional variable to BossAI class to make sure the boss calls for help

elthehablo:
Some bosses make sure all nearby friendly creatures assist the boss when engaged. The problem here can be that this is only called on being engaged, which means that if a hunter uses FD right after pull, the boss will move to other targets (due to SetInCombatWithZone) but all the aggroed creatures will reset. This PR makes sure this call for help is called every 2 seconds during combat.

Problems may arise when we implement this on bosses that can move around a lot, as they would start calling every creature in the vicinity as they move around, which wouldn't be intentional. For the current 2 bosses I added this to in the PR, however, this is not an issue, as they have a boundary. And even at the max range of that boundary the range for call to help does not reach creatures it shouldn't.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/6867

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] implement for other relevant bosses (but that's as easy as giving a value for the float in the constructor, but bearing in mind my comment in the beginning about movement)

### DIFF
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp
@@ -571,6 +571,7 @@ BossAI::BossAI(Creature* creature, uint32 bossId) : ScriptedAI(creature),
     summons(creature),
     _bossId(bossId)
 {
+    callForHelpRange = 0.0f;
     if (instance)
         SetBoundary(instance->GetBossBoundary(bossId));
 
@@ -630,6 +631,13 @@ void BossAI::_JustEngagedWith()
     me->setActive(true);
     DoZoneInCombat();
     ScheduleTasks();
+    if (callForHelpRange)
+    {
+        ScheduleTimedEvent(0s, [&]
+        {
+            me->CallForHelp(callForHelpRange);
+        }, 2s);
+    }
     if (instance)
     {
         // bosses do not respawn, check only on enter combat
--- a/src/server/game/AI/ScriptedAI/ScriptedCreature.h
+++ b/src/server/game/AI/ScriptedAI/ScriptedCreature.h
@@ -462,6 +462,8 @@ public:
     BossAI(Creature* creature, uint32 bossId);
     ~BossAI() override {}
 
+    float callForHelpRange;
+
     InstanceScript* const instance;
 
     bool CanRespawn() override;
--- a/src/server/scripts/Outland/TempestKeep/Eye/boss_astromancer.cpp
+++ b/src/server/scripts/Outland/TempestKeep/Eye/boss_astromancer.cpp
@@ -60,6 +60,7 @@ struct boss_high_astromancer_solarian : public BossAI
 {
     boss_high_astromancer_solarian(Creature* creature) : BossAI(creature, DATA_ASTROMANCER)
     {
+        callForHelpRange = 105.0f;
         scheduler.SetValidator([this]
         {
             return !me->HasUnitState(UNIT_STATE_CASTING);
@@ -118,7 +119,6 @@ struct boss_high_astromancer_solarian : public BossAI
     {
         Talk(SAY_AGGRO);
         BossAI::JustEngagedWith(who);
-        me->CallForHelp(105.0f);
         me->GetMotionMaster()->Clear();
 
         scheduler.Schedule(3650ms, [this](TaskContext context)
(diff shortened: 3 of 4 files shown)


Nyeriah (approved):

balleny:
this kind of "making sure groups arent split by buguse" should be implemented also for trash groups in raids and selectively applied to the group (so not calling the room but calling the npcs that belong together). e.g. the 4 initial mobs in AQ40, literally ALL groups formations in raids

elthehablo:
> this kind of "making sure groups arent split by buguse" should be implemented also for trash groups in raids and selectively applied to the group (so not calling the room but calling the npcs that belong together). e.g. the 4 initial mobs in AQ40, literally ALL groups formations in raids

Yes but that is not a trivial fix. So this comes first

balleny:
of course its not trivial, thats also why this issue reappears every raid again.
unfortunately, all implementations so far have been hacky, incorrect or simply not working

bwl
https://github.com/chromiecraft/chromiecraft/issues/3321 into https://github.com/chromiecraft/chromiecraft/issues/4172 (last was also never reopened)

AQ
https://github.com/chromiecraft/chromiecraft/issues/4297
https://github.com/chromiecraft/chromiecraft/issues/6422

Johaine:
Well, is there some consensus on how it should actually work? As far as I can tell, there a several issues, that are similar but not the same.

1. Calling for help: E.g. are NPCs "linked" dynamically on pull? What are the conditions and rules for that?
2. Separating statically linked groups.
3. Respawning statically linked groups

balleny:
tldr:

1) when you engage a group, the full group engages (this usually works fine)
2) the group stays in combat together.

typical scenarios are: 
1) hunter pulls & feign deaths, rogue pulls and vanish. everything resets
2) same as 1) but before the hunter feign deaths or the rogue vanishes, 1 player gets 1 or more ability / hit on a certain mob. that causes the feign death/vanish to reset all "other" mobs except for the one hit by the 2nd player.

and in 2) is the issue. I guess, the easiest solution would be as following: IF a mob is part of a group, engaging with that pack (e.g. by attacking a single mob, healing, buffing etc.) should add the player to the threat table of all mobs of that group or formation. It should not cause them to change their current aggro target, but they should be on the aggro table.

that way, if hunter pulls a group and someone attacks 1 target, a feign death will cause all mobs to go for that 2nd player after feign death.
technically, I dont know if 0 threat can be added on "first" interaction with a pack for every player or if 0 as a value would be ignored.
but from technical point of view, i guess it would be the easiest to cause the first "engagement" with a mob that is part of a formation to become part of the aggro table of each mob of that group.

alternatively, 1 approach would be that mobs "resetting", re-engage on the current target of a mob that is part of a formation and already attacking a target. so either the mobs check - or an mob infight calls their formation for assistance every second.

balleny:
that way, you would be closer to blizzlike from what I would say, and remove a lot of exploit behavior - given mobs are "grouped" coherently across all raids

Johaine:
Well, it is exactly that "alternatively, 1 approach would be that [...]" part that I'm curious about. Does anyone know if attacking a mob of a group actually adds threat for all mobs of that group of official servers? Or if its the latter, mobs automatically assisting their group members if their own threat table is empty for example.

Because both solutions don't really sound that complicated to implement but it's usually nice to know the correct approach before starting to code.

balleny:
i understand that approach. its the decision between a temporarily solution or the correct solution.
talking from point of view of ChromieCraft, the exploitation would surely valid a temp. solution but AZ point of view might be different.

Johaine:
It's not so much about temporary solution vs permanent solution but about correct vs incorrect solution. (Although a temporary fix is usually bad as "Nothing is more permanent than a temporary solution." When something is fixed 90% people tend to care much less about the remaining 10% and it will never be fixed completely.)

E.g., a temporary solution might be added to the current threat system. A permanent solution could probably be part of a complete threat rewrite ( #5985 ).

Even a permanent solution would probably not be accepted if it was incorrect and lead to regressions. From my experience, people can live with long-time bugs but will scream loudly as soon as something that worked before, suddenly doesn't.
So it must be abundantly clear what the correct behaviour is before thinking about any solution.

But, for example, can you tell me what should happen in this case:
Mob A and mob B are in the same group.
Hunter ("P1") shoots mob A. Mob B aggros as well.
Warrior ("P2") taunts and tanks Mob B.
Priest ("P3") heals P2 a lot.
P1 feigns death and disengages with mob A.

Currently, mob B would reset. Instead, which player should mob B attack now? P2 or P3?

balleny:
Definite answer I cannot give cause I cannot test it on retail, but:
If P3 was in range of mob A while casting the heals, the heals will have added up to the threat table (because heals generate threat). The same way, rage gain etc. will have caused threat gain on mob A passively.

in case both have been out of range / distance to generate threat on mob A, imo mob A should start attacking the next closest target that is also on the threat table of mob B (or in cause of a group, on the next closest target that is on the combined threat table of the mobs forming a group)

PR #18647, opened 2024-04-01
feat(Core/Optimization): Correctly document sendBuffer size and optimize it.

Johaine:
While looking at #18628 and trying to understand, how the buffer works, I first noticed that the comment in the current code is wrong.
The buffer is not statically set to 4kB but instead initialized with 4kB and later resized to 64kB. This makes the impact on the CPU time to allocate that buffer (previously on every update even without queued packets) even more pronounced.

So the first commit is a definite fix of the documentation.

The following two commits are a proposal and up for discussion. When players are alone in the wilds, allocating a 64kB buffer every time seems wasted. On the other hand, in raids or cities, having a small buffer also doesn't seem that efficient.
Therefore, we could start with a 4kB sized buffer (as previously implied with the comment) and then increase its size (permanently) when we encounter a single very large packet that needs resizing anyway. Still limit the maximum size we regularly allocate to 64kB.

I cannot test if the reduced default buffer size leads to a measurable performance increase with many players or if it causes more buffers to be required leading to a performance decrease. This should be tested by someone with the means to it. In general, it should lead to a slight decrease in memory usage with many players.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -364,11 +364,11 @@ Network.OutKBuff = -1
 
 #
 #    Network.OutUBuff
-#        Description: Amount of memory (in bytes) reserved in the user space per connection for
-#                     output buffering.
-#         Default:    65536
+#        Description: Amount of memory (in bytes) reserved initially in the user space per
+#                     connection for output buffering.
+#         Default:    4096
 
-Network.OutUBuff = 65536
+Network.OutUBuff = 4096
 
 #
 #    Network.TcpNoDelay:
--- a/src/server/game/Server/WorldSocket.cpp
+++ b/src/server/game/Server/WorldSocket.cpp
@@ -187,14 +187,18 @@ bool WorldSocket::Update()
                 if (!queued->empty())
                     buffer.Write(queued->contents(), queued->size());
             }
-            else    // single packet larger than standard buffer size (Network.OutUBuff)
+            else    // Single packet larger than current buffer size
             {
-                MessageBuffer packetBuffer(currentPacketSize);
-                packetBuffer.Write(header.header, header.getHeaderLength());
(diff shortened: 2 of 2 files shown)


pklloveyou:
I don't think we can sacrifice performance to save memory,
I configured Network.OutUBuff=524288 on the server
If I use the default configuration or a lower value than 256K, the DIFF increases significantly

Johaine:
Memory allocations have a performance cost as well. And your 512kB buffer really seems very large. That's 8 times the default.
What is the performance difference for you between 64kB (the default) and your 512kB setting? How many players and additional modules are installed?

I'm fine with keeping the buffer size as it is right now and only updating the documentation. But it would still be good if someone could test the proposed change and could quanitify its impact.

pklloveyou:
@Johaine When I test the PR, should I change the # back to the default?

pklloveyou:
Can you give me some input if I do?
```
            if (buffer.GetRemainingSpace() < currentPacketSize)
            {
                QueuePacket(std::move(buffer));
                buffer.Resize(_sendBufferSize);
            }

            // Single packet larger than current buffer size
            if (buffer.GetRemainingSpace() < currentPacketSize)
            {
                // Resize buffer to fit current packet
                buffer.Resize(currentPacketSize);

                // Grow future buffers to current packet size if still below limit
                if (currentPacketSize <= 65536)
                    _sendBufferSize = currentPacketSize;
            }

            buffer.Write(header.header, header.getHeaderLength());
            if (!queued->empty())
                buffer.Write(queued->contents(), queued->size());

            delete queued;
```

PkllonG:
No players reported bugs

Nyeriah (approved):

PR #18672, opened 2024-04-05
feat(Core/Scripts): Optimize PlayerScripts by calling only overridden/implemented functions.

walkline:
## Changes Proposed:

At the moment, our scripting system will try to call all available hooks for a registered script. So, if you created and registered a player script that implements only the OnLogin hook, the core would try to call all other possible hooks for this script, such as OnBeforePlayerUpdate, OnPlayerUpdate, OnIsPvP, etc. (but it would execute dummy hooks). And the more scripts you register, the greater the performance impact.

With the suggested changes for PlayerScripts, we would call only hooks that are enabled.
To use this optimization, PlayerScripts should add additional parameter during script initialization. For example
```
class ServerMailReward : public PlayerScript
{
public:
    ServerMailReward() : PlayerScript("ServerMailReward", {PLAYERHOOK_ON_LOGIN}) { }

    void OnLogin(Player* player) override
    {
```

If you don't provide this parameter, then all hooks would be enabled for the script, which makes the change backward compatible.

I'm a little bit exhausted to write boilerplate code, so feel free to update other scriptable objects (like UnitScript, GuildScript, etc.) by utilizing this approach.

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


sudlud:
Nice

pangolp:
After this change, we may have to update the modules, right?

heyitsbench:
We may want to update the modules, but it wouldn't be required for them to continue to function.

pangolp:
Update the branch, with the master branch, because some changes to the tests were added. You should pass the tests now.

sudlud:
What's the performance improvement on clean AC with this patch? :) did you do any tests already, just wondering

walkline:
@sudlud In my case, with around 1k idle players, I see approximately a 11% performance improvement. With real players, I believe the numbers would be different.

sudlud:
@walkline alright, thanks for the feedback

kissingers:
After update, will it still work if the original module not change?

walkline:
@kissingers Yes, they should work.

PkllonG:
I thought about it for a while, and I was afraid that you would be unhappy when you heard my opinion.
I'm sorry I haven't been testing this PR, Because
If I had to choose, I'd rather remove the unused hooks.

sudlud:
@PkllonG removing is not really an option imo

Well, the PR actually "removes" the unused hooks of each PlayerScript instance from the to-execute-list, so the unused hooks of each script don't get called in runtime.

E.g. core already has at least 4 playerscripts which only each use 1-2 hooks actively and the PR disables the rest of the 200 for each script.

But just because the core is not using all 200 hooks of PlayerScript doesn't mean they have no purpose at all. Modules also use scripts and hooks.

PkllonG:
> @PkllonG removing is not really an option imo
> 
> Well, the PR actually "removes" the unused hooks of each PlayerScript instance from the to-execute-list, so the unused hooks of each script don't get called in runtime.
> 
> E.g. core already has at least 4 playerscripts which only each use 1-2 hooks actively and the PR disables the rest of the 200 for each script.
> 
> But just because the core is not using all 200 hooks of PlayerScript doesn't mean they have no purpose at all. Modules also use scripts and hooks.
> 
> Edit: The whole purpose of hooks is that they exist if anyone wants to use them, but they should have zero impact on performance if unused.

Well, it may be that everyone's choice is not in the same direction. I understand how PR works, and thank you for your careful explanation.

sudlud:
@PkllonG  Thanks for your feedback!

Would you still be willing to test the PR? 🙂 

Your frequent testing has been quite valuable on recent PRs.

PkllonG:
> @PkllonG Thanks for your feedback!
> 
> Would you still be willing to test the PR? 🙂
> 
> Your frequent testing has been quite valuable on recent PRs.

okey.

PkllonG:
With or without it ```std::vector<uint16>{PLAYERHOOK_ON_LOGIN}``` , module works fine.

PkllonG:
As for performance. I didn't test on the server.

sudlud:
@PkllonG thanks for testing!

@walkline do you want to merge master branch into PR to get all checks passing?

pangolp:
I don't know if it will pass the tests, because in the core, a series of modules are also validated. I just made a pull request, so that the emulator stops validating those modules. The modules already have tests. It is not necessary for the emulator to validate them every time someone wants to make a contribution. There are people who do not use modules, who use the clean core, blizzlike. https://github.com/azerothcore/azerothcore-wotlk/pull/18690

sudlud:
I tested this PR in regards of the PlayerScript in midsummer.cpp.

The `OnUpdateZone`-Hook still works as expected.

I've then updated the hook list, replacing the hook with a random not-implemented hook of the PlayerScript:
```diff
-MidsummerPlayerScript() : PlayerScript("MidsummerPlayerScript", std::vector<uint16>{PLAYERHOOK_ON_UPDATE_ZONE})
+MidsummerPlayerScript() : PlayerScript("MidsummerPlayerScript", std::vector<uint16>{PLAYERHOOK_ON_LOGIN})
```

The `OnUpdateZone`-Hook was not beeing called anymore (as expected).

I then removed the hook list (to check backwards-compatibility):
```diff
-MidsummerPlayerScript() : PlayerScript("MidsummerPlayerScript", std::vector<uint16>{PLAYERHOOK_ON_UPDATE_ZONE})
+MidsummerPlayerScript() : PlayerScript("MidsummerPlayerScript")
```

The `OnUpdateZone`-Hook works as expected (again).

So I couldn't find any issues, PR works as expected (for this test scenario / PlayerScript).

Nyeriah (approved):

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -710,30 +710,42 @@ class ScriptRegistry
...
     // after server startup.
     static ScriptMap ScriptPointerList;
     // After database load scripts
     static ScriptVector ALScripts;
+    // The list of hook types with the list of enabled scripts for this specific hook.
+    // With this approach, we wouldn't call all available hooks in case if we override just one hook.
+    static EnabledHooksVector EnabledHooks;
 
-    static void AddScript(TScript* const script)
+    static void AddScript(TScript* const script, std::vector<uint16> enabledHooks = std::vector<uint16>())
```
Winfidonarleyan, line 728:
```suggestion
    static void AddScript(TScript* const script, std::vector<uint16> enabledHooks = {})
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -710,30 +710,42 @@ class ScriptRegistry
...
     {
         ASSERT(script);
 
         if (!_checkMemory(script))
             return;
 
+        if (EnabledHooks.size() == 0)
```
Winfidonarleyan, line 735:
```suggestion
        if (EnabledHooks.empty())
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -710,30 +710,42 @@ class ScriptRegistry
...
+            EnabledHooks.resize(script->GetTotalAvailableHooks());
```
Winfidonarleyan, line 736:
```suggestion
            EnabledHooks.reserve(script->GetTotalAvailableHooks());
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -710,30 +710,42 @@ class ScriptRegistry
...
+
         if (script->isAfterLoadScript())
         {
-            ALScripts.push_back(script);
+            ALScripts.push_back(std::pair<TScript*,std::vector<uint16>>(script, enabledHooks));
```
Winfidonarleyan, line 740:
```suggestion
            ALScripts.emplace_back(script, std::move(enabledHooks));
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -776,6 +788,14 @@ class ScriptRegistry
                     // If the script is already assigned -> delete it!
                     if (oldScript)
                     {
+                        for (std::vector<TScript*>& vIt : EnabledHooks)
```
Winfidonarleyan, line 791:
```suggestion
                        for (auto& vIt : EnabledHooks)
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -798,6 +818,9 @@ class ScriptRegistry
             }
             else
             {
+                for (uint16 v : (*it).second)
+                    EnabledHooks[v].push_back(script);
```
Winfidonarleyan, line 822:
```suggestion
                    EnabledHooks[v].emplace_back(script);
```

src/server/game/Scripting/ScriptObject.h
```diff
@@ -53,15 +53,18 @@ class ScriptObject
 
     [[nodiscard]] const std::string& GetName() const { return _name; }
 
+    uint16 GetTotalAvailableHooks() { return _totalAvailableHooks; }
```
Winfidonarleyan, line 56:
```suggestion
    [[nodiscard]] uint16 GetTotalAvailableHooks() { return _totalAvailableHooks; }
```

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -21,1736 +21,1228 @@
 
 void ScriptMgr::OnBeforePlayerDurabilityRepair(Player* player, ObjectGuid npcGUID, ObjectGuid itemGUID, float& discountMod, uint8 guildBank)
 {
-    ExecuteScript<PlayerScript>([&](PlayerScript* script)
-    {
+    CALL_ENABLED_HOOKS(PlayerScript, PLAYERHOOK_ON_BEFORE_DURABILITY_REPAIR, {
         script->OnBeforeDurabilityRepair(player, npcGUID, itemGUID, discountMod, guildBank);
```
Winfidonarleyan, line 25:
```suggestion
    CALL_ENABLED_HOOKS(PlayerScript, PLAYERHOOK_ON_BEFORE_DURABILITY_REPAIR,
    {
        script->OnBeforeDurabilityRepair(player, npcGUID, itemGUID, discountMod, guildBank);
```

src/server/game/Scripting/ScriptMgr.h
walkline: Here we need resize, since later we are accessing elements like this: `EnabledHooks[v].`

Winfidonarleyan: Ah, yes, my bad

Winfidonarleyan (approved):

src/server/scripts/World/server_mail.cpp
```diff
@@ -24,7 +24,7 @@
 class ServerMailReward : public PlayerScript
 {
 public:
-    ServerMailReward() : PlayerScript("ServerMailReward") { }
+    ServerMailReward() : PlayerScript("ServerMailReward", {PLAYERHOOK_ON_FIRST_LOGIN}) { }
```
PkllonG, line 27: PLAYERHOOK_ON_LOGIN

pangolp:
Excellent work, thank you very much.

PkllonG:
nice!

PkllonG (approved):
Great job

pangolp:
@walkline maybe I can annoy you a little bit, let's see if we can write something on the wiki about this (if necessary). I asked if we had to update the modules, but not as a criticism, but as a question, if we really had to do it or not, and it would be interesting, if we have to do it, we can give some example just in case or something.

PR #18679, opened 2024-04-06
feat(Battlegrounds): make it compatible with custom arena type

Helias:
## Changes Proposed:

This PR proposes changes to:
-  make different checks ONLY for standard arena type

## Issues Addressed:
- incompatibility with [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)

## Tests Performed:
This PR has been:
- [ ] Tested in-game by the author ONLY using the module (so far)

## How to Test the Changes:

- test the standard arenas

- [ ] we should make more tests with standard arenas


laasker:
The changes in BattlegroundQueue.cpp causes BG queues to not work

Helias:
> The changes in BattlegroundQueue.cpp causes BG queues to not work

please, check now if this has been solved with the last commit

laasker:
I tested pretty quickly but yea it works, now BG queues pops, normal 2v2/3v3/5v5 queue pops and Solo 3v3 (module) queue pops only when there are 6 people on the queue.

Also can you take a look at my last comment in https://github.com/azerothcore/mod-arena-3v3-solo-queue/issues/8 ?

29Hido:
Just tested it, seems like this changes are breaking https://github.com/azerothcore/mod-1v1-arena

pangolp:
We cannot take the 1vs1 arena module as a reference, because we had to do many things to that module that shouldn't be there, due to other changes in the emulator. If these changes cause errors in the 1v1 arena module, then a pull request is opened within the module. But that doesn't mean we're going to stop making these changes, which may be important for other modules.

Nyeriah (approved):

Helias:
I've merged it because I've tested it a lot

PR #18690, opened 2024-04-10
chore. Remove core module validations

pangolp:
## Changes Proposed:

This PR proposes changes to:
- [x] Ci

image

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Currently, every time the tests are run, this list of modules, which have nothing to do with the emulator tests, is checked. Each module has its own validations, which must be passed before merging it. Checking it again, here in the core, does not really make sense. Because every time one of the modules fails, the validations will not be passed, and that can generate confusion in the people who make contributions.


pangolp:
If you do not want to delete the files, I can restore them, and we can leave the list empty in any case, but I think that leaving the list empty does not make sense.

sudlud:
One benefit is that it compiles with all these module at the same time, which a module CI will not do.

I could imagine that this CI was created to ensure that everything will compile with the modules e.g. beeing used on CC.

But from the view of AC alone, which has the main focus on the core itself, it's correct that it should not really "care" about any collection of modules compiling together nicely.

pangolp:
> One benefit is that it compiles with all these module at the same time, which a module CI will not do.
> 
> I could imagine that this CI was created to ensure that everything will compile with the modules e.g. beeing used on CC.
> 
> But from the view of AC alone, which has the main focus on the core itself, it's correct that it should not really "care" about any collection of modules compiling together nicely.

This is not the complete list either. There are more than 150 modules, plus those that people create and that the community has not made a fork of them. It would be unnecessary to check them all. The modules already have validations.

55Honey:
These are mandatory checks for a basic set of modules. Many of them being used on CC. 
I am not sure about all of them, but there's no way we can just remove the whole check.

pangolp:
> These are mandatory checks for a basic set of modules. Many of them being used on CC. I am not sure about all of them, but there's no way we can just remove the whole check.

I left you my answer privately, to avoid misunderstandings. Unfortunately, I do not agree.

pangolp:
@azerothcore/developers Please, if you can express your opinion about this change, it would help me a lot. My argument, in case it is not clear, is that the modules already have validations. I don't see why, the emulator also has to validate that this list of modules (which are not all), have to pass the tests every time a person makes a change within the emulator. His thoughts or any type of change that can be made in this regard is appreciated.

55Honey:
It is mandatory for us to notice common modules not compiling before the build fails. This PR can not be accepted.

55Honey:
It is mandatory for us to notice common modules not compiling before the build fails. This PR can not be accepted.

FrancescoBorzi:
> My argument, in case it is not clear, is that the modules already have validations.

The modules already have validations indeed, but this is testing the INTEGRATION between AC and modules so this is another kind of check.

I'm against removing it.

Issue #18741, opened 2024-04-13
Crash when sending items by collect mail


pangolp:
### Current Behaviour

When sending a collect email to a player who is then deleted, the server crashes.

![WoWScrnShot_041324_163516](https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/3ff31308-9bdb-4182-9431-c9f142a3170e)

![image](https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/fe7d0512-3947-4ccb-b0e2-165dd7a1ffca)

### Expected Blizzlike Behaviour

Basically, we should corroborate that the player to be deleted has no pending mails. That way, the failure would possibly not occur. 

### Source

None

### Steps to reproduce the problem

Currently, the emulator has a bug, which consists of the following:
1. With one character, send a collect item to another character.
2. Delete the character that should receive the mail.
3. The emulator does not check before deleting the character that the character has pending mails.
4. Crashes the server.

### Extra Notes

None

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/723596c6734a

### Operating system

Windows 10

### Custom changes or Modules

None

Nyeriah:
Can't reproduce following the steps provided on https://github.com/azerothcore/azerothcore-wotlk/commit/cdc72216c8b9282906fdd3634364a29c56973fd0

pangolp:
> Can't reproduce following the steps provided on [cdc7221](https://github.com/azerothcore/azerothcore-wotlk/commit/cdc72216c8b9282906fdd3634364a29c56973fd0)

I am testing with the latest version. Mine was 3 days old.

sudlud:
So the crash should happen when you try to login with the sender of the mail, after the recipient was deleted?

pangolp:
Tested on the latest version of the emulator.

https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/417375c5-4ca3-4a15-b46b-f85bda2ff2b6

pangolp:
> So the crash should happen when you try to login with the sender of the mail, after the recipient was deleted?

For me, the emulator should not let you delete the character if it has a pending mail. The same thing it does when you change faction. You can't change until you delete the email. Although it could be another reason. I mention that, but I'm not 100% sure if it would be the way.

MarioJvPz98:
[...]

PR #18744, opened 2024-04-14
fix(Core/MySQL): _Query m_mStmt==null Crash

PkllonG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18741

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -220,7 +220,10 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
     //ASSERT(m_mStmt); // Can only be null if preparation failed, server side error or bad query
     if (!m_mStmt)
+    {
+        LOG_ERROR("sql.sql", "Could not fetch prepared statement {} on database `{}`.", index, m_connectionInfo.database);
         return false;
+    }
 
     m_mStmt->BindParameters(stmt);
 
@@ -273,7 +276,10 @@ bool MySQLConnection::_Query(PreparedStatementBase* stmt, MySQLPreparedStatement
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
     //ASSERT(m_mStmt);            // Can only be null if preparation failed, server side error or bad query
     if (!m_mStmt)
+    {
+        LOG_ERROR("sql.sql", "Could not fetch prepared statement {} on database `{}`.", index, m_connectionInfo.database);
         return false;
+    }
 
     m_mStmt->BindParameters(stmt);
     *mysqlStmt = m_mStmt;
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -220,7 +220,9 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
     //ASSERT(m_mStmt); // Can only be null if preparation failed, server side error or bad query
     if (!m_mStmt)
+    {
         return false;
+    }
 
     m_mStmt->BindParameters(stmt);
 
@@ -273,7 +275,9 @@ bool MySQLConnection::_Query(PreparedStatementBase* stmt, MySQLPreparedStatement
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
     //ASSERT(m_mStmt);            // Can only be null if preparation failed, server side error or bad query
     if (!m_mStmt)
+    {
         return false;
+    }
 
     m_mStmt->BindParameters(stmt);
     *mysqlStmt = m_mStmt;
--- a/src/server/database/Database/MySQLConnection.cpp
+++ b/src/server/database/Database/MySQLConnection.cpp
@@ -218,7 +218,6 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
     uint32 index = stmt->GetIndex();
 
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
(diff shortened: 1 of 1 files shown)


PkllonG:
@pangolp Can you tell us what you think?

kissingers:
Can add a log if null, so can know what sql problem?

pangolp:
> @pangolp Can you tell us what you think?

As I mentioned in the pull request that I made, I really want the problem to be solved. It really makes no difference to me whether they approve of your method or mine. As long as the fix arrives and the error does not continue, I am satisfied.

pangolp:
> Can add a log if null, so can know what sql problem?

This relationship is a problem, which causes the emulator to crash.

PkllonG:
```2024-04-14 12:41:26 Could not fetch prepared statement 65 on database `characters`, connection type: synchronous.```
There is this

**Comment by PkllonG** [comment]
Player characters can be deleted normally,
I've observed that in the database, the mail that crashed the server was also deleted normally.

**Comment by PkllonG** [comment]
Can you help me if you need to add logs here? I don't speak English very well.

**Comment by kissingers** [comment]
> Can you help me if you need to add logs here? I don't speak English very well. @kissingers

It's horrible, my English is even worse and I totally rely on my browser's automatic translation and deepl

**Comment by PkllonG** [comment]
Can you help me @pangolp

**Comment by PkllonG** [comment]
> It doesn't crash after PR There is this `` 2024-04-14 12:41:26 Could not fetch prepared statement 65 on database `characters`, connection type: synchronous. ``

image
Hints have already been given here

**Comment by PkllonG** [comment]
I'm not crashing after testing

**Comment by pangolp** [comment]
The doubt I have is. Should both pull requests be applied or just one? Because this pull request is fine, however, we must tell the player why the player cannot be removed. In fact, we are debating that the message that we are giving today in the other PR does not have enough information for the player to understand what the cause of the problem is, and correct it before being able to continue with the elimination. In fact, you suggested the message that appears when changing faction, and I changed it, because at first, I also thought about it. But I checked the DELETE CHARACTER messages and was guided by the one I chose in the first instance. I explained the problem, so that we can all determine the best solution and we can solve the problem.

**Comment by PkllonG** [comment]
This PR should also prevent server crashes caused by other reasons

**Comment by pangolp** [comment]
> This PR should also prevent server crashes caused by other reasons

Did that mean we should add both changes or just one?

**Comment by PkllonG** [comment]
> > This PR should also prevent server crashes caused by other reasons
> 
> Did that mean we should add both changes or just one?

I recommend one. Because it doesn't crash, and it allows the player to delete the character normally

**Comment by PkllonG** [comment]
> > > This PR should also prevent server crashes caused by other reasons
> > 
> > 
> > Did that mean we should add both changes or just one?
> 
> I recommend one. Because it doesn't crash, and it allows the player to delete the character normally

The main thing is that it doesn't give players vague error prompts

**Comment by PkllonG** [comment]
[c143330](https://github.com/azerothcore/azerothcore-wotlk/pull/18744/commits/c143330fe2057fe486e5eaa95e514ae6e01a2699) We don't need to pop up the duplicate error log again.

File: src/server/database/Database/MySQLConnection.cpp [hunk 1]
```diff
@@ -218,7 +218,9 @@ bool MySQLConnection::Execute(PreparedStatementBase* stmt)
     uint32 index = stmt->GetIndex();
 
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
-    ASSERT(m_mStmt); // Can only be null if preparation failed, server side error or bad query
+    //ASSERT(m_mStmt); // Can only be null if preparation failed, server side error or bad query
+    if (!m_mStmt)
+        return false;
```
FrancescoBorzi, line 223:
```suggestion
    if (!m_mStmt)
    {
        return false;
    }
```

src/server/database/Database/MySQLConnection.cpp
```diff
@@ -269,7 +271,9 @@ bool MySQLConnection::_Query(PreparedStatementBase* stmt, MySQLPreparedStatement
     uint32 index = stmt->GetIndex();
 
     MySQLPreparedStatement* m_mStmt = GetPreparedStatement(index);
-    ASSERT(m_mStmt);            // Can only be null if preparation failed, server side error or bad query
+    //ASSERT(m_mStmt);            // Can only be null if preparation failed, server side error or bad query
+    if (!m_mStmt)
+        return false;
```
FrancescoBorzi, line 276:
```suggestion
    if (!m_mStmt)
    {
        return false;
    }
```

kissingers:
This is a good fix, because even I use eluna to change some functions, when I write a wrong lua database insert, also will lead server crash,  if can avoid these database crash, and give the obvious mention, the server will be more robust.

pangolp:
Link the solution to the issue. So that in case of a merge, it closes automatically

src/server/database/Database/MySQLConnection.cpp
walkline: @FrancescoBorzi you should check our latest code style convention. The first one was correct.

walkline:
I don't like this solution. You shouldn't have a situation with a nil prepared statement, and you don't need to handle this case.

pangolp:
> I don't like this solution. You shouldn't have a situation with a nil prepared statement, and you don't need to handle this case.

You have another alternative, linked in the topic. And if you don't like any of the two, you can open a third. We have to give you some solution. Because the greats do not have this fault. Even TC has this problem in its core 335 at least (in the rest of the nose). But, the servers, which have been online for years, magically, are all immune... because several people, upon seeing the problem, tried to see if they were also victims and found that they are not. They are an elite, and we are the scum.

Nyeriah:
Thanks for the PR, but I don`t think this is an acceptable solution. If a statement fails rather have it crash than compromise DB data, thus the assert

PR #18754, opened 2024-04-14
fix(Core/Characters): Fix crash on characters deletion with COD mails.

walkline:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18741

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/database/Database/Implementation/CharacterDatabase.cpp
+++ b/src/server/database/Database/Implementation/CharacterDatabase.cpp
@@ -115,7 +115,7 @@ void CharacterDatabaseConnection::DoPrepareStatements()
     // End LoginQueryHolder content
 
     PrepareStatement(CHAR_SEL_CHARACTER_ACTIONS_SPEC, "SELECT button, action, type FROM character_action WHERE guid = ? AND spec = ? ORDER BY button", CONNECTION_ASYNC);
-    PrepareStatement(CHAR_SEL_MAILITEMS, "SELECT creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, playedTime, text, item_guid, itemEntry, ii.owner_guid, m.id FROM mail_items mi INNER JOIN mail m ON mi.mail_id = m.id LEFT JOIN item_instance ii ON mi.item_guid = ii.guid WHERE m.receiver = ?", CONNECTION_ASYNC);
+    PrepareStatement(CHAR_SEL_MAILITEMS, "SELECT creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, playedTime, text, item_guid, itemEntry, ii.owner_guid, m.id FROM mail_items mi INNER JOIN mail m ON mi.mail_id = m.id LEFT JOIN item_instance ii ON mi.item_guid = ii.guid WHERE m.receiver = ?", CONNECTION_BOTH);
     PrepareStatement(CHAR_SEL_AUCTION_ITEMS, "SELECT creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, playedTime, text, itemguid, itemEntry FROM auctionhouse ah JOIN item_instance ii ON ah.itemguid = ii.guid", CONNECTION_SYNCH);
     PrepareStatement(CHAR_SEL_AUCTIONS, "SELECT id, houseid, itemguid, itemEntry, count, itemowner, buyoutprice, time, buyguid, lastbid, startbid, deposit FROM auctionhouse ah INNER JOIN item_instance ii ON ii.guid = ah.itemguid", CONNECTION_SYNCH);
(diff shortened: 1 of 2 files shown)


pangolp:
What operating systems did you test on, and are the steps the same as for the other pull request? Please, if you can complete and give more information, we could ask or perform tests. Complete if possible the fields (even if not all) but the necessary ones so that whoever reads the contribution, knows how to test it. Thank you.

walkline:
@pangolp The steps from your issue (https://github.com/azerothcore/azerothcore-wotlk/issues/18741) are valid, but I could reproduce it only on Windows (but not on macOS).

pangolp:
> @pangolp The steps to reproduce from your issue (https://github.com/azerothcore/azerothcore-wotlk/issues/18741) are valid, but I could reproduce it only on Windows (but not on macOS).

The problem was apparently always Windows. I thought it was something global, but it seems not.

PkllonG:
The server didn't crash after PR (windows10 x64)

pangolp:
> The server didn't crash after PR (windows10 x64)

Is the character deleted without deleting the mail then?

PkllonG:
> > The server didn't crash after PR (windows10 x64)
> 
> Is the character deleted without deleting the mail then?

Press the test in your video, crash before PR, don't crash after PR

pangolp:
> > > The server didn't crash after PR (windows10 x64)
> > 
> > Is the character deleted without deleting the mail then?
> 
> Press the test in your video, crash before PR, don't crash after PR

Beyond the fix, which I have to try, today the emulator does not fail. Unless you had the emulator without updating or using the command as mentioned in the topic. I promise to try it now in a while, because as I mentioned before, the important thing is that it is solved, regardless of who solves it.

sudlud:
just tested this PR

- no crash when deleting the character with pending COD mail
- after character deletion the pending mail still exists in table `acore_characters.mail` and still references the guid of the now-deleted character

is this the expected result?

heyitsbench:
> is this the expected result?

From what I can tell, that is the expected and correct result. Based on old forum posts, it does seem to be possible to delete a character with mail destined to it and the mail would return after thirty days.

sudlud:
> > is this the expected result?
> 
> From what I can tell, that is the expected and correct result. Based on old forum posts, it does seem to be possible to delete a character with mail destined to it and the mail would return after thirty days.

when I login with the mail sender after deletion of the mail receiver, the mail is actually instantly returned.

sudlud (approved):

sudlud:
also no crash when using `.character erase`

pangolp:
I just tested on Windows 10, and apparently, with this change, the crash no longer occurs. However I wonder what happens with the mail... if after 30 days, it returns to the player who sent it, or if in the future, we will have to review that issue as well, but for the moment, I would say that this change is fine. In the rest of the operating systems was it also tested?

pangolp (approved):
Tested on Windows 10

sudlud:
> I just tested on Windows 10, and apparently, with this change, the crash no longer occurs. However I wonder what happens with the mail... if after 30 days, it returns to the player who sent it, or if in the future, we will have to review that issue as well, but for the moment, I would say that this change is fine. In the rest of the operating systems was it also tested?

Please see my earlier comment - the mail gets returned instantly when the mail sender logs in again.

pangolp:
> > I just tested on Windows 10, and apparently, with this change, the crash no longer occurs. However I wonder what happens with the mail... if after 30 days, it returns to the player who sent it, or if in the future, we will have to review that issue as well, but for the moment, I would say that this change is fine. In the rest of the operating systems was it also tested?
> 
> Please see my earlier comment - the mail gets returned instantly when the mail sender logs in again.

I did 2 tests. In the first one, what you said happened, however, it seemed to me that the money was not returned, only the item. So I took the test again, and since the email didn't arrive, I continued with another topic and that's why I asked. Did the gold also reach you? Either way, it may be cause for another pull request. It doesn't have to be repaired right here. But we would have to take it into account to at least create the issue so that the precedent of the failure remains and then we try to repair it.

sudlud:
This also confused me, because I didn't understand what "C.O.D." Mail is.
You actually send an item and request a certain amount of gold from the receiver, that he has to pay to get the item from the mail.

So the returning mail does not have any gold in it, just the item.

pangolp:
> This also confused me, because I didn't understand what "C.O.D." Mail is.
> You actually send an item and request a certain amount of gold from the receiver, that he has to pay to get the item from the mail.
> 
> So the returning mail does not have any gold in it, just the item.

Exact. In case of non-payment or deletion, both must be returned. But I'm just saying this, so that we create the issue and then resolve it in another pull request. This pull request seems to solve the problem and should be merged. But let's not forget to do the problem, to review and solve the other problem.

PR #18812, opened 2024-04-26
fix(Core/Map): remove hackfix that blocked update of GOs beeing set active

sudlud:
- this just broke the whole purpose of setActive(true) if the gameobjects is still not really beeing set active afterwards
- GetGridActivationRange() returns 0.0f for gameobjects anyways
- so CalculateCellArea() will result in the minimal cell area around the gameobject's position
- if the cell around the gameobject should not be updated, the gameobject should not have been set active in the first place

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

fun fact:
Battlefield::SpawnGameObject() sets all objects active after spawning them

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -702,9 +702,6 @@ void Map::VisitNearbyCellsOf(WorldObject* obj, TypeContainerVisitor<Acore::Objec
     if (!obj->IsPositionValid())
         return;
 
-    if (obj->GetGridActivationRange() <= 0.0f) // pussywizard: gameobjects for example are on active lists, but range is equal to 0 (they just prevent grid unloading)
-        return;
-
     // Update mobs/objects in ALL visible cells around object!
     CellArea area = Cell::CalculateCellArea(obj->GetPositionX(), obj->GetPositionY(), obj->GetGridActivationRange());
 


PkllonG:
What needs to be tested.

sudlud:
> What needs to be tested.

Well this should actually not have any negative impact at all.

It might increase cpu load slightly (maybe?) as all gameobjects beeing set active now actually are beeing handled accordingly in the map update.

So if theres no cpu load issues in a high population server for example, then this should be fine.

sudlud:
Testing could also be done by assigning a custom gameobject script to a gameobject, do setActive(true) inside.
When leaving the map where the gameobject is in, it should still update the script.

Before PR script execution would stop until the player comes back.

PkllonG:
1 player, CPU usage is almost comparable

pangolp:
Can this cause additional load understand? Or did the translator misinterpret your words?

sudlud:
> Can this cause additional load understand? Or did the translator misinterpret your words?

It depends.

I didn't check the whole core to see which gameobjects are set active.
So any setActive gameobject will be handled as "active" correctly after this PR.
So on each map update, anything in the same cell as the gameobject will also be updated. Or should be just one cell because the "activision range" of gameobjects is 0.0f.

So this might have a slight increase in load, but most likely right now it should bot make any difference.

Also the "increased load argument" is irrelevant in this case, as the hackfix has to be removed in my opinion and of any gameobjects should not be set active, that's the topic of another PR.

pangolp:
> > Can this cause additional load understand? Or did the translator misinterpret your words?
> 
> It depends.
> 
> I didn't check the whole core to see which gameobjects are set active. So any setActive gameobject will be handled as "active" correctly after this PR. So on each map update, anything in the same cell as the gameobject will also be updated. It should be just one cell because the "activision range" of gameobjects is 0.0f.
> 
> So this might have a slight increase in load, but most likely right now it should not make any difference.
> 
> Also the "increased load argument" is irrelevant in this case, as the hackfix has to be removed in my opinion and if any gameobjects should not be set active, that's the topic of another PR.

Thanks for the reply. Honestly, I don't have enough knowledge to make a decision on a topic like this. I prefer that another colleague look at it. I trust his judgment to make the decision, but I would be calmer if someone else reviewed it for me.

Nyeriah (approved):

PkllonG:
CPU usage is doubled

Nyeriah:
Objects shouldn’t be set active indiscriminately and should also be set inactive once they’re no longer in use. Higher resource usage is likely due to indiscriminate/arbitrary use of the active flag. I still see no issue with this PR in particular, the faulty cases/bad scripting needs to be addressed instead

Issue #17186, opened 2023-09-04
Goblins do not attack the player when pvp inside Area 52


Pyromanihack:
Original issue: https://github.com/chromiecraft/chromiecraft/issues/5120

### What client do you play on?

enUS

### Faction

Alliance

### Content Phase:

70

### Current Behaviour

When you fight inside Area 52 goblin guards turns red but they don't attack you and and you can't attack them
![image](https://user-images.githubusercontent.com/124716025/222922670-f81deedd-2c78-411d-912d-d60bccc0ea4b.png)

### Expected Blizzlike Behaviour

As a neutral city the guards should stop any pvp action near

### Source

https://www.youtube.com/watch?v=UvqAYdOSMbs&ab_channel=TheSurprisingGirl
https://www.wowhead.com/wotlk/npc=19541/netherstorm-agent
https://www.wowhead.com/wotlk/npc=22494/cosmowrench-bruiser

### Steps to reproduce the problem

1. `.tele area52` with 2 opposite faction player
2. Fight inside Area 52
3. Goblins stay quiet

### AC rev. hash/commit

6cd3b2a32cd9

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
[...]

PR #18847, opened 2024-04-30
fix(Spells/SpellInfoCorrections) PVP in area 52

pangolp:
Co-authored-by: Jelle Meeus <jellemeeus@hotmail.com>

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17186
- Closes https://github.com/chromiecraft/chromiecraft/issues/5120

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. .`tele area52` with 2 opposite faction player
2. Fight inside Area 52
3. The guards will start hitting you if you put up a fight.

https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/a38a081a-b080-432a-add6-4312123e19d6


pvp100com:
I tested this PR, when I attack hostile players, the guards will immediately attack me. after I am killed by the guards and resurrected, they still attack me. I remain unable to attack these guards.

I'm not sure if this meets expectations. thank you.

pangolp:
> I tested this PR, when I attack hostile players, the guards will immediately attack me. after I am killed by the guards and resurrected, they still attack me. I remain unable to attack these guards.
> 
> I'm not sure if this meets expectations. thank you.

According to the videos we were watching, no player can attack the guards. The guards are there to prevent pvp from being done in the area. So we assume that's what should happen, but if anyone has more information, we'll check it out.

pvp100com:
After being killed by guards, you will still be marked as hostile by them and they will continue to attack. and waiting for a few minutes, the guards will return to a friendly state. If that's as expected, It worked really well.

pangolp:
> After being killed by guards, you will still be marked as hostile by them and they will continue to attack. and waiting for a few minutes, the guards will return to a friendly state. If that's as expected, It worked really well.

Thanks for the tests. Could you try the other gobbin areas, such as `Booty Bay` and `Tanaris`? Just in case, those mechanics have not been broken, although that should not happen.

pvp100com:
After testing, it was found that the “Faction” of guards in Tanaris is “475”. Therefore, when I attack hostile players, the guards will retaliate against me. In contrast to this PR (Faction = 1857 and 1806), when I die and respawn, the guards will no longer continue to attack me and will treat me as friendly.

If possible, I think the effect of this PR should be consistent with Tanaris. thank you.

pangolp:
> After testing, it was found that the “Faction” of guards in Tanaris is “475”. Therefore, when I attack hostile players, the guards will retaliate against me. In contrast to this PR (Faction = 1857 and 1806), when I die and respawn, the guards will no longer continue to attack me and will treat me as friendly.
> 
> If possible, I think the effect of this PR should be consistent with Tanaris. thank you.

It is difficult to achieve this effect, because the problem is that the area 52 faction is friendly by default. On the other hand, in Tanaris, she is not friendly. The problem is in the DBC. We have to eliminate the default behavior. I don't know if we can achieve the same effect as in Tanaris, however, I would fix the problem, and then if anything, I would improve it. But today it doesn't even work. I would like to add it, and then improve it in another pull request.

Thanks for the tests.

pangolp:
Anyway, there's no rush to merge it. We can wait, find out a little more, and see if we can achieve the same effect as in Tanaris. Although it will depend on the evidence we have in this regard. I really don't know how it works on the official servers.

pvp100com:
Ratchet guard (637) and Booty Bay guard (121) are the same as Tanaris. For your reference, thank you.

jellemeeus:
The goblin guards at K3 Storm Peaks suffer from the same issue. I can't think of any other goblin guards

K3 Bruiser, faction 1812

tp to K3
```
.go xyz 6124 -1023 409.1 571 4.67
```

pangolp:
And then they say that I don't like giving credits, but this time, I think I got out of hand.

pangolp:
This is a disaster, nothing is understood, I better close this and create it again.

PR #18893, opened 2024-05-13
Emerald Dragons no longer have 0% health events

Day36512:
Events happen at 75, 50 or 25 percent health and should not include a 0% event.

## Changes Proposed:

This PR proposes changes to:

## SOURCE:

Video showing Dot event does not happen at 0% as an example:
https://youtu.be/Wos1qtt280g?si=U6Kif8SeUgVI51lR

## Tests Performed:
This PR has been:


elthehablo:
Might as well change the script to use ScheduleHealthCheckEvent. It's the more modern and preferred approach. It can be called on reset

elthehablo (dismissed):
Use ScheduleHealthCheckEvent

elthehablo:
Actually I 100% forgot this is not BossAI and cannot be and I don't think it is implemented for ScriptedAI

Nyeriah (approved):

Issue #8921, opened 2021-11-03
ICC Saurfang combat start dialog with Muradin Bronzebeard are broken


zqgmzq:
### Current Behaviour

ICC Saurfang combat start dialog Muradin Bronzebeard are broken

### Expected Blizzlike Behaviour

ICC Saurfang combat start dialog Muradin Bronzebeard is ok

### Steps to reproduce the problem

Go ICC
Start batle with Saurfang

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/bfe290477ce4c4787555671021a1a280daf00be7

### Operating system

Debian 10

Shineslippers:
I guess he is mixing Horde/Allince text
![WoWScrnShot_110421_185130](https://user-images.githubusercontent.com/85256266/140372612-836006af-d312-4589-b138-b97b072378ca.jpg)

Wowwiki [Saurfang quotes](https://wowwiki-archive.fandom.com/wiki/Deathbringer_Saurfang?oldid=2386527#Quotes)

Ravenguard13:
Another thing is - the ending of the fight is broken too. 

After he dies, the Big Saurfang has to show up and take his son's body. Unfortunately, this script is absent, for both factions.

Not sure if i'm supposed to make a new topic about this or just say it here

Ravenguard13:
@Azcobu  it's "Content 80", not 60 ))))

Shineslippers:
It's auto. Don't understand. I labled it 80.

pangolp:
If I make this query
```sql
UPDATE `creature` SET `id1`=37200 WHERE `guid`=200982;
```
The dialogue in the alliance is perfect.
Although I will surely fail in the horde.

I'm not saying that should be the answer.
But it's funny, how when you make that change, it "fixes itself"
I think the problem is in the `Creature::UpdateEntry`.
But I have already tried several times to find the solution and I have not been able to.

pangolp:
Another change I made was this.

```cpp
uint32 Criatura::GetScriptId() constante
{
     if (CreatureData const* criaturaData = GetCreatureData())
         if (uint32 scriptId = criaturaData->ScriptId)
             devolver scriptId;

     devolver sObjectMgr->GetCreatureTemplate(GetEntry())->ScriptID;
}
```

For this

```cpp
uint32 Criatura::GetScriptId() constante
{
[...]

PR #18931, opened 2024-05-20
fix(ICC/Deathbringer Saurfang) NPC Muradin Dialogues

pangolp:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/8921
- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/13506

## Tests Performed:

This PR has been:

WoWScrnShot_052024_041835

WoWScrnShot_052024_041928

## How to Test the Changes:

1. `.tele IcecrownCitadel`
2. Eliminate the first 3 bosses, and then start the fourth fight. Without the pull request, the script is not updated, and the dialogs are displayed incorrectly. I don't know if there is another way to do it, but I looked for thousands of ways, and the only one that occurred to me, as a final point, to fix it, until a better way appears, was this.


PkllonG:
https://github.com/azerothcore/azerothcore-wotlk/pull/13506

pangolp:
> https://github.com/azerothcore/azerothcore-wotlk/pull/13506
> You try this and it works fine on my server.

I'll try it. I don't understand why they closed their pull request. I've been trying to figure this out for quite some time. I made many changes, before uploading this PR. Since I couldn't find a solution, I leaned towards this path.

pangolp:
@PkllonG Could you please upload an image of the `.npc info` command from your server? Thank you. I have uploaded images of the results of mine. However, I would like to see the result you get.

PkllonG:
You try, it's definitely good.

PkllonG:
image

pangolp:
It seems like your change would be correct, I'll try it later. Thank you.

pangolp:
Your code seems to be correct. I only made a small modification. The `GetScriptId` was added differently within the `UpdateEntry`.

pangolp:
I just made the change I mentioned, if you want, you can try it to see if you get the same result. I just tried it. It seems to work. We can also close this pull request, and re-open yours. I gave him credits for the commit. But I leave it to your discretion. It's a shame not to have seen it before. I have been trying to solve this for several months and carried out tests of all kinds.

PkllonG:
You can come,

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
```diff
@@ -399,11 +399,10 @@ class instance_icecrown_citadel : public InstanceMapScript
                 case NPC_SE_HIGH_OVERLORD_SAURFANG:
                     if (TeamIdInInstance == TEAM_ALLIANCE)
                     {
-                        creature->UpdateEntry(NPC_SE_MURADIN_BRONZEBEARD, creature->GetCreatureData());
+                        creature->UpdateEntry(NPC_SE_MURADIN_BRONZEBEARD, creature->GetCreatureData(), true, true);
                         creature->LoadEquipment();
                     }
                     DeathbringerSaurfangEventGUID = creature->GetGUID();
-                    creature->LastUsedScriptID = creature->GetScriptId();
```
PkllonG, line 406: Don't delete here

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -594,6 +594,8 @@ bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data, bool changele
         AIM_Initialize();
     }
 
+    LastUsedScriptID = GetScriptId();
```
PkllonG, line 597: Please delete here

pangolp:
@PkllonG do you want to make those changes if the result is the same? Did you try it? Does it affect something I'm not considering? I need context because you're asking me for those changes. If it affects something that doesn't, then it would be good to know.

PkllonG:
The result is the same

pangolp:
> The result is the same

All the more reason, if the result is the same, when you consult the script inside the UpdateEntry, you save yourself from doing it outside it. Unless it affects another mechanic, I think it would be better that way.

Unless we're making a mistake, and then yes, we can do it the way you indicate. But if the result is the same, I prefer to do it this way.

pangolp:
If I have to make those changes, the truth is, I don't care, I have no problem doing it. What I want is for it to solve the problem. I don't care about the method, I'm not interested in being right. I'm just looking for the fault to be corrected.

pangolp:
Restore how to update the script. Actually, as long as the problem is solved, it is indistinct to me, if it is updated in creature or in the instance. I still believe that it is better within UpdateEntry because I would always update the script of the npc, but to avoid confessions, I left it as it was.

pangolp:
https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/0bbd5e9b-6bc8-4957-b765-ae0e25498edf

https://github.com/azerothcore/azerothcore-wotlk/assets/2810187/d8988465-0271-4ded-a180-751743fa76d5

PkllonG:
Not all UpdateEntry need to be loaded with AI,
Changing only partial code makes it easier to merge code and safer.

PkllonG (commented):
It's great now

PkllonG (commented):
Good job

pangolp:
> Good job

It was your job. My code was different. But if that way works, and we don't have to add a spawn, it is better to do it this way.

PkllonG:
It has been working on my server since very early days and has not found bugs in a few years

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -3015,8 +3015,11 @@ std::string Creature::GetScriptName() const
 uint32 Creature::GetScriptId() const
 {
     if (CreatureData const* creatureData = GetCreatureData())
-        if (uint32 scriptId = creatureData->ScriptId)
+    {
+        uint32 scriptId = creatureData->ScriptId;
+        if (scriptId && GetEntry() == creatureData->id1)
```
Nyeriah, line 3020: Please explain this change

pangolp: From what I could see, through this change, what can be achieved is that if the npc has an updateEntry, then it takes the script that corresponds to it, otherwise, it always takes the script from the npc before the update. That's why the dialogues worked poorly. Initially, I had proposed something else, which was to spawn the npc, but this is not the first one to have this problem. What is not entirely clear to me, if necessary, put a true in the `bool changelevel`, which is the third parameter of the method.
```cpp
bool Creature::UpdateEntry(uint32 Entry, const CreatureData* data, bool changelevel, bool updateAI)
```

pangolp:
In the image of the pull request, you can see how the npc updates its entry and also the id, as the name of the script it should use. Currently, without this change, muradin, despite doing the updateEntry, continues to retain the id of the original npc, as well as its script.

pangolp:
Currently, in the master branch of the repository, the npc, when doing an updateEntry, updates several of the data, but does not apply changes to id1 and consequently, does not update the script it should use either. That is why the dialogue is broken, and for that reason, this pull request is being created to solve that problem.

muradin

After applying the changes, you can confirm that the id1 of the npc is updated, and also its script. If this pull request is approved, we may have to review several more npc. Mainly, in those instances, where updateEntry occur.

muradin_fix

src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
```diff
@@ -401,7 +401,7 @@ class instance_icecrown_citadel : public InstanceMapScript
                 case NPC_SE_HIGH_OVERLORD_SAURFANG:
                     if (TeamIdInInstance == TEAM_ALLIANCE)
                     {
-                        creature->UpdateEntry(NPC_SE_MURADIN_BRONZEBEARD, creature->GetCreatureData());
+                        creature->UpdateEntry(NPC_SE_MURADIN_BRONZEBEARD, creature->GetCreatureData(), true, true);
```
Nyeriah, line 404:
```suggestion
                        creature->UpdateEntry(NPC_SE_MURADIN_BRONZEBEARD, true);
```

src/server/game/Entities/Creature/Creature.cpp
pangolp: How else do we solve the problem? Because if we set that value in True, the script is not updated and there is a flaw that can be seen in the videos. I uploaded several videos explaining the difference when applying the changes, but if there's another way to solve the problem, I'd like to know. The method might not be right, it can be changed. The goal is to correct the problem.

pangolp: From the mobile phone, I didn't see that you had sent a suggestion for change. I applied it, I'll try it. I'll see if I get the same result. Thanks.

pangolp:
@Nyeriah after the change is applied, mainly in the `Creature.cpp` file, I will continue updating several npcs, from other instances, or within this same one, that have the same problem. But first, we have to test if this works before we can make the other changes.

pangolp:
If necessary, we can split the issue into 2 pull requests. One to update the changes in Creature.cpp and another to make the change to the script. In case it needs to be reverted, I have no problem creating a second pull request if necessary and splitting the issue. If you need anything, let me know.

pangolp:
I've been testing the changes, and there seems to be no difference. The script works as it should. However, the `id1` inside the parenthesis is not updated. It is the only value that remains unchanged, and is still the id of the original npc.

`Current Entry: 37200 of (37187, 0, 0).`

https://www.wowhead.com/wotlk/npc=37200/muradin-bronzebeard
https://www.wowhead.com/wotlk/npc=37187/high-overlord-saurfang

I don't know if we should update that as well, or if it's not necessary to do so. In the script, no changes are made on that subject.

Nyeriah:
I'll look into this over the weekend.

pangolp:
> I'll look into this over the weekend.

Okay. Look into it and let me know if you have any questions, so I can modify this pull request and after it is merged, create another one with the changes I mentioned, because I feel that it won't be the only NPC that we have to make this type of change to. There may be more within the instance and in other instances as well. For example, in Pit of Saron

PkllonG:
Maybe I can help a little because the relevant changes have been working fine on my server for a long time. Bugs have all been fixed.

pangolp:
> Maybe I can help a little because the relevant changes have been working fine on my server for a long time. Bugs have all been fixed.

We could only make the changes to the Creature.cpp file in this PR and then create another PR with the changes to the npc that we must update. So, if we were to reverse one, it wouldn't affect the other.

Nyeriah (approved):

pangolp:
Thanks.

PR #18971, opened 2024-05-29
fix(core/character-creation): get main part of cyrillic name

obergodmar:
## Changes Proposed:

This PR proposes changes to:

Currently, in the Cyrillic version of the server and client, there is an issue with the declension of a name when creating a new character. You can watch the video to see what the problem is:

https://github.com/azerothcore/azerothcore-wotlk/assets/33424304/7117aee2-830c-4fff-9338-1588e8417aad

This PR solves the problem.

## Issues Addressed:

I didn't find a similar issue.

## SOURCE:

I applied them on my server and everything worked fine. The DB table `character_declinedname` was populated with correct declined names.

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided below

1. Configure the server locale to be Russian.
2. Run ruRU client.
3. Create a character with a declinable name.
4. Apply declension after the character is being created.
5. There is no error dialog and you are joining the world with created character.

### DIFF
--- a/src/common/Utilities/Util.cpp
+++ b/src/common/Utilities/Util.cpp
@@ -383,7 +383,7 @@ void wstrToLower(std::wstring& str) { std::transform(std::begin(str), std::end(s
 void strToUpper(std::string& str) { std::transform(std::begin(str), std::end(str), std::begin(str), charToUpper); }
 void strToLower(std::string& str) { std::transform(std::begin(str), std::end(str), std::begin(str), charToLower); }
 
-std::wstring GetMainPartOfName(std::wstring const& wname, uint32 declension)
+std::wstring GetMainPartOfName(std::wstring const& wname, uint32_t declension)
 {
     // supported only Cyrillic cases
     if (wname.empty() || !isCyrillicCharacter(wname[0]) || declension > 5)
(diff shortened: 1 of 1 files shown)


src/common/Utilities/Util.cpp
```diff
@@ -392,22 +392,22 @@ std::wstring GetMainPartOfName(std::wstring const& wname, uint32 declension)
     }
 
     // Important: end length must be <= MAX_INTERNAL_PLAYER_NAME-MAX_PLAYER_NAME (3 currently)
-    static std::wstring const a_End    = { wchar_t(0x0430), wchar_t(0x0000) };
```
obergodmar, line 395: Endings are initialized with a wchar_t(0x0000) at the end. This is not necessary and might cause issues with string comparisons. In C++, strings are null-terminated automatically, so we don't need to include the null character explicitly.

src/common/Utilities/Util.cpp
```diff
@@ -421,16 +421,21 @@ std::wstring GetMainPartOfName(std::wstring const& wname, uint32 declension)
...
+        }
+
+        std::wstring const& ending = *endingPtr;
         std::size_t const endLen = ending.length();
         if (endLen > thisLen)
         {
             continue;
         }
 
-        if (wname.substr(thisLen - endLen, thisLen) == ending)
```
obergodmar, line 433: The comparison logic might be flawed because the substring method is used incorrectly. It should compare the correct substring of the name with the ending.

heyitsbench:
TIL about declension. Good find!

Winfidonarleyan:
Work
image

Winfidonarleyan (approved):

Nyeriah (approved):

Winfidonarleyan:
Thanks for contribute

PR #18978, opened 2024-05-30
fix(Scripts/IcecrownCitadel): The Lich King should switch phased instantly after reaching specefic life

qwe900:
The Phases of the Lich King should be instant switch after reaching :
-> under 70% of life 
-> under 40% of life
-> under 10% of life

Everthing he is doing except when players are down in the rooms should be canceld and next phase beginn.

Before it will not much effect the gameplay except u have a full raid off BiS chars, then u reach to the point where u normale
jump from p2 to p3  while p2 is running and then u are forced to play like rumorelesswinter longer as you should even the boss is under 10% already.

In cases where u come easily to these scenario like u use mod-autobalance u make stuk as the boss is already dead and and phase of outro didnt even beginn.

## Changes Proposed:

This PR proposes changes to:

## SOURCE:

On Retail just go with good char and Kill the LK as fast as pos, he will not force you into these scenario.
Lot of BFA solo videos shows that the LK cancels the remorelesswinter and jumps into nextphaes p2->p3 and p3->outro

## Tests Performed:

This PR has been:

I only checked it now by myself as iam now longer have a community server to test these changes with multiple people.

## How to Test the Changes:

1. Kill all bosses before LK
2. Do the LK fight with a lot of good people or use autobalance or stat increasing moduls
3. profit

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
@@ -765,6 +765,13 @@ public:
             }
         }
 
+        uint32 CalculateHealthPercentageAfterHit(uint32 damage)
+        {
+            uint32 expectedHealthAfterDamage = me->GetHealth() > damage ? me->GetHealth() - damage : 0;
+            float healthPercentageAfterHit = (static_cast<float>(expectedHealthAfterDamage) / me->GetMaxHealth()) * 100.0f;
+            return static_cast<uint32>(healthPercentageAfterHit);
+        }
+
         void DamageTaken(Unit* attacker, uint32& damage, DamageEffectType, SpellSchoolMask) override
         {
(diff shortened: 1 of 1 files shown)


src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
```diff
@@ -765,6 +765,13 @@ class boss_the_lich_king : public CreatureScript
             }
         }
 
+        uint32 CalculateHealthPercentageAfterHit(uint32 damage)
```
sogladev, line 768: This method already exists in Unit, e.g. `me->HealthBelowPctDamaged(70, damage)`

https://github.com/azerothcore/azerothcore-wotlk/blob/0f5992045ccd8e7c9fc320cd21f2cd4184552de4/src/server/game/Entities/Unit/Unit.h#L1539

qwe900: cool didnt now that,

PR #9281, opened 2021-11-22
feat(Core/Modules): add separated lib for modules

Winfidonarleyan:
## Changes Proposed:
1. General
- Make sepated library `modules`
- Make new manager `ModuleMgr` for check enable modules and other
- Enable loading modules configs if module enable. (Previously, all configurations were always loaded regardless of whether the module was disabled or not)
2. Eluna
- Move all `ELUNA` code to module
- Added new hooks for support `mod-eluna`
- Fix using double core hooks

## Issues Addressed:
- Part dynamic linking

## Tests Performed:
- In game

## How to Test the Changes:
1. Startup core with/without modules
2. Check any scripts
3. Check any modules
4. Check eluna scripts with https://github.com/azerothcore/mod-eluna-lua-engine/pull/71

## Linked PR's:
- [ ] https://github.com/azerothcore/mod-eluna-lua-engine/pull/71

## Media:
- Add new info in command `.ser debug`
image

- New structure for modules
image

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #19081, opened 2024-06-17
fix(Scripts/AllCommandScript) CanExecuteCommand renamed and moved to be executed for all commands.

AnchyDev:
## Changes Proposed:

This PR proposes changes to:

When using the AllCommands::CanExecuteCommand hook, only failed commands are captured.
This is because the call is inside of the failure routine, so I moved it.

## Issues Addressed:

None

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Set a breakpoint on line 328 of ChatCommand.cpp
2. Observe it is hit for all commands, failed or not.
3. Yay?

Alternatively you can download [this](https://github.com/AnchyDev/ButWhy) module and observe that only failed commands are blocked.

I don't see any issue with capturing all commands.


src/server/game/Chat/ChatCommands/ChatCommand.cpp
```diff
@@ -322,8 +322,14 @@ namespace Acore::Impl::ChatCommands
         oldTail = newTail;
     }
 
+    /* if we matched a command at some point, invoke it */
     if (cmd)
-    { /* if we matched a command at some point, invoke it */
+    {
+        if (!sScriptMgr->CanExecuteCommand(handler, cmdStr))
+        {
```
sudlud, line 329: You can remove the braces

AnchyDev: > You can remove the braces

Alright, it looks like the AzerothCore code style changed lol. I'll update it

src/server/game/Chat/ChatCommands/ChatCommand.cpp
```diff
@@ -345,9 +344,7 @@ namespace Acore::Impl::ChatCommands
     }
 
     if (!sScriptMgr->CanExecuteCommand(handler, cmdStr))
```
sudlud, line 346: Why are we having this twice here? 👀

sudlud (commented):
Why calling the hook twice

src/server/game/Chat/ChatCommands/ChatCommand.cpp
AnchyDev: I don't know, it was originally there so I just left it. It can probably be removed.

AnchyDev:
> Why calling the hook twice

Monkeys

sudlud:
looking again at this, it feels like this hook was specifically placed to be called only on unknown / failed command executions.
looks like it was introduced with #9281 and replacing `sEluna->OnCommand(...)`, so there is at least a naming difference which implies a different use case

sudlud (approved):

Issue #11719, opened 2022-05-11
Noggenfogger Elixir effect may disable AA


Si1ker:
Originally reported by Sílker at https://github.com/chromiecraft/chromiecraft/issues/3495

### Current Behaviour

Noggenfogger has **the chance** to disable auto-attacks.

When you drink [Noggenfogger Elixir](https://classic.wowhead.com/item=8529/noggenfogger-elixir) you have the chance to get one of the following three effects:

[You feel light](https://classic.wowhead.com/spell=16593/noggenfogger-elixir)
[You feel smaller](https://classic.wowhead.com/spell=16595/noggenfogger-elixir)
[You've stopped breathing](https://classic.wowhead.com/spell=16591/noggenfogger-elixir)

The third one, once you turn into a skeleton, has a low-rate bug (it is not a 100% reproducible bug), it may disable your auto-attacks until you log back into the game. For personal testing, I'd say a 20% repro-rate.

The [first](https://classic.wowhead.com/spell=16593/noggenfogger-elixir) and [second](https://classic.wowhead.com/spell=16595/noggenfogger-elixir) effects have not triggered this bug in 100 tests. Only the [third](https://classic.wowhead.com/spell=16591/noggenfogger-elixir) one.

Note:  **/startattack** , does not work when you're affected by the bug.

Tested in chromiecraft after latest update 5/11 on [c7f74e5](https://github.com/chromiecraft/azerothcore-wotlk/commit/c7f74e563758241e29721e95b86aed0093a8856a)

https://user-images.githubusercontent.com/61223313/167906947-a5cd84e4-f623-4c99-8b04-c173ac0b9eb4.mp4

### Expected Blizzlike Behaviour

[You've stopped breathing - Aura 16591](https://classic.wowhead.com/spell=16591/noggenfogger-elixir) shouldn't have the chance to disable your auto-attacks

### Source

Is not planned for an esthetic potion to break your gameplay

### Steps to reproduce the problem

.add 8529 100

Drink 'till you get the aura 16591 -> try AA to a target
Were you able to attack?
- Yes -> repeat again.
- No -> found the bug.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/c7f74e563758241e29721e95b86aed0093a8856a

### Operating system

Ubuntu 20.04

### Custom changes or Modules

N/A

UltraNix:
@Si1ker Still a bug? I remember I fixed something very similar bug.

[...]

PR #19267, opened 2024-07-04
fix(Core/Handlers): Adjust start swing handler.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Core (handlers).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11719
- Closes https://github.com/chromiecraft/chromiecraft/issues/3495
- Likely closes https://github.com/chromiecraft/chromiecraft/issues/7070
- Might close https://github.com/chromiecraft/chromiecraft/issues/7001

## Tests Performed:

This PR has been:

## How to Test the Changes:

So this fix is weird. From what I understand, it's trying to avoid a null pointer being floated along the way, but the issue is that when it gets to the removed function and stops the attack, it breaks the client and the client can no longer auto attack anything ever again. The behavior can easily be seen (iirc) by doing something like casting the aether ray wrangling rope but removing the target before the cast finishes, or by using the macro in the Noggenfogger issue. I don't know if the proposed change is necessarily correct, but I do know that it fixes the issues seen. I also know I would like thorough testing to be done on this PR if at all possible.

### DIFF
--- a/src/server/game/Handlers/CombatHandler.cpp
+++ b/src/server/game/Handlers/CombatHandler.cpp
@@ -35,6 +35,13 @@ void WorldSession::HandleAttackSwingOpcode(WorldPacket& recvData)
 
     Unit* pEnemy = ObjectAccessor::GetUnit(*_player, guid);
 
+    if (!pEnemy)
+    {
+        // stop attack state at client
+        SendAttackStop(nullptr);
+        return;
+    }
+
     if (!_player->IsValidAttackTarget(pEnemy))
     {
         // stop attack state at client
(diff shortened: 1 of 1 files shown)


Nyeriah:
Need to check if the remaining code doesn’t crash or cause issues after it’s removed (the variable is used below and will be dereferenced/null if there’s no enemy)

heyitsbench:
I did test it a bit and I could not get it to crash trying everything I observed causing the auto attack issue, nor did I find anything not working as (I believe) it should, but yea, if anyone else can give this a solid test, that'd be swell.

heyitsbench:
Tested once again, still could not find any issues and had no crashes.

PkllonG:
I'm using the Currently no players have found the bug

PR #19323, opened 2024-07-06
Sqlparser for modules

slavanorm:
## Changes Proposed:
small app to collect sql from modules into data/sql/custom and remember this fact.
configurable via cli, 0 dependencies.
has dry run which doesnt touch anything just showing proposed changes.

```
usage: sqlparser.py [-h] [-y [Y]] [--glob GLOB] [--repo_folder REPO_FOLDER]
                    [--source_suffix SOURCE_SUFFIX]
                    [--target_suffix TARGET_SUFFIX]
                    [--report_suffix REPORT_SUFFIX]
                    [{dry-run,run,None}]

Collect and copy SQL files to new folder, generate report. Skips collecting
pulled updates based on report.

positional arguments:
  {dry-run,run,None}    Simulate without changes OR run parser

options:
  -h, --help            show this help message and exit
  -y [Y]                Run without confirmation
  --glob GLOB           glob pattern to find scripts
  --repo_folder REPO_FOLDER
                        repository folder
  --source_suffix SOURCE_SUFFIX
                        location of modules folder, appends to repo_folder
  --target_suffix TARGET_SUFFIX
                        location of target folder, appends to repo_folder
  --report_suffix REPORT_SUFFIX
                        location of module results, appends to repo_folder
```

This PR proposes changes to:

## Issues Addressed:
- unclear way to import sql from scripts, mostly implying running sql by hand.

## Tests Performed:
This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]

### DIFF
--- a/apps/sqlparser/sqlparser.py
+++ b/apps/sqlparser/sqlparser.py
@@ -1,0 +1,167 @@
+import argparse
+import shutil
+import csv
+import sys
+
+from datetime import datetime
+from pathlib import Path
+from collections import Counter
+
+# constants, can be overriden using arguments
+
+## implied folder structure: modules/*/sql/dbname/*/*.sql
+glob = "**/sql/**/*.sql"
+
+repo_folder = Path("/Users/v0/Py/_hobby/wow/server")
+
+source_suffix = "modules"
+source_dir = lambda: repo_folder / source_suffix
+target_suffix = "data/sql/custom/test"
+target_dir = lambda: repo_folder / target_suffix
+report_suffix = "modules_sql_parsed.csv"
+report_path = lambda: repo_folder / report_suffix
+
+
+counters = dict(
+    written=0, skipped=0, total=0, modules=Counter()  # module: counter
+)
+
+
+def get_db_name(arg):
+    if arg.startswith("db"):
+        # assumes db_*.sql or db-*.sql filename
+        arg = arg[3:]
+    return f"db_{arg}"
+
+
+def collect_sql_files(full_fp, files_to_ignore=None, dry=False):
+    # fp means filepath
+    global counters
+
+    fp = full_fp.relative_to(source_dir())
+
+    module = fp.parents[-2]
+    path_to_parse = fp.parts[fp.parts.index("sql") + 1 :]
+    database = get_db_name(path_to_parse[0])
+    other_folders = "".join(path_to_parse[1:-1])
+
+    target_file = (
+        target_dir()
+        / database
+        / f"{module}_{counters['written']}_{other_folders}_{fp.parts[-1]}"
+    )
+    fp_out = target_file.relative_to(target_dir())
+
+    if files_to_ignore and str(fp) in files_to_ignore:
+        counters["skipped"] += 1
+        print(f"skipped existing module:{module} name:{fp_out} ")
+        return
+
+    if not dry:
+        # required by shutil.copy2
+        target_file.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(full_fp, target_file)
+
+    result = dict(
+        index=counters["total"],
+        module_index=counters["modules"][str(module)],
+        module=module,
+        database=database,
+        applied=datetime.now().isoformat(timespec="seconds"),
+        fp_in=fp,
+        fp_out=fp_out,
+        target_dir=target_dir(),
+    )
+    counters["written"] += 1
+    counters["total"] += 1
+    counters["modules"][str(module)] += 1
+    print(f"parsed module:{module} name:{fp_out} ")
+    return result
+
+
+def write_report(data):
+    with report_path().open("w") as csvfile:
+        writer = csv.DictWriter(csvfile, fieldnames=data[0].keys())
+        writer.writeheader()
+        writer.writerows(data)
+
+
+def read_report():
+    # reads report, updates counters, returns report and files_to_ignore
+    global counters
+    if not report_path().exists():
+        return None, None
+
+    with report_path().open() as csvfile:
+        reader = csv.DictReader(csvfile)
+        report = list(reader)
+
+    counters["total"] = len(report)
+
+    files_to_ignore = []
+    for e in report:
+        files_to_ignore.append(e["fp_in"])
+        counters["modules"][e["module"]] += 1
+
+    return report, files_to_ignore
+
+
+def main(dry=False):
+    report, files_to_ignore = read_report()
+    collected = [
+        collect_sql_files(e, files_to_ignore, dry)
+        for e in source_dir().glob(glob)
+    ]
+    collected = [e for e in collected if e is not None]
+    if not dry and collected:
+        write_report(collected)
+
+    print("report stats", counters)
+
+
+def argparsed():
+    global glob, repo_folder, source_suffix, target_suffix, report_suffix
+
+    default = ["--help"]
+    # default = "-y", "", "--glob", "test"
+    parser = argparse.ArgumentParser(
+        description="Collect and copy SQL files to new folder, generate report. Skips collecting pulled updates based on report."
+    )
+    parser.add_argument(
+        "option",
+        nargs="?",
+        choices=["dry-run", "run", None],
+        help="Simulate without changes OR run parser",
+    )
+    parser.add_argument("-y", nargs="?", help="Run without confirmation")
+
+    globals_to_parse = [
+        ["glob", "glob pattern to find scripts"],
+        ["repo_folder", "repository folder"],
+        ["source_suffix", "location of modules folder, appends to repo_folder"],
+        ["target_suffix", "location of target folder, appends to repo_folder"],
+        ["report_suffix", "location of module results, appends to repo_folder"],
+    ]
+
+    for e in globals_to_parse:
+        parser.add_argument(f"--{e[0]}", help=e[1])
+
+    args = parser.parse_args(args=None if sys.argv[1:] else default)
+
+    if args.option == "run":
+        prompt = "This script will copy SQL files. are you sure? [y/N]\n"
+        if input(prompt).lower() != "y":
+            print("Aborting")
+            return
+    elif args.option == "--help":
+        return
+    for e in globals_to_parse:
+        value = getattr(args, e[0])
+        if value:
+            globals()[e[0]] = value
+
+    main(args.option == "dry-run")
+
+
+if __name__ == "__main__":
+    argparsed()


slavanorm:
the idea was to bake this script into docker compose inbefore service applying updates.
i just dont know which service does it

hermensbas:
the code has references to external mods? As the ignore list. Isnt the core code independent from external mods and thereby any reference to external mods is unwanted?

slavanorm:
No references, only python builtins

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+    parser.add_argument("-y", nargs="?", help="Run without confirmation")
+
+    globals_to_parse = [
+        ["glob", "glob pattern to find scripts"],
+        ["repo_folder", "repository folder"],
+        ["source_suffix", "location of modules folder, appends to repo_folder"],
+        ["target_suffix", "location of target folder, appends to repo_folder"],
+        ["report_suffix", "location of module results, appends to repo_folder"],
+        ["modules_to_ignore", "list of modules to ignore"],
+    ]
```
michaeldelago, line 150: This is better of a list of tuples

```suggestion
    globals_to_parse = [
        ("glob", "glob pattern to find scripts"),
        ("repo_folder", "repository folder"),
        ("source_suffix", "location of modules folder, appends to repo_folder"),
        ("target_suffix", "location of target folder, appends to repo_folder"),
        ("report_suffix", "location of module results, appends to repo_folder"),
        ("modules_to_ignore", "list of modules to ignore"),
    ]
```

better yet though, I'm pretty sure you don't need to do whatever this is doing and you should just `parser.add_argument` for each one of these

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+        for e in source_dir().glob(glob)
+    ]
+    collected = [e for e in collected if e is not None]
+    if not dry and collected:
+        write_report(collected)
+
+    print("report stats", counters)
+
+
+def argparsed():
```
michaeldelago, line 127: missing docstring. Doesn't need anything fancy but something short and descriptive would be nice

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+    return report, files_to_ignore
+
+
+def main(dry=False):
+    report, files_to_ignore = read_report()
+    collected = [
+        collect_sql_files(e, files_to_ignore, dry)
+        for e in source_dir().glob(glob)
+    ]
+    collected = [e for e in collected if e is not None]
```
michaeldelago, line 120: Please use a more descriptive variable name than `e`

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+    if not dry and collected:
```
michaeldelago, line 121: an empty list is false-y, but it's shows the intent in the code a bit better to do `len(collected) == 0`

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+        write_report(collected)
```
michaeldelago, line 122: This might be a spot where it makes sense to pass dry as an argument to `write_report`. In `write_report`, if `dry == True`, it can print to stdout or something instead of to a file so the user can see what changes are going to be made.

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def get_db_name(arg):
+    if arg.startswith("db"):
+        # assumes db_*.sql or db-*.sql filename
+        arg = arg[3:]
+    return f"db_{arg}"
+
+
+def collect_sql_files(full_fp, files_to_ignore=None, dry=False):
```
michaeldelago, line 42: default arg for files_to_ignore can be an empty list so it's more clear as to what should be passed in

beware, common footgun with a list as a default argument is that all function invocations share the same list (as in, the pointer to that list), but as long as you're not doing anything to the list (adding/removing things) that's fine. 

```suggestion
def collect_sql_files(full_fp, files_to_ignore=[], dry=False):
```

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+    # fp means filepath
```
michaeldelago, line 43: > fp_means_filepath

it doesn't need to be abbreviated, you can just do `full_filepath` and so on

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+    global counters
+
+    fp = full_fp.relative_to(source_dir())
+
+    module = fp.parents[-2]
+    path_to_parse = fp.parts[fp.parts.index("sql") + 1 :]
+    database = get_db_name(path_to_parse[0])
+    other_folders = "".join(path_to_parse[1:-1])
```
michaeldelago, line 51: A little bit unclear what this block is doing. 

Short comment above it with the input and expected output would be appreciated.

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+target_suffix = "data/sql/custom"
+target_dir = lambda: repo_folder / target_suffix
+report_suffix = "modules_sql_parsed.csv"
+report_path = lambda: repo_folder / report_suffix
+
+modules_to_ignore = ["mod-playerbots"]
+
+counters = dict(
+    written=0, skipped=0, total=0, modules=Counter()  # module: counter
+)
```
michaeldelago, line 32: why `Counter()` only for `modules`? I assume since the `pathlib.Path` need to be hashed?

it might be more clear to split this out into a few variables instead of a dict:

```suggestion
sql_files_written = 0
sql_files_skipped = 0
modules_counter = Counter()
```

I think `total` could just be `written` + `skipped`? Not sure, worth a poke.

I probably sound like a broken record but globals can/should be avoided if possible

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def get_db_name(arg):
```
michaeldelago, line 35: parameter `arg` should be named something more descriptive.

apps/sqlparser/sqlparser.py
michaeldelago, line 1: Please add a `README.md` with a contrib disclaimer and instructions on how to use.

in the context of docker, ideally it'd be `docker compose -f docker-compose.yaml -f apps/sqlparser/compose.yaml ACTION` or adding an `include` statement into your `docker-compose.override.yml`

michaeldelago, line 1: Please add a `compose.yaml` with some comments 

Benefit of using stdlib only is that the container just needs to be the default `python:3` and then this file is bind-mounted in.

michaeldelago: It might be nice add some notes about linting/typechecking as well, such as pyright (my personal preference) or mypy, and the code formatter being used (personal preference is `black`, but having one in the first place is most important)

michaeldelago (changes requested):
Thanks for this contribution!

I'm a bit of a stickler for some of the things you can do with python, but the overall gist and flow of this script makes sense.

My main requested changes are
- don't use globals wherever possible, and instead rely on function parameters and return values
- ensure that relative type safety is maintained
- ensure that code is optimized for readability, ensuring the intent of the code is clear at all times.

apps/sqlparser/sqlparser.py
slavanorm: regarding compose yaml:
1) i just checked that none of apps in apps/* folder use such yaml.
2) i initially suggested to add this python code to db-import container.

do you still insist on making dedicated compose step for this?

is it possible to add this code into the normal compose routine of db-import?

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+# default_argparsed_args = "-y", ""  # , "--glob", "test"
+
+source_suffix = "modules"
+source_dir = lambda: repo_folder / source_suffix
+target_suffix = "data/sql/custom"
+target_dir = lambda: repo_folder / target_suffix
+report_suffix = "modules_sql_parsed.csv"
+report_path = lambda: repo_folder / report_suffix
+
+modules_to_ignore = ["mod-playerbots"]
```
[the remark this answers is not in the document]
slavanorm: it is included in CLI. i think all of CLI args should be also passable via env variables.

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+
+def argparsed():
+    global glob, repo_folder, source_suffix, target_suffix, report_suffix
+
+    default = ["--help"]
+    # default = "-y", "", "--glob", "test"
+    parser = argparse.ArgumentParser(
+        description="Collect and copy SQL files to new folder, generate report. Skips collecting pulled updates based on report."
+    )
+    parser.add_argument(
```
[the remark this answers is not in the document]
slavanorm: didnt comprehend this, you prefer dry-run over help as default?

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+        ["target_suffix", "location of target folder, appends to repo_folder"],
+        ["report_suffix", "location of module results, appends to repo_folder"],
+        ["modules_to_ignore", "list of modules to ignore"],
+    ]
+
+    for e in globals_to_parse[:-1]:
+        parser.add_argument(f"--{e[0]}", help=e[1])
+
+    e = globals_to_parse[-1]
+    parser.add_argument(f"--{e[0]}", help=e[1], nargs="+")
```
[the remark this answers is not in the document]
slavanorm: it does the same thing as in line 153, but for the last argument i wanted to add nargs="+", allowing modules to ignore to be a list of arguments

apps/sqlparser/sqlparser.py
slavanorm: comparing to dry as argument, 
this code is more explicit about app control flow and has less code lines count.

are you sure we need to make dry a write_report argument?

slavanorm: i think always printing is good thing for logging/debugging, thats why i didnt put it inside write_report func.
report is only the csv file

apps/sqlparser/sqlparser.py
slavanorm: default arg as list is not allowed by some PEP and flake, i dont see big reason to violate it here.
are you sure we need this change?

slavanorm:
i will also work to employ .conf file to conform to project apps structure

apps/sqlparser/sqlparser.py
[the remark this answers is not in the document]
slavanorm: im gonna remove this now

apps/sqlparser/sqlparser.py
slavanorm: this is used afterwards to read args too, thats why i decided to decrease boilerplate.

are we still required to drop this?

apps/sqlparser/sqlparser.py
slavanorm: regarding total, you are not 100% right.
total is pulled from report, and is basically an global index for all migrations.
its value is in the fact that you have a global index showing all migrations order.
therefore, once you have 1 entry in report, it will not match with written and skipped, which describe current script run.


all counters keys have very close purpose.
is there any value in splitting them? 
it will surely reduce readability and bloat the function annotations and returns.

apps/sqlparser/sqlparser.py
```diff
@@ -0,0 +1,178 @@
...
+        ["source_suffix", "location of modules folder, appends to repo_folder"],
+        ["target_suffix", "location of target folder, appends to repo_folder"],
+        ["report_suffix", "location of module results, appends to repo_folder"],
+        ["modules_to_ignore", "list of modules to ignore"],
+    ]
+
+    for e in globals_to_parse[:-1]:
+        parser.add_argument(f"--{e[0]}", help=e[1])
+
+    e = globals_to_parse[-1]
```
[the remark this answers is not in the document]
slavanorm: e is used as short for element, explicitly denoting absence of value of naming an item.
i'm open to use any other conventional name for such case, which one do you prefer?

slavanorm:
created readme.md, documented all issues without questions raised in this discussions.
resolved, committed.

apps/sqlparser/sqlparser.py
michaeldelago: I prefer this to have a separate compose file/service because I think this should ultimately be opt-in for the following reasons

1. What happens if this script ever breaks? Who is going to support and maintain it?
2. This repo doesn't require python for the build process. It can be a pain to require yet another software installation on an already large/heavy project. 
    1. A lot of people already have python installed already, and it's fantastic that this is stdlib only, but it's still adding another language and another dependency.
    2.  There are python scripts already in the repo, but to my understanding those aren't often executed/part of the build process
3. Installing the SQL from modules is generally quick, easy, and a one-time thing. 
    1. Installing SQL for modules in both docker and traditional installs has been only semi-automatic for a very long time.
4. One of the goals for the docker installation is to be as similar to the "traditional" install as possible. 
    1. Lots of people already see docker as complex magic, and adding extra steps to the docker build process can just compound on that perception. 
    2. Every time something new is added to the docker install that isn't added to the traditional install, this perception gets stronger.

I think this script looks really good, but I think in its current state and goals it should be managed in a separate repo rather than an enabled-by-default addition to this one.

PR #19358, opened 2024-07-08
chore(Scripts/Archimonde): removed unused variable

FrancescoBorzi:
The `result` variable seems not used (this is also triggering a new warning when compiling with clang 15 no-pch)


sudlud:
I already removed the unused variable yesterday with azerothcore/azerothcore-wotlk#19240

Issue #19429, opened 2024-07-18
Players are stuck in an orientation after melee attacks


noisiver:
### Current Behaviour

When attacking in melee, other players sees the meleeing character facing the wrong way. Before attacking it's fine, the character rotates however the player is rotating it, but after melee combat they get locked in a certain orientation while standing still. When they're movign it's fine, but when they stop it goes back to the locked rotation.

### Expected Blizzlike Behaviour

They should face whatever way they're actually facing when seen by other players.

### Steps to reproduce the problem

1. Create and log in two different characters on different accounts
2. Attack an enemy with one of the characters
3. After killing the mob, turn with the character that was attacking the mob while observing with the other character

### Extra Notes

After testing many commits, reverting 05f4fa1a351917ecd2e091064b812824750f1922 seems to fix it.

This is what the hunter sees
![image](https://github.com/user-attachments/assets/f1aa8e2a-1bad-4214-84fd-0baefac6be31)
and this is what the other player sees
![image](https://github.com/user-attachments/assets/807f80c3-221f-4805-8603-3d0805f111ee)

### AC rev. hash/commit

3c77defe78671892e71bab4ebb105ac08e9a2c09

### Operating system

Ubuntu 22.04

### Custom changes or Modules

None

noisiver:
Updated the title because this doesn't seem to be specific to hunters. I just tried meleeing as a shaman and it has the same effect.

noisiver:
The culprit is 05f4fa1a351917ecd2e091064b812824750f1922
I reverted the changes in that one and it's no longer an issue.

PR #19434, opened 2024-07-18
fix(Core): Correct packet structure for combat stop messages.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19429

## Tests Performed:

This PR:
- [X] Has not been tested.

## How to Test the Changes:

### DIFF
--- a/src/server/game/Handlers/CombatHandler.cpp
+++ b/src/server/game/Handlers/CombatHandler.cpp
@@ -86,10 +86,10 @@ void WorldSession::SendAttackStop(Unit const* enemy)
 {
     WorldPacket data(SMSG_ATTACKSTOP, (8 + 8 + 4)); // we guess size
     data << GetPlayer()->GetPackGUID();
-    uint8 nowDead = enemy->isDead();
 
     if (enemy)
     {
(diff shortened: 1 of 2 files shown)


Nyeriah (approved):

src/server/game/Handlers/CombatHandler.cpp
```diff
@@ -86,10 +86,12 @@ void WorldSession::SendAttackStop(Unit const* enemy)
 {
     WorldPacket data(SMSG_ATTACKSTOP, (8 + 8 + 4)); // we guess size
     data << GetPlayer()->GetPackGUID();
+    uint8 nowDead = enemy->isDead();
```
Nyeriah, line 89: This should probably be inside the IF below

Nyeriah: Because the if checks if enemy is null but this will use it before checking

heyitsbench: Good call, corrected.

Issue #19062, opened 2024-06-15
Add spawns missing AI delay


Alethrion:
### Current Behaviour

Practically all ads on the server spawn in and attack instantly, practically getting an auto-attack off before they've even phased in. (This may also apply to respawns in the open world, but I cannot find a source outside private servers for the expected behavior)

### Expected Blizzlike Behaviour

Many ads are supposed to have a short (0.5-1 second) delay on their ai. The bosses that I found with add spawning that have a delayed AI: Hydross, Nightbane, Curator, Aran, Midnight, Maybe Vashj.

### Source

Oof, gimme a second
Hydross: https://youtu.be/X4-VeHZ1DsE?t=39
Nightbane: https://youtu.be/D2jz0prp32o?t=90
Curator: https://youtu.be/M7K8HkxJ99o?t=810
Aran: https://youtu.be/gYANql-4mYo?t=90
Midnight: https://youtu.be/9gXlyEFhQg0?t=23
Vashj: https://youtu.be/MPTtcBScs9g?t=293

### Steps to reproduce the problem

1. Find a boss that spawns ads and meet the requirements for the ads to spawn
2. View behavior

### Extra Notes

This is definitely not an exhaustive list, but it seems to be fairly common. Cases can also be difficult to find sources for as when the mob is hit (such as by aoe) it's ai is instantly activated, hence cases can likely be found where the same mobs instantly act as well.

### AC rev. hash/commit

[85a2812](https://github.com/azerothcore/azerothcore-wotlk/commit/85a2812dbc31a131942a6a9633f828f1fedd6f62)

### Operating system

Windows 10 x64

Grimdhex:
Do we know situations where isn't the case ? This is important for the method being considered. Either we generalise this behaviour to all summons or it becomes an option to be set (where it's defined by a flag or something like that).

heyitsbench:
If the behavior *isn't* global across all summons, I'd personally argue the delay is more common than not, so if a flag were to be the decided implementation, I'd say it should be set to remove the delay, not add it.

Nyeriah:
Is there something already existing about this e.g summon properties?

If not flags extra might be the way to go, if there’s a bit available 

Mobs are static until the first DoZoneInCombat() call, so we just need to delay it in JustSummoned()

most mobs if not all are supposed be summoned by spells so there’s little room to change it in the script files themselves

Rorschach91:
We can add the Flames of Azzinoth in this report.

PR #19469, opened 2024-07-23
fix(Scripts/HyjalSummit): Adjust Doomfire Spirit movement logic.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Scripts (creature scripts).

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

- [X] Sniffs.
- [X] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.).

## Tests Performed:

This PR:
- [X] Has not been tested.

## How to Test the Changes:

1. Engage Archimonde.
2. Wait for Doomfire.
3. Observe its movement.


heyitsbench:
Doomfire sucks by the way.

src/server/scripts/Kalimdor/CavernsOfTime/BattleForMountHyjal/boss_archimonde.cpp
```diff
@@ -145,32 +145,29 @@ struct npc_doomfire_spirit : public ScriptedAI
...
+            float dist = 8.f;
 
-        fAngle = urand(0, M_PI * 2);
-    }
+            while (dist >= 0)
+            {
+                pos = me->WorldObject::GetFirstCollisionPosition(dist, nextOrientation);
+                if (fabsf(dist - me->GetExactDist2d(pos)) < 0.001) // Account for small deviation
+                    break;
+                dist -= 2.0; // Distance drops by two units with each unsuccessful attempt
```
Kitzunu, line 163:
```suggestion
                dist -= 2.0f; // Distance drops by two units with each unsuccessful attempt
```

Gultask:
does this conflict or make https://github.com/azerothcore/azerothcore-wotlk/pull/19467 obsolete in any way? :pepeinspect:

heyitsbench:
> does this conflict with or make 19467 obsolete in any way?

Not that I can discern, no.

Tereneckla:
> does this conflict with or make #19467 obsolete in any way? :pepeinspect:

That one is the starting direction and this one is behavior around obstacles

heyitsbench:
So the way Doomfire seems to work is like this (just to summarize a bunch of things learned yesterday):
Doomfire and Doomfire Spirit both spawn at a given time, the Spirit might face Archimonde with its initial spawn.
~0.01s later, the Spirit does a 1y teleport away and has its orientation flipped 180 degrees.
~0.01s later, the Spirit then does an 8y teleport in the same style as its normal teleports.
~1.6s later, the Doomfire takes on its aura to create Doomfire area aura patches on the ground and begins following the Doomfire Spirit. At the same time, the Spirit teleports again in its usual fashion.

The way the teleport works is the Spirit seems to pick a random direction, either straight, 45 degrees left, or 45 degrees right of its current orientation. The selection of directions does not seem to be weighted based on the sample size we have. Once a direction is selected, it does a check in that direction with a distance of 8y if the coordinates are clear of obstruction. If it's successful, it teleports. If not, it checks again at 6y, then 4y, then 2y. If all three fail, then the Doomfire will still teleport, but nothing about its position nor orientation will change. Its subsequent teleport(s) will still pick direction based off the unchanged orientation.

After this PR, the Doomfire does not wait to "wake up" and take on its aura, nor do the first two Spirit teleports occur, only at 1.6s forward. The Doomfire issue is, in my opinion, [a symptom of larger missing behavior](https://github.com/azerothcore/azerothcore-wotlk/issues/19062), but the Spirit issue is likely easy to solve with some scheduling stuff.

heyitsbench:
After the recent commits (thanks to Tereneckla), the first two Spirit teleports now are occurring. A misuse of `WorldObject::GetFirstCollisionPosition` was also corrected. The Doomfire itself still does not wait to start acting, but looking roughly at the fire, it does seem to exhibit closer behavior to what I've seen in videos and on Classic.

Nyeriah (approved):

Issue #19500, opened 2024-07-28
[character_achievement_progress] Segmentation fault.


manstfu:
### Current Behaviour

Criteria `12630` (not only) in table `character_achievement_progress` causing a segmentation fault (world crash).
https://gist.github.com/manstfu/6c8e861b60101dc38a48d23506af8f16
https://gist.github.com/manstfu/c112ea620594127d2d5871597a9ebef8
### Expected Blizzlike Behaviour

.

### Steps to reproduce the problem

Set criteria `12630` to any character in table `character_achievement_progress`

### AC rev. hash/commit

AzerothCore rev. 17d61664fd6a+ 2024-07-25 23:52:41 +0000 (master branch) (Unix, RelWithDebInfo, Static)

### Operating system

Debian 12

PR #19504, opened 2024-07-28
fix(Core/AchievementMgr): Check for null achievement before being derefenced.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Core (Game systems).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19500

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/src/server/game/Achievements/AchievementMgr.cpp
+++ b/src/server/game/Achievements/AchievementMgr.cpp
@@ -2473,13 +2473,15 @@ bool AchievementGlobalMgr::IsStatisticAchievement(AchievementEntry const* achiev
 
 bool AchievementGlobalMgr::IsAverageCriteria(AchievementCriteriaEntry const* criteria) const
 {
-    if (!sAchievementStore.LookupEntry(criteria->referredAchievement))
+    auto referencedAchievement = sAchievementStore.LookupEntry(criteria->referredAchievement);
+
+    if (!referencedAchievement)
         return false;
 
-    if ((sAchievementStore.LookupEntry(criteria->referredAchievement))->flags & ACHIEVEMENT_FLAG_AVERAGE)
+    if (referencedAchievement->flags & ACHIEVEMENT_FLAG_AVERAGE)
         return true;
 
-    if (AchievementEntryList const* achRefList = GetAchievementByReferencedId(criteria->referredAchievement))
+    if (AchievementEntryList const* achRefList = GetAchievementByReferencedId(referencedAchievement->ID))
         for (AchievementEntryList::const_iterator itr = achRefList->begin(); itr != achRefList->end(); ++itr)
             if ((*itr)->flags & ACHIEVEMENT_FLAG_AVERAGE)
                 return true;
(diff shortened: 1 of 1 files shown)


AnchyDev (commented):
It's possible criteria can be a nullptr as well.

src/server/game/Achievements/AchievementMgr.cpp
```diff
@@ -2473,6 +2473,9 @@ bool AchievementGlobalMgr::IsStatisticAchievement(AchievementEntry const* achiev
 
 bool AchievementGlobalMgr::IsAverageCriteria(AchievementCriteriaEntry const* criteria) const
 {
+    if (!sAchievementStore.LookupEntry(criteria->referredAchievement))
+        return false;
+
     if ((sAchievementStore.LookupEntry(criteria->referredAchievement))->flags & ACHIEVEMENT_FLAG_AVERAGE)
         return true;
```
Kitzunu, line 2481: instead of calling it twice you could do

```cpp
auto referencedAchievement = sAchievementStore.LookupEntry(criteria->referredAchievement);

if (!referencedAchievement)
    return false;
    
if (referencedAchievement->flags & ACHIEVEMENT_FLAG_AVERAGE)
    return true;

**Review by Kitzunu — APPROVED** [review]

File: src/server/game/Achievements/AchievementMgr.cpp (hunk 1 above)
Comment (heyitsbench) [reply src/server/game/Achievements/AchievementMgr.cpp]: Corrected, also made use of the variable in the if statement below.

**Comment by heyitsbench** [comment]
> It's possible criteria can be a nullptr as well.

Maybe I'm wrong, but looking at the parts of code that use this function, it doesn't *look* like they'd pass in a nullptr (i.e. being dereferenced already or being contained in CriteriaProgressMap, hopefully with correct data).

**Comment by sudlud** [comment]
looks like a total of 161 achievement criteria refer to a non-existing achievement
```sql
select * from achievement_criteria_dbc acd where acd.Achievement_Id not in
(
    select ad.ID from achievement_dbc ad where 1
)
```
(obviosly requires sql import of the relevant dbc filles)

AnchyDev:
> > It's possible criteria can be a nullptr as well.
> 
> Maybe I'm wrong, but looking at the parts of code that use this function, it doesn't _look_ like they'd pass in a nullptr (i.e. being dereferenced already or being contained in CriteriaProgressMap, hopefully with correct data).

In `AchievementMsg::BuildAllDataPacket` there is some code that looks up an entry and passed it directly to `IsAverageCriteria` with no care for whether the store returned a nullptr or not. It is not de-referenced before this point.

```cpp
        if (sAchievementMgr->IsAverageCriteria(sAchievementCriteriaStore.LookupEntry(iter->first)))
            *data << uint32(now - GetPlayer()->GetCreationTime().count());    // for average achievements
        else
            *data << uint32(now - iter->second.date);
```
I just figured it would have been worth putting a check in there just in case a rare error occurs where the criteria does not exist in the store (failed to load from DBC) but the player still has the criteria in database. It may not really be the responsibility of IsAverageCriteria to check for the nullptr here anyway, but no matter it's merged now :smile:

Issue #14367, opened 2022-12-25
Most (every?) AoE ranges are wrong (too high)


Rdyx:
### Current Behaviour

AoEs are hitting further than they should. Radius ranges seems correct from DBC. Don't know if it's server or client related.
i.e: holy nova should heal/hit in a 10y range. Yet, when checking with .dist, got 11.33y (example), then casting holy nova heal/hit.

### Expected Blizzlike Behaviour

AoEs should not hit above determined range.

### Source

![image](https://user-images.githubusercontent.com/29793773/209460973-7b88fac3-bea0-4972-acbc-6ae76ac03da5.png)
![image](https://user-images.githubusercontent.com/29793773/209460983-aa2e2321-f9d5-48b9-961a-5ceec4e8bc58.png)

### Steps to reproduce the problem

1. Learn any AoE spell
2. Get someone to be able to test (either friendly or ennemy, depending on the used spell)
3. Check range with .dist while targetting your buddy
4. Get a little above the determined range told in the tooltip (i.e: 10y in the tooltip, get somewhere around 11y)
5. Use your spell, check if it heal/hits

### Extra Notes

Reducing all ranges in SpellRadius.dbc (iirc) by 1-1.5y could do the trick but I think it would not be a real proper solution as it could lead into unintended too small radius for some other AoEs that are working properly maybe (but found none).

### AC rev. hash/commit

AzerothCore rev. d0561c56df19+ 2022-07-26 14:29:17 -0300 (master branch) (Unix, Release, Static)
Using SSL version: OpenSSL 1.1.1f  31 Mar 2020 (library: OpenSSL 1.1.1f  31 Mar 2020)
Using Boost version: 1.71.0
Using MySQL version: 80031
Using CMake version: 3.16.3
Compiled on: Linux 5.15.0-1023-aws
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in ./maps. Total size: 291014951 bytes
vmaps directory located in ./vmaps. Total size: 658130721 bytes
mmaps directory located in ./mmaps. Total size: 2320900676 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 
Using World DB: ACDB 335.7-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
[...]

PR #19529, opened 2024-07-30
fix (Core/Spells) remove object size for hit calculation of aoe aura spells that target enemies

Tereneckla:
## Changes Proposed:

This PR proposes changes to:

The check that starts in GridNotifiers.h is used by by all kinds of aoe aura spells that target enemies (persistent ground (consecration), npc aura(totem of wrath)
ground targeted aoe (blizzard) also passes it but the check is not used???

the one in Spell.cpp then is used by spells such as blizzard

## Issues Addressed:

- Progress #14367

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

This only deals with the spells where players easily see a mismatch and get annoyed. Every type of spell should be checked if they should or should not include caster/target object size

Ground targeted aoe goes through two checks for some reason

These functions need a refactor. On some the base function does calculation with ObjectSize and then another function removes Size to do a check with that. Some are the other way around.
The names also are not very helpful

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -9012,7 +9012,7 @@ namespace Acore
             if (!target->ToGameObject()->IsInRange(_position->GetPositionX(), _position->GetPositionY(), _position->GetPositionZ(), _range))
                 return false;
         }
-        else if (!target->IsWithinDist3d(_position, _range))
+        else if (!target->IsInDist(_position, _range))
             return false;
         else if (target->GetTypeId() == TYPEID_UNIT && target->ToCreature()->IsAvoidingAOE()) // pussywizard
             return false;
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -1102,7 +1102,7 @@ namespace Acore
 
             }
 
-            if (i_funit->_IsValidAttackTarget(u, _spellInfo, i_obj->GetTypeId() == TYPEID_DYNAMICOBJECT ? i_obj : nullptr) && i_obj->IsWithinDistInMap(u, i_range))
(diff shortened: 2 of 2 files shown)


src/server/game/Spells/Spell.cpp
```diff
@@ -9012,7 +9012,7 @@ namespace Acore
             if (!target->ToGameObject()->IsInRange(_position->GetPositionX(), _position->GetPositionY(), _position->GetPositionZ(), _range))
                 return false;
         }
-        else if (!target->IsWithinDist3d(_position, _range))
+        else if (!target->IsInDist(_position, _range))
```
Grimdhex, line 9015: was it really necessary? Normally this shouldn't change anything, as IsWithin3d simply calls IsInDist by adding GetCombatReach to the range. But GetCombatReach is normally only overridden in Unit.cpp and should be equal to 0.0f  here (default value)

Tereneckla: IsWithinDist3d adds GetObjectSize which has no overrides anywhere

https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Object/Object.cpp#L1295

Nyeriah (approved):

src/server/game/Spells/Spell.cpp
Grimdhex: Yes, I remembered TC, but I didn't necessarily think that the function was different on AC (and it's normal because it's an old version of TC). In researching the reason for the change I discovered that the outcome appears to be much more complex. I'm not saying that TC's fix is necessarily the right one, but I don't think that simply replacing "IsWithinDist3d" by "IsInDist"

source: https://github.com/TrinityCore/TrinityCore/pull/16290 and https://github.com/WoWManiaUK/Redemption/issues/3061

Tereneckla: Yes, all the functions around Distance can use a refactor

PR #19554, opened 2024-08-04
fix(Core/GameObject) Ritual of Summoning

### LINKED ISSUE #17012: (Spells): Ritual of Summoning.
### Current Behaviour

1. The portal created by the warlock is at the same position as the warlock.
2. Players who click the warlock's portal do not play a casting animation.
3. The ritual is completed immediately when the third person clicks the portal.
4. There is a cast time to create portals.
5. The teleported player has no change to their orientation.
6. Only one other person can click the portal.

1. It should be a few yards in front of the warlock when they cast the spell.
2. They should play the same animation as the warlock when they've clicked the portal.
3. [There should be a small delay between when the third person clicks (they still play an animation) and when the closet is created](https://youtu.be/DhEIJi5Bgfo?t=101).
3a. The ritual can be interrupted if the third player (likely the second as well, didn't check) clicks the portal (plays the animation) and steps away before it is completed. This does not consume a soul shard from the warlock.
4. Creating portals to summon players should be instant.
5. The teleported player should have their orientation changed to match the warlock when they clicked the closet/created the portal.
6. While not consistent (and likely not intended,) [it is sometimes possible for a third person to click the portal and play the casting animation](https://youtu.be/DhEIJi5Bgfo?t=47).

### Source

Wrath Classic primarily, light double checking from retail videos of the time

1. `.learn 698`
2. Play around with the ritual.
3. Observe the results.

### Extra Notes

Split from https://github.com/azerothcore/azerothcore-wotlk/issues/17010.

### AC rev. hash/commit

4f015d09678038be49e9d12114c3147e499d82ef

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

Saqra1:
- Ritual of Summoning object should spawn in front of the caster
- Clicking on the object should play an animation
  - There is a delay after the third player clicks before the closet is created as expected
- Opening a summoning portal should be instant

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes #17012 
- Closes https://github.com/chromiecraft/chromiecraft/issues/5346

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -198,10 +198,23 @@ void GameObject::RemoveFromWorld()
     }
 }
 
+SummoningRitual GameObject::GetSummoningRitualInfo()
+{
+    SummoningRitual summoningRitual = GetGOInfo()->summoningRitual;
+    if (summoningRitual.spellId == 62330)
+    {
+        summoningRitual.spellId = 61993;
+        summoningRitual.animSpell = 32783;
+    }
+    return summoningRitual;
+}
+
 void GameObject::CheckRitualList()
 {
     if (m_unique_users.empty())
         return;
+
+    uint32 animSpell = GetSummoningRitualInfo().animSpell;
 
     for (GuidSet::iterator itr = m_unique_users.begin(); itr != m_unique_users.end();)
     {
@@ -214,7 +227,7 @@ void GameObject::CheckRitualList()
         bool erase = true;
         if (Player* channeler = ObjectAccessor::GetPlayer(*this, *itr))
             if (Spell* spell = channeler->GetCurrentSpell(CURRENT_CHANNELED_SPELL))
-                if (spell->m_spellInfo->Id == GetGOInfo()->summoningRitual.animSpell)
+                if (spell->m_spellInfo->Id == animSpell)
                     erase = false;
 
         if (erase)
@@ -226,7 +239,7 @@ void GameObject::CheckRitualList()
 
 void GameObject::ClearRitualList()
 {
-    uint32 animSpell = GetGOInfo()->summoningRitual.animSpell;
+    uint32 animSpell = GetSummoningRitualInfo().animSpell;
     if (!animSpell || m_unique_users.empty())
         return;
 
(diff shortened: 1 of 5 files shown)


data/sql/updates/pending_db_world/rev_1722736427173275700.sql
```diff
@@ -0,0 +1,2 @@
+--
+UPDATE `gameobject_template` SET `Data1` = 61993, `Data2` = 32929 WHERE `entry` = 194108;
```
heyitsbench, line 2: This change is incorrect. The values for this gameobject are correct in current latest master AC according to a 53788 sniff, so they cannot be changed.

Saqra1: I see, what would be the best place to override these values?

heyitsbench: IMO they shouldn't be overridden.

Saqra1: Currently one of those values is overridden in code, at GameObject::Update, and it is necessary for the summon ritual to work.
Data1 which gets mapped to spellId contains a value for a spell that does not seem to exist, 62330.

heyitsbench: > [...] contains a value for a spell that does not exist, 62330.

The spell it points to is [a serverside spell](https://wowgaming.altervista.org/aowow/?spell=62330). Do we know what it does? No, but that's what is used and it shouldn't be overridden. Maybe a more knowledgeable contributor has a better idea or disagrees with me, but that is my opinion.

Saqra1: Ok, I agree not to change this value in the database, I still need to learn a lot about the project and codebase.

So, how can I fix this?

What I found while investigating the issue is that the server checks specifically for that spell id (`if (spellId == 62330)`) and replaces the value to cast `61993` instead. This is how it currently works, how the server handles that serverside spell, it casts something else. Whether is the right way or not, I am not sure. Is there a similar scenario that I could take a look at?

heyitsbench: You didn't pick a simple issue to try to resolve. 😛

It looks to me like the 62330 casts should be scripted out in some manner. According to a sniff, it appears that the characters assisting with the ritual (i.e. clicking the portal) are channeling the 698 spell alongside the warlock until it is satisfied with the player count clicking the portal.

As well, I believe [this line](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/GameObject/GameObject.cpp#L1921) replaces the spell used for the 'closet' itself to the wrong one. It appears that it should also be `23598`, like regular meeting stones.

Saqra1: What I ended up doing is moving these values into code. I don't think we would be worse than before, compared to the current implementation. It is an small improvement, it behaves as expected.

Saqra1:
I believe I found a better approach.

Issue #18042, opened 2023-12-24
Eclipsion Dragonhawk's fire breath hitting harder than it should


Stump876:
Original Cromiecraft report https://github.com/chromiecraft/chromiecraft/issues/6391 Confirmed on my local 

### What client do you play on?

enUS

### Faction

Both

### Content Phase:

70

### Current Behaviour
The [Eclipsion Dragonhawk](https://wowgaming.altervista.org/aowow/?npc=20502), located in Eclipse Point, Shadowmoon Valley, is dealing an extreme amount of damage with their [Fire Breath](https://wowgaming.altervista.org/aowow/?spell=38309) spell.
Witnessed in person on two separate characters. Two images of the combat log are included.
![eclipsiondragonhawkbugreportexample1](https://github.com/chromiecraft/chromiecraft/assets/153065400/155f9eda-f0c0-426f-aa88-e2b3e1131d6f)
![eclipsiondragonhawkbugreportexample2](https://github.com/chromiecraft/chromiecraft/assets/153065400/e2212d71-a312-4083-b970-038fe126ee90)

**Tester notes:** Current spell used is 38309 tooltip is stated as below.
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/68164076/fe931c57-2dab-499d-b0a9-9f159b61d09a)
I don't believe this is the correct spell. The tooltip also looks to be unfinished, and there is no referance to this spell being used by anything
https://www.wowhead.com/spell=38309/fire-breath
https://wowgaming.altervista.org/aowow/?spell=38309

### Expected Blizzlike Behaviour

Several sources have spell 37985 listed as the ability used by this mob
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/68164076/94dddd5e-96bf-4a5e-b1ff-f142bceced6d)

### Source
https://www.wowhead.com/npc=20502/eclipsion-dragonhawk#abilities
https://wowgaming.altervista.org/aowow/?spell=37985#used-by-npc

### Steps to reproduce the problem

1. .go c 72587
2. engage in combat
3. Wait until it casts Fire Breath
4. Pull out marshmallows
5. Check combat logs for the spell used

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/123bf37ee7e6763d9d7f4b778cfe8453577bce6c

### Operating system

Ubuntu 20.04

### Modules

[...]

PR #19605, opened 2024-08-13
fix(Core/SpellInfo): Disallow damage scaling trigger spells to get scaled twice.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18042
- Closes https://github.com/chromiecraft/chromiecraft/issues/6391
- Closes https://github.com/chromiecraft/chromiecraft/issues/7263

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:


heyitsbench:
So this kinda works. This solves the double scaling issue for spells like Eclipsion Dragonhawk's Fire Breath and Ethereum Archon's Energy Flux, but in the case of Wrath Speaker's Rain of Chaos, this breaks it, it seems in the situation where both a trigger spell and its triggered spell have the scale based on caster level, only the trigger spell actually gets scaled, but in the case of a trigger spell having the attribute but its triggered spell not having the attribute, the final damage should still be scaled. Not sure how to fix this yet.

heyitsbench:
Tested with an included double scaling evaluation, all the spells previously mentioned seem to be doing much more accurate damage.

Nyeriah (approved):

PR #19740, opened 2024-08-25
fix(Core/Handlers): Only remove emote if player is moving.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Core (Handlers).

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

- [X] Live research (checked on live servers, i.e. Classic Cataclysm)

## Tests Performed:

This PR has been:

- [ ] This pull request doesn't entirely fix this issue, but it does improve it. Currently, players cannot turn while doing emotes. With this PR, you can turn while emoting, but it seems only with the mouse, using the turn keys still breaks the emote. As well, this applies to all emotes equally. It shouldn't, at least according to Cata Classic testing. E.g. /clap should not break with any movement (turn, walk, jump), but /dance should break with actual positional movement, but not any sort of turning. Requires more thorough testing, but not something I'm confident matches retail Wrath nor do I really want to do that testing at this moment. If/when this PR is merged, I'll make an issue to remind myself to do it at some point.

### DIFF
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -373,7 +373,7 @@ void WorldSession::HandleMovementOpcodes(WorldPacket& recvData)
     // Stop emote on move
     if (Player* plrMover = mover->ToPlayer())
     {
-        if (plrMover->GetUInt32Value(UNIT_NPC_EMOTESTATE) != EMOTE_ONESHOT_NONE)
+        if (plrMover->GetUInt32Value(UNIT_NPC_EMOTESTATE) != EMOTE_ONESHOT_NONE && movementInfo.HasMovementFlag(MOVEMENTFLAG_MASK_MOVING))
         {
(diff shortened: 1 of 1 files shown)


heyitsbench:
So I've done some experimentation on Cata Classic and I've got some findings. First, there are two main categories of emotes, being emotes and stand states. Stand states including things like `/sleep`, `/sit`, `/kneel`, and emotes including almost everything else pretty much. Emotes are not supposed to be interrupted by anything movement wise it seems, not by turning, walking, or jumping. Stand states get interrupted by walking and jumping, but not by turning, nor do they allow the player to actually rotate. 

The two outliers to these categories are `/dance` and `/beg`. Both of these are not interrupted by turning but *do* allow the player to rotate, but they are interrupted by jumping and walking. What complicates this further is that `/dance` clearly uses an emote state according to a sniff, but `/beg` only uses the emote opcode. I've checked the DBCs, and I can't see anything obvious that's specific to these two that would explain this behavior. Still TBD whether this behavior is unchanged from retail Wrath.

PR #19749, opened 2024-08-26
Talentpoints hook

ianmeesters:
## Changes Proposed:
Added hook in Player.CalculateTalentsPoints() function to be able to overwrite the "talentPointsForLevel" variable. resulting in being able to make custom talent points distributions throughout leveling. 

This PR proposes changes to:
-  [x ] Core (units, players, creatures, game systems).

## Tests Performed:

This PR has been:

## How to Test the Changes:
For an easy to add module i grabbed the mod-random-enchants module since it already uses some hooks. so i just added the one i created to the random_enchants.h file in the public function declerations:
```
void OnCalculateTalentsPoints(Player const* player, uint32& talentPointsForLevel) override; 
```
then at the bottom of the random_enchants.cpp file implemented the hook as such:
```
void RandomEnchantsPlayer::OnCalculateTalentsPoints(Player const* player, uint32& talentPointsForLevel)
{
    talentPointsForLevel = player->GetLevel();
}
```
it compiled and runs fine, i created a character and set it to lvl 20. and as expected the available talent points are set to 20. 

I also tested to make sure nothing broke without the module active. but with my hooks in place. And when i logged on the talent points were set to 11 which is the correct default if the talentPointsForLevel var isn't altered.

Added info:
source random enchants module i used to implement newly created hook: https://github.com/azerothcore/mod-random-enchants

i am part of the discord server: amyfang.
please feel free to contect me if there are any questions!

added context:
i plan on making DK a playable class starting from lvl 1, this is still a long ways to go, but this hook is needed to alter the talent points calculation since there is a specific bit for deadknights in the calculation that would not make talentpoints function below lvl 55. ALSO quite a while ago i was running a server with my buddies and we altered the talent points to give extra points every few levels to make the leveling experience snappy and more fun/powerfull. I know there is also a config flag to set rate, but with this hook there are so many more fun things that can be implemented.

thanks for your time, love the project!

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -13584,6 +13584,7 @@ uint32 Player::CalculateTalentsPoints() const
     }
 
     talentPointsForLevel += m_extraBonusTalentCount;
+    sScriptMgr->OnPLayerCalculateTalentsPoints(this, talentPointsForLevel);
     return uint32(talentPointsForLevel * sWorld->getRate(RATE_TALENT));
 }
 
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -13584,7 +13584,7 @@ uint32 Player::CalculateTalentsPoints() const
     }
 
     talentPointsForLevel += m_extraBonusTalentCount;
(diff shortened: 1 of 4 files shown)


ianmeesters:
i created my own module which can be used to clone and test the new hook:

https://github.com/ianmeesters/lvl1DK

sudlud (approved):

Issue #18442, opened 2024-02-28
(Core/Pooling): Creatures in a pool of their own don't seem to respawn.


heyitsbench:
### Current Behaviour

Currently, [Voidhunter Yar](https://wowgaming.altervista.org/aowow/?npc=18683) does not respawn once killed. This creature only has one spawn location and it is in a pool on its own.

### Expected Blizzlike Behaviour

He should admittedly not be in a pool, but even in a pool of their own, a creature should still be able to respawn.

### Source

Assumption.

### Steps to reproduce the problem

1. `.go c id 18683`
2. Kill him.
3. Wait twelve hours.
4. Observe he is not respawned.

### Extra Notes

Originally reported at https://github.com/chromiecraft/chromiecraft/issues/6689.

This creature has a `spawntimesecs` of twelve hours. Once killed, this creature *did* have an entry made in `characters.creature_respawn` with an epoch time ~twelve hours after it was killed (a little more than twelve, but who's counting). Once the time was passed, the entry was removed, but the creature was not spawned. This was also tested with the `spawntimesecs` changed to fifteen, which had similar results. Once the creature was removed from the pool, he spawned fine (still with `spawntimesecs` set to fifteen cause I'm not waiting another twelve hours).

### AC rev. hash/commit

b9c1dc95dd8f150a77673d9f23235b235dd700e4

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
A list of creature pools in a similar situation:
```sql
INSERT INTO `pool_template` (`entry`, `max_limit`, `description`) VALUES
(1028, 1, 'Gilmorian (14447)'),
(1042, 1, 'Teremus the Devourer (7846)'),
(1087, 1, 'Fumblub Gearwind Rare Spawn (1 out 1)'),
(1089, 1, 'Crazed Indu le Survivor Rare Spawn (1 out 1)');

INSERT INTO `pool_creature` (`guid`, `pool_entry`, `chance`, `description`) VALUES
(48204, 1028, 0, 'Gilmorian (14447) - Spawn 1'),
(134380, 1042, 0, 'Teremus the Devourer (7846) - Spawn 1'),
(151938, 1087, 100, 'Fumblub Gearwind - Spawnlocation 1'),
(151942, 1089, 100, 'Crazed Indu le Survivor - Spawnlocation 1');
```

[...]

PR #19750, opened 2024-08-26
fix(Core/Pool): Fix degradation of pools over time

walkline:
## Changes Proposed:

We are currently experiencing an issue with the pool system where a pool may fail to spawn an object when it should. This causes the pool to become inactive until the server is restarted. On servers with long uptimes, this could result in maps without active pools, leading to missing mining and herbalism nodes.

This problem arises in the following scenario:

Consider a pool group with the following objects and spawn chances:
* ObjectA: 75% chance
* ObjectB: 15% chance
* ObjectC: 10% chance

If the last spawned object was ObjectC and the roll results in ObjectC again, the logic here https://github.com/azerothcore/azerothcore-wotlk/blob/83f8871c31fd2b9f374587f921c11b2e00cbf733/src/server/game/Pools/PoolMgr.cpp#L327-L338 results in an empty rolledObjects list. As a consequence, no new objects spawn, but the old one despawns, effectively leaving the pool group in a bugged state.

Another related issue is that the current logic prevents the same object from being spawned consecutively in a pool group. For instance, if ObjectA (with a 75% chance) is spawned and collected, the next spawn will have ObjectB with a 90% chance and ObjectC with a 10% chance. This behavior is suboptimal and does not align with expected spawning dynamics.

The suggested changes address both of these issues.

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Pools/PoolMgr.cpp
+++ b/src/server/game/Pools/PoolMgr.cpp
@@ -330,7 +330,7 @@ void PoolGroup<T>::SpawnObject(ActivePoolData& spawns, uint32 limit, uint32 trig
 
                 // Triggering object is marked as spawned at this time and can be also rolled (respawn case)
                 // so this need explicit check for this case
-                if (roll < 0 && (/*obj.guid == triggerFrom ||*/ !spawns.IsActiveObject<T>(obj.guid)))
+                if (roll < 0 && (obj.guid == triggerFrom || !spawns.IsActiveObject<T>(obj.guid)))
                 {
                     rolledObjects.push_back(obj);
                     break;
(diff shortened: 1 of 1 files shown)


sudlud:
how does this relate to single-spawn pools? these also had the problem thad they would never respawn as there's only one spawn inside with 100%

walkline:
I didn't investigate that part, but maybe it would solve this issue as well - https://github.com/azerothcore/azerothcore-wotlk/issues/18442

Nyeriah (approved):

heyitsbench:
Tested this, it does in fact resolve #18442, though I didn't test for degradation of other pools.

walkline:
By the way, for my tests, I created a command to rewind time. You can create a macro (`.debug timeff 1200` rewinds time for 20 minutes) along with `.cheat casttime` to grind a node until it's bugged. I believe this can be useful for testing.

https://gist.github.com/walkline/1382a2542cb2e2ba5e1dd64a710d7c6e

Issue #13330, opened 2022-10-06
Triggered Spells put you into combat


BeefSteaker:
original issue https://github.com/chromiecraft/chromiecraft/issues/3905. confirmed on ptr
### Current Behaviour

Triggered spells, like Mage Ignite put both sides into combat when they trigger
This leads to a very long combat timer when a mage has living bomb on a target and it crits

### Expected Blizzlike Behaviour

Triggered Spells have no influence on the combat timer

### Source

https://youtu.be/qavkjivhSkQ?t=696

Here you see the mage leaving combat with a crit of living bomb causing ignite to trigger 1 second ago

Proof that he has the living bomb glyph (ignite on the new target that had a previous casted living bomb on):
https://youtu.be/qavkjivhSkQ?t=436

### Steps to reproduce the problem

    Get 2 characters, 1 of them mage
    On the mage learn ignite and living bomb
    .add 18970 2
    .add 45737
    Equip those rings
    Learn the Glyph
    Duel
    Cast Living Bomb
    Look at the combat status

### Extra Notes

https://wowgaming.altervista.org/aowow/?spell=44457
https://wowgaming.altervista.org/aowow/?spell=12848

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/e6dd5d3e22d92dddc841079f9108275fd0359be1

### Operating system

Ubuntu 20.04

### Custom changes or Modules

[mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
    [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
    [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
    [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
    [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
    [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
    [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
    [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
    [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
    [mod-eluna](https://github.com/azerothcore/mod-eluna)
    [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
    [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
[...]

PR #19819, opened 2024-08-31
fix(Core/Spells): Resolve a case of triggered spells prolonging combat.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/13330

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -20189,13 +20189,13 @@ void Unit::OutDebugInfo() const
 class AuraMunchingQueue : public BasicEvent
 {
 public:
-    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId) { }
+    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId, AuraEffect* aurEff) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId), _aurEff(aurEff) { }
 
     bool Execute(uint64 /*eventTime*/, uint32 /*updateTime*/) override
     {
         if (_owner.IsInWorld() && _owner.FindMap())
             if (Unit* target = ObjectAccessor::GetUnit(_owner, _targetGUID))
-                _owner.CastCustomSpell(_spellId, SPELLVALUE_BASE_POINT0, _basePoints, target, TriggerCastFlags(TRIGGERED_FULL_MASK & ~TRIGGERED_NO_PERIODIC_RESET), nullptr, nullptr, _owner.GetGUID());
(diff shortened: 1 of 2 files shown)


Helias (approved):

kissingers:
@PkllonG Why you Revert the PR? It seems case crash at my server, are you same?

LannyE:
It appears this commit causes ACCES_VIOLATION crashes.  For reference my issue ticket on the Trinity-bots GitHub page https://github.com/trickerer/Trinity-Bots/issues/876.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -20221,9 +20223,9 @@ void Unit::CastDelayedSpellWithPeriodicAmount(Unit* caster, uint32 spellId, Aura
 
     // xinef: delay only for casting on different unit
     if (this == caster || !sWorld->getBoolConfig(CONFIG_MUNCHING_BLIZZLIKE))
-        caster->CastCustomSpell(spellId, SPELLVALUE_BASE_POINT0, addAmount, this, TriggerCastFlags(TRIGGERED_FULL_MASK & ~TRIGGERED_NO_PERIODIC_RESET), nullptr, nullptr, caster->GetGUID());
+        caster->CastCustomSpell(spellId, SPELLVALUE_BASE_POINT0, addAmount, this, TriggerCastFlags(TRIGGERED_FULL_MASK & ~TRIGGERED_NO_PERIODIC_RESET), nullptr, aurEff, caster->GetGUID());
     else
-        caster->m_Events.AddEvent(new AuraMunchingQueue(*caster, GetGUID(), addAmount, spellId), caster->m_Events.CalculateQueueTime(400));
+        caster->m_Events.AddEvent(new AuraMunchingQueue(*caster, GetGUID(), addAmount, spellId, aurEff), caster->m_Events.CalculateQueueTime(400));
```
walkline, line 20228: @heyitsbench Here is your crash related issue.
As you can see, you are adding an event for the future (+400ms, if I'm not mistaken) and capturing the `auraEff` pointer. 
But what would happen if, for example, after 200ms, the aura is removed and the auraEff pointer is deleted? Then, in 400ms the `AuraMunchingQueue` tries to use `auraEff`, but you SHOULD NOT use it, since `auraEff` pointer points to a deleted object in the memory.

One way to solve this is in the AuraMunchingQueue::Execute function, where you can validate whether the auraEff pointer is still valid. You can do this by iterating over the `m_modAuras` of the `_owner `unit and checking if the captured `auraEff` pointer is still in the list. If the `auraEff` pointer is not in the list, it means that the object was deleted and removed from the list.

heyitsbench: https://github.com/azerothcore/azerothcore-wotlk/compare/master...heyitsbench:munching-redux
``
Here's my attempt at implementing what you're suggesting here. Gave a quick test and the behavior I aimed to fix originally is still being corrected, but I was never able to reproduce the crash locally, so I can't confirm whether it fixes that yet. Anything you can point out is immediately wrong/poorly done?

Issue #15849, opened 2023-04-06
Feature: Port Sparring System


Gultask:
### Describe your feature request or suggestion in detail

https://github.com/TrinityCore/TrinityCore/commit/0750b7f8455df39a64462636ca296c6f2aa2b048

Kitzunu:
https://github.com/TrinityCore/TrinityCore/pull/17673 335 pr

heyitsbench:
> TrinityCore 17673 335 pr

Eh, from what I understand in [this issue](https://github.com/TrinityCore/TrinityCore/issues/18576) that PR only added initial support for it for the 335 branch, looks like only master has full support for it through that commit.

Gultask:
What Bench said

Not all creatures need or use this system, though. I'm almost certain it only starts being used WotLK+
That's when perma-sparring first appears afaik. Before that they stop combat after reaching a HP %, many such cases (Demon Hunter Supplicants, Shattered Hand Gladiators, Black Temple Trash...)

PR #19824, opened 2024-08-31
feat(Core/Creatures): implement a sparring system

Grimdhex:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15849

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

The test needs some DB tweaks:
1. You need two creatures hostile to each other (for example, a stormwind guard against a wolf).
2. Add to creature_sparring the guid of creatures that will not receive damage below a certain percentage.
3. Look if it's working and check various situations (should receive damages if the player attacks or mindcontrols the victim or the attackers, for example).

DB Query (go to goldshire to test the change):
```
INSERT INTO `creature_sparring` (`GUID`, `SparringPCT`) VALUES
(4356, 75);

UPDATE `creature` SET `position_x` = -9481.81, `position_y` = 72.47, `position_z` = 56.4849 WHERE `guid` = 4356
``` 

- [ ] I didn't do research on live servers, so without additional information, the behaviour is maybe not 100 % blizzlike.


Grimdhex:
Ready to be tested.

I know that TC handles calculations elsewhere in the code (HasBreakableByDamageCrowdControlAura and DoDamageAndTrigger) so we need to check these situations to see if it's necessary. But I think it should all be good because I'm integrating it right at the end of the process with "DealDamage".

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2783,6 +2788,17 @@ bool Creature::LoadCreaturesAddon(bool reload)
     return true;
 }
 
+void Creature::LoadSparringPct()
+{
+    ObjectGuid::LowType spawnId = GetSpawnId();
+    const auto& sparringData = sObjectMgr->GetSparringData();
```
Kitzunu, line 2794: auto const

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2783,6 +2788,17 @@ bool Creature::LoadCreaturesAddon(bool reload)
...
+
+    auto itr = sparringData.find(spawnId);
+    if (itr != sparringData.end() && !itr->second.empty()) {
```
Kitzunu, line 2797:
```suggestion
    if (itr != sparringData.end() && !itr->second.empty())
    {
```

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1019,6 +1020,17 @@ uint32 Unit::DealDamage(Unit* attacker, Unit* victim, uint32 damage, CleanDamage
         }
     }
 
+    // Sparring
+    if(victim->CanSparringWith(attacker))
```
Kitzunu, line 1024:
```suggestion
    if (victim->CanSparringWith(attacker))
```

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3921,6 +3939,24 @@ void Unit::_UpdateAutoRepeatSpell()
     }
 }
 
+bool Unit::CanSparringWith(const Unit* attacker) const
```
Kitzunu, line 3942: unit const

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3921,6 +3939,24 @@ void Unit::_UpdateAutoRepeatSpell()
...
+
+    if(!attacker)
+        return false;
+
+    if(attacker->GetTypeId() != TYPEID_UNIT || attacker->IsCharmedOwnedByPlayerOrPlayer())
+        return false;
+
+    if(Creature const* me = ToCreature())
+        if(me->GetSparringPct() == 0.0f)
+            return false;
```
Kitzunu, line 3955: spacing with if.

!IsUnit()

!attacker->IsUnit()

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2297,6 +2297,36 @@ void ObjectMgr::LoadCreatures()
     LOG_INFO("server.loading", " ");
 }
 
+void ObjectMgr::LoadCreatureSparring()
```
Kitzunu, line 2300: could this table use ingame reloading? if so add support for it

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -44,6 +44,7 @@
 #include "MoveSplineInit.h"
 #include "MovementGenerator.h"
 #include "ObjectAccessor.h"
+#include "Object.h"
 #include "ObjectMgr.h"
```
Kitzunu, line 48: should be above ObjectAccessor

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2783,6 +2788,17 @@ bool Creature::LoadCreaturesAddon(bool reload)
     return true;
 }
 
+void Creature::LoadSparringPct()
```
Kitzunu, line 2791: On second thought I don't get why we need this. Why can't we just get the sparringpct from SparringData directly in GetSparringPct

data/sql/updates/pending_db_world/rev_1724592400426298600.sql
```diff
@@ -0,0 +1,7 @@
+-- 
+DROP TABLE IF EXISTS `creature_sparring`;
+CREATE TABLE `creature_sparring` (
+  `Entry` int unsigned NOT NULL,
+  `SparringPct` float NOT NULL CHECK(SparringPct between 0 and 100),
+  PRIMARY KEY (`Entry`,`SparringPct`)
```
Kitzunu, line 6:
```suggestion
  `SparringPct` float NOT NULL,
  PRIMARY KEY (`Entry`,`SparringPct`),
  CONSTRAINT `creature_sparring_chk_1` CHECK (`SparringPct` BETWEEN 0 AND 100)
```

Kitzunu:
You should also check in load that guid you are loading exists in creature

src/server/game/Entities/Unit/Unit.cpp
Grimdhex: First we doesn't have IsUnit() then here it is: IsCreature (that we don't have also) because a "Unit" can return: TYPEID_UNIT or TYPEID_PLAYER.

src/server/game/Globals/ObjectMgr.cpp
Grimdhex: I will look to add reloading support.

Grimdhex:
> You should also check in load that guid you are loading exists in creature

fixed by correctly binding the primary key in the sql query

src/server/game/Entities/Creature/Creature.cpp
Grimdhex: If we do this, you'll be performing an operation every time GetSparringPct is called (and with DealDamage it's call very often). I think it's better to load a sparringPct variable for all creatures and check its state as many others variables (spawnId, guid...).

Kitzunu: Load it into store at server startup

src/server/game/Globals/ObjectMgr.cpp
Kitzunu: It is easy if you load them into store at startup

Grimdhex: But I can't reload all creature IG. At least not with the current system. It's easy if we don't store the sparring pct as variable and if acces directory to the store. The only worse thing is if thé store start to be really, you will iterate very often the store and the CPU could  drasticly increase.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -272,7 +272,7 @@ Creature::Creature(bool isWorldObject): Unit(isWorldObject), MovableMapObject(),
     m_transportCheckTimer(1000), lootPickPocketRestoreTime(0), m_combatPulseTime(0), m_combatPulseDelay(0), m_reactState(REACT_AGGRESSIVE), m_defaultMovementType(IDLE_MOTION_TYPE),
     m_spawnId(0), m_equipmentId(0), m_originalEquipmentId(0), m_AlreadyCallAssistance(false),
     m_AlreadySearchedAssistance(false), m_regenHealth(true), m_regenPower(true), m_AI_locked(false), m_meleeDamageSchoolMask(SPELL_SCHOOL_MASK_NORMAL), m_originalEntry(0), m_moveInLineOfSightDisabled(false), m_moveInLineOfSightStrictlyDisabled(false),
-    m_homePosition(), m_transportHomePosition(), m_creatureInfo(nullptr), m_creatureData(nullptr), m_detectionDistance(20.0f), m_waypointID(0), m_path_id(0), m_formation(nullptr), _lastDamagedTime(nullptr), m_cannotReachTimer(0),
+    m_homePosition(), m_transportHomePosition(), m_creatureInfo(nullptr), m_creatureData(nullptr), m_detectionDistance(20.0f),_sparringPct(0.0f), m_waypointID(0), m_path_id(0), m_formation(nullptr), _lastDamagedTime(nullptr), m_cannotReachTimer(0),
```
Helias, line 275: are you sure that you are not breaking other Creature() object initialization? would be better to put the new `_sparringPct` variable as last parameter?

Grimdhex: This order is imposed by gcc/clang, which requires the same order as in the header file declaration (msvc doesn't care about this). In any case, it doesn't change anything. Nothing is deleted or modified.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2792,6 +2797,18 @@ bool Creature::LoadCreaturesAddon(bool reload)
...
 
+void Creature::LoadSparringPct()
+{
+    ObjectGuid::LowType spawnId = GetSpawnId();
+    auto const& sparringData = sObjectMgr->GetSparringData();
+
+    auto itr = sparringData.find(spawnId);
+    if (itr != sparringData.end() && !itr->second.empty())
+    {
+        _sparringPct = itr->second[0];
```
Winfidonarleyan, line 2808: Need select random element from container instead of first

Grimdhex: Why ? One SpawnId, one pct value ?

Kitzunu:
> > You should also check in load that guid you are loading exists in creature
> 
> fixed by correctly binding the primary key in the sql query

While it is good, I do prefer checking it on load to minimize SQL errors and favoring logging errors

Grimdhex:
Not tested it yet, but I found a solution for the reloading system. It should works.

For your suggestion @Kitzunu, it works but as I said when the list grows it causes a heavy CPU charge because for tick damage on all creatures, we need to access the list to verify if the creature has a sparring entry. And parse an unordered map cost more than just checking a variable in a object...

TheSCREWEDSoftware:
@Grimdhex are you still working on this or is a state of being tested?

Grimdhex:
Could be tested

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3954,6 +3971,24 @@ void Unit::_UpdateAutoRepeatSpell()
...
+    if (!IsCreature() || IsCharmedOwnedByPlayerOrPlayer())
+        return false;
+
+    if (!attacker)
+        return false;
+
+    if (!attacker->IsCreature() || attacker->IsCharmedOwnedByPlayerOrPlayer())
+        return false;
+
+    if (Creature const* me = ToCreature())
```
Kitzunu, line 3985: I personally dont like using "me" outside of scripts. I would do creature instead

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -3954,6 +3971,24 @@ void Unit::_UpdateAutoRepeatSpell()
...
+        if (me->GetSparringPct() == 0.0f)
```
Kitzunu, line 3986:
```suggestion
        if (!me->GetSparringPct())
```

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2298,6 +2298,42 @@ void ObjectMgr::LoadCreatures()
...
+    do
+    {
+        Field* fields = result->Fetch();
+
+        ObjectGuid::LowType spawnId     = fields[0].Get<uint32>();
+        float sparringHealthPct         = fields[1].Get<float>();
+
+        if (!sObjectMgr->GetCreatureData(spawnId))
+        {
+            LOG_ERROR("sql.sql", "Entry {} has a record in `creature_sparring` but doesn't exist in 'creatures' table");
```
Kitzunu, line 2324: use same backticks around creatures

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2298,6 +2298,42 @@ void ObjectMgr::LoadCreatures()
...
+            continue;
+        }
+
+        _creatureSparringStore[spawnId].push_back(sparringHealthPct);
+
+        ++count;
+    } while (result->NextRow());
+
+    LOG_INFO("server.loading", ">> Loaded {} Creatures in {} ms", count, GetMSTimeDiffToNow(oldMSTime));
```
Kitzunu, line 2333: should be loaded x creature sparrings or something. otherwise this will be confused with creature data

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -2298,6 +2298,42 @@ void ObjectMgr::LoadCreatures()
...
 
+void ObjectMgr::LoadCreatureSparring()
+{
+    uint32 oldMSTime = getMSTime();
+
+    QueryResult result = WorldDatabase.Query("SELECT GUID, SparringPCT FROM creature_sparring");
+
+    if (!result)
+    {
+        LOG_WARN("server.loading", ">> Loaded 0 creatures. DB table `creature_sparring` is empty.");
```
Kitzunu, line 2309: same as above

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1032,6 +1032,17 @@ uint32 Unit::DealDamage(Unit* attacker, Unit* victim, uint32 damage, CleanDamage
         }
     }
 
+    // Sparring
+    if (victim->CanSparringWith(attacker))
+    {
+        if(damage >= victim->GetHealth())
```
Kitzunu, line 1038: space after if

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2635,6 +2646,12 @@ void Unit::AttackerStateUpdate(Unit* victim, WeaponAttackType attType /*= BASE_A
             Unit::DealDamageMods(victim, damageInfo.damages[i].damage, &damageInfo.damages[i].absorb);
         }
 
+        // Related to sparring system. Allow attack animations even if there are no damages
+        if (victim->CanSparringWith(damageInfo.attacker))
+        {
+            damageInfo.HitInfo |= HITINFO_FAKE_DAMAGE;
+        }
```
Kitzunu, line 2653: remove brackets

Exitare:
Tested the PR using your provided example. 
I tested the following things:

- [x] Just observed whether the npcs will fight without dying (Works)
- [x] Attacked with my player attacker using a meele attack. (Works)
- [x] Used a priest with mind control to attack the opponent to kill it. (Works)  
- [x] Used a range spell to kill one of the opponents (Works)
- [x] Killed both opponents, respawned them by using .respawn all and observed whether they would engange in fighting again without killing each other. (Works)

Grimdhex:
> Tested the PR using your provided example. I tested the following things:
> 
> * [ ]  Just observed whether the npcs will fight without dying (Works, kind of). The wolf if able to kill the guard.
> * [x]  Attacked with my player attacker using a meele attack. Able to kill the wolf (Works)
> * [x]  Used a priest with mind control to attack the opponent to kill it. Able to kill the guard(s) (Works)
> * [x]  Used a range spell to kill one of the opponents (Works)
> * [ ]  Killed both opponents, respawned them by using .respawn all and observed whether they would engange in fighting again without killing each other. (Works, kind of). Again, same as above. The wolf is able to kill the guard but the guard can not kill the wolf.

Thanks, so it's work as intended. This PR allows to an npc to not receive damage under a certain percentage (very rare cases of use before a cataclysm, but they do exist).

Grimdhex:
Hi @Kitzunu  I will be busy for a long time. To avoid that this PR fall in limbs. Is it good for you? Just to merge this.

Kitzunu (approved):

PR #19809, opened 2024-08-31
fix(Scripts/BlackTemple): Ensure Illidari council health is balanced across wipes.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Scripts (bosses, spell scripts, creature scripts).

## Issues Addressed:

- Closes unreported issue. Depending on how the fight is reset, the council creature can break and no longer balance health between the council members.

## SOURCE:

The changes:
- [X] Have not been validated.

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.

## How to Test the Changes:

- [X] This pull request requires further testing.
1. Engage the Illidari council.
2. Wipe.
3. Reset the fight.
4. Observe the council.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #19849, opened 2024-09-02
fix(Scripts/BlackTemple): Adjust order of damage calculations for Shared Rule.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/19809#issuecomment-2323729656

## SOURCE:

The changes:
- [X] Have not been validated.

## Tests Performed:

This PR has been:

- [X] I don't think this fully matches behavior observed on official servers, but it makes the fight completable as an interim solution.
  - Upon studying of a sniff, it definitely appears a council member with effectively no health remaining should still be relaying full damage through the Shared Rule spell. This fix definitely seems more Blizzlike than current behavior.

### DIFF
--- a/src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp
+++ b/src/server/scripts/Outland/BlackTemple/boss_illidari_council.cpp
@@ -254,13 +254,13 @@ struct boss_illidari_council_memberAI : public ScriptedAI
     {
         InstanceScript* instance = me->GetInstanceScript();
 
-        if (me->GetHealth() <= damage)
-            damage = me->GetHealth() - 1;
-
         int32 damageTaken = damage;
         Creature* target = instance->GetCreature(DATA_ILLIDARI_COUNCIL);
 
         me->CastCustomSpell(target->ToUnit(), SPELL_SHARED_RULE_DMG, &damageTaken, &damageTaken, &damageTaken, true, nullptr, nullptr, me->GetGUID());
+
+        if (me->GetHealth() <= damage)
+            damage = me->GetHealth() - 1;
     }
 
     void KilledUnit(Unit*) override


kissingers:
another problem is. because can't kill the 4 boss, I use die kill all, but the instance can't out of combat.
this is a old problem, at my fix before, I modify and work ok

            if (!UpdateVictim())
                return;

also when time is 120s, the spell can get by magic class also 120s, so there is 100% after 120s, the magic will no buff and might die , so I changed to 
events.ScheduleEvent(EVENT_SPELL_DAMPEN_MAGIC, 110000);

also when the SPELL_BLESSING_OF_SPELL_WARDING at the magic boss, the magic class cant get buff form the boss, might need cant do the buff to the magic boss

this my fix before half a year
https://github.com/azerothcore/azerothcore-wotlk/commit/a69ba616f1871ed43978072a1c23d4e09ccba982

heyitsbench:
This PR should make the council killable, so your first fix should no longer be needed.

kissingers:
> This PR should make the council killable, so your first fix should no longer be needed.



> This PR should make the council killable, so your first fix should no longer be needed.

I know, but cant out of combat seem still will happen.
and the mage stole the buff after only 120 seconds and the boss refreshes 120 time is the same, will certainly lead to the mage without buff instantly be killed situation.
I didn't see if the mage boss gets magic protection now, if it does, it means that the mage's spell steal can't steal the buff, and uncontrollable surprises can happen as well

kissingers:
and my fix only on my branch not pull to master branch, but it does solve the uncontrollable problems of not being able to get out of combat and mages not being able to steal buffs or not being able to steal buffs in combat

PkllonG:
Tested

Nyeriah (approved):

heyitsbench:
> I use .die kill all

For the record, this could be the reason why you can't leave combat, because the fight doesn't actually end properly. These creatures are supposed to be killed by script, their death isn't what actually causes the encounter to end.

PR #19878, opened 2024-09-06
fix(DB): Missing spanish texts in some city guardian directions

Varoso:
In some cities guard directions had some options not translated to spanish

## Changes Proposed:

This PR proposes changes to:

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Set the WoW client to esES and speak with a guard of a city (Ex: Silvermoon), some texts have a missing spanish translation and are in english
2. Execute the script added to the PR and restart the server
3. Speak again with the guard and all the texts are now in spanish


pangolp:
Whenever an INSERT INTO is performed, it must be accompanied by a DELETE FROM, so that the script is executed as many times as necessary, without failing. In the current way, the first time it is executed without problems, but the second time it is executed, it will give an error, because it is going to try to insert something that already exists. It would be rare for this to happen, but it is always a good idea to prevent this type of failure. Also, when doing an INSERT INTO it is advisable to define the list of attributes.

Finally, if you give me information on which NPC I should test, I will run the tests to determine what you are mentioning. Technically, the translations should go in another table, but just in case, I would like to know which NPC is having problems in this case.

pangolp:
image

I'm using an esES client, and the problem is the same as the Dalaran trainers had. But the change you would have to make is different. If you do the following query, you will see that the gossips do not have the broadcasttextid, so they are only shown in English. But if you assign those values ​​to them, the translations will be available, not only in Spanish, but also in other languages.

https://www.wowhead.com/wotlk/npc=16222/silvermoon-city-guardian

```sql
SELECT `gossip_menu_id` FROM `creature_template` WHERE `entry`=16222;

SELECT * FROM `gossip_menu_option` WHERE `MenuID`=7633;

SELECT `MenuID`, `OptionID`, `OptionIcon`, `OptionText` FROM `gossip_menu_option` WHERE `MenuID`=7633 AND `OptionBroadcastTextID`=0;
```

data/sql/updates/pending_db_world/rev_1725625860N.sql
```diff
@@ -0,0 +1,61 @@
+-- Script to add missing spanish options in some city guardians directions
+INSERT INTO acore_world.gossip_menu_option_locale VALUES
```
Helias, line 2: never include database name when you do a PR, moreover use the backtick ` and add the related name field behind `VALUES`, like (`field1`, `field2`, `field3`, `field4`, `field5`)
```suggestion
INSERT INTO `gossip_menu_option_locale` VALUES
```

sudlud:
Is there an update to this? Assigning the correct broadcasttextids as @pangolp suggested would be the preferred solution here.

pangolp:
> Is there an update to this? Assigning the correct broadcasttextids as @pangolp suggested would be the preferred solution here.

I couldn't push the texts, because of permission, you don't have other people's contributions enabled. If you enable them, I push them or someone who has permissions, who approves the recommendations and I think you can merge it seamlessly.

sudlud:
@pangolp please feel free to post the changes here in the PR as a review with "suggested" changes so they can be committed

pangolp:
> @pangolp please feel free to post the changes here in the PR as a review with "suggested" changes so they can be committed

Try to make a commit, but do not accept changes from contributors. But the suggestions are made I think. Can't you accept them?

pangolp:
If not, what I do, is another PR with the changes, we add the person as co-author, and then we close this pull request, after the other one merges. Because the suggestions for changes were already there to make the commits, because I let myself upload the changes to the branch.

sudlud:
> If not, what I do, is another PR with the changes, we add the person as co-author, and then we close this pull request, after the other one merges. Because the suggestions for changes were already there to make the commits, because I let myself upload the changes to the branch.

Yes please do this.

Varoso:
I'm very sorry, I've been out of town for work and haven't been able to see your comments until now. From what I see, there is a secondary pull request with the appropriate changes to resolve this. I'll close this one so that only the other one remains. Thank you very much for your help!

Issue #19904, opened 2024-09-07
[Loot] Unyielding mobs low drop of Runecloth


alecsci:
### Current Behaviour

Reported by: https://github.com/chromiecraft/chromiecraft/issues/5784
> Unyielding Mobs - Footman, Knight, and Sorcerer - all drop excessively low Runecloth (and too high level gear, which will be submitted in another ticket)

### Expected Blizzlike Behaviour

> All Unyielding mobs have slightly more than a 50% chance to drop 1-3 pieces of runecloth. (Sources below)

### Source

https://www.wowhead.com/wotlk/npc=16904/unyielding-footman
https://www.wowhead.com/wotlk/npc=16905/unyielding-sorcerer
https://www.wowhead.com/wotlk/npc=16906/unyielding-knight

### Steps to reproduce the problem

Farm these:
1. go c id 16904
2. go c id 16905
3. go c id 16906

### Extra Notes

Keira shows 20% chance drop for runecloth for these 3 creatures.

### AC rev. hash/commit

cf95c1dfbdb252cfd18e32f8bdfef0da8ff01092

### Operating system

Windows 11 x64

PR #19910, opened 2024-09-08
fix(DB/Creature): [Loot] Unyielding mobs low drop of Runecloth

Carriola-wotlk:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19904
- Closes chromiecraft/chromiecraft#5784

I also changed drop rate of Netherweave Cloth (it was too high, it was 60%). Here are the changes:

Unyielding Footman
- Runecloth chance 50 (chance was 20)
- Netherweave Cloth chance 25 (chance was 60)
- GroupId 1

Unyielding Sorcerer
- Runecloth chance 50 (chance was 20)
- Netherweave Cloth chance 25 (chance was 60)
- GroupId 1

Unyielding Knight
- Runecloth chance 50 (chance was 20)
- Netherweave Cloth chance 25 (chance was 60)
- GroupId 1

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/data/sql/updates/pending_db_world/2024_09_08_00.sql
+++ b/data/sql/updates/pending_db_world/2024_09_08_00.sql
@@ -1,40 +1,40 @@
 -- Unyielding Footman
 
--- Runecloth chance 53 (chance was 20)
+-- Runecloth chance 50 (chance was 20)
 UPDATE `creature_loot_template`
-SET `Chance` = 53
+SET `Chance` = 50, `GroupId` = 1
 WHERE `Entry` = 16904 AND `Item` = 14047;
 
 -- Netherweave Cloth chance 25 (chance was 60)
 UPDATE `creature_loot_template`
-SET `Chance` = 25
+SET `Chance` = 25, `GroupId` = 1
 WHERE `Entry` = 16904 AND `Item` = 21877;
 
 
 
 -- Unyielding Sorcerer
 
--- Runecloth chance 54 (chance was 20)
+-- Runecloth chance 50 (chance was 20)
 UPDATE `creature_loot_template`
-SET `Chance` = 54
+SET `Chance` = 50, `GroupId` = 1
 WHERE `Entry` = 16905 AND `Item` = 14047;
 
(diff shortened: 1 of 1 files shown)


elthehablo:
What are the sources on this?

elthehablo:
I see. Wowhead. Unfortunately not a reliable source. It's best to get some second source to check these numbers with. We can either check cmangos databases or using the law of big numbers by killing these on a TBC/wotlk/cataclysm live server

alecsci:
@elthehablo, isn't wotlk wowhead good enough ?

elthehablo:
It depends on the issue. The problem is that you just cannot trust those numbers.

Items that shouldn't even drop off of those creatures sometimes show up there.

Gultask:
My take on this is that cloth at these numbers we can find the intuit the chances by looking at Wowhead's numbers. In this case:
50% for Runecloth
25% for Netherweave

My only question is: Can they drop both cloth at the same time?
Because if they can, then the current method is fine.
If they **cannot** there are 2 possibilities:
1. Only one cloth can drop at a time, at the same percentages, so they'd only need the same GroupID
2. Level 59 creatures drop Runecloth while level 60 creatures drop Netherweave. We don't have a clean system for that so we normally just go with the solution to the first possibility :P

I'm trying to recall if Hellfire Peninsula creatures can drop both cloths at the same time, but as a rule of thumb creatures normally can't so what I'd do is set all Runecloth to 50, all Netherweave to 25, and set both to the same GroupId, so only one cloth can drop at any one point.

Carriola-wotlk:
> I'm trying to recall if Hellfire Peninsula creatures can drop both cloths at the same time, but as a rule of thumb creatures normally can't so what I'd do is set all Runecloth to 50, all Netherweave to 25, and set both to the same GroupId, so only one cloth can drop at any one point.

ok, so I edit chances (Runecloth to 50% and Netherweave to 25%) and set GroupId 1 for both

sudlud (approved):

PR #19922, opened 2024-09-09
feat(ScriptMgr): add OnArenaStart hook

Helias:
## Changes Proposed:

This PR proposes changes to:
-  Add new hook OnArenaStart() because OnBattlegroundStart() runs only when a BG (non-arena) starts.
 
## Tests Performed:

I've only compiled it locally

## How to Test the Changes:

Use the hook through a module, ex. 3v3soloQ


laasker:
Tested with BGs and Arenas, the hook calls when the doors of an arena open.

PR #19923, opened 2024-09-09
fix(Core/AI): Reduce movement leash extension interval.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Progresses leashing issue.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Engage mobs.
2. A lot of mobs.
3. Ensure they don't prematurely leash too much.

- [ ] It seems the only issue I've seen contributing to premature leashing on AC (as has been shown in [this video](https://github.com/chromiecraft/chromiecraft/issues/7345#issuecomment-2337661304)) is creature fanning, i.e. spreading themselves around a player. This feature afaik is not implemented in any capacity on VMaNGOS, so it doesn't cause any problems there, but on AC it causes creatures to never stop moving if there are enough of them or they are big enough. This doesn't fix that issue, but it does mitigate it hopefully significantly enough to cause a noticeable difference.

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -79,7 +79,10 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
         owner->StopMoving();
         _lastTargetPosition.reset();
         if (cOwner)
+        {
+            cOwner->UpdateLeashExtensionTime();
             cOwner->SetCannotReachTarget();
+        }
         return true;
     }
 
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -78,8 +78,8 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     {
         owner->StopMoving();
         _lastTargetPosition.reset();
-        if (Creature* cOwner2 = owner->ToCreature())
-            cOwner2->SetCannotReachTarget();
+        if (cOwner)
+            cOwner->SetCannotReachTarget();
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


heyitsbench:
CC: @PkllonG

PkllonG:
I found another bug where the mage used a spell to attack a creature into battle, and the creature would leave the fight after 1 second. There was this error before the this PR

PkllonG:
It works better after PR than before PR, but there are still the bugs I described above

heyitsbench:
> [...] but there are still the bugs I described above

Good to know, but not related to this PR. Thank you for testing. ❤️

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -154,10 +154,10 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
     i_leashExtensionTimer.Update(time_diff);
     if (i_leashExtensionTimer.Passed())
     {
-        i_leashExtensionTimer.Reset(5000);
+        i_leashExtensionTimer.Reset(1500);
         if (owner->HasUnitState(UNIT_STATE_CHASE) && owner->movespline->Finalized() && !target->isMoving())
```
walkline, line 158: Why do we have this `target->isMoving()` check?

Originally, in vMangos, this timer only ticks when owner->movespline->Finalized() is true (without checking owner->HasUnitState(UNIT_STATE_CHASE)). Otherwise, it resets the timer.

heyitsbench: If @PkllonG would like to chime in and provide the reasoning, we can add it back, but for now I've adjusted the checks for parity with VMaNGOS.

PkllonG: It was added because, I think, it is forbidden to work when the player is running away. Otherwise, the player cannot leave the battle.

PkllonG: If there's a better way, feel free to change or delete it

heyitsbench: The system as I understand it should not rely on the target's movement hardly at all. If the creature stands still, then it should reset its leash. The `i_recalculateTravel` (I believe) indicates the mob is fanning/spreading out in front of the player, so that also resets the movement leash timer check.

I'm also not convinced about the reset interval either used in this PR or as used by VMaNGOS, but most videos I've seen describing the system suggest it is 1.5s, so that's what I've chosen to replace it with here, as it does seem to help mitigate premature leashing issues.

Nyeriah (approved):

PR #19938, opened 2024-09-11
fix: default bgarenas visibility distance default value

Helias:
fix startup error `Visibility.Distance.BGArenas can't be greater 250`

## Changes Proposed:

Set properly the default value of Visibility.Distance.BGArenas

## SOURCE:
Core implementation

## Tests Performed:

I only run a build

## How to Test the Changes:

A successful build should be enough

### DIFF
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -1283,11 +1283,11 @@ Visibility.GroupMode = 1
 #                     Min limit is max aggro radius (45) * Rate.Creature.Aggro
 #        Default:     100  - (Visibility.Distance.Continents)
 #                     170 - (Visibility.Distance.Instances)
-#                     533 - (Visibility.Distance.BGArenas)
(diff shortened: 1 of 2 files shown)


sudlud:
follow-up to #19708, wondering why the startup error didn't make CI fail

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -1283,11 +1283,11 @@ Visibility.GroupMode = 1
 #                     Min limit is max aggro radius (45) * Rate.Creature.Aggro
 #        Default:     100  - (Visibility.Distance.Continents)
 #                     170 - (Visibility.Distance.Instances)
-#                     533 - (Visibility.Distance.BGArenas)
+#                     250 - (Visibility.Distance.BGArenas)
```
sudlud, line 1286: The number must also be updated right above in a comment

Helias: done, nice catch!

sudlud (approved):

Issue #19966, opened 2024-09-14
[GM COMMAND] Level up


manstfu:
### Current Behaviour

`.levelup $character 79` and `.char level $character 80` doesn't add level achievement to character if he is offline, it works for online character!

### Expected Blizzlike Behaviour

Add level achievement even if character is offline

### Source

My own mind

### Steps to reproduce the problem

.char level $name 80 while character is offline

### AC rev. hash/commit

rev. af776520dbd5+

### Operating system

Debian 12

manstfu:
Playerbot related!

manstfu:
Seems not playerbot related, can someone confirm this?

Nyeriah:
Yeah, it's not granting achievements.

manstfu:
Any possibility to be related to https://github.com/azerothcore/azerothcore-wotlk/blob/4fc60bbbdfb6e40b4e174333065c8365f639dfef/data/sql/updates/db_characters/2024_09_03_00.sql#L4 ?

PR #19969, opened 2024-09-14
fix(Scripts/Commands): Fix level up achievements not being granted

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19966

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Handlers/CharacterHandler.cpp
+++ b/src/server/game/Handlers/CharacterHandler.cpp
@@ -987,8 +987,12 @@ void WorldSession::HandlePlayerLoginFromDB(LoginQueryHolder const& holder)
 
     if (pCurrChar->HasAtLoginFlag(AT_LOGIN_CHECK_ACHIEVS))
     {
-        pCurrChar->RemoveAtLoginFlag(AT_LOGIN_CHECK_ACHIEVS, true);
-        pCurrChar->CheckAllAchievementCriteria();
+        // If we process the check while players are loading they won't be notified of the changes.
+        pCurrChar->m_Events.AddEventAtOffset([pCurrChar]
+        {
+            pCurrChar->RemoveAtLoginFlag(AT_LOGIN_CHECK_ACHIEVS, true);
+            pCurrChar->CheckAllAchievementCriteria();
+        }, 1s);
     }
 
     bool firstLogin = pCurrChar->HasAtLoginFlag(AT_LOGIN_FIRST);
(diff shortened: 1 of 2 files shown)


sudlud:
@walkline

Nyeriah:
I might have misunderstood the offline achievement implementation but I don't think it is reasonable here to check every player level, against every achievement ID/criteria ID, so the criterias are updated when the check on login does effectively the same, which is different from where the offline implementation has been used so far, as in the AH criterias it would be just plain impossible to update with offline players

src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -987,8 +987,12 @@ void WorldSession::HandlePlayerLoginFromDB(LoginQueryHolder const& holder)
 
     if (pCurrChar->HasAtLoginFlag(AT_LOGIN_CHECK_ACHIEVS))
     {
-        pCurrChar->RemoveAtLoginFlag(AT_LOGIN_CHECK_ACHIEVS, true);
-        pCurrChar->CheckAllAchievementCriteria();
+        // If we process the check while players are loading they won't be notified of the changes.
+        pCurrChar->m_Events.AddEventAtOffset([pCurrChar]
```
Kitzunu, line 991: would we be able to do `if (!pCurrChar->GetSession()->PlayerLoading())` instead of

Nyeriah: No, because then it would just return false and not run at all 🤷

walkline:
> I might have misunderstood the offline achievement implementation but I don't think it is reasonable here to check every player level, against every achievement ID/criteria ID, so the criterias are updated when the check on login does effectively the same. I also believe this differs from the cases where the offline implementation has been used so far, as in the AH criterias it would be just plain impossible to update with offline players

I believe this should do the same thing, but I can’t test this at the moment:
```
// inside command handler for offline player
sAchievementMgr->UpdateAchievementCriteriaForOfflinePlayer(playerGuid.GetCounter(), ACHIEVEMENT_CRITERIA_TYPE_REACH_LEVEL);
```

But I don’t mind to use the current suggested approach.

Nyeriah:
I'll check the suggestion and report back once I am able

Nyeriah:
Works

src/server/game/Handlers/CharacterHandler.cpp
Nyeriah: I'm leaving this here because it still fixes achievements not being reported to the client on the first login after flag application

sudlud (approved):

PR #19991, opened 2024-09-17
fix(Core/Spells): Resolve a case of triggered spells prolonging combat.

heyitsbench:
Attempt two from https://github.com/azerothcore/azerothcore-wotlk/pull/19819.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/13330

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] This PR absolutely cannot be merged until testing is done.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -20233,19 +20233,19 @@ public:
     bool Execute(uint64 /*eventTime*/, uint32 /*updateTime*/) override
     {
         if (_owner.IsInWorld() && _owner.FindMap())
-        {
-            bool auraFound = false;
-            Unit::AuraEffectList const& auraEffects = _owner.GetAuraEffectsByType(_auraType);
-            for (Unit::AuraEffectList::const_iterator j = auraEffects.begin(); j != auraEffects.end(); ++j)
-                if ((*j) == _aurEff)
-                    auraFound = true;
-
-            if (!auraFound)
-                _aurEff = nullptr;
-
             if (Unit* target = ObjectAccessor::GetUnit(_owner, _targetGUID))
+            {
+                bool auraFound = false; // Used to ensure _aurEff exists to avoid nullptr dereference/crash
+                Unit::AuraEffectList const& auraEffects = target->GetAuraEffectsByType(_auraType);
+                for (Unit::AuraEffectList::const_iterator j = auraEffects.begin(); j != auraEffects.end(); ++j)
+                    if ((*j) == _aurEff)
+                        auraFound = true;
+
+                if (!auraFound)
+                    _aurEff = nullptr;
+
                 _owner.CastCustomSpell(_spellId, SPELLVALUE_BASE_POINT0, _basePoints, target, TriggerCastFlags(TRIGGERED_FULL_MASK & ~TRIGGERED_NO_PERIODIC_RESET), nullptr, _aurEff, _owner.GetGUID());
-        }
+            }
 
         return true;
     }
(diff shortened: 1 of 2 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -20228,13 +20228,24 @@ void Unit::OutDebugInfo() const
...
 {
 public:
-    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId) { }
+    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId, AuraEffect* aurEff, AuraType auraType) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId), _aurEff(aurEff), _auraType(auraType) { }
 
     bool Execute(uint64 /*eventTime*/, uint32 /*updateTime*/) override
     {
         if (_owner.IsInWorld() && _owner.FindMap())
+        {
+            bool auraFound = false;
```
sudlud, line 20237: I feel like this section could maybe need a comment describing why this search is beeing done

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -20228,13 +20228,24 @@ void Unit::OutDebugInfo() const
...
+            Unit::AuraEffectList const& auraEffects = _owner.GetAuraEffectsByType(_auraType);
```
walkline, line 20238: I think I mixed up the owner and the target. This aura belongs to the target (right?), so we need to check the aura on the target, not the owner.

heyitsbench: You seem to be correct, adjusted the PR accordingly.

src/server/game/Entities/Unit/Unit.cpp
heyitsbench: Added a comment, let me know if it's not descriptive enough.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -20228,13 +20228,24 @@ void Unit::OutDebugInfo() const
...
-    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId) { }
+    AuraMunchingQueue(Unit& owner, ObjectGuid targetGUID, int32 basePoints, uint32 spellId, AuraEffect* aurEff, AuraType auraType) : _owner(owner), _targetGUID(targetGUID), _basePoints(basePoints), _spellId(spellId), _aurEff(aurEff), _auraType(auraType) { }
 
     bool Execute(uint64 /*eventTime*/, uint32 /*updateTime*/) override
     {
         if (_owner.IsInWorld() && _owner.FindMap())
             if (Unit* target = ObjectAccessor::GetUnit(_owner, _targetGUID))
-                _owner.CastCustomSpell(_spellId, SPELLVALUE_BASE_POINT0, _basePoints, target, TriggerCastFlags(TRIGGERED_FULL_MASK & ~TRIGGERED_NO_PERIODIC_RESET), nullptr, nullptr, _owner.GetGUID());
+            {
+                bool auraFound = false; // Used to ensure _aurEff exists to avoid nullptr dereference/crash
```
walkline, line 20238: This is a minor nitpick, so you can probably ignore it, but the issue we're trying to solve here is called a "dangling/wild pointer/reference".

heyitsbench: Don't want to be putting incorrect information in the comments. Thanks for teaching me something today! ❤️

kissingers:
So I wonder after fix the crash, Is it still can work fine for  "resolve a case of triggered spells prolonging combat" ?

heyitsbench:
> So I wonder after fix the crash, Is it still can work fine for "resolve a case of triggered spells prolonging combat" ?

I did test it locally and the issue I was after *was* still solved by this PR.

Issue #20010, opened 2024-09-21
Error while parsing SQL > .go creature name Kaz'rogal <


TheSCREWEDSoftware:
### Current Behaviour

If attempt to teleport to that npc and i would assume any other NPC that breaks SQL query due to the ' as example: 
`.go creature name Kaz'rogal` You will crash and get the following message.

![image](https://github.com/user-attachments/assets/694f7315-ccf6-4474-b74a-b582a6a2f8c2)

### Expected Blizzlike Behaviour

I don't think blizzard at the time would make the try and expect of the query for someone using quotes within a apostrophe.

### Steps to reproduce the problem

1) Start the server Auth and World
2) Open the client and join, make sure you have GM access commands.
3) .go creature name Kaz'rogal

or 

4) . go creature name "Sinister Reflection"

They will both crash you worldserver with the same error.

### Extra Notes

For me .go creature name Kaz'rogal

### AC rev. hash/commit

```
AzerothCore rev. 64c9e4a0a4d5 2024-09-15 21:16:14 +0000 (master branch) (Win64, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
Server uptime: 38 second(s)
Update time diff: 2ms. Last 500 diffs summary:
|- Mean: 5ms
|- Median: 2ms
|- Percentiles (95, 99, max): 12ms, 19ms, 130ms
Using SSL version: OpenSSL 3.3.2 3 Sep 2024 (library: OpenSSL 3.3.2 3 Sep 2024)
Using Boost version: 1.86.0
Using CMake version: 3.29.8
Using MySQL version: 80039
Found MySQL Executable: C:/Program Files/MySQL/MySQL Server 8.0/bin/mysql.exe
Compiled on: Windows 10.0.17763
Worldserver listening connections on port %hu
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in Data/maps. Total size: 291014951 bytes
vmaps directory located in Data/vmaps. Total size: 658130721 bytes
mmaps directory located in Data/mmaps. Total size: 2192910844 bytes
Default DBC locale: enUS.
All available DBC locales: enUS
Using World DB: ACDB 335.12-dev
Latest LoginDatabase update: acore_cms_subscriptions.sql
Latest CharacterDatabase update: trasmorg.sql
[...]

PR #20012, opened 2024-09-21
fix(Scripts/Commands): Prevent crash if you use doublequotes in go cr…

Kitzunu:
…eature name

* closes https://github.com/azerothcore/azerothcore-wotlk/issues/20010

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Commands/cs_go.cpp
+++ b/src/server/scripts/Commands/cs_go.cpp
@@ -123,7 +123,14 @@ public:
         if (!name.data())
             return false;
 
-        QueryResult result = WorldDatabase.Query("SELECT entry FROM creature_template WHERE name = \"{}\" LIMIT 1" , name.data());
+        // Make sure we don't pass double quotes into the SQL query. Otherwise it causes a MySQL error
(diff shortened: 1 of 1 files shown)


Nyeriah (approved):

walkline:
What do you think about using a prepared statement here? I believe that would eliminate the need for the additional validation logic and prevent SQL injection vulnerability.

Kitzunu:
Wouldn't it be the same issue? Or am I thinking wrong?

walkline:
Not exactly. With a prepared statement, we’re not sending the entire query because MySQL has already analyzed the prepared statement and is only expecting parameters. When it receives a parameter, such as a string, MySQL will escape it to ensure it fits correctly into the query without causing any issues. As a result, a string like `"Some name"` would likely be transformed into something like `\"Some name\"` or similar.

PR #20030, opened 2024-09-24
refactor(SQL/Quest) Rescuing the Rescuers ID 11244

pangolp:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/AdonisWOW/bugtracker/issues/6
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20037

## SOURCE:

- https://www.wowhead.com/wotlk/quest=11244/rescuing-the-rescuers
- https://www.wowhead.com/wotlk/object=186565/ceremonial-dragonflayer-harpoon
- https://www.wowhead.com/wotlk/spell=42968/removing-dragonflayer-harpoon
- https://www.wowhead.com/wotlk/npc=24077/impaled-valgarde-scout

Currently, and the reason I was actually reviewing this quest, is that sometimes the smartAI is not executed, making it so that with only 1 npc, the quest can be completed without the need to go to the rest. I can't find that flaw, to attach it to this solution. If anyone can think of a possible cause, let me know so I can review it. But despite not being able to find a solution to that problem, I took the opportunity to attach some translations and make a modification to the smartAI.

https://github.com/user-attachments/assets/92a12a00-152f-43f8-950b-a7a1164cbe8a

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Initially, gameobjects are spawned in the world. In this case, they will be applied to NPCs. The loading bar was also translated into Spanish, and the broadcast_text was assigned to the 2 missing texts, so that they can be seen in all languages ​​and not just in English.


pangolp:
I have the feeling that for some reason it cannot select the target, and therefore the smartAI cannot be executed, for that reason the spear remains in place and can be used infinitely. But I do not know the reason why this could be happening.

Helias (approved):

pangolp:
Thanks @Helias

PR #20038, opened 2024-09-24
chore: change bg type variable uint8 -> uint16 for 3rd modules compatibility

Helias:
This will allow values bigger than 255 to be a `BattlegroundTypeId`.
For retro-ported arenas like Tol'Viron or Tiger's Peak, where the MoP IDs are very big, it is necessary to consider this type as uint16.

I set it as WIP because I would like to make some further tests to see if other changes are needed.

related modules:
- https://github.com/Helias/mod-arena-tolviron
- https://github.com/Helias/mod-arena-tigerspeak

## Changes Proposed:

This PR proposes changes to: a variable type from `uint8` to `uin16`

## Tests Performed:
tested with modules and 

## How to Test the Changes:
A successful build should be enough

### DIFF
--- a/src/server/game/Scripting/ScriptDefines/AllBattlegroundScript.h
+++ b/src/server/game/Scripting/ScriptDefines/AllBattlegroundScript.h
@@ -44,7 +44,7 @@ enum AllBattlegroundHook
 };
 
 enum BattlegroundBracketId : uint8;
-enum BattlegroundTypeId : uint8;
+enum BattlegroundTypeId : uint16;
 enum TeamId : uint8;
 
 class AllBattlegroundScript : public ScriptObject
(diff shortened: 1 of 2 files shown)


Grimdhex:
I see that BattlegroundId is define by dbc. Does it send somewhere to the client ? Because it's maybe something that we couldn't change without breaking an opcode.

sudlud:
does this also require changing lines like this:
https://github.com/azerothcore/azerothcore-wotlk/blob/510422a4f5ecca4afdcc41bedac2b5f4bae1706d/src/server/game/Battlegrounds/BattlegroundMgr.cpp#L142

sudlud:
> I see that BattlegroundId is define by dbc. Does it send somewhere to the client ? Because it's maybe something that we couldn't change without breaking an opcode.

it's probably worth checking the code source before merging this.

e.g. it's used within a world packet in `BattlegroundMgr::BuildBattlegroundListPacket()` but there it's beeing sent as uint32, so this should be alright

Kitzunu:
You would need some core refactor because updating in-match rated arenas iterate on a uint8 for loop, which wouldn't update any bg with uint16 values

sudlud:
Closing this PR for now, please feel free to reopen and continue working on it anytime in the future :)

PR #20039, opened 2024-09-24
new BattlegroundDesertionTypes for hook OnBattlegroundDesertion

Helias:
… into new-BattlegroundDesertionTypes

original author @laasker

## Changes Proposed:
Add more desertion type values to distinguish them Battleground desertion vs Arena desertion

## Issues Addressed:
No issues

## Tests Performed:
Laasker did some tests using 3v3soloQ module

## How to Test the Changes:
A successful build should be enough.

The web app "[BG queue abuser viewer](https://github.com/Helias/BG-Queue-Abuser-Viewer)" could be moved under azerothcore organization and it should be updated with these new types, but it's not blocking for this PR 

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -637,13 +637,8 @@ void WorldSession::LogoutPlayer(bool save)
                     sScriptMgr->OnBattlegroundDesertion(_player, BG_DESERTION_TYPE_INVITE_LOGOUT);
                 }
 
-                if (bgQueueTypeId >= BATTLEGROUND_QUEUE_2v2 && bgQueueTypeId < MAX_BATTLEGROUND_QUEUE_TYPES)
-                {
-                    if (_player->IsInvitedForBattlegroundQueueType(bgQueueTypeId))
-                    {
-                        sScriptMgr->OnBattlegroundDesertion(_player, ARENA_DESERTION_TYPE_INVITE_LOGOUT);
-                    }
-                }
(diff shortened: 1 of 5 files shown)


src/server/game/Server/WorldSession.cpp
```diff
@@ -637,6 +637,9 @@ void WorldSession::LogoutPlayer(bool save)
                     sScriptMgr->OnBattlegroundDesertion(_player, BG_DESERTION_TYPE_INVITE_LOGOUT);
                 }
 
+                if (bgQueueTypeId >= BATTLEGROUND_QUEUE_2v2 && bgQueueTypeId < MAX_BATTLEGROUND_QUEUE_TYPES && _player->IsInvitedForBattlegroundQueueType(bgQueueTypeId))
+                    sScriptMgr->OnBattlegroundDesertion(_player, ARENA_DESERTION_TYPE_INVITE_LOGOUT);
+
```
sudlud, line 642: this relies on the order of entries in `BattlegroundQueueTypeId` to detect if it's an arena match? is there probably a "nicer" way to decide if it's arena? also `BattlegroundQueueTypeId` currently has a big gap between `BATTLEGROUND_QUEUE_5v5`  and `MAX_BATTLEGROUND_QUEUE_TYPES`, adding entries there would also trigger this call.
but maybe it's just irrelevant

Helias: it's done on purpose, so we can add more custom entries between those two values

sudlud (approved):

Issue #18787, opened 2024-04-22
Fears and Roots breaking on damage behavior


balleny:
### Current Behaviour

abilities like fear https://wowgaming.altervista.org/aowow/?search=fear#abilities and roots https://wowgaming.altervista.org/aowow/?search=roots#abilities should break if the target receive damage (not on 1 damage but certain threshold).

this behavior right now seems to be based on an amount relative to the total hp of the affected target. this results e.g. in cases were warlocks can fear elite mobs and solo them a lot easier than supposed, or using a druids root on Lady Vashjs Striders which rarely break

### Expected Blizzlike Behaviour

there should surely be multiple triggers for the effect to end, including the damage taken relative to the hp of the affected target AND the absolute damage taken, etc.

### Source

retail gaming experience

### Steps to reproduce the problem

root or fear high hp target, deal damage and observe cc effect not breaking

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/e2ecbf44cc19450109bee1ef7ce6aaa7621c5af4

### Operating system

ubuntu

amed80:
I remember finding it weird that [Socrethar](https://www.wowhead.com/wotlk/npc=20132/socrethar) (350k hp) never broke out from [Turn Evil](https://www.wowhead.com/wotlk/spell=10326/turn-evil) , every other mob (with less HP) it break after some hits but this specific elite didn't break once when i tried to solo him the first time.

balleny:
because currently this mechanic seems to respect only % damage taken but not absolute values or some mechanic like it was common classic which was called "heartbeat" (but that was a classic thing only e.g. with mind control)

sogladev:
I've listed my research below. So far `2` is the most accurate and straight forward to implement. Not sure how accurate `3` is?

### 1. Heartbeat resist
This requires 2 changes: implement heartbeat[1] (5s timer per unit) and make use of `SPELL_ATTR0_HEARTBEAT_RESIST`[2]

Looks like progress has been made in TC335 (unmerged) to implement these 

https://www.github.com/TrinityCore/TrinityCore/pull/22380
https://www.github.com/TrinityCore/TrinityCore/pull/25822
https://www.github.com/TrinityCore/TrinityCore/pull/20206


In cmangos-wotlk these changes are present:
[1] Sep 16, 2021 cmangos implementation of a TC335 unmerged/closed PR 
[...]

PR #20153, opened 2024-10-06
fix(Core/SpellAuraEffects): use caster's level to scale amount to break crowd control auras

sogladev:
## Changes Proposed:

This PR proposes changes to:

CC Auras which use their amount to drop are using 10% of the target's hp as a threshold. Propose to use an amount that scales with caster's level instead based on cmangos' implementation

new values for these levels:
lvl80 2650
lvl70 1890

I left the 10% line in case caster is null?

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/18787

## SOURCE:

per wiki 
> Fear effects have a damage limit, which appears to be based on level. Once the target takes a certain amount of damage, the fear effect will be broken, regardless of any remaining duration.  https://wowpedia.fandom.com/wiki/Fear_effect

[cmangos commit introducing this kind of calculation](https://github.com/cmangos/mangos-wotlk/commit/53efce2202df71a3c96986d607ce5f91873ff38d#diff-b43c43801be9ee8ddc22dee662e046f00dfe78de6a540f40be86314aa74a19bdR116)

## Tests Performed:

This PR has been:

tested with debug

## How to Test the Changes:

vashj strider

add fear immune
.cast 22663
.npc add temp 22056

1. go find an elite with large amount of hp e.g. vashj strider
2. fear it
3. observe it now breaks easier

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -500,6 +500,7 @@ int32 AuraEffect::CalculateAmount(Unit* caster)
             amount = int32(GetBase()->GetUnitOwner()->CountPctFromMaxHealth(10));
             if (caster)
             {
+                amount = sObjectMgr->GetCreatureBaseStats(caster->GetLevel(), Classes::CLASS_WARRIOR)->BaseHealth[EXPANSION_WRATH_OF_THE_LICH_KING] / 4.75f;
                 // Glyphs increasing damage cap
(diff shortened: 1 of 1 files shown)


Grimdhex:
You could remove the line with CountPctFromMaxHealth but maybe add a check to return if the caster is not valid. 

There are no sense to continue the calculation without valid caster.

Grimdhex (approved):

Issue #16884, opened 2023-08-02
Mounts: Magic Rooster can be summoned with other monts 


wowquester:
### Current Behaviour

I know i know, you may say that this mount is not available in game, but it is still part of the game and it seem to be bugged. Just try it yourself. When you are on this mount (Magic Rooster), you can summon any other mount and the icon is still there. Also, if you summon any other mount and then magic rooster, the indicator will remain. 
![obrazek](https://github.com/azerothcore/azerothcore-wotlk/assets/138876396/1c152ccf-93e8-4468-9210-5bd874ab0ad4)

### Expected Blizzlike Behaviour

Chances are it should act as any other mount. 

### Source

It is a mount. 

### Steps to reproduce the problem

.add 49290
Summon mount.
Summon any other mount.

### AC rev. hash/commit

bda9b404038e

### Operating system

w11

heyitsbench:
Unfortunately this is a blizzlike bug. Please see [this issue for details](https://github.com/azerothcore/azerothcore-wotlk/issues/15985).

Issue #15985, opened 2023-04-17
(Spells): Magic Rooster is castable while mounted.


heyitsbench:
### Current Behaviour

[Magic Rooster](https://wowgaming.altervista.org/aowow/?spell=65917) is castable while mounted.

### Expected Blizzlike Behaviour

This mount should not be castable while mounted.

### Source

Assumption, but it seems likely that it would have been mentioned somewhere, like Wowhead comments if it was.

### Steps to reproduce the problem

1. `.additem 49290`, `.additem 8631`
2. Learn the mounts.
3. Cast the saber mount.
4. Cast the rooster mount.
5. Observe both mounts active.

### Extra Notes

There is an [open PR](https://github.com/TrinityCore/TrinityCore/pull/28902) for TC's 3.3.5 branch to remove this flag.

### AC rev. hash/commit

9fcacc5c06910305e0175445f486aca77a9e7c92

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
This apparently [is a blizzlike bug that was found during retail Wrath](https://www.wowhead.com/wotlk/item=49290/magic-rooster-egg#comments:id=929888), no idea if this is a case we want to correct.

CC: @Kitzunu

Kitzunu:
harmless bug

PR #20164, opened 2024-10-09
fix(Core/SpellGeneric) Magic Rooster can be cast while mounted

pangolp:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/AdonisWOW/bugtracker/issues/12
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15985
- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/16134
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16884
## SOURCE:

- https://www.wowhead.com/wotlk/item=49290/magic-rooster-egg#comments:id=929888
```
Interestingly enough, you can mount to this mount from another and not be dismounted on the cast.
Probably a glitch, but all other mounts will dismount your current mount before mounting you on your new one.
```

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Learning the mammoth with sellers (`.additem 44235` or `.additem 44234`)
2. Learning to saddle a chicken (`.additem 46778` or `.additem 49290`)
3. While on the mammoth, cast the chicken mount (although it also works on other mounts, such as flying ones, allowing you to fly with the chicken, when it shouldn't be like that or use the mammoth vendors on it)
4. Applying the fix does not prevent the player from using the chicken on other mounts, but it does prevent it from having both auras.
5. The other solution would be to prevent the chicken from being used in this case, while already mounted. But I would have to test if this error is also occurring on the broom of the Halloween event.

```cpp
ApplySpellFix({
    47977, // Magic Broom
    65917  // Magic Rooster
    }, [](SpellInfo* spellInfo)
{
    // First two effects apply auras, which shouldn't be there
    // due to NO_TARGET applying aura on current caster (core bug)
    // Just wipe effect data, to mimic blizz-behavior
    spellInfo->Effects[EFFECT_0].Effect = 0;
    spellInfo->Effects[EFFECT_1].Effect = 0;
    spellInfo->Attributes &= ~SPELL_ATTR0_ALLOW_WHILE_MOUNTED;
});
```

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1728482907199736600.sql
+++ b/data/sql/updates/pending_db_world/rev_1728482907199736600.sql
@@ -1,0 +1,4 @@
+--
+DELETE FROM `spell_script_names` WHERE `spell_id`=65917;
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(65917, 'spell_magic_rooster_egg');
--- a/src/server/scripts/Spells/spell_generic.cpp
+++ b/src/server/scripts/Spells/spell_generic.cpp
@@ -5311,6 +5311,30 @@ class spell_gen_steal_weapon : public AuraScript
     }
 };
 
+// 65917 - Magic Rooster
+// Due to a flaw in the spell design
+// The mount can be used, they are on another mount
+class spell_magic_rooster_egg : public SpellScript
+{
+    PrepareSpellScript(spell_magic_rooster_egg);
+
+    void HandleBeforeCast()
+    {
+        Unit* caster = GetCaster();
+
+        if (caster)
+        {
+            caster->Dismount();
+            caster->RemoveAurasByType(SPELL_AURA_MOUNTED);
+        }
+    }
+
+    void Register() override
+    {
+        BeforeCast += SpellCastFn(spell_magic_rooster_egg::HandleBeforeCast);
+    }
+};
+
 void AddSC_generic_spell_scripts()
 {
     RegisterSpellScript(spell_silithyst);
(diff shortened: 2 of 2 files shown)


heyitsbench:
https://github.com/azerothcore/azerothcore-wotlk/pull/16134 for reference.

pangolp:
> #15985, #16134 for reference.

Thanks.

pangolp:
The script isn't bad, because blizzlike talking, the mount can be sung over another mount (which is strange). Preventing casting, maybe would be the best, but it wouldn't be blizzlike the solution anymore. But what is clear is that the player cannot have the 2 auras, because if he castes a flying mount and then the chicken, he could fly with the second, when that should not be possible or even move faster on the water, if he castes the turtle mount.

heyitsbench:
Current behavior prior to this PR *is* Blizzlike.

Nyeriah:
It is a blizzlike bug and we have decided not to fix it in another instance this was brought up.

pangolp:
Apply the solution to it then in a particular way, because while it's not serious at all, some things look a little weird.

pangolp:
Anyway, I'm not preventing what might caste the frame, because I couldn't make that. All I do is I prevent it from having the 2 auras. I'll add it to a module then and use it that way.

pangolp:
If we put a variable in the worldserver.conf either? The module remains acceptable as a solution. Anything, close the PR or let me know to close it. Thank you.

Kitzunu:
Blizzlike bug which is not gamebreaking

pangolp:
I respect the decision to preserve the blizzlike code, so I just created a module, for all those who want to prevent this from happening. Greetings. https://github.com/pangolp/mod-non-blizzlike-fixes

Issue #20111, opened 2024-09-30
[DK Starter Chain Quest] A Special Surprise. Some issues.


Rorschach91:
### Current Behaviour

Quest can be completed, but:
All npcs in the room can be killed by the player.
Player's race npc can be instantly killed.
All Npcs have a chance to dodge, parry or miss your attack.

Evidence: [Youtube](https://youtu.be/VvS0Nq_GOtA)

### Expected Blizzlike Behaviour

You can only kill your race's npc.
You can be able to kill your race's npc only when it finishes its dialogue line.
Npcs shouldn't be able to parry, dodge or miss your attack, they are basically dummies.

How it should work: [Youtube](https://www.youtube.com/watch?v=9lzb-Gz5bD4&t)

### Steps to reproduce the problem

1. Start the quest.
2. Go into the house

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/784e80668d5b57640a4c7a4d6e2c7b898e4f804d

### Operating system

Windows 11

TheSCREWEDSoftware:
Also to add to your video, not happening to the same race as but to the gnome, double checking (whenever this get fixed) making sure that none of the NPCs can actually parry or block your attack (like the gnome did).

Rorschach91:
> Also to add to your video, not happening to the same race as but to the gnome, double checking (whenever this get fixed) making sure that none of the NPCs can actually parry or block your attack (like the gnome did).

I checked again and I can confirm that all npcs have a chance to parry, dodge or miss an attack.

PR #20178, opened 2024-10-10
Fix(Script) A Special Surprise Npcs issues.

Rorschach91:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20111#issue-2556469731

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Create a DK.
2. Use .mod phase 4
3. Go in the right area and use .quest add id quest ([here](https://www.wowhead.com/wotlk/quest=12739/a-special-surprise=))
4. Check the quest and npcs inside the house.

- [ ]
- [ ]


src/server/scripts/EasternKingdoms/ScarletEnclave/chapter2.cpp
```diff
@@ -623,7 +623,7 @@ class npc_a_special_surprise : public CreatureScript
             ExecuteSpeech_Counter = 0;
             PlayerGUID.Clear();
 
-            me->SetImmuneToPC(false);
+            me->SetUnitFlag(UNIT_FLAG_NON_ATTACKABLE);
```
Nefertumm, line 626:
```suggestion
            me->SetUnitFlag(UNIT_FLAG_NON_ATTACKABLE);
```

data/sql/updates/pending_db_world/Asp_Npcs_Flags_change.sql
```diff
@@ -0,0 +1,30 @@
+
+UPDATE `creature_template` SET `unit_flags` = `unit_flags` & ~(256|512) WHERE `entry` = 29072;
+UPDATE `creature_template` SET `unit_flags` = `unit_flags` |2|131072 WHERE `entry` = 29072;
```
heyitsbench, line 3: These changes do not match sniffs, observed value in a 9.1.5 41079 sniff (when walking up to the creature before the RP begins, so probably default flags) is 33536 (256+512+32768) as it currently is on AC.

src/server/scripts/EasternKingdoms/ScarletEnclave/chapter2.cpp
```diff
@@ -623,7 +623,8 @@ class npc_a_special_surprise : public CreatureScript
             ExecuteSpeech_Counter = 0;
             PlayerGUID.Clear();
 
-            me->SetImmuneToPC(false);
+            me->RemoveUnitFlag(UNIT_FLAG_PACIFIED);
+            me->SetUnitFlag(UNIT_FLAG_IMMUNE_TO_PC);
```
heyitsbench, line 627:
```suggestion
            me->SetImmuneToPC(true);
```
This should be all that's required. This pacified unit flag doesn't get touched in sniffs either. We also prefer to use helper methods instead of setting/removing the raw flags ourselves.

heyitsbench: On second glance, maybe instead of using the pacified flag, just setting their react state to passive would accomplish a similar effect.

heyitsbench (approved):

Issue #20149, opened 2024-10-06
[Creature Spell] creature_template_spell "index" = 0 will not appear and work


kissingers:
### Current Behaviour

creature_template_spell "index" = 0 will not appear and work.
the creature 25653, if you try control the blue dragon and see the spell, the index 0 spell in Database will not appear
but if change index big than 0, such as 4, will appear. the bug about 2 month. and before all ok.

### Expected Blizzlike Behaviour

N/A

### Steps to reproduce the problem

.go c 55066
fight with the boss kljd and wait 2m, try control the blue dragon and see the spell, the index 0 spell in Database will not appear

### AC rev. hash/commit

AzerothCore rev. 01b86ec7a6c9 2024-10-04 11:01:36 -0300 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 1.
Connection peak: 0.
服务器运行时间: 45 second(s)
Update time diff: 100ms. Last 38 diffs summary:
|- Mean: 101ms
|- Median: 100ms
|- Percentiles (95, 99, max): 101ms, 133ms, 152ms
Using SSL version: OpenSSL 3.0.14 4 Jun 2024 (library: OpenSSL 3.0.14 4 Jun 2024)
Using Boost version: 1.74.0
Using CMake version: 3.25.1
Using MySQL version: 80402
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.8.12-1-pve
Worldserver listening connections on port %u
Realmlist (Realm Id: 3) configured in port 54347
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in /root/wow3/bin/maps. Total size: 291014339 bytes
vmaps directory located in /root/wow3/bin/vmaps. Total size: 658130553 bytes
mmaps directory located in /root/wow3/bin/mmaps. Total size: 2192911264 bytes
Default DBC locale: zhCN.
All available DBC locales: zhCN 
Using World DB: ACDB 335.12-dev
Latest LoginDatabase update: 2024_01_20_00.sql
Latest CharacterDatabase update: 2024_09_22_00.sql
Latest WorldDatabase update: mod-zone-difficult.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-difficult
|- mod-eluna

### Operating system

Debian 12 with PVE

[...]

PR #20227, opened 2024-10-16
fix(Core/CharmInfo): Adjust how spells are given to charmed creatures.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20149

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/src/server/game/Entities/Unit/CharmInfo.cpp
+++ b/src/server/game/Entities/Unit/CharmInfo.cpp
@@ -78,21 +78,7 @@ void CharmInfo::InitPossessCreateSpells()
 {
     if (_unit->IsCreature())
     {
-        // Adding switch until better way is found. Malcrom
-        // Adding entrys to this switch will prevent COMMAND_ATTACK being added to pet bar.
-        switch (_unit->GetEntry())
-        {
-            case 23575: // Mindless Abomination
-            case 24783: // Trained Rock Falcon
-            case 27664: // Crashin' Thrashin' Racer
-            case 40281: // Crashin' Thrashin' Racer
-            case 23109: // Vengeful Spirit
-            case 25653: // Power of the Blue Flight
-                break;
-            default:
-                InitEmptyActionBar();
-                break;
-        }
+        InitEmptyActionBar(false);
 
         for (uint32 i = 0; i < MAX_CREATURE_SPELLS; ++i)
         {
@@ -160,11 +146,14 @@ bool CharmInfo::AddSpellToActionBar(SpellInfo const* spellInfo, ActiveStates new
 {
     uint32 spell_id = 0;
     uint32 first_id = 0;
+    bool autocastable = false;
 
     if (spellInfo)
     {
         spell_id = spellInfo->Id;
         first_id = spellInfo->GetFirstRankSpell()->Id;
+        if (spellInfo->IsAutocastable())
+            autocastable = true;
     }
 
     // new spell rank can be already listed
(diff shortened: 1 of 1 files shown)


heyitsbench:
@kissingers Please give this PR a test if/when you're able. ❤️

kissingers:
Test OK, but might have some little problem:
By default now will not display normal attack icon.
If change index 3 to 6(total 0-3), the 6 spell will not close the spell 2
image

kissingers:
image

heyitsbench:
I don't understand the expected behavior of your message.

kissingers:
It work, but I think if there is need a mealattack icon at last pision not olny just replace by spell at index 0
and if index not continue, just 0-2 then 6, if can automove 6 near 2?
image

heyitsbench:
I don't believe there is any official instance of a creature having their melee attack in any position other than index 0 (1 on hotbar). If there's evidence to the contrary, then it would need to be investigated, but I've yet to see it myself.

kissingers:
> I don't believe there is any official instance of a creature having their melee attack in any position other than index 0 (1 on hotbar). If there's evidence to the contrary, then it would need to be investigated, but I've yet to see it myself.

I haven't noticed if that's the case either, but I'm assuming there's a melee fight where the priest controls some creature? I try test both fix and not fix.

kissingers:
after fix
image

before fix
image

before fix might have spell idex problem, but the creature can not attack after fix by press keyboard, because the spell 2 depend the next meele attack, and there no meele attack ico(but can use mouse left click to start attack)

heyitsbench:
Cases like that need to have the index manually adjusted.

kissingers:
just re-index all spell, and if index does not exist, skip index it, and add meele attack at end.
or fist add meele attack, then index the spell except does not exist

heyitsbench:
> and add meele attack at end.

But this isn't correct, again, I haven't seen any creature in the game with a melee attack anywhere but index 0. For the creatures that are supposed to have melee attack available, they need to have their indices manually adjusted, as either way their data would be wrong.

kissingers:
> > and add meele attack at end.
> 
> But this isn't correct, again, I haven't seen any creature in the game with a melee attack anywhere but index 0. For the creatures that are supposed to have melee attack available, they need to have their indices manually adjusted, as either way their data would be wrong.

might left the keyboard 1 not add other spell, and add spell index 0 to key beboard 2 , and others flow if exist, and not left blank

heyitsbench:
> how about Blizzlike behaviour?

Blizzlike would have index 0 be clear for any creature they want to have melee attack available for. If they want a spell in index 0 (see all of the manual creature overrides this PR removes), then it always overrides the melee attack in index 0. Again, I have never seen the melee attack in any index other than 0.

kissingers:
> > how about Blizzlike behaviour?
> 
> Blizzlike would have index 0 be clear for any creature they want to have melee attack available for. If they want a spell in index 0 (see all of the manual creature overrides this PR removes), then it always overrides the melee attack in index 0. Again, I have never seen the melee attack in any index other than 0.

So the fix is suite for the bug. how about the blank keyboard? might  Blizzlike database will always not let table index not continue?

heyitsbench:
> [...] how about the blank keyboard? might Blizzlike database will always not let table index not continue?

I don't understand the question.

kissingers:
if index only have 0-2 then skip to 6-7, not have 3-5, will Blizzlike keep the keyboard 4-6 blank?

heyitsbench:
> if index only have 0-2 then skip to 6-7, not have 3-5, will Blizzlike keep the keyboard 4-6 blank?

Yes, can be seen [here](https://www.youtube.com/watch?v=g5Lo1PPLz5k&t=6s) with Death Knight Understudies and [here](https://www.youtube.com/watch?v=fVT3gYZM2AI&t=77s) with Vengeful Spirits.

kissingers:
So it all ok, very thanks again.

Nyeriah (approved):

PR #20245, opened 2024-10-19
Revert "fix(Core/Spells): SPELL_ATTR1_IMMUNITY_PURGES_EFFECT should not remov…"

Nyeriah:
Reverts azerothcore/azerothcore-wotlk#20032

Causes issues with warrior instances. Cant think of a solution to this atm.


heyitsbench:
This also resolves a server crash when attempting to dismount vehicles such as the death knight starting experience Frostbrood Vanquisher.

sudlud (approved):

Issue #20240, opened 2024-10-18
Potential memory leak in  worldserver


PaoloBenigni:
### Current Behaviour

The ram consumption of the worldserver is continually increasing even when players have never logged into the system. 
![Picture1](https://github.com/user-attachments/assets/dbde9758-49d7-4bde-b017-f6ee5ef9db93)

Analysis of the heap memory shows that the big spikes in memory usage are related to the map manager.
`MMapMgr.cpp, line 174 (C:\Temp\source\azerothcore-wotlk\src\common\Collision\Management\MMapMgr.cpp)`

However what seems unusual is that there is an ever increasing usage of memory from the RandomMovementGenerator.
`RandomMovementGenerator.cpp, line 73 (C:\Temp\source\azerothcore-wotlk\src\server\game\Movement\MovementGenerators\RandomMovementGenerator.cpp)`
The final number of hits in memory for this was 223,300. I am unsure if this is expected behavior, but was expecting that there would be a 1:1 relationship between creatures and generators, or at least that it wouldnt increase after everything was properly loaded. 

Here is the full stack trace. 
```
ntdll.dll!RtlAllocateHeap 00007ffd2801a9a0		
ucrtbase.dll!malloc_base + 0x30 bytes 00007ffd257ffde0		
worldserver.exe!operator new Line 36 + 0x3 bytes 00007ff68d838f36	D:\a\_work\1\s\src\vctools\crt\vcstartup\src\heap\new_scalar.cpp, line 36	
worldserver.exe!RandomMovementGenerator<Creature>::_setRandomLocation Line 73 + 0x79 bytes 00007ff68d5be849	C:\Temp\source\azerothcore-wotlk\src\server\game\Movement\MovementGenerators\RandomMovementGenerator.cpp, line 73	
worldserver.exe!RandomMovementGenerator<Creature>::DoUpdate Line 298 + 0x6 bytes 00007ff68d5bdc2a	C:\Temp\source\azerothcore-wotlk\src\server\game\Movement\MovementGenerators\RandomMovementGenerator.cpp, line 298	
worldserver.exe!MotionMaster::UpdateMotion Line 114 + 0x17 bytes 00007ff68d1ab0dc	C:\Temp\source\azerothcore-wotlk\src\server\game\Movement\MotionMaster.cpp, line 114	
[...]

PR #20258, opened 2024-10-21
fix(Core/Movement): fix a small memory leak in RandomMovementGenerator

Grimdhex:
Thank to Ovahlord for the catch.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20240

## SOURCE:

## Tests Performed:

This PR has been:
- [] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

1.
2.
3.

- [ ] Possibly don't fix the issue but the catch is correct
- [ ] 

### DIFF
--- a/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
@@ -27,17 +27,16 @@
 #include "Util.h"
 
 template<class T>
-RandomMovementGenerator<T>::~RandomMovementGenerator() { }
+RandomMovementGenerator<T>::~RandomMovementGenerator() = default;
 
-template<>
-RandomMovementGenerator<Creature>::~RandomMovementGenerator()
-{
-    delete _pathGenerator;
-}
+template RandomMovementGenerator<Creature>::~RandomMovementGenerator();
 
 template<>
 void RandomMovementGenerator<Creature>::_setRandomLocation(Creature* creature)
 {
+    if (!creature)
+        return;
+
     if (creature->_moveState != MAP_OBJECT_CELL_MOVE_NONE)
         return;
 
@@ -135,7 +134,7 @@ void RandomMovementGenerator<Creature>::_setRandomLocation(Creature* creature)
         else // ground
         {
             if (!_pathGenerator)
-                _pathGenerator = new PathGenerator(creature);
+                _pathGenerator = std::make_unique<PathGenerator>(creature);
             else
                 _pathGenerator->Clear();
 
@@ -242,6 +241,8 @@ void RandomMovementGenerator<Creature>::DoInitialize(Creature* creature)
     if (!_wanderDistance)
         _wanderDistance = creature->GetWanderDistance();
 
+    _pathGenerator = nullptr;
+
     _nextMoveTime.Reset(creature->GetSpawnId() && creature->GetWanderDistance() == _wanderDistance ? urand(1, 5000) : 0);
     _wanderDistance = std::max((creature->GetWanderDistance() == _wanderDistance && creature->GetInstanceId() == 0) ? (creature->CanFly() ? MIN_WANDER_DISTANCE_AIR : MIN_WANDER_DISTANCE_GROUND) : 0.0f, _wanderDistance);
 
@@ -280,6 +281,7 @@ bool RandomMovementGenerator<Creature>::DoUpdate(Creature* creature, const uint3
     {
         _nextMoveTime.Reset(0);  // Expire the timer
         creature->StopMoving();
+        _pathGenerator = nullptr;
         return true;
     }
 
--- a/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
@@ -241,7 +241,7 @@ void RandomMovementGenerator<Creature>::DoInitialize(Creature* creature)
     if (!_wanderDistance)
         _wanderDistance = creature->GetWanderDistance();
 
-    _pathGenerator = nullptr;
+    _pathGenerator.reset();
 
     _nextMoveTime.Reset(creature->GetSpawnId() && creature->GetWanderDistance() == _wanderDistance ? urand(1, 5000) : 0);
     _wanderDistance = std::max((creature->GetWanderDistance() == _wanderDistance && creature->GetInstanceId() == 0) ? (creature->CanFly() ? MIN_WANDER_DISTANCE_AIR : MIN_WANDER_DISTANCE_GROUND) : 0.0f, _wanderDistance);
@@ -281,7 +281,7 @@ bool RandomMovementGenerator<Creature>::DoUpdate(Creature* creature, const uint3
     {
         _nextMoveTime.Reset(0);  // Expire the timer
         creature->StopMoving();
-        _pathGenerator = nullptr;
+        _pathGenerator.reset();
         return true;
     }
 
--- a/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
@@ -27,7 +27,7 @@
 #include "Util.h"
 
 template<class T>
-RandomMovementGenerator<T>::~RandomMovementGenerator() = default;
+RandomMovementGenerator<T>::~RandomMovementGenerator() { }
 
 template RandomMovementGenerator<Creature>::~RandomMovementGenerator();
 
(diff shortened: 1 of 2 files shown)


Grimdhex:
@PaoloBenigni could you test the PR ?

PaoloBenigni:
On it

walkline:
In general, refactoring it to use smart pointers is not a bad idea, but do you see a scenario where `_pathGenerator` could leak? I don't see one.

As for the crash, I don't think that check is necessary. The movement generator is bound to the creature's lifetime, so I would expect the pointer to be valid. Null cases should be handled at higher layers (if any).

As I mentioned in the Discord channel, the issue that @PaoloBenigni brought up is related to `_preComputedPaths`. It's a large cache, containing, I believe, 144 possible paths. Due to the randomness involved, filling in the cache takes a lot of time, and during this process, allocations are occurring.

kissingers:
Comparing TC, found that there are a lot of differences,  the AZ does a lot of calculations on random moves, which may be the reason why the CPU load is higher.

https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp

PaoloBenigni:
@walkline It seems like the calls are two types of object in memory, one related to line 72, and another in 137. The _precomputed is smaller, but there are many more hits than line 137.  Here are the relevant memory allocations. 


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/Paolo/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/Paolo/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">


File | Line | Hits | Bytes | Bytes/hit
-- | -- | -- | -- | --
RandomMovementGenerator.h | 35 | 177 | 11328 | 64
RandomMovementGenerator.cpp | 72 | 254846 | 16310144 | 64
RandomMovementGenerator.cpp | 137 | 176 | 171776 | 976



</body>

</html>

Grimdhex:
> In general, refactoring it to use smart pointers is not a bad idea, but do you see a scenario where `_pathGenerator` could leak? I don't see one.
> 
> 
> 
> As for the crash, I don't think that check is necessary. The movement generator is bound to the creature's lifetime, so I would expect the pointer to be valid. Null cases should be handled at higher layers (if any).
> 
> 
> 
> As I mentioned in the Discord channel, the issue that @PaoloBenigni brought up is related to `_preComputedPaths`. It's a large cache, containing, I believe, 144 possible paths. Due to the randomness involved, filling in the cache takes a lot of time, and during this process, allocations are occurring.

That why I note that will maybe don't fix the issue. But it's stay a safe case. We can't say that we expert with the management of our pointer (and the core code in general). Samething for the check. All of this cost nothing and avoid weird issues (or just mistake by our).

PaoloBenigni:
Observations so far
image
Size of objects in memory is the same as before, It looks like there are fewer overall in memory. 
I poked around the world and didnt see any issues gameside.

Grimdhex:
For _precuptedPath, honestly SetRandomPosition is so ugly that I'm not going to bother trying to figure out where it's stuck. I've imported TC's function, which is much cleaner and gets rid of a lot of variables. I've also made a few personal improvements to the rest of the class

I've tested 5min in games, and it doesn't seem to cause any crash.

PaoloBenigni:
Ill pull and build now.

walkline:
> Comparing TC, found that there are a lot of differences, the AZ does a lot of calculations on random moves, which may be the reason why the CPU load is higher.
> 
> https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp

> Comparing TC, found that there are a lot of differences, the AZ does a lot of calculations on random moves, which may be the reason why the CPU load is higher.
> 
> https://github.com/TrinityCore/TrinityCore/blob/master/src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp

Your assumption regarding CPU consumption is wrong.

I compared speed of this TC port and the current implementation. Here is results:

| Name | median | average |
|---|---|---|
| This TC port | 58.292 µs |  70.297 µs |
| AC cache miss | 50.355 µs |  54.416 µs |
| AC cache hit | 15.125 µs |  16.182 µs |

The data is based on over 1000 calculations.

PaoloBenigni:
I tested commit [37af537](https://github.com/azerothcore/azerothcore-wotlk/pull/20258/commits/37af537722e3018e6a5ba1147f84798fcc536dab) and found it to work. I teleported around and saw npcs moving around in what appeared a normal way.

Grimdhex:
@walkine has absolutely right. TC doesn't seem to fire “less” calculation, it's just that they are attributed to other Core functions and not done directly in the method. So it's don't optimize the CPU calculation.

The only advantage with this PR is a more readable and understandable code and the removal of all those rather cumbersome variables that did seem to be the cause of this memory increase. But if someone feel to rewrite the class to optimize also the CPU usage, do it :)

PaoloBenigni:
Tested latest. Built with VS. No observed issues.

walkline:
> @walkine has absolutely right. TC doesn't seem to fire “less” calculation, it's just that they are attributed to other Core functions and not done directly in the method. So it's don't optimize the CPU calculation.
> 
> The only advantage with this PR is a more readable and understandable code and the removal of all those rather cumbersome variables that did seem to be the cause of this memory increase. But if someone feel to rewrite the class to optimize also the CPU usage, do it :)
> 
> Otherwise, ready to be review :)

I agree that the TC solution is easier to read and simple like bubble sort. The issue that @PaoloBenigni brought up doesn’t seem to be a problem, but rather intended behavior. It also appears that there’s a config option to disable the cache for the random movement generator if you’re limited on RAM.

Do we have any other issues with the current implementation? If not, then why consider changing the current efficient implementation to something less efficient, but easier to read (like bubble sort)?

Grimdhex:
> > @walkine has absolutely right. TC doesn't seem to fire “less” calculation, it's just that they are attributed to other Core functions and not done directly in the method. So it's don't optimize the CPU calculation.
> 
> > 
> 
> > The only advantage with this PR is a more readable and understandable code and the removal of all those rather cumbersome variables that did seem to be the cause of this memory increase. But if someone feel to rewrite the class to optimize also the CPU usage, do it :)
> 
> > 
> 
> > Otherwise, ready to be review :)
> 
> 
> 
> I agree that the TC solution is easier to read and simple like bubble sort. The issue that @PaoloBenigni brought up doesn’t seem to be a problem, but rather intended behavior. It also appears that there’s a config option to disable the cache for the random movement generator if you’re limited on RAM.
> 
> 
> 
> Do we have any other issues with the current implementation? If not, then why consider changing the current efficient implementation to something less efficient, but easier to read (like bubble sort)?

@PaoloBenigni do you have test by disabling the option ?

PaoloBenigni:
Ill test this now, but which part of the change is getting stuck with licensing issues?

FrancescoBorzi:
@azerothcore/developers

src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
```diff
@@ -242,6 +241,8 @@ void RandomMovementGenerator<Creature>::DoInitialize(Creature* creature)
     if (!_wanderDistance)
         _wanderDistance = creature->GetWanderDistance();
 
+    _pathGenerator = nullptr;
```
Winfidonarleyan, line 244:
```suggestion
    _pathGenerator.reset();
```

src/server/game/Movement/MovementGenerators/RandomMovementGenerator.cpp
```diff
@@ -280,6 +281,7 @@ bool RandomMovementGenerator<Creature>::DoUpdate(Creature* creature, const uint3
     {
         _nextMoveTime.Reset(0);  // Expire the timer
         creature->StopMoving();
+        _pathGenerator = nullptr;
```
Winfidonarleyan, line 284:
```suggestion
        _pathGenerator.reset();
```

Winfidonarleyan (approved):

Ovahlord:
Don't use std::unique_ptr::reset(); like this... just go with = nullptr.

Also these reset calls are redundant and can be removed entirely. If you allocate a new object with std::make_unique, the previous object is automatically being destroyed and no longer needs manual management.

The pathGenerator class is doing nothing on its own so there is no need to destroy it unless you need a new one for a new path calculation. So tldr: kill the reset calls. Unnecessary memory meddling

PR #20457, opened 2024-11-06
Exception handling for DB failure

AaronGulman:
Hi,
This PR is aimed at exception handling for the DB initialization process within the server's authentication module. By wrapping the `StartDB()` function in a try-catch block, this change ensures that any exceptions thrown during DB startup are properly caught and logged, preventing the server from failing silently if the database connection cannot be established. Specifically, it logs an error message and exits with code 1 if the initialization fails.

The change has been tested by me in a controlled dev env to ensure the exception handling works as expected.

I hope you find this tiny change useful ! I apologize if it is not (I am mostly a full stack web dev!) 

Good luck !

### DIFF
--- a/src/server/apps/authserver/Main.cpp
+++ b/src/server/apps/authserver/Main.cpp
@@ -118,8 +118,14 @@ int main(int argc, char** argv)
     }
 
     // Initialize the database connection
+   try {
     if (!StartDB())
         return 1;
+} catch (const std::exception& ex) {
+    LOG_ERROR("server.authserver", "Database initialization failed: {}", ex.what());
+    return 1;
+}
+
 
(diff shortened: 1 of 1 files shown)


src/server/apps/authserver/Main.cpp
```diff
@@ -118,8 +118,14 @@ int main(int argc, char** argv)
     }
 
     // Initialize the database connection
+   try {
     if (!StartDB())
         return 1;
+} catch (const std::exception& ex) {
+    LOG_ERROR("server.authserver", "Database initialization failed: {}", ex.what());
+    return 1;
+}
```
FrancescoBorzi, line 127: hello and thanks for your contribution! can you please fix the code indentation?

PR #20470, opened 2024-11-06
chore(DB): move 2023 update files

FrancescoBorzi:


Kitzunu:
You need to set the files as ARCHIVED in https://www.azerothcore.org/wiki/updates

FrancescoBorzi:
@Kitzunu done

Nyeriah (approved):

Nyeriah:
Do we need to update the base database files before this?/along with this? @Kitzunu @FrancescoBorzi

Nyeriah:
Just afraid we might stash files not included in the base database files. This is confusing :/

Kitzunu:
Anything 2023 is in base.
But we might want to update `updates.sql` in the base to reflect these changes

Kitzunu:
I agree it is confusing, as it has never been done before 😅 Who know what might happen.

But I don't think it will cause any issues and the base `updates` file says all these files are already imported anyways. So it will still only run files that are not in the table, so all files created after the last squash

FrancescoBorzi:
> Anything 2023 is in base. But we might want to update `updates.sql` in the base to reflect these changes

yes, in my opinion that should have been done already when squashing those updates inside the base files. @Kitzunu can you please take care of it?

sudlud:
just tested this on an existing db_world and a freshly generated db_world without any issues.

the base `updates.sql` does **not** need an update for this PR imo.

the updater automatically recognizes that the update files that are marked as `already-applied` / `RELEASED` in the DB have been moved to the `archive` directory.

it will then mark the updates as `ARCHIVED`:
```
Checking update "2023_04_26_00.sql"...
>> Updating the state of "2023_04_26_00.sql" to 'ARCHIVED'...
>> Update is already applied and matches the hash '8356D5D'
```

once / if the **archived** files should be deleted with the next squash, then they need to be removed from the base `updates.sql` too.

Issue #9300, opened 2021-11-23
Feature: Spell Queue System


lingxDEV:
### Describe your feature request or suggestion in detail

On retail servers you are able to queue spells, meaning you can press a spell 0.1-0.4 seconds before a finishing current cast / global and the spell will be cast after the current cast.

It was implemented it on the Preservation Project (4.3.4): https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/4430863a1e9a787a3a702b58adaf4c7bb639fa29

I guess it would have to be implemented differently in 3.3.5 since 4.3.4 supports it.

### Additional context

Spell queue system on Warmane: https://www.youtube.com/watch?v=lLIypI3P3mY

Nyeriah:
Not sure how would this be implemented without breaking anything since we don't even have that opcode https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/4430863a1e9a787a3a702b58adaf4c7bb639fa29#diff-d4f5deb80bb2d190160370c721b6c35de8f753e307193da6a2ef3cc539936098R229

Kitzunu:
"**As of patch 4.1 spell casts can now be queued up 400ms before a ongoing cast finishes** or the global cooldown expires to deliver a smooth transitions between casts though emulators do not really benefit from it.
"

Going to close, as it is not blizzlike for 3.3.5

T1ti:
This is wrong, there was a small queue system in blizzlike 3.3.5
It worked differently and was much shorter but it was there

> This used to be an in-game option, starting in Cataclysm, in the form of a slider named "Custom Lag Tolerance". It was unfortunately removed in Legion because it was deemed "too confusing".
> 
> **Spell Queuing is a feature that allows several spells to be queued up on the server - so that they can be executed immediately with no delay, regardless of how high your latency is. This is a great feature introduced in patch 2.3.0 (TBC)** and put an end to all casters needing to use /stopcasting macros and cast-bar latency indicators to perform properly. An unfortunate side-effect is that if you are a character with abilities that are cast very fast back-to-back, sometimes your character will perform attacks that you didn't want to.
> 
> As an example, Feral Druids with high haste, Heroism, and so on can sometimes cast 2x Shred instead of 1x Shred, even though you changed your mind and started mashing Ferocious Bite instead. This can be pretty annoying and can mess with your rotation. This is true for many other specs as well.
> 
[...]

PR #20797, opened 2024-12-01
feat(Core/Player): implement Spell Queue

sogladev:
## Changes Proposed:

This PR proposes changes to:

Implement a system to queue and process spell/item cast requests.
- Queue can store multiple casts prioritized by time requested (FIFO), allowing interactions such as using an item + casting a spell immediately after a cast finishes
- Window is set at [400ms, based on this discussion](https://github.com/azerothcore/azerothcore-wotlk/issues/9300#issuecomment-977211452)

Uses [similar flow logic as TC master](https://github.com/TrinityCore/TrinityCore/pull/29409) and [Cata preservation](https://github.com/The-Cataclysm-Preservation-Project/TrinityCore/commit/4430863a1e9a787a3a702b58adaf4c7bb639fa29#diff-b4a23722c5b449d326b1c0e5a75f3b51de931e42e688793fd9f1d233536a02c6R17)

One key difference is that the 335a client does not send `CMSG_CAST_SPELL`  packets while on GCD, making queueing spells during casts that last the duration of GCD not possible, such as instant cast corruption. Only spells queued during the end of a current spell cast (<400ms) will be queued. 

For item casts, `CMSG_USE_ITEM` are sent when on GCD, so items can be queued during GCD e.g., casting a 1.5s spell and queuing a potion is possible

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9300

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

tested rotations of some specs:
- affli warlock
- moonkin
- surv hunter
- warrior

1.
2.
3.

- [ ]
- [ ]

You can help by testing PRs and writing your feedback here on the PRs page on GitHub. Follow the instructions here:


walkline:
As I understood it works in the next way:
IMG_9242
Every player has its own independent queue with its own time frame for processing.

I assume it should be global for all players, like this:
IMG_9243

My assumption is incorrect?

Tereneckla:
No, you are mistaken. The first graphic is correct

Tereneckla:
No, it's just a way to circumvent input lag when casting spells. 400ms window to enqueue the spell so that it starts casting immediately after your last

sudlud:
Can you merge master please

src/server/game/Handlers/SpellHandler.cpp
```diff
@@ -85,6 +85,41 @@ void WorldSession::HandleUseItemOpcode(WorldPacket& recvPacket)
...
+
+    if (!spellInfo)
+    {
+        LOG_ERROR("network.opcode", "WORLD: unknown spell id {}", spellId);
+        recvPacket.rfinish(); // prevent spam at ignore packet
+        return;
+    }
+
+    // fail if we are cancelling pending request
+    if (_player->SpellQueue.size())
```
Winfidonarleyan, line 98: Need check via empty

Grimdhex:
Great job on the feature. I've played with this PR for 15min and I didn't see any particular issue. Work as intended with spells and items. 

Just why disable it by default? To me, as it's a blizz feature, it should be enabled (and let the players disable it if they don't want it).

heyitsbench:
I do appreciate the feature being put behind a config option, but I do agree with defaulting it to enabled, as it is a Blizzlike feature.

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -2163,6 +2163,21 @@ NoResetTalentsCost = 0
 
 ToggleXP.Cost = 100000
 
+#
+#    SpellQueue.Enabled
+#        Description: Enable SpellQueue.
+#        Default:     0 - (Disabled)
+#                     1 - (Enabled)
+
+SpellQueue.Enabled = 0
```
Nyeriah, line 2172:
```suggestion
SpellQueue.Enabled = 1
```

src/server/game/World/World.cpp
```diff
@@ -1487,6 +1487,10 @@ void World::LoadConfigSettings(bool reload)
     // Realm Availability
     _bool_configs[CONFIG_REALM_LOGIN_ENABLED] = sConfigMgr->GetOption<bool>("World.RealmAvailability", true);
 
+    // SpellQueue
+    _bool_configs[CONFIG_SPELL_QUEUE_ENABLED] = sConfigMgr->GetOption<bool>("SpellQueue.Enabled", false);
```
Nyeriah, line 1491:
```suggestion
    _bool_configs[CONFIG_SPELL_QUEUE_ENABLED] = sConfigMgr->GetOption<bool>("SpellQueue.Enabled", true);
```

Nyeriah (approved):

Grimdhex (approved):

boom8866:
It has a problem that when the player casts a spell too quickly, the second spell is not cast in the normal sequence, and keeping cycling through the first spell or other spells will result in the spell not being released, or even if the player presses the spell faster, it will result in the loss of the spell

kissingers:
> It has a problem that when the player casts a spell too quickly, the second spell is not cast in the normal sequence, and keeping cycling through the first spell or other spells will result in the spell not being released, or even if the player presses the spell faster, it will result in the loss of the spell

Yes, this is the big problem, many players say they can not cast spell right as they think, some spell always can not cast if use default as 1

Nyeriah:
It defaults to ENABLED because this is not a "custom" feature but actually part of the core. Users can either disable it if they want to or use a lower ms delay.

kissingers:
> It defaults to ENABLED because this is not a "custom" feature but actually part of the core 335 features. Users can either disable it if they want to or use a lower ms delay.


I know that enabling this option is quite good and can increase the player's DPS, but there's a problem: if you keep pressing multiple skills, you can't cast the desired skill when needed. For example, with a warlock's abilities, if you keep pressing Incinerate, you won't be able to cast Chaos Bolt when you want to. so if this problem can fix, it will more good.

Grimdhex:
> > It defaults to ENABLED because this is not a "custom" feature but actually part of the core 335 features. Users can either disable it if they want to or use a lower ms delay.
> 
> 
> 
> 
> 
> I know that enabling this option is quite good and it can increase the player's DPS. However, there's a problem: if you keep pressing multiple skills, you can't cast the desired skill when needed. For example, with a warlock's abilities, if you keep pressing Incinerate, when the skill Chaos Bolt becomes available, pressing it won't cast it; instead, it will continue to cast Incinerate.
> 
> 
> 
> It would be perfect if the system could be improved so that if no new skill is pressed, it continues to queue other skills. But if a new skill is pressed and the server receives it in time, it would be inserted at the front of the queue for casting.
> 
> 
> 
> just same as  [boom8866] description before.

Probably a bug so please open an issue on it.

Tereneckla:
You can't queue up spells for multiple casts so I don't see how spamming a spell would hinder you from casting another spell.
Unless your players are spamming Incinerate in the 400ms window only to then realize Chaos Bolt goes of CD. I'd quite frankly call that a skill issue.

PR #20815, opened 2024-12-02
fix(Tools/MapExtractor) Update MPQ patch iterator to also load patch files 4 through 9

NathanHandley:
The client can support up to 10 patch files (such as patch-enUS.MPQ through patch-enUS-9.MPQ).  This change is needed if MapExtractor is to recognize all 10.

## Changes Proposed:
-  [X] Extractor (MapExtractor) 

## Issues Addressed:
Fixes an issue where the mapextractor will not read/extract from any patch files beyond patch-4.  This change will allow it to extract from all (which goes up to *-9.mpq). 

## SOURCE:
I ran into issues locally when extracting map data inside "patch-enUS-5.MPQ" (it stopped at patch-4, with the needed map data being in patch-5).

## Tests Performed:
This has been tested by the change author.  Maps were successfully exported with the mapextractor.exe following this change.

## How to Test the Changes:
Create (or rename existing to) a /Data/[localization]/patch-[localization]-5.MPQ file, then run mapextractor

No known issues.  It's an extremely simple change.

### DIFF
--- a/src/tools/map_extractor/System.cpp
+++ b/src/tools/map_extractor/System.cpp
@@ -1150,11 +1150,9 @@ void LoadLocaleMPQFiles(int const locale)
         if (i > 1)
             sprintf(ext, "-%i", i);
 
+        sprintf(filename, "%s/Data/%s/patch-%s%s.MPQ", input_path, langs[locale], langs[locale], ext);
         if (FileExists(filename))
-        {
-            sprintf(filename, "%s/Data/%s/patch-%s%s.MPQ", input_path, langs[locale], langs[locale], ext);
             new MPQArchive(filename);
-        }
     }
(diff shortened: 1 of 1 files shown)


heyitsbench:
This doesn't add support for any patch files with an alphabetical character, correct? If so, that may be something to add support for down the line.

NathanHandley:
> This doesn't add support for any patch files with an alphabetical character, correct? If so, that may be something to add support for down the line.

Correct, it doesn't.  It does increase it from -4 to -9, but yeah something more complicated would be needed for -A through -Z.

sudlud:
Looks like the tools build is not with us today

sudlud (approved):

Issue #5068, opened 2021-03-31
Feature: Auction House getAll scan


eTzmNcbkrng:
**Describe your feature request or suggestion in detail**
Add the Auction House feature to run a getAll scan which was added to WoW in 2.3

**Describe a possible solution to your Feature or Suggestion in detail**
https://github.com/TrinityCore/TrinityCore/commit/3aaeb574050668e5a240078f6e40337c3975d110

**Additional context**

**getAll**
- Download the ENTIRE auction house as one single page. Only allowed when CanSendAuctionQuery() returns true as its second return (~every 15 minutes). May disconnect people with low bandwidth. Also see notes on client-to-server traffic in GetAuctionItemInfo() (Added in 2.3, though initially disabled. Appears to be enabled as of 2.4)

 \- https://wowpedia.fandom.com/wiki/API_QueryAuctionItems

necropola:
I'd really like to have **AuctionHouse Search with GetAll** working on AC and the commit from TC would apply with very little modifications, if there weren't these weird **pussywizard** modifications:
https://github.com/azerothcore/azerothcore-wotlk/blob/5906222fd6910beec4b5d6c07f4c69f69bb094a4/src/server/game/AuctionHouse/AuctionHouseMgr.cpp#L550-L585
* A special *optimization for a simplified case* (Which case and why handle it differently?)
* A special/additional thread AsyncAuctionListing (Why can't BuildListAuctionItems just be directly called from AuctionHouseHandler like all the other stuff that writes to world packets?)

It looks like the initial import into github already had these special modifications (which are spread over various parts of the code including **worldserver/Master**), but so far nobody touched it, not even to fix obvious typos.  
https://github.com/azerothcore/azerothcore-wotlk/blob/5906222fd6910beec4b5d6c07f4c69f69bb094a4/src/server/worldserver/Master.cpp#L192-L194

Does any of the veteran devs know/remember, why this sepcial code is needed in AC?

tstuver2000:
Just wanted to chime in and add my support for this feature. Here's what I wrote on #Suggestions on discord:

AH API Suggestion: Have AzerothCore support the "GetAll" one-click AH scan. This is a type of all-in-one auction house scan / API dump (sorry if I'm not using the best terms for this) that blizzard implemented to allow faster scanning of the AH and lesser load on the servers.

[...]

PR #20830, opened 2024-12-03
feat(Core/AuctionHouse): Rework auctionhouse search threading

### LINKED ISSUE #20538: Feature: Rewrite Auction House search

Currently, the auction search implementation has race conditions that lead to crashes (like [this one](https://gist.github.com/Nyeriah/d331e335c0c99427755712cc482ff295)).
The race conditions happens in this thread - https://github.com/azerothcore/azerothcore-wotlk/blob/f8d53d4cd87a887237986e4a2619b54c701f1dc0/src/server/apps/worldserver/Main.cpp#L713
In this thread, we interact with auction listings and players/sessions without proper synchronization.

I see several solutions here:

1. Move search requests back to the world thread. This isn't the best option, as searches would affect the world update tick (diff).
2. Use an appropriate threading model for search requests. I outlined my ideas in this scheme, but it's not the only way to achieve this.  
AH drawio
I believe the suggested solution is scalable and would require minimal locking. Since each worker maintains its own cache, it has a space (RAM) complexity of NxM, where N is the number of workers and M is the number of listings in the auction.
3. Delegate Auction House functionality to a new service (process). I have long-term plans to implement this as part of [TC9](https://github.com/walkline/ToCloud9) project, but I think we need a solution now.

### Additional context

_No response_

Takenbacon:
Complete rework of the auctionhouse search threading as the old system has been proven to be obviously very unsafe (#20538). 

Simple model, spawns a configurable number of worker threads that each contain a pointer to a shared copy of AuctionEntry with an almost completely decoupled data structure (except for a pointer to ItemTemplate, which will be fine as long as we don't actively invalidate ItemTemplate data during run time. This can be "fixed" if needed by copying relevant item template data at the cost of increased memory usage and inability to modify existing data).

Scales well and is safe. I have used a similar model on servers with xx,xxx+ players and 150k+ auctions, however the implementation is quite different. I have done some initial local testing but I would highly recommend more testing before any merging. No danger to auctionhouse item data integrity during as we do not modify any auction entry data. Compile tested using MSVC only currently, GCC/Clang untested.

The auctionhouse system itself could use more changes but I kept the scope of this PR limited to just the searches.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes #20538 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Simply use the auctionhouse (specifically anything related to searches, owned and bidded auction lists)
2.
3.

None currently.

- [ ]
- [ ]


heyitsbench:
Looks like this also closes https://github.com/azerothcore/azerothcore-wotlk/issues/5068.

src/server/game/AuctionHouse/AuctionHouseMgr.h
```diff
@@ -28,9 +28,13 @@
 
 class Item;
 class Player;
+class AuctionHouseSearcher;
 
 #define MIN_AUCTION_TIME (12*HOUR)
 #define MAX_AUCTION_ITEMS 160
+#define MAX_AUCTIONS_PER_PAGE 50
+#define AUCTION_SEARCH_DELAY 300
+#define MAX_GETALL_RETURN 55000
```
heyitsbench, line 37: Is there any particular reasoning for this value in particular?

Takenbacon: Yes, the max allowable single packet size in 3.3.5 client protocol is 0x7FFFFF. A single BuildAuctionInfo structure has a size of 148 bytes. 148 * 55000 = 8140000 which gives us just under the max size of 8388607 with a little bit of margin. The 55k is just a nice arbitrary number close to the limit. There has been previous research to confirm the "getAll" feature did not fully work correctly on high population live realms during WoTLK with lots of auctions but I don't have an exact source at the moment as this was long ago.

heyitsbench: Gotcha, thank you for the explanation. ❤️

Takenbacon:
Looks like I finally got most of the build check issues resolved, only problem remaining appears to be a couple small compatibility issues with ahbot, in which it'd probably be best just to update the ahbot module.

src/server/game/AuctionHouse/AuctionHouseSearcher.cpp
```diff
@@ -0,0 +1,722 @@
...
+    _responseQueue->Enqueue(searchResponse);
+}
+
+void AuctionHouseWorkerThread::SearchBidderListRequest(AuctionSearchBidderListRequest const& searchBidderListRequest)
+{
+    SearchableAuctionEntriesMap const& searchableAuctionMap = GetSearchableAuctionMap(searchBidderListRequest.listFaction);
+
+    AuctionSearcherResponse* searchResponse = new AuctionSearcherResponse();
+    searchResponse->playerGuid = searchBidderListRequest.ownerGuid;
+    searchResponse->packet.Initialize(SMSG_AUCTION_BIDDER_LIST_RESULT, (4 + 4 + 4) + 30000); // pussywizard: ensure there is enough memory
```
walkline, line 241: I know that this change (upfront allocation of 30 012 bytes) doesn't come from your side, but does it make sense to you? I thought the average bidder list shouldn't be that long.
On the other hand, I think upfront allocation can be useful for getAll requests.

src/server/game/AuctionHouse/AuctionHouseSearcher.cpp
```diff
@@ -0,0 +1,722 @@
...
+{
+    // pussywizard: optimization, this is a simplified case
+    if (searchRequest.searchInfo.itemClass == 0xffffffff && searchRequest.searchInfo.itemSubClass == 0xffffffff
+        && searchRequest.searchInfo.inventoryType == 0xffffffff && searchRequest.searchInfo.quality == 0xffffffff
+        && searchRequest.searchInfo.levelmin == 0x00 && searchRequest.searchInfo.levelmax == 0x00
+        && searchRequest.searchInfo.usable == 0x00 && searchRequest.searchInfo.wsearchedname.empty())
+    {
+        for (auto const& pair : auctionMap)
+            auctionEntries.push_back(pair.second.get());
+    }
```
walkline, line 287: Small nitpick: if we add `return` here, we can get rid of the `else` statement below.

sudlud:
Thanks for the update, could you please also prepare a PR for ahbot? Then we could get both merged when everything's ready :)

src/server/game/Handlers/AuctionHouseHandler.cpp
```diff
@@ -267,25 +267,27 @@ void WorldSession::HandleAuctionSellItem(WorldPacket& recvData)
...
         if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_AUCTION))
-            AH->houseId = AUCTIONHOUSE_NEUTRAL;
+            AH->houseId = AuctionHouseId::Neutral;
         else
         {
             CreatureData const* auctioneerData = sObjectMgr->GetCreatureData(creature->GetSpawnId());
             if (!auctioneerData)
             {
                 LOG_ERROR("network.opcode", "Data for auctioneer not found ({})", auctioneer.ToString());
+                delete AH;
```
walkline, line 277: Good catch!

src/server/game/Handlers/AuctionHouseHandler.cpp
```diff
@@ -736,24 +707,63 @@ void WorldSession::HandleAuctionListItems(WorldPacket& recvData)
...
     if (_player->HasUnitState(UNIT_STATE_DIED))
-    {
         _player->RemoveAurasByType(SPELL_AURA_FEIGN_DEATH);
-    }
 
-    // pussywizard:
-    const Milliseconds delay = 2s;
-    const Milliseconds now = GameTime::GetGameTimeMS();
-    Milliseconds diff = GetMSTimeDiff(_lastAuctionListItemsMSTime, now);
-    if (diff > delay)
```
walkline, line 749: I was thinking maybe we could add rate limiter for some auction house opcodes somewhere [here](https://github.com/azerothcore/azerothcore-wotlk/blob/47580c0fab3d704ca5b2e4618984712e52e38950/src/server/game/Server/WorldSession.cpp#L1370). What do you think? Maybe we can address it in a separate PR if needed.

walkline (commented):
I think you already know that I love this implementation 😁 I just have a few small nitpicks (some of them related to the old code).

src/server/game/World/World.cpp
```diff
@@ -2339,18 +2337,9 @@ void World::Update(uint32 diff)
...
-        _timers[WUPDATE_AUCTIONS].Reset();
-
-        // pussywizard: handle expired auctions, auctions expired when realm was offline are also handled here (not during loading when many required things aren't loaded yet)
-        sAuctionMgr->Update();
-    }
-
-    AsyncAuctionListingMgr::Update(Milliseconds(diff));
+    // pussywizard: handle expired auctions, auctions expired when realm was offline are also handled here (not during loading when many required things aren't loaded yet)
+    METRIC_TIMER("world_update_time", METRIC_TAG("type", "Update expired auctions"));
+    sAuctionMgr->Update(diff);
```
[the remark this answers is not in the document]
Takenbacon: Done. https://github.com/azerothcore/azerothcore-wotlk/pull/20830/commits/6cb4d9ce5dee56f7ecc577efec9d518f9cd23721

src/server/game/AuctionHouse/AuctionHouseSearcher.cpp
Takenbacon: I did see that but opted to not change it at the time, but yes, it's a bit excessive and ultimately useless. At 148 bytes per auction entry it's effectively preallocating the buffer size for ~202 bidder auction entries which is unlikely to be the case 99% of the time. It's a bit insane to pre-optimize for ultra edge cases especially when the optimization is overall extremely insignificant. Also, we don't do the same for either SMSG_AUCTION_OWNER_LIST_RESULT or SMSG_AUCTION_LIST_RESULT. As such, I have now removed it https://github.com/azerothcore/azerothcore-wotlk/pull/20830/commits/ed83ecd3422511645e2cfdeb93d7cf63d2cab512

src/server/game/AuctionHouse/AuctionHouseSearcher.cpp
Takenbacon: Done. https://github.com/azerothcore/azerothcore-wotlk/pull/20830/commits/f2e96081fbd46657f48a54a24d81512146600339

src/server/game/Handlers/AuctionHouseHandler.cpp
Takenbacon: The auction list opcodes definitely should be throttled, in fact there's even a blizzlike client side delay that we add to the list packets (now defined as AUCTION_SEARCH_DELAY) but that's obviously only on the client side. Personally, I've seen the antidos system and I can't say I'm a huge fan of how it works for a variety of reasons that would take too long to elaborate on, but I suppose it's the best we have for now. It could still use some refinement if we had a better system but as a temporary measure I've added them https://github.com/azerothcore/azerothcore-wotlk/pull/20830/commits/2a51d5629f534e3c0d4e8145dff9853bf0f1d2ab

CMSG_AUCTION_LIST_ITEMS is still lacking the blizzlike throttle for "getAll" requests (something around ~15 minutes if I recall correctly?). I don't believe it to be too important at the moment but it should also be implemented at some point. (Note: If someone other than me eventually implements it, there's a "gotcha" that CMSG_AUCTION_LIST_ITEMS is currently handled as PROCESS_THREADSAFE, any global containers used in the handler must be locked or the opcode moved to PROCESS_THREADUNSAFE)

walkline: Yeah, agree. Thanks for addressing the comments!

src/server/game/AuctionHouse/AuctionHouseMgr.h
sudlud: Probably worth adding this as a comment to the define value? :)

Takenbacon:
> @Takenbacon just so I know if i need to update the labels. Is this ready to be tested or still being changed/developed?

Ready to be tested, nothing left on my end except for any requested changes during reviews/testing that may come up.

walkline:
I ran a small smoke test - no issues were found.

Nyeriah (approved):

Grimdhex:
Why use std::wstring instead string ? Most of the core manipulate char and not wchar_t. Even with OS differences, that shouldn't be an issue ? Does I'm wrong ?

Takenbacon:
> Why use std::wstring instead string ? Most of the core manipulate char and not wchar_t. Even with OS differences, that shouldn't be an issue ? Does I'm wrong ?

Mostly because that is what was there in the original implementation. Use of a wide string is not new from this PR. 
https://github.com/azerothcore/azerothcore-wotlk/blob/69c85ca8a2387498d87730b4a4791aa3c56200ba/src/server/game/Misc/AsyncAuctionListing.cpp#L56

Now, why we use a wide string in general? It's going to be for supporting the various languages/character encoding inputs received from the client. But, I couldn't really give you the specifics and I can't say I've bothered to really look into it. Character encoding is not a particular strong suit of mine as it is a confusing mess and varies significantly. You'll notice `Utf8toWStr` has quite a few uses in the codebase, mostly converting client string inputs (who list, channels, chat, etc)

Grimdhex (approved):

Nyeriah:
Please check the merge conflicts
Also we need someone to please take a look into the AH bot module so we can merge it

Takenbacon:
> Please check the merge conflicts Also we need someone to please take a look into the AH bot module so we can merge it (the CI is failing)

Conflicts resolved.

src/server/game/AuctionHouse/AuctionHouseMgr.h
sudlud: Did you add the comment already?

Takenbacon: > Did you add the comment already? I would appreciate it

Oops, I read this comment while I was originally at work and forgot about it. Will add now.

sudlud (approved):

Kitzunu:
Awesome sauce

Nyeriah:
https://github.com/azerothcore/mod-eluna needs an update

PR #20883, opened 2024-12-09
fix(Scripts/Zuldrak) Spell: Scourge Disguise and related stuff

p-tkachuk:
## Changes Proposed:
This PR proposes changes to:

## Issues Addressed:
1. Scourge Disguise spell [51966, 52192] is "unstable" and should be removed after some time with message informing player about it: "Scourge Disguise Failing! Find a safe place!"
2. Quest "Dark horizon" [12664]: When NPC Gorebag [28666] sends player on a fly mount to see some places it casted another copy of "Scourge Disguise" on player, which caused 2 active auras being on a player at the same time. And if 1 aura was removed - second aura didnt had any effect.

## SOURCE:

## Tests Performed:
This PR has been:

## How to Test the Changes:

TELEPORT
.go xyz 6278.02 -1933.10 239.60

HOW TO TEST ITEM/SPELL:
// Choker
.additem 38699

Use choker, wait 3-5 minutes. Now you should see that spell is going to end, and it will be actually removed.

HOW TO TEST QUEST:
// Choker
.additem 38699
// dark horizon quest
.quest add 12664
// Gorebag
.go creature id 28666 (beware of agro, use choker first)

Talk to NPC, take a ride, see that you have only 1 active aura of "Scourge Disguise" after ride is over.

After the ride event from quest dark horizon[12664] player end up with "Scourge Disguise"[52192] aura, which is different one that player gets when he use Choker ("Scourge Disguise"[51966]). So if player uses choker while having 52192 aura - it will be removed after 8 seconds.

### DIFF
--- a/src/server/scripts/Northrend/zone_zuldrak.cpp
+++ b/src/server/scripts/Northrend/zone_zuldrak.cpp
@@ -976,7 +976,7 @@ class spell_scourge_disguise_instability : public AuraScript
             if (player->HasAura(SPELL_SCOURGE_DISGUISE) || player->HasAura(SPELL_SCOURGE_DISGUISE_INSTANT_CAST))
             {
                 uint32 textId = Acore::Containers::SelectRandomContainerElement(scourgeDisguiseTextIDs);
-                player->Talk(textId, CHAT_MSG_RAID_BOSS_EMOTE, 0, player);
+                player->Unit::Whisper(textId, player, true);
 
                 player->CastSpell(player, SPELL_SCOURGE_DISGUISE_EXPIRING, true);
             }
--- a/src/server/scripts/Northrend/zone_zuldrak.cpp
+++ b/src/server/scripts/Northrend/zone_zuldrak.cpp
@@ -23,6 +23,7 @@
 #include "ScriptedGossip.h"
 #include "SpellAuras.h"
 #include "SpellInfo.h"
+#include "SpellScriptLoader.h"
 #include "Vehicle.h"
 
 // Ours
@@ -322,6 +323,7 @@ enum overlordDrakuru
     SPELL_THROW_BRIGHT_CRYSTAL          = 54087,
     SPELL_TELEPORT_EFFECT               = 52096,
     SPELL_SCOURGE_DISGUISE              = 51966,
+    SPELL_SCOURGE_DISGUISE_INSTANT_CAST = 52192,
     SPELL_BLIGHT_FOG                    = 54104,
     SPELL_THROW_PORTAL_CRYSTAL          = 54209,
     SPELL_ARTHAS_PORTAL                 = 51807,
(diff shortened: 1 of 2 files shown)


src/server/scripts/Northrend/zone_zuldrak.cpp
```diff
@@ -948,6 +950,39 @@ class npc_storm_cloud : public CreatureScript
...
+    void HandleRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        Unit* caster = GetCaster();
+        if (caster && caster->IsPlayer())
+        {
+            Player* player = caster->ToPlayer();
+            if (player->HasAura(SPELL_SCOURGE_DISGUISE) || player->HasAura(SPELL_SCOURGE_DISGUISE_INSTANT_CAST))
+            {
+                WorldPacket data;
+                ChatHandler::BuildChatPacket(data, CHAT_MSG_RAID_BOSS_EMOTE, LANG_UNIVERSAL, player, player, "Scourge Disguise Failing! Find a safeplace!");
```
Kitzunu, line 971: Thanks for the PR but we don't hardcode messages like this. They are picked up from the DB

src/server/scripts/Northrend/zone_zuldrak.cpp
```diff
@@ -948,6 +950,39 @@ class npc_storm_cloud : public CreatureScript
     }
 };
 
+class spell_scourge_disguise_instability : public AuraScript
+{
+    PrepareAuraScript(spell_scourge_disguise_instability)
+
+    void HandleApply(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        SetDuration(urand(3*60*1000, 5*60*1000));
```
Kitzunu, line 959: Use already defined values, so 3 * MINUTE * IN_MILLISECONDS

src/server/scripts/Northrend/zone_zuldrak.cpp
p-tkachuk: I'm kinda new here... would you care to share a specific table to look?

Kitzunu: Well actually we ideally dont use chathandler here. 

What you want to do is
```cpp
player->Say(TEXT_ID);
```
text_id is from broadcast_text table

Kitzunu: and remember to not hardcode the textid, put it in an enum

Kitzunu:
```sql
SELECT * FROM broadcast_text WHERE maletext LIKE '%Scourge Disguise%'

Comment (p-tkachuk) [reply src/server/scripts/Northrend/zone_zuldrak.cpp:971]: Wow, so it is actually not 1, but 3 different texts. Probably should be used randomly.

File: src/server/scripts/Northrend/zone_zuldrak.cpp (hunk 2 above)
Comment (p-tkachuk) [reply src/server/scripts/Northrend/zone_zuldrak.cpp]: Done

File: src/server/scripts/Northrend/zone_zuldrak.cpp (hunk 1 above)
Comment (p-tkachuk) [reply src/server/scripts/Northrend/zone_zuldrak.cpp]: @Kitzunu I've replaced hardcoded string with TEXT_ID from Enum.
But I would like to use player->Talk instead of player->Say, because I need to set specific message type. This message should not be a regular chat message from player, but a big yellow text in the center of the screen:

image
Video from 2008: https://www.wowhead.com/wotlk/quest=12669/so-far-so-bad#videos:id=944

Comment (Kitzunu) [reply src/server/scripts/Northrend/zone_zuldrak.cpp]: Talk isn't a member of Player class

Comment (Kitzunu) [reply src/server/scripts/Northrend/zone_zuldrak.cpp]: So the correct way of fixing it is to know how it should be sent which can be obtained in a sniff.
That way we can figure who is the sender, the server or the player.

If it is the player, then we currently don't have a supported way.
But assuming it isn't you can do
```cpp
ChatHandler(player->GetSession()).PSendSysMessage(TEXT_ID);
```

Kitzunu: Remember to include ChatHandler.h

Kitzunu: and even if Talk was a part of Player class it doesn't do notifications

Kitzunu: So maybe acore_string and PSendSysMessage is the way to go...

@Nyeriah Assist! I am getting brain farts

p-tkachuk: > Talk isn't a member of Player class

But it is a member of Unit class, which is a parent class for Player.
Is it "bad manners" to use parent class methods?

p-tkachuk: > and even if Talk was a part of Player class it doesn't do notifications

I have ~0 knowledge of code base at this point, so I do not know what you mean by telling "it doesn't do notifications" :)

BUT! With my current implementation I have result pretty close to what Blizzard have in their serevers
image

Kitzunu: my bad then

src/server/scripts/Northrend/zone_zuldrak.cpp
```diff
@@ -948,6 +950,46 @@ class npc_storm_cloud : public CreatureScript
     }
 };
 
+enum ScourgeDisguiseInstability
+{
+    SCOURGE_DISGUISE_FAILING_MESSAGE_1       = 28552, // Scourge Disguise Failing! Find a safe place!
+    SCOURGE_DISGUISE_FAILING_MESSAGE_2       = 28758, // Scourge Disguise Failing! Run for cover!
+    SCOURGE_DISGUISE_FAILING_MESSAGE_3       = 28759, // Scourge Disguise Failing! Hide quickly!
+};
+const std::vector<uint32> scourgeDisguiseTextIDs = { SCOURGE_DISGUISE_FAILING_MESSAGE_1, SCOURGE_DISGUISE_FAILING_MESSAGE_2, SCOURGE_DISGUISE_FAILING_MESSAGE_3 };
```
Kitzunu, line 959: std::vector .... const scourgeDisguiseTextIDs

p-tkachuk: replaced

PR #20955, opened 2024-12-17
fix(Core/Grids): Grid improvements

Takenbacon:
**!! PR is WIP and not ready to be merged**

1. Dynamically allocates grid cells. At (64 * 64) * (8 * 8) = **262,144 cells per map** improving the memory efficiency of the structure has a significant impact considering most cells are usually empty anyways. When compiled in RelWithDebug and using the `PreloadAllNonInstancedMapGrids = 1` config option, RAM usage decreased from 9,810MB to 4,645MB (no mmaps) and startup time improved from 19 seconds to 16 seconds.
2. Greatly simplify ObjectGridLoader

While the changes as of this moment are testable, there is more I'd still like to do, thus PR is WIP for now.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
4.
5.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -280,9 +280,9 @@ void Map::AddToGrid(T* obj, Cell const& cell)
 {
     NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
     if (obj->IsWorldObject())
-        grid->GetGridType(cell.CellX(), cell.CellY()).template AddWorldObject<T>(obj);
+        grid->AddWorldObject<T>(cell.CellX(), cell.CellY(), obj);
     else
-        grid->GetGridType(cell.CellX(), cell.CellY()).template AddGridObject<T>(obj);
+        grid->AddGridObject<T>(cell.CellX(), cell.CellY(), obj);
 }
 
 template<>
@@ -290,9 +290,9 @@ void Map::AddToGrid(Creature* obj, Cell const& cell)
 {
     NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
     if (obj->IsWorldObject())
-        grid->GetGridType(cell.CellX(), cell.CellY()).AddWorldObject(obj);
+        grid->AddWorldObject(cell.CellX(), cell.CellY(), obj);
     else
-        grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
+        grid->AddGridObject(cell.CellX(), cell.CellY(), obj);
 
     obj->SetCurrentCell(cell);
 }
@@ -301,7 +301,7 @@ template<>
 void Map::AddToGrid(GameObject* obj, Cell const& cell)
 {
     NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
-    grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
+    grid->AddGridObject(cell.CellX(), cell.CellY(), obj);
 
     obj->SetCurrentCell(cell);
 }
@@ -311,9 +311,9 @@ void Map::AddToGrid(DynamicObject* obj, Cell const& cell)
 {
     NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
     if (obj->IsWorldObject())
-        grid->GetGridType(cell.CellX(), cell.CellY()).AddWorldObject(obj);
+        grid->AddWorldObject(cell.CellX(), cell.CellY(), obj);
     else
-        grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
+        grid->AddGridObject(cell.CellX(), cell.CellY(), obj);
 
     obj->SetCurrentCell(cell);
 }
@@ -331,9 +331,9 @@ void Map::AddToGrid(Corpse* obj, Cell const& cell)
     if (grid->isGridObjectDataLoaded())
     {
         if (obj->IsWorldObject())
-            grid->GetGridType(cell.CellX(), cell.CellY()).AddWorldObject(obj);
+            grid->AddWorldObject(cell.CellX(), cell.CellY(), obj);
         else
-            grid->GetGridType(cell.CellX(), cell.CellY()).AddGridObject(obj);
+            grid->AddGridObject(cell.CellX(), cell.CellY(), obj);
     }
 }
 
@@ -362,18 +362,16 @@ void Map::SwitchGridContainers(Creature* obj, bool on)
     NGridType* ngrid = getNGrid(cell.GridX(), cell.GridY());
     ASSERT(ngrid);
 
-    GridType& grid = ngrid->GetGridType(cell.CellX(), cell.CellY());
-
     obj->RemoveFromGrid(); //This step is not really necessary but we want to do ASSERT in remove/add
 
     if (on)
     {
-        grid.AddWorldObject(obj);
+        ngrid->AddWorldObject(cell.CellX(), cell.CellY(), obj);
         AddWorldObject(obj);
     }
     else
     {
-        grid.AddGridObject(obj);
+        ngrid->AddGridObject(cell.CellX(), cell.CellY(), obj);
         RemoveWorldObject(obj);
     }
 
@@ -400,18 +398,16 @@ void Map::SwitchGridContainers(GameObject* obj, bool on)
     NGridType* ngrid = getNGrid(cell.GridX(), cell.GridY());
     ASSERT(ngrid);
 
-    GridType& grid = ngrid->GetGridType(cell.CellX(), cell.CellY());
-
     obj->RemoveFromGrid(); //This step is not really necessary but we want to do ASSERT in remove/add
 
     if (on)
     {
-        grid.AddWorldObject(obj);
+        ngrid->AddWorldObject(cell.CellX(), cell.CellY(), obj);
         AddWorldObject(obj);
     }
     else
     {
-        grid.AddGridObject(obj);
+        ngrid->AddGridObject(cell.CellX(), cell.CellY(), obj);
         RemoveWorldObject(obj);
     }
 }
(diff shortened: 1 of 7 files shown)


Grimdhex:
One small change if you want for more readability: go out the GridMap inside the Map file. And put it in this own file.

Takenbacon:
> One small change if you want for more readability: go out the GridMap inside the Map file. And put it in this own file.

Yeah I saw that, I'm currently weighing how much I should change in a single PR, and also considering this is specifically marked "grids" whereas some of my changes would be more related to map class.

Kitzunu:
Interesting 👍

src/server/game/Maps/Map.cpp
```diff
@@ -447,7 +443,7 @@ void Map::EnsureGridCreated_i(const GridCoord& p)
     if (!getNGrid(p.x_coord, p.y_coord))
     {
         // pussywizard: moved setNGrid to the end of the function
-        NGridType* ngt = new NGridType(p.x_coord * MAX_NUMBER_OF_GRIDS + p.y_coord, p.x_coord, p.y_coord);
+        NGridType* ngt = new NGridType(p.y_coord * MAX_NUMBER_OF_GRIDS + p.x_coord, p.x_coord, p.y_coord);
```
Grimdhex, line 446: Why reverse the coordinates here?

Takenbacon: To synchronize NGrid ID with CoordPair's ID which enables the ObjectGridLoader simplifications. I believe this to have just been a simple mistake in the original grid system implementation as GridId isn't used much let alone in conjunction with a grid CoordPair. Previously, spawned objects were stored by cell ID and the object grid loader would loop all cells in the grid and do a lookup on the _mapObjectGuidsStore map for each cell in the grid. Since we're always loading every cell in a grid anyways, it's easy to cut out the bloat and just store the objects by grid ID instead, simplifying the process.

CoordPair:
`[[nodiscard]] uint32 GetId() const
{ return y_coord * LIMIT + x_coord; }`

Grimdhex:
I see that you have moved some grid definitions outside of maps. If you don't these data are duplicated at differents places and can be merged. I've a PR that port these changes available only on TC master branch (https://github.com/TrinityCore/TrinityCore/commit/65f82c7c274a7b7058c065dfffe224c59855f41d#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52)

Takenbacon:
> I see that you have moved some grid definitions outside of maps. If you don't these data are duplicated at different places in the framework, and it can be merged. I've a PR that port these changes available only on TC master branch ([TrinityCore/TrinityCore@65f82c7#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52](https://github.com/TrinityCore/TrinityCore/commit/65f82c7c274a7b7058c065dfffe224c59855f41d#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52)); It's just for information, not necessary for this PR.

Sorry, I don't quite understand. Are you saying you want me to merge in that TC commit?

Grimdhex:
> > I see that you have moved some grid definitions outside of maps. If you don't these data are duplicated at different places in the framework, and it can be merged. I've a PR that port these changes available only on TC master branch ([TrinityCore/TrinityCore@65f82c7#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52](https://github.com/TrinityCore/TrinityCore/commit/65f82c7c274a7b7058c065dfffe224c59855f41d#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52)); It's just for information, not necessary for this PR.
> 
> 
> 
> Sorry, I don't quite understand. Are you saying you want me to merge in that TC commit?

Not necessary. If you want other improvements for grids, it's another possibility. But no real impact IG. Just code management.

Takenbacon:
> > > I see that you have moved some grid definitions outside of maps. If you don't these data are duplicated at different places in the framework, and it can be merged. I've a PR that port these changes available only on TC master branch ([TrinityCore/TrinityCore@65f82c7#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52](https://github.com/TrinityCore/TrinityCore/commit/65f82c7c274a7b7058c065dfffe224c59855f41d#diff-ad66842f3a1bb1684807e622c9f129d330fa7cd3daf4ba9a62254cd2de340c52)); It's just for information, not necessary for this PR.
> > 
> > 
> > Sorry, I don't quite understand. Are you saying you want me to merge in that TC commit?
> 
> Not necessary. If you want other improvements for grids, it's another possibility. But no real impact IG. Just code management.

Ah gotcha. Yeah, I still have a bunch of local WIP changes to reorganize some more of the code. But I am definitely starting to get tired of messing with it so I might just end up finishing what I'm doing and calling it "good" for now, haha.

Nyeriah:
Every small improvement is a good improvement, absolutelly no need to address all that in one go 😅 Thanks for this all other PRs btw.

Takenbacon:
My most recent commit will be the bulk of the refactoring I'll do in regards to grids I think. There's still a couple smaller things I'm planning on looking into cleaning up/improving but I could probably spend the rest of my life on this and I'd like to move on to other things eventually haha. It does compile currently (on MSVC) and I was able to login and test a couple things. I will be out of town this upcoming weekend so I'll hopefully get this wrapped up the following if all goes well.

Takenbacon:
I FINALLY got the build fixed on gcc/clang. Holy crap. I really just need to get unlazy and set up my own environment instead of depending on the automated build scripts lol. Anyways, that should be it for this PR in terms of changes. Ready to be tested and WIP tag removed when convenient.

src/server/game/Grids/GridTerrainData.h
```diff
@@ -0,0 +1,193 @@
...
+    uint32 Flags{ 0 };
+    float  Level{ INVALID_HEIGHT };
+    float  DepthLevel{ INVALID_HEIGHT };
+    LiquidStatus Status{ LIQUID_MAP_NO_WATER };
+};
+
+class GridTerrainData
+{
+    uint32  _flags;
+    union
```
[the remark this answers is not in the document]
Takenbacon: Not sure if variant works too well in this particular case with pointers to arrays. Either way, I've done some refactoring to the file and removed the union into separate structures, as it makes a little more sense as to how the data is loaded anyways.

Takenbacon:
Made a couple small changes and a fix.

Interesting tidbit: Previously (before my PR), when a grid terrain map failed to load for any reason the GridMap remained created and accessible - either in a partially loaded or default-initialized state, which I think is not great and could've led to some odd behavior. I fixed that but it did have an unintended consequence of at least in one case in `Map::GetWaterOrGroundLevel` having a different default return value causing an issue with random movement in areas without terrain map data (RFC instance for example). I've gone through all the uses of `GetGridTerrainData` and think I got all the uses checked now, I hope.

I've done about as much local testing as I can, just looking for others to now give it a test and feedback results so this PR hopefully doesn't stay in an idle state for too long of a duration.

Kitzunu:
@azerothcore/testers Do your thing

blinkysc:
Been trying this out for a few hours, seems like a very big improvement! Haven't encountered any bugs. Also no joke with the memory improvements...

before
before

after
after

Also some perf reports (I was interested if many more cycles or anything else) 
[before_perf.txt](https://github.com/user-attachments/files/18776028/before_perf.txt)
[after_perf.txt](https://github.com/user-attachments/files/18776029/after_perf.txt)

I wanted to try with many thousand bots but seems many of the cpp/h changed runs into changes to the playerbot fork. May take awhile to try and "plug and play" those difference to give it a go with that.

PkllonG:
image
Before PR, they are lit

Kitzunu:
Conflict

Takenbacon:
> image Before PR, they should be lit
> 
> Before PR image

Fixed, thank you for checking. I was loading grids a little too early in the process before instance scripts so they weren't being properly assigned to the script. Good catch.

> Conflict

Fixed

PkllonG:
I ran it for a while longer and didn't find any errors for a while

PkllonG:
So far, it's been working well

Nyeriah (approved):

Nyeriah:
I will need some help with this one. We are populating maps before they have any players inside now, so instance maps can no longer tell wether they should spawn one NPC or another, where the spawn is conditioned by faction inside the instance

https://github.com/azerothcore/azerothcore-wotlk/issues/21529

Takenbacon:
> I will need some help with this one. We are populating maps before they have any players inside now, so instance maps can no longer tell wether they should spawn one NPC or another, where the spawn is conditioned by faction inside the instance
> 
> #21529

Ah I had forgotten about faction instances, I'm guessing the hook that is setting faction Id is being called after the grids get loaded. Should be a fairly straightforward fix, either moving the hook call or maybe alternatively moving it to a different hook (OnInstanceCreate?) which might be a better solution anyways. I will take a look in the morning after work if possible.

Nyeriah:
I tried looking into it, but the issue is that the instance is always empty when they are spawned, so we would need to add players before the creatures are spawned (?). Or get that information during the creation process itself, when the map is created for the first player to enter

Takenbacon:
> I tried looking into it, but the issue is that the instance is always empty when they are spawned, so we would need to add players before the creatures are spawned (?). Or get that information during the creation process itself, when the map is created for the first player to enter (but we don't have the instance script initialized at that point?)

https://github.com/azerothcore/azerothcore-wotlk/blob/28f5ead7f308a6a301d3dd8798c5031e8a52dcb1/src/server/game/Maps/MapInstanced.cpp#L216

Grids are spawned on the hook here, right after instance script is created. Probably need to pass player to this function and add a new hook just before grids are loaded.

Takenbacon:
@Nyeriah https://github.com/Takenbacon/azerothcore-wotlk/commit/e607d1cde037214a4229577efe01e3a69ef33e2a I threw this together really quickly if you or someone want to test, it should work or at the very least lead someone in the right direction, I just unfortunately don't have enough time until after my work week is done, if it's still a problem I can look at it this upcoming weekend but I am quite confident it should take care of it.

PR #21002, opened 2024-12-21
refactor(CI/Codestyle): Move the SQL check to python

Kitzunu:
this also fixes the issue where you couldnt have comments after the last semicolon


Kitzunu:
image

Exitare:
Thanks for this adjustment :)

Grimdhex:
I've tested with edge case and it's fine. But I will probably rewrite the script later to handle more complex behavior (as if you find INSERT comeback in the file and find DELETE...). But it's a more complex logic than parsing from start to the end. 

Thanks :)

Grimdhex (approved):

Issue #21007, opened 2024-12-22
[Quest] Counterattack! (4021) - Banner (Object 164690) not spawning


Exitare:
Linked CC issue: https://github.com/chromiecraft/chromiecraft/issues/7404

### Current Behaviour

There is a report on this quest but for different reasons, and it's the fact that quest/roleplay can't be started if a player did the battle recently. This, however is a different issue:

Currently it's possible to start the quest and get Centaurs to spawn.  The whole battle scene is missing, and as you can see on [THIS VIDEO](https://www.youtube.com/watch?v=BGt6jHSVVjc) it involves a lot of centaurs and allies fighting them in a quite epic way. After killing a few of them, Warlord Krom'zar will appear with two "bodyguards" (By the way, these two adds can be pulled separatedly from the boss, which I find a little strange).
Anyway, the problem is that after killing [Warlord Krom'zar](https://www.youtube.com/watch?v=BGt6jHSVVjc), the [banner](https://www.wowhead.com/wotlk/object=164690/kromzars-banner) doesn't spawn so we can't loot the [Piece of Krom'zar's Banner](https://www.wowhead.com/wotlk/item=11227/piece-of-kromzars-banner). Quest cannot be completed at all.

![ 2024-10-09 a las 11 21 45](https://github.com/user-attachments/assets/53622228-d9d0-41ac-bcc0-bffadc9d5a8c)

### Expected Blizzlike Behaviour

After killing [Warlord Krom'zar](https://www.wowhead.com/wotlk/npc=9456/warlord-kromzar), his banner should spawn right on top of his corpse to be looted by the player so we can obtain the [Piece of Krom'zar's Banner](https://www.wowhead.com/wotlk/item=11227/piece-of-kromzars-banner).

### Source

https://www.wowhead.com/wotlk/quest=4021/counterattack
https://www.wowhead.com/wotlk/npc=9456/warlord-kromzar
https://www.youtube.com/watch?v=BGt6jHSVVjc

### Steps to reproduce the problem

1) Have a Horde character of at least level 11 (recommended 20+).
2) Get quest 4021 from Regthar Deathgate.
3) Kill centaurs until Warlord Krom'zar spawns.
4) Kill Warlord Krom'zar and observe no banner spawns on his corpse.

### Extra Notes

The quest in itself behaves funky. The general didn't spawn for me at times when I tested it.

### AC rev. hash/commit

[...]

PR #21050, opened 2024-12-28
fix(DB/SAI) - Fixes quest Counterattack! sometimes not being completed and adds 1 ability to the creature.

TheSCREWEDSoftware:
Removes `NOT_REPEATABLE` and `WHILE_CHARMED` `event_flag`, allowing to always drop the flag no matter if it's CC'd or not from `Warlord Krom'zar`

and

Adds `Strike` to the `Warlord Krom'zar` sniffed from: `4.4.1.58158` and `1.15.5.57979`

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/21007
- Closes https://github.com/chromiecraft/chromiecraft/issues/7404

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1735350584050913500.sql
+++ b/data/sql/updates/pending_db_world/rev_1735350584050913500.sql
@@ -1,6 +1,6 @@
 --
 -- Removes NOT_REPEATABL` and WHILE_CHARMED event_flag, allowing to always drop the flag no matter if it's CC'd or not.
-UPDATE smart_scripts SET `event_flags` = `event_flags` &~1 &~512 WHERE `entryorguid` = 23576 AND ´source_type` = 0 AND `id` = 1;
+UPDATE smart_scripts SET `event_flags` = `event_flags` &~1 &~512 WHERE `entryorguid` = 23576 AND `source_type` = 0 AND `id` = 1;
 
 -- Adds Strike sniffed from: 4.4.1.58158 and 1.15.5.57979
 INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
--- a/data/sql/updates/pending_db_world/rev_1735350584050913500.sql
+++ b/data/sql/updates/pending_db_world/rev_1735350584050913500.sql
@@ -2,6 +2,6 @@
 -- Removes NOT_REPEATABL` and WHILE_CHARMED event_flag, allowing to always drop the flag no matter if it's CC'd or not.
 UPDATE smart_scripts SET `event_flags` = `event_flags` &~1 &~512 WHERE `entryorguid` = 23576 AND ´source_type` = 0 AND `id` = 1;
 
--- Adds `Strike` sniffed from: 4.4.1.58158 and 1.15.5.57979
+-- Adds Strike sniffed from: 4.4.1.58158 and 1.15.5.57979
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
@Exitare if you want to give a test it should always drop the flag. My testing was done via `.npc add temp 9456` indifferent if you have `.quest add 4021`

data/sql/updates/pending_db_world/rev_1735350584050913500.sql
```diff
@@ -0,0 +1,7 @@
+--
+-- Removes NOT_REPEATABL` and WHILE_CHARMED event_flag, allowing to always drop the flag no matter if it's CC'd or not.
+UPDATE smart_scripts SET `event_flags` = `event_flags` &~1 &~512 WHERE `entryorguid` = 23576 AND `source_type` = 0 AND `id` = 1;
```
Kitzunu, line 3: Back ticks

TheSCREWEDSoftware: Done

TheSCREWEDSoftware: Also changed from 2 indiviual values to grouped

image

data/sql/updates/pending_db_world/rev_1735350584050913500.sql
```diff
@@ -0,0 +1,7 @@
+--
+-- Removes NOT_REPEATABL` and WHILE_CHARMED event_flag, allowing to always drop the flag no matter if it's CC'd or not.
+UPDATE `smart_scripts` SET `event_flags` = `event_flags` &~(1|512) WHERE `entryorguid` = 23576 AND `source_type` = 0 AND `id` = 1;
+
+-- Adds Strike sniffed from: 4.4.1.58158 and 1.15.5.57979
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
```
sudlud, line 6: missing delete before

TheSCREWEDSoftware: Done

sudlud (approved):

data/sql/updates/pending_db_world/rev_1735350584050913500.sql
```diff
@@ -0,0 +1,8 @@
+--
+-- Removes NOT_REPEATABLE and WHILE_CHARMED event_flag, allowing to always drop the flag no matter if it's CC'd or not.
+UPDATE `smart_scripts` SET `event_flags` = `event_flags` &~(1|512) WHERE `entryorguid` = 9456 AND `source_type` = 0 AND `id` = 1;
```
sudlud, line 3: taking another look here this can't be correct.
regarding flag 512 the core implementation clearly states:
```cpp
SMART_EVENT_FLAG_WHILE_CHARMED         = 0x200,                     // Event occurs even if AI owner is charmed
```

so removing flag 512 should actually make the sai event not work anymore while charmed.

sudlud (changes requested):

data/sql/updates/pending_db_world/rev_1735350584050913500.sql
TheSCREWEDSoftware: Currently if you cc the creature and kill him he won't drop the flag

TheSCREWEDSoftware: @sudlud 

https://github.com/user-attachments/assets/2043beaa-a9f5-4cb2-9291-c042c39e5eea

sudlud:
so i looked into this and did some testing myself.

Even with your fix i was able to not get the flag spawned in some occasions.

i would say the root cause is probably some LOS missing or  something similiar, because i can get it fixed by just setting the target for the SAI entry to `SELF` instead of `ACTION INVOKER`.

This would be my preferred approach here. Can you please check if you get the same results when changing the sai target (without changing the event_flags).

TheSCREWEDSoftware:
> so i looked into this and did some testing myself.
> 
> Even with your fix i was able to not get the flag spawned in some occasions.
> 
> i would say the root cause is probably some LOS missing or something similiar, because i can get it fixed by just setting the target for the SAI entry to `SELF` instead of `ACTION INVOKER`.
> 
> This would be my preferred approach here. Can you please check if you get the same results when changing the sai target (without changing the event_flags).

I will change the PR to match this changes. I just tested it and it works (i tested with a stun also I think you mentioned in discord if not mistaken) he always drops the flag. 

https://github.com/user-attachments/assets/dc23b400-eba3-44ac-8cde-59dafed5bc44

sudlud (approved):

PR #14842, opened 2023-02-01
feat (core \log\db): areatrigger.dbc, sAreaTriggerStore, areatrigger_dbc

acidmanifesto:
Alrighty so I am not certain who thought it would be wonderful to not use the areatrigger.dbc file but in its place use a areatrigger table with 1217 entries where the dbc has 1220 entries. I am not doing a comparison to figure out the missing shit. It is excessive and not necessary. 

## Changes Proposed:
-  get rid of areatrigger table 
-  use and implement areatrigger.dbc
- implement areatrigger_dbc table
- Implement sAreaTriggerStore

## Issues Addressed:

- Closes areatrigger.dbc not being used
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15432

## Tests Performed:

- builds
- runs

## How to Test the Changes:

1. This is a heavily global core wide change.
2. Test affect boss scripts
3. test any areas of areatriggers.
4. Run a affect BG
5. SMART_SCRIPT_TYPE_AREATRIGGER needs to be tested
6. battlegroundEY and WS need to be tested.
7. HandleGoTriggerCommand gm cmd needs to be tested
8. various bosses needs to be tested that were adjusted in the file changes.
9. Continue betting pool with how long it will be before review and merge.  Currently it is between 6 months and 10 months.

Restore for original world areatrigger table that gets dropped with this PR:
https://gist.github.com/acidmanifesto/b5bffecf2446c707d386af23057d8cd9

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #13058, opened 2022-09-19
feat(DB/*_dbc): Update loc column names to reflect blizzlike assignments.

heyitsbench:
## Changes Proposed:
-  Changes ordering and names of localized column names in *_dbc tables.
Original order:
`enUS, enGB, koKR, frFR, deDE, enCN, zhCN, enTW, zhTW, esES, esMX, ruRU, ptPT, ptBR, itIT, Unk`
Proposed order:
`enUS/enGB, koKR, frFR, deDE, enCN/zhCN, zhTW, esES, esMX, ruRU, Unk, Unk, Unk, Unk, Unk, Unk, Unk`
Note that for the first and fifth column in the PR, only one locale is in the column name. E.g. `Name_Lang_zhCN`.

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

DBCs from various localized clients.

## Tests Performed:

- Tested that the server would not crash on loading the unfilled tables.

## How to Test the Changes:

1. Run the query.
2. Run the worldserver and make sure it runs as you'd expect.
Filling the tables shouldn't be necessary, since I don't know of any situation where the localized fields get sent to the client or are used for anything visible to the client.

## Known Issues and TODO List:

- [ ] enTW locale is nowhere to be found. It's very likely it's a part of that sixth column, but not known for certain.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #21091, opened 2025-01-04
feat(Dbc/Db): moved dbc tables to separate database

Razor2142:
## Changes Proposed:
A proposal to move the dbc override tables from the world database to separate one.

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] Consider removing the _dbc suffix from the table names. This requires changes to [this logic](https://github.com/azerothcore/azerothcore-wotlk/blob/712e4af5c2c9826fc604aaea5e49073a4165678e/src/server/database/Database/Field.cpp#L235).
- [ ] Add migration logic between databases to avoid possible data loss on existing databases. I would have prefered to move tables but there is no way to know the name of the databases from the sql update files.


data/sql/base/db_dbc/achievement_category_dbc.sql
```diff
@@ -0,0 +1,66 @@
...
+-- Table structure for table `achievement_category_dbc`
+--
+
+DROP TABLE IF EXISTS `achievement_category_dbc`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!50503 SET character_set_client = utf8mb4 */;
+CREATE TABLE `achievement_category_dbc` (
+  `ID` int NOT NULL DEFAULT '0',
+  `Parent` int NOT NULL DEFAULT '0',
+  `Name_Lang_enUS` varchar(100) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
```
heyitsbench, line 26: Small thing, and likely out of scope to be fixed by this PR, but just a reminder that these locale columns are incorrectly named and do not reflect the actual locale order as seen in client distributions, can be seen in [this PR](https://github.com/azerothcore/azerothcore-wotlk/pull/13058).

heyitsbench:
Of note is that `AreaTrigger.dbc` is not currently read by AC afaik, instead using the table `areatrigger` which just seems to mirror the data in that DBC iirc (more info can be found with [this PR](https://github.com/azerothcore/azerothcore-wotlk/pull/14842)). At some point, the handling of this data should be adjusted to match the other DBC files.

Nyeriah:
I'm not sure spell_dbc should be moved elsewhere, it's not a "replacement" of the dbc file, but rather the storage of serverside spells

heyitsbench:
> I'm not sure spell_dbc should be moved elsewhere, it's not a "replacement" of the dbc file, but rather the storage of serverside spells

IMO I think serverside spells should be split off into their own table, left in `world` DB, with `spell_dbc` being moved into this proposed separate DBC DB.

sudlud:
What I'm concerned about is the migration procedure for existing servers. 

- You would have to manually create a new database (most likely requiring to still know the root password of your MySQL installation)
- also correctly assign the rights for this database to the existing database user
- And then also correctly update the server configs with the new setting

Something like this? Or am I making wrong assumptions here

heyitsbench:
> What I'm concerned about is the migration procedure for existing servers.

Given that the various DBC tables are meant only to act as overrides to existing DBC files, I'm curious if a connection to a DBC DB could be made optional.

sudlud:
Some thoughts on this
- keep server side spells in the table in world db
- only drop empty dbc tables in this pr to not delete custom
Content automatically
- make dbc overwrites completely optional feature
- set feature default off with this pr
- provide a guide for people using the feature in how to migrate, enable, configure it

The impact on the normal user must be as minimal as possible imo and should never require any manual steps from them to continue using AC after an update.
Also we should never surprise the user with a drop table if there is custom content in their db imo.

sudlud:
Before this should continue any further, can you please elaborate on the advantages this pr would bring?

If it's just "having less to scroll through in the database viewer" then the impact on the normal user would probably not be justifying this change.

Kitzunu:
At this moment, these changes will not be considered. To not leave false hope I will close the PR 

Thanks for your work though. Maybe in the future it could be revised

PR #21103, opened 2025-01-06
fix(Player/SpellQueue): bandaid crashfix

sogladev:
## Changes Proposed:

This PR proposes changes to:

if there's an invalid spell in the queue, clear the queue and return

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/PlayerUpdates.cpp
+++ b/src/server/game/Entities/Player/PlayerUpdates.cpp
@@ -2335,7 +2335,7 @@ void Player::ProcessSpellQueue()
         SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(request.spellId);
         if (!spellInfo)
         {
-            LOG_ERROR("entities.player", "Player::ProcessSpellQueue: Invalid request");
+            LOG_ERROR("entities.player", "Player::ProcessSpellQueue: Invalid spell {}", request.spellId);
             SpellQueue.clear();
             break;
         }
--- a/src/server/game/Entities/Player/PlayerUpdates.cpp
+++ b/src/server/game/Entities/Player/PlayerUpdates.cpp
@@ -2333,6 +2333,12 @@ void Player::ProcessSpellQueue()
     {
         PendingSpellCastRequest& request = SpellQueue.front(); // Peek at the first spell
         SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(request.spellId);
+        if (!spellInfo)
+        {
+            LOG_ERROR("entities.player", "Player::ProcessSpellQueue: Invalid request");
+            SpellQueue.clear();
+            break;
+        }
         if (CanExecutePendingSpellCastRequest(spellInfo))
         {
             ExecuteOrCancelSpellCastRequest(&request);


src/server/game/Entities/Player/PlayerUpdates.cpp
```diff
@@ -2333,6 +2333,12 @@ void Player::ProcessSpellQueue()
     {
         PendingSpellCastRequest& request = SpellQueue.front(); // Peek at the first spell
         SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(request.spellId);
+        if (!spellInfo)
+        {
+            LOG_ERROR("entities.player", "Player::ProcessSpellQueue: Invalid request");
```
Kitzunu, line 2338: Should probably post request.spellId in the error

Kitzunu (approved):

Nyeriah:
cc @walkline

walkline:
From [this](https://gist.github.com/Nyeriah/11365392125a771ec9cd05a0d6493d32) crashlog, we see that the spell cast request has invalid values. The more interesting question is how the values became invalid.

I saw some other recent crashlogs - [one](https://gist.github.com/walkline/0472659b0b2ea583239847a51c1fcb70) and [two](https://gist.github.com/walkline/10ce55ef869879fe7d88bef65ed95728) - which make me think that something bad is happening, such as using a dangling pointer (using a deleted player) or a race condition (e.g., double parallel updates of the same map or instance).

Nyeriah:
Suddenly this crash became quite frequent, do we think this will help easen up the crashes?

Takenbacon:
From the vague research I did earlier in the week it shouldn't even be possible for spellInfo to be null in the container to begin with (under normal circumstances). Once I'm done with my work week I'll try to spend a little more time on it if possible.

Nyeriah:
Yes, it didnt help https://gist.github.com/Nyeriah/bbcd980360e462731503bb23f1c79c4c

Takenbacon:
> Yes, it didnt help https://gist.github.com/Nyeriah/bbcd980360e462731503bb23f1c79c4c

I spent a little time on it tonight and while there's a couple issues I saw nothing that would lead to a crash. I'd be curious to try disabling spell queue (it's a config option) and see if similar crashes are still occurring, would narrow it down specifically to spell queue or if spell queue container is just "unlucky" and being corrupted from something else.

sudlud:
so for the latest crashlog nyeriah provided this bandaid pr actually catched the error and then crashed on `SpellQueue.clear();` inside the bandaid fix😅

could probably use another "safer" way to clear `SpellQueue` to a working state until the root cause has been found

Takenbacon:
> so for the latest crashlog nyeriah provided this bandaid pr actually catched the error and then crashed on `SpellQueue.clear();` inside the bandaid fix😅
> 
> could probably use another "safer" way to clear `SpellQueue` to a working state until the root cause has been found

Not possible. An element in the container or even the player object itself point to invalid memory, all paths will inevitably lead to a seg fault once the memory is accessed during deallocation or use. The only solution is to fix the underlying issue, it's just a matter of trying to narrow down where it is or what causes it. It's significantly more difficult to do based off a crashlog without being able to check a memory dump in gdb.

PR #21150, opened 2025-01-12
fix(Core/Player): SpellQueue use a LockedQueue instead of std::deque

sogladev:
## Changes Proposed:

This PR proposes changes to:

Use a LockedQueue to prevent race conditions

Use the copy operator instead of a move assignment when adding elements to the SpellQueue

- Reverts as it doesn't fix the crash https://github.com/azerothcore/azerothcore-wotlk/pull/21103

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

Tested rotation of affli lock

## How to Test the Changes:

1. 
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/common/Threading/LockedQueue.h
+++ b/src/common/Threading/LockedQueue.h
@@ -150,5 +150,19 @@ public:
         std::lock_guard<std::mutex> lock(_lock);
         return _queue.empty();
     }
+
+    template<class Predicate>
+    bool contains(Predicate pred)
+    {
+        std::lock_guard<std::mutex> lock(_lock);
+        return std::any_of(_queue.begin(), _queue.end(), pred);
+    }
+
+    // Clears the queue.
+    void clear()
+    {
+        std::lock_guard<std::mutex> lock(_lock);
+        _queue.clear();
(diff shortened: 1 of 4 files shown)


walkline:
That crash remains a mystery to me 😁 Have you identified the race condition?

sogladev:
> That crash remains a mystery to me 😁 Have you identified the race condition?

No, my guess is there could be a race condition with how spellqueue is accessed. A map thread is consuming the queue in Player::Update and some other thread adds to the queue. Some lock should be required in this case. 

You posted a log on discord where a player is updated after deletion. This is more likely the cause to me, but I haven't found what causes it.

Takenbacon:
This is not the issue and will not fix anything. The queue is updated in the handler and will always be "safe" as it's updated in the same map thread as the player.

sogladev:
> This is not the issue and will not fix anything. The queue is added to in the packet handlers and will always be "safe" as it's updated in the same map thread as the player.

Thanks for your input @Takenbacon. Closing

Issue #21140, opened 2025-01-11
Rework auctionhouse Update causes people to be kicked if buy to many items at once


TheSCREWEDSoftware:
Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/7657

Triage Notes: https://github.com/chromiecraft/chromiecraft/issues/7657#issuecomment-2585396150

- I tested this with 2 different accounts.
- Made sure the "buyer" account had [auctionator](https://felbite.com/addon/4868-auctionator/)
- The "seller" account would have 40 linen clothes (.additem 2589 40) and would make 40 stacks of 1 auctions with 10 copper for bid and 20 bid for buyout (you will need to .modify money 100g)
- With the buyer try buying 11 linen clothes, you will be kicked but the items will be in your mail still (no error on the the worldserver)

Original Post Below: 

---

### What client do you play on?

enUS

### Faction

Alliance

### Content Phase:

Generic

### Current Behaviour

Buying 10 or more items at the same time from AH using Auctionator disconnects the player. Items are succesfully bought and can be found in mailbox after relog.

This problem started happening on Chromiecraft within the last month or so. For some reason server kicks the player. I did the same tests on other 3.3.5a servers with the same addon and there's no such issue.

### Expected Blizzlike Behaviour

Server shouldn't kick the player.

### Source

No source required.

### Steps to reproduce the problem

1. Install Auctionator addon (from Felbite or somewhere else)
2. Login to chromiecraft live server
3. On "Buy" tab, search for an item of which you can buy 10 or more (eg. Light Armor Kit, any potion etc.). It has to be 10 x 1 (10 stacks of 1) and not 1 x 10 (1 stack of 10) (Screenshot: https://imgur.com/a/CGKe6Py)
4. Enter the amount to buy (10+) and click buy.
5. You are disconnected.

### Extra Notes

There are different working versions of Auctionator but it doesn't matter which one you are using for this test. I will put 2 links below for easy downloading:
1- https://forum.warmane.com/showthread.php?t=412904&p=3030654&viewfull=1#post3030654
2- https://felbite.com/addon/4868-auctionator/

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/5f020de5466e3894914fe424229c4b9f3d138b2f

### Operating system

Ubuntu 22.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
[...]

PR #21280, opened 2025-01-27
fix(Core/Auction):  Fix disconnect  when people try to buy/sell to many items

skelUA:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/21140

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -1548,11 +1548,16 @@ uint32 WorldSession::DosProtection::GetMaxPacketCounterAllowed(uint16 opcode) co
         case CMSG_SOCKET_GEMS:                          // not profiled
         case CMSG_WRAP_ITEM:                            // not profiled
         case CMSG_REPORT_PVP_AFK:                       // not profiled
+            {
+                maxPacketCounterAllowed = 10;
+                break;
+            }
         case CMSG_AUCTION_LIST_ITEMS:                   // not profiled
         case CMSG_AUCTION_LIST_BIDDER_ITEMS:            // not profiled
         case CMSG_AUCTION_LIST_OWNER_ITEMS:             // not profiled
+        case CMSG_AUCTION_REMOVE_ITEM:
             {
-                maxPacketCounterAllowed = 10;
+                maxPacketCounterAllowed = 200;
                 break;
             }
 


sudlud:
if 100 is sufficient, we could remove the packets from the list so it uses the default value for those

Vtkachenko-Informa:
Some users and addons needs more 200 packets

TheSCREWEDSoftware:
I shall ping @Takenbacon for some input

TheSCREWEDSoftware:
> I shall ping @Takenbacon for some input

Context:

> AH Query packets were added to the throttler (here: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Server/WorldSession.cpp#L1551-L1553). I'm guessing a buy also seems to send a query request. We can just increase the limits for now, it was noted in the AH PR that the current packet throttler implementation is not ideal for this use case to begin with but that'll need a good bit more work to properly handle.

Takenbacon:
> Some users and addons needs more 200 packets

Which addons and were they designed to work on retail 3.3.5? I have a _very_ hard time believing a non-private server addon would allow for sending 200 query packets a second, retail servers are generally way more strict than we are and would cause disconnections.

Either way, increasing the limit is a bit of a mixed bag but I think it is the right move for now as a temporary solution. The concern is a malicious user could theoretically spam query opcodes and since the AH queries are quite slow with a big AH it will backlog the searcher threads effectively stopping AH searches. In the long run we need a generic solution for queuing up certain packets (not just AH but a couple other systems as well), but to do it properly and cleanly will take a little bit of work.

Vtkachenko-Informa:
> > Some users and addons needs more 200 packets
> 
> Which addons and were they designed to work on retail 3.3.5? I have a _very_ hard time believing a non-private server addon would allow for sending 200 query packets a second, retail servers are generally way more strict than we are and would cause disconnections.
> 
> Either way, increasing the limit is a bit of a mixed bag but I think it is the right move for now as a temporary solution. The concern is a malicious user could theoretically spam query opcodes and since the AH queries are quite slow with a big AH it will backlog the searcher threads effectively stopping AH searches. In the long run we need a generic solution for queuing up certain packets (not just AH but a couple other systems as well), but to do it properly and cleanly will take a little bit of work.

Test on auctinator try to canxel more than  100 auctions

skelUA:
> > Some users and addons needs more 200 packets
> 
> Which addons and were they designed to work on retail 3.3.5? I have a _very_ hard time believing a non-private server addon would allow for sending 200 query packets a second, retail servers are generally way more strict than we are and would cause disconnections.
> 
> Either way, increasing the limit is a bit of a mixed bag but I think it is the right move for now as a temporary solution. The concern is a malicious user could theoretically spam query opcodes and since the AH queries are quite slow with a big AH it will backlog the searcher threads effectively stopping AH searches. In the long run we need a generic solution for queuing up certain packets (not just AH but a couple other systems as well), but to do it properly and cleanly will take a little bit of work.

Test on auctinator try to cancel more than 100 auctions

sudlud:
i would imagine that you can set auctionator to some throttling limits for this?

also my comment from above still stands - why use magic number 200, just remove the packet type from the special handling and let it use the default value of 100. but maybe there#s other opinions on this

Takenbacon:
I did a little bit of testing with Auctionator, and while I'm not familiar with the addon, I am unable to reproduce any opcode spam with it. You'll need some more in depth reproduction steps. Also, you should know Auctionator technically isn't a 3.3.5 addon. It's 4.x back ported for private servers which is already questionable enough as is.

skelUA:
> I did a little bit of testing with Auctionator, and while I'm not familiar with the addon, I am unable to reproduce any opcode spam with it. You'll need some more in depth reproduction steps. Also, you should know Auctionator technically isn't a 3.3.5 addon. It's 4.x back ported for private servers which is already questionable enough as is.

try to add 200+ items on auction and cancel they  all from auctinator

sudlud:
@Takenbacon were you able to reproduce the issue maybe?

Takenbacon:
> @Takenbacon were you able to reproduce the issue maybe?

I have not quite got back to this. I'll see about  implementing packet throttling as one of my next changes.

PR #21285, opened 2025-01-28
fix(CORE) Add mutex protection to _pendingTimeSyncRequests map access

### LINKED ISSUE #21284: WordSession::Update() Crash
### Current Behaviour

https://gist.github.com/Nyeriah/cd98829965340ccb8f6c053231a30be2

No crash

### Source

_No response_

Unknown

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/dfd09a3541f3be672c3e2ff3a0a431547a229d3b

### Operating system

Ubuntu

### Custom changes or Modules

_No response_

blinkysc:
WorldSession::SendTimeSync() function was experiencing a segmentation fault when accessing the _pendingTimeSyncRequests map from multiple threads. Added mutex synchronization to prevent concurrent map access that was causing seg fault

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes #21284

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -1635,13 +1635,14 @@ void WorldSession::ResetTimeSync()
 
 void WorldSession::SendTimeSync()
 {
+    std::lock_guard<std::mutex> guard(_timeSyncLock);
+
+    // Create the packet and increment the counter under the lock
     WorldPacket data(SMSG_TIME_SYNC_REQ, 4);
     data << uint32(_timeSyncNextCounter);
     SendPacket(&data);
 
-    std::lock_guard<std::mutex> guard(_timeSyncLock);
-    _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
-
+    _pendingTimeSyncRequests.emplace(_timeSyncNextCounter, getMSTime());
     // Schedule next sync in 10 sec (except for the 2 first packets, which are spaced by only 5s)
     _timeSyncTimer = _timeSyncNextCounter == 0 ? 5000 : 10000;
     _timeSyncNextCounter++;
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -1639,10 +1639,8 @@ void WorldSession::SendTimeSync()
     data << uint32(_timeSyncNextCounter);
     SendPacket(&data);
 
-    {
-        std::lock_guard<std::mutex> guard(_timeSyncLock);
-        _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
-    }
+    std::lock_guard<std::mutex> guard(_timeSyncLock);
+    _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
 
     // Schedule next sync in 10 sec (except for the 2 first packets, which are spaced by only 5s)
     _timeSyncTimer = _timeSyncNextCounter == 0 ? 5000 : 10000;
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -1639,7 +1639,7 @@ void WorldSession::SendTimeSync()
     SendPacket(&data);
 
     {
-        std::lock_guard<std::mutex> guard(_timeSyncLock); 
+        std::lock_guard<std::mutex> guard(_timeSyncLock);
         _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
     }
 
--- a/src/server/game/Server/WorldSession.cpp
+++ b/src/server/game/Server/WorldSession.cpp
@@ -1638,7 +1638,10 @@ void WorldSession::SendTimeSync()
     data << uint32(_timeSyncNextCounter);
     SendPacket(&data);
 
-    _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
+    {
+        std::lock_guard<std::mutex> guard(_timeSyncLock); 
+        _pendingTimeSyncRequests[_timeSyncNextCounter] = getMSTime();
+    }
 
     // Schedule next sync in 10 sec (except for the 2 first packets, which are spaced by only 5s)
     _timeSyncTimer = _timeSyncNextCounter == 0 ? 5000 : 10000;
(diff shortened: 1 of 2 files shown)


blinkysc:
@nl-saw look good?

Takenbacon:
Why would this need to be locked? Where do you see this being accessed by multiple threads?

blinkysc:
> Why would this need to be locked? Where do you see this being accessed by multiple threads?

#4  WorldSession::SendTimeSync (this=this@entry=0x62d2089d600) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:1641
#5  0x00005c50383281e2 in WorldSession::Update (this=0x62d2089d600, diff=18, updater=...) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:494
#6  0x00005c5038254581 in Map::Update (this=0x62d97045e80, t_diff=0, s_diff=18) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/Map.cpp:753
#7  0x00005c503826aacd in MapUpdateRequest::call (this=0x6275e115340) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:44
#8  0x00005c503826a428 in MapUpdater::WorkerThread (this=0x5c5038fc7a00 <MapMgr::instance()::instance+216>) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:158
#9  0x0000062e3b2dc253 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6

Takenbacon:
> > Why would this need to be locked? Where do you see this being accessed by multiple threads?
> 
> #4 WorldSession::SendTimeSync (this=this@entry=0x62d2089d600) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:1641 #5 0x00005c50383281e2 in WorldSession::Update (this=0x62d2089d600, diff=18, updater=...) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:494 #6 0x00005c5038254581 in Map::Update (this=0x62d97045e80, t_diff=0, s_diff=18) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/Map.cpp:753 #7 0x00005c503826aacd in MapUpdateRequest::call (this=0x6275e115340) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:44 #8 0x00005c503826a428 in MapUpdater::WorkerThread (this=0x5c5038fc7a00 MapMgr::instance()::instance+216) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:158 #9 0x0000062e3b2dc253 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6

That is the standard process flow of the server, there is nothing to indicate a threading crash or a necessity for locking.

Maybe we need to add this to a wiki or something as there is often misunderstanding the threading design of the server but the general rule of thumb is:

Every player/worldsession can only belong to one map thread at a time (obviously), thus member values within a player or worldsession object 99.99% of the time are thread safe, which is why we don't have a billion arbitrary locks everywhere.

There are two slight exceptions to this that are special cases:

Every session belongs to a network thread that feeds the incoming (locked) packet queue. Anything the network thread touches on the worldsession must be locked, there are only a couple like the packet queue and latency values.

Finally, and this is generally the problem most of the time, we have cross map interaction between objects. Because Maps are updated in parallel we shouldn't be accessing another player/session object from another map during the Map update. It's not often but sometimes you'll run into something that violates this rule. Something like iterating and accessing group members or casting a spell on a player on another map is a big no-no which leads to all sorts of odd and semi rare segfaults. The main point is this tends to be a fundamental design flaw in that section of code and needs to be reworked rather than just needlessly slapping locks everywhere instead of addressing the bigger problem.

blinkysc:
> > > Why would this need to be locked? Where do you see this being accessed by multiple threads?
> > 
> > 
> > #4 WorldSession::SendTimeSync (this=this@entry=0x62d2089d600) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:1641 #5 0x00005c50383281e2 in WorldSession::Update (this=0x62d2089d600, diff=18, updater=...) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Server/WorldSession.cpp:494 #6 0x00005c5038254581 in Map::Update (this=0x62d97045e80, t_diff=0, s_diff=18) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/Map.cpp:753 #7 0x00005c503826aacd in MapUpdateRequest::call (this=0x6275e115340) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:44 #8 0x00005c503826a428 in MapUpdater::WorkerThread (this=0x5c5038fc7a00 MapMgr::instance()::instance+216) at /root/env/chromiecraft/azerothcore-wotlk/src/server/game/Maps/MapUpdater.cpp:158 #9 0x0000062e3b2dc253 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
> 
> That is the standard process flow of the server, there is nothing to indicate a threading crash or a necessity for locking.
> 
> Maybe we need to add this to a wiki or something as there is often misunderstanding the threading design of the server but the general rule of thumb is:
> 
> Every player/worldsession can only belong to one map thread at a time (obviously), thus member values within a player or worldsession object 99.99% of the time are thread safe, which is why we don't have a billion arbitrary locks everywhere.
> 
> There are two slight exceptions to this that are special cases:
> 
> Every session belongs to a network thread that feeds the incoming (locked) packet queue. Anything the network thread touches on the worldsession must be locked, there are only a couple like the packet queue and latency values.
> 
> Finally, and this is generally the problem most of the time, we have cross map interaction between objects. Because Maps are updated in parallel we shouldn't be accessing another player/session object from another map during the Map update. It's not often but sometimes you'll run into something that violates this rule. Something like iterating and accessing group members or casting a spell on a player on another map is a big no-no which leads to all sorts of odd and semi rare segfaults. The main point is this tends to be a fundamental design flaw in that section of code and needs to be reworked rather than just needlessly slapping locks everywhere instead of addressing the bigger problem.

Ok cool. Yeah guess would be locks everywhere if you wanted to catch edge cases. I can close PR. Thanks for exp @Takenbacon

PR #21287, opened 2025-01-28
fix(Scripts/AzjolNerub): Boss Herald Volazj - Insanity spell cast

p-tkachuk:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:
Insanity spell wasn't cast due to incorrect check for agro (by calling IsInCombatWith() method for wrong target)
Also now visage will look like other party members

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:
1. Make two characters and level them to 80 (.level 79)
2. Put them in one party.
3. `.go xyz 519 -478 27 619`  (both players)
4. Fight with the boss.
5. Boss will cast Insanity (which works now), each party member will be phase shifted and will fight "evil" versions of party members.

- [ ]


src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
```diff
@@ -362,7 +361,7 @@ class spell_herald_volzaj_insanity : public SpellScript
         {
             targets.remove_if([this](WorldObject* targetObj) -> bool
             {
-                return !targetObj || !targetObj->IsPlayer() || !targetObj->ToPlayer()->IsInCombatWith(GetCaster()) ||
+                return !targetObj || !targetObj->IsPlayer() || !GetCaster()->IsInCombatWith(targetObj->ToPlayer()) ||
```
sudlud, line 364: why do you think this is necessary? i would imagine that the order should be irrelevant here as both should be in combat with each other

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
```diff
@@ -405,19 +404,19 @@ class spell_herald_volzaj_insanity : public SpellScript
             for (std::list<WorldObject*>::const_iterator itr2 = targets.begin(); itr2 != targets.end(); ++itr2)
             {
                 // Should not make clone of current player target
-                Player const* plrClone = *itr2 ? (*itr2)->ToPlayer() : nullptr;
-                if (!plrClone || plrClone == plrTarget)
+                Player* plrClone = *itr2 ? (*itr2)->ToPlayer() : nullptr;
```
sudlud, line 407: why remove the const?

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
```diff
@@ -405,19 +404,19 @@ class spell_herald_volzaj_insanity : public SpellScript
...
+                if (!plrClone || plrClone == plrTarget || !plrClone->IsAlive())
                 {
                     continue;
                 }
 
                 if (Unit* summon = caster->SummonCreature(NPC_TWISTED_VISAGE, plrClone->GetPosition(), TEMPSUMMON_CORPSE_DESPAWN, 0))
                 {
+                    plrClone->CastSpell(summon, SPELL_CLONE_PLAYER, true);
+
```
sudlud, line 416: why move this line up here and not just change `plrTarget` to `plrClone` right in the line where it was?

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
p-tkachuk: Yes, you are right, it should be irrelevant... but it is not. That what actually causing a bug in the first place.
That method check ThreatList, and for player it is empty. (I'm still new here... can player have a threat list at all?).
So this method call should be reversed, and called for the boss. In that case the threat list is not empty and returns correct value.

https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Unit/Unit.cpp#L21111

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
p-tkachuk: Again, I'm new in c++ :)
I was not able to call "CastSpell" method on const variable.

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
p-tkachuk: I've changed the order because it looks more logical to me - first of all lets change appearance, then do everything else.

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
p-tkachuk: Oh, for the const we can use "." instead of "->", right?

src/server/scripts/Northrend/AzjolNerub/ahnkahet/boss_herald_volazj.cpp
sudlud: alright, i just thought it would make the diff of this pr cleaner if the line is not beeing moved.

sudlud (approved):

PR #21288, opened 2025-01-28
fix(Core/Threading) Refactored Map class - some code optimization

nl-saw:
~Changed i_scriptLock to mutex and~ some general code refactoring

## Changes Proposed:
- Removed unnecessary iteration
- Some code cleaning

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:
- [X] Tested in-game by other community members/someone else other than the author/has been live on production servers.

## How to Test the Changes:

No functional changes have been made. The changes primarily involve code refactoring aimed at potential performance improvements, with enhanced code structure and more stability.
There should be no issues or deficiencies resulting from these changes.

1.
2.
3.

- [ ]
- [ ]


Takenbacon:
I do have a couple possible concerns but it's super hard to review off my phone, will try once I'm back in town.

Main one being with merging the two loops, sessions and player objects are destroyed in the session update, which means adding the player update after might be pointing to invalid memory. I see it retains that odd "optimization" with checking for t_diff so it may not be consistent

Takenbacon:
Ok I had time to glance at it.

You can ignore my first comment, I forgot session/player deletions are handled in "unsafe" context (world tick) so there shouldn't be any issue with player being deleted during the map session update.

As for i_scriptLock, I suspect (but I am not 100% sure) the original intention looks to be to prevent recursive calls rather than actually a threading/mutex protection. With a standard mutex a recursive call will deadlock.

Takenbacon:
> I checked the files making use of i_scriptlock. The only recursive call I found is
> 
> ```
> bool Map::HasEnoughWater(WorldObject const* searcher, float x, float y, float z) const
> {
>     LiquidData const& liquidData = const_cast<Map*>(this)->GetLiquidData(searcher->GetPhaseMask(), x, y, z, searcher->GetCollisionHeight(), MAP_ALL_LIQUIDS);
>     return (liquidData.Status & MAP_LIQUID_STATUS_SWIMMING) != 0 && HasEnoughWater(searcher, liquidData);
> }
> ```
> 
> Could this be problematic in your experience?

Couldn't tell you, the code originated from commit https://github.com/TrinityCore/TrinityCore/commit/bd0dbcc9ad239e608ebc881558f64877b60d14d8

The concern from the original author appears to have been a recursive/dead loop call from ScriptsProcess calling ScriptsStart. No idea if it ever was an actual problem or just a precaution. A mutex however changes this functionality though as a standard mutex into a mutex will deadlock, causing exactly what the original author was trying to prevent.

Takenbacon:
> I see, my main goal there was to eliminate potential race condition. What would you suggest, keep current logic around the script lock and just keep the rest?

Yeah that would be the safest bet for now. There shouldn't be any race conditions in regards to that section of code anyways, it's all processed through the same map update thread. Everything else looks good.

src/server/game/Maps/Map.cpp
```diff
@@ -624,90 +617,88 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
...
 
-    // the player iterator is stored in the map object
-    // to make sure calls to Map::Remove don't invalidate it
+    // Update players and their associated objects
     for (m_mapRefIter = m_mapRefMgr.begin(); m_mapRefIter != m_mapRefMgr.end(); ++m_mapRefIter)
     {
         Player* player = m_mapRefIter->GetSource();
-
-        if (!player || !player->IsInWorld())
-            continue;
```
walkline, line 662: Hm, you don't like the early return/continue approach? As for me, it improves readability.

src/server/game/Maps/Map.cpp
```diff
@@ -588,34 +588,27 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
...
         {
-            //player->Update(t_diff);
+            // Update session
             WorldSession* session = player->GetSession();
             MapSessionFilter updater(session);
             session->Update(s_diff, updater);
+
+            // update players at tick
+            if (!t_diff)
+                player->Update(s_diff);
```
walkline, line 604: Nice. This is a major reason why we should merge this.

walkline (commented):
I like it. Just have a minor nitpick overall.

src/server/game/Maps/Map.cpp
Kitzunu: Early return is to be preferred

src/server/game/Maps/Map.cpp
```diff
@@ -694,20 +685,21 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
                             updateList.push_back(cre);
                 ref = ref->next();
             }
-            for (std::vector<Creature*>::const_iterator itr = updateList.begin(); itr != updateList.end(); ++itr)
-                VisitNearbyCellsOf(*itr, grid_object_update, world_object_update, grid_large_object_update, world_large_object_update);
+
+            for (Creature* cre : updateList) {
```
sudlud, line 689: Put brace on new line

@Kitzunu codestyle check did not catch this somehow, probably because it's a loop?

src/server/game/Maps/Map.cpp
```diff
@@ -694,20 +685,21 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
...
+            }
         }
     }
 
-    for (_transportsUpdateIter = _transports.begin(); _transportsUpdateIter != _transports.end();) // pussywizard: transports updated after VisitNearbyCellsOf, grids around are loaded, everything ok
+    // Update transports - pussywizard: transports updated after VisitNearbyCellsOf, grids around are loaded, everything ok
+    for (_transportsUpdateIter = _transports.begin(); _transportsUpdateIter != _transports.end();)
     {
         MotionTransport* transport = *_transportsUpdateIter;
         ++_transportsUpdateIter;
```
sudlud, line 699: Can't this loop be shortened like the creature loop above, too? Or am I missing something here

walkline (approved):

PR #21296, opened 2025-01-30
fix(Core/Transport): Fix for passengers dropped in the middle of nowhere 

p-tkachuk:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:
Fix issue when vehicle passengers where dropped in the middle of nowhere on Transport teleport

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:
TWO players required!

0. Create 2 characters, level up to 80. (.level 79)
1. `.go xyz -4263 -11331 5.4 530`
2. Obtain vehicle with additional seats on it ( `.additem 44083` )
3. Invite another player to party
4. Second player should sit on your vehicle
5. Enter any transport (like boat or zeppelin) 
6. Wait until it teleports to another location

As a result you party member will be teleported along with you, unlike before the fix when he was left in previous location

- [ ] It would be probably awesome to teleport player without ejecting him from the seat
- [ ]

### DIFF
--- a/src/server/game/Entities/Transport/Transport.cpp
+++ b/src/server/game/Entities/Transport/Transport.cpp
@@ -584,7 +584,8 @@ void MotionTransport::DelayedTeleportTransport()
                     TransportBase::CalculatePassengerPosition(destX, destY, destZ, &destO, x, y, z, o);
 
                     Player* player = obj->ToPlayer();
-                    // Vehicle passengers are dropped in the midle of nowhere, so lets try to eject them, add to the transport and teleport
+                    // Vehicle passengers are dropped in the middle of nowhere, so lets try to eject them, add to the transport and teleport
+                    // this needs to be revisited to properly restore vehicles with passengers after transport teleportation
                     if (player->IsVehicle())
                         if (Vehicle* vehicleKit = player->GetVehicleKit())
                             for (SeatMap::iterator itr = vehicleKit->Seats.begin(); itr != vehicleKit->Seats.end(); ++itr)
--- a/src/server/game/Entities/Transport/Transport.cpp
+++ b/src/server/game/Entities/Transport/Transport.cpp
@@ -582,7 +582,21 @@ void MotionTransport::DelayedTeleportTransport()
                     float destX, destY, destZ, destO;
                     obj->m_movementInfo.transport.pos.GetPosition(destX, destY, destZ, destO);
                     TransportBase::CalculatePassengerPosition(destX, destY, destZ, &destO, x, y, z, o);
-                    if (!obj->ToPlayer()->TeleportTo(newMapId, destX, destY, destZ, destO, TELE_TO_NOT_LEAVE_TRANSPORT))
+
+                    Player* player = obj->ToPlayer();
+                    // Vehicle passengers are dropped in the midle of nowhere, so lets try to eject them, add to the transport and teleport
+                    if (player->IsVehicle())
+                        if (Vehicle* vehicleKit = player->GetVehicleKit())
+                            for (SeatMap::iterator itr = vehicleKit->Seats.begin(); itr != vehicleKit->Seats.end(); ++itr)
+                                if (Player* passenger = ObjectAccessor::GetPlayer(*player, itr->second.Passenger.Guid))
+                                {
+                                    passenger->ExitVehicle();
+                                    AddPassenger(passenger, true);
+                                    if (!passenger->TeleportTo(newMapId, destX, destY, destZ, destO, TELE_TO_NOT_LEAVE_TRANSPORT))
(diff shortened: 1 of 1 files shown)


sudlud:
what would be the blizzlike behaviour on transport teleport with players in a vehicle? you're ejecting them with this PR and then teleporting them with the transport, which is of course better than not teleporting them and letting them fall to the ground.

but i would assume that the blizzlike handling would be to teleport player, vehicle and all passengers 1:1 with the transport?

p-tkachuk:
> what would be the blizzlike behaviour on transport teleport with players in a vehicle? you're ejecting them with this PR and then teleporting them with the transport, which is of course better than not teleporting them and letting them fall to the ground.
> 
> but i would assume that the blizzlike handling would be to teleport player, vehicle and all passengers 1:1 with the transport?

Unfortunately I do not know how it works on blizz servers. But yes, most likely passengers should be teleported keeping their position on vehicle.

On AC, during teleportation, vehicle is destroyed and created in new spot, so we are loosing information about vehicle passengers. We need to keep this data somehow and put players back on vehicle after teleportation... but I'm not that good yet to implement such thing. So at this point I just came up with a simpler solution.

src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -582,8 +582,25 @@ void MotionTransport::DelayedTeleportTransport()
...
+                        if (player->IsVehicle())
+                            if (Vehicle* vehicleKit = player->GetVehicleKit())
+                                for (SeatMap::iterator itr = vehicleKit->Seats.begin(); itr != vehicleKit->Seats.end(); ++itr)
+                                    if (Unit* unit = ObjectAccessor::GetUnit(*player, itr->second.Passenger.Guid))
+                                        if (Player* passenger = unit->ToPlayer())
+                                        {
+                                            passenger->ExitVehicle();
+                                            AddPassenger(passenger, true);
+                                            if (!passenger->TeleportTo(newMapId, destX, destY, destZ, destO, TELE_TO_NOT_LEAVE_TRANSPORT))
+                                                _passengers.erase(obj);
```
sudlud, line 598: i would say this can't be correct here as you're handling a passenger and not the main obj/player.

this would also lead to calling `_passengers.erase(obj);` multiple times for one player `obj`

src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -582,8 +582,25 @@ void MotionTransport::DelayedTeleportTransport()
...
-                    if (!obj->ToPlayer()->TeleportTo(newMapId, destX, destY, destZ, destO, TELE_TO_NOT_LEAVE_TRANSPORT))
-                        _passengers.erase(obj);
+
+                    if (Player* player = obj->ToPlayer())
+                    {
+                        // Vehicle passengers are dropped in the midle of nowhere, so lets try to eject them, add to transport and teleport
+                        if (player->IsVehicle())
+                            if (Vehicle* vehicleKit = player->GetVehicleKit())
+                                for (SeatMap::iterator itr = vehicleKit->Seats.begin(); itr != vehicleKit->Seats.end(); ++itr)
+                                    if (Unit* unit = ObjectAccessor::GetUnit(*player, itr->second.Passenger.Guid))
```
sudlud, line 592: i think there should be `ObjectAccessor:GetPlayer()` available to directly fetch a player by guid, so you don't need the 2 steps here

src/server/game/Entities/Transport/Transport.cpp
```diff
@@ -582,8 +582,25 @@ void MotionTransport::DelayedTeleportTransport()
                     float destX, destY, destZ, destO;
                     obj->m_movementInfo.transport.pos.GetPosition(destX, destY, destZ, destO);
                     TransportBase::CalculatePassengerPosition(destX, destY, destZ, &destO, x, y, z, o);
-                    if (!obj->ToPlayer()->TeleportTo(newMapId, destX, destY, destZ, destO, TELE_TO_NOT_LEAVE_TRANSPORT))
-                        _passengers.erase(obj);
+
+                    if (Player* player = obj->ToPlayer())
```
sudlud, line 586: Also at this point it's 100 % safe that obj is a player, in the previous version it also wasn't checked. Just a note

src/server/game/Entities/Transport/Transport.cpp
p-tkachuk: Oops, this is a copy-pasta of a line from below... fixed.

src/server/game/Entities/Transport/Transport.cpp
p-tkachuk: Removed the if statement.

sudlud (approved):

PR #21380, opened 2025-02-10
fix(Core/Pet): Implemented sound playback when dismissing warlock's pet.

r4d1sh:
Implememted `SMSG_PET_DISMISS_SOUND` which triggers the corresponding sound when a warlock dismisses their pet.

The pull request is provided 'as is' in the interest of the entire community. If the AzerothCore maintainers refuse to accept it due to incorrect formatting (templates etc) or some other nonsense, you can add it to your server manually.


Nyeriah:
It takes NO effort filling the pull request template, literaly. We offer support to hundreds of users, opening issues and pull requests, I am not sure what makes you think you are better than them

PR #21382, opened 2025-02-10
fix(Core/Chat): Fix lower security check for GM commands.

r4d1sh:
Fixed the system that prevents the use of certain GM commands on targets whose access level (GMLevel) is higher than yours.

For example, the .summon command. Before the fix, you could use it on characters with a higher access level than yours. After the fix, you will receive a notification that your access level is too low.

**The pull request is provided 'as is' in the interest of the entire community, in the spirit of open-source. If the AzerothCore maintainers refuse to accept it due to incorrect formatting (templates etc) or some other nonsense, you can add it to your server manually.**


sudlud:
thanks for the PR, please use the PR template for future PRs.

sudlud (approved):

PR #21385, opened 2025-02-10
fix(Core/PacketIO): Implemented CMSG_BUSY_TRADE and CMSG_IGNORE_TRADE.

r4d1sh:
When a player's game client receives a message that another player wants to initiate a trade, the client can send various responses:

`CMSG_BEGIN_TRADE` - Agree to trade.
`CMSG_BUSY_TRADE` - Refusal to trade and the message "$n is busy right now" for the player initiating the trade. This occurs if you are busy in trading at the AH or have enabled automatic trade rejection in the client settings.
`CMSG_IGNORE_TRADE` - Refusal to trade and a message to the initiating player that their target is ignoring them.

Currently, `CMSG_BUSY_TRADE` is not processed by the server. This means that the trade window will not open if you offer it to a player who is busy trading at the AH. Meanwhile, the server will consider that the trade has already started and will not allow both players to initiate it further until they log out. The proposed changes implement `CMSG_BUSY_TRADE` and now everything works fine. This also fixes the functionality of the automatic trade rejection option in the game client settings.

As for `CMSG_IGNORE_TRADE`, it is not processed by the server, but the server does not allow the initiation of a trade if you are being ignored (`WorldSession::HandleInitiateTradeOpcode`). The proposed changes remove the old check for ignoring and allow the client to make the decision using `CMSG_IGNORE_TRADE`, as it was probably originally intended.

Original PR: https://github.com/TrinityCore/TrinityCore/pull/29570

**The pull request is provided 'as is' in the interest of the entire community, in the spirit of open-source. If the AzerothCore maintainers refuse to accept it due to incorrect formatting (templates etc) or some other nonsense, you can add it to your server manually.**


Kitzunu:
While I disagree with your childish behavior in the slightest, I do appreciate your contribution.

Kitzunu (approved):
lgtm

PR #21412, opened 2025-02-12
feat(DB/npc_vendor) Add comment column

aradep:
## Changes Proposed:
Adds comment column to `npc_vendor`

## How to Test the Changes:
Run the SQL file and check the npc_vendor table.

I know the IDs are already there but this makes it a bit easier to search and read at a glance.

image

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1739324024224393200.sql
+++ b/data/sql/updates/pending_db_world/rev_1739324024224393200.sql
@@ -1,0 +1,10 @@
+-- Add comment column to npc_vendor
+ALTER TABLE `npc_vendor`
+ADD COLUMN `Comment` VARCHAR(255) NULL;
+
+-- Fill comment with "Vendor name - Item name".
+UPDATE `npc_vendor`
+INNER JOIN `creature_template` ON `npc_vendor`.`entry` = `creature_template`.`entry`
+INNER JOIN `item_template` ON `npc_vendor`.`item` = `item_template`.`entry`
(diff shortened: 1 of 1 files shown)


sogladev:
This [this violates Third Normal Form (3NF)](https://www.geeksforgeeks.org/third-normal-form-3nf/) as creature_template.name and item_template.name are already present in their respective tables. Imo it's not a good idea to add a redundant column. 

This looks like a use case for a view instead. 

> A view can dynamically generate the Comment as needed without altering the underlying table structure or introducing redundancy.

You can query the npc_vendor_view to get the Comment column without modifying the npc_vendor table directly.

```
-- drop `Comment` if you already have it
ALTER TABLE `npc_vendor`
DROP COLUMN `Comment`;
CREATE VIEW npc_vendor_view AS
SELECT 
    nv.*,
    CONCAT(ct.name, ' - ', it.name) AS Comment
FROM 
    npc_vendor nv
INNER JOIN 
    creature_template ct ON nv.entry = ct.entry
INNER JOIN 
    item_template it ON nv.item = it.entry;
```

Alternatively, [Keira3](https://www.azerothcore.org/Keira3/) already shows a good amount in the vendor tab.

TheSCREWEDSoftware:
This change brings nothing of good and only will bring pollution to database. A quick query will achieve the same result. 

Check sogla's exampel of a query or use keira as both would achieve the same result.

aradep:
I agree it's not best practice but it's the same as the comment column in the loot_template tables.

Issue #14605, opened 2023-01-09
No odd acore_characters/item_instance guid in mysql


Conan1996-2:
### Current Behaviour

I created 10 Characters, looked at the item_guid and see they are even numbered, Is this intentional?

### Expected Blizzlike Behaviour

the guid should be sequential.

### Steps to reproduce the problem

Create an account
Create characters
Open database and look at the acore_characters/item_instance

### AC rev. hash/commit

This doesn't matter

### Operating system

Ubuntu

### Custom changes or Modules

![odd](https://user-images.githubusercontent.com/71103778/211428670-1a590f1e-539c-4484-8fca-d45d26117004.png)

heyitsbench:
> This doesn't matter

Yes it does.

Conan1996-2:
AzerothCore rev. aae46cc8afeb 2023-01-02 20:08:19 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 3.0.2 15 Mar 2022 (library: OpenSSL 3.0.2 15 Mar 2022)
Using Boost version: 1.74.0
Using MySQL version: 80031
Using CMake version: 3.22.1
Compiled on: Linux 5.15.0-1021-raspi
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in /mnt/Plexdata/WOW/wrath/azerothcore/build/Data/maps. Total size: 291014951 bytes
vmaps directory located in /mnt/Plexdata/WOW/wrath/azerothcore/build/Data/vmaps. Total size: 658130721 bytes
mmaps directory located in /mnt/Plexdata/WOW/wrath/azerothcore/build/Data/mmaps. Total size: 2320900676 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS 
Using World DB: ACDB 335.8-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
No modules enabled

Nyeriah:
At first glance seems to be true 
![image](https://user-images.githubusercontent.com/47818697/211432990-826c7c56-d8a3-40e0-b81e-7a3563cd3cf9.png)

PR #21473, opened 2025-02-16
feat(Core/Characters): Cleanup character item_instance and character_…

Kitzunu:
…inventory guid gaps

* closes https://github.com/azerothcore/azerothcore-wotlk/issues/14605

This is defaulted to off with config.

It will reorder item_instance guid and update the responding in character_inventory at startup

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

- Enable the config
- create characters and see guid gaps in item_instance
- restart server
- see that gaps are gone
- see that you still have your items

- [ ]
- [ ]


src/server/game/Tools/CharacterDatabaseCleaner.cpp
```diff
@@ -154,3 +157,11 @@ void CharacterDatabaseCleaner::CleanCharacterQuestStatus()
 {
     CharacterDatabase.DirectExecute("DELETE FROM character_queststatus WHERE status = 0");
 }
+
+void CharacterDatabaseCleaner::CleanCharacterItem()
+{
+    CharacterDatabase.DirectExecute("CREATE TEMPORARY TABLE temp_guid_mapping AS SELECT guid AS old_guid, ROW_NUMBER() OVER (ORDER BY guid) AS new_guid FROM item_instance;");
+    CharacterDatabase.DirectExecute("UPDATE item_instance i JOIN temp_guid_mapping m ON i.guid = m.old_guid SET i.guid = m.new_guid;");
+    CharacterDatabase.DirectExecute("UPDATE character_inventory ci JOIN temp_guid_mapping m ON ci.item = m.old_guid SET ci.item = m.new_guid;");
```
walkline, line 165: We need to do the same for `mail_items`, `guild_bank_item`, `recovery_item`, `auctionhouse` and probably some more tables (use a collective brain?).

walkline (commented):
I have concerns about large databases. I assume this will attempt to update every record. Imagine a database with 10+ million items - updating all of them at once would be extremely slow, especially with primary keys and indexes affecting the update process.

To me, this approach could work if performed occasionally, like once a year during scheduled maintenance, with downtime depending on the database size. However, I think it would be too slow to run on startup.

For startup, we could use a different approach, such as:
1. Identifying the gaps.
2. Taking the newest items (with the highest GUIDs) and using them to fill the gaps without modifying the others.

src/server/game/Tools/CharacterDatabaseCleaner.cpp
Kitzunu: ah true they would all hava record in item_instance

Kitzunu:
@Winfidonarleyan do you have any suggestions on how to improve it?

Kitzunu:
We dont necessarily need to have it in the core. We could make a Python script for it.

The upside to having it on load is that it probably isnt safe to run it during runtime

Winfidonarleyan:
> @Winfidonarleyan do you have any suggestions on how to improve it?

This is db. I do not know how to make it better.

Nyeriah:
Probably worth a disclaimer that this can slow down start up considerably.

Nyeriah:
What we should probably do instead is implement proper guid recycle, not just for items but everything else, similar to how instances recycle IDs.

 I think adding the procedure to start up like this is a bit self defeating. It would require a largely populated server for it to make any difference, but then again going through 1mil+ item entries probably slows it down to the point it’s unpractical on restarts

Kitzunu:
The way guid is set today for objectguid is a variable on startup gets MAX() + 1 from DB and then the count is remembered coreside during runtime. And the cause of the gaps is due to temporary items are created when checking if it can be equipped which upps the guid count coreside.

Kitzunu:
I will revise this later. Dont have time to spend on digging into it rn

Issue #3971, opened 2019-04-15
[DEV] Use sphere-plane collision tests for movement spells


Pursche:
This issue is supposed to be a discussion about heightmaps, height checks and navigation.
Movement spells include blink, shadowstep, charge etc.

I have been looking into height testing on WoW terrains and I believe that I have found the reason that spells like Blink sometimes teleports the player through the ground.

The way height testing is implemented serverside right now is that we figure out what terrain triangle the player is standing on (from the players X and Y position), after finding the triangle we get the vertices of the triangle, and then we use interpolation to figure out the height of a certain point inside of that triangle. This can all be seen in GridMap::getHeightFromFloat.

This is a bit of a simplistic implementation of height testing and it does not match up with the client, which I will argue below. If they do not match up and the client thinks we're under the terrain then we will simply fall under the world. To account for this all cores I've seen add an offset to the height to prevent the player from falling through the ground, you can usually see this as an offset of +0.5f or +2.0f depending on the surrounding code. For years now, the standard way of "fixing" issues like Blink falling through the map has been to increase this offset, but this adds the drawback of the player falling a bit after blinking on flat ground.

Now lets look at how the client detects height, please take a look at this image: https://imgur.com/xvNT4Jv

In the chat you can see my equivalence to the .gps command, first it shows the position the client reported and then it shows the height the server calculated on that X and Y coordinate. After that comes some debug stuff, but it's not important for this issue. Please notice how the clientside test and the serverside test doesn't match up and the center of the players feet seem to hover over the ground, this happens because the client uses a more complicated collision test that uses a sphere-plane intersection rather than a triangle interpolation. According to the chat message, if this was the destination of a blink and the height offset (read: hacky fix) for this test is +0.5f we would fall through the ground.

So, to finally fix this I propose keeping the current GetHeight function that uses triangle interpolation, but also adding a GetAccurateHeight function that builds a plane from the triangle the player is standing on and does a sphere-plane intersection test on it. GetAccurateHeight would be used in most places where you want to teleport a player, seeing as a player falling through the ground is a pretty unacceptable bug compared to an NPC slightly clipping into the ground. Keep the current GetHeight function for hot paths like NPC height testing.

Now, the obvious response to all of this is "why don't you fix it?". I would love to fix this issue for AC, but I am not well versed in your code and I have way too many projects on my plate already, and on top of all of this I have quite a lot of extra tasks at work that will require overtime the next 1-2 months.

[...]

PR #21537, opened 2025-02-19
fix(Core/Spell): Prevent Blink from causing character to fall through ground

walkline:
Added GROUND_HEIGHT_TOLERANCE to handle cases where the player's Z position is slightly below the vmap ground at the start of Blink. Without this, the algorithm incorrectly detects a large fall height, leading to unintended fall-through issues.

To reproduce the issue:
1. `.go xyz 840.172485 -449.381317 -56.342438 349 0.601184`
2. Jump and press blink, in some cases it leads to the issue.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
3.
4.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -1431,7 +1431,16 @@ void Spell::SelectImplicitCasterDestTargets(SpellEffIndex effIndex, SpellImplici
                 float destx = pos.GetPositionX() + distance * cos(pos.GetOrientation());
                 float desty = pos.GetPositionY() + distance * sin(pos.GetOrientation());
 
-                float ground = map->GetHeight(phasemask, pos.GetPositionX(), pos.GetPositionY(), pos.GetPositionZ());
+                // Added GROUND_HEIGHT_TOLERANCE to account for cases where, during a jump,
+                // the Z position may be slightly below the vmap ground level.
+                // Without this tolerance, a ray trace might incorrectly attempt to find ground
+                // beneath the actual surface.
+                //
+                // Example:
+                //    actual vmap ground: -56.342392
+                //    Z position:         -56.347195
+                float searchGroundZPos = pos.GetPositionZ()+GROUND_HEIGHT_TOLERANCE;
+                float ground = map->GetHeight(phasemask, pos.GetPositionX(), pos.GetPositionY(), searchGroundZPos);
 
                 bool isCasterInWater = m_caster->IsInWater();
                 if (!m_caster->HasUnitMovementFlag(MOVEMENTFLAG_FALLING) || (pos.GetPositionZ() - ground < distance))


Pursche:
This "fix" is fundamentally flawed, you are just adding a band-aid by adding an arbitrary height offset when the actual problem is that the collision test is assuming that the player is a infinitely small point in space, rather than having a collider with an actual size.

This might "fix the problem" in this specific place, but sooner or later you will find another spot that you can blink through and you'll just need to keep increasing the offset.

I've tried getting emulators to understand and fix this properly for the past 6 years but nobody does, so people keep PRing this kind of shortsighted hack fix.

More information (with expertly drawn Paint drawings) here:
https://github.com/azerothcore/azerothcore-wotlk/discussions/3971#discussioncomment-200081

heyitsbench:
> with expertly drawn Paint drawings

Darn right they are.

walkline:
> This "fix" is fundamentally flawed, you are just adding a band-aid by adding an arbitrary height offset when the actual problem is that the collision test is assuming that the player is a infinitely small point in space, rather than having a collider with an actual size.
> 
> This might "fix the problem" in this specific place, but sooner or later you will find another spot that you can blink through and you'll just need to keep increasing the offset.
> 
> I've tried getting emulators to understand and fix this properly for the past 6 years but nobody does, so people keep PRing this kind of shortsighted hack fix.
> 
> More information (with expertly drawn Paint drawings) here: [#3971 (comment)](https://github.com/azerothcore/azerothcore-wotlk/discussions/3971#discussioncomment-200081)

I started looking into this issue after reading your topic about cylinder collision tests, hoping to reproduce the exact problem and possibly adapt or at least understand how to fit your proposal into the solution.

However, it turned out to be a different issue, or feel free to correct me if I'm wrong.

The problem occurs when trying to get the player's ground position before actual Blink. This fails due to strange behavior during a regular jump (even a standstill jump).

When a player performs a standstill jump, you would expect their Z position to be higher than when they are simply standing in the same spot, or at the very least, the same. However, for some reason, during both standstill and moving jumps, the Z position can be slightly lower than the standing Z position and even lower than the actual ground position.

Additionally, since mmaps are not valid in that location (I don't know the implementation details here), vmaps are used to determine the height. Unlike a 2D check, vmaps perform a 3D check. Because we pass a Z position that is below the actual surface, the result is -inf.

As a result, [this](https://github.com/azerothcore/azerothcore-wotlk/blob/a60331d10985181da97d580ad2d35000b933542d/src/server/game/Spells/Spell.cpp#L1446) check fails, and the core assumes the player is falling into a deep hole, skipping all the collision checks that would normally be performed in a "regular case".

kissingers:
I use before way hack fix, just add 0.5 so always no fall, if can use GROUND_HEIGHT_TOLERANCE fix that good.
                    bool col = VMAP::VMapFactory::createOrGetVMapMgr()->GetObjectHitPos(mapid, pos.GetPositionX(), pos.GetPositionY(), z + 0.5f, destx, desty, z + 0.5f, destx, desty, z, -0.5f);
                    // check dynamic collision
					bool dcol = m_caster->GetMap()->GetObjectHitPos(phasemask, pos.GetPositionX(), pos.GetPositionY(), z + 0.5f, destx, desty, z + 0.5f, destx, desty, z, -0.5f);

Kitzunu:
Is this still WIP?

walkline:
> Is this still WIP?

Well, there is a more elegant and general way to implement the fix, but it's not a priority for me. As for the suggested fix, it is still applicable and would work.

If you prefer the more elegant solution and want to reduce the PR list, I'm fine with closing this. Perhaps it can be addressed in a separate PR.

kissingers:
This simple fix is quite good. The problem of falling to ground is too frequently now. If there is a better solution, later can remove this part of the change.

sudlud (approved):

Pursche:
Once again a hack fix won out over actually addressing the issue. ¯\_(ツ)_/¯

kissingers:
> Once again a hack fix won out over actually addressing the issue. ¯\_(ツ)_/¯

I think the difference between AzerothCore and TC is that AzerothCore is more practical. If we can't fix it completely, a temporary hack fix is not a problem. Imagine how many players fall and die every day. The practicality is really low.

blinkysc:
I implemented this for singularity and I just now though of this PR, maybe something like this would be better for the blink issue? https://github.com/azerothcore/azerothcore-wotlk/pull/21928

walkline:
@blinkysc I would say it's the other way around. You can use the approach that Blink's implementation uses, with `Map::GetObjectHitPos` and height checks (though I'm not sure about the terrain where the spell is used).

As for the Blink issue, I haven't experienced any problems with detecting vertical collisions, so there's nothing to address on that front.

kissingers:
Not just the mage’s Blink, but also the warrior’s Charge and Intercept, the druid’s Feral Charge, and the rogue’s Killing Spree all have this issue.

PR #19451, opened 2024-07-20
refactor(Deps/MySQL): Drop MariaDB and MySQL 5.7/8.1 and add MySQL 8.…

Kitzunu:
…4 support

## Changes Proposed:

* Support MySQL LTS version 8.4

* Bump min Server and Client version to MySQL 8.0 (LTS)

* Drop support for MySQL 5.7 (EOL)

* Drop support for MySQL 8.1 (EOL)

* Drop support for MariaDB as it is using MySQL 5.7 functions

---

Before all the MariaDB and Linux fans try to gut me in the comments...

- AzerothCore cannot satisfy all users' preferred dependencies
- AzerothCore has always been MySQL support first. The reason behind previous support of MariaDB is that they have been closely compatible.
- MariaDB is built upon MySQL 5.7 functions which will inevitably become compatible with MySQL 8.4 as they will continue to deprecate 5.7 functionality.
- MariaDB may be shipped with most linux distros, but there is no issues in installing MySQL. Just do it.
- If MariaDB decides one day to adapt MySQL 8 functions, then we can support it again I guess :D

## Test done

Build
========== Build: 1 succeeded, 0 failed, 18 up-to-date, 1 skipped ==========

Using SSL version: OpenSSL 3.3.1 4 Jun 2024 (library: OpenSSL 3.3.1 4 Jun 2024)
Using Boost version: 1.85.0
Using CMake version: 3.27.4
Using MySQL version: 80401
Found MySQL Executable: C:/Program Files/MySQL/MySQL Server 8.4/bin/mysql.exe

Log into game

Used some commands

## Todo

- Merge https://github.com/azerothcore/wiki/pull/976

PR #21595, opened 2025-02-23
fix(Scripts/Ulduar) Fix Kologarn script

p-tkachuk:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:
1. On first enter to Ulduar instance boss Kologarn could be in IN_PROGRESS state, which will cause player to freeze.
2. Focused Eyebeam spell casted first ray from first eye on target and second ray from second eye to random place. Rays not always follow people.
3. If raid suicided by jumping in pit near Kologarn, on reset his body would be rotated to wrong position.

Issues were resolved by next changes:
1. IN_PROGRESS state was set due to cast of "SPELL_ARM_DEAD" spell, which caused damage to Boss. Now this spell only casted when fight is in progress.
2. Correct effects targets were set for the spell. Also bunny npc now instructed to follow player.
3. On reset event boss will return to its home position. (he is stationary, so only orientation will change, which is what we need)

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:
1. Create 2+ characters, level up them to 80 `.level 79`
2. `.tele ulduar` all of them
3. Create party -> raid with those people
4. Go to Kologarn and kill all other bosses on your way.
5. Begin fight with Kologar and see that described issues are fixed

- [ ] Focused Eyebeam spell probably should be cast in random period, rather then every 20 seconds.
- [ ]

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
@@ -709,9 +709,9 @@ class spell_kologarn_focused_eyebeam : public SpellScript
         targets.clear();
 
         // Select a random target
-        auto front = newTargets.begin();
-        advance(front, urand(0, newTargets.size() - 1));
-        targets.push_back(*front);
+        std::list<Unit*>::iterator head = newTargets.begin();
+        advance(head, urand(0, newTargets.size() - 1));
+        targets.push_back(*head);
     }
 
     void Register() override
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
@@ -478,12 +478,8 @@ public:
                 case EVENT_FOCUSED_EYEBEAM:
                 {
                     events.ScheduleEvent(EVENT_FOCUSED_EYEBEAM, 20s);
-
-                    if (Unit* eyebeamTarget = SelectTarget(SelectTargetMethod::MaxDistance, 0, 0, true))
-                    {
-                        me->CastSpell(eyebeamTarget, SPELL_FOCUSED_EYEBEAM_SUMMON, false);
-                        Talk(EMOTE_EYES);
-                    }
+                    me->CastSpell(me, SPELL_FOCUSED_EYEBEAM_SUMMON, false);
+                    Talk(EMOTE_EYES);
                     return;
                 }
                 case EVENT_RESTORE_ARM_LEFT:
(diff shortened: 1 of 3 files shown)


Nyeriah:
the target changes look like a hack

p-tkachuk:
> the target changes look like a hack, would be cool to have better insight on why its necessary

@Nyeriah if remove all "corrections" from spell - it would work too. But... during fight the skill would sometimes hit players in Kolagarn hand, which is incorrect. And also it would hit in melee zone, which agian, according to comments in wowhead is incorrect. It should hit most ranged player.

Nyeriah:
Then we need to filter the area object selection instead of hacking the spell attribute

p-tkachuk:
> Then we need to filter the area object selection instead of hacking the spell attribute

@Nyeriah not sure what you mean by that... I'm still a rookie here.
I took another look at spell 63342 (the one with the hack) and couldn't find any flag indicating that this spell should target the most distant players. Based on comments on Wowhead, it seems to select one player from a pool of the three most distant players. Without a hack (or a custom script for this spell?), achieving the intended behavior doesn't appear feasible.
If you can address this issue, that would be fantastic! In the meantime, I believe the current implementation is a solid workaround. It significantly improves the user experience during the boss fight. The lasers now consistently target a single player instead of unpredictably shooting from the back of a boss's head to a random location.

Nyeriah:
We can create a spell script and change the object area selection code to filter the targets it should not be hitting

PkllonG:
tested, It can be closed here https://github.com/azerothcore/azerothcore-wotlk/issues/21604

PkllonG:
Has anything changed?

p-tkachuk:
> Has anything changed?

yep. This line:
https://github.com/azerothcore/azerothcore-wotlk/pull/21595/files#diff-50a962a7b3d47029063bbd485ad4490ecc5b43692749603edfd84e95fa993e82R298

pretty important change... without it hands will be spawned if Kolagarn is dead.

PkllonG:
force pushed It's hard to observe the changes every time

p-tkachuk:
> force pushed It's hard to observe the changes every time

Old habit. Sorry)

p-tkachuk:
@Nyeriah I've removed the hack for spell, please take another look

Rastrian:
LGTM

Nyeriah:
You got the idea right, but the code could be improved a bit. I will check with you again later, unless someone else drops by in the meantime to review this.

src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
```diff
@@ -640,29 +647,77 @@ class boss_kologarn_eyebeam : public CreatureScript
...
+{
+    PrepareSpellScript(spell_kologarn_focused_eyebeam);
+
+    bool Load() override
+    {
+        if (!GetCaster()->IsCreature())
+            return false;
+
+        return true;
+    }
```
Nyeriah, line 684:
```suggestion
    bool Load() override
    {
        return GetCaster()->IsCreature();
    }
```

src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
```diff
@@ -640,29 +647,77 @@ class boss_kologarn_eyebeam : public CreatureScript
             }
         }
 
-        void UpdateAI(uint32 diff) override
+        void UpdateAI(uint32 /*diff*/) override
         {
             if (justSpawned)
             {
-                me->DespawnOrUnsummon(10000);
+                if (ObjectGuid summoner = me->GetSummonerGUID())
```
Nyeriah, line 654: You could use ´´´void IsSummonedBy()``` to handle this part of the code

src/server/scripts/Northrend/Ulduar/Ulduar/boss_kologarn.cpp
p-tkachuk: done

PkllonG:
Tested, but I didn't merge PR https://github.com/azerothcore/azerothcore-wotlk/pull/19451

Nyeriah (approved):

TheSCREWEDSoftware:
@p-tkachuk for future PRs if you refer the issue(s) in the "Closes" part of the form it will automatically the issue(s) mentioned.

Issue #21604, opened 2025-02-25
(Instance/Ulduar): Unable to enter instance


noisiver:
### Current Behaviour

Trying to enter Ulduar using a regular player account (non-GM) prints `Map 603 could not be created for player`. A GM can enter just fine.

### Expected Behaviour

You should be able to enter Ulduar.

### Steps to reproduce the problem

1. Create two new characters on accounts where at least one do not have GM privileges
2. Form a raid group
3. Enter Ulduar using the non-GM character (You can use `tele name [character-name] ulduarinstance` using a GM character or in the worldserver terminal

### AC rev. hash/commit

AC> AzerothCore rev. 9f99e89bc325 2025-02-24 15:49:54 -0300 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 2. Characters in world: 2.
Connection peak: 2.
Server uptime: 1 minute(s) 15 second(s)
Update time diff: 1ms. Last 500 diffs summary:
|- Mean: 1ms
|- Median: 1ms
|- Percentiles (95, 99, max): 2ms, 4ms, 4ms
Using SSL version: OpenSSL 3.0.13 30 Jan 2024 (library: OpenSSL 3.0.13 30 Jan 2024)
Using Boost version: 1.83.0
Using CMake version: 3.28.3
Using MySQL version: 80040
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.8.0-1018-oracle
Worldserver listening connections on port 29728
Realmlist (Realm Id: 4) configured in port 29728
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in /home/ubuntu/azerothcore/data/ptr/maps. Total size: 291014951 bytes
vmaps directory located in /home/ubuntu/azerothcore/data/ptr/vmaps. Total size: 658130721 bytes
mmaps directory located in /home/ubuntu/azerothcore/data/ptr/mmaps. Total size: 2192910844 bytes
Default DBC locale: enUS.
All available DBC locales: enUS zhCN zhTW esMX
Using World DB: ACDB 335.13-dev
Latest LoginDatabase update: 2025_02_16_01.sql
Latest CharacterDatabase update: 2025_02_16_00.sql
Latest WorldDatabase update: 2025_02_24_00.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
No modules are enabled

[...]

PR #21606, opened 2025-02-25
fix(Map/Grids): grids need to be loaded before Initialize the instance script

PkllonG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/21604

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -2266,6 +2266,9 @@ void InstanceMap::CreateInstanceScript(bool load, std::string data, uint32 compl
     if (!instance_data)
         return;
 
+    //grids need to be loaded before Initialize the instance script
+    OnCreateMap();
+
     // use mangos behavior if we are dealing with Eluna AI
     // initialize should then be called only if load is false
     if (!isOtherAI || !load)
--- a/src/server/game/Maps/MapInstanced.cpp
+++ b/src/server/game/Maps/MapInstanced.cpp
@@ -223,8 +223,6 @@ InstanceMap* MapInstanced::CreateInstance(uint32 InstanceId, InstanceSave* save,
                 map->GetInstanceScript()->SetTeamIdInInstance(leader->GetTeamId());
     }
 
-    map->OnCreateMap();
-
     if (!save) // this is for sure a dungeon (assert above), no need to check here
         sInstanceSaveMgr->AddInstanceSave(GetId(), InstanceId, difficulty);
 


PkllonG:
@noisiver please test

PkllonG:
@Takenbacon Please take a look at whether this is okay to do? The problem was solved after I did so.

Takenbacon:
> @Takenbacon Please take a look at whether this is okay to do? The problem was solved after I did so.

There is a necessary order, grids must be loaded after the instance script. This commit will break faction instances (ICC, etc) as it's not setting the instance faction before loading objects.

While #21604 is technically discovered by grid creation, it is absolutely not a grid issue. There is no case where that script in particular should only be working based off luck just because the object happens to be far enough away not to be initially loaded at instance enter. Fundamentally it's a script and or possibly a vehicle logic issue. I debugged it earlier and it's caused by the vehicle kit being Reset upon creation which ultimately causes the Kologarn PassengerBoarded script hook to cast a spell and makes the boss enter combat with its own vehicle accessories, causing instance in progress state.

noisiver:
I've enterered Ulduar 3 times (clearing the instance-related tables in the DB between tests) and it works. I don't think multiple tests would show anything but figured it doesn't hurt.

PkllonG:
> > @Takenbacon Please take a look at whether this is okay to do? The problem was solved after I did so.
> 
> There is a necessary order, grids must be loaded after the instance script. This commit will break faction instances (ICC, etc) as it's not setting the instance faction before loading objects.
> 
> While #21604 is technically discovered by grid creation, it is absolutely not a grid issue. There is no case where that script in particular should only be working based off luck just because the object happens to be far enough away not to be initially loaded at instance enter. Fundamentally it's a script and or possibly a vehicle logic issue. I debugged it earlier and it's caused by the vehicle kit being Reset upon creation which ultimately causes the Kologarn PassengerBoarded script hook to cast a spell and makes the boss enter combat with its own vehicle accessories, causing instance in progress state.

So, do you need this to close the PR?

p-tkachuk:
Fix for this issue is available here:
https://github.com/azerothcore/azerothcore-wotlk/pull/21595

What happens?
When Kologarn added to world, he ejects all of his passengers (because he is a vehicle). And on ejection spell is casted to make a damage to Kologarn, which starts a fight.
Proposed changes in linked PR fixes an issue by preventing a spell cast if he is not in fight yet.

Issue #15499, opened 2023-03-20
Shaman Lightning Overload is currently behaving as it did in 3.0-3.1 (proccing off itself) with the 3.2 proc rates


neifion-00000000:
### Current Behaviour

Lightning Overload can currently proc off itself infinitely.

### Expected Blizzlike Behaviour

https://www.youtube.com/watch?v=xbuuSdOMPUU is a pretty clear video of it behaving as it should

In 3.2, Lightning Overload was buffed significantly.  Alongside this buff came a removal of proc-chaining which was not mentioned but visible in any fights with clear view of the elemental shaman's damage (with the improved rates, if it were possible it would happen every fight at a rate of 1 in 9, you'd find evidence in every ele shaman video, yet I can find none where it occurs).

### Source

https://www.youtube.com/watch?v=xbuuSdOMPUU again is my best source, you can see procs occurring usually right before the player's next cast goes off on the right hand side (procs are 50% in strength of a normal bolt)

Additionally, the 3.2 chance increase is documented as a blue post (working already), and there is a wowhead comment mentioning the removal of chain procs.  https://www.wowhead.com/wotlk/spell=30679/lightning-overload#comments:id=779835:reply=423919

Additional additionally, this is how it functions on wotlk classic.

### Steps to reproduce the problem

While speced into Lightning Overload, Spam Lightning bolt or chain lightning on a target dummy or mobs until it happens.  It shouldn't take long.  Not only is a double proc 1 in 9, a triple proc is currently 1 in 27 (and continue multiplying by 3 for quad, quint, etc)

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/29fba5fe38e96667b67597f02b7dd58724650b64

### Operating system

Win10

### Custom changes or Modules

reported to me via CC, then confirmed

neifion-00000000:
notes/tips for that video if you plan on rewatching, its been sped up and if you're having trouble keeping up with the casts you can slow it down

neifion-00000000:
@Nyeriah this is still occurring unfortunately 

![image](https://github.com/azerothcore/azerothcore-wotlk/assets/91289495/e661afa6-a699-4785-9d1a-c28709b4d0c5)

neifion-00000000:
sorry I meant @UltraNix

PR #21625, opened 2025-02-28
fix (Core/Spells) remove custom loop so that one lightning damage instance can only proc one lightning overload

Tereneckla:
## Changes Proposed:

This PR proposes changes to:

Remove proc "propagating" of lightning overload

The triggered spells 45297 and 45284 don't every proc lightning overload, handled before already (tested with a few hundred chain lightnings and lightning bolts)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/15499

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

.learn 30679
.learn 49271
Spam that spell

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8412,37 +8412,17 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
                     if (!procSpell || !IsPlayer() || !victim)
                         return false;
 
+                    uint32 spell = procSpell->SpellFamilyFlags[0] & 0x2 ? 45297 : 45284;
+
                     if (procEx & PROC_EX_CRITICAL_HIT)
                         damage /= 2;
 
-                    // do not proc off from itself
-                    if (procSpell->Id == 45297 || procSpell->Id == 45284)
-                    {
-                        return false;
-                    }
+                    // do not reduce damage-spells have correct basepoints
+                    damage /= 2;
+                    int32 dmg = damage;
 
-                    do
-                    {
-                        uint32 spell = 0;
-
-                        if (procSpell->SpellFamilyFlags[0] & 0x2)
-                        {
-                            // 1/3 of 33% if 11%
-                            if (!roll_chance_i(33))
-                                return false;
-
-                            spell = 45297;
-                        }
-                        else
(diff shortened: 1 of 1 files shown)


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -8412,37 +8412,17 @@ bool Unit::HandleDummyAuraProc(Unit* victim, uint32 damage, AuraEffect* triggere
...
+                    uint32 spell = procSpell->SpellFamilyFlags[0] & 0x2 ? 45297 : 45284;
+
                     if (procEx & PROC_EX_CRITICAL_HIT)
                         damage /= 2;
 
-                    // do not proc off from itself
-                    if (procSpell->Id == 45297 || procSpell->Id == 45284)
-                    {
-                        return false;
-                    }
```
Nyeriah, line 8422: I think this part shouldnt be removed

Tereneckla: I casted both Chain Lighting and Lightning Bolt a few hundred times with prints of the procSpell-Ids, those 2 never showed up

wetbrownsauce: The procSpell never seems to be 45297 or 45284, so this check can be removed currently. Maybe something in spell_proc_event prevents this already?  

That check was added in response to an old bug report that lightning overload was procc'ing off of itself, but I don't think it ever was, I think it was always that do-while loop triggering additional hits and the original reporter thinking it was procc'ing off of itself. see: https://github.com/azerothcore/azerothcore-wotlk/issues/15499

However, both with and without this PR, chain lightning seems to be able to trigger lightning overload off of its chain jumps. Against 3 enemies, I sometimes see 6 hits (lightning overload proc), but I'm also sometimes seeing 9 hits (lightning overload procs an extra time?). Unsure if that is blizzlike or not.

Tereneckla: https://youtu.be/5hLcz1-9L4U?t=65

Chain lightning with 3 hits (proof that it is not glyph), followed by 8 hits

Nyeriah (approved):

FrancescoBorzi:
thanks for this contribution and nice to see you guys around @Tereneckla @wetbrownsauce

PR #21676, opened 2025-03-09
feat(Core/Entities): Add OnPlayerSendListInventory script hook

VincentVanclef:
## Changes Proposed:

Simply adds a new script hook
-  Would like to know when a player interacts with a vendor. Going to need this for a custom auction module.


sudlud:
please update with latest master and resolve the conflicts

src/server/game/Handlers/ItemHandler.cpp
```diff
@@ -1036,6 +1036,8 @@ void WorldSession::SendListInventory(ObjectGuid vendorGuid, uint32 vendorEntry)
 {
     LOG_DEBUG("network", "WORLD: Sent SMSG_LIST_INVENTORY");
 
+    sScriptMgr->OnPlayerSendInventoryList(GetPlayer(), vendorGuid, vendorEntry);
```
sudlud, line 1039: `OnPlayerSendListInventory()` (same as function name it is beeing called in)

is there no other hook available that could accomplish your desired functionality

VincentVanclef: I checked, and sadly not.

The reason is my module is displaying a custom vendor list based on the players inventory and then im adding a callback method that triggers on the next item bought from OnPlayerBeforeBuyItemFromVendor. So if the player interacts with any normal vendor, i need to clear this callback.

So this was the only way

src/server/game/Scripting/ScriptDefines/PlayerScript.cpp
```diff
@@ -920,6 +920,11 @@ bool ScriptMgr::OnPlayerCanGiveLevel(Player* player, uint8 newLevel)
     CALL_ENABLED_BOOLEAN_HOOKS(PlayerScript, PLAYERHOOK_ON_CAN_GIVE_LEVEL, !script->OnPlayerCanGiveLevel(player, newLevel));
 }
 
+void ScriptMgr::OnPlayerSendListInventory(Player* player, ObjectGuid vendorGuid, uint32 vendorEntry)
```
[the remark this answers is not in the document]
VincentVanclef: good idea, thats now implemented

PR #21698, opened 2025-03-13
feature/on-name-query-hooks

VincentVanclef:
## Changes Proposed:

Simply adds a new script hook
-  This is very useful for adding custom info to players name, e.g. Paragon Level or if using a Guild Level module then display the guilds level in the guild name

### DIFF
--- a/src/server/game/Handlers/QueryHandler.cpp
+++ b/src/server/game/Handlers/QueryHandler.cpp
@@ -24,6 +24,7 @@
 #include "Opcodes.h"
 #include "Pet.h"
 #include "Player.h"
+#include "ScriptMgr.h"
 #include "World.h"
 #include "WorldPacket.h"
 #include "WorldSession.h"
@@ -43,8 +44,11 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
 
     Player* player = ObjectAccessor::FindConnectedPlayer(guid);
 
+    std::string name = playerData->Name;
+    sScriptMgr->OnPlayerSendNameQueryOpcode(player, name);
+
     data << uint8(0);                               // name known
-    data << playerData->Name;                       // played name
+    data << name;                                   // player name
     data << uint8(0);                               // realm name - only set for cross realm interaction (such as Battlegrounds)
     data << uint8(player ? player->getRace() : playerData->Race);
     data << uint8(playerData->Sex);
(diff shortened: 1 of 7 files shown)


sudlud:
idk i'm unsure about this, we can't just slab hooks into packet handlers everywhere.

i would assume that there's another way to achive a similiar result maybe? so you wand to manipulate the displayed guild and player names?

also do you have any work-in-progress of module(s) that you require these hooks for?

VincentVanclef:
> idk i'm unsure about this, we can't just slab hooks into packet handlers everywhere.
> 
> i would assume that there's another way to achieve a similiar result maybe? so you wand to manipulate the displayed guild and player names?
> 
> also do you have any work-in-progress of module(s) that you require these hooks for?

Yes both paragon and guild level module fully working:

image


the : P 6 in my char name is my characters Paragon Level

the : Lv 0 in my guild name is my guilds level (currently 0)

i have just added the scripthooks in my own fork for now, but would love them added to AZ to prevent merge conflicts and then i can release the modules to the public after.

sudlud:
Interesting, is your module publicly available? I'd like to take a look

VincentVanclef:
> Interesting, is your module publicly available? I'd like to take a look

https://github.com/VincentVanclef/mod-paragon-levels

Just turned this public, its modified a bit to work with my fork so won't work off the bat but can release a new one more in line with this:

https://github.com/VincentVanclef/mod-paragon-levels/blob/5a8632a2f1d1e8d3e70ee5872c5ba63983438133/src/paragon_levels.cpp

Which has isolated SQL so it can work as a plug-n-play module.

sudlud:
i could imagine that you can accomplish the same result for custom player name by using `CaracterCache::UpdateCharacterData()` to change the cached character name, e.g. at login.

VincentVanclef:
> i could imagine that you can accomplish the same result for custom player name by using `CaracterCache::UpdateCharacterData()` to change the cached character name, e.g. at login.

I would argue that is a very hacky approach and will mess up whatever else uses the CharacterCacheData.

src/server/game/Handlers/QueryHandler.cpp
```diff
@@ -43,8 +44,11 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
 
     Player* player = ObjectAccessor::FindConnectedPlayer(guid);
 
+    std::string name = playerData->Name;
```
sudlud, line 47: Unfortunately I will not accept a string copy just for the sake of having a new hook.

sudlud:
Thanks again for sending a PR but I'll close this for the time beeing. Feel free to address the latest comments / concerns and reopen with an adjusted approach.

PR #21707, opened 2025-03-14
feature/OnPlayerGetXpForLevel-hook

VincentVanclef:
## Changes Proposed:

Simply adds a new script hook
-  For cases like Paragon leveling or implementing true individual xp rates this is required. Would be very nice to have this hook.

### DIFF
--- a/src/server/game/Globals/ObjectMgr.cpp
+++ b/src/server/game/Globals/ObjectMgr.cpp
@@ -7508,6 +7508,14 @@ uint32 ObjectMgr::GetXPForLevel(uint8 level) const
     return 0;
 }
 
+uint32 ObjectMgr::GetXPForLevel(Player* player) const
+{
+    uint8 level = player->GetLevel();
+    uint32 xpForNextLevel = GetXPForLevel(level);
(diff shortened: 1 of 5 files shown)


sudlud:
I'm a bit lost on this one.

why should we add a new function to the core, which is called nowhere in the core, just so that you can place a new hook into it, which will also never be called? or am i missing something here.

you could probably introduce new functions directly in your module in some `.h` / `.cpp` file instead

sudlud:
thanks again for sending a pr - i must say that unfortunately I am still not a fan of this hook implementation.

you copy the function `GetXPForLevel()` but now with a `player` parameter, so the function name is probably not too appropriate anymore.
also there's at least 2 other occasions in the core where `GetXPForLevel()` is beeing called - so it's not a generic hook for that function as the hook name suggests.
as of know i would suggest to keep this a custom change in your fork as I unfortunately do no see it beeing merged into the core in its current state.

TheSCREWEDSoftware:
Status of this PR, will any changes will be made from the review or this is to be closed?

TheSCREWEDSoftware:
Be free to re-open this in the future

PR #19858, opened 2024-09-03
feat(Core/Arena): Add support for arena seasons completion with progression in runtime.

walkline:
## Changes Proposed:
With this change, we can manage arena seasons in the game without needing to restart the server. The main feature is the ability to complete an arena season by distributing rewards to players and deleting arena teams. Additionally, you can progress through seasons without a server restart.

This update introduces 4 new commands:
```
.arena season start $season_id
.arena season reward 2,3 
.arena season reward all
.arena season reward 2
.arena season deleteteams
.arena season set state $state
```

Some arena season related data (such as the current season and season state) is now stored in the database, so the corresponding configuration options have been removed from the config file.

Example usage to end and start a new season by rewarding only the 2v2 bracket:
```
.arena season set state 0
// The arena will be disabled, and we need to wait until all ongoing rated arenas are finished.
// Wait...
.arena season reward 2
// This will delete all teams, including 3v3 and 5v5.
.arena season deleteteams
// Place the correct arena vendors and start the new season.
.arena season start 8
```

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:

- Closes 

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [ ] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

[...]

PR #21960, opened 2025-04-19
fix(DB): Add and correct arena season rewards.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

## Tests Performed:

This PR:
- [X] Has not been tested.

- [ ] Seasons 1 and 2 don't have the same style of mail message as the future seasons, not sure how the drakes were provided if not by mail, but don't really see a good mail template to use for them.

### DIFF
--- a/data/sql/updates/pending_db_world/arena-rewards.sql
+++ b/data/sql/updates/pending_db_world/arena-rewards.sql
@@ -14,12 +14,12 @@ INSERT INTO `arena_season_reward_group` (`id`, `arena_season`, `criteria_type`, 
 (30, 3, 'pct', 10, 35, 0, '', '', 0),
 -- Season 2 (Merciless)
 (31, 2, 'abs', 1, 1, 0, '', '', 0),
-(32, 2, 'pct', 0, 0.5, 182, '', '', 0), -- Mail needs verification
+(32, 2, 'pct', 0, 0.5, 0, 'Congratulations!', 'On behalf of the Steamwheedle Fighting Circuit, we congratulate you for your successes in this arena season.$B$BIn recognition of your skill and savagery, we hereby bestow upon you this Merciless Nether Drake.  May it serve you well.', 0),
 (33, 2, 'pct', 0.5, 3, 0, '', '', 0),
 (34, 2, 'pct', 3, 10, 0, '', '', 0),
 (35, 2, 'pct', 10, 35, 0, '', '', 0),
 -- Season 1
-(36, 1, 'pct', 0, 0.5, 182, '', '', 0), -- Mail needs verification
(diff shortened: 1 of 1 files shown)


data/sql/updates/pending_db_world/arena-rewards.sql
```diff
@@ -0,0 +1,72 @@
...
+DELETE FROM `arena_season_reward` WHERE `group_id` IN (1, 6, 11, 16, 21, 22, 22, 23, 24, 25, 26, 27, 27, 28, 29, 30, 31, 32, 32, 33, 34, 35, 36, 36, 37, 38, 39);
+INSERT INTO `arena_season_reward` (`group_id`, `type`, `entry`) VALUES
+-- Season 8 (Wrathful)
+(1, 'achievement', 4599),
+-- Season 7 (Relentless)
+(6, 'achievement', 3758),
+-- Season 6 (Furious)
+(11, 'achievement', 3436),
+-- Season 5 (Deadly)
+(16, 'achievement', 3336),
```
heyitsbench, line 36: CC: @walkline, was all four of these seasons using the [Deadly Gladiator achievement](https://github.com/azerothcore/azerothcore-wotlk/pull/19858/files#diff-353c1135e51c7fb21c471acdeb028d0debce5652a038c9a9c587ba74bbc12c4eR55-R76) an oversight or am I missing something?

walkline: Most likely, I didn't have the data for the older seasons, so I just used the same values.

data/sql/updates/pending_db_world/arena-rewards.sql
```diff
@@ -0,0 +1,72 @@
...
+(25, 4, 'pct', 10, 35, 0, '', '', 0),
+-- Season 3 (Vengeful)
+(26, 3, 'abs', 1, 1, 0, '', '', 0),
+(27, 3, 'pct', 0, 0.5, 211, '', '', 0),
+(28, 3, 'pct', 0.5, 3, 0, '', '', 0),
+(29, 3, 'pct', 3, 10, 0, '', '', 0),
+(30, 3, 'pct', 10, 35, 0, '', '', 0),
+-- Season 2 (Merciless)
+(31, 2, 'abs', 1, 1, 0, '', '', 0),
+(32, 2, 'pct', 0, 0.5, 182, '', '', 0), -- Mail needs verification
```
walkline, line 17: Mail template 182 doesn't seem right to me, not that I care much. But as an alternative, we could use `reward_mail_subject` and `reward_mail_body` instead of the template.

walkline (approved):

data/sql/updates/pending_db_world/arena-rewards.sql
heyitsbench: That's a good call, at least it would show up differently in the DB to distinguish it from the likely good mail template usages.

Issue #21781, opened 2025-03-26
[Sunwell Plateau] Conflagration and Shadow Nova wrong targeting


amed80:
### Current Behaviour

In Phase 1 : 
- [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166)'s [Conflagration](https://wowgaming.altervista.org/aowow/?spell=45342) currently target the 2nd person on its own threat table.

- [Lady Sacrolash](https://wowgaming.altervista.org/aowow/?npc=25165)'s [Shadow Nova](https://wowgaming.altervista.org/aowow/?spell=45329) currently target the 2nd person on its own threat table.

In Phase 2:
- [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166)'s [Shadow Nova](https://wowgaming.altervista.org/aowow/?spell=45329) currently target the 2nd person on its own threat table.

- [Lady Sacrolash](https://wowgaming.altervista.org/aowow/?npc=25165)'s [Conflagration](https://wowgaming.altervista.org/aowow/?spell=45342) currently target the 2nd person on its own threat table.

### Expected Behaviour

In Phase 1:
- [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166)'s [Conflagration](https://wowgaming.altervista.org/aowow/?spell=45342) should target randomly one of the top 6 on [Lady Sacrolash](https://wowgaming.altervista.org/aowow/?npc=25165) threat table apart from the current [Lady Sacrolash](https://wowgaming.altervista.org/aowow/?npc=25165)'s target.

- [Lady Sacrolash](https://wowgaming.altervista.org/aowow/?npc=25165)'s [Shadow Nova](https://wowgaming.altervista.org/aowow/?spell=45329) should target randomly one of the top 6 on [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166) threat table apart from the current [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166)'s target.

In Phase 2:

- [Grand Warlock Alythess](https://wowgaming.altervista.org/aowow/?npc=25166)'s [Shadow Nova](https://wowgaming.altervista.org/aowow/?spell=45329) should target a random person on its own threat table apart from the current target.

[...]

PR #21964, opened 2025-04-20
fix(Scripts/SunwellPlateau): Make Twin special abilities target one of the highest 6 in threat of the other twin

Tereneckla:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/21781

## SOURCE:

## Tests Performed:

This PR has been:

Tested with the number set to 2 and 4 characters

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
+++ b/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
@@ -157,7 +157,7 @@ struct boss_sacrolash : public BossAI
             if (Creature* alythess = instance->GetCreature(DATA_ALYTHESS))
             {
                 std::list<Unit*> targets;
-                alythess->AI()->SelectTargetList(targets, 6, SelectTargetMethod::MaxThreat, 1, 100.0f);
+                alythess->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 1, 100.0f);
                 target = Acore::Containers::SelectRandomContainerElement(targets);
             }
             if (!target)
@@ -290,7 +290,7 @@ struct boss_alythess : public BossAI
             if (Creature* sacrolash = instance->GetCreature(DATA_SACROLASH))
             {
                 std::list<Unit*> targets;
-                sacrolash->AI()->SelectTargetList(targets, 6, SelectTargetMethod::MaxThreat, 1, 100.0f);
+                sacrolash->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 1, 100.0f);
                 target = Acore::Containers::SelectRandomContainerElement(targets);
             }
             if (!target)
--- a/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
+++ b/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
@@ -157,7 +157,7 @@ struct boss_sacrolash : public BossAI
             if (Creature* alythess = instance->GetCreature(DATA_ALYTHESS))
             {
                 std::list<Unit*> targets;
-                alythess->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 1, 100.0f);
+                alythess->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
                 target = Acore::Containers::SelectRandomContainerElement(targets);
             }
             if (!target)
@@ -290,7 +290,7 @@ struct boss_alythess : public BossAI
             if (Creature* sacrolash = instance->GetCreature(DATA_SACROLASH))
             {
                 std::list<Unit*> targets;
-                sacrolash->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 1, 100.0f);
+                sacrolash->AI()->SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
                 target = Acore::Containers::SelectRandomContainerElement(targets);
             }
             if (!target)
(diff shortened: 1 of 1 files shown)


Tereneckla:
My alternative solution would be introducing a new <num> parameter to SelectTarget like it already is present in SelectTargetList

That is then used instead of then provided to SelectTargetList instead of GetThreatListSize if it is >0 and forces SelectRandomContainerElement instead of first element

```cpp
    template <class PREDICATE>
    Unit* SelectTarget(SelectTargetMethod targetType, uint32 position, uint32 num, PREDICATE const& predicate)
    {
        ThreatMgr& mgr = GetThreatMgr();

        // shortcut: if we ignore the first <offset> elements, and there are at most <offset> elements, then we ignore ALL elements
        if (mgr.GetThreatListSize() <= position)
            return nullptr;

        uint32 targetNum = num > 0 ? num : mgr.GetThreatListSize();
        std::list<Unit*> targetList;
        SelectTargetList(targetList, targetNum, targetType, position, predicate);

        // maybe nothing fulfills the predicate
        if (targetList.empty())
            return nullptr;

        if(num > 0){
            targetType = SelectTargetMethod::Random;
        }

        switch (targetType)
        {
        case SelectTargetMethod::MaxThreat:
        case SelectTargetMethod::MinThreat:
        case SelectTargetMethod::MaxDistance:
        case SelectTargetMethod::MinDistance:
            return targetList.front();
        case SelectTargetMethod::Random:
            return Acore::Containers::SelectRandomContainerElement(targetList);
        default:
            return nullptr;
        }
    }
```

Nyeriah:
I have no issues with this approach, although I preferred having it added to the target options (in the enum). MaNGOS also has it as an enum option (as far as I could see). But if we cant think of any other bosses that should behave like this, then this ad hoc solution is completely fine

Tereneckla:
> I have no issues with this approach, although I preferred having it added to the target options (in the enum). MaNGOS also has it as an enum option (as far as I could see). But if we cant think of any other bosses that should behave like this, then this ad hoc solution is completely fine

Just checked cmangos, their enum are 
https://github.com/cmangos/mangos-classic/blob/fd03e05fbc2dbea7cf75ea6bcaefab41d748eea9/src/game/Entities/Unit.h#L1018C1-L1026C3

My problem with an enum is that we then have 4 options which are 100% useless in SelectTargetList (and also still need the parameter to actually specify the range of random).

Their solution has a roll before knowing the threatlist size, which will run into issues when people are doing the raid with lower group sizes
https://github.com/cmangos/mangos-wotlk/blob/98a5437bdb3826a267b1dae93dc8cf90c9ded83c/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/sunwell_plateau/boss_eredar_twins.cpp#L442

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
```diff
@@ -119,7 +119,10 @@ struct boss_sacrolash : public BossAI
 
             scheduler.CancelGroup(GROUP_SPECIAL_ABILITY);
             ScheduleTimedEvent(20s, [&] {
-                Unit* target = SelectTarget(SelectTargetMethod::MaxThreat, 1, 100.0f);
+                Unit* target = nullptr;
+                std::list<Unit*> targets;
+                SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
+                target = Acore::Containers::SelectRandomContainerElement(targets);
```
Nyeriah, line 125:
```suggestion
                std::list<Unit*> targets;
                SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
                Unit* target = Acore::Containers::SelectRandomContainerElement(targets);
```

Tereneckla: shouldn't abuse copy-paste without thinking

Rorschach91 (approved):

Rorschach91:
@Nyeriah if you think that the code don't have issues, merge it.

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
```diff
@@ -119,7 +119,9 @@ struct boss_sacrolash : public BossAI
 
             scheduler.CancelGroup(GROUP_SPECIAL_ABILITY);
             ScheduleTimedEvent(20s, [&] {
-                Unit* target = SelectTarget(SelectTargetMethod::MaxThreat, 1, 100.0f);
+                std::list<Unit*> targets;
+                SelectTargetList(targets, 5, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
```
sogladev, line 123:
```suggestion
                SelectTargetList(targets, 6, SelectTargetMethod::MaxThreat, 0, 100.0f, true, false);
```

The issue reads to me as 6. @amed80 could you confirm whether this is 5 or 6?
> Grand Warlock Alythess's Conflagration should target randomly one of the top 6 on Lady Sacrolash threat table apart from the current Lady Sacrolash's target.

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
```diff
@@ -119,7 +119,9 @@ struct boss_sacrolash : public BossAI
...
+                Unit* target = Acore::Containers::SelectRandomContainerElement(targets);
```
sogladev, line 124: Can `targets` be empty here? If so, test if targets is empty before proceeding


```suggestion
                Unit* target;
                if (!targets.empty())
                    target = Acore::Containers::SelectRandomContainerElement(targets);
```

https://github.com/Tereneckla/azerothcore-wotlk/blob/twin-target/src/common/Utilities/Containers.h#L130
> container cannot be empty

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_eredar_twins.cpp
Tereneckla: In the examples the lowest in threat that was selected was Top6. So we need to select 5 targets because we exclude the tank

sogladev (approved):

PR #21998, opened 2025-04-27
Linux: Support systemd socket activation for the worldserver game socket

ovv:
Using systemd socket activation allows the worldserver to start only when a connection is made to the socket. Coupled with a script to auto-shutdown the server when no characters are connected this would ensure the worldserver is only running when players are connected.

A new dependency is added on linux system `libsystemd-dev`

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

N/A

## SOURCE:

N/A

## Tests Performed:

This PR has been:

This PR should be tested on Windows to ensure it doesn't have an impact there as systemd and socket activation is not available.

## How to Test the Changes:

1. Build server
2. Ensure server start and bind to the expected socket

- [ ]
- [ ]


src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -72,27 +90,31 @@ class AsyncAcceptor
...
 
-#if AC_PLATFORM != AC_PLATFORM_WINDOWS
-        _acceptor.set_option(boost::asio::ip::tcp::acceptor::reuse_address(true), errorCode);
-        if (errorCode)
-        {
-            LOG_INFO("network", "Failed to set reuse_address option on acceptor {}", errorCode.message());
-            return false;
-        }
-#endif
+            #if AC_PLATFORM != AC_PLATFORM_WINDOWS
```
sudlud, line 103: Keep these aligned to the front please

walkline:
Thanks for the contribution!
I’m wondering if we can get rid of libsystemd-dev dependency.
As far as I understand all we need to do is to read these two environment variables - LISTEN_PID and LISTEN_FDS.

Can we use something like this?
```
#if defined(__linux__)
#include <cstdlib>
#include <unistd.h>

int get_listen_fds()
{
    const char* listen_pid = std::getenv("LISTEN_PID");
    const char* listen_fds = std::getenv("LISTEN_FDS");
    if (!listen_pid || !listen_fds)
        return 0;

    pid_t pid = static_cast<pid_t>(std::stoi(listen_pid));
    if (pid != getpid())
        return 0;

    // Unset after reading (systemd convention for security)
    unsetenv("LISTEN_PID");
    unsetenv("LISTEN_FDS");

    return std::stoi(listen_fds);
}
#else
// On non-Linux systems, just return 0 (no socket activation)
int get_listen_fds()
{
    return 0;
}
#endif
```

ovv:
> I’m wondering if we can get rid of libsystemd-dev dependency. As far as I understand all we need to do is to read these two environment variables - LISTEN_PID and LISTEN_FDS.
> 
> Can we use something like this?

Yes that's the gist of it. In my opinion relying on libsystemd-dev is easier

1. `libsystemd-dev` library is optional. If you don't care about this feature you don't need it. See the docker CI that builds without it.
2. Servers that care about this functionality already have systemd installed. Installing the header is a small addition
3. If we want to introduce socket activation for the telnet and soap API we probably want to use `sd_listen_fds_with_names`. That's slightly harder to implement ourselves.

So I'd rather we keep it unless there is a clear goal of keeping dependencies to the strict minimum

walkline:
From my point of view, fewer dependencies are better. If we can replace this dependency with 10 lines of code or so, I would prefer that approach.

Also adding this dependency would require updates to the documentation. It might confuse users if they decide to enable this feature later and discover they need to install an additional dependency first and recompile the project.

That said, this is just my perspective — I'd be glad to hear what others think.

benoitheylens:
Hello, I was looking to  do something similar, how fun :) !
Question though: will it work with Playerbots (or any mod simulating players) ?

ovv:
> Question though: will it work with Playerbots (or any mod simulating players) ?

I don't use Playerbots, so I can't confirm, but it shouldn't be an issue as long as it doesn't slow down the server startup too much.

sudlud:
Can you please try to avoid force pushing on PRs, it makes tracking changes after reviews quite hard

sudlud:
I'd also agree to avoid additional dependencies if possible

ovv:
This is ready for a second round of review.

I removed the dependency on libsystemd, and it's no longer marking the server as offline when shutting down and socket activation is used. I can make it a setting if that's preferred.

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -33,10 +34,22 @@ class AsyncAcceptor
...
     typedef void(*AcceptCallback)(tcp::socket&& newSocket, uint32 threadIndex);
 
-    AsyncAcceptor(Acore::Asio::IoContext& ioContext, std::string const& bindIp, uint16 port) :
+    AsyncAcceptor(Acore::Asio::IoContext& ioContext, std::string const& bindIp, uint16 port, bool supportSocketActivation = false) :
         _acceptor(ioContext), _endpoint(Acore::Net::make_address(bindIp), port),
-        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); })
+        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); }),
+        _supportSocketActivation(supportSocketActivation)
     {
+        int listen_fd = get_listen_fds();
```
sudlud: Can probably be int const?

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -33,10 +34,22 @@ class AsyncAcceptor
...
+        if (_supportSocketActivation && listen_fd > 0)
+        {
+            LOG_DEBUG("network", "Using socket from systemd socket activation");
+            boost::system::error_code errorCode;
+            _acceptor.assign(boost::asio::ip::tcp::v4(), listen_fd, errorCode);
+            if (errorCode)
+            {
+                LOG_WARN("network", "Failed to assign socket {}", errorCode.message());
+            }
```
sudlud: Remove braces

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -33,10 +34,22 @@ class AsyncAcceptor
...
+        int const listen_fd = get_listen_fds();
+        if (_supportSocketActivation && listen_fd > 0)
+        {
+            LOG_DEBUG("network", "Using socket from systemd socket activation");
+            boost::system::error_code errorCode;
+            _acceptor.assign(boost::asio::ip::tcp::v4(), listen_fd, errorCode);
+            if (errorCode)
+            {
+                LOG_WARN("network", "Failed to assign socket", errorCode.message());
+            }
```
sudlud: I think you misunderstood, we're not doing braces on single line after an if-statement (ac code style). The fmt braces should still be required here

ovv: Indeed, my bad. I'm mixing a bunch of languages at the moment and getting confused.

src/common/Utilities/Systemd.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <string>
+
+int get_listen_fds()
+{
+    const char* listen_pid = std::getenv("LISTEN_PID");
+    const char* listen_fds = std::getenv("LISTEN_FDS");
+    if (!listen_pid || !listen_fds)
+        return 0;
+
+    pid_t pid = static_cast<pid_t>(std::stoi(listen_pid));
```
Winfidonarleyan: https://github.com/azerothcore/azerothcore-wotlk/blob/92094eec01f3618762997bd954114119f102349f/src/common/Utilities/StringConvert.h#L244

sudlud:
please. no force pushes, it makes tracking reviews and following changes impossible.

src/common/Utilities/Systemd.cpp
```diff
@@ -0,0 +1,46 @@
...
+#if defined(__linux__)
+#include "StringConvert.h"
+#include <cstdlib>
+#include <unistd.h>
+#include <string>
+
+int get_listen_fds()
+{
+    const char* listen_pid = std::getenv("LISTEN_PID");
+    const char* listen_fds = std::getenv("LISTEN_FDS");
```
sudlud:
```suggestion
    char* const listen_pid = std::getenv("LISTEN_PID");
    char* const listen_fds = std::getenv("LISTEN_FDS");
```

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -124,6 +139,7 @@ class AsyncAcceptor
     tcp::socket _socket;
     std::atomic<bool> _closed;
     std::function<std::pair<tcp::socket*, uint32>()> _socketFactory;
+    bool _supportSocketActivation = false;
```
sudlud:
```suggestion
    bool _supportSocketActivation;
```

sudlud: already beeing initialized in the constructor, right?

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -33,10 +34,20 @@ class AsyncAcceptor
...
+    AsyncAcceptor(Acore::Asio::IoContext& ioContext, std::string const& bindIp, uint16 port, bool supportSocketActivation = false) :
         _acceptor(ioContext), _endpoint(Acore::Net::make_address(bindIp), port),
-        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); })
+        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); }),
+        _supportSocketActivation(supportSocketActivation)
     {
+        int const listen_fd = get_listen_fds();
+        if (_supportSocketActivation && listen_fd > 0)
+        {
+            LOG_DEBUG("network", "Using socket from systemd socket activation");
```
sudlud, line 45: would it make sense to also print the value of `listen_fd` to the debug output? or is that useless information

sudlud:
please update the pr description for the latest version of the pr (e.g. no newly added dependency)

i would also propose to make this configurable in worldserver.conf so it is disabled by default.

sudlud:
Also could you please add some test steps on how to test this on Linux

sudlud:
hey, what's the state of this PR :) can you please address the review comments

ovv:
> hey, what's the state of this PR :) can you please address the review comments

I still want to get it merge. I'll try to get to it in the coming days

TheSCREWEDSoftware:
@ovv when this is ready to be reviewed, ping me so i can set the tag for it

sudlud:
I will close this for the time being, please feel free to reopen and address the review comments so we can get it merged in the future :)

src/server/shared/Network/AsyncAcceptor.h
ovv: I think its useless information. If you only use one socket it should always be 3 as far as I know

ovv:
How to test this on linux:

1. Set `Network.UseSocketActivation = 1` in the worldserver.conf
2. Configure azerothcore as a systemd service with socket activation

`/etc/systemd/system/worldserver.service`
```
[Unit]
Description=AzerothCore Worldserver
After=network.target mysql.service
Requires=worldserver.socket

[Service]
Type=exec
Restart=always
RestartSec=60
User=azerothcore
WorkingDirectory=/opt/azerothcore
ExecStart=/usr/bin/stdbuf -o0 /opt/azerothcore/dist/bin/worldserver --config /opt/azerothcore/dist/etc/worldserver.conf
SyslogIdentifier=worldserver

ProtectSystem=strict
PrivateTmp=true
ProtectHome=yes
ProtectClock=yes
ProtectKernelLogs=yes
ProtectKernelModules=yes
UMask=0077
LockPersonality=yes
RestrictRealtime=yes
MemoryDenyWriteExecute=yes

[Install]
WantedBy=multi-user.target
``` 

`/etc/systemd/system/worldserver.socket`
```
[Unit]
Description=AzerothCore Worldserver Socket

[Socket]
ListenStream=0.0.0.0:8086

[Install]
WantedBy=sockets.target
```

Enable both and start the socket
```
$ systemctl enable worldserver.service
$ systemctl enable worldserver.socket
$ systemctl start worldserver.socket
```

3. Connecting to the port 8086 should automatically start the worldserver and pass the opened socket

src/server/shared/Network/SocketMgr.h
```diff
@@ -42,7 +43,8 @@ class SocketMgr
         std::unique_ptr<AsyncAcceptor> acceptor;
         try
         {
-            acceptor = std::make_unique<AsyncAcceptor>(ioContext, bindIp, port);
+            bool supportSocketActivation = sConfigMgr->GetOption<bool>("Network.UseSocketActivation", false, true);
```
sudlud: Does the default value here match the default in the config?

src/server/apps/worldserver/Main.cpp
```diff
@@ -406,7 +407,8 @@ int main(int argc, char** argv)
     sScriptMgr->OnShutdown();
 
     // set server offline
-    LoginDatabase.DirectExecute("UPDATE realmlist SET flag = flag | {} WHERE id = '{}'", REALM_FLAG_OFFLINE, realm.Id.Realm);
+    if (get_listen_fds() == 0)
+        LoginDatabase.DirectExecute("UPDATE realmlist SET flag = flag | {} WHERE id = '{}'", REALM_FLAG_OFFLINE, realm.Id.Realm);
```
sudlud, line 411: Will a realm always show online if this option is enabled?

ovv: Yes. I need to update this so it uses the `Network.UseSocketActivation` option

src/server/shared/Network/SocketMgr.h
ovv: Yes, it default to disable

src/server/shared/Network/AsyncAcceptor.h
```diff
@@ -33,10 +34,20 @@ class AsyncAcceptor
...
-        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); })
+        _socket(ioContext), _closed(false), _socketFactory([this](){ return DefaultSocketFactory(); }),
+        _supportSocketActivation(supportSocketActivation)
     {
+        int const listen_fd = get_listen_fd();
+        if (_supportSocketActivation && listen_fd > 0)
+        {
+            LOG_DEBUG("network", "Using socket from systemd socket activation");
+            boost::system::error_code errorCode;
+            _acceptor.assign(boost::asio::ip::tcp::v4(), listen_fd, errorCode);
```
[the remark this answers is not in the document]
ovv: @sudlud I did this but I'm not sure if that's fine with the rest of azerothcore. Are IPv6 fine ?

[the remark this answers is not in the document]
sudlud: Azerothcore is ipv4 only as far as I can tell. IPv6 was not a thing in 2010 for wotlk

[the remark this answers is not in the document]
ovv: Do you want me to change that back ?

I guess it would fail later rather than at the socket creation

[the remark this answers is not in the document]
sudlud: Im not sure of its relevant, but we probably don't need any ipv6 handling in the core right now

maybe-dragon:
Tested on Linux with `Network.UseSocketActivation = 1`.
Works great.
But I removed the `WantedBy=multi-user.target` from the worldserver.service, so that it does not start on its own...

sudlud:
Can you please update the file headers to GPLv2 (#23655) then this should be ready for merge imo

ovv:
> Can you please update the file headers to GPLv2 (#23655) then this should be ready for merge imo

done

sudlud:
So I consulted ChatGPT in regards of handling the systemd licensing correctly, this is the updated version:

```cpp
/*
 * This file is part of the AzerothCore Project. See AUTHORS file for Copyright information.
 *
 * Portions of this file are derived from systemd, licensed under:
 * - GPL-2.0-or-later (if the original systemd file was GPL-2.0+)
 *   OR
 * - LGPL-2.1-or-later, relicensed under GPL-2.0-or-later as permitted by LGPL-2.1+
 *
 * Original systemd copyright:
 *   Copyright (c) the systemd contributors.
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
 * more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program. If not, see <http://www.gnu.org/licenses/>.
 */
```

ovv:
> So I consulted ChatGPT in regards of handling the systemd licensing correctly, this is the updated version:
> 
> ```c++
> /*
>  * This file is part of the AzerothCore Project. See AUTHORS file for Copyright information.
>  *
>  * Portions of this file are derived from systemd, licensed under:
>  * - GPL-2.0-or-later (if the original systemd file was GPL-2.0+)
>  *   OR
>  * - LGPL-2.1-or-later, relicensed under GPL-2.0-or-later as permitted by LGPL-2.1+
>  *
>  * Original systemd copyright:
>  *   Copyright (c) the systemd contributors.
>  *
>  * This program is free software; you can redistribute it and/or modify
>  * it under the terms of the GNU General Public License as published by
>  * the Free Software Foundation; either version 2 of the License, or
>  * (at your option) any later version.
>  *
>  * This program is distributed in the hope that it will be useful, but WITHOUT
>  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
>  * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
>  * more details.
>  *
>  * You should have received a copy of the GNU General Public License along
>  * with this program. If not, see <http://www.gnu.org/licenses/>.
>  */
> ```

I'm not sure this is necessary since we are not using any systemd library directly. I'll put (or feel free to do it yourselves if it's easier) whatever works for the project.

sudlud:
To my understanding that function was pulled and maybe adapted from the systemd sources, right? so I'd like to have the licensing as suggested :)

ovv:
@sudlud done

sudlud (approved):

FrancescoBorzi:
@ovv thank you very much for your contribution!

sudlud:
@ovv this probably also needs to be added to `authserver.conf.dist`, right?

PR #22029, opened 2025-04-30
fix(Scripts): Use distinct sound entry for BRD L70ETC Music Doodad.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

## Tests Performed:

This PR:
- [X] Has not been tested.


src/server/scripts/World/go_scripts.cpp
```diff
@@ -457,14 +457,20 @@ class go_bear_trap : public GameObjectScript
...
 };
 
 enum L70ETCMusicEvents
 {
     EVENT_ETC_START_MUSIC = 1
 };
 
+enum L70ETCMusicAreas
+{
+    BLACKROCK_DEPTHS = 1584
```
sudlud, line 471: What about AREA_

heyitsbench: ur an area_

heyitsbench: I didn't add that prefix since it wasn't present in the Brewfest enums further down in the file, consistency and all that.

sudlud: Makes me wonder why the brewfest stuff is still in the world file

sudlud: In this case I also don't really like the brewfest enum down below, having enum values like ORGRIMMAR just is too generic imo. I'd prefer a prefix

sudlud: sounds like a great plan and also like something bench would prepare a pr for

PkllonG:
Tested

src/server/scripts/World/go_scripts.cpp
heyitsbench: Corrected.

TheSCREWEDSoftware (approved):

Issue #22030, opened 2025-04-30
Outdated AUTHORS file


jackpoz:
### Current Behaviour

The AUTHORS file at https://github.com/azerothcore/azerothcore-wotlk/blob/master/AUTHORS lists TrinityCore as 2008-2012 at https://github.com/azerothcore/azerothcore-wotlk/blob/8797e8ad280aabc8314bc175c3a6284aa2b6deaf/AUTHORS#L17

This is incorrect, as code from TrinityCore after 2012 has been merged into AzerothCore, the most recent one being https://github.com/azerothcore/azerothcore-wotlk/commit/38ff8bae2803da83d0df7a1fadf7edf050ff9f6a which is a cherry-pick from https://github.com/TrinityCore/TrinityCore/commit/15598723810f76a56b1ea5f3198b482dc0c19d65 from 2024.

Please update the AUTHORS file with 2024 as latest year for TrinityCore, or 2025 is any more recent commit was included.

There is no template that fits this kind of issues, so forgive me if not all fields are filled in correctly.

### Logs and Screenshots

No logs

### Operating System

N/A

### Dependencies & versions

N/A

### Commit

8797e8ad280aabc8314bc175c3a6284aa2b6deaf

PR #22031, opened 2025-04-30
chore(Docs): Update TrinityCore year

jackpoz:
Update TrinityCore year in AUTHORS with the most recent commit included in this project

## Changes Proposed:

This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).
-  [x] None of the above

## Issues Addressed:

- Closes #22030

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [ ] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #22034, opened 2025-05-01
docs(AUTHORS): add Cross-project Collaboration section

FrancescoBorzi:
- Outcome of https://github.com/azerothcore/azerothcore-wotlk/pull/22031
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22030

Co-authored-by: jackpoz <giacomopoz@gmail.com>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,22 @@ The third-party libraries have their own way of addressing authorship, and the a
 a third-party library reflects who did the importing instead of who wrote the code within the commit.
 
 The Authors of third-party libraries are not explicitly mentioned, and usually is possible to obtain from the files belonging to the third-party libraries.
+
+## Cross-project collaboration
+
+At AzerothCore, we actively promote collaboration with other MaNGOS-based open-source projects that are actively maintained. This also includes cherry-picking commits from relevant projects such as:
+
+* TrinityCore: https://github.com/trinitycore
+* CMaNGOS: https://github.com/cmangos/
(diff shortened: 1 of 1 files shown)


Rorschach91 (approved):

Helias (approved):

TheSCREWEDSoftware (approved):

heyitsbench:
Given that cherrypicks are likely to continue, maybe would be good to drop the year from TC (and other projects if added) and replace it with `CURRENT` as AC is currently listed from the section discussed in the linked PR.

Nyeriah (approved):

Francesco-Borzi:
> Given that cherrypicks are likely to continue, maybe would be good to drop the year from TC (and other projects if added) and replace it with `CURRENT` as AC is currently listed from the section discussed in the linked PR in addition to the changes proposed in this PR, that way it avoids the need of updating the year(s) annually when a commit is cherrypicked.

I don't have a strong opinion about this specifically, but the way I see it, that section is more about the software BEFORE the AzerothCore project has been created, where we unfortunately don't have the git history. And it's still meaningful to show that it was based on SWC which was based on a 2012 version of TC.

The new section that I'm adding in this PR shows the currently active projects where we mostly cherry-pick stuff from, so IMHO I would not mix it up and leave the other sections the way it is now.

IntelligentQuantum (approved):

sudlud (approved):

PR #22104, opened 2025-05-11
refactor(Core): Use database tables for Shaman totems and Druid shapeshift forms.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Play around with Shaman/Druids and their totems/shapeshifts to try and break something.


data/sql/updates/pending_db_world/get-model-for-player.sql
```diff
@@ -0,0 +1,173 @@
...
+(29, 6, 255, 2, 20872);
+
+DELETE FROM `player_totem_model`;
+INSERT INTO `player_totem_model` (`TotemID`, `RaceID`, `ModelID`) VALUES
+-- Orc
+(1, 2, 30758), -- Fire
+(2, 2, 30757), -- Earth
+(3, 2, 30759), -- Water
+(4, 2, 30756), -- Air
+-- Dwarf
```
PkllonG, line 154: Dwarf, Only here cannot be tested, there are no dwarf shamans.

PkllonG:
Tested

data/sql/updates/pending_db_world/get-model-for-player.sql
heyitsbench: Yea, I'm not too sure why they were defined in the original function, considering removing them from the DB data, but they're not doing any harm there, I suppose. 😛

PkllonG: Leave it there, it might be useful.

data/sql/updates/pending_db_world/get-model-for-player.sql
```diff
@@ -0,0 +1,173 @@
...
+DROP TABLE IF EXISTS `player_totem_model`;
+
+CREATE TABLE IF NOT EXISTS `player_shapeshift_model` (
+  `ShapeshiftID` TINYINT unsigned NOT NULL,
+  `RaceID` TINYINT unsigned NOT NULL,
+  `CustomizationID` TINYINT unsigned NOT NULL,
+  `GenderID` TINYINT unsigned NOT NULL,
+  `ModelID` INT unsigned NOT NULL,
+  PRIMARY KEY (`ShapeshiftID`, `RaceID`, `CustomizationID`, `GenderID`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 PACK_KEYS=0;
```
PkllonG, line 11: In addition, there should be someone requesting the use of 'InnoDB' here.

data/sql/updates/pending_db_world/get-model-for-player.sql
```diff
@@ -0,0 +1,173 @@
...
+
+CREATE TABLE IF NOT EXISTS `player_totem_model` (
+  `TotemID` TINYINT unsigned NOT NULL,
+  `RaceID` TINYINT unsigned NOT NULL,
+  `ModelID` INT unsigned NOT NULL,
+  PRIMARY KEY (`TotemID`, `RaceID`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 PACK_KEYS=0;
```
PkllonG, line 18: And here too.

data/sql/updates/pending_db_world/get-model-for-player.sql
heyitsbench: You're probably right, I just used MyISAM here since it's what I used last time I added a table, and had no opposition.

heyitsbench: Corrected.

data/sql/updates/pending_db_world/get-model-for-player.sql
heyitsbench: Corrected.

data/sql/updates/pending_db_world/get-model-for-player.sql
heyitsbench: May be worth adding that to the SQL codestyle check at some point too, just for consistency.

heyitsbench:
> So theortical anyone could customise the model used by druid forms and the shaman totems?

That's the idea, yes.

TheSCREWEDSoftware:
Pretty cool addition

src/server/game/Globals/ObjectMgr.cpp
```diff
@@ -1785,6 +1785,139 @@ void ObjectMgr::LoadCreatureModelInfo()
...
+
+    itr = _playerShapeshiftModel.find(std::make_tuple(form, player->getRace(), 255, player->getGender()));
+    if (itr != _playerShapeshiftModel.end())
+        return itr->second; // Default gender-dependent model
+
+    itr = _playerShapeshiftModel.find(std::make_tuple(form, player->getRace(), 255, GENDER_NONE));
+    if (itr != _playerShapeshiftModel.end())
+        return itr->second; // Last resort
+
+    LOG_DEBUG("misc", "ShapeshiftForm {} with RaceID ({}) have no shapeshift model data defined, using fallback data.", form, player->getRace());
```
Kitzunu, line 1917: should probably use log type `player.something` or just leave it `player`

Kitzunu:
I like it. In the future, we could make a handler for player stuff so we can shrink the usage of ObjectMgr

heyitsbench:
> [...] so we can shrink the usage of ObjectMgr

Me likey.

PkllonG:
Tested

TheSCREWEDSoftware:
Removing wiki tag so it doesnt duplicate a via the bot another issue.

FrancescoBorzi:
very good idea @heyitsbench , we should move all kind of static data like this to the DB

TheSCREWEDSoftware:
Disregard @PkllonG

TheSCREWEDSoftware (approved):

RAORelewise:
Since we already going this route, why not complete it and also allow setting a ClassID? Or even better, make both RaceMask and ClassMask. 

ClassMask could just be 0 by default = all classes.

Would make it perfectly customizeable at will.

VincentVanclef:
Since we already going this route, why not complete it and also allow setting a ClassID? Or even better, make both RaceMask and ClassMask.

ClassMask could just be 0 by default = all classes.

Would make it perfectly customizeable at will.

VincentVanclef:
> > Since we already going this route, why not complete it and also allow setting a ClassID? Or even better, make both RaceMask and ClassMask.
> 
> RaceMask I could see, but ClassMask is...not something I would do. IMO if someone is going that far down the modding route, they should be able to adjust the table/usage in the core themselves to achieve what they're going for. Feels a step too far, again, just in my opinion.

This feature is already opening up for customization, its too late to draw a line, it's already been crossed, now lets finish the feature. Heck even adding a Level restraint would be cool. Can have much cooler cat form at level 80 than at level 20 etc. 😄

TheSCREWEDSoftware:
While the RaceID eventually could be expanded to RaceMask it wouldn't be a bad change (even if i prefer row per entry).

This change wasn't made to support custom races, because this could be already achieved in the core, just moves this hardcored values and functionality into the database.

RaceMask shouldn't be an addition, only druids shapeshift and only shaman have totems, being the base of the functionality anyone that decides they want this to work (shapeshift and totems) to other classes e.g a warrior can have specific totem, it's amtter of looking at this pr and make the addition in their core with the changes they need

PR #22108, opened 2025-05-11
feat(Core): make BGs and Arena prep time configurable

NoxMax:
I hate waiting for too long. Since the first time I played BGs many years ago I wondered if we actually need 2 whole minutes to prep. So I made it configurable.

## Changes Proposed:
PR makes BGs and Arena prep time configurable in worldserver.conf. Keeping the default values of 120s/60s for BGs and Arena respectively.
Making the timing configurable was simple, but making the timing announcements accurate is what required most of the work. Previous code logic had announcements triggered based on event flags rather than actual time values (full/half/quarter/none) The new code makes _ProcessJoin in Battleground.cpp actually aware how much time is left till match begins.

This PR proposes changes to:

## SOURCE:
N/A Default behavior has not been changed.

## Tests Performed:
Compiled correctly on Debian.
This PR has been:

## How to Test the Changes:

1. At the default values of 120s/60s for BGs and Arena, there should be no observed change in game.
2. Change the prep time of BGs and Arena to something other than 120s/60s and restart the server. The prep time should be whatever you changed the value to.
3. The announcement messages should be accurate. If it says 30 seconds till start, match must start in 30 seconds.
4. BGs have announcements at 120/60/30/0 seconds. Arena has announcements at 60/30/15/0 seconds. If you set a prep time that is less than one of the announcement times, you should not see an announcement for a time that is greater than the configured perp time. Example: If you set the prep time to 45 seconds for BGs, you should only get the 30/0 second announcements, and at their correct time.

Please advise after testing if anything has been missed.

### DIFF
--- a/src/server/game/Spells/SpellInfoCorrections.cpp
+++ b/src/server/game/Spells/SpellInfoCorrections.cpp
@@ -4897,8 +4897,10 @@ void SpellMgr::LoadSpellInfoCorrections()
         spellInfo->Effects[EFFECT_0].RadiusEntry = sSpellRadiusStore.LookupEntry(EFFECT_RADIUS_70_YARDS);
     });
 
-    // Encapsulate
-    ApplySpellFix({ 45662 }, [](SpellInfo* spellInfo)
+    ApplySpellFix({
+        45662, // Encapsulate
+        45642  // Fire Bloom
+        }, [](SpellInfo* spellInfo)
     {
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
very nice idea, thanks for your submission @NoxMax !

FrancescoBorzi:
@NoxMax what happened to this PR?

heyitsbench:
> what happened to this PR?

"Apologies for reposting the PR. I accidentally closed the original -_-"

https://github.com/azerothcore/azerothcore-wotlk/pull/22136

Issue #22117, opened 2025-05-12
[Tempest Keep] Al'ar bug at T1 if tank died


kissingers:
### Current Behaviour

When only 1 tank at top side vi Al'ar, if tank died, the boss will lost target and will not cast  spell Flame Strike (34121), and also will not cast any other spell, below side player can attack till boss die.

https://www.youtube.com/shorts/tfxxuKyfcJ8

### Expected Behaviour

If no any body at meele attack range, the boss will cast Flame Strike (34121) till next Action or stage.

### Steps to reproduce the problem

1. at least 2 players 
2.  .go c 158041
3.  top side player attack boss and died
4. look the boss status

### AC rev. hash/commit

AzerothCore rev. 8b4b37fdbbb0 2025-05-12 00:27:24 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
服务器运行时间: 7 minute(s) 45 second(s)
Update time diff: 100ms. Last 500 diffs summary:
|- Mean: 100ms
|- Median: 100ms
|- Percentiles (95, 99, max): 101ms, 101ms, 101ms
Using SSL version: OpenSSL 3.0.15 3 Sep 2024 (library: OpenSSL 3.0.15 3 Sep 2024)
Using Boost version: 1.74.0
Using CMake version: 3.25.1
Using MySQL version: 80405
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 
Worldserver listening connections on port 54347
Realmlist (Realm Id: 3) configured in port 54347
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in /root/wow3/bin/maps. Total size: 291014339 bytes
vmaps directory located in /root/wow3/bin/vmaps. Total size: 658130553 bytes
mmaps directory located in /root/wow3/bin/mmaps. Total size: 2192911264 bytes
Default DBC locale: zhCN.
All available DBC locales: zhCN 
Using World DB: ACDB 335.13-dev
Latest LoginDatabase update: 2025_02_16_01.sql
Latest CharacterDatabase update: charactersdb_anticheat.sql
Latest WorldDatabase update: npc_beastmaster.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-anticheat
|- mod-bg-reward
[...]

PR #22118, opened 2025-05-12
fix(Scripts/Tempest Keep/Eye) Al'ar don't attack other players if only one tank died

kissingers:
Because SPELL_FLAME_BUFFET is a range spell, target any player will take effect for full instance, so change DoCastVictim to DoCastRandomTarget can always cast the spell.

Also if scheduler.Update(diff) run after if (!UpdateVictim()) , if no victim will not update diff, so boss stop do anything, move to before will always try time diff first. so the DoCastVictim can work property.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

https://github.com/azerothcore/azerothcore-wotlk/issues/22117
- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Nyeriah:
It should cast it on its target if the target is away, not on a random player, so this is wrong

TheSCREWEDSoftware:
Shouldn't the change be if there's no near current target then it should cast flame buffet on a random target.

kissingers:
> Shouldn't the change be if there's no near current target then it should cast flame buffet on a random target.

Before fix, the boss will do nothing, ignore all below side players.

TheSCREWEDSoftware:
Even if this PRs fixes that, it will complety work as is not supposed

kissingers:
> Even if this PRs fixes that, it will complety work as is not supposed

No. with this PR, the boss work as supposed. 

another fix way is the boss can target the one of the live players.  but might chang much more than this fix way and break the boss workflow.

Nyeriah:
You are right, it's an AOE spell, therefore it has no target. Should be changed to DoCastAOE() instead in that case

kissingers:
> You are right, it's an AOE spell, therefore it has no target. Should be changed to DoCastAOE() instead in that case

I try use DoCastAOE() test. if DoCastAOE() work fine. is more good than DoCastRandomTarget.

kissingers:
> You are right, it's an AOE spell, therefore it has no target. Should be changed to DoCastAOE() instead in that case

It work fine, already update.

Nyeriah (approved):

PR #22127, opened 2025-05-15
fix(Core/SharedDefines): Define creature static flags.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

  - https: //github.com/cmangos/mangos-wotlk/commit/1fae6e3e6e70fa615226b223e7cb148473ebfab0

## Tests Performed:

This PR:
- [X] Has not been tested.

- [ ] This PR does not aim to implement any of these flags, nor create any helper functions to interact with them, only define them. I did not elect to mark them as not implemented since...all of them are unimplemented. If/when any are implemented, then the rest should probably be marked as unimplemented.

### DIFF
--- a/src/server/shared/SharedDefines.h
+++ b/src/server/shared/SharedDefines.h
@@ -2726,6 +2726,245 @@ enum CreatureTypeFlags
     CREATURE_TYPE_FLAG_QUEST_BOSS                        = 0x80000000    // Not verified
 };
 
+enum class CreatureStaticFlags : uint32
+{
+    MOUNTABLE                = 0x00000001,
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
What does this exactly do, is this flags can be used by creatures that are static?

And from the commit this is taken from, doesn't this require a DB change to support those changes in the creature_template?

Seamthesis:
Mangos seems to only use max 4 CreatureStaticFlags. Are you sure that the client have 8 flag groups for Woltk ?

Issue #5116, opened 2021-04-03
Creatures leash system not blizzlike


T1ti:
## EXPECTED BLIZZLIKE BEHAVIOUR:

1. There's a small radius around the aggro point where creature will never leave combat
2. Outside of that radius, the leash is timer based(around ~12sec _(seems to be based on creature level and actually ranges from 11 to 16sec)_ . When that timer expires, the creature resets, as simple as that.
3. several actions will reset that leash timer : 
  -the creature standing still/attacking for 5seconds(not in chase state)
  -the creature being hit by an hostile action from the player : either an attack, or a debuff(even non damaging)
Those actions will just reset the leash timer and the creature will chase for ~12seconds again

## CURRENT BEHAVIOUR:
1. Kind of working, creatures will never reset n a radius around the creature, but see point 2
2. There doesn't seem to be a timer, it looks purely distance based, creatures seem to reset only when they reach a distance from the original aggro location. There is an enormous radius around the original creature location where it will never reset, and the creature just resets when you step outside that radius
3. Player attacks are extending the leash outside the radius a bit currently, by about 10sec. The creature attacking does not extend the leash at all, if you bring the creature out of range and just let it attack you, it will reset after a bit(should not happen)
**-Debuffs don't extend the leash, it has been verified that they do on both Classic AND retail** (debuffs pretty much count as an attack)

## SOURCE:
Based on testing on Classic servers by me and the vmangos project.
![image](https://user-images.githubusercontent.com/40864460/113485279-017aa500-94ad-11eb-884f-cee5957cd466.png)
Made a drawing to show a bit how it works, the sizes are guessed
edit : new source : blizard's official creature tool from 2009, I believe the "timeout" is actually the leash timer, it's set to 0 for instance creatures (unlimited)
![image](https://user-images.githubusercontent.com/40864460/116008968-3fd03380-a617-11eb-99f2-68d87ea81347.png)
I described those settigns here : https://github.com/azerothcore/azerothcore-wotlk/issues/5455 

## STEPS TO REPRODUCE THE PROBLEM:

1. just test the points

## AC HASH/COMMIT:
4b7454574412cfe23bb5902b9b3d42d2313b3541

T1ti:
Note : the current AC system feels extremly obnoxious in cavesa nd tunnels with turns thata ren't a straight line, i had creatures chasing me for nearly 40 seconds in the jasperlode mine without hitting them because I was theorycally not existing the radius

[...]

PR #22129, opened 2025-05-15
fix(Core/Creature): Leashing improvements.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/5116

## SOURCE:

  - https://github.com/cmangos/mangos-wotlk/commit/57fd85d43c0addbf6b1fcf13d8979fe45140a3d3
  - https://github.com/cmangos/mangos-wotlk/commit/0690df3310df84c5ae95670c1e0956308accf583

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] So many problems to fix.


heyitsbench:
This PR specifically targets the safe distance check (i.e. the distance from where the mob is pulled, it should never leash) and the leash timer for improvement.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2671,29 +2671,28 @@ bool Creature::CanCreatureAttack(Unit const* victim, bool skipDistCheck) const
...
+    GetMotionMaster()->GetMotionSlot(MOTION_SLOT_IDLE)->GetResetPosition(x, y, z);
+
+    if (!dist)
+        return true;
 
     if (Unit* unit = GetCharmerOrOwner())
     {
         dist = std::min<float>(GetMap()->GetVisibilityRange() + GetObjectSize() * 2, 150.0f);
-        return victim->IsWithinDist(unit, dist);
+        return IsWithinDist(unit, dist);
```
PkllonG, line 2692:
```
    if (isWorldBoss() || GetCharmerOrOwner())
    {
        dist = std::min<float>(GetMap()->GetVisibilityRange() + GetObjectSize() * 2, 150.0f);
        return IsWithinDist(victim, dist);
    }
```

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2671,29 +2671,28 @@ bool Creature::CanCreatureAttack(Unit const* victim, bool skipDistCheck) const
...
 
-    // xinef: added size factor for huge npcs
-    float dist = std::min<float>(GetDetectionRange() + GetObjectSize() * 2, 150.0f);
+    float dist = sWorld->getFloatConfig(CONFIG_CREATURE_LEASH_RADIUS);
+
+    float x, y, z;
+    GetMotionMaster()->GetMotionSlot(MOTION_SLOT_IDLE)->GetResetPosition(x, y, z);
+
+    if (!dist)
+        return true;
```
PkllonG, line 2687:
```
    if (!dist)
        return true;
```
I don't think it should exist here

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2671,29 +2671,28 @@ bool Creature::CanCreatureAttack(Unit const* victim, bool skipDistCheck) const
...
 
     if (Unit* unit = GetCharmerOrOwner())
     {
         dist = std::min<float>(GetMap()->GetVisibilityRange() + GetObjectSize() * 2, 150.0f);
-        return victim->IsWithinDist(unit, dist);
+        return IsWithinDist(unit, dist);
     }
     else
```
PkllonG, line 2694:
```
-    else
+    if (!dist)
+        return true;
```
It's more appropriate here

heyitsbench: I'm inclined to agree, corrected.

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2671,29 +2671,28 @@ bool Creature::CanCreatureAttack(Unit const* victim, bool skipDistCheck) const
...
-        return victim->IsWithinDist(unit, dist);
-    }
-    else
-    {
-        // to prevent creatures in air ignore attacks because distance is already too high...
-        if (GetMovementTemplate().IsFlightAllowed())
-            return victim->IsInDist2d(&m_homePosition, dist);
-        else
-            return victim->IsInDist(&m_homePosition, dist);
+        return IsWithinDist(unit, dist);
```
PkllonG, line 2689: There's no possibility that a creature and its owner will attack each other, right?
So it should be, the creatures and the owners of the creatures, together attack their targets :
```
return unit->IsWithinDist(victim, dist);
```

PkllonG:
Whether this can improve performance?
image

src/server/game/Entities/Creature/Creature.cpp
heyitsbench: I don't particularly agree with this one. The two instances I can think of where this code could come up is with Mind Control targets and pets/summons.
I'm not sure how/if this function is used elsewhere for the case of charms (haven't looked yet), but iirc on Classic, creatures you Mind Control can't get very far from either where they're pulled or from the charming player before breaking the charm, don't recall any other thing affecting them in a way dissimilar to a player as far as leashing goes.
As for pets, player pets can be attacking a creature and despawn when the owner gets out of range. The pet can be commanded to attack a target on the edge of visibility range, but will return to the player if the enemy leaves the player's visibility range before the pet enters combat with it.
Creature pets on the other hand are an interesting case. If the owner of the pet leaves visibility range while the pet is still engaged with the player (say a caster owner and a melee pet), the pet will stay engaged with the player while the owner returns to its pull position, and the pet will be in its own evade mode. (Can be seen in [this timestamped video](https://www.youtube.com/watch?v=xnrUkbEbiJk&t=573s).)

That all being said, maybe
```cpp
return IsWithinDist(victim, dist);
```
would suit this case? Not too sure.

PkllonG: Well, we can try your way first

PkllonG:
Tested

kissingers:
This PR case a new bug, when test DPS with the NPC fack man, always out of combit if only use melle attack.
Can test this way
.go c 48323  or  other fack man npc with world boss flags, try attack.

PR #22130, opened 2025-05-16
Kil'Jaeden Ability Timers

LuckyLuc96:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

https://github.com/chromiecraft/chromiecraft/issues/8037
- The issue mentioned was the boss not casting his cycle of spells for phase 1 every 20 seconds, but every 40 seconds. The bug appears to be simply the timer having been too lenient in the script. This small change does appear to resolve the issue.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. This fight looked normal to me, but I am not familiar with it. Please also observe his behavior.
2.
3.

- [ ]
- [ ]


Nyeriah:
If this proves correct, the ternary IF no longer serves a purpose in that statement

LuckyLuc96:
I think I may have underestimated the bug and it's possible the issue is the phase is not being properly read. I will keep testing and come back. So my change may be causing this cycle of ability to be the same throughout the fight, instead of changing on the last phase.

LuckyLuc96:
> If this proves correct, the ternary IF no longer serves a purpose in that statement

I agree. I'll change that and test the fight another time to make sure everything looks okay.

LuckyLuc96:
Testing the boss fight more and everything seems to be working well. He doesn't cast Fire Bloom so much toward the end of the fight, because he is busy with a channeled spell. But it does get cast consistently around every 20 seconds.

Apologies for the 5 commits for the small change.

kissingers:
Yes, the current skills of Kil'jaeden are quite weak. He is the easiest boss in the Sunwell because the timing intervals of some channeled abilities have not been taken into account.

amed80:
First [Fire Bloom](https://wowgaming.altervista.org/aowow/?spell=45641) is happening too early in P3 and P4, it's ~37s after the phase start , it should be ~48s at least (sometimes even +60s)
In P5, first [Fire Bloom](https://wowgaming.altervista.org/aowow/?spell=45641) happens after ~50s , when in videos i never saw it happen before the first darkness cast (+70s)

If [Fire Bloom](https://wowgaming.altervista.org/aowow/?spell=45641) gets delayed the next one will be casted too early such as 14s , the shortest timer between 2 [Fire Bloom](https://wowgaming.altervista.org/aowow/?spell=45641) i have seen in videos is 17s.

Videos for reference: [Video1](https://youtu.be/tdjp6WUc14U) , [Video2](https://youtu.be/XCUGF8P1S8I) , [Video3](https://youtu.be/JbkS3uN0Zlc) , [Video4](https://youtu.be/Kl23FCk1uO8) , [Video5](https://youtu.be/uwBw08p01Xw) , [Video6](https://youtu.be/tbOlOAuEnGU) , [Video7](https://youtu.be/zY3KVfovgaA)

LuckyLuc96:
I am going off my DBM addon to see how the fight is going along with expectations. Phases 3-5 seem right on target. Phase 2, it appears I may be pushing his shield ability to not trigger when expected. 

Phase 1 seems normal. DBM still shows the flame ability as if it were a 40 second interval, but the original bug report provided evidence it should be 20 seconds.

amed80:
i think , you are changing the repeat timer with latest changes instead of the initial timer, it's why Fire Bloom "doesn't happen" in p5 as if it skips the first initial timer , it will have to wait 70s to repeat it?

LuckyLuc96:
> i think , you are changing the repeat timer with latest changes instead of the initial timer, it's why Fire Bloom "doesn't happen" in p5 as if it skips the first initial timer , it will have to wait 70s to repeat it?
> 
> Also for the timers , keep in mind there is always a delay by ~28s at the start of P3,P4 and P5, (caused by Shadow Spike)

I meant that oddly enough, my fight doesn't seem to ever get to phase 5. At least according to the deadly boss mods announcements.

LuckyLuc96:
In [this video](https://youtu.be/tdjp6WUc14U) from retail's TBC classic, this player is hit with Fire Bloom during phase 3 at 3:37 and you can see it cast again at 4:00, and is a 23 second delay

In [this timestamp](https://youtu.be/tdjp6WUc14U?t=405) from same player is hit with Fire Bloom at 6:45 mark. Kil'jaeden casts firebloom around 7:08 after that. This is in phase 4 of 5, and is a 23 second delay

Phase 5, for this specific group at least, I can only see firebloom being cast a single time. I have the same effect during my testing with my alteration of setting a "cooldown" of 70 seconds with my edit of the Fire Bloom timer. 



I tried to edit my timers to match this 23 second delay, but it resulted in him casting the fire bloom every 18ish seconds - which is too early. I'll try 25 to 27 and see how it works. 

Setting the timers to 20 (phase 2), 25 (phases 3/4), and 40 (phase 5) gets me a pretty consistent cast of Fire Bloom every 20 seconds for phase 1, every 20 seconds for phase 3 and 4, and every 50ish seconds for phase 5. I am not sure why the timer seems to have a negative affect for phase 3/4, but I think the way it works out ingame is pretty good.

src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kiljaeden.cpp
```diff
@@ -441,10 +441,27 @@ struct boss_kiljaeden : public BossAI
             DoCastRandomTarget(SPELL_LEGION_LIGHTNING, 0, 40.0f);
         }, _phase == PHASE_SACRIFICE ? 15s : 30s);
 
+        /* Phase names for reference
+        PHASE_DECEIVERS                 = 1,
+        PHASE_NORMAL                    = 2,
+        PHASE_DARKNESS                  = 3,
+        PHASE_ARMAGEDDON                = 4,
+        PHASE_SACRIFICE                 = 5,
+        */
```
heyitsbench, line 450: This comment block isn't necessary for the final PR, given that the phases are already enumerated elsewhere in the file.

LuckyLuc96: You're right. I'll take it out.

Issue #22173, opened 2025-05-22
[DK Starting Area] Npcs go in standby if there are no players in visual range or in map.


Rorschach91:
### Current Behaviour

NPCs within the Scarlet Enclave (DK starting area. MapID = 609) do not start their routines unless a DK is playing in that area and the NPC is in the player's line of sight.
In addition, these npc's deactivate their routines when they leave the player's line of sight.

This is a problem because some npc's such as Citizen of Havenshire, Havenshire Stalions/Mares/Colts, Scarlet Miners and Cars etc... may not complete their routines, causing some problems for players.

### Expected Behaviour

Npcs within the Scarlet Enclave area must perform their routines regardless of the presence of players in the area or in sight.

### Steps to reproduce the problem

1. Make a DK
2. Check npcs behaviour inside Scarlet Enclave

### AC rev. hash/commit

[a953710](https://github.com/azerothcore/azerothcore-wotlk/commit/a953710d3a562fb5a9c9a062584e75ce112d89b9)

### Operating system

Windows 11

TheSCREWEDSoftware:
This only applies at the start an server running or at any time there's no player around? (i thought grids when loaded, they dont get unloaded)

Rorschach91:
> This only applies at the start an server running or at any time there's no player around? (i thought grids when loaded, they dont get unloaded)
> 
> Edit: If it's only at the start (when the grids are not loaded) may in the script file having all the area loaded would be the solution

This is an AC function used to avoid unnecessary memory waste, but it is a general function and cannot be enabled/disabled on individual maps (so it currently affects the whole world).

@Yehonal said that we should implement a method to prevent some areas from going to standby.

Yehonal:
> > This only applies at the start an server running or at any time there's no player around? (i thought grids when loaded, they dont get unloaded)
> > Edit: If it's only at the start (when the grids are not loaded) may in the script file having all the area loaded would be the solution
> 
> No, the "feature" you're thinking of is almost purely terrain related. Objects also get created during grid loading but unless they're marked as active they do not remain updated if there is no players within range. Just be warned use of the active flag can be very problematic as it keeps all nearby objects updated as well and is ultimately a hack. There are much better solutions, just need time to implement them.

You mean that the `SMART_ACTION_SET_ACTIVE` is an hack? what would be a better solution in your opinion

Yehonal:
[...]

PR #22175, opened 2025-05-22
fix(DB/Creature) Configure Set_Active on various npcs in dk starting area.

Rorschach91:
I added Set_Active only on those npcs that really need it for an offy-like behaviour.
I don't know if there are other npcs in that area that should need set_active. If this is the case I will open another PR in the future.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Related To https://github.com/azerothcore/azerothcore-wotlk/issues/22173

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/SE_Npcs_Set_Active.sql
+++ b/data/sql/updates/pending_db_world/SE_Npcs_Set_Active.sql
@@ -1,0 +1,188 @@
+
+-- Citizen of Havenshire
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE (`entry` IN (28576, 28577));
+
+DELETE FROM `smart_scripts` WHERE (`entryorguid` IN (28576, 28577)) AND (`source_type` = 0) AND (`id` IN (16));
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+(28576, 0, 16, 0, 25, 0, 100, 0, 0, 0, 0, 0, 0, 0, 48, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Citizen of Havenshire - On Reset - Set Active On'),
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
Leaving comment from @Takenbacon from the reference issue just for any reference on this commit.

> > > This only applies at the start an server running or at any time there's no player around? (i thought grids when loaded, they dont get unloaded)
> > > Edit: If it's only at the start (when the grids are not loaded) may in the script file having all the area loaded would be the solution
> > 
> > 
> > No, the "feature" you're thinking of is almost purely terrain related. Objects also get created during grid loading but unless they're marked as active they do not remain updated if there is no players within range. Just be warned use of the active flag can be very problematic as it keeps all nearby objects updated as well and is ultimately a hack. There are much better solutions, just need time to implement them.
> 
> You mean that the `SMART_ACTION_SET_ACTIVE` is an hack? what would be a better solution in your opinion

Correct. Anything that is setting an object to active isn't great. With enough abuse it's easy to end up with thousands/tens of thousands of objects remaining updated which can lead to a *massive* performance penalty, among other problems. It's also a flag that tries to do too many things, functionality needs to be split up. Sometimes people use it to keep an object updated, sometimes it's to load a grid, sometimes it's for visibility, etc.

The solution would be a revised object updater system that doesn't rely on "marking" a cell, allowing singular objects rather than the entire area to remain updated plus other benefits but it takes some work to do so.

Nyeriah (approved):

Rorschach91:
> Leaving comment from @Takenbacon from the reference issue just for any reference on this commit.
> 
> > > > This only applies at the start an server running or at any time there's no player around? (i thought grids when loaded, they dont get unloaded)
> > > > Edit: If it's only at the start (when the grids are not loaded) may in the script file having all the area loaded would be the solution
> > > 
> > > 
> > > No, the "feature" you're thinking of is almost purely terrain related. Objects also get created during grid loading but unless they're marked as active they do not remain updated if there is no players within range. Just be warned use of the active flag can be very problematic as it keeps all nearby objects updated as well and is ultimately a hack. There are much better solutions, just need time to implement them.
> > 
> > 
> > You mean that the `SMART_ACTION_SET_ACTIVE` is an hack? what would be a better solution in your opinion
> 
> Correct. Anything that is setting an object to active isn't great. With enough abuse it's easy to end up with thousands/tens of thousands of objects remaining updated which can lead to a _massive_ performance penalty, among other problems. It's also a flag that tries to do too many things, functionality needs to be split up. Sometimes people use it to keep an object updated, sometimes it's to load a grid, sometimes it's for visibility, etc.
> 
> The solution would be a revised object updater system that doesn't rely on "marking" a cell, allowing singular objects rather than the entire area to remain updated plus other benefits but it takes some work to do so.

I had read it, but at the moment we have no alternative. If a configuration is implemented in the future, these set_active can be removed, but for now we are forced to do it this way.

PR #22204, opened 2025-05-25
feat(Core/Commands): Add group revive command

brian8544:
## Changes Proposed:

Adds command to revive the whole party/raid. Inspired by other emulation projects. (https://streamable.com/oqfios)
This PR proposes changes to:

## Tests Performed:

This PR has been:
- [x] This pull request requires further testing and may have edge cases to be tested. **-- I checked this to be sure.**

## How to Test the Changes:

1. Create party/raid.
2. Kill players in party/raid.
3. Execute new command `.group revive`
4. See if it crashes, I've tested: targetting npc's, players _not_ in group and targetting myself.


TheSCREWEDSoftware:
Just be clear when you say `Inspired by other emulation projects` this wasn't copied (the code) and just altered to match the AC structure right?

src/server/scripts/Commands/cs_group.cpp
```diff
@@ -257,6 +258,44 @@ class group_commandscript : public CommandScript
...
+        {
+            Player* target = it->GetSource();
+            if (target && target->GetSession())
+            {
+                target->RemoveAurasDueToSpell(27827); // Spirit of Redemption
+                target->ResurrectPlayer(!AccountMgr::IsPlayerAccount(target->GetSession()->GetSecurity()) ? 1.0f : 0.5f);
+                target->SpawnCorpseBones();
+                target->SaveToDB(false, false);
+            }
+            else if (target)
```
Nyeriah, line 290: this is set for offline resurrect, but if the player is offline the target object will return null

brian8544: Thank you, @Nyeriah for the code review. 

I haven't tested resurrecting offline characters, so that's my fault. It's good that you've noticed it.

Offline revive functionality has been removed, other projects don't have this function either and to be fair, I can't think of a reason to revive offline players in a group. Since for specific cases, the singular revive function can be used.

Let me know if there's anything else

Nyeriah: We would need to at least have access to the player guid at that point, I'm not sure if the GroupReference holds it. If it does then it would be easy to implement it, but it's fine as it is.

Issue #22220, opened 2025-05-27
Auras of type 79 - SPELL_AURA_MOD_DAMAGE_PERCENT_DONE double dip on weapon damage % based magic abilities


Macs-Account:
### Current Behaviour

If I gain an aura of enum 79, gaining % damage increase, it will both increase my weapon damage **as well as** the total damage of abilities that are based on weapon damage, if they do not deal physical damage (frost strike).

This means that an ability that deals its damage based on your weapon damage and deals magic damage will double dip this effect, causing them to deal significantly too much damage.

### Expected Behaviour

When I gain a damage buff that increases my percentage damage dealt, the total damage should be increased by the amount of the buff

### Steps to reproduce the problem

1: Cast Frost Strike on a training dummy
2: Give yourself a % damage increase that is done via aura enum 79 (higher is better to demonstrate -- Blood presence works fine though to verify this issue)
3: Cast Frost Strike again on the same training dummy

In my example runs I get the following results
1: Frost strike for 1278 on a target dummy
2: Gain a 799% damage buff
3: Frost Strike for 99288 on the same target dummy, an increase of 7769%
Again with a different buff
2a: Gain a 1999% damage buff
3a: Frost Strike for 538923 on the same target dummy, an increase of 42169%

This result can be observed on any physical weapon damage % ability in the game if you change its damage type to anything besides physical

### Extra Notes

if you flip the bool for whether to consider percentage damage increases for weapon damage based abilities at SpellEffects.cpp at line 3623
weaponDamage = m_caster->CalculateDamage(m_attackType, normalized, true);
to
weaponDamage = m_caster->CalculateDamage(m_attackType, normalized, false);

weapon damage based magic abilities will work properly, but physical weapon damage based abilities will then deal significantly too *little* damage when you gain a % damage buff

### AC rev. hash/commit

987cc31f8adc8519d249fee2696db0500c240c69

### Operating system

Windows 10

PR #22232, opened 2025-05-28
fix(Core/Spells): weapon damage based magic abilities gain too much effect from spell aura % damage increase

Macs-Account:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22220

## SOURCE:

## Tests Performed:

This PR has been:

**Methodology**
List of SpellIds to use while testing this
49143 - Frost Strike (Magic damage based on both Weapon % damage and flat damage)
45902 - Blood Strike (Physical damage based on both Weapon % damage and flat damage)
45477 - Icy Touch (Magic damage dealing spell)
53209 - Chimera Shot (Magic damage based on Weapon % damage only)

48266 - Blood presence for damage buff

To test, get an average damage of a spell without the aura (for spells that have damage ranges). Write it down as X
Apply a % damage buff to yourself. Write it down as Y
Get an average damage of a spell with the aura active (for spells that have damage ranges). Write that down as Z

Z/X should equal very close to Y (within statistical variation)

For ease of testing you can optionally turn up the damage % buff on blood presence or make a custom spell with a really high value of aura name 79

## How to Test the Changes:

1. Without this change, go to a target dummy and use a magic weapon damage ability like Frost Strike, and a physical weapon damage ability like blood strike
2. Gain a aura name 79 Percent damage modifier (such as blood presence)
3. Cast those spells again and calculate how much extra damage was **actually** added to both spells from blood presence (Your physical spell will work as expected, the magic spell will gain way too much benefit)
4. With this change, repeat the above steps

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13194,8 +13194,7 @@ uint32 Unit::MeleeDamageBonusDone(Unit* victim, uint32 pdamage, WeaponAttackType
             if (!spellProto || (spellProto->ValidateAttribute6SpellDamageMods(this, *i, false) &&
                 sScriptMgr->IsNeedModMeleeDamagePercent(this, *i, DoneTotalMod, spellProto)))
             {
-                int32 bonusSchoolMask = (*i)->GetMiscValue();
-                if (!(bonusSchoolMask & SPELL_SCHOOL_MASK_NORMAL) && (bonusSchoolMask & damageSchoolMask))
+                if (((*i)->GetMiscValue() & damageSchoolMask))
                 {
                     if ((*i)->GetSpellInfo()->EquippedItemClass == -1)
                         AddPct(DoneTotalMod, (*i)->GetAmount());
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13194,7 +13194,8 @@ uint32 Unit::MeleeDamageBonusDone(Unit* victim, uint32 pdamage, WeaponAttackType
             if (!spellProto || (spellProto->ValidateAttribute6SpellDamageMods(this, *i, false) &&
                 sScriptMgr->IsNeedModMeleeDamagePercent(this, *i, DoneTotalMod, spellProto)))
             {
-                if (((*i)->GetMiscValue() & damageSchoolMask))
+                int32 bonusSchoolMask = (*i)->GetMiscValue();
+                if (!(bonusSchoolMask & SPELL_SCHOOL_MASK_NORMAL) && (bonusSchoolMask & damageSchoolMask))
                 {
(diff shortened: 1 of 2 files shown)


Tereneckla:
I'd say the problem lies in the implementation of MeleeDamageBonusDone, where it doesn't apply %damage auras when the spell only is physical, but still applies auras that mod physical to other spells

https://github.com/azerothcore/azerothcore-wotlk/blob/8d9636e1fa2b82f9612cfa78ce263741f3416c2c/src/server/game/Entities/Unit/Unit.cpp#L13187-L13208

So if we change L13197, we can filter out all auras that mod physical
```cpp
               // mods for SPELL_SCHOOL_MASK_NORMAL still are already factored in base melee damage calculation
                int32 bonusSchoolMask = (*i)->GetMiscValue();
                if (!(bonusSchoolMask & SPELL_SCHOOL_MASK_NORMAL) && (bonusSchoolMask & damageSchoolMask))```

Macs-Account:
Thank you for the code change suggestion, applying it certainly put things on the right track.
However, After reverting my code change and using yours the damage dealt by frost strike now does just *slightly* too little damage based on what I would expect after gaining a percentage damage boost, whereas 
icy touch (plain spell)
blood strike (physical damage with flat bonus)
shadow bolt (plain spell)
chimera shot (magic damage based on weapon damage with no flat damage)
deal exactly what I expect.

If I manually add the flat damage bonus of frost strike * damage modifier to the actual damage of frost strike I get the exact value I would expect though

So the flat damage portion of frost strike is now not gaining the bonus damage when it should

I'll try to figure it out but I am a bit new at modifying any core code on this project so I will be a little slow, and would certainly accept help

Macs-Account:
Yea, I can see the reasoning behind that but that's misrepresenting the data, which I could see leading to problems down the road.

I haven't had time to sit down and trace out the code path of how every effect on the spells get processed, but if I had to guess any spell that has weapon % damage on it has the entire spell processed through the same means, so the weapon % and flat school damage components both get sent down the same method chain.

If that's the case the solution is probably to have the school damage effect portion of the ability get treated as a spell.

Tereneckla (approved):

Macs-Account:
@sogladev Done! Added to this PR

sogladev (approved):

Macs-Account:
@sogladev I am new to this process, is there something I need to do at this point for this PR or is anything left done by the team?

sogladev:
> @sogladev I am new to this process, is there something I need to do at this point for this PR or is anything left done by the team?

Nothing needs to be done on your end. The label "To Be Merged" means it will get merged soon^tm.

If you want more details, the process is described on the wiki here:
https://www.azerothcore.org/wiki/faq https://www.azerothcore.org/wiki/merge-process

PR #22258, opened 2025-06-01
fix(DB): Significantly increase opcode limit for guild member note opcodes.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

The changes in this PR may be a bit controversial, but in my opinion they are valid, given the existence back during retail 3.3.5 of the EP/GP addon. This addon uses both the public/officer notes to save the values used for the loot distribution system, and when changes are made in mass (e.g. when doing the decay operation), changes are made to these notes very quickly for each member in the guild. Currently, this limit is very low and causes disconnects for users of the addon attempting to do decay for their guild, and can possibly interrupt the process and corrupt the point values.

## Tests Performed:

This PR:
- [X] Has not been tested.

### DIFF
--- a/data/sql/updates/pending_db_world/member-note-antidos.sql
+++ b/data/sql/updates/pending_db_world/member-note-antidos.sql
@@ -1,0 +1,1 @@
+UPDATE `antidos_opcode_policies` SET `MaxAllowedCount` = 1000 WHERE `Opcode` IN (564, 565);


heyitsbench:
Noticed the cause of the disconnects myself was antiDoS policies based on [this commit](https://github.com/skelUA/wotlk/commit/86a9168dea497e9a2187830c96f5f5a2547f764e) authored by @skelUA, though I'm leaning towards being a little more conservative with the count value.

TheSCREWEDSoftware (approved):

PR #22350, opened 2025-06-24
fix(Core/Movement): Use attack speed for leash reset period and only extend timer if in melee range.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/5116

## SOURCE:

## Tests Performed:

This PR:
- [X] Requires further testing and may have edge cases to be tested.

## How to Test the Changes:

### DIFF
--- a/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
+++ b/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
@@ -300,8 +300,8 @@ void ChaseMovementGenerator<Creature>::DoInitialize(Creature* owner)
     i_path = nullptr;
     _lastTargetPosition.reset();
     i_recheckDistance.Reset(0);
-    owner->SetWalk(false);
     i_leashExtensionTimer.Reset(owner->GetAttackTime(BASE_ATTACK));
+    owner->SetWalk(false);
     owner->AddUnitState(UNIT_STATE_CHASE);
 }
 


wetbrownsauce:
I did some testing, here are the results with the PR:

Test 1: Player aggro's mob at its spawn point, but remains at its spawn point without attacking.
Result: Mob continues to attack player indefinitely as it should.

Test 2: Player aggro's a mob and runs a medium distance away, then stops moving and allows the mob to catch up and hit the player. 
Result: Mob reaches player and attacks indefinitely as it should.

Test 3: Player hauls ass and tries to get away from the mob.
Result: Mob resets after a certain distance as it should.

Test 4: Player aggro's mob, begins running away, but periodically stops moving and allows the mob to hit them for a couple seconds. 
Result: Mob will chase indefinitely so long as the player continues to stop and allow the mob to hit them a few times, as it should be. 

Test 5: Player aggro's a mob, is permanently slowed, and tries to run away and escape the mob.
Result: Mob resets after about same distance from Test 3

Test 6: Player aggro's a large mob and attempts to escape it.
Result: Mob chases player and eventually resets as it should.

Test 7: Player aggro's a VERY large mob (Thrym in Zul'Drak) and attempts to run away.
Result: Due to the player being dazed periodically while running away, Thrym will never reset. He is so large, that when he catches up to the player and stops to fight, the time it takes a dazed player to escape his combat reach seems to be enough for him to count as stopped for long enough to reset his timer. If daze is never applied to the player, they are able to escape him. 

Test 8: Player attempts to kite a mob a long distance by running and attacking it.
Result: Mob can be kited indefinitely so long as player keeps periodically damaging it. If the player stops for long enough, the mob resets as it should. 

Test 9: A warlock player dots a large amount of mobs in an area, including casters, and stands still while fighting one of them.
Result: In some rare circumstances, I sometimes have caster mobs randomly reset if I'm not directly hitting them in this scenario. However, I can't seem to get it to happen reliably, and it might be more related to long cast times and distance from initials spawn. Here's a video: [Leash Test](https://www.youtube.com/watch?v=Vyr42lAcATQ)

Test 10: A player aggro's a caster mob with a long cast time for their main spell (4 seconds in this test), and then stands still about 20 yards away. 
Result: The mob's casts will keep it from resetting, provided they don't miss. Had one instance where a cast missed, and then during the follow-up cast, the mob reset. This seems to only happen if the mob was moving prior to the missed cast, so likely its reset timer was close to expiring and the 4 sec cast took too long to go off after the first miss. 

In most cases, this PR seems to be an improvement over the original behavior. Current issues are with very large mobs + slowed players, and casters with long cast times. Let me know if you'd like any additional tests or feedback.

Nyeriah:
Have you tested mobs with really fast attack timers, too? e.g cats Teldrassil cats were an issue before (chasing indefinitely)

wetbrownsauce:
> Have you tested mobs with really fast attack timers, too? e.g cats Teldrassil cats were an issue before (chasing indefinitely)

Just tested. Cats in teldrassil reset after a normal amount of time if you're running away from them. Even if I keep myself permanently dazed, I escape them after a short while. 

Edit: I'm signing off for the night. If any more tests are needed, I'll do them tomorrow!

Nyeriah:
Cool, thanks for testing it!

heyitsbench:
> Current issues are with very large mobs + slowed players, and casters with long cast times.

Firstly, thank you very much for testing! ❤️ I will have to test the behavior of very large mobs like Thrym on official servers to see what should be aimed for. As for casters, they're generally a question mark in terms of expected behavior, as I haven't done much testing myself yet nor have I seen much in the way of specifics regarding how they interact with the leash timer.

PR #22351, opened 2025-06-24
fix(Core): string handling and use smart pointer for strand

FrancescoBorzi:
## Summary
- avoid buffer overflow by replacing `strcpy` with `std::strncpy` in map extractor
- manage logging strand with `std::unique_ptr`

## Testing
- `cmake -S . -B build` *(fails: Could NOT find Boost)*
- `bash apps/ci/ci-run-unit-tests.sh` *(fails: unit_tests not found)*

------
https://chatgpt.com/codex/tasks/task_e_685a7e8173c08325bdedd1d78098a77d

### DIFF
--- a/src/common/Logging/Log.cpp
+++ b/src/common/Logging/Log.cpp
@@ -29,6 +29,7 @@
 #include "Timer.h"
 #include "Tokenize.h"
 #include <chrono>
+#include <memory>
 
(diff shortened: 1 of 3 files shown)


TheSCREWEDSoftware:
If the chatgpt link is supposed to work, doesnt show me anything.

walkline (approved):
Looks good!
I’d call it a refactoring though, since it’s more about improving the code than fixing a bug.

Issue #20448, opened 2024-11-05
Feature: Add support for InfluxDB v2.x for metrics


dlmw:
### Describe your feature request or suggestion in detail

I have an InfluxDB v2 service running, where all clients must provide an [API token](https://docs.influxdata.com/influxdb/v2/admin/tokens/) to read/write data.

I followed the [docs for gathering AzerothCore metrics](https://www.azerothcore.org/wiki/monitoring-azerothcore-with-grafana), but it doesn't look like it supports sending a token.

### Describe a possible solution to your feature or suggestion in detail

The line to provide the credentials look like this:
```
# "hostname;port;database"
Metric.ConnectionInfo = "127.0.0.1;8086;worldserver"
```
I'd love to be able to provide it this way:
```
# "hostname;port;database;token"
Metric.ConnectionInfo = "127.0.0.1;8086;worldserver;DF3F1..."

### Additional context

_No response_

Kitzunu:
The guide says we only support influxdb v1

PR #22358, opened 2025-06-24
BREAKINGCHANGE(Metrics): Support InfluxDB v2

Kitzunu:
### BREAKING CHANGE
- Config setting Metric.ConnectionInfo has been renamed to Metric.InfluxDB.Connection to emphesise that we use InfluxDB

### Notable changes:
- AC Now supports sending data to InfluxDB v2
- Config setting Metric.ConnectionInfo has been renamed to Metric.InfluxDB.Connection to emphesise that we use InfluxDB

By implementing it this way we keep "legacy" support for InfluxDB v1 while also allowing to send data to InfluxDB v2

> [!IMPORTANT]
>  While InfluxDB v2 works, the grafana dashboard is not updated so it will not work with InfluxDB v2

> [!NOTE]
> The following screenshot is taken from the InfluxDB dashboard as proof that the connection works and sends data correctly. (NOT GRAFANA) 
image

* Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20448

### TODO
- The Grafana json files here (https://github.com/azerothcore/azerothcore-wotlk/tree/master/apps/grafana) needs to be updated to use `Flux` queries to be able to access InfluxDB v2.
    - Current implementation uses `InfluxQL` which is only supported by InfluxDB v1

### DIFF
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -812,6 +812,34 @@ Log.Async.Enable = 0
 Metric.Enable = 0
 
 #
+#    Metric.InfluxDB
+#        Description: Connection settings for InfluxDB.
+#
+#        For InfluxDB v1:
+#                     Only fill in Metric.InfluxDB.Connection.
+#
+#        Example:
+#                     Metric.InfluxDB.Connection = "hostname;port;database"
+#
+#        For InfluxDB v2:
+#                     Fill in every field.
+#
+#        Example:
+#                     Metric.InfluxDB.Connection = "hostname;port"
+#                     Metric.InfluxDB.v2 = 0 - (Disabled)
+#                                          1 - (Enabled)
+#                     Metric.InfluxDB.Org = "my-org"
+#                     Metric.InfluxDB.Bucket = "my-bucket"
+#                     Metric.InfluxDB.Token = "my-token"
+#
+
+Metric.InfluxDB.Connection = "127.0.0.1;8086;worldserver"
+Metric.InfluxDB.v2 = 0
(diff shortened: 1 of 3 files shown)


Kitzunu:
New issue to update grafana json: https://github.com/azerothcore/azerothcore-wotlk/issues/22359

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -811,6 +811,37 @@ Log.Async.Enable = 0
...
+#                     Metric.InfluxDB.Connection = "hostname;port"
+#                     Metric.InfluxDB.v2 = 0 - (Disabled)
+#                                          1 - (Enabled)
+#                     Metric.InfluxDB.Org = "my-org"
+#                     Metric.InfluxDB.Bucket = "my-bucket"
+#                     Metric.InfluxDB.Token = "my-token"
+#
+
+Metric.InfluxDB.Connection = "127.0.0.1;8086;worldserver"
+Metric.InfluxDB.v2 = 0
```
heyitsbench, line 841: Should this be defaulted to enabled if v1 is meant to be legacy support?

Kitzunu: no, because our grafana scripts can't fetch data from v2 atm

heyitsbench: Maybe a note should be made to swap it once the issue is resolved?

Kitzunu: there is a disclaimer in the config info, that's enough I recon. But can be added in the grafana issue for sure

IntelligentQuantum (approved):

Issue #22367, opened 2025-06-27
[Core / DBC defined spells] May occur into character loss and an ending error


Plutonicwave:
### Current Behaviour

"Character GUID Full: 0x00000000000003f3 Type: Player Low: 1011 has skill 118 that does not exist."
The log spouts this error everytime you log into the character with the dual wield spell. Unlearning the spell doesn't work, and resetting all spells doesn't work either. Harmless? because I've been experiencing weird & random behavior with quest stages reverting (Turned in quests appear in my quest log again, completed or not)

### Expected Behaviour

No error spam or allow me to simply remove the spell.

### Steps to reproduce the problem

1. Add dual wield spell to a class that isn't supposed to have it (Druid for example)
2. Logout
3. Log back in

### AC rev. hash/commit

AzerothCore rev. 13edac562c57 2025-06-01 14:55:09 +0800 (Playerbot branch) (Unix, Release, Static)
Connected players:1. Characters in world: 1.
Connection peak: 1.
Server uptime: 29 second(s)
Update time diff: 2ms. Last 500 diffs summary:
|- Mean: 2ms
|- Median: 2ms
|- Percentiles (95, 99, max): 4ms, 8ms, 11ms
Using SSL version: OpenSSL 3.0.16 11 Feb 2025 (library: OpenSSL 3.0.16 11 Feb 2025)
Using Boost version: 1.74.0
Using CMake version: 3.25.1
Using MySQL version: 80405
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.1.0-37-amd64
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in ./maps. Total size: 291014951 bytes
vmaps directory located in ./vmaps. Total size: 658130721 bytes
mmaps directory located in ./mmaps. Total size: 2192910844 bytes
Default DBC locale: enUS
All available DBC locales: enUS
Using World DB: ACBD 335.13-dev
Using Playerbots DB Revision:
Latest LoginDatabase update: 2025_02_16_01.sql
Latest CharacterDatabase update: playerbots_names.sql
Latest WorldDatabase update: world_playerbots_rpg_races.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
[...]

PR #22370, opened 2025-06-27
fix(Core/Player): Unlearn invalid spells for race/class on login

Kitzunu:
Added checks in Player::_addSpell to ensure that skills learned from spells are valid for the player's race and class. If a skill is not allowed for the current race/class combination, the spell is removed and an error is logged to prevent invalid skill acquisition.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. be druid
2. .learn 674
3. logout and login - see error in console and the char dont have the spell in the spellbook

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/PlayerStorage.cpp
+++ b/src/server/game/Entities/Player/PlayerStorage.cpp
@@ -6479,7 +6479,7 @@ void Player::_LoadSpells(PreparedQueryResult result)
 
             if (CheckSkillLearnedBySpell(spellId))
                 addSpell(spellId, specMask, true);
-             else
+            else
                 removeSpell(spellId, SPEC_MASK_ALL, false);
         } while (result->NextRow());
     }


TheSCREWEDSoftware:
image

It works and i dont think is working as intended fully

If you learn the spell and relog it gives you the error, working, but now if you try to add the spell again (after the relog or reloging after)

(im .learn)
image

image

I think it was confusing what i said.

Learn dual wield, relog, is gone and gives you the error message nice!
Without reloging (or even relogin) if you try to learn it again it will say unlearn.

Kitzunu:
because you are not supposed to have it

TheSCREWEDSoftware:
But if i do the same e.g for picklock which also a dbc one, 

Shows i've learnt it then unlearns it (but keeps in my spellbook) and bars
image

Kitzunu:
got you, let me see if I can do something about it

kissingers:
I think it would be more reasonable to make it an optional config, so sometime we can set learning skills across races.

TheSCREWEDSoftware:
This are DBC spells, spells that require your class-combo to be in the DBC, for those that mod their client and server doesn't make a difference, you're not supposed to have dual wield without having that DBC changes

TheSCREWEDSoftware:
-- Learnt and logout
image

-- gone
image

-- and you can re added on login (if you wish) and gets removed again :)

image

with more than 1 DBC spell

image

image

image

It's working!

Nyeriah:
Why are we loading spells again? Can't we do this when they are loaded on login already, without additional new query?

Kitzunu:
brainlag or something like that

src/server/game/Entities/Player/PlayerStorage.cpp
```diff
@@ -6472,9 +6472,16 @@ void Player::_LoadSpells(PreparedQueryResult result)
...
     {
         do
-            // xinef: checked
-            addSpell((*result)[0].Get<uint32>(), (*result)[1].Get<uint8>(), true);
-        while (result->NextRow());
+        {
+            Field* fields = result->Fetch();
+            uint32 spellId = fields[0].Get<uint32>();
+            uint8 specMask = fields[1].Get<uint8>();
+
```
Kitzunu:
```suggestion

```

src/server/game/Entities/Player/PlayerStorage.cpp
```diff
@@ -6472,9 +6472,15 @@ void Player::_LoadSpells(PreparedQueryResult result)
...
-            addSpell((*result)[0].Get<uint32>(), (*result)[1].Get<uint8>(), true);
-        while (result->NextRow());
+        {
+            Field* fields = result->Fetch();
+            uint32 spellId = fields[0].Get<uint32>();
+            uint8 specMask = fields[1].Get<uint8>();
+            addSpell(spellId, specMask, true);
+
+            if (!CheckSkillLearnedBySpell(spellId))
+                removeSpell(spellId, SPEC_MASK_ALL, false);
```
p-tkachuk, line 6483: Just wondering - why always add spell and then conditionally remove it? Why not conditionally add it? 
To clean up something, right?

Kitzunu: Having it to clean up on load is mainly for being able to learn abilities while you are ingame for like testing purposes. Otherwise it would always check on addSpell, which imo isnt needed or desirable.

Nyeriah:
```suggestion

            if (CheckSkillLearnedBySpell(spellId))
                addSpell(spellId, specMask, true);
             else
                removeSpell(spellId, SPEC_MASK_ALL, false);
```

I think this is what he meant. We are still only dealing with it in the load, but we dont add if its invalid anyway

Macx-Lio:
Hello Everyone,

I hope this is an optional Functionallity and could be disabled if not wished.

If this realy comes without an Option to disable it, it will destroy the "Mod-Npc-Beastmaster".
The Module "NPC-Beastmaster" allows you to place a NPC called "Withe-Fang" into your World.
This NPC trains you the Beastmaster-Spells and you also can select a Pet from him.
The Module will be useless, if you are realy delete not Class related Spells.

It also will kill my new Project of the MeshUp-Character.
I am working on a Module to implement a Hero-Class, which could pick 3 Skilltrees from any Class.
You will be possible to learn the Spells which depends on the selected Skilltrees.

But I wont be able to genereate every possible Spell-Combination within a single Class into the Database.
As long a Character could learn a Spell, even it is not in the List for his Class, my Module would be possible to work.

So I hope you will make it Optional.

Have Fun :D

TheSCREWEDSoftware:
> Hello Everyone,
> 
> I hope this is an optional Functionallity and could be disabled if not wished.
> 
> If this realy comes without an Option to disable it, it will destroy the "Mod-Npc-Beastmaster". The Module "NPC-Beastmaster" allows you to place a NPC called "Withe-Fang" into your World. This NPC trains you the Beastmaster-Spells and you also can select a Pet from him. The Module will be useless, if you are realy delete not Class related Spells.
> 
> It also will kill my new Project of the MeshUp-Character. I am working on a Module to implement a Hero-Class, which could pick 3 Skilltrees from any Class. You will be possible to learn the Spells which depends on the selected Skilltrees.
> 
> But I wont be able to genereate every possible Spell-Combination within a single Class into the Database. As long a Character could learn a Spell, even it is not in the List for his Class, my Module would be possible to work.
> 
> So I hope you will make it Optional.
> 
> Have Fun :D

I'm unsure if it would break the module, test the module with this PR to see the results

src/server/game/Entities/Player/PlayerStorage.cpp
Kitzunu: ah valid

Raz0r1337:
I think it would be useful if this feature could be disabled in worldserver.conf. I know of some modules that wouldn't work otherwise. Prestige modules, for example, would be particularly affected. If you can simply disable it, then there won't be any problems. 😉

Vtkachenko-Informa:
@Kitzunu 
2025-07-21 20:21:50 Player Riman (GUID: 56743), has spell (19577) that teach skill (270) which is invalid for the race/class combination (Race: 11, Class: 3). Will be deleted.

https://www.wowhead.com/spell=19577. - hunter spell

NoxMax:
@Vtkachenko-Informa could you elaborate more on the character "Riman"? It is a hunter now, but has it always been a hunter? Any custom changes that have been made to Riman within the database?

TheSCREWEDSoftware:
This is happening in CC for intidimation

This issue is the skill 270 "Pet - Generic Hunter"

heyitsbench:
Is it possible it's seeing the line for skill `Pet - Generic Hunter` and returning early before it sees the same spell for skill `Beast Mastery`?

CC: @Kitzunu

Tereneckla:
> Is it possible it's seeing the line for skill `Pet - Generic Hunter` and returning early before it sees the same spell for skill `Beast Mastery`?
> 
> CC: @Kitzunu

It returns if any **one** of the Skills do not match for the player

NoxMax:
I mean in that case... I humbly ask that this PR is reconsidered and possibly reverted. Overall it seems to create more problems than the niche problem scenario it tries to prevent.

Tereneckla:
Or we adjust the logic a little bit https://github.com/azerothcore/azerothcore-wotlk/pull/22537

NoxMax:
That would deal with one side effect, but remains the matter of it breaking intentional commands use and a number of other modules as mentioned by others here.

To be clear, I don't simply think that this PR should be reverted and calling it a day. The issue that started all this #22367 should be handled, by maybe ensuring that learning spells that require DBC edits also require some extra parameter (just to confirm the user/script knows what they're doing) and also making sure `.unlearn` is much more capable in handling the cleaning of those spells.

While this PR is a simpler approach, it's also a sledgehammer that straight up prevents things whether they are desired or not by the server admin.

Nyeriah:
Hopefully the fix provided and the new config are enough to satisfy everybody's needs and we can settle this issue.

fullofstarschen:
Hunter's skill Intimidation(19577) will be wrongly deleted automatically.

TheSCREWEDSoftware:
> Hunter's skill Intimidation(19577) will be wrongly deleted automatically.

Fixed in https://github.com/azerothcore/azerothcore-wotlk/pull/22537

PR #22392, opened 2025-06-29
feat(Core/Maps): Improve map object updater

### LINKED ISSUE #3269: Errors present in all instances
Errors present in all instances

##### CURRENT BEHAVIOUR:
When there is no player in the instance, the creature does not return to its original position

When there is no player in the example, the living creature should immediately return to its original position

Enter the instance, Make creatures follow and attack you, leaving instances, enter the instance again to view the biological position.
1. Enter the instance
image
2. Make creatures follow and attack you, leaving instances
image
3. Leave the instance and enter the instance again to see the biological location. 
image
You can see that the creature is still in position when you left the instance is not moving

##### AC HASH/COMMIT:
https://github.com/azerothcore/azerothcore-wotlk/commit/3cce44469aaf46a9884c276e29dc262c5722f9bb

##### OPERATING SYSTEM:
Windows 7/10

##### OTHER CUSTOMIZATIONS:

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92266819-errors-present-in-all-instances?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Takenbacon:
Decouples map object updater from grid searchers allowing more granular control over object updating. This will enable fixing bugs such as waypoint npcs not moving when out of range of players among other issues. Also is a much better and lighter-weight solution than using setActive for such similar cases.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

Closes #3269

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


TheSCREWEDSoftware:
If you're able point me towards the how to test the PR i don't mind doing at this state so some feedback can be collected

Takenbacon:
> If you're able point me towards the how to test the PR i don't mind doing at this state so some feedback can be collected

It's pretty simple, mostly just making sure objects still function (specifically creatures, gameobjects and dynamic objects). You'll know when an object isn't being properly updated, it won't respond to actions usually.

Takenbacon:
I have completed what I had left to do, I think the WIP tag can be removed and ready for proper testing. As always this is another one of those PR's that has a moderate risk of potentially breaking something so a decent amount of testing before merging would be highly encouraged.

PkllonG:
I used this PR and have found no bugs so far.
Except that the CPU usage has increased a bit.

Takenbacon:
> I used this PR and have found no bugs so far. Except that the CPU usage has increased a bit.

Yeah I expected there may be a very slight bump in initial cpu usage as this PR will maintain more objects being updated in the world. It shouldn't make the server run any slower, just more load at "idle" because less dependent on players updating nearby objects. Ultimately we need to work on removing a couple hacks that have been abused quite a bit over the years, namely setActive and visibilityOverride in which you'd probably end up seeing a net gain.

Takenbacon:
I experimented with it a bit just to double check and as an example all I did was remove active objects from marking nearby cells (active objects themselves are still being updated, just not the area around them anymore) and it cut the objects being updated in half. I think in the vast majority of cases an active object doesn't need to keep the surrounding area active but I wouldn't trust making that change without significant testing as I don't know why many of these objects were set as active to begin with and what it fixed.

src/server/game/Maps/Map.cpp
```diff
@@ -728,6 +629,101 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
...
+    ASSERT(mapUpdatableObject && mapUpdatableObject->GetUpdateState() == UpdatableMapObject::UpdateState::Updating);
+
+    if (obj != _updatableObjectList.back())
+    {
+        dynamic_cast<UpdatableMapObject*>(_updatableObjectList.back())->SetMapUpdateListOffset(mapUpdatableObject->GetMapUpdateListOffset());
+        std::swap(_updatableObjectList[mapUpdatableObject->GetMapUpdateListOffset()], _updatableObjectList.back());
+    }
+
+    _updatableObjectList.pop_back();
+    mapUpdatableObject->SetUpdateState(UpdatableMapObject::UpdateState::NotUpdating);
```
walkline, line 712: Should we invalidate offset index to simplify debugging, what do you thing?

Takenbacon: @walkline I had that exact same thought when writing it. Functionally it makes no difference. if `_mapUpdateState != Updating` then _mapUpdateListOffset is to be considered invalid, but I do agree from a debugging/easier to see what's going on stand point. What stopped me is I am using `std::size_t` which I think makes the most sense in the use context, but it's unsigned. There isn't going to really be a invalid offset, 0 is valid. With that in mind I was like eh, good enough 🤣. I could change it to an int or something but ultimately not sure if it really matters too much, someone trying to access a -1 offset would be just as dangerous as anything else anyways. Alternatively, I could assert GetMapUpdateListOffset if not correct status to ensure validity

src/server/game/Maps/Map.cpp
```diff
@@ -728,6 +629,101 @@ void Map::Update(const uint32 t_diff, const uint32 s_diff, bool  /*thread*/)
...
+            {
+                ++i;
+                continue;
+            }
+
+            obj->Update(diff);
+
+            if (!obj->IsUpdateNeeded())
+            {
+                _RemoveObjectFromUpdateList(obj);
```
walkline, line 653: So, if we're removing an object from the list, it can only be added again through `Player::UpdateVisibilityOf`, right?
I assume `Player::UpdateVisibilityOf` doesn't work the same way as visiting active cells did before, correct?
If that's the case, what's the difference? Will we have a smaller and more accurate set of objects to update?

Takenbacon:
> Interesting approach. I assume it should work faster because we no longer rebuild the update list (which was a linked list) on every map update iteration, and iterating over a vector is faster than over a linked list. I see that this implementation makes creatures with waypoints "active". But aside from that, looking on your previous comments, why are we updating more objects than in the previous implementation during the "idle" state?

Yes, it should theoretically be slightly faster but that was not really an objective of the PR, just a side effect. Even more so considering AC source code specifically was walking the list _twice_ due to the fairly insane "large object" implementation.

VisibilityOverride is mainly what is causing that, it's more or less a temporary solution to just let those objects stay updated until some work can be done to the visibility system to remove that hack as well (similar to how this PR removes the need for the "active" hack in most cases). As mentioned above the object updater would walk the list twice, once for normal objects within normal visibility range, and a 2nd time with an extended range for "large" objects. It's just an unfortunate symptom of many years of abuse on poorly designed flags/systems that need some work to clean up.

src/server/game/Maps/Map.cpp
Takenbacon: Actually more or less it does function the same. Previously the update object grid searchers effectively just iterated every object within visibility range. We are doing the same thing (without having a grid searcher specifically for updating objects) and just marking objects to be updated in the visibility grid searcher loop.

By default all objects upon creation will be updated, and if they do not meet the requirements in IsUpdateNeeded by the next timed update they will be removed from the list (gives scripts, combat, movement etc an opportunity to take effect). From there the only way to be added back to the update list is to be within visibility range. If in the future we find opportunities where it might need to be applied else where, say, maybe a timed event that should run even with no one nearby and without marking the object as active it's very easy to apply.

PkllonG:
have found no bugs so far.

PkllonG:
PR Can fix this https://github.com/azerothcore/azerothcore-wotlk/issues/3269

Takenbacon:
> PR Can fix this #3269

By adding an evade flag check yes it'll fix it

PkllonG:
> > PR Can fix this #3269
> 
> By adding an evade flag check yes it'll fix it

I mean the PR has fixed it

Takenbacon:
> > > PR Can fix this #3269
> > 
> > 
> > By adding an evade flag check yes it'll fix it
> 
> I mean the PR has fixed it

Ah, well probably only partially. It recalculates if objects need updated every 30seconds so they probably get part way back to home position but not quite,  still needs evade flag check to fully return. I'll add it when I can.

PkllonG:
have found no bugs so far.

Yehonal:
Hey @Takenbacon GJ here, I hope more people can test it

Is it the PR you wanted to work on to fix this issue? https://github.com/azerothcore/azerothcore-wotlk/issues/22173

@Rorschach91 worked on a temporary solution for that here: https://github.com/azerothcore/azerothcore-wotlk/pull/22175 
Do you think we should revert those changes or is it safe to have both?

Takenbacon:
> Hey @Takenbacon GJ here, I hope more people can test it
> 
> Is it the PR you wanted to work on to fix this issue? #22173
> 
> @Rorschach91 worked on a temporary solution for that here: #22175 Do you think we should revert those changes or is it safe to have both?

Yes, it is to help support changes such as #22173 in a way that isn't going to be so impactful.

@Rorschach91 changes need to stay most likely, I haven't looked at exactly what those objects are doing that requires the need to remain active however this PR's goal isn't exactly to replace setActive but rather make it more sustainable by removing the limitation that all nearby objects become active as well.

I'd almost consider this PR phase one of two. I've intentionally retained all active functionality for compatibility reasons, eventually we need to go through the uses of setactive and clean them up, once we do we can remove the nearby objects cell markings which will greatly lighten the load they cause.

PkllonG:
have found no bugs so far.
882e689f96bd0f43ca68f4c12d5cb7b7

walkline (approved):
I think the PR is in a good state. 

Let's set the testing deadline for the end of this week, and then merge it if no major issues found.

Takenbacon:
Added a couple small updates: evading objects will remain active and asserted GetMapUpdateListOffset if someone was crazy enough to use the offset when it's not valid.

Kitzunu:
Merging as no bugs has been reported from pkl in over a week

PkllonG:
Except that the CPU usage was a bit higher, no errors were found

LannyE:
(Minor Issue) I think this PR must have forgot to remove the "SetAllCreaturesWithWaypointMovementActive" from the WorldConfig.cpp file.  I now get a "Missing Property" warning about it on startup of the server after applying this change.

skelUA:
@Takenbacon  have a high diff problem after some time . Online players  ~ 250-300 now
image

sogladev:
@Takenbacon This PR introduced a bug related to gameobjects "Acherus Soul Prison" (ids below) as part of the DK quest "The Endless Hunger"

The item "Acherus Shackle Key" is used to unlock the gameobjects, but some are unable to be interacted with. 


https://github.com/user-attachments/assets/a6aa321a-0307-486e-a9df-fdfd51faf7b6



While testing, this is a random amount on each server restart. It's reported that this could affect all of the gobjects, causing players to be stuck on this part of the chain.

open issue: https://github.com/azerothcore/azerothcore-wotlk/issues/22473 

How to reproduce:

```
.quest add 12848
.go c id 29520
```
Try to use the Acherus Shackle Key and see if any of the Acherus Soul Prisons are unusable.

I have not found what's causing this based on the changes in this PR. Do you have any ideas on what may cause some Acherus Soul Prisons to become unusable?

ids:
"Acherus Soul Prison" (IDs: 191577, 191580, 191581, 191582, 191583, 191584, 191585, 191586, 191587, 191588, 191589, 191590)
"The Endless Hunger" Quest 12848
unlock spell "ID - 54669 Unlocking Soul Prison"

Takenbacon:
> @Takenbacon This PR introduced a bug related to gameobjects "Acherus Soul Prison" (ids below) as part of the DK quest "The Endless Hunger"
> 
> The item "Acherus Shackle Key" is used to unlock the gameobjects, but some are unable to be interacted with.
> acherus_soul_prison.mp4
> 
> While testing, this is a random amount on each server restart. It's reported that this could affect all of the gobjects, causing players to be stuck on this part of the chain.
> 
> open issue: #22473
> 
> How to reproduce:
> 
> ```
> .quest add 12848
> .go c id 29520
> ```
> 
> Try to use the Acherus Shackle Key and see if any of the Acherus Soul Prisons are unusable.
> 
> I have not found what's causing this based on the changes in this PR. Do you have any ideas on what may cause some Acherus Soul Prisons to become unusable?
> 
> ids: "Acherus Soul Prison" (IDs: 191577, 191580, 191581, 191582, 191583, 191584, 191585, 191586, 191587, 191588, 191589, 191590) "The Endless Hunger" Quest 12848 unlock spell "ID - 54669 Unlocking Soul Prison"

Seems a bit odd to be related - especially from my understanding this is the only reported issue, I will check regardless.

Takenbacon:
@sogladev https://github.com/azerothcore/azerothcore-wotlk/pull/22474#issuecomment-3071041881

kissingers:
for npcbots crash problem, I ask the AI, and add this to creature.cpp end of function Creature::RemoveFromWorld()
if (Map* map = GetMap())
map->RemoveObjectFromMapUpdateList(this);

will no crash. might not the best fix, but it work.

Takenbacon:
@kissingers I suggest you make an issue on the NPC bots branch, not here. We cannot for obvious reasons prevent custom modules/patches from doing dangerous things, that's up to them to address.

Issue #18213, opened 2024-01-22
(DeathKnight): Risen Ghoul has incorrect auras.


heyitsbench:
### Current Behaviour

Currently, the death knight's [Risen Ghoul](https://wowgaming.altervista.org/aowow/?npc=26125) when [summoned](https://wowgaming.altervista.org/aowow/?spell=52150) has the following auras on it:
```
51996 - Death Knight Pet Scaling 02
54566 - Death Knight Pet Scaling 01
61017 - Hunter Pet Scaling 04 <---
62137 - Avoidance
```

### Expected Blizzlike Behaviour

The Ghoul should instead have the following auras:
```
36492 - Serverside - Summon Heal <---
47466 - Risen Ghoul Self Stun <---
51996 - Death Knight Pet Scaling 02
54566 - Death Knight Pet Scaling 01
61697 - Death Knight Pet Scaling 03 <---
62137 - Avoidance
65221 - Command <---
67557 - Serverside - Pet Scaling - Master Spell 03 - Intellect, Spirit, Resilience <---
67561 - Serverside - Pet Scaling - Master Spell 06 - Spell Hit, Expertise, Spell Penetration <---
```
These spells may also need implementation.

### Source

Wrath Classic (50664)

### Steps to reproduce the problem

1. `.learn 52150`
2. Cast Raise Dead.
3. .list auras on the Risen Ghoul.

### AC rev. hash/commit

d0eae390b61377c569af4bcf2ac647a23a21800b

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
Looks like [Risen Ghoul Self Stun](https://wowgaming.altervista.org/aowow/?spell=47466) is active as a part of https://github.com/azerothcore/azerothcore-wotlk/issues/17684 and not a constant aura.

Joshmash:
Does [Army of the dead](https://wowgaming.altervista.org/aowow/?spell=42650) [ghouls](https://wowgaming.altervista.org/aowow/?npc=24207)  have any incorrect auras?

heyitsbench:
> Does Army of the dead ghouls  have any incorrect auras?

If they do, I'll make an issue for them once I get home.

Issue #18212, opened 2024-01-22
(DeathKnight): Gargoyle has incorrect auras.


heyitsbench:
### Current Behaviour

Currently, the death knight's [Ebon Gargoyle](https://wowgaming.altervista.org/aowow/?npc=27829) when [summoned](https://wowgaming.altervista.org/aowow/?spell=49206) has the following auras on it:
```
51996 - Death Knight Pet Scaling 02
54566 - Death Knight Pet Scaling 01
61017 - Hunter Pet Scaling 04 <---
61697 - Death Knight Pet Scaling 03
62137 - Avoidance
65221 - Command
```

### Expected Blizzlike Behaviour

The Gargoyle should instead have the following auras:
```
36492 - Serverside - Summon Heal <---
51996 - Death Knight Pet Scaling 02
54566 - Death Knight Pet Scaling 01
61697 - Death Knight Pet Scaling 03
62137 - Avoidance
65221 - Command
67557 - Serverside - Pet Scaling - Master Spell 03 - Intellect, Spirit, Resilience <---
67561 - Serverside - Pet Scaling - Master Spell 06 - Spell Hit, Expertise, Spell Penetration <---
```
These spells may also need implementation.

### Source

Wrath Classic (50664)

### Steps to reproduce the problem

1. `.learn 49206`
2. Cast Summon Gargoyle.
3. `.list auras` on the Ebon Gargoyle.

### AC rev. hash/commit

d0eae390b61377c569af4bcf2ac647a23a21800b

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
Possibly related (at least partially) to https://github.com/azerothcore/azerothcore-wotlk/issues/16420.

Joshmash:
Will these fixes change the intervals between pet attacks? This problem was mentioned here https://github.com/azerothcore/azerothcore-wotlk/issues/16420#issuecomment-1902190937

heyitsbench:
> Will these fixes change the intervals between pet attacks?

Without a proper implementation, changing the auras the gargoyle uses means basically nothing. The scaling spells likely need their formulas corrected on top of the spells being properly applied.

sogladev:
```
36492 - Serverside - Summon Heal <---
```
[...]

Issue #17684, opened 2023-11-12
(DeathKnight): Ghoul is not rooted while rising.


heyitsbench:
### Current Behaviour

If you summon a ghoul using [Raise Dead](https://wowgaming.altervista.org/aowow/?spell=46584), the ghoul will follow you while rising from the ground.

### Expected Blizzlike Behaviour

The ghoul should be rooted while rising out of the ground. Once risen, he will begin following the summoner.

### Source

Wrath Classic

### Steps to reproduce the problem

1. `.learn 46584`
2. `.additem 43673`
3. Learn the glyph.
4. Start running.
5. Cast Raise Dead.
6. Observe the ghoul remaining at your side.

### AC rev. hash/commit

bbadc32bea2ad8b60ea591b8bb7b7387babef458

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
This may be a global issue across all ghouls, as far as I know both the pet and open world ghouls should be rooted while rising, but I'd have to double check.

Issue #16420, opened 2023-06-01
(Talent) Gargoyle of Death Knight does low damage


therelus:
### Current Behaviour

The https://wowclassicdb.com/wotlk/spell/49206 of DK does very low damage, i watched it on ofi wow that it deals more damage

### Expected Blizzlike Behaviour

Gargoyle must do more damage

### Steps to reproduce the problem

1. Have talent https://wowclassicdb.com/wotlk/spell/49206 activated
2. Use the spell when you have runicpower
3. Watch damage, you will sense that is too lower

### AC rev. hash/commit

AzerothCore rev. 96a5224d927f+ 2023-04-22 07:25:31 +0700 (npcbots_3.3.5 branch) (Win64, Release, Static) (worldserver-daemon)

### Operating system

x64

Mrming1994:
I have the same problem

Mrming1994:
I find that database-acore_world not mindmg & maxdmg field

therelus:
Hmmm, i think it may be fixed in scripting but i don´t know really

Felcaster:
Gargoyle still have not enough dmg. I create full pvp DK 6651 gears score with Shadowmourne and two best trinkets.
Dr.Damage shows that strike must be ~2684 and total lifetime dmg ~38525 (with all procs around 80000).
But recount shows 1595-1614 strike dmg and 36071 total dmg (gargoyle called with all procs). So it miss ~ 50%dmg.
![WoWScrnShot_080823_155109](https://github.com/azerothcore/azerothcore-wotlk/assets/141717213/69022421-ffad-4551-b9ee-b9ef276441dd)
Screen with procs ~83657 lifetime dmg
![WoWScrnShot_080823_160419](https://github.com/azerothcore/azerothcore-wotlk/assets/141717213/9d1872bf-3d1f-4265-9091-0897ad4a3f4d)

Joshmash:
i hope its gonna be fixed ASAP

Revan2107:
Some player has given me interesting observation. He was testing UDK and spotted that minimal interval between pet's attack is **1 second**. To achieve such result you should have **approximate 500 haste** rating and DK buff for **20% of melee haste**. 

So there is weird thing: usually UDK are being built via melee haste + using haste pot to speed up Gargoyle. Maybe it is a part of problem with Garg damage

Mrming1994:
Is there anyone who can come and fix this bug?

Racknar0:
[...]

PR #22398, opened 2025-06-30
fix(Scripts/Spells): Update Death Knight Ebon Gargoyle auras and damage, and Ghoul auras

sogladev:
## Changes Proposed:

This PR proposes changes to:

Applies auras based on sniffs (https://github.com/azerothcore/azerothcore-wotlk/issues/18212), minus handling for the  4s duration server-side heal aura

Omitted handling Resilience scaling, as pets use the owner's resilience rating
- https://github.com/azerothcore/azerothcore-wotlk/blob/904ddc72cc92bbd668c2a2a0b4da8e2a09de02cc/src/server/game/Entities/Unit/Unit.cpp#L18985

spell_dbc attributes are based from `67561 - spell_pet_spellhit_expertise_spellpen_scaling`, added in PR https://github.com/azerothcore/azerothcore-wotlk/pull/19795

I moved permanent auras from `creature_addon` from the database for consistency.

Damage fixes:
- Applied haste with clamping on DK Pet scaling 02, allowing faster than 1.0 seconds casts
- Fixes regression (https://github.com/azerothcore/azerothcore-wotlk/pull/9718) by changing the EffectHandler of Gargoyle Strike spellscript. This increases damage per cast roughly
```
naked: 250 -> 322
s6 pvp gear: 750 -> 840
```

## Issues Addressed:

Gargoyle
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18212
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/16420

Ghoul
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18213
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17684

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

Lvl80 uhDK with PvP resilience gear

haste effects
```
.cast 26297
.cast 21049
.cast 13491
.cast 3631
```

1. Summon Gargoyle
2. Observe Gargoyle's cast time and damage
3.

- [ ]
- [ ]


src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -800,8 +800,11 @@ class spell_dk_pet_scaling : public AuraScript
     {
         // xinef: scale haste with owners melee haste
         if (Unit* owner = GetUnitOwner()->GetOwner())
-            if (owner->m_modAttackSpeedPct[BASE_ATTACK] < 1.0f) // inherit haste only
-                amount = std::min<int32>(100, int32(((1.0f / owner->m_modAttackSpeedPct[BASE_ATTACK]) - 1.0f) * 100.0f));
+        {
+            float modSpeed = owner->m_modAttackSpeedPct[BASE_ATTACK];
+            modSpeed = std::ranges::clamp(modSpeed, 1e-6f, 1.0f);
+            amount = static_cast<int32>(((1.0f / modSpeed) - 1.0f) * 100.0f);
+        }
```
Tereneckla, line 807: While clamp to 1s casttime is incorrect, it is a rough approximation of a cap/dumb ai the gargoyle actually had

The AI definitely needs a change as the wotlk gargoyle was not very efficient with the casts. One could easily get him to melee if not careful with positioning during the cast

It should only start a new cast once the previous one has landed https://www.mmo-champion.com/threads/697868-Gargoyal-and-Haste?p=7382837&viewfull=1#post7382837

https://web.archive.org/web/20100704014722/http://elitistjerks.com/f72/t72364-unholy_dps_3_3_5_only_exception/p138/#post1683140
image

sogladev: I refrained from modifying the Gargoyle AI in this, so this should be done in a new PR addressing the AI issues you mention.

>  the gargoyle should most likely be affected by gcd (have not found proof for that)

The lower bound could be clamped (to `0.2727...` so the cast time of 2.0 is never lower than hasted GCD 0.75? 

Choosing to melee attack is missing indeed. This seems like generic decision-making behavior, but could be hacked to force players to reposition.

> new cast once the previous one has landed

Do other caster mobs behave this way? This could be hacked together with `OnSpellHit`, but this sounds like generic behavior that's missing.

Tereneckla: Hasted gcd in wotlk is 1s, but the elitistjerk comment disproved my gcd theory.
They still found the gargoyle to cast way less Gargoyle Strikes then it theoretically could

sogladev: I read the new archive link. The archive link provides no evidence that suggests a cast time lower than 0.75 is impossible. 

With Wrath Classic, different, but a cast time of around 0.35 was achieved.

https://www.mmo-champion.com/content/11149-Wrath-Classic-PTR-Testing-a-Death-Knight-Gargoyle-Adjustment

[The response to the linked post](https://web.archive.org/web/20100704014722/http://elitistjerks.com/f72/t72364-unholy_dps_3_3_5_only_exception/p138/#post1683140)
> and do so chain casting
suggests "new cast once the previous one has landed" is likely wrong


I can only conclude that Gargoyle Strike should chain cast and is not affected by a GCD. With this change, a cast time of 0.79 (confirmed during wrath) is possible and down to the value achieved on wrath PTR.

Tereneckla: I never said lower than 0.75s casttime is impossible

My mmo-champion link talks about the travel time adding ~0.05s delay. That could very much still be described as chain casting

sogladev:
I added Risen Ghoul auras to this PR as it relies on the same spell script implementation

Tereneckla:
The PR in itself is ok. I just want to avoid the kind of situation that occurred at the start of wotlk classic and for that a timely change on the gargoyle spellcasting behavior is needed
image

heyitsbench:
I would maybe hold off on adding the ghoul self stun aura for now, I didn't test but I assume the birth animation isn't happening with this PR, and it looks a bit janky without it.

sogladev:
> I would maybe hold off on adding the ghoul self stun aura for now, I didn't test but I assume the birth animation isn't happening with this PR, and it looks a bit janky without it.

@heyitsbench  This is the animation it plays with the stun aura. It should fix
- https://github.com/azerothcore/azerothcore-wotlk/issues/17684

https://github.com/user-attachments/assets/5f3c754c-06df-4e15-a493-83ccd8ff7c7d

> I just want to avoid the kind of situation that occurred at the start of wotlk classic and for that a timely change on the gargoyle spellcasting behavior is needed

@Tereneckla The 335 behavior proposed in this PR should be blizzlike. Problematic haste snapshotting can be addressed in a separate issue/PR to either hardcap the speed, dynamically adjust haste (wrath classic fix),... through a config option.

heyitsbench:
> This is the animation it plays with the stun aura. It should fix #issue

My bad, for some reason in my attempt to add the aura, it did not play the animation, but clearly I must have missed something.

Tereneckla:
> The 335 behavior proposed in this PR should be blizzlike. Problematic haste snapshotting can be addressed in a separate issue/PR to either hardcap the speed, dynamically adjust haste (wrath classic fix),... through a config option.

I believe I gave plenty evidence that the gargoyle AI is not blizzlike. There shouldn't be a need for a config option if that is implemented properly

Melanholyx:
I have tested it on my server , damage is still low like 50% what it should do

Tereneckla:
Looking at it again. The AP scaling of the gargoyle is wrong (at least I comparison to wowsims, which I think is well researched)

We are giving the Gargoyle 33% or 40% (already wrong, impurity has 5 ranks) of AP as SP
Gargoyle strike has a 45% SP scaling
https://github.com/azerothcore/azerothcore-wotlk/blob/d786fe613351a8c34be88f49fedd7e9c0996a542/src/server/scripts/Spells/spell_dk.cpp#L777-L797

wowsims has it scale straight of the AP, no extra 45% mod from Gargoyle strike
https://github.com/wowsims/wotlk/blob/1b5c329859ff2ede795bcf98185e82f8766e54e3/sim/deathknight/summon_gargoyle.go#L139-L165

sogladev:
I checked out wowsims and did some experimenting with a lvl80 t9 geared uhdk and a naked uhdk. The new changes better align with the calculations from the above-linked Wowsims commit (`baseDamage`, `attackPowerModifier`). The damage also more closely matches the "Gargoyle Strike" in the Results tab.

~Previously, Gargoyle Strike was only doing roughly half the damage. 
| Profile | Sim (Avg cast) | After | Before |
|:-------:|:-----:|:-----:|:------:|
| T9      | ~2500 | ~2480  | ~1100   |
| Naked   | ~520 | ~573   | ~280    |

I have 1 unknown `SPELL_ATTR0_SCALES_WITH_CREATURE_LEVEL` which is disabled for pets 

- https://github.com/azerothcore/azerothcore-wotlk/blob/4cbddc073ab18c413b8ae15d26d0ee0d2eae2345/src/server/game/Spells/SpellInfo.cpp#L462

This adds ~300 damage if scaling is applied, but we currently have a spellscript to applies some scaling (~80) instead, and the damage scaling looks to be OK with setting just the scaling to 1

<details>
<summary>
wowsims links
</summary>

T9: https://wowsims.github.io/wotlk/deathknight/#eJzjYnBiMmAUYpC6z9rEyBaQk1iZWsTBKMAldZmRi1/g7S8miZeySmzbDjNd2snExSnw6BOTEjOIySPwHij3QFaJeT2Yt+czk8R3KI9f4DVQ7gdQ37yTTEDExStw/wmTRKucEgtQmgHIfQOUnyKjxHIJxOUXOAo0lRMotewZE0i/kMDLT0wS72WVOK/9BwlcAps5H2jENBklNqDAerBbtsLcwilw7zGQCbKJgYtFYOUTkNCKz0wQ1yAxGZQYFjF9ZeI6wMQwCigCQgak6njFxFDF8IuRX4pXECLwwV6LwYCxi1FX19jIwMDE2NQACIx0jYwNDIyBEMg1NjI0BUoZGALZhsaGxkBBw0eMEhxzrzAJzLoCTA1XmBRmXGHSmHGVyWDmVaYXjCwfGNlnMTIcYmKSYDzBdILxFiPDTWhUqzkodTNzgpkNaQ5QJ4g4SM6aCQIn7S0hIhfsFdPA4Jq90W0mhWAJVi0aJxaNTgdqGNNw8ggBc4KpYg85ABK2e987OEL4P+wTGAFGOIMm


Naked: https://wowsims.github.io/wotlk/deathknight/#eJzjYnBiMmAUYpDaz9LEyBaQk1iZWsTBKMAlZcLFgAdyCqz4zKTEvH4nEzKTQYlhEdNXJq4DTAyjgCIgZECqjldMDFUMvxj5pXgFIQIf7LUYDBi7GHV1jY0MDEyMTQ2AwEjXyNjAwBgIgVxjI0NToJSBIZBtaGxoDBQ0fMQowTH3CpPArCtMEtOuMCnMuMKkMeMqk8HMq0wvGFk+MLLPYmQ4xMQkwXiC6QTjLUaGm9CoVnNQ6mbmBDMb0hygThBxkJw1EwRO2ltCRC7YK6aBwTV7o9tMCsESrFo0TiwanQ7UMKbh5BEC5gRTxR5yACRs9753cITwf9gnMAIApzVD9Q==
</details>

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -784,9 +784,8 @@ class spell_dk_pet_scaling : public AuraScript
         {
             int32 modifier = 33;
```
Tereneckla: Is it actually confirmed that the gargoyle gains 33% of AP as SP. I think we should leave the scaling of the spell alone and change the inherited SP instead.

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -784,9 +784,8 @@ class spell_dk_pet_scaling : public AuraScript
...
 
-            // xinef: impurity
-            if (owner->GetDummyAuraEffect(SPELLFAMILY_DEATHKNIGHT, 1986, 0))
-                modifier = 40;
+            if (AuraEffect* impurityEff = owner->GetDummyAuraEffect(SPELLFAMILY_DEATHKNIGHT, 1986, 0))
+                modifier += CalculatePct(modifier, impurityEff->GetSpellInfo()->Effects[EFFECT_0].CalcValue());
```
Tereneckla:
```suggestion
            if (AuraEffect* impurityEff = owner->GetDummyAuraEffect(SPELLFAMILY_DEATHKNIGHT, 1986, EFFECT_0))
                AddPct(modifier, impurityEff->GetAmount());
```

Tereneckla (changes requested):

src/server/scripts/Spells/spell_dk.cpp
sogladev: L144, L166 https://github.com/wowsims/wotlk/blame/fa745fa59d9b28e49f85f0ce58c0ed1b097fe2be/sim/deathknight/summon_gargoyle.go

`attackPowerModifier := (1.0 + 0.04*float64(garg.dkOwner.Talents.Impurity)) / 3.0`, previously 
`attackPowerModifier := 0.3333333333333333 * (1.0 + 0.04*float64(garg.dkOwner.Talents.Impurity))`

This suggests the 0.33 inheritance is likely correct

`TargetSpellCoefficient: 1` suggests a SP coefficient of 1. I don't see how the DBC scaling value of 0.45 fits? It seems as if this was is no longer used, which is not that uncommon, and scaling auras were used instead.

Tereneckla: In the end a 0.33 mod from DK AP to Gargoyle Strike damage should be achieved. The question just is how blizzard did it. I honestly am fine with either way
The sim obviously can cut corners in the implementation. They're letting the gargoyle inherit attack power 1 to 1 and scale the spell of attack power

TargetSpellCoefficient is a parameter the sim dropped that modified how much a spell would benefit of spellpower increases on the target (think of the old Judgement of the Crusader https://www.wowhead.com/tbc/spell=21183/judgement-of-the-crusader )

The don't add coefficients as a parameter to spells. it just is part of the damage calculation in ApplyEffects
```			baseDamage := sim.Roll(1185, 1377) + spellCoeff*spell.SpellPower()```

sogladev: Thanks for clarifying the `TargetSpellCoefficient`.

Yes, it's not confirmed that the 0.33 SP inheritance is correct. 

I would rather keep the 0.45SP on Gargoyle Strike and up the inherited SP. The result should be the same

Tereneckla (approved):

Rorschach91 (approved):

PR #22401, opened 2025-06-30
feat: startup-scripts reworked + bash scripts workflow  integration

Yehonal:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


apps/installer/main.sh
```diff
@@ -72,7 +74,11 @@ function _switch() {
             printf "AzerothCore Rev. %s\n" "$ACORE_VERSION"
             exit
             ;;
-        ""|"quit"|"15")
+        ""|"sm"|"service-manager"|"15")
```
sogladev, line 77: Using `./acore.sh 14` (service-manager) prints the version.

14 and 15 look to be switched

console output
```
==== ACORE DASHBOARD ====
1) init (i): First Installation                                                                 9) module-remove: (mr): Module Remove by name
2) install-deps (d): Configure OS dep                                                          10) client-data: (gd): download client data from github repository (beta)
3) pull (u): Update Repository                                                                 11) run-worldserver (rw): execute a simple restarter for worldserver
4) reset (r): Reset & Clean Repository                                                         12) run-authserver (ra): execute a simple restarter for authserver
5) compiler (c): Run compiler tool                                                             13) docker (dr): Run docker tools
6) module-search (ms): Module Search by keyword                                                14) service-manager (sm): Run service manager to run authserver and worldserver in background
7) module-install (mi): Module Install by name                                                 15) version (v): Show AzerothCore version
8) module-update (mu): Module Update by name                                                   16) quit: Exit from this menu
```

.vscode/settings.json
```diff
@@ -115,5 +115,12 @@
         "xutility": "cpp",
         "*.ipp": "cpp",
         "resumable": "cpp"
-    }
+    },
+    "deno.enable": true,
+    "deno.path": "deps/deno/bin/deno",
+    "deno.lint": true,
+    "C_Cpp.default.compileCommands": "${workspaceFolder}/build/compile_commands.json",
```
sogladev, line 122: `compile_commands.json` is not created by the dashboard compile with the default config. Should this be added to the CMakeLists.txt or to the dist/config.sh?

.vscode/settings.json
```diff
@@ -115,5 +115,12 @@
         "xutility": "cpp",
         "*.ipp": "cpp",
         "resumable": "cpp"
-    }
+    },
+    "deno.enable": true,
```
sogladev, line 119: deno is no longer used

walkline:
Thanks for putting in all this work on improving things! I can see there's a lot of effort here. I do have some concerns about the direction though.

Honestly, adding such a comprehensive bash app feels like it's making things more complicated than they need to be. Right now we already have some bash in our CI and it's kind of a black box for newer devs - you need to dig through scripts to understand what's happening. Also it's hard to debug and when something breaks it's genuinely hard to figure out what went wrong.

Most of our contributors seem to be more into C++ development than heavy bash scripting, so having a bash app handling so many different problems might work better as a separate project. That way we keep our maintenance efforts focused on the main codebase while people who want this specific opinionated approach to tooling can still use and maintain it on their own. Though if we have enough volunteers who are willing to own and maintain this bash codebase long-term, then I'm fine with keeping it in the main repo (not sure if that is the case here).

If we do keep parts of this, I'd say the dev environment stuff like devcontainer and vscode settings makes sense, but maybe we should rethink the bigger bash parts.

What do you think about this direction?

apps/installer/main.sh
Yehonal: This should be solved in the recent commits. Thanks!

Yehonal:
> Thanks for putting in all this work on improving things! I can see there's a lot of effort here. I do have some concerns about the direction though.
> 
> Honestly, adding such a comprehensive bash app feels like it's making things more complicated than they need to be. Right now we already have some bash in our CI and it's kind of a black box for newer devs - you need to dig through scripts to understand what's happening. Also it's hard to debug and when something breaks it's genuinely hard to figure out what went wrong.
> 
> Most of our contributors seem to be more into C++ development than heavy bash scripting, so having a bash app handling so many different problems might work better as a separate project. That way we keep our maintenance efforts focused on the main codebase while people who want this specific opinionated approach to tooling can still use and maintain it on their own. Though if we have enough volunteers who are willing to own and maintain this bash codebase long-term, then I'm fine with keeping it in the main repo (not sure if that is the case here).
> 
> If we do keep parts of this, I'd say the dev environment stuff like devcontainer and vscode settings makes sense, but maybe we should rethink the bigger bash parts.
> 
> What do you think about this direction?

Thanks for the thoughtful feedback!

You're totally right that bash can be confusing, especially for people not used to scripting. But the goal here is to bring back something we've lost over the years: a zero-config setup that lets people get started with AzerothCore in minutes.

We had that in the past, but it got partially removed due to lack of maintenance. I’d like to avoid that happening again by being the point of reference - so feel free to ping me anytime something breaks. I’m happy to support it and even open a dedicated Discord channel if needed.

I agree that bash isn't the best in terms of DX or debugging. I’ve been experimenting with TypeScript and Deno to offer a friendlier alternative, but even that got removed for unclear reasons. Still, I think it’s a good direction in the future: bash should remain for scripts that cannot be converted, but a better DX layer on top, maybe with TypeScript for those who are more familiar with it.

That said, I believe this scripting layer should live in the main repo, because ops/dev tooling is not something we can do in C++ and it needs to evolve with the rest of the code. We always keep the fallback install method for “experts”, but the default path should be as easy as possible (that's also why AC is probably the only emulator maintaining a docker layer natively as well)

This is just one piece of the puzzle I'm working on these days.

Again, just ping me on Discord when anything in this regard is needed. 

Open to feedback and improvements anytime! :)

.vscode/settings.json
Yehonal: Will probably be soon, again. Let's keep it for now

.vscode/settings.json
Yehonal: The VSCode IDE is creating it automatically with a side-configuration job at the startup

But I've added it in the CmakeFile too just in case :)

FrancescoBorzi (approved):
@Yehonal good job. Approving this counting on your promise to actively support this from now on in the future.

> I’d like to avoid that happening again by being the point of reference - so feel free to ping me anytime something breaks.

Issue #20786, opened 2024-11-30
[Creature Leash] - Creature will not Leash if itself deals self-inflicted damage like a dot


TheSCREWEDSoftware:
### Current Behaviour

If you're fighting a creature and try to run away and for some reason they self-afflict damage (like a dot) the timer will be reseted per tick of damage (being 1 time damage 1 usage 1 tick  or dot damage for each tick). So it wont leash.

### Expected Blizzlike Behaviour

I would assume that the leash timer would not be reset by the creature's self-inflected damage.

### Source

None at the moment. It's hard to find sources on this because probably don't often run away from these to notice that behaviour. Also because probably because it works as intended (not extending the leash) people dont report it because doesnt happen.

### Steps to reproduce the problem

Two ways.

You can do the same tests like i did with the video below.
- Choose whatever mob you want to try it
- Aggro it, run away, see it leash
- Do the same but as you agro it run the following command (with it selected) `.cast self 8699` // casts Unholy Frenzy on the creature
- Now try to leash while the debuff is up (you can repeat casting the `.cast self` if you want to pronlonge the issue.

or you can
- `.go xyz -10418 383 45 0 1.3`
- Go down the crypt and find `Skeletal Warder`
- Pull `Skeletal Warder` and pull other mobs. 
- Walk away to leash. While `Skeletal Warder` is alive because keeps inflicting self damage, the social aggro creatures wont ever reset until he dies.

### Extra Notes

Video of my testing/findings. [here](https://www.youtube.com/watch?v=vCH0Vq2Yq7M&feature=youtu.be)

Went deeper in this matter because of this issue: https://github.com/chromiecraft/chromiecraft/issues/7560?notification_referrer_id=NT_kwDOAQKW4bQxMzU5MTQ4Mzg4MjoxNjk0NjkxMw#issuecomment-2509143501

It's not the ONLY reason of the issue, is part of it.

### AC rev. hash/commit

AzerothCore rev. 55f0630c09ad 2024-11-26 21:53:32 +0000 (master branch) (Win64, RelWithDebInfo, Static)

```
Connected players: 1. Characters in world: 0.
Connection peak: 1.
Server uptime: 55 minute(s) 50 second(s)
Update time diff: 1ms. Last 500 diffs summary:
|- Mean: 2ms
|- Median: 2ms
|- Percentiles (95, 99, max): 4ms, 5ms, 5ms
[...]

PR #22449, opened 2025-07-09
fix(Core/Unit): Don't extend leash when taking self-inflicted damage.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/20786

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -13730,7 +13730,7 @@ void Unit::CombatStart(Unit* victim, bool initialAggro)
         victim->SetInCombatWith(this);
 
         // Update leash timer when attacking creatures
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
I've to test this when i get the chance, but i would be happy to no longer have to revivie HC chars in the skeleton place below westfall

Nyeriah:
I'm not sure if this is possible at all, in this case this isnt the damage function but the one that sets them in combat, and it would need to set them in combat with... themselves, I dunno

heyitsbench:
I did test the case mentioned in the linked issue, and it did resolve it.

Nyeriah (approved):

PR #22462, opened 2025-07-11
fix(Core/FleeingMovementGenerator): Prevent fleeing players from moving off cliffs

walkline:
Resolved an edge case where the movement generator could direct fleeing targets
over steep ledges or cliffs, causing unintended falls.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

https://www.youtube.com/watch?v=yF-j_2epviQ

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


heyitsbench:
CC: @Apathyxia

Apathyxia:
Still managing to do this, but it's definitely way more difficult to do now. I already see way less abuse potential compared to current situation.

https://github.com/user-attachments/assets/1bf2569f-81d0-4bee-83d7-b384e4cc9adb

I know of several places where this was an issue, and I have tested them now. Only the lumber mill is where I am somewhat able to create an issue, and even that is not competely consistent. 

I tested:
- Blade's Edge Arena's pillars: 
        Fear did not bring you down from the pillars or the ropes. You also didn't fall through the ground.
- Blade's Edge Arena's long wooden platform's edges: 
        Polymorph did not make you fall down. 
- Warsong Gulch flag room roof edges: 
        You did not fall down through the structures.
- Arathi Basin Lumber Mill: 
        Apart from above example, it works very well. You are not feared through the ground.
- Eye of the Storm flag spawn area: 
        You do not fall off the sides or through the ground.

This is a tremendous improvement over the current system. I can't say with 100% certainty that above cases always work properly, but in my tests (~30-50 casts on each place) it looks very promising.

heyitsbench:
If you can, try and record coordinates (`.gps`), they may help with debugging/fixing the issue. Walkline might say I'm wrong though, if I am then don't listen to me. 😛

Apathyxia:
Well, if it helps, here are two rough positions where I got feared to and the character fell down from:

image

image

walkline:
Our mmaps generator uses a 60 angle as the maximum slope, and as a result, we are getting the following "valid" nodes: image
If we reduce it to 50°, it would look like this::
image
However, changing this value globally might break movement in other areas. So I'm not sure what the best approach is...
But maybe we can address this in some other issue.

Apathyxia:
Reducing it to 50° definitely looks better, but as you said, it might have undesired effects elsewhere. Perhaps it would still be an interesting thing to research at some point. 

That being said, I think the current PR is a massive improvement and would solve most of the cases happening. 
If it is up to me, I'd approve it.

heyitsbench (approved):

walkline:
@Apathyxia with this PR https://github.com/azerothcore/azerothcore-wotlk/pull/22506 Lumber Mill area should be safe.

PR #22474, opened 2025-07-14
fix(Scripts/ScarletEnclave): Always start Unworthy Initiate event when interacting with Soul Prison.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Scripts (bosses, spell scripts, creature scripts).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22473

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.
- [X] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [X] This pull request can be tested by following the reproduction steps provided in the linked issue.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

Issue #22473, opened 2025-07-14
[DK Chain Quest] Acherus Soul Prisons could become unblockable.


Rorschach91:
### Triage

Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/8144
Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/8143

### Current Behaviour

The key is recieved, but when trying to free the Unworthy Initiate, it does absolutely nothing on any of them.

[The Endless Hunger] : https://wowgaming.altervista.org/aowow/?quest=12848

### Expected Blizzlike Behaviour

You should be able to right click the Archerus Soul Prison, and then begin to free the shackled npc.

https://www.youtube.com/watch?v=RAYGrKpsMbY

### Source

![Image](https://github.com/user-attachments/assets/05a643fd-ca6c-4e3a-9c83-3f0f57cfae50)

Screenshot taken in CC server

### Steps to reproduce the problem

1. Pick up quest: [The Endless Hunger] by questgiver "Instructor Razuvious"
2. Go to the pit and attempt to use [Acherus Shackle Key] by right clicking Acherus Soul Prison

### AC rev. hash/commit

[5e30c5a](https://github.com/azerothcore/azerothcore-wotlk/commit/5e30c5abaaa2977a670da883a114338107800122)

### Operating system

Windows 11

### Customizations

None

### Server

ChromieCraft

PR #22496, opened 2025-07-16
fix(Script/ScarletEnclave) Acherus Soul Prisons now reset properly.

Rorschach91:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22473
- Closes https://github.com/azerothcore/azerothcore-wotlk/pull/22474

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -712,7 +712,7 @@ public:
                             }
                         }
 
-                        if (prison)
+                        if (prison && prison->getLootState() != GO_NOT_READY)
                             prison->ResetDoorOrButton();
                     }
                     break;
--- a/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
+++ b/src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
@@ -711,7 +711,8 @@ public:
                                 }
                             }
                         }
-
+                        // Must check for loot state as out of order updates will reset
+                        // the prison gameobject during spawn causing invalid state
                         if (prison && prison->getLootState() != GO_NOT_READY)
                             prison->ResetDoorOrButton();
                     }


src/server/scripts/EasternKingdoms/ScarletEnclave/chapter1.cpp
```diff
@@ -712,7 +712,7 @@ class npc_unworthy_initiate : public CreatureScript
                             }
                         }
 
-                        if (prison)
+                        if (prison && prison->getLootState() != GO_NOT_READY)
```
Yehonal, line 716:
```suggestion
                        if (prison && prison->getLootState() == GO_ACTIVATED)
```

The [ResetDoorOrButton](https://github.com/azerothcore/azerothcore-wotlk/blob/7712c2a97346a69ff1e2bca731d3f7e2aa09c494/src/server/game/Entities/GameObject/GameObject.cpp#L1404) works only when the object is GO_ACTIVATED, maybe we can check that directly

Takenbacon: > The [ResetDoorOrButton](https://github.com/azerothcore/azerothcore-wotlk/blob/7712c2a97346a69ff1e2bca731d3f7e2aa09c494/src/server/game/Entities/GameObject/GameObject.cpp#L1404) works only when the object is GO_ACTIVATED, maybe we can check that directly

Not quite, when a gameobject is first created (but before the first update) loot state is set to GO_NOT_READY which is what triggers this issue.

I do believe ResetDoorOrButton function in general probably should check for GO_NOT_READY (or by just checking if state != GO_ACTIVATED like you mentioned) but who knows what that could potentially break.

Rorschach91: I only know that with this fix the bug cannot be reproduced.

Yehonal (approved):

sudlud:
what about including the comment by @Takenbacon also
https://github.com/azerothcore/azerothcore-wotlk/pull/22474#issuecomment-3071041881

it's good to have the comment there imo to understand why this check is beeing done

sudlud:
and does this also close https://github.com/azerothcore/azerothcore-wotlk/pull/22474 ?

heyitsbench:
This does close the linked PR, and I do agree with adding the comment.

Issue #22468, opened 2025-07-13
Feature: provide config file for mmaps generator


walkline:
### Describe your feature request or suggestion in detail

It seems that the global config doesn't work well for some specific places, like here: https://github.com/azerothcore/azerothcore-wotlk/pull/22462#issuecomment-3067024680.

It would be great to have a YAML or JSON file (I prefer YAML 😉) where we could define values like walkableSlopeAngle and other parameters from [this section](https://github.com/azerothcore/azerothcore-wotlk/blob/88342576e070476f4866a82873bc9b1080e1ab99/src/tools/mmaps_generator/MapBuilder.cpp#L1060-L1076) — both at the global level and optionally per map or per tile.

TheSCREWEDSoftware:
Wouldn't this provide false-positives for people that extract their own maps? when report bugs. Should these numbers be tweaked and be fixed, instead of configurable? 

Numbers that match more overhaul the expected behaviour? And the exampel you shown from 60 to 50º change, can this currently cause "holes" on the map?

`.go xyz -11 1199 109 1 3`
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ac837e24-fa27-47c6-8f92-fe8ac546cc3d" />

This is a commun place where people die (assumed getting stuck under the getting flip up again and dying, or just charge anywhere in borean tundra eventually you will fall between any creater)

walkline:
What do you mean by false positives? Do you mean cases where users modify the config, generate mmaps, and then report invalid pathfinding? If so, you're right — that can happen. But in those cases, we can always request their config to verify.

That said, I believe making these kinds of settings more accessible would encourage more users to experiment, which could ultimately lead to improvements in map quality. I think that’s a worthwhile tradeoff.

We could also add a mechanism in the core to display information about the config used — for example, a config hash or even the full config content.

As for the specific area you mentioned, I’m not sure what you mean. The only strange thing I noticed was odd pathfinding behavior in that area. 

![Image](https://github.com/user-attachments/assets/16ed83d8-19c1-45ea-9576-56a31ff2633d)

In recastnavigation, a different path is generated, but that could be due to slight config differences between recast and the core.

<img width="1125" height="764" alt="Image" src="https://github.com/user-attachments/assets/7104023f-fc0e-4550-b071-54c3abd68253" />

Kitzunu:
[...]

PR #22506, opened 2025-07-18
feat(Core/mmaps): Add configuration file for mmaps-generator.

walkline:
With this PR, I tried to make mmaps more accessible for debugging and modifications. I’ve been thinking through some internal configuration parameters and exposing them to users via a config file, along with (hopefully) clear explanations. Not sure if it all turned out well — I’d appreciate a fresh perspective to see if things are clear and make sense.

This change introduces a YAML config file for the mmaps-generator.
You can place the config file in the same folder as the executable (handled by make install), or specify it explicitly using the --config argument: `./mmaps-generator --config /my/path/config.yaml`.

With this config file, it’s also possible to override configuration values per map or per tile.

Additionally, I modified the tile header to include the configuration values used during generation. This allows us to identify which tiles need regeneration after config changes and makes it possible to expose this debug information in-game.
Because of this change, I bumped the mmaps version number — so existing mmaps need to be regenerated (or redownloaded, if someone provides them).

Here’s how the debug information looks in-game (just type .mmap loc):photo_2025-07-18 18 29 45

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22468
- Fixes https://github.com/azerothcore/azerothcore-wotlk/pull/22462#issuecomment-3067024680


src/tools/mmaps_generator/mmaps-config.yaml
```diff
@@ -0,0 +1,142 @@
...
+  #    - RecastDemo expects this folder to be in the same directory as its executable.
+  # 7. In RecastDemo:
+  #    - Click "Input Mesh" and select the `.obj` file
+  #    - Choose "Solo Mesh" in the Sample selector
+  # 8. (Optional) Reuse the Recast config values from step 3:
+  #    - `cellSizeHorizontal` → "Cell Size"
+  #    - `walkableSlopeAngle` → "Max Slope"
+  #    - `walkableClimb` → "Max Climb"
+  #    - and so on
+  # 9. Scroll to the bottom of RecastDemo UI and press "Build" to generate the navigation mesh
```
walkline, line 144: Maybe someone can volunteer to create a wiki page based on this, with a clearer explanation and a tested flow for Windows (I'm using macOS).

PkllonG:
complicated

PkllonG:
complicate
Unfortunately, this is already the best way

walkline:
Thanks for looking into this!

Could you clarify what exactly you find complicated?
Is it the regular extraction process, modifying the mmap parameters, the PR description itself, or perhaps the code?
Or maybe I added too much detail in the comments, which could make things more confusing—especially if you're not a native English speaker.

> Unfortunately, this is already the best way

What did you mean by this?

That the process was clearer before this PR? If so, let’s compare both approaches using two typical use cases.
**Use Case 1:** Regular User Extracting mmaps
_Current approach_: Build the mmaps-extractor, place it in the folder with maps and vmaps, then run the executable.
_This PR approach_: Same as before, but now you also copy a config file. You don’t even need to open or edit it unless you’re interested.

This doesn’t seem like an overcomplication to me—what do you think?

**Use Case 2**: Developer Investigating Movement Issues
This is based on my experience a few weeks ago.
**_Current approach_**:
Suppose you get a report that a creature behaves oddly in a specific area. A good first step would be to visualize the navigation mesh for that location.
But how?
I knew of the RecastNavigation project, but I didn’t know how to use it to visualize our maps.
After collecting bits and pieces of info online, I was still unsure what parameters we use in our system, where to find them in the code, and how to reproduce the setup in RecastDemo.

**_This PR approach_**:
There’s now guidance on how to visualize navigation data: https://github.com/azerothcore/azerothcore-wotlk/pull/22506/files#diff-7e699f5fb22754aaa65619c81367db5c12b8393019b0d285cc9d8fa9818e6598R122
Also, the most important parameters are exposed in a config file with descriptions.
You can tweak them at the global, map, or tile level without recompilation. The debug command (.mmap loc) provides values for RecastDemo.

So from my perspective, we previously had no clear guidance on handling this kind of issue. With this PR, we at least have a starting point and some structure.

That’s why I believe it’s a step forward.

PkllonG:
It was a translation error,
Sorry for misunderstanding what I mean.

src/tools/mmaps_generator/mmaps-config.yaml
TheSCREWEDSoftware: I can give a shot, just need to try the PR (or when is merged) when I get time @Kitzunu any specific page place i can create wiki page for this?

TheSCREWEDSoftware: @walkline current values (default) in the mmaps generator config matches the current core values?

walkline: Yeah, they should be the same.

TheSCREWEDSoftware: Just needed to know, to see if wowgaming needed to release AC Data V17(?) :v

Nyeriah: Yes, maps will need to be reextracted.

Nyeriah:
Since maps need to be reextracted, after the testing phase, we will need to coordinate the release process to ensure someone is available to update them (to whoever happens to be the person to press the green button).

TheSCREWEDSoftware:
Than i think i miss understood walkine's changes. If we do not change / update current MMAPS this information cannot be obtained? (with the new debug information).

Functionality wise, what does this change from current MMAPS vs re-compiled MMAPS. (with the same settings / values)

Related to the 

`we will need to coordinate the release process to ensure someone is available to update them (to whoever happens to be the person to press the green button).`

I've no issue in doing this

TheSCREWEDSoftware:
I will see if i get around this PR this month.

Rorschach91:
> New Extracted AC_Data_V17: [Here](https://mega.nz/file/4hd1QTDC#44NxUKTWHthrsC7Ry2GZtqLeNea5GPDtt0wRThuw4Zc)

I added this link with new extracted maps.

kissingers:
If bulid select with tools, copy mmaps-config.yaml to mmaps_generator.exe same path more good.

PR #22136, opened 2025-05-17
feat(Core): make BGs and Arena prep time configurable

NoxMax:
**Apologies for reposting the PR. I accidentally closed the original -_-**

## Changes Proposed:
PR makes BGs and Arena prep time configurable in worldserver.conf. Keeping the default values of 120s/60s for BGs and Arena respectively. Now you don't have to wait if the default times are too long for you.
Making the timing configurable was simple, but making the timing announcements accurate is what required most of the work. Previous code logic had announcements triggered based on event flags rather than actual time values (full/half/quarter/none) The new code makes _ProcessJoin in Battleground.cpp actually aware how much time is left till match begins.
This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:
None on GH, but requested feature on Discord.

## SOURCE:
N/A Default behavior has not been changed.

## Tests Performed:
Compiled correctly on Debian.
Changes behaved as expected in-game.
This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [x] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

1. At the default values of 120s/60s for BGs and Arena, there should be no observed change in game.
2. Change the prep time of BGs and Arena to something other than 120s/60s and restart the server. The prep time should be whatever you changed the value to.
3. The announcement messages should be accurate. If it says 30 seconds till start, match must start in 30 seconds.
4. BGs have announcements at 120/60/30/0 seconds. Arena has announcements at 60/30/15/0 seconds. If you set a prep time that is less than one of the announcement times, you should not see an announcement for a time that is greater than the configured perp time. Example: If you set the prep time to 45 seconds for BGs, you should only get the 30/0 second announcements, and at their correct time.

Please advise after testing if anything has been missed.

## Known Issues and TODO List:

None for now.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

[...]

PR #22521, opened 2025-07-21
feat(Core/Player): Allow spell validation to be skipped at server owner's risk

LuckyLuc96:
There are risks involved with this and it does reintroduce the patched bug. But I think the server owner should be allowed to risk this. And I really want to keep playing my warrior who has the mage teleport spells. It works.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

Reintroduces issue: https://github.com/azerothcore/azerothcore-wotlk/issues/22367
Related to PR: https://github.com/azerothcore/azerothcore-wotlk/pull/22370

## Images

DEFAULT:
ValidateSkillLearnedBySpells = 1 -- When a character does indeed have invalid spells.
Screenshot From 2025-07-21 16-01-52

ValidateSkillLearnedBySpells = 0 -- The same character is allowed to have the invalid spells. 
Screenshot From 2025-07-21 16-00-37

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [x] Reintroduces https://github.com/azerothcore/azerothcore-wotlk/issues/22367
- [ ]


src/server/game/Entities/Player/Player.cpp
```diff
@@ -3103,6 +3103,9 @@ bool Player::addSpell(uint32 spellId, uint8 addSpecMask, bool updateActive, bool
 
 bool Player::CheckSkillLearnedBySpell(uint32 spellId)
 {
+    if (sConfigMgr->GetOption<bool>("ValidateSkillLearnedBySpells", true))
```
Tereneckla:
```suggestion
    if (!sConfigMgr->GetOption<bool>("ValidateSkillLearnedBySpells", true))
```

src/server/game/World/WorldConfig.h
```diff
@@ -472,7 +472,7 @@ enum ServerConfigs
     RATE_MISS_CHANCE_MULTIPLIER_TARGET_CREATURE,
     RATE_MISS_CHANCE_MULTIPLIER_TARGET_PLAYER,
     CONFIG_NEW_CHAR_STRING,
-
+    CONFIG_VALIDATE_SKILL_LEARNED_BY_SPELLS,
```
Tereneckla, line 477:
```suggestion
    CONFIG_VALIDATE_SKILL_LEARNED_BY_SPELLS,
    
```

Tereneckla (changes requested):

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -1765,6 +1765,16 @@ CleanCharacterDB = 0
 
 PersistentCharacterCleanFlags = 0
 
+#
+#    ValidateSkillLearnedBySpells
+#        Description: If enabled (1), players will lose spells that are invalid for their race/class.
```
Tereneckla:
```suggestion
#        Description: If enabled, players will lose spells that are invalid for their race/class.
```

src/server/game/Entities/Player/Player.cpp
LuckyLuc96: Fixed thank you. Will push soon

LuckyLuc96:
I may have the config backwards do not merge just yet

LuckyLuc96:
Okay after recompiling and testing both versions of the settings. And also trying to not confuse myself with boolean logic that includes a !not in there I am certain this is all working correctly as is.

TheSCREWEDSoftware:
@amed80 want to give a try to this see if it's fixes the previous issues you've shared? (about getting your spells deleted)

amed80:
yes, this works

NoxMax:
To be honest, I think this should be disabled by default for now (i.e: override the original patch) With respect to @Kitzunu work, I don't think the patch was the right approach to the issue. It just goes after all "invalid" spells, regardless of whether it was acquired by mistake or deliberately.

And on the original PR, one person is even reporting a false positive targeting of a valid spell.

Kitzunu:
Then the dbc data is wrong, not a false positive.

I will review this tonight, it is alright. And should be default off, blizzlike.

NoxMax:
While Blizzlike should be the standard, I have to say that the original patch goes far too broadly and strongly in enforcing that, in a way that might be antithesis to the standard itself.

Consider an actual Blizzard GM on an official server. If they used the .learn command to add a spell that their GM toon class doesn't typically have access to, would the server's default behavior deny and undo the GM's command?

sudlud:
could probably exclude GMs from this in general.

NoxMax:
Which kinda loops back to the original issue #22367, and what would be the point of the patch then. That person after all used GM commands to create the issue, as opposed to by gameplay means.

The way I see it, if there is something in game that would a give a toon on a non-GM account something that they shouldn't have, then that thing should specifically and directly be addressed, rather than some general solution that targets many things.

Meanwhile with GM accounts, the reality is that there are quite a few commands that can break elements of the server if misused. And that should be an acceptable risk for commands misuse.

heyitsbench:
> Consider an actual Blizzard GM on an official server. If they used the .learn command to add a spell that their GM toon class doesn't typically have access to, would the server's default behavior deny and undo the GM's command?

You're assuming the existence of a `.learn` command on Blizz's end.

NoxMax:
I think it is a safe assumption that some `.learn` equivalent is accessible to Blizz GMs, at least based on back in the day when they used to showup in world and cast some wacky stuff for some players' entertainment, and even take requests for things to cast.

Regardless, the larger point is about server-GM relation. That the server shouldn't deny a GM's command just because it would be considered invalid for a regular player.

TheSCREWEDSoftware:
> I think it is a safe assumption that some `.learn` equivalent is accessible to Blizz GMs, at least based on back in the day when they used to showup in world and cast some wacky stuff for some players' entertainment, and even take requests for things to cast.
> 
> Regardless, the larger point is about server-GM relation. That the server shouldn't deny a GM's command just because it would be considered invalid for a regular player.

I really doubt outside of development realms or builds (like the leaked alpha version of wow) that any GM in the world would have access to any tools that can directly affect the player outcome in game.

Learning spells, giving quests, giving items (this a known, but is either assume from an external blizzard tool, not in-game), we have all the gm commands implement AC for convience, probably more than 90% of this can't even be accessed by highest GM in official servers in-game, the same way speculate how they do certain parts of the database, GM usage would probably complety different than is on the emulators side.

heyitsbench:
> I really doubt outside of development realms or builds (like the leaked alpha version of wow) that any GM in the world would have access to any tools that can directly affect the player outcome in game.

They absolutely did have these tools, even the lower tiers of GMs. Here's a couple videos from a former GM recalling their experience and how they carried out the job: [video 1](https://youtu.be/tVJqcFCIzlw), [video 2](https://youtu.be/m4-xdjS21hw).

NoxMax:
> that can directly affect the player outcome in game.

To be clear I was talking about the variety of spells they would give to their own selves to do fun demonstrations. This is based on player stories and videos I have seen of GMs doing that on live servers, in pre-Cata times. But yes as far as what they gave to players, it was always nonconsequential and they had it removed before leaving. Something like turning the entire raid party into animals as they were prepping (actual video I've seen them do)

heyitsbench:
> we have all the gm commands implement AC for convience

This also isn't true.

Nyeriah:
Sorry to interrupt what seems to be a heated discussion but @Kitzunu has let us know that they will be reviewing this PR later this evening, so let’s wait for his response regarding this issue

Kitzunu:
This is what I would do, then I think it is good to be merged.
As for false positives - they need to be fixed seperatly as they would be an issue with the original DBC files.

```diff
diff --git a/src/server/game/Entities/Player/Player.cpp b/src/server/game/Entities/Player/Player.cpp
index 89e766860c..32c98cba71 100644
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -3101,9 +3101,12 @@ bool Player::addSpell(uint32 spellId, uint8 addSpecMask, bool updateActive, bool
     return true;
 }
 
-bool Player::CheckSkillLearnedBySpell(uint32 spellId)
+bool Player::CheckSkillLearnedBySpell(uint32 spellId, Player* player)
 {
-    if (!sConfigMgr->GetOption<bool>("ValidateSkillLearnedBySpells", true))
+    if (!sWorld->getBoolConfig(CONFIG_VALIDATE_SKILL_LEARNED_BY_SPELLS))
+        return true;
+
+    if (player->IsGameMaster())
         return true;
 
     SkillLineAbilityMapBounds skill_bounds = sSpellMgr->GetSkillLineAbilityMapBounds(spellId);
```
```diff
diff --git a/src/server/game/Entities/Player/Player.h b/src/server/game/Entities/Player/Player.h
index a098eaf1ce..1e497d0f74 100644
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1720,7 +1720,7 @@ public:
     void learnQuestRewardedSpells();
     void learnQuestRewardedSpells(Quest const* quest);
     void learnSpellHighRank(uint32 spellid);
-    bool CheckSkillLearnedBySpell(uint32 spellId);
+    bool CheckSkillLearnedBySpell(uint32 spellId, Player* player);
     void SetReputation(uint32 factionentry, float value);
     [[nodiscard]] uint32 GetReputation(uint32 factionentry) const;
     std::string const& GetGuildName();
```
```diff
diff --git a/src/server/game/Entities/Player/PlayerStorage.cpp b/src/server/game/Entities/Player/PlayerStorage.cpp
index 43f11d14e7..8e0402b7e7 100644
--- a/src/server/game/Entities/Player/PlayerStorage.cpp
+++ b/src/server/game/Entities/Player/PlayerStorage.cpp
@@ -6477,7 +6477,7 @@ void Player::_LoadSpells(PreparedQueryResult result)
             uint32 spellId = fields[0].Get<uint32>();
             uint8 specMask = fields[1].Get<uint8>();
 
-            if (CheckSkillLearnedBySpell(spellId))
+            if (CheckSkillLearnedBySpell(spellId, this))
                 addSpell(spellId, specMask, true);
             else
                 removeSpell(spellId, SPEC_MASK_ALL, false);
```

LuckyLuc96:
> Player::CheckSkillLearnedBySpell(uint32 spellId, Player* player)
>  {
> -    if (!sConfigMgr->GetOption<bool>("ValidateSkillLearnedBySpells", true))
> +    if (!sWorld->getBoolConfig(CONFIG_VALIDATE_SKILL_LEARNED_BY_SPELLS))
> +        return true;

Does this change make it so that the bool that is being checked acquire that value from the hardcoded cache, rather than the worldserver.conf?

> /src/server/game/World/WorldConfig.cpp
>SetConfigValue<bool>(CONFIG_VALIDATE_SKILL_LEARNED_BY_SPELLS, "ValidateSkillLearnedBySpells", true);

Kitzunu:
It makes sure you use the cached version, so you dont call the config everytime

LuckyLuc96:
> It makes sure you use the cached version, so you dont call the config everytime

That makes sense thank you. I was not aware that it updated. I got the changes ready

LuckyLuc96:
But I'd like to recompile and check for my sanity before merge

NoxMax:
```
if (player->IsGameMaster())
         return true;
```
Alright so if I understand the new proposed changes, now the question is, what is the problem that #22370 is actually solving?

Because the issue in #22367 was caused by game master behavior, meanwhile there are pre-existing checks to prevent non-GMs from causing such conflict in-game to begin with. Soooo...

Kitzunu:
To validate that a player cannot have a spell that teaches a skill that's not allowed for their race class combination. No matter how they obtained it. Can't be more clear

NoxMax:
Sure, but how could they obtain it without GM intervention?

If there is such bug somewhere in the game, wouldn't it be better to address that bug itself, rather than apply a broad generalized fix that can interfere with unrelated (but desired) situations?

LuckyLuc96:
I'm getting a core dump when trying to run with latest changes.

>> World database is up-to-date! Containing 1153 new and 622 archived updates.
 
>Loading World Information...
> RealmID:              1
> Version DB world:     ACDB 335.14-dev
> Config: Bad value defined for name 'AutoBalance.LevelScaling.EndGameBoost', going to use 'true' instead
mod-autobalance: `AutoBalance.LevelScaling.EndGameBoost` is enabled in the configuration, but is not currently implemented. No effect.
AutoBalance::OnBeforeConfigLoad: Config loaded. Global config time set to (1753210716728463).
Server Config (Index: 450) is defined but not loaded, unable to continue.
--------------------------------------------------------------------
 
>> ASSERTION FAILED

> Location: /home/luc/azerothcore/src/common/Configuration/ConfigValueCache.h:134
Function: VerifyAllConfigsLoaded
Condition: false
 
---------------------------------------------------------------------
>Segmentation fault (core dumped)

TheSCREWEDSoftware:
> Sure, but how could they obtain it without GM intervention?
> 
> If there is such bug somewhere in the game, wouldn't it be better to address that bug itself, rather than apply a broad generalized fix that can interfere with unrelated (but desired) situations?

Without a gm intervention? a npc via script, npc_trainer or eluna script that gives you the spell as an example, no GM intervention, but you will obtain something you shouldn't (if you havent updated the DBCs)

LuckyLuc96:
I also forgot to add the GM check. Will add that now. But I am at a loss on why I had a core dump on running worldserver and why the checks are failing.

Tereneckla:
unused parameter player bool Player::CheckSkillLearnedBySpell(uint32 spellId, Player* player)

NoxMax:
> but you will obtain something you shouldn't

In this case allow me to define GM intervention as not simply GM commands, but actions that have been taken to modify the server, be it just commands, or in the background as someone with server admin authority.

So with that in mind, could a non-GM player, client side only, get something they shouldn't without such intervention? My opinion is that if there is such in-game bugs, then they should be addressed specifically, not the way #22370 goes about it.

LuckyLuc96:
Reversion is good - compiles, launches, and worldserver.conf config effects reflect in the worldserver.

TheSCREWEDSoftware:
Make sure you update your branch as this: https://github.com/azerothcore/azerothcore-wotlk/pull/22136 was merged which has worldserver.config changes

Nyeriah (approved):

PR #22578, opened 2025-07-27
feat(Core/Account): Save accountFlags only when  LogoutPlayer

PkllonG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


PkllonG:
PR has been running on my server for a day and no players have reported bugs

Takenbacon:
Out of curiosity but what was wrong with the previous use? Seems like now it'll be always updating increasing useless db queries where as before it was only as needed.

PkllonG:
> Out of curiosity but what was wrong with the previous use? Seems like now it'll be always updating increasing useless db queries where as before it was only as needed.

Rewriting to the database every time is wasteful.

sudlud:
But aren't the checks set up so that the flag is not written every time (before this pr)? At least for the death knight flag for example - under normal operation it should only be written once per account total.

Also there's lots of other stuff that gets instantly synced into the database already, I don't really see what's special about the account flags now.

Am I missing something that is constantly changing the account flags during runtime?

PkllonG:
> But aren't the checks set up so that the flag is not written every time (before this pr)? At least for the death knight flag for example - under normal operation it should only be written once per account total.
> 
> Also there's lots of other stuff that gets instantly synced into the database already, I don't really see what's special about the account flags now.
> 
> Am I missing something that is constantly changing the account flags during runtime?

You mean just one flag, but it has many flags, each of which is written once every time? When you have a lot of players online, this increases database usage,

Tereneckla:
Death Knight is the only flag that can be gained automatically. All the others have to be set manually

PkllonG:
If they all oppose it, I will close it.

kissingers:
Indeed, server scalability for large player counts needs consideration. I've heard that servers exceeding 1,000 players often face stability issues and high load.

sudlud:
> Indeed, server scalability for large player counts needs consideration. I've heard that servers exceeding 1,000 players often face stability issues and high load.

this might be a true fact in general, but imo it's completely irrelevant for the topic covered in this pr. account flags don't get changed that often that the process of storing them would need optimization. i rather see more disadvantages in not storing upated account flags immediately.

PkllonG:
Additionally, I have a concern about how to determine the conditions for accountFlags when there are multiple realm servers.
Maybe accountFlags should be stored in the characters database.

heyitsbench:
But...they're flags for accounts.

PkllonG:
> But...they're flags for accounts.

multiple realm ACCOUNT_FLAG_DEATH_KNIGHT_OK
An account can only create one DK. 
What are you going to do about this?

PkllonG:
> realm

The correct state is that each realm can only create one DK.

heyitsbench:
I would think it likely that the assumption that the flag means you can create a DK on a given realm is wrong. Maybe it was a leftover from a previous idea of DKs, maybe it's a flag similar to owning the Wrath expansion, we don't know for sure. What we do know is that these are *account flags*, and having them be specific to each realm is, in my opinion, not a likely scenario.

Nyeriah:
You can create a DK on *any* realm once you have a level 55. Might be mandela effect but that's how I remember it

heyitsbench:
> Might be mandela effect but that's how I remember it

You seem to be correct, at least according to an older version of the [Wowpedia article](https://wowpedia.fandom.com/wiki/Death_knight?oldid=2389943#Unlocking_your_death_knight). I don't believe that was the case for Wrath Classic, but seemed to have been for retail Wrath.

PkllonG:
> You can create a DK on _any_ realm once you have a level 55. Might be mandela effect but that's how I remember it

If that's the case, then my understanding was wrong.

PkllonG:
Sorry, thank you everyone for explaining to me.

Issue #22670, opened 2025-08-11
[Death Knight] Death and Decay interrupts the channeling of Army of the Dead.


Rorschach91:
### Current Behaviour

If a Death Knight uses Army of the Dead while Death and Decay is active, the Army of the Dead channel will be interrupted.

### Expected Behaviour

It shouldn't happen.

### Source

Not needed. 

### Steps to reproduce the problem

You can use a simple dummy.
1. Attack the dummy.
2. Use Death and Decay
3. Use Army of the Dead.

### Extra Notes

The player must be (and must remain) in combat.

With Death and Decay:
https://github.com/user-attachments/assets/d23cbfa6-06c4-4956-836d-96a2ab828999

Without Death and Decay:
https://github.com/user-attachments/assets/199e02dd-e549-4529-9b33-4986be8c9d71

### AC rev. hash/commit

[1d70c6a](https://github.com/azerothcore/azerothcore-wotlk/commit/1d70c6acc3192684ac1665057cfa9bfd4f7aaffa)

### Operating system

Windows 11

PR #22671, opened 2025-08-11
fix(Scripts/Spell): Prevent Death and Decay from interrupting Army of…

EricksOliveira:
… the Dead

When casting Army of the Dead while Death and Decay is active, the Army channel was being interrupted because the core still considered Death and Decay as an active generic spell cast.

This fix clears the CURRENT_GENERIC_SPELL slot after casting Death and Decay, without removing its ground aura, preventing unintended interruptions.

## Changes Proposed:

This PR proposes changes to:
-  [x ] Scripts (bosses, spell scripts, creature scripts).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22670

## SOURCE:

- [x ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [x ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)

## Tests Performed:

This PR has been:
- [x ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [x ] This pull request can be tested by following the reproduction steps provided in the linked issue

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -4574,6 +4574,11 @@ void Spell::finish(bool ok)
         m_caster->AttackStop();
 
     CleanupTargetList();
+
+    if (m_caster && m_spellInfo->IsChanneled())
+    {
+        m_caster->RemoveAurasDueToSpell(m_spellInfo->Id);
+    }
 }
 
 void Spell::WriteCastResultInfo(WorldPacket& data, Player* caster, SpellInfo const* spellInfo, uint8 castCount, SpellCastResult result, SpellCustomErrors customError)


Nyeriah:
This doesn't look right. What causes it to be interrupted in the first place?

EricksOliveira:
To resolve the issue locally and safely, the solution was to use the AfterCast hook in the Death and Decay script to, immediately after casting, free up the CURRENT_GENERIC_SPELL slot held by the core by calling InterruptSpell(CURRENT_GENERIC_SPELL, false).

This call doesn't remove the Death and Decay aura from the ground; it simply frees up the slot that was blocking other channels. This way, when the player attempts to cast Army of the Dead, the core no longer detects the previous active cast and allows the channel to play without interruption.

Takenbacon:
Have you checked why only death and decay seems to keep a lingering reference even after the cast is completed and not other spells of similar types? Seems more like a generic issue that should be addressed than handled in a spellscript.

EricksOliveira:
I only looked for the Spell correction. However, I can check something generic.

EricksOliveira:
I investigated whether this could be a more generic issue in the core’s spell handling logic rather than something exclusive to Death and Decay.
During testing, I compared it with other similar AoE ground-targeted spells (e.g., Blizzard, Rain of Fire, Flamestrike). These spells correctly release their references once the cast is finished and don’t show the same lingering behavior.

In the case of Death and Decay, the reference to the caster/spell context seems to persist beyond the normal cleanup phase. This suggests there’s either an extra condition or missing event in its specific implementation path that prevents the cleanup from triggering as expected.

Because this behavior wasn’t reproducible with other AoE channel or duration-based spells, I handled the fix in the spellscript to ensure it only targets the affected case without unintentionally altering global mechanics.
If we find more spells showing this lingering reference problem, then addressing it at the core level would be the better long-term solution.

Rorschach91:
In any case, I just tested the PR and the bug has not been fixed.

Rorschach91:
Still bugged.

Takenbacon:
Out of curiosity here with how many revisions we are at, are you not debugging and testing locally or just guessing at what the problem might be?

EricksOliveira:
> Por curiosidade, considerando quantas revisões já fizemos, você não está depurando e testando localmente ou apenas tentando adivinhar qual pode ser o problema?

Sorry, I can't test it. My laptop is really bad, I can't compile and test it.

Talthos:
On Chromiecraft (which uses AzerothCore), I've noticed Death and Decay seems to get interrupted by Knockdown type stuns. Makes me wonder if this issue has the same (or a similar) underlying cause as that one.

Nyeriah:
Just a quick question: did you actually investigate any of this, or was it simply generated and handed over by ChatGPT? I only ask because none of the conclusions in this pull request hold up. It would spare your time as well as everyone else’s if we were clear about that from the outset, given that people are spending their time compiling, testing, and reviewing what turns out to be AI by-product that doesn’t actually fix anything.

PR #22678, opened 2025-08-12
fix(Core/Trainer): Fixes display of ranks in Trainer (Horn of Winter and similar)

### LINKED ISSUE #4590: [DK] Core/Spell: Horn of Winter appears trainable
Hello, friends this is my first report on this community.

##### CURRENT BEHAVIOUR:
Death Knight trainers offer Horn of Winter Rank 1 (57330)
for training even if the player already has the spell trained. (visual bug)

it should appear as learned ever.

Create a DK and train all on Trainer

##### EXTRA NOTES:
in Trinitycore They fixed it with this commit
https://github.com/TrinityCore/TrinityCore/commit/baebb2d602ca2f10159ea0039ec3a6d03155c3bd

##### AC HASH/COMMIT:
https://github.com/azerothcore/azerothcore-wotlk/commit/75813d7a23223b9dee07c7b7df9469be05fedfb9

##### OPERATING SYSTEM:
Window 10

##### MODULES:
No

##### OTHER CUSTOMIZATIONS:
No

EricksOliveira:

https://github.com/TrinityCore/TrinityCore/commit/baebb2d602ca2f10159ea0039ec3a6d03155c3bd

Implements proper sending of the SMSG_SEND_UNLEARN_SPELLS packet to correct the display of spell ranks on NPC Trainers.

Before this change, when training the second rank of Horn of Winter, the first rank remained listed in the Trainer until the window was closed and reopened, causing visual inconsistency.

The changes introduce:

Player::SendUnlearnSpells to send to the client spells that should no longer be displayed.

Player::IsUnlearnNeededForSpell to check if a given spell is part of a rank chain and needs to be handled in the packet.

This means the Trainer now correctly hides previous ranks after learning a new one, ensuring expected behavior for chained spells.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/4590

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Rorschach91:
This seems to work.
work

Nyeriah:
This iteration already happens a few lines below https://github.com/azerothcore/azerothcore-wotlk/pull/22678/files#diff-baadebd8cd1117ca48225f316a5ab3fd5fd55b20963394d302341147183db067R3943

So it should be fixed instead imo

Rorschach91:
Edit: it seems that this solution causes more problems.
I retested the PR with a fresh DK.

EricksOliveira:
Thank you all for the feedback

TheSCREWEDSoftware:
@EricksOliveira are you able do a quick resume what does this fix and how to test it?

EricksOliveira:
> @EricksOliveira are you able do a quick resume what does this fix and how to test it?

Previously, the Rank 1 Spell Horn of Winter would reappear after training, but the Rank 2 Spell Horn of winter not appear. This no longer occurs. A requirement was introduced for the Rank 2 Spell Horn to only appear when the Rank 1 Spell Horn is trained.

Teknishun7:
Fresh DK, go see trainer, only rank 1 is there, learn it, rank 2 appears and rank 1 is gone, learn rank 2, rank 1 reappears, if you close trainer window and reopen it, rank 1 is now gone. The PR does seem to fix the issue at hand, least at first glance, you can learn both rank now and everything is in order, but its creating a new bug where rank 1 reappears at the end after learning rank 2. So I would say the issue remains. Here's a demo.
[Video]
https://www.youtube.com/watch?v=hK4h9Jv0c9o

EricksOliveira:
> Fresh DK, go see trainer, only rank 1 is there, learn it, rank 2 appears and rank 1 is gone, learn rank 2, rank 1 reappears, if you close trainer window and reopen it, rank 1 is now gone. The PR does seem to fix the issue at hand, least at first glance, you can learn both rank now and everything is in order, but its creating a new bug where rank 1 reappears at the end after learning rank 2. So I would say the issue remains. Here's a demo. [Video] https://www.youtube.com/watch?v=hK4h9Jv0c9o

Thanks for the feedback. The fix was only partially implemented, and I was still investigating. But I managed to uncover the real issue. The fix has been submitted.

Teknishun7:
All fixed now. Working as intended! Thanks for the PR.

src/server/game/Handlers/NPCHandler.cpp
```diff
@@ -264,6 +264,10 @@ void WorldSession::HandleTrainerBuySpellOpcode(WorldPacket& recvData)
     else
         _player->learnSpell(spellId);
 
+    // Send "unlearn" packet only if necessary
```
heyitsbench: Comment redundant.

src/server/game/Entities/Player/Player.h
```diff
@@ -1656,6 +1656,9 @@ class Player : public Unit, public GridObject<Player>
     void UpdateNextMailTimeAndUnreads();
     void AddNewMailDeliverTime(time_t deliver_time);
 
+    void SendUnlearnSpells();
+    static bool IsUnlearnSpellsPacketNeededForSpell(uint32 spellId);
```
heyitsbench:
```suggestion
    static bool IsUnlearnNeededForSpell(uint32 spellId);
```
Not a fan of the name being as long as it is, feels unnecessary.

heyitsbench:
Please attach the commit and co-author where this came from.

src/server/game/Handlers/NPCHandler.cpp
EricksOliveira: Ok

src/server/game/Entities/Player/Player.h
EricksOliveira: Ok

sogladev:
I tested the current implementation, and the bug sometimes remains after a fresh login:
- open trainer
- learn a skill
- horn of winter shows (bug), or shows briefly and disappears.

When I compared with TC's 335 branch, this PR seems to be missing some logic for `IsUnlearnSpellsPacketNeededForSpell`. I tried continuing from this PR in https://github.com/azerothcore/azerothcore-wotlk/commit/f3d6ead637aaf8a9368527b6aed7693fbe310c30 to match TC better, but I wasn't able to fix the issue.

I found that the challenge here is knowing under which conditions TC sends the spell unlearn packets. TC only sends them on login and in `bool Player::AddSpell`](https://github.com/TrinityCore/TrinityCore/blob/6057f339229742476011a2ec046f327f08aac82c/src/server/game/Entities/Player/Player.cpp#L3019).


@EricksOliveira Would you be interested in continuing this? I suggest these directions:

1. Try to continue the current PR by also sending unlearn on login to fix the issue described above
2. Try to replicate TC by not sending unlearn in `HandleTrainerBuySpellOpcode`, but by sending unlearn packets both on login and in AC's `bool Player::addSpell`

EricksOliveira:
New fix. Awaiting testing.

Nyeriah:
This PR requires testing since latest changes before further review.

pangolp:
The error is clear. It doesn't only happen with the Death Knight (DK); it also happens now with the Druid, who, without having the Feral quest, is shown by the trainer using spells that can only be obtained in that form. The DK issue is the most common error, and anyone who has created a character knows it exists. I don't know if this is the solution or not, nor what consequences this implementation might have.

If anyone has a series of steps to test and leave a review... as far as I remember, it's confirmed for the Death Knight, and it recently happened to me with the Druid as well. But I don't know if it occurs with other classes or trainers.

In a configuration with x1 rates, these problems may not be seen because the spells are learned more slowly, but if the rates increase and several levels are gained together, that's where these errors appear and become visible.

sudlud:
Thanks for working on this. The fix is on the right track but a few things are blocking it.

**1. The login fix is overwritten.** `SendInitialSpells()` now sends the populated `SMSG_SEND_UNLEARN_SPELLS`, but `SendInitialPacketsBeforeAddToMap()` (`Player.cpp:11584-11586`) still sends an empty `SMSG_SEND_UNLEARN_SPELLS` right after it, which clobbers the populated one. This is why the bug still shows after a fresh login. The TC commit fixes exactly this by replacing that empty packet with `SendUnlearnSpells()`, rather than adding the call inside `SendInitialSpells()`.

**2. Missing `SupercededBySpell` check.** TC only includes a spell in the packet when none of its `SkillLineAbility` entries have `SupercededBySpell` set (the client already hides those ranks on its own). Neither `IsUnlearnNeededForSpell` nor the `SendUnlearnSpells` loop checks `SkillLineAbilityEntry::SupercededBySpell`, so superseded ranks get pushed redundantly and flicker. This matches the "shows briefly and disappears" report.

**3. Attribution.** This is a port of TC commit `baebb2d` by Shauren and must credit the original author properly. The current `Co-authored-by:` trailer is malformed (`Co-authored-by: Name [@Shauren](mailto:...)`) and needs to be a valid trailer in the commit itself, e.g. `Co-authored-by: Shauren <shauren.trinity@gmail.com>`, per the project's cherry-pick/attribution policy.

Could you update the PR to address all three? Items 1 and 2 are needed for the fix to actually work on the login path; item 3 is required for the PR to be accepted.

sogladev (approved):
✅ I wasn't able to reproduce the issue anymore by doing https://github.com/azerothcore/azerothcore-wotlk/pull/22678#issuecomment-3338708807c

Issue #22675, opened 2025-08-11
Crash: Server crashes randomly because of possible 'overload' issues


Ramvaris:
### Steps to Reproduce

The server crashes randomly. Looking at the crash reports showed something interesting - pointing back to the AzerothCore itself even though I am running Azerothcore-wotlk-with-NPCBots. If the server load is high enough then multiple GridRemovals can happen at the same tick - so it crashes.

It didn't happen before and started suddenly after shutting it down to get some food (only 2 people playing here - but 120 Bots in the world) and the BOT-Build is a bit behind - but I didn't find anything related here. So I guess I open a ticket - just to be sure.

It may be that this does NOT happen on NON-Bot servers because of the lesser load and therefore less often appearing situation?

The fixes I made to check this out were:

In the GridNotifiers.h:

```
void operator()(Creature* u)
{
    // Prevents Chain Reactions to mobs - causing a cascade error
    if (u->IsInCombat())
        return;
        // Rest of the Code
```

In the Map.cpp
```
template<class T>
void Map::RemoveFromMap(T* obj, bool remove)
{
    // Previous Code
    // This should prevent double removals and therefore crashes under heavy-load
    if (!obj->IsInGrid())
        return;
    // Rest of the Code
```

Until now it is running stable - but it could very well be that I am completly wrong here.

### Logs and Screenshots

[730c61714dda+_worldserver.exe_\[10-8_23-11-41\].txt](https://github.com/user-attachments/files/21719951/730c61714dda%2B_worldserver.exe_.10-8_23-11-41.txt)
[730c61714dda+_worldserver.exe_\[10-8_23-1-5\].txt](https://github.com/user-attachments/files/21719947/730c61714dda%2B_worldserver.exe_.10-8_23-1-5.txt)
[730c61714dda+_worldserver.exe_\[10-8_22-42-45\].txt](https://github.com/user-attachments/files/21719950/730c61714dda%2B_worldserver.exe_.10-8_22-42-45.txt)
[730c61714dda+_worldserver.exe_\[10-8_22-34-38\].txt](https://github.com/user-attachments/files/21719955/730c61714dda%2B_worldserver.exe_.10-8_22-34-38.txt)
[730c61714dda+_worldserver.exe_\[10-8_22-30-31\].txt](https://github.com/user-attachments/files/21719948/730c61714dda%2B_worldserver.exe_.10-8_22-30-31.txt)
[...]

PR #22683, opened 2025-08-13
Fix: Add safeguards against recursion and race condition crashes

Ramvaris:
This commit introduces three key stability fixes: 1. A re-entrancy guard in SmartScript::ProcessEvent to prevent stack overflows from recursive event calls. 2. An idempotency check in Map::RemoveFromMap to prevent assertion failures from double-removal race conditions. 3. An in-combat check in GridNotifiers.h to prevent call-for-help chainreactions.

## Changes Proposed:
This PR proposes changes to:

## Issues Addressed:
https://github.com/azerothcore/azerothcore-wotlk/issues/22675

## Tests Performed:
This PR has been:

## How to Test the Changes:

1. Compile
2. Start the server and play
3. Check your Tracelogs for far less SmartScript calls and crashes

None

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -43,6 +43,15 @@
 //  see: https://github.com/azerothcore/azerothcore-wotlk/issues/9766
 #include "GridNotifiersImpl.h"
 
+// Definition of some variables / structs for the recursion railguard
+thread_local bool g_smartScriptProcessing = false;
+
+struct SmartScriptProcessorGuard
+{
+    SmartScriptProcessorGuard() { g_smartScriptProcessing = true; }
+    ~SmartScriptProcessorGuard() { g_smartScriptProcessing = false; }
+};
+
 SmartScript::SmartScript()
 {
     go = nullptr;
@@ -4037,6 +4046,12 @@ void SmartScript::GetWorldObjectsInDist(ObjectVector& targets, float dist) const
 
 void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, uint32 var1, bool bvar, SpellInfo const* spell, GameObject* gob)
 {
+	// A recursion protection
+	if (g_smartScriptProcessing)
+        return;
+
+    SmartScriptProcessorGuard guard;
+	
     if (!e.active && e.GetEventType() != SMART_EVENT_LINK)
         return;
 
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -4046,10 +4046,10 @@ void SmartScript::GetWorldObjectsInDist(ObjectVector& targets, float dist) const
 
 void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, uint32 var1, bool bvar, SpellInfo const* spell, GameObject* gob)
 {
-	// A recursion protection
-	if (g_smartScriptProcessing)
+    // A recursion protection
+    if (g_smartScriptProcessing)
         return;
-
+    
     SmartScriptProcessorGuard guard;
 	
     if (!e.active && e.GetEventType() != SMART_EVENT_LINK)
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -4049,9 +4049,9 @@ void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, ui
     // A recursion protection
     if (g_smartScriptProcessing)
         return;
-    
+
     SmartScriptProcessorGuard guard;
-	
+
     if (!e.active && e.GetEventType() != SMART_EVENT_LINK)
         return;
 
--- a/src/server/game/AI/SmartScripts/SmartScript.cpp
+++ b/src/server/game/AI/SmartScripts/SmartScript.cpp
@@ -43,13 +43,40 @@
 //  see: https://github.com/azerothcore/azerothcore-wotlk/issues/9766
 #include "GridNotifiersImpl.h"
 
-// Definition of some variables / structs for the recursion railguard
-thread_local bool g_smartScriptProcessing = false;
+// Maximum allowed nested calls of any kind to prevent a stack overflow
+#define MAX_GLOBAL_RECURSION_DEPTH 20
+// Maximum allowed nested calls for the same event to prevent simple infinite loops
+#define MAX_PER_EVENT_RECURSION 3
+
+// Global depth counter variable
+thread_local uint32_t g_smartScriptGlobalDepth = 0;
+
+// The counter map (per event)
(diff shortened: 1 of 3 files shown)


src/server/game/Maps/Map.cpp
```diff
@@ -651,6 +651,10 @@ void Map::RemoveFromMap(T* obj, bool remove)
 {
     obj->RemoveFromWorld();
 
+    // This should prevent double removals and therefore crashes under heavy-load
```
Takenbacon, line 654: Do you have a crashlog for this one? I suspect this is because of NPCBots which is known to do some dumb unsafe stuff. The goal is to always fix the underlying issue and not mask it by early returns - if we are trying to double remove an object from the map something has gone horribly wrong elsewhere. For reference, the mentioned assert exists for a reason, to safeguard against greater issues and expose potential architectural problems, by early returning you are defeating the job of that sanity-check.

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -4037,6 +4046,12 @@ void SmartScript::GetWorldObjectsInDist(ObjectVector& targets, float dist) const
 
 void SmartScript::ProcessEvent(SmartScriptHolder& e, Unit* unit, uint32 var0, uint32 var1, bool bvar, SpellInfo const* spell, GameObject* gob)
 {
+    // A recursion protection
```
Takenbacon: Not entirely sure how I feel about this one tbh, this might actually break some stuff. While it's probably not ideal I really wouldn't be surprised if certain smartAI actions end up calling others. The problem is when it gets stuck on a infi loop but inherently a little bit of recursion may not be technically "wrong".

src/server/game/Maps/Map.cpp
Ramvaris: The crash happened many times before this fix but only after a certain update point on the AzerothCore. But yeah, I surely would NOT rule the NPCBots out as culprits - especially since Trickerer made a patch some days ago because of heavy crashing with the last AzerothCore update. A patch that seems connected to this.

This was his commit: https://github.com/trickerer/AzerothCore-wotlk-with-NPCBots/commit/730c61714dda150469b19b9bb221ea0bfe09a84c

Here is one crash-log you requested as example - I have multiple - those crashes were gone after the map.cpp 'fix':
[730c61714dda+_worldserver.exe_[10-8_20-28-20].txt](https://github.com/user-attachments/files/21772465/730c61714dda%2B_worldserver.exe_.10-8_20-28-20.txt)

But completly beside that that - wouldn't it be better to just write an ERROR and catching the 'exception' instead of letting it crash the server? I mean it's about a GridRemoval. I can't think of any bigger, permanent events that could severly corrupt the server if you don't immediately stop it. Most of the potential problems would solve by reloading the grid or restarting the server - as those things usually aren't even written in the database? 

Edit: Sorry if my theory is not fitting, but I don't know the AzerothCore very well yet. I only modify the code here and there locally for my personal purposes. Like letting me fly by changing the SpellInfo.cpp check (adding mapID 0 and 1 as exception). So nothing that big as long as Eluna also does the trick.

src/server/game/AI/SmartScripts/SmartScript.cpp
Ramvaris: I didn't experience ANY problems with the new recursion protection. On the contrary. The mobs seem to react WAY faster in chaining their actions - from talking to casting. At least on the official stuff. As far as I am aware SmartScripts should run sequential and not recursive anyways... at least when they are done in the correct way.

The NPCBots - I know, not part of this repo but relevant nowadays - seem to be a bit more... sluggy. They tend to stand around for a moment re-orienting themselves before continuing to act. I guess that hints at a problem based on Trickerers design of those. Maybe way to much SmartScripts running here from them or the environment reacting to them. I have 120 of them running around in my world to let it seem more 'lively'. Who knows. But even if it is that way it only shows the inherent weakness of the unprotected function... but on the other hand those crashes were not present before the forementioned 'update time'.

The crashes I had after 'fixing' the map.cpp were those (another log) - fixed by this code adaption (no crashes over 3 days):
[730c61714dda+_worldserver.exe_[11-8_19-17-51].txt](https://github.com/user-attachments/files/21772619/730c61714dda%2B_worldserver.exe_.11-8_19-17-51.txt)

Ramvaris: Little addendum: There IS a problem. The event in Uldaman for example where the Titan comes out didn't work. I will rewrite the guardrail to accept recursions - but only to a specific depth and not with the same parameters over and over again. That should keep all the intended recursions in a working state.

Ramvaris: Final: I added (with the help of Gemini Pro AI - as I am more of a Python not a C++ guy - sorry ;)) a guard that just protects scripts from being run to often. It should also give an error IF that happens so you can pinpoint problematic scripts more easily. That brought up my personal endless script problem nearly instantly. 

Example Entry from my NPCBots log: 2025-08-14 14:47:06 SmartScript: Terminating script processing - Exceeded max per-event recursion of 3. Infinite loop detected for script Entry/GUID: 18944, SourceType: 0, Event: 12.

So it works. The Ironaya Uldaman event chain also worked perfectly fine again.

I hope that's a better solution!

src/server/game/Maps/Map.cpp
Takenbacon: NPCbot issues must be fixed in NPCbots code, not upstream.

As for your question, asserts are powerful debugging tool to sanity-check the state of the server, a production server should theoretically have asserts disabled anyways. It is bad practice to simply return to bypass an assert check unless you are 100% confident in all code paths and behavior. In this case, you can't be certain by returning you aren't now going to do a double delete or leak memory later on in execution. This also follows [Offensive programming](https://en.wikipedia.org/wiki/Offensive_programming) methodology.

Your crashlog is useless without debugging symbols.

src/server/game/AI/SmartScripts/SmartScript.cpp
Takenbacon: A global guard isn't a terrible idea but I do have a little hesitation still. I believe it would be better to fix the underlying source (which seems to be one NPC in particular misbehaving).

Regardless, if it is to remain it needs at minimum a couple changes: 

- The map needs to be cleared, the current use and key will grow the map exponentially to hundreds of thousands of elements leading to a significant resource leak and slowdown.
- Use std::unordered_map over map
- Use AC code style&naming conventions (there's a wiki page somewhere)

src/server/game/Maps/Map.cpp
Ramvaris: Urghs. Yeah I just noticed. Forgot to copy the symbol file while testing. I guess I will just remove the map.cpp check then tomorrow when I am at my PC and just let it live at my home brew system to keep playing. I'd rather have some memory leaks than not being able to play as this is kind of a neccesary therapy for me right now - no joke. Just thought its a good idea to share it with others :). But you are right here. It probably is a NPCBot based problem, so maybe Trickerer will fix it in his release somehow.

Ramvaris:
Closed due to seemingly a NPCBot problem and not needed here.
Thanks for all the good advices, but I will follow the more 'defensive variant'.
Things need to be debugged, that's right but on a live system to keep it running is also important.
I will put the fixes in my own fork of NPCBots that I will create.
For people that need them.

PR #22686, opened 2025-08-13
fix(Core/GroupHandler): Adjust maximum value for rolls.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Roll.

- [ ] Not sure about the 32768 urand comment, maybe an older version of it only supported input values that size? 1,000,000 worked fine in my testing, sooo.


heyitsbench:
Actually could maybe tie this into a config if preferred, with default value being 1,000,000.

TheSCREWEDSoftware:
Wouldn't be a bad a idea to configurable

heyitsbench:
Adjusted, interestingly the 3.3.5a client seems to do its own clamp to 1,000,000. Setting the config value any higher still did not allow any higher rolls in my testing.

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -3141,6 +3141,16 @@ LeaveGroupOnLogout.Enabled = 0
 
 Group.Raid.LevelRestriction = 10
 
+#
+#    Group.RandomRollMaximum
+#
+#     The maximum value for use with the client '/roll' command.
```
TheSCREWEDSoftware, line 3147: I'm not the greatest with phrasing, possibility of refer /roll $Group.RandomRollMaximum and/or /roll $minValue $Group.RandomRollMaximum

TheSCREWEDSoftware:
> Adjusted, interestingly the 3.3.5a client seems to do its own clamp to 1,000,000. Setting the config value any higher still did not allow any higher rolls in my testing.

At least you can roll over 9'000

src/server/apps/worldserver/worldserver.conf.dist
heyitsbench: Might just be me, but I don't really like the idea of sticking command syntax (especially built in client commands) in the config descriptions.

TheSCREWEDSoftware: I mentioned because not everyone know that roll can take 2 parametres, but i understand your side and might be etter not stick client command syntxa into the worldserver config

Issue #22048, opened 2025-05-04
ICC Lich King: Vile spirits target closest pet/totem to attack when spawn


tabooo:
### Current Behaviour

Vile spirits target closest pet/totem to attack when spawned

### Expected Behaviour

Vile spirits should not attack nearest pets/totems when spawn

### Steps to reproduce the problem

1. Go to lich king battle (.go creature id 36597)
2. Have pet/totem
3. Wait for Vile Spirits to spawn (Phase 4, after 20 seconds of second quake)

### AC rev. hash/commit

AzerothCore rev. 8358e17ac179+ 2025-04-30 20:54:37 -0300 (master branch) (Win64, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 2.
Server uptime: 1 day(s) 2 hour(s) 55 minute(s) 4 second(s)
Update time diff: 1ms. Last 500 diffs summary:
|- Mean: 2ms
|- Median: 2ms
|- Percentiles (95, 99, max): 6ms, 6ms, 8ms
Using SSL version: OpenSSL 3.3.2 3 Sep 2024 (library: OpenSSL 3.3.2 3 Sep 2024)
Using Boost version: 1.81.0
Using CMake version: 3.30.3
Using MySQL version: 80402
Found MySQL Executable: C:/Program Files/MySQL/MySQL Server 8.4/bin/mysql.exe
Compiled on: Windows 10.0.19045
Worldserver listening connections on port 8090
Realmlist (Realm Id: 3) configured in port 8090
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in D:/wowserver/server-test/data/maps. Total size: 291014951 bytes
vmaps directory located in D:/wowserver/server-test/data/vmaps. Total size: 658130721 bytes
mmaps directory located in D:/wowserver/server-test/data/mmaps. Total size: 2192910844 bytes
Default DBC locale: enUS.
All available DBC locales: enUS
Using World DB: ACDB 335.13-dev
Latest LoginDatabase update: acore_cms_subscriptions.sql
Latest CharacterDatabase update: zone_difficulty_completion_logs.sql
Latest WorldDatabase update: zone_difficulty_spelloverrides.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-1v1-arena
[...]

PR #22693, opened 2025-08-15
fix(Scripts/ICC): Prevent Vile Spirits from attacking pets/totems whe…

EricksOliveira:
…n selecting initial target

This PR fixes an issue where Vile Spirits (and potentially other summoned aggressive creatures) would incorrectly target pets, guardians, and totems instead of players.

The root cause was in CreatureAI::MoveInLineOfSight, which considered any Unit as a valid target.

Solution: Added a check to ensure only players can be targeted in MoveInLineOfSight.

Prevents Vile Spirits from chasing non-player units.

Preserves correct encounter behavior on The Lich King fight.

Avoids unintended attacks on pets, totems, or guardians globally.

## Changes Proposed:

This PR proposes changes to:
-  [x ] Scripts (bosses, spell scripts, creature scripts).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22048

## SOURCE:

- [x ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [x ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)

## Tests Performed:

This PR has been:
- [x ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [x ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
@@ -528,8 +528,8 @@ public:
 
     bool Execute(uint64 /*time*/, uint32 /*diff*/) override
     {
-        _owner->CastSpell(_owner, SPELL_VILE_SPIRIT_MOVE_SEARCH, true);
         _owner->SetReactState(REACT_AGGRESSIVE);
+        _owner->CastSpell(_owner, SPELL_VILE_SPIRIT_MOVE_SEARCH, true);
         _owner->CastSpell((Unit*)nullptr, SPELL_VILE_SPIRIT_DAMAGE_SEARCH, true);
         return true;
     }


Nyeriah:
The spell has the attribute to only hit players

EricksOliveira:
> The spell has the attribute to only hit players

Although the spell used by Vile Spirits has a player-only target, recent changes to the grid system allowed pets and totems to be included in the target list before final selection. This update adds explicit filtering to spell_the_lich_king_vile_spirit_move_target_search to only target players, preventing the spirits from fixating on pets/totems when summoned.

Nyeriah:
The issue was first flagged on the 4th of May, and there’d been no grid updates up to that point. What’s more, this very problem has cropped up in previous revisions, so it’s rather doubtful the grid updates are the culprit. In any case, if the fault lies with the searcher, then this addition to the spellscript really doesn’t belong there.

EricksOliveira:
You’re right, thanks for pointing that out.
If the underlying issue is indeed in the target searcher not respecting the spell’s attributes, then fixing it there would be the proper approach instead of patching it in the spell script.
My intention here was to provide a safeguard so that Vile Spirits behave correctly during the encounter, but I agree it would be better to fix the root cause in the target selection logic.
Do you think the best approach would be to enforce the TARGET_FLAG_PLAYER check directly in the target searcher?

Nyeriah:
Question: if you are absoltely outside attack range for any of the spirits and just let them active on their own through the spell sequence, do they still go for anything but players? The little investigation I've done about this in the past has led me to find there was nothing wrong with the spell itself. This can be easily verified by checking the target selector list (that you previously filtered), it only contains players.

EricksOliveira:
> Question: if you are absoltely outside attack range for any of the spirits and just let them active on their own through the spell sequence, do they still go for anything but players? The little investigation I've done about this in the past has led me to find there was nothing wrong with the spell itself. This can be easily verified by checking the target selector list (that you previously filtered), it only contains players.

Yes, the spell itself is fine and the target selector only includes players, as you noted. The issue occurs because the Vile Spirits enter REACT_AGGRESSIVE immediately upon spawning, before the SPELL_VILE_SPIRIT_MOVE_SEARCH actually runs. In that small window, the AI may select any nearby valid target, including pets and totems, which causes them to attack these unintended units.

The fix ensures that the spell is cast first to select a player target, and only after that the Vile Spirit enters REACT_AGGRESSIVE, preventing it from attacking anything other than players.

Teknishun7:
Let me know if this needs testing.

EricksOliveira:
I tested it and it worked well for me.

GrenderG:
Wouldn't this change prevent ANY pet from aggroing as they should? :thinking:

Nyeriah:
Looks like more ChatGPT to me, this breaks every existing creature, let us know once there is a proper fix attempt and we will reopen it

PR #22719, opened 2025-08-23
refactor(Core/Packets): Rewrite various query packets to modern class.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## Tests Performed:

This PR has been:
- [X]Tested in-game by the author.

- [X] I honestly would like certain pieces to have some sort of boolean instead of things like `uint8(0)`, since makes more sense to a reader, but not sure if it'd have the same bit representation when sent as a packet, so also not sure if the client would accept it and work with it properly.


src/server/game/Server/WorldSession.h
```diff
@@ -661,9 +668,9 @@ class WorldSession
     void HandleGameObjectUseOpcode(WorldPacket& recPacket);
     void HandleGameobjectReportUse(WorldPacket& recvPacket);
 
-    void HandleNameQueryOpcode(WorldPacket& recvPacket);
+    void HandleNameQueryOpcode(WorldPackets::Query::NameQuery& packet);
 
-    void HandleQueryTimeOpcode(WorldPacket& recvPacket);
+    void HandleQueryTimeOpcode(WorldPackets::Query::TimeQuery& packet);
```
sudlud: Querytime? Instead of timequery, so it matches the function name and also the packet define

src/server/game/Handlers/QueryHandler.cpp
```diff
@@ -32,50 +33,44 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
...
-    data << uint8(0);                               // name known
-    data << playerData->Name;                       // played name
-    data << uint8(0);                               // realm name - only set for cross realm interaction (such as Battlegrounds)
-    data << uint8(player ? player->getRace() : playerData->Race);
-    data << uint8(playerData->Sex);
-    data << uint8(playerData->Class);
+    nameQueryResponse.NameUnknown = uint8(0);
+    nameQueryResponse.Name = playerData->Name;
+    nameQueryResponse.RealmName = uint8(0); // Only set for cross realm interaction (such as Battlegrounds)
+    nameQueryResponse.Race = uint8(player ? player->getRace() : playerData->Race);
```
sudlud: The castings when filling the member variables can be removed imo. The target data type is already known because of the member variable data type

src/server/game/Handlers/QueryHandler.cpp
```diff
@@ -32,50 +33,44 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
...
     Player* player = ObjectAccessor::FindConnectedPlayer(guid);
 
-    data << uint8(0);                               // name known
-    data << playerData->Name;                       // played name
-    data << uint8(0);                               // realm name - only set for cross realm interaction (such as Battlegrounds)
-    data << uint8(player ? player->getRace() : playerData->Race);
-    data << uint8(playerData->Sex);
-    data << uint8(playerData->Class);
+    nameQueryResponse.NameUnknown = uint8(0);
+    nameQueryResponse.Name = playerData->Name;
```
sudlud, line 48: This introduces an additional string copy, right?

src/server/game/Server/WorldSession.h
heyitsbench: I'd be more in favor of renaming the function to be in line with the NameQuery function above.

src/server/game/Handlers/QueryHandler.cpp
heyitsbench: Should be corrected.

src/server/game/Handlers/QueryHandler.cpp
heyitsbench: Believe so, is there an issue with that?

heyitsbench:
Failing check seems to be due to [mod-war-effort](https://github.com/azerothcore/mod-war-effort), unrelated to this PR.

src/server/game/Handlers/QueryHandler.cpp
Takenbacon: > Believe so, is there an issue with that?

It's always going to be preferable to avoid copies, you can probably pass it as a string view

heyitsbench: That's fair, I've replaced the datatype in the packet structure, retested and still seems to work fine for the client. I assume this is a suitable correction?

src/server/game/Server/WorldSession.h
sudlud: Yeah I think that's what I meant, having the same naming like the name query

src/server/game/Handlers/QueryHandler.cpp
Takenbacon: > That's fair, I've replaced the datatype in the packet structure, retested and still seems to work fine for the client. I assume this is a suitable correction?

Yeah should be fine. The only stipulation with string views is they're basically a pointer to the string so there's lifetime considerations involved in the use, e.g. you'd have a problem using a string view on a temporary string that goes out of scope from where the string view is used. As long as the original string is valid in the same context it's safe and more efficient

src/server/game/Handlers/QueryHandler.cpp
```diff
@@ -32,62 +33,55 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
...
+        nameQueryResponse.NameUnknown = true;
+        SendPacket(nameQueryResponse.Write());
         return;
     }
 
     Player* player = ObjectAccessor::FindConnectedPlayer(guid);
 
-    data << uint8(0);                               // name known
-    data << playerData->Name;                       // played name
-    data << uint8(0);                               // realm name - only set for cross realm interaction (such as Battlegrounds)
```
heyitsbench, line 48: Down the road, this should be added back with something like this:

```cpp
if (player->HasSameNativeRealmAs(queryingPlayer))
    nameQueryResponse.RealmName = player->GetNativeRealmName();
```

But AC has no cross-realm support atm afaik, so these methods don't functionally exist. The client does properly show the (*) and the realm name on the player when this field is set, so nothing else needs doing it looks like to support that piece.

For now, I've just set the RealmName to be "" in the packet definition.

TheSCREWEDSoftware: Should this ^ (the above said) be added into the QueryHandler as TODO comment?

heyitsbench: I don't personally think so, I feel that the comment in QueryPackets.h conveys the nature of the field and what exactly it needs to be implemented.

src/server/game/Handlers/QueryHandler.cpp
```diff
@@ -32,62 +33,55 @@ void WorldSession::SendNameQueryOpcode(ObjectGuid guid)
...
-    data << uint8(player ? player->getRace() : playerData->Race);
-    data << uint8(playerData->Sex);
-    data << uint8(playerData->Class);
+    nameQueryResponse.NameUnknown = false;
+    nameQueryResponse.Name = playerData->Name;
+    nameQueryResponse.Race = player ? player->getRace() : playerData->Race;
+    nameQueryResponse.Sex = player ? player->getGender() : playerData->Sex;
+    nameQueryResponse.Class = player ? player->getClass() : playerData->Class;
```
sudlud, line 51: This is a change to the previous implementation, on purpose?

heyitsbench: Yes, done for consistency across the attribute fields.

PR #22707, opened 2025-08-19
fix(SERVER/Main): Implementation for version param

SAS2000:
**Implementation for -v/--version parameters were missing**

IssueId: #22706

On branch fix-issue-22706
  Changes to be committed:
	modified:   src/server/apps/authserver/Main.cpp
	modified:   src/server/apps/worldserver/Main.cpp

## Changes Proposed:
This PR proposes changes to:
**-  [ ] Core (units, players, creatures, game systems).**
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:
https://github.com/azerothcore/azerothcore-wotlk/issues/22706

## Tests Performed:
This PR has been:
**- [ ] Tested in-game by the author.**
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:
Built and start any of the servers (auth/world) with -v commandline parameter.

**- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue**

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #22727, opened 2025-08-24
fix(Core/Commands): Waypoint show on/off bug

### LINKED ISSUE #22726: Waypoint show on/off implementation try to GetCreature(guid) by dbGuid -> can never delete the waypoints from the map
### Current Behaviour

I report two errors related to "_.wp show_":

**1.** When you write "_.wp show off_" according to the description and according to the implementation the intention seems that it should delete all the Waypoints so we do not need any target or guid. But if we ommit both we now got an error: "_You should select a creature_"

**2.** When I create waypoints with "_.wp show on_" on a target, if I later try to remove the waypoints I got an error that "_Warning: Could not delete WP from the world with ID: {}_" for every waypoint. This is stable and we never can delete the waypoints (only a server restart can eleminate them)

The same problem occurs when you try to ".wp show on" twice on a creature because "show on" first tries to remove the old waypoints the same (wrong way) as "show off".

### Expected Behaviour

**1.** I should delete all waypoints without selecting a target or specifying a path_id

**2.** When I add waypoints with "_.wp show on_" later I should be able to delete them using "**.wp show off**" without any error.

### Source

I created a fix for both problems.

But some notes on the error:
When you add the waypoints of a creature to the game with ".wp show on" it creates the waypoint creature, save it to the database and add it to the map. It also saves the current dbGuid(spawnId) to the waypoint_data table. 
Adding a map also calls: GetMap()->GetObjectsStore().Insert<Creature>(GetGUID(), this);

When we then try to delete the waypoints with ".wp show off" it cannot find the waypoints in the map because we try to find them by dbGuid (the way previously saved to database):
Creature* creature = handler->GetSession()->GetPlayer()->GetMap()->GetCreature(ObjectGuid::Create<HighGuid::Unit>(VISUAL_WAYPOINT, guid));
Here guid is a dbGuid (spawnId) and not the ObjectGuid for units.

1. .tele brill -> run to the bridge -> target Gordo -> .wp show on
2. Untarget Gordo -> .wp show off -> you get the error "You should select a creature"
3. Retarget Gordo -> .wp show off -> you get 30 errors: "Warning: Could not delete WP from the world with ID: {}"

### Extra Notes

I created a fix just want to create this bug report to use its number for the branch name.

### AC rev. hash/commit

AzerothCore rev. 9ed31bd63e56+ 2025-08-24 14:50:16 +0200 (master branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 11 x64

### Custom changes or Modules

_No response_

SAS2000:
Current implementation cannot delete waypoints

Closes AzerothCore issue #22707

On branch fix-issue-22726
  Changes to be committed:
    modified:   src/server/scripts/Commands/cs_wp.cpp

## Changes Proposed:

This PR proposes changes to:
**-  [ ] Core (units, players, creatures, game systems).**

## Issues Addressed:

- Closes #22726

## Tests Performed:

This PR has been:
**- [ ] Tested in-game by the author.**

## How to Test the Changes:

**- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue**

### DIFF
--- a/src/server/scripts/Commands/cs_wp.cpp
+++ b/src/server/scripts/Commands/cs_wp.cpp
@@ -49,6 +49,57 @@ public:
         };
         return commandTable;
     }
+
+    /**
+    * Remove all waypoint from a query result.
+    *
+    * @param result: If the result not empty it should constains the creature database guids in the first column in all rows.
+    *
+    */
+    static void RemoveWaypointsFromQueryResult(ChatHandler* handler, const PreparedQueryResult& result)
+    {
+        if (result)
+        {
+            bool hasError = false;
+
+            do
+            {
+                const Field* fields = result->Fetch();
+                const ObjectGuid::LowType guid = fields[0].Get<uint32>();
+                const auto creatureRange = handler->GetSession()->GetPlayer()->GetMap()->GetCreatureBySpawnIdStore().equal_range(guid);
+                if (creatureRange.first != creatureRange.second)
+                {
+                    std::for_each(
+                        creatureRange.first,
+                        creatureRange.second,
+                        [&](auto& guidCreaturePair) {
+                            Creature* creature = guidCreaturePair.second;
+                            creature->CombatStop();
+                            creature->DeleteFromDB();
+                            creature->AddObjectToRemoveList();
+                        }
+                    );
+                }
+                else
+                {
+                    handler->PSendSysMessage(LANG_WAYPOINT_NOTREMOVED, guid);
+                    hasError = true;
+
+                    WorldDatabasePreparedStatement* stmt = WorldDatabase.GetPreparedStatement(WORLD_DEL_CREATURE);
+                    stmt->SetData(0, guid);
+                    WorldDatabase.Execute(stmt);
+                }
+            } while (result->NextRow());
+
+            if (hasError)
+            {
+                handler->PSendSysMessage(LANG_WAYPOINT_TOOFAR1);
+                handler->PSendSysMessage(LANG_WAYPOINT_TOOFAR2);
+                handler->PSendSysMessage(LANG_WAYPOINT_TOOFAR3);
+            }
+        }
+    }
+
     /**
     * Add a waypoint to a creature.
     *
@@ -752,41 +803,45 @@ public:
         uint32 pathid = 0;
         Creature* target = handler->getSelectedCreature();
 
-        // Did player provide a PathID?
+        const std::string show = show_str;
 
-        if (!guid_str)
+        // If command equals "off" we do not need a guid or target, just delete all WPs
+        if (show != "off")
         {
-            // No PathID provided
-            // -> Player must have selected a creature
-
-            if (!target)
+            // Did player provide a PathID?
+            if (!guid_str)
             {
-                handler->SendErrorMessage(LANG_SELECT_CREATURE);
-                return false;
+                // No PathID provided
+                // -> Player must have selected a creature
+                if (!target)
+                {
+                    handler->SendErrorMessage(LANG_SELECT_CREATURE);
+                    return false;
+                }
+
+                pathid = target->GetWaypointPath();
             }
+            else
+            {
+                // PathID provided
+                // Warn if player also selected a creature
+                // -> Creature selection is ignored <-
+                if (target)
+                    handler->SendSysMessage(LANG_WAYPOINT_CREATSELECTED);
 
-            pathid = target->GetWaypointPath();
+                pathid = atoi((char*)guid_str);
+            }
         }
-        else
-        {
-            // PathID provided
-            // Warn if player also selected a creature
-            // -> Creature selection is ignored <-
-            if (target)
-                handler->SendSysMessage(LANG_WAYPOINT_CREATSELECTED);
-
-            pathid = atoi((char*)guid_str);
-        }
-
-        std::string show = show_str;
 
         //handler->PSendSysMessage("wpshow - show: {}", show);
+
+        constexpr uint32 wpEntry = VISUAL_WAYPOINT;
 
         // Show info for the selected waypoint
         if (show == "info")
         {
             // Check if the user did specify a visual waypoint
-            if (!target || target->GetEntry() != VISUAL_WAYPOINT)
+            if (!target || target->GetEntry() != wpEntry)
             {
                 handler->SendErrorMessage(LANG_WAYPOINT_VP_SELECT);
                 return false;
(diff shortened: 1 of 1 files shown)


src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -49,6 +49,57 @@ class wp_commandscript : public CommandScript
...
         return commandTable;
     }
+
+    /**
+    * Remove all waypoint from a query result.
+    *
+    * @param result: If the result not empty it should constains the creature database guids in the first column in all rows.
+    *
+    */
+    static void RemoveWaypointsFromQueryResult(ChatHandler* handler, const PreparedQueryResult& result)
```
SAS2000, line 59: I created this new function because this code (~100%) was duplicated in ".wp show on" when we wanted to delete current waypoints before we add the new ones and in ".wp show off" when we wanted to delete all waypoints.

Because only the DB query was different and the query result parsing was important part of the error handling I decided to start the function from the PreparedQueryResult.

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -49,6 +49,57 @@ class wp_commandscript : public CommandScript
...
+    {
+        if (result)
+        {
+            bool hasError = false;
+
+            do
+            {
+                const Field* fields = result->Fetch();
+                const ObjectGuid::LowType guid = fields[0].Get<uint32>();
+                const auto creatureRange = handler->GetSession()->GetPlayer()->GetMap()->GetCreatureBySpawnIdStore().equal_range(guid);
```
SAS2000: Here is the fix for the "cannot remove waypoints" bug.

Originally this line was: 
"Creature* creature = handler->GetSession()->GetPlayer()->GetMap()->GetCreature(ObjectGuid::Create<HighGuid::Unit>(VISUAL_WAYPOINT, guid));"

But the guid came from line 68 is a dbGuid (see line 929:  stmt->SetData(0, wpCreature->GetSpawnId()); )
So the original implementation can never find the waypoints in game thats why the error message appears.

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -49,6 +49,57 @@ class wp_commandscript : public CommandScript
...
+                        creatureRange.second,
+                        [&](auto& guidCreaturePair) {
+                            Creature* creature = guidCreaturePair.second;
+                            creature->CombatStop();
+                            creature->DeleteFromDB();
+                            creature->AddObjectToRemoveList();
+                        }
+                    );
+                }
+                else
```
SAS2000: After the fix this else case should never happen but we need more testing before deleting the whole if statement and only use the for_each.

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -752,41 +803,45 @@ class wp_commandscript : public CommandScript
         uint32 pathid = 0;
         Creature* target = handler->getSelectedCreature();
 
-        // Did player provide a PathID?
+        const std::string show = show_str;
 
-        if (!guid_str)
+        // If command equals "off" we do not need a guid or target, just delete all WPs
+        if (show != "off")
```
SAS2000, line 809: This is the fix for: the ".wp show off" should work without any path_id or any target. 
This is the intention based on the help (see command table 619 entry) and the implementation (line 1053: WorldDatabase.GetPreparedStatement(WORLD_SEL_CREATURE_BY_ID); This query does not contain path_id only queries on all waypoint entries.

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -752,41 +803,45 @@ class wp_commandscript : public CommandScript
...
-            pathid = atoi((char*)guid_str);
+                pathid = atoi((char*)guid_str);
+            }
         }
 
-        std::string show = show_str;
-
         //handler->PSendSysMessage("wpshow - show: {}", show);
 
+        constexpr uint32 wpEntry = VISUAL_WAYPOINT;
```
SAS2000, line 838: I put this name here instead of defining it in every if subpart

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -1036,7 +1051,7 @@ class wp_commandscript : public CommandScript
         if (show == "off")
         {
             WorldDatabasePreparedStatement* stmt = WorldDatabase.GetPreparedStatement(WORLD_SEL_CREATURE_BY_ID);
-            stmt->SetArguments(1, 1, 1);
+            stmt->SetArguments(wpEntry, wpEntry, wpEntry);
```
SAS2000, line 1054: Replace magic numbers with meaningful variables.

src/server/scripts/Commands/cs_wp.cpp
sudlud: Const comes after the type, also please no auto for simple data types

Also I am not sure if this call might need some null checks of the intermediate values

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -752,41 +803,45 @@ class wp_commandscript : public CommandScript
         uint32 pathid = 0;
         Creature* target = handler->getSelectedCreature();
 
-        // Did player provide a PathID?
+        const std::string show = show_str;
```
sudlud: Const after type, could you use string_view here?

src/server/scripts/Commands/cs_wp.cpp
sudlud: Should there be an early return if result or handler are not set probably?

sudlud (approved):

src/server/scripts/Commands/cs_wp.cpp
SAS2000: Fixed

src/server/scripts/Commands/cs_wp.cpp
SAS2000: There are 12 other examples for "GetCreatureBySpawnIdStore().equal_range" in the code and all of them use "auto" for the type that's why I choose auto here.

We test the result at the begining. After that both fields and guid should have a valid value.

SAS2000: I added a check to handler also in the first line.

SAS2000: image

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -49,6 +49,57 @@ class wp_commandscript : public CommandScript
...
+
+    /**
+    * Remove all waypoint from a query result.
+    *
+    * @param result: If the result not empty it should constains the creature database guids in the first column in all rows.
+    *
+    */
+    static void RemoveWaypointsFromQueryResult(ChatHandler* handler, const PreparedQueryResult& result)
+    {
+        if (handler && result)
```
sudlud: It is usually preferred to use early return for these checks to get rid of the giant if sections. e.g.

if (!handler || !result)
  return;

With this you don't have to indent the rest of the function, which improves readability a lot

sudlud:
do you want to update the function(s) you're updating in this pr to the new parameter standard? :D

e.g. like

```diff
-static bool HandleWpShowCommand(ChatHandler* handler, const char* args)
+static bool HandleWpShowCommand(ChatHandler* handler, std::string show, Optional<ObjectGuid::LowType> guid)
```

and the updated handling, usually makes the code much cleaner. you can take a look at other commands already updated, you can basically add the parameters directly to the function declaration now instead of using `const char* args`

Nyeriah:
Tiny one but it supports parsing on/off to Boolean

src/server/scripts/Commands/cs_wp.cpp
SAS2000: Ah ok, I misunderstood your point.

SAS2000:
> Tiny one but it supports parsing on/off to Boolean

It would be great but "wp show" also supports info and first/last.

SAS2000:
> do you want to update the function(s) you're updating in this pr to the new parameter standard? :D
> 
> e.g. like
> 
> ```diff
> -static bool HandleWpShowCommand(ChatHandler* handler, const char* args)
> +static bool HandleWpShowCommand(ChatHandler* handler, std::string show, Optional<ObjectGuid::LowType> guid)
> ```
> 
> and the updated handling, usually makes the code much cleaner. you can take a look at other commands already updated, you can basically add the parameters directly to the function declaration now instead of using `const char* args`

This is a great idea! But in the meantime I found a couple of other issues realted to "wp show first/last" so I suggest to deliver this changeset now as is and I will create a new one, cleaning this "wp show" a bit further and I also made this change.

SAS2000:
As I mentioned earlier during the testing of this PR I found a couple other problems in "wp show" so I created another pull request based on this. 

Please do not waste time on testing this alone use that branch and thest these two PRs together.

https://github.com/azerothcore/azerothcore-wotlk/pull/22741

sudlud:
so are these 2 prs seperate or should they actually be one pr? if they aim to fix the same thing / topic, maybe they should be combined please. or can they be tested individually

SAS2000:
> so are these 2 prs seperate or should they actually be one pr? if they aim to fix the same thing / topic, maybe they should be combined please. or can they be tested individually

I have two separate tickets that fixes separate issues but they all related because all issues are connected to the waypoint system.
Because at the code level I used some common refactorings all the final changes are on this branch:

https://github.com/azerothcore/azerothcore-wotlk/pull/22741

But I didn't want to merge the two tickets because the tickets separate the issues.

So how to test: only use the version from 22741 for all testing and you can test the fixes on both issues (#22740, #22726) there.

At the end if you approve this fix I only merge the latter one (#22741)

sudlud:
so if i understand you correctly, we could close this pr (#22727) and just use #22741 and there we could put both issues as beeing closed with the pr? it's alright of one pr closes multiple issues

PR #22741, opened 2025-08-28
fix(Core/Commands): Waypoint show command on/off/first/last #22740

### LINKED ISSUE #22740: Waypoint show command on/off/first/last
### Current Behaviour

This ticket is a direct follow up of my #22726 issue.

I furter tested the implementation and I found the following problems:

1. "wp show first/last" option was implemented (and working although buggy) but not appears in the description.
2. the "wp show first/last" behaviour significantly differs from the "show on" which is not the best practice. These commands should work exactly the same as "show on". The major differences:
2. a. The wps not added to the database so you cannot use "wp show info" on them.
2. b. The wps do not show their id in the path (wp show on shows every wps id as level)űűűűűűűűűűűűűűűűű
3. Now if you delete the first wp and call wp show first the query does not work (currently the SQL queries the id 1 and not the smallest Id as it should be in an environment where you can delete points)
4. After some testing on my last fix I realised that the error handling of "wp show off" is wrong. We should not delete wps from the database if we are far away because otherwise later when we are near we cannot delete them. Only the delete macro or a server restart can delete these wps. (Originally when there was a bug and we cannot delete wps at all this was a more or less OK workaround but now as it is stable this is not good anymore)

### Expected Behaviour

0. wp show first/last should appear in the command help
1. wp show on / first / last should work the same (using the same code)
2. wp show first / last should create a wp that can be "info"-ed and has a proper id and can be deleted as the other wps
3. wp show first should work if you delte the first point also.
4. As "wp show off" now works stable and only deletes wps from the current map we shuld not delete other wps because when we are near later we will not able to delete those wps also.

### Source

_No response_

0. wp gm on
1. .tele brill 
2. (go to the bridge, target Gordo) wp show first -> the wps level is not 1 (its id) as it should be
3. (target teh first wp) wp show info -> because we didn't added it to the db it fails
4. wp show off -> another error because it is not in the database

For the on/off issue

0. .tele brill
1. (target Gordo) wp show on
2. .tele dalaran
3. wp show off ->30 errors and the wps deleted from the db but not from the world
4. .tele brill
5. .wp show off ->the wps still there and you either delete tehm by gm commands or wait for server restart.

### Extra Notes

For me the following pattern seems wrong:

Creature* wpCreature = new Creature;
wpCreature->Create(...)
wpCreature->SaveToDB(...)
wpCreature->LoadCreatureFromDB(

First we create the creature, then we save it to db.
Then we loaded from db which means we created a new guid (and simply drop the first one -> waste) and inside the  load function we add it to the map.

My question: instead of the loadFromDb function why we not simply call 
GetMap()->AddToMap(wpCreature)

Its clean, shows intention and more performant.
If something missing from the saveDB that later on loadDB configure than it is a flaw of the saveDB funtion.

### AC rev. hash/commit

AzerothCore rev. f556278c6c96+ 2025-08-26 21:33:54 +0200 (Win64, RelWithDebInfo, Static)

### Operating system

Windows 11 x64

### Custom changes or Modules

_No response_

SAS2000:
NOTE: This PR is a direct update on my previous issue: https://github.com/azerothcore/azerothcore-wotlk/pull/22727

## Changes Proposed:

This PR proposes changes to:
**-  [ ] Core (units, players, creatures, game systems).**

## Issues Addressed:

- Closes #22740 

## Tests Performed:

This PR has been:
**- [ ] Tested in-game by the author.**

## How to Test the Changes:

**- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue**

- The wp show on/first/last implementation still not perfect. If we would like make it more robust we should also save the map_id and not just the dbGuid to the waypoint table. That way we can delete all waypoints in one go not just the ones that are in the same map as we are. But because it is a gm tool i think the proposed implementation is good enough.


TheSCREWEDSoftware:
Does this PR require https://github.com/azerothcore/azerothcore-wotlk/pull/22727 to work?

SAS2000:
> Does this PR require #22727 to work?

No, but I used that branch as a base. All the issues mentioned there (and how can you tested the fixes) are still valid here but I added some new improvements.

Only this code should be reviewed and merged (other one just stay there for reference)

data/sql/updates/db_world/2025_10_12_07.sql
```diff
@@ -0,0 +1,5 @@
+-- DB update 2025_08_28_00 -> 2025_08_28_01
+-- add wp show first/last to command help
+UPDATE `command` SET
+`help`='Syntax: .wp show $option\nOptions:\non $pathid (or selected creature with loaded path) - Show path\nfirst $pathid (or selected creature with loaded path) - Show first waypoint\nlast $pathid (or selected creature with loaded path) - Show last waypoint\noff - Hide path (all waypoints in current map)\ninfo $selected_waypoint - Show info for selected waypoint.'
+WHERE `name` = 'wp show';
```
SAS2000, line 5: Originally the first/last option was missing from the description

src/server/database/Database/Implementation/WorldDatabase.cpp
```diff
@@ -53,12 +53,12 @@ void WorldDatabaseConnection::DoPrepareStatements()
...
-    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_BY_ID, "SELECT point, position_x, position_y, position_z FROM waypoint_data WHERE id = ?", CONNECTION_SYNCH);
-    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_FIRST_BY_ID, "SELECT position_x, position_y, position_z FROM waypoint_data WHERE point = 1 AND id = ?", CONNECTION_SYNCH);
-    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_LAST_BY_ID, "SELECT position_x, position_y, position_z, orientation FROM waypoint_data WHERE id = ? ORDER BY point DESC LIMIT 1", CONNECTION_SYNCH);
+    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_BY_ID, "SELECT point, position_x, position_y, position_z, orientation, wpguid FROM waypoint_data WHERE id = ?", CONNECTION_SYNCH);
+    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_FIRST_BY_ID, "SELECT point, position_x, position_y, position_z, orientation, wpguid FROM waypoint_data WHERE id = ? ORDER BY point ASC LIMIT 1", CONNECTION_SYNCH);
+    PrepareStatement(WORLD_SEL_WAYPOINT_DATA_POS_LAST_BY_ID, "SELECT point, position_x, position_y, position_z, orientation, wpguid FROM waypoint_data WHERE id = ? ORDER BY point DESC LIMIT 1", CONNECTION_SYNCH);
```
SAS2000, line 58: First and last was different. First used point=1 which can be error prone if the first point was deleted. Also query the point number is important so we can show it in the UI as level.
I unified these three queries so we can use the same loop to parse the result.

src/server/database/Database/Implementation/WorldDatabase.cpp
```diff
@@ -53,12 +53,12 @@ void WorldDatabaseConnection::DoPrepareStatements()
...
     PrepareStatement(WORLD_SEL_WAYPOINT_DATA_BY_WPGUID, "SELECT id, point FROM waypoint_data WHERE wpguid = ?", CONNECTION_SYNCH);
     PrepareStatement(WORLD_SEL_WAYPOINT_DATA_ALL_BY_WPGUID, "SELECT id, point, delay, move_type, action, action_chance FROM waypoint_data WHERE wpguid = ?", CONNECTION_SYNCH);
-    PrepareStatement(WORLD_UPD_WAYPOINT_DATA_ALL_WPGUID, "UPDATE waypoint_data SET wpguid = 0", CONNECTION_ASYNC);
+    PrepareStatement(WORLD_UPD_WAYPOINT_DATA_BY_WPGUID, "UPDATE waypoint_data SET wpguid = 0 WHERE wpguid = ?", CONNECTION_ASYNC);
```
SAS2000, line 61: Separating the deleteion of the wp in the world and in the DB caused problems (you try to delte wps one-by-one in the world, and if you cannot you report it in the chat but after that you wipe all wps from the database)

Instead of this we delete wps from the db only if the deletion from the world was successful.

src/server/game/Miscellaneous/Language.h
```diff
@@ -287,8 +287,8 @@ enum AcoreStrings
     LANG_WAYPOINT_REMOVED               = 241,
     LANG_WAYPOINT_NOTREMOVED            = 242,
     LANG_WAYPOINT_TOOFAR1               = 243,
-    LANG_WAYPOINT_TOOFAR2               = 244,
-    LANG_WAYPOINT_TOOFAR3               = 245,
+    LANG_WAYPOINT_TOOFAR2               = 244,  // Not used
+    LANG_WAYPOINT_TOOFAR3               = 245,  // Not used
```
SAS2000, line 291: I wasn't sure how to handle this. Because I didn't want to delte them I marked them as unused.

src/server/scripts/Commands/cs_wp.cpp
```diff
@@ -49,6 +53,122 @@ class wp_commandscript : public CommandScript
         };
         return commandTable;
     }
+
+    static bool AddWaypointsByPreparedStatement(ChatHandler* handler, WorldDatabaseConnection::Statements index, uint32 pathid, Creature* target = nullptr)
```
SAS2000, line 57: This function is the common part of on/first/last

SAS2000:
I sync-ed my branch but after that one check is failing. I didn't touch that part of the code. Can someone help me finding out what can be the problem?

sudlud (commented):
Hi @SAS2000, thank you for this contribution and for the detailed issue write-up that goes with it. I am sorry it took so long to get a proper second look at this. The refactor is a clear improvement: unifying `on`/`first`/`last` behind a single `AddWaypointsByPreparedStatement` helper, making `first`/`last` DB-backed and info-able, switching `first` to `ORDER BY point ASC LIMIT 1` so it survives deletion of point 1, and scoping `wp show off` to the current map are all the right calls and they match what #22740 asked for.

There are a few things that need to be sorted before this can be merged, mostly mechanical, plus a few cleanups.

## Required before merge

1. **Rebase onto current master (the branch is conflicting).** The branch is based on an older snapshot, and two things have moved since:
   - **RBAC permissions.** The command table here registers the `wp` subcommands with `SEC_ADMINISTRATOR`, but master has migrated them to `rbac::RBAC_PERM_COMMAND_WP_*`. As-is, merging would revert that. Please keep the `rbac::RBAC_PERM_COMMAND_WP_SHOW` (and siblings) when you rebase.

2. **Move the SQL update into `pending_db_world/`.** The PR adds `data/sql/updates/db_world/2025_10_12_07.sql` directly into the merged directory, with a hand-written `-- DB update 2025_10_12_06 -> 2025_10_12_07` header. That directory is import-only, and the date is now well behind master, which breaks the update chain and is part of why the branch conflicts. Please:
   - delete `db_world/2025_10_12_07.sql`,
   - run `./create_sql.sh` inside `data/sql/updates/pending_db_world/` and put the same `UPDATE command ...` statement into the generated `rev_<timestamp>.sql`.

   The import step assigns the correct date and chains the header at merge time. The SQL content itself is fine.

3. **CodeFactor (the failing check you asked about).** It is flagging style. A couple of spots to check against `apps/codestyle/codestyle-cpp.py`:
   - `namespace {` and the lambda in `RemoveWaypointById` use K&R braces; the project uses Allman.
   - `const ObjectGuid::LowType guid` should be `ObjectGuid::LowType const guid` (the `Type const` rule).

## Cleanups requested in the Copilot review (all valid)

4. **Remove the `DEBUG:` chat messages.** `wp on`/`first`/`last`/`info` and the final fallthrough all print `DEBUG:` lines to the GM's chat. Some predate this PR, but since this rewrites the function it is a good time to drop them (or route to logs). While there, the final `"no valid command found"` branch returns `true`; returning `false` would surface the command help on a bad subcommand.

5. **Hoist the `waypoint_data` update out of the loop in `RemoveWaypointById`.** When `deleteFromWaypointData` is true, `WORLD_UPD_WAYPOINT_DATA_BY_WPGUID` runs once per entry in the `equal_range`, even though it is identical each time. It only needs to run once.

6. **Drop the redundant `if (result)` in the `off` branch.** It is immediately preceded by `if (!result) return false;`, so the inner check is dead.

## Edge cases worth a note (non-blocking, your call)

7. **Orphan rows when replacing a visual on another map.** In the replace path, `RemoveWaypointById(handler, wpguid)` is called with `deleteFromWaypointData = false`. If that old visual creature is not in the current map, the call returns `false`, the old creature row is left in the DB, and then `wpguid` is overwritten with the new spawn, leaking the old row. This is the tension you already noted in your TODO about storing `map_id`, so I am fine treating it as a known limitation rather than a blocker. A short comment in the code acknowledging it would be enough.

8. **`LoadCreatureFromDB` failure path.** If `SaveToDB` succeeds but `LoadCreatureFromDB` fails, the saved creature row and the `wpguid` reference are left behind. This is extremely unlikely (it reloads the row just written) and already exists in master, so I am not asking you to fix it here, just flagging it.

9. Minor: the `Optional<ObjectGuid::LowType> guid` parameter is actually a path id. Renaming it to `pathId` would read better.

Thanks again for sticking with this and for the careful issue analysis. Once the rebase, the SQL relocation, and the CodeFactor and DEBUG cleanups are done, this should be in good shape.

Kitzunu:
I will close as the author has not adressed the reviews by copilot that was made 3 months ago. Feel free to pick this up at a later date :)

Issue #22742, opened 2025-08-28
[NPC: Bubulo Acerbus] is missing from the world


TheSCREWEDSoftware:
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/8290

Triage Notes: 

Gnome's creature template exists, but no spawn in the world, requires sniff.

https://wowgaming.altervista.org/aowow/?npc=14724

wowhead.com/wotlk/npc=14724/bubulo-acerbus?fromClassic

Original Post Below: 

---

### Current Behaviour

<img width="1111" height="743" alt="Image" src="https://github.com/user-attachments/assets/1cfe6b23-13a1-4c65-95e1-50ec0009852b" />
The npc for cloth turnins for Gnome rep, Bubulo Acerbus, has gone missing.

### Expected Blizzlike Behaviour

Should be there.

### Source

https://www.wowhead.com/classic/npc=14724/bubulo-acerbus

### Steps to reproduce the problem

1. Go to tinker town
2. /target Bubulo, confirm he is not there.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/71c22ff6cf8c9d442579ec5905c5eee73408b029

### Operating system

Ubuntu 20.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-chromiecraft-smartstone](https://github.com/chromiecraft/mod-chromiecraft-smartstone)
- [mod-costumes](https://github.com/azerothcore/mod-costumes)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[...]

PR #22768, opened 2025-09-02
fix(DB/Creature): Add spawn for Bubulo Acerbus.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22742
- Closes https://github.com/chromiecraft/chromiecraft/issues/8290

## SOURCE:

- [X] Sniffs.

## Tests Performed:

This PR:
- [X] Has not been tested.


TheSCREWEDSoftware (approved):
All good, if you feel like adding to the creature's comment `Bubulo Acerbus` would be a plus, if not is still good to merge

heyitsbench:
I'm not entirely sure what the initial idea of the comment column was for, but I definitely don't agree with sticking the creature's name in there. Can easily be found with a join query.

sudlud:
> I'm not entirely sure what the initial idea of the comment column was for, but I definitely don't agree with sticking the creature's name in there. Can easily be found with a join query.

E.g. marking spawns with guid specific SAI / SAI targets

TheSCREWEDSoftware:
> I'm not entirely sure what the initial idea of the comment column was for, but I definitely don't agree with sticking the creature's name in there. Can easily be found with a join query.

Disregard my previous statement for the comment

image

I guess is just to be aware of that creature's GUID is being used in SAI

Issue #22771, opened 2025-09-02
[Item: Guild Charters] If the GUID, Enchamtement (first column) or character_inventory itemID are too high, they dont work.


TheSCREWEDSoftware:
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/8275#issuecomment-3246416065

Triage Notes: https://github.com/chromiecraft/chromiecraft/issues/8275#issuecomment-3246416065

<img width="239" height="227" alt="Image" src="https://github.com/user-attachments/assets/7dbc52fc-3f63-424b-a3ac-1a341dc02bad" />

Sometimes tooltip, sometimes not, but it doesnt retain the the charter's information at all, one of the things on the title is the issue, im unsure which one.

Original Post Below: 

---

### Current Behaviour

Guild Chanter won't open pop out window to get signs from people.

### Expected Blizzlike Behaviour

Guild Chanter opens up and allows to get signs from people.

### Source

Its too common to provide source.

### Steps to reproduce the problem

1. Go to NPC that create guilds
2. Write the guild name
3. Receive guild chanter
4. Click right click to open guild chanter
5. Guild chanter won' open

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/c67d4efa4a19625a3d0bbe902dd3ba9f6980d2a2

### Operating system

Ubuntu 20.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-chromiecraft-smartstone](https://github.com/chromiecraft/mod-chromiecraft-smartstone)
[...]

PR #22774, opened 2025-09-02
feat(core): implement 31-bit safe petition_id for improved database integrity

Yehonal:
## The problem on ChromieCraft

🔍 What’s happening now

Some players can buy a guild charter, rename it, but signatures don’t work.

After logging out and back in, the charter often won’t open anymore.

🧪 Tests done

We reproduced this on empty/local servers, so it’s not related to data corruption or player load.

We confirmed that the first enchantment slot of the charter item stores its GUID.

When that GUID goes above 2,147,483,647 (the signed 32-bit integer limit), the 3.3.5a client misreads it.

That explains why renaming works (value still in memory) but signing and reopening after relog break (client interprets the value as invalid).

⚙️ Why this matters

The database and core handle these values correctly as uint32.

The problem is very likely to be client-side only, a legacy limitation of the 3.3.5a client we guess

## Changes Proposed:

TLDR:

Introduce a new 31-bit safe petition identifier to enhance database integrity and decouple it from the item instance GUID. Update relevant database tables and queries to utilize the new identifier.

FULL:

This pull request introduces a new 31-bit safe `petition_id` to decouple petition handling from item GUIDs, improving database integrity and future compatibility. The changes refactor both database schema and code logic to consistently use `petition_id` for all petition operations, including creation, lookup, and deletion. Additionally, new helper methods and mappings are added to efficiently convert between item GUIDs and petition IDs.

**Database schema and migration:**

* Added a new `petition_id` column to the `petition` and `petition_sign` tables, backfilled existing data, and created supporting indexes for efficient lookups and deletes. Existing item GUIDs are mapped into the 31-bit safe range, and charter items are updated to store the new `petition_id` in their enchantments. (`data/sql/updates/pending_db_characters/rev_1756841669142726711.sql`)

**Codebase refactoring for petition_id usage:**

* All prepared statements and queries in `CharacterDatabase.cpp` are updated to use `petition_id` instead of item GUIDs for petition operations, including insertion, updating, and deletion of petitions and signatures. (`src/server/database/Database/Implementation/CharacterDatabase.cpp`, `src/server/database/Database/Implementation/CharacterDatabase.h`) [[1]](diffhunk://#diff-642c906075729a1eb0ec3dcdcf01e08de47d00ca88464b74b876f14f468f086eL356-R357) [[2]](diffhunk://#diff-642c906075729a1eb0ec3dcdcf01e08de47d00ca88464b74b876f14f468f086eL459-R461) [[3]](diffhunk://#diff-3f9e61636f74100b903abddf7d5805c185e0b5f814979de53ebc1868a53b3fe3L384-R385)

**Petition manager enhancements:**

* The `PetitionMgr` class now tracks both `petition_id` and item GUID mappings, supports generating unique 31-bit safe petition IDs, and includes new helper methods for bidirectional lookup between IDs and GUIDs. Petition loading and signature loading routines are updated to use these mappings and skip orphaned signatures. (`src/server/game/Petitions/PetitionMgr.cpp`, `src/server/game/Petitions/PetitionMgr.h`) [[1]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014R44-R46) [[2]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014R55-R72) [[3]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014L70-R82) [[4]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014L82-R110) [[5]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014L105-R127) [[6]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014R168-R176) [[7]](diffhunk://#diff-cba4d48325f8985e7289d1d8955b44ed72e766bb61b079177e2c731fa771f014R223-R256) [[8]](diffhunk://#diff-7049a4fd5eeaf4297d24f2f4a36144a395c673d56ef11cad08106f5e9b837f1bR40-R54) [[9]](diffhunk://#diff-7049a4fd5eeaf4297d24f2f4a36144a395c673d56ef11cad08106f5e9b837f1bL68-R79) [[10]](diffhunk://#diff-7049a4fd5eeaf4297d24f2f4a36144a395c673d56ef11cad08106f5e9b837f1bR90-R100)

**Game handler logic updates:**

* All petition-related game handler code now uses `petition_id` for packet serialization, DB operations, and in-game logic, ensuring consistent use of the new identifier throughout the petition lifecycle (buy, show, sign, rename, turn-in, etc.). (`src/server/game/Handlers/PetitionsHandler.cpp`) [[1]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L186-R188) [[2]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L214-R227) [[3]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L252-R256) [[4]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L289-R293) [[5]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L376-R380) [[6]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L500-R504) [[7]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L628-R632) [[8]](diffhunk://#diff-ee3e6ecd6de2320a3b7f941d58ba06e0f9ae2657e1c5ed9ae703eb276d5e75d9L793-R802)

**Type and structure improvements:**

* The `Petition` struct now includes explicit documentation for each field, including the new `petitionId`, to clarify its role and improve code readability. (`src/server/game/Petitions/PetitionMgr.h`)

These changes collectively modernize the petition system, making it safer and more maintainable by removing its dependency on item GUIDs and introducing a robust, unique identifier for all petition operations.

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


heyitsbench:
This PR targets https://github.com/azerothcore/azerothcore-wotlk/issues/22771 for the sake of tying things together.

However, I don't think this approach is correct. Assuming the comments and variable names in various locations are to be believed, I think the more likely value being sent in the packets in PetitionsHandler *is* the guild/arena team GUID. Would it not be preferable to create the guild/team when the petition/charter is acquired and provide the ID there rather than create a whole separate value to be tracked and taken care of, especially if it's not accurate?

Yehonal:
> This PR targets #22771 for the sake of tying things together.
> 
> However, I don't think this approach is correct. Assuming the comments and variable names in various locations are to be believed, I think the more likely value being sent in the packets in PetitionsHandler _is_ the guild/arena team GUID. Would it not be preferable to create the guild/team when the petition/charter is acquired and provide the ID there rather than create a whole separate value to be tracked and taken care of, especially if it's not accurate?

In WotLK the value handled in PetitionsHandler is actually the charter item’s low GUID (basically a client-facing token, not the real guild/team GUID)

If we created the guild right when the charter is purchased, we’d change the expected flow and end up with orphan guilds and extra cleanup logic.

Using a dedicated 31-bit petition_id avoids the client’s signed-int overflow and keeps the current server lifecycle untouched.

An alternative could be to just mask bit 31 on the item GUID, but the petition_id approach is cleaner and more future-proof IMHO.

heyitsbench:
> In WotLK the value handled in PetitionsHandler is actually the charter item’s low GUID (basically a client-facing token, not the real guild/team GUID)

Source? Because that definitely isn't the case on retail, and it doesn't seem like a system that's changed much between now and then.

<details>
<summary>Example from sniff (11.2.0)</summary>
ServerToClient: SMSG_UPDATE_OBJECT (0x460000)

[0] UpdateType: CreateObject1

[0] Object Guid: Full: 0x0 Item/0 R0/S0 Low: 59809686613

[0] Object Type: 1 (Item)

[0] EntryID: 5863

[0] (Enchantment) [0] ID: 107214264
</details>

Yehonal:
> > In WotLK the value handled in PetitionsHandler is actually the charter item’s low GUID (basically a client-facing token, not the real guild/team GUID)
> 
> Source? Because that definitely isn't the case on retail, and it doesn't seem like a system that's changed much between now and then.
> 
> Example from sniff (11.2.0)
> ServerToClient: SMSG_UPDATE_OBJECT (0x460000)
> [0] UpdateType: CreateObject1
> 
> [0] Object Guid: Full: 0x0 Item/0 R0/S0 Low: 59809686613
> 
> [0] Object Type: 1 (Item)
> 
> [0] EntryID: 5863
> 
> [0] (Enchantment) [0] ID: 107214264

Retail 11.2.0 sniffs aren’t comparable to WotLK 3.3.5a because the protocol (GUID size/packing, update fields) changed a lot: retail uses wider GUIDs and different packet layouts, so the “enchantment” ID you see there doesn’t map 1:1
with Wrath. 
In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in. Our change swaps that token with a 31‑bit‑safe petition_id to avoid the client’s signed‑int overflow, without altering when the guild is actually created.

heyitsbench:
> In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in.

I know exactly what AC currently does with this value, my point is that I don't believe that it's correct for 3.3.5, based not only on the idea that sending the charter item GUID along with the item itself (which already has the GUID baked into its data, because of course it does) makes no sense, and the comments in AC's code definitely imply that it *should* be using the ID/GUID of the guild/arena team it wants to make, and I don't have a second account I can easily check with atm, but I have a feeling that retail does the same thing. Using AC's code as a source for 3.3.5 behavior is not a valid source.

Takenbacon:
> > In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in.
> 
> I know exactly what AC currently does with this value, my point is that I don't believe that it's correct for 3.3.5, based not only on the idea that sending the charter item GUID along with the item itself (which already has the GUID baked into its data, because of course it does) makes no sense, and the comments in AC's code definitely imply that it _should_ be using the ID/GUID of the guild/arena team it wants to make, and I don't have a second account I can easily check with atm, but I have a feeling that retail does the same thing. Using AC's code as a source for 3.3.5 behavior is not a valid source.

Attempting to create an empty guild/arena team would be significantly more complicated, both from a memory management and just resource wastefulness point of view. I do have a hard time believing blizzard would be that dumb to do so but who knows, they tend to do very questionable things sometimes just like equipment sets apparently use a global guid for some reason even though they're very much player specific.

To me this PR has no downsides, it's an id that is completely transparent to a player. Even if we aren't sure if it's technically correct or not it really wouldn't matter.

Yehonal:
> > In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in.
> 
> I know exactly what AC currently does with this value, my point is that I don't believe that it's correct for 3.3.5, based not only on the idea that sending the charter item GUID along with the item itself (which already has the GUID baked into its data, because of course it does) makes no sense, and the comments in AC's code definitely imply that it _should_ be using the ID/GUID of the guild/arena team it wants to make, and I don't have a second account I can easily check with atm, but I have a feeling that retail does the same thing. Using AC's code as a source for 3.3.5 behavior is not a valid source.



> > In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in.
> 
> I know exactly what AC currently does with this value, my point is that I don't believe that it's correct for 3.3.5, based not only on the idea that sending the charter item GUID along with the item itself (which already has the GUID baked into its data, because of course it does) makes no sense, and the comments in AC's code definitely imply that it _should_ be using the ID/GUID of the guild/arena team it wants to make, and I don't have a second account I can easily check with atm, but I have a feeling that retail does the same thing. Using AC's code as a source for 3.3.5 behavior is not a valid source.

Sorry, but are you suggesting to allocate (and reserve) a guild ID right when the petition is created? That would likely produce many orphan IDs because a lot of charters are never turned in, and it forces us to pre-reserve rows in the guild table (or maybe I'm missing something)

src/server/game/Petitions/PetitionMgr.h
```diff
@@ -79,9 +87,17 @@ class PetitionMgr
...
 
+    uint32 GeneratePetitionId();
+    uint32 GetPetitionIdByItemGuid(ObjectGuid petitionItemGuid) const;
+    ObjectGuid GetItemGuidByPetitionId(uint32 petitionId) const;
+
 protected:
     PetitionContainer PetitionStore;
     SignatureContainer SignatureStore;
+    // Mapping id -> item-guid to support DB-id lookups
+    std::map<uint32, ObjectGuid> PetitionIdToItemGuid;
```
Takenbacon: unordered_map would probably be better

Yehonal:
> > > In 3.3.5a (AzerothCore), the charter flow explicitly uses the petition item’s low 32‑bit GUID as the client‑visible token: the code sets it into the first enchantment and sends the same value in petition packets, while the real guild/team object is only created at turn‑in.
> > 
> > 
> > I know exactly what AC currently does with this value, my point is that I don't believe that it's correct for 3.3.5, based not only on the idea that sending the charter item GUID along with the item itself (which already has the GUID baked into its data, because of course it does) makes no sense, and the comments in AC's code definitely imply that it _should_ be using the ID/GUID of the guild/arena team it wants to make, and I don't have a second account I can easily check with atm, but I have a feeling that retail does the same thing. Using AC's code as a source for 3.3.5 behavior is not a valid source.
> 
> Attempting to create an empty guild/arena team would be significantly more complicated, both from a memory management and just resource wastefulness point of view. I do have a hard time believing blizzard would be that dumb to do so but who knows, they tend to do very questionable things sometimes just like equipment sets apparently use a global guid for some reason even though they're very much player specific.
> 
> To me this PR has no downsides, it's an id that is completely transparent to a player. Even if we aren't sure if it's technically correct or not it really wouldn't matter.

I missed this reply earlier, +1. 

From a design perspective it also makes sense that the petition token stays small: it’s a short-lived identifier used only until turn-in and then discarded, conceptually similar to an instance/raid token. Also to avoid guild IDs orphans ofc.

heyitsbench:
> Sorry, but are you suggesting to allocate (and reserve) a guild ID right when the petition is created?

I want things to be implemented that are accurate to Blizzard's servers, that is all. This implementation does not seem accurate to me.

src/server/game/Petitions/PetitionMgr.h
Yehonal: done, thanks!

heyitsbench:
Confirmed with the help of @TheSCREWEDSoftware that no, the field attached to the petition is neither the charter item GUID nor the GUID of the final guild. I have no issue with this PR.

Nyeriah (approved):

Issue #16442, opened 2023-06-01
(Creature): Certain creatures spontaneously have wrong display ID.


heyitsbench:
### Current Behaviour

Sometimes certain creatures seem to have the wrong display ID from what they should according to the database. This has been observed with [Zephyr](https://wowgaming.altervista.org/aowow/?npc=25967) and [Gelanthis](https://wowgaming.altervista.org/aowow/?npc=16624). Zephyr sometimes shows up with a display ID of 19278, Gelanthis sometimes appears with ID 15505. This seems to be a *very* inconsistent issue.

### Expected Blizzlike Behaviour

Zephyr's native model ID is 19282, Gelanthis' expected ID is 15504. They should never deviate.

### Source

Assumption, values from the database

### Steps to reproduce the problem

1. `.go c id 25967/16624`
2. `.npc info`
3. Observe their display ID.
4. Restart AC until it deviates from what's in the DB.

### Extra Notes

I noticed the issue with Zephyr in [this AC issue](https://github.com/azerothcore/azerothcore-wotlk/issues/14915). Vyperhand observed the issue with Gelanthis in [this CC issue](https://github.com/chromiecraft/chromiecraft/issues/5532#issuecomment-1544571835).

This is an inconsistent issue, but it most definitely *does* happen. Don't know if the model can swap while the server is running (I doubt it), but it can swap between restarts.

### AC rev. hash/commit

a696227e8de1b26088d3a2abbb29d861aeb48672

### Operating system

Windows 10 x64

### Custom changes or Modules

None.

Nyeriah:
Model_other_gender in the model table, pretty common issue (e.g captain greenskin)

heyitsbench:
> Model_other_gender in the model table, pretty common issue (e.g captain greenskin)

Well that's dumb.

Gultask:
Those values should be sniffed as well, no? I can't see why someone would decide to add a random model there for fun.

Gultask:
I remember another similar issue with model_other_gender I had to fix by forcing the model on script

wzw1990:
So why don't we simply remove that *dumb* code?

heyitsbench:
> So why don't we simply remove that _dumb_ code?

It's not dumb code, it's just a possibly bad database entry.

avarishd:
Was this solved by #19068 ?

heyitsbench:
Don't believe so, I think the issue with this is incorrect 'model_other_gender' value.

[...]

Issue #7769, opened 2021-09-07
Milton Beats Randomly Spawning With Female Model


Azcobu:
### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [X] 30-39
- [ ] 40-49
- [ ] 50-59

### Current Behaviour
Original report: https://github.com/chromiecraft/chromiecraft/issues/1617

Milton Beats (NPC ID 13082) is spawned by opening a chest in the tunnel that leads to Ravenholdt Manor (he's used in a rogue quest).
He can spawn with a female model currently.
![WoWScrnShot_090321_120827](https://user-images.githubusercontent.com/19631518/132030085-77f87f9d-464e-49aa-a3a3-6708cc230a25.png)

Tester: spawned this mob in the open world and they did appear using the female model. Spawned a few dozen and they seemed to come out roughyl 50-50, though had one long run of ~7 males in a row.

![WoWScrnShot_090721_234653](https://user-images.githubusercontent.com/81782124/132360674-67085092-a2c9-4fe5-bf11-6511b04fcad7.jpg)

Slap spam makes this inadvisable:
![WoWScrnShot_090821_011838](https://user-images.githubusercontent.com/81782124/132374497-f16830be-bdb9-4332-aa7c-b764f071f684.jpg)

It's the same model as the male/female Syndicate Prowlers. (4016/7) However, only a single model, 4016, is set in `creature_template` for Milton. Checking `creature_model_info` for 4016 shows it is set to gender 0 (= male). @JamesGent suggested on the CC tracker that the model might be flagged as M/F, but it doesn't seem to be, though I don't know how model 4017 is being used otherwise. I even had a look in `conditions`, but the only mention of Milton there is to reset SmartAI. 

### Expected Blizzlike Behaviour

He should always be male, and have [this model](https://wow.zamimg.com/uploads/screenshots/normal/64199-milton-beats-milton-beats-this-friendly-guy-guards-a-chest.jpg). 

### Source

https://www.wowhead.com/npc=13082/milton-beats

### Steps to reproduce the problem

If you have access to GM commands
1. `.npc add temp 13082`
2. Check the model

If you don't have access to GM commands (he's an aggressive elite NPC level 33, if you are lower level he will kill you)
[...]

PR #22782, opened 2025-09-03
fix(DB/Creature): Remove incorrect other gender display from Vurtok Axebreaker.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/8300

## SOURCE:

- [X] Sniffs.

## Tests Performed:

This PR:
- [X] Has not been tested.

### DIFF
--- a/data/sql/updates/pending_db_world/vurtok-axebreaker.sql
+++ b/data/sql/updates/pending_db_world/vurtok-axebreaker.sql
@@ -1,0 +1,1 @@
+UPDATE `creature_template_model` SET `DisplayID_Other_Gender` = 0 WHERE `DisplayID` = 16294;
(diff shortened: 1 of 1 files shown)


Rorschach91:
Can we touch this table? I thought not.

heyitsbench:
It has no VerifiedBuild column, and data definitely has been wrong in there before.

Rorschach91:
> It has no VerifiedBuild column, and data definitely has been wrong in there before.

In this case, can we resolve these issues in this way?
https://github.com/azerothcore/azerothcore-wotlk/issues/16442
https://github.com/azerothcore/azerothcore-wotlk/issues/7769
https://github.com/chromiecraft/chromiecraft/issues/1617
https://github.com/azerothcore/azerothcore-wotlk/issues/22090

```sql
UPDATE `creature_model_info` SET `DisplayID_Other_Gender` = 0 WHERE `DisplayID` IN (2981, 2983, 4016, 4017, 15504, 15505, 19278, 19282);
```

heyitsbench:
> In this case, can we resolve these issues?

This instance I inspected the creatures that make use of the affected display IDs and ensured no issues would come from the change. The same would have to be done for those issues to make sure functionality is still accurate for all affected creatures.

Nyeriah:
The problem is that some models are shared by various other creatures

TheSCREWEDSoftware (approved):

PR #22794, opened 2025-09-04
feat(Core/Restricted Flight Area): implements verification logic

grimgravy:
## Changes Proposed:

This PR proposes changes to:

This commit introduces flight restriction logic to the route planning system, specifically for the zones Dalaran and Wintergrasp.
- When a player falls from a significant height within these restricted areas:
- If they are at a high altitude, a parachute is automatically deployed to prevent fall damage.
- If they are at a low altitude, flight ability is disabled without triggering a fall.

## SOURCE:

## Tests Performed:

This PR has been:

### DIFF
--- a/src/server/game/Spells/Auras/SpellAuraEffects.cpp
+++ b/src/server/game/Spells/Auras/SpellAuraEffects.cpp
@@ -5644,7 +5644,25 @@ void AuraEffect::HandleAuraDummy(AuraApplication const* aurApp, uint8 mode, bool
                         case 58600: // Restricted Flight Area
                         case 58730: // Restricted Flight Area
                             if (aurApp->GetRemoveMode() == AURA_REMOVE_BY_EXPIRE)
-                                target->CastSpell(target, 58601, true);
+                            {
+                                float posX = target->GetPositionX();
+                                float posY = target->GetPositionY();
+                                float posZ = target->GetPositionZ();
+                                float groundZ = target->GetMapHeight(posX, posY, posZ);
+                                float HeightAboveGround = posZ - groundZ;
+                                float MinParachuteHeight = 8.0f;
+
+                                if (HeightAboveGround >= MinParachuteHeight)
+                                {
+                                    target->CastSpell(target, 58601, true);
+                                }
+
+                                if (HeightAboveGround < MinParachuteHeight)
+                                {
+                                    target->RemoveAurasByType(SPELL_AURA_FLY);
+                                    target->RemoveAurasByType(SPELL_AURA_MOD_INCREASE_MOUNTED_FLIGHT_SPEED);
+                                }
+                            }
                             break;
                         case 46374: // quest The Power of the Elements (11893)
                             {


TheSCREWEDSoftware:
Are you able to provide / share the sources on this behaviour?

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -5644,7 +5644,25 @@ void AuraEffect::HandleAuraDummy(AuraApplication const* aurApp, uint8 mode, bool
...
                         case 58730: // Restricted Flight Area
                             if (aurApp->GetRemoveMode() == AURA_REMOVE_BY_EXPIRE)
-                                target->CastSpell(target, 58601, true);
+                            {
+                                float posX = target->GetPositionX();
+                                float posY = target->GetPositionY();
+                                float posZ = target->GetPositionZ();
+                                float groundZ = target->GetMapHeight(posX, posY, posZ);
+                                float HeightAboveGround = posZ - groundZ;
+                                float MinParachuteHeight = 8.0f;
```
sudlud: These can all be float const, also names should start with lower case

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -5644,7 +5644,25 @@ void AuraEffect::HandleAuraDummy(AuraApplication const* aurApp, uint8 mode, bool
...
+
+                                if (HeightAboveGround >= MinParachuteHeight)
+                                {
+                                    target->CastSpell(target, 58601, true);
+                                }
```
sudlud: Remove braces, please add comment with spell name or better add spell to enum with name

grimgravy:
> Are you able to provide / share the sources on this behaviour?

I couldn't find it, but I remember a video from years ago.

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -5644,7 +5644,25 @@ void AuraEffect::HandleAuraDummy(AuraApplication const* aurApp, uint8 mode, bool
...
+
+                                if (HeightAboveGround < MinParachuteHeight)
```
sudlud: Replace by "else"

sudlud:
so basically if you're too close to ground you should just start falling?

grimgravy:
> so basically if you're too close to ground you should just start falling?

does not activate the parachute

Teknishun7:
Afaik, the PR grimgravy is proposing is blizzlike

Looks like this can at least be tested. I will also try and find confirmation of this behavior, but Im like 90% it is how its supposed to be.

Teknishun7:
I spent way too long on trying to merge this, no matter how I tried this just wouldnt merge, usually stubborn PRs I end up successfully merging. This one simply does not share the same history as Master so it was just not possible. Since its only a very small change in a single file, I fixed the merge conflict manually.

No matter how I tested it, the PR didnt work.

sudlud:
i am closing this PR for now, please feel free to reopen and update it anytime in the future to address the last comments / issues.

PR #22821, opened 2025-09-06
feat(CI): support macos-15

FrancescoBorzi:


TheSCREWEDSoftware (approved):
Build for it's failing, but irrelevant to this PR

sudlud:
Are you sure? Build should pass

TheSCREWEDSoftware:
To my understanding this is just to update mac-os to support 15 not that it change it to work as 15, corrcet me if wrong @FrancescoBorzi .

If the case is to try to fix it this requires cmake changes and probably core changes as we getting `Undefined symbols for architecture arm64:` so this is a cmake and maybe a worldserver changes required to put working it.

sudlud:
This pr changes our mac CI to macOS 15, so it must not be merged with failing mac ci.

FrancescoBorzi:
this needs more attention: postponed

Issue #19769, opened 2024-08-28
Players can get parried while behind a mob


Rorschach91:
Bug Triager Notes:

Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/7310#issue-2485449119

This is a Generic issue that regard some bosses. 

### What client do you play on?

enGB

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

While standing behind a mob, players can still be parried

[skillissue.webm](https://github.com/user-attachments/assets/b950d978-387d-4665-8f81-6ca32c1fa80d)

![image](https://github.com/user-attachments/assets/a699b5a5-5f95-40df-a73a-7a086d0b5cf5)

### Expected Blizzlike Behaviour

Mobs cannot parry while you are behind them. 

### Source

https://wowpedia.fandom.com/wiki/Parry?oldid=2363574
![image](https://github.com/user-attachments/assets/7f380130-f8bb-482c-a9b7-c53e8f292acb)
Note they do say bugs/quick turn, but this seems like a new issue as haven't really seen before. 

### Steps to reproduce the problem

To replicate closer to ticket; 
1. Zone into Hyjal
2. Go to Kaz'rogal
3. Test attacks from behind (if can, use Execute?)
Else if on own AC, just summon the boss and test out attacks from behind. 

### Extra Notes

Hadn't really seen this before when standing pretty visibly behind boss. My guess was a somewhat recent hitbox change to AoE spells. 

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/30a790117be921ed0eff513322af1793995b2b4d

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-BreakingNewsOverride](https://github.com/AnchyDev/BreakingNewsOverride)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
[...]

PR #22500, opened 2025-07-16
fix(Core/Player): allow attacking target within boundary radius when…

sogladev:
… not facing the target

## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

changes can be illustrated by screenshots: 
https://github.com/TrinityCore/TrinityCore/pull/16885#issuecomment-225691166

This change allows a player to attack a target when standing very close to the target ignoring facing requirements. Cone targets like Cone of Cold are included in this. The screenshot below should no longer occur.
<details>
<summary>
Screenshot AC current behavior
</summary>

WoWScrnShot_071625_150423

</details>

## Issues Addressed:

- Closes 

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [x] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [x] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

cherry-pick of "Game/Entities: Players can attack targets when they are not facing them as long as they are in boundaryradius, this also applies for cone effect spells
"
https://github.com/TrinityCore/TrinityCore/commit/86275a397a478a30dad7e00af73029e2b15d4668

old PvP footage and see if abilities or attacks that should display 'You are facing the wrong way!' still succeed when the player is positioned on top of the target.

1. Eviscerate is cast without facing the target
https://youtu.be/0wnC_3AkMfU?si=GhYyMUblNP7z9IOH&t=111

2. MS casts when the player runs behind the target
https://youtu.be/AW_tY_sRdz0?si=zZtfa6Sp0EvaJkVG&t=125

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1 or 2 characters

[...]

PR #22826, opened 2025-09-06
Port HasInArc from Trinity

v-mstrs:
https://github.com/azerothcore/azerothcore-wotlk/blob/fa087ff05460b08ccc2de3065a77f2afa8f54b51/src/server/game/Entities/Unit/Unit.cpp#L2902-L2908
https://github.com/azerothcore/azerothcore-wotlk/blob/fa087ff05460b08ccc2de3065a77f2afa8f54b51/src/server/game/Entities/Object/Position.cpp#L161-L171
**Haven't tested** but seems about right.
Could also just remove the above lines but might as well update it while we are at it.
Need to add co-authors
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19769

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.Zone into Hyjal
2. Go to Kaz'rogal
3. Test attacks from behind (if can, use Execute?)
4. Else if on own AC, just summon the boss and test out attacks from behind.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Object/Position.cpp
+++ b/src/server/game/Entities/Object/Position.cpp
@@ -138,25 +138,27 @@ bool Position::IsWithinBox(const Position& center, float xradius, float yradius,
     return true;
 }
 
-bool Position::HasInArc(float arc, Position const* obj, float border) const
+bool Position::HasInArc(float arc, const Position* obj, float targetRadius) const
 {
     // always have self in arc
     if (obj == this)
         return true;
 
     // move arc to range 0.. 2*pi
-    arc = NormalizeOrientation(arc);
+    arc = Position::NormalizeOrientation(arc);
+
+    float angle = GetAngle(obj);
+    angle -= m_orientation;
 
     // move angle to range -pi ... +pi
-    float angle = GetRelativeAngle(obj);
+    angle = Position::NormalizeOrientation(angle);
+    if (angle > M_PI)
+        angle -= 2.0f * M_PI;
 
-    if (angle > float(M_PI))
-        angle -= 2.0f * float(M_PI);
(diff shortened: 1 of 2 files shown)


PkllonG:
Can you try revert https://github.com/azerothcore/azerothcore-wotlk/pull/22500?

PkllonG:
Can you try revert #1https://github.com/azerothcore/azerothcore-wotlk/pull/22500?

heyitsbench:
This issue existed before that commit.

sudlud:
Please add co authors :) but I see you already have this in your todo for this pr

kissingers:
As if 
bool Position::HasInArc(float arc, const Position* obj, float targetRadius) const
same as
bool Position::HasInArc(float arc, Position const* obj, float targetRadius) const

why do not change any of this define, so only port and change here, and do not change define, also do not change call for const Position* obj or  Position const* obj, so the call seem more same...

    float lborder = -1 * (arc / targetRadius);                        // in range -pi..0
    float rborder = (arc / targetRadius);                             // in range 0..pi

only change 1 file and so clear

heyitsbench:
> why do not change any of this define, so only port and change here, and do not change define, also do not change call for const Position* obj or Position const* obj, so the call seem more same...

Maybe I'm alone, but I'm very much not a fan of seeing functions that do not match the definitions in this codebase. If the way this function is being used is different between AC/TC, then the usages of it should also be changed. And if the changes to the variable names make them more accurate to how they're used, then they're also fine.

sudlud:
Also putting the const after the variable is following ac codestyle.

PR #22884, opened 2025-09-11
refactor(Core/Packets): Rewrite various calendar and complaint packets to modern class.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## Tests Performed:

This PR has been:
- [X] Has not been tested.

- [ ] Some of these packets suck, like GetCalendar we fake a CMSG_GET_CALENDAR from each world session when instance resets happen. That's no good, maybe needs a redesign to instead use a function that does the same thing as the get calendar handler. Also all these packets with arrays of data suck and I don't know how to refactor em, so it'll wait for later. I want to do the complaint system first.


src/server/game/Server/Packets/MiscPackets.h
```diff
@@ -210,6 +210,32 @@ namespace WorldPackets
...
+
+            void Read() override;
+
+            uint8 SpamType = 0; // 0 - mail, 1 - chat
+            ObjectGuid SpammerGuid;
+            uint32 Unk1 = 0;
+            uint32 Unk2 = 0;
+            uint32 Unk3 = 0;
+            uint32 Unk4 = 0;
+            std::string Description = "";
```
sudlud, line 227: could we use string_view here?

sudlud: should not really matter here for this packet type though

heyitsbench: I get compile errors trying to set it to a string_view that I don't feel like chasing atm. Not worth it IMO.

PR #22927, opened 2025-09-15
fix(DB/Creature): Verify various creature models with no verified build.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

- [X] Sniffs.

## Tests Performed:

This PR:
- [X] Has not been tested.

## How to Test the Changes:

1. Verify that any creatures listed with a change to model or scale have no visual issues.


heyitsbench:
I did take a look, but I don't see an obvious, good way of redesigning the way model data is handled to avoid the error seen here. We could mangle the probability data, but I don't want to make the data inaccurate, even if it ends up with the same functionality.

TheSCREWEDSoftware:
We probably could see if there are more than 1 index that's has 1 as probablity or higher, if any others are 0, it should warning not a error.

At least 1 correct index / visible model, good no matter if the rest are 0.

PR #22934, opened 2025-09-15
feat(DB): Add `VerifiedBuild` column for `creature_model_info`

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.


data/sql/updates/pending_db_world/cmi-verified-build.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `creature_model_info` ADD COLUMN `VerifiedBuild` MEDIUMINT NULL DEFAULT NULL AFTER `DisplayID_Other_Gender`;
+UPDATE `creature_model_info` SET `VerifiedBuild` = 0;
```
sudlud, line 2: why not make the column default `0` instead

heyitsbench: Consistency with existing VerifiedBuild columns.

sudlud (approved):

Issue #21486, opened 2025-02-17
[Zul'Drak] You Reap What You Sow -- not completable


Rorschach91:
Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/7806

![Image](https://github.com/user-attachments/assets/199f4e86-c396-475b-99dc-a6f15ceaf537)

### Current Behaviour

(https://www.wowhead.com/wotlk/quest=12685/you-reap-what-you-sow), the Ritual is not able to be started. It gives the player an error message that says it required the corpse and the prophet not engaged.

https://www.youtube.com/watch?v=KgGz3kvuZFk

You can complete the quest by command kill the prophet, then using the Quetz'lun's Ritual (https://www.wowhead.com/wotlk/item=39187/quetzluns-ritual) item, then .respawn the prophet. The rest of the event works correctly after that. 

### Expected Blizzlike Behaviour

Using the quest item on the corpse should start the event. 

### Source

https://www.youtube.com/watch?v=UhgzcHAxCAo

### Steps to reproduce the problem

1. .quest add 12685
2. .go c id 28672
3. Try to do the event without commands. 

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/076e32292f184eea6609d50d47be875859b6ec2e

### Operating system

Ubuntu 22.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-costumes](https://github.com/azerothcore/mod-costumes)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
[...]

PR #22944, opened 2025-09-16
fix(DB/gameobject): Adjust Quetz'lun's Corpse Spell Focus location

blinkysc:
Sets the spell focus to same position as gameobject 190739 or Quetz'lun's Corpse

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/21486

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. .quest add 12685
2. .go c id 28672
3. Use the item near Quetz'lun

- [ ]
- [ ]


TheSCREWEDSoftware:
It will complain because of the backticks, and im unsure if this values can be sniffed.

Nyeriah:
gameobject_template data can't be changed without sniffs, with the rare exception of trap objects, is it the case? if we don't have sniffs of it then we need to drop the go template changes

blinkysc:
> gameobject_template data can't be changed without sniffs, with the rare exception of trap objects, is it the case? if we don't have sniffs of it then we need to drop the go template changes

I just got the position and updated with this
```SQL
SELECT guid, id, position_x, position_y, position_z FROM gameobject WHERE id = 190739;
```
I can remove the go template changes for sure!

blinkysc:
I changed to 38 radius because this looked like what would be the "equivalent" when not updating gameobject_template

Nyeriah:
I'm not sure I understand, you are still changing gameobject_template protected data

blinkysc:
> I'm not sure I understand, you are still changing gameobject_template protected data

I'm dumb... Lol ok... Now just changing location "Quetz'lun's Corpse Spell Focus"

data/sql/updates/pending_db_world/rev_1758041289461617900.sql
```diff
@@ -0,0 +1 @@
+UPDATE `gameobject` SET `position_x` = 5716.26, `position_y` = -4369.34, `position_z` = 385.885 WHERE `guid` = 99747;
```
sudlud: and id = ...

chauusie:
So I found this quest *did* work, but only once per server restart?

Anyway, I tested it with this fix, and found it works however many times I try it!

sudlud (approved):

Issue #22694, opened 2025-08-15
Error 1064 When Populating Database


jrelyt93:
### Current Behaviour

For anyone who gets ERROR 1064 when using authserver.exe or worldserver.exe to populate the necessary databases (acore_auth, acore_characters, & acore_world), check your MySQL version. For whatever reason, it seems as though MySQL 9.1 has beef with AzerothCore. I had this issue and dropping down to MySQL 8.4 fixed it immediately.

.txt log of the issue included.

### Logs and Screenshots

Opening DatabasePool 'acore_auth'. Asynchronous connections: 1, synchronous connections: 1.
Could not connect to MySQL database at 127.0.0.1: Unknown database 'acore_auth'
Database "acore_auth" does not exist
Do you want to create it? [yes (default) / no]:
yes
Creating database "acore_auth"...
ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SOURCE create_table.sql' at line 1

Applying of file 'create_table.sql' to database 'acore_auth' failed! If you are a user, please pull the latest revision from the repository. Also make sure you have not applied any of the databases with your sql client. You cannot use auto-update system and import sql files from AzerothCore repository with your sql client. If you are a developer, please fix your sql query.
Failed to create database acore_auth! Does the user (named in *.conf) have `CREATE`, `ALTER`, `DROP`, `INSERT` and `DELETE` privileges on the MySQL server?
DatabasePool Login NOT opened. There were errors opening the MySQL connections. Check your log file for specific errors

### Operating System

Windows 10

### Dependencies & versions

MySQL 9.1, MySQL 8.4

### Commit

azerothcore-wotlk main

TheSCREWEDSoftware:
@jrelyt93 even this being a mysql version error, please provide a commit (if you run your worldserver `.server debug` and copy that content and paste it here or change the description "commit" field

LuckyLuc96:
in windows command shell or linux terminal you can also type "git log" to see the last commit as long as you are in the right directory

<img width="934" height="330" alt="Image" src="https://github.com/user-attachments/assets/c2f9ef3a-5217-4e90-8f51-6afeb0b1b8e0" />

jrelyt93:
Terribly sorry for the delay, I've gone ahead and updated it properly now, I think. I hope this is helpful!

PR #22949, opened 2025-09-16
fix(CORE/DBUpdater): Use stdin instead of -e SOURCE

arsinspace:
DBUpdater previously always used `-e "SOURCE file.sql"`, which is unsupported in non-interactive mode and caused errors like:

ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SOURCE ...'

- Changed DBUpdater<T>::ApplyFile to pass the SQL file via stdin
- Removed usage of `-e SOURCE` when stdin is provided
- Verified that SQL updates apply successfully on macOS/Homebrew MySQL as well as on Linux

Error 2025-09-11 в 01 18 02

### DIFF
--- a/src/server/database/Updater/DBUpdater.cpp
+++ b/src/server/database/Updater/DBUpdater.cpp
@@ -496,17 +496,13 @@ void DBUpdater<T>::ApplyFile(DatabaseWorkerPool<T>& pool, std::string const& hos
     if (ssl == "ssl")
         args.emplace_back("--ssl-mode=REQUIRED");
 
-    // Execute sql file
-    args.emplace_back("-e");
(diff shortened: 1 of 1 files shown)


Mkps:
Tested.
Used to have this ERROR 1064 (42000) with my config using MySQL 9.4. This PR fixes it.

## BEFORE - Fails on db update:

worldserver_before

## AFTER - Successfully started after applying db update:

### Authserver:
authserver_after

### Worldserver:
worldserver_after

Tereneckla:
also closes https://github.com/azerothcore/azerothcore-wotlk/issues/22694

sudlud:
> also closes #22694

i added it to the pr description

PR #22969, opened 2025-09-18
feat(Core/Config): Infinite ammo config setting

aradep:
## Changes Proposed:
Adds a worldserver.conf setting to disable ammo consumption for ranged attacks and thrown weapons.

https://github.com/user-attachments/assets/1824f12f-9706-441a-88f3-e42500f677ed

## Tests Performed:
Bow/crossbow/gun - Autoshot
Bow/crossbow/gun - ranged spell
Thrown weapon - Throw
Thrown weapon - ranged spell (Deadly Throw)

## How to Test the Changes:
Make a hunter and shoot some stuff.

### DIFF
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -5400,15 +5400,16 @@ void Spell::TakeAmmo()
                 // decrease durability for non-stackable throw weapon
                 m_caster->ToPlayer()->DurabilityPointLossForEquipSlot(EQUIPMENT_SLOT_RANGED);
             }
-            else
+            else if (!sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO))
             {
                 // decrease items amount for stackable throw weapon
-                uint32 count = (sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO) ? 0 : 1);
+                uint32 count = 1;
                 m_caster->ToPlayer()->DestroyItemCount(pItem, count, true);
             }
         }
-        else if (uint32 ammo = m_caster->ToPlayer()->GetUInt32Value(PLAYER_AMMO_ID))
-            m_caster->ToPlayer()->DestroyItemCount(ammo, sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO) ? 0 : 1, true);
+        else if (!sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO))
+            if (uint32 ammo = m_caster->ToPlayer()->GetUInt32Value(PLAYER_AMMO_ID))
+                m_caster->ToPlayer()->DestroyItemCount(ammo, 1, true);
     }
 }
 


TheSCREWEDSoftware:
If the ammo is set ot infinite, how the physicall item behave in banks or guild banks (if stackable and splitting)?

aradep:
Items behave the same. The only change is that when the core says "remove 1 of the ammos" if this setting is enabled 0 ammos are removed instead.

kissingers:
Is the damage from ammo still applied?

sogladev:
Currently, with the config value enabled, players will still lose durability for non-stackable thrown weapons. This doesn't seem right to me. Can this be disabled as well and simplify the code as: 

- https://github.com/azerothcore/azerothcore-wotlk/blob/f884f6489611b99d9747db3507040c9cad5696c0/src/server/game/Spells/Spell.cpp#L5397
```cpp
void Spell::TakeAmmo()
{
    if (sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO))
        return;
```

aradep:
Bows/crossbows/guns also continue to lose durability with the setting enabled so I think it's best to keep it in for thrown weapons too. Having an ammo consumption setting affect durability could lead to confusion and there's already a section of the .conf to modify durability rates.

I do prefer your one-line solution but we can't return at the beginning of TakeAmmo since thrown weapon durability happens at the end of it.

src/server/game/Spells/Spell.cpp
```diff
@@ -5403,12 +5403,12 @@ void Spell::TakeAmmo()
             else
             {
                 // decrease items amount for stackable throw weapon
-                uint32 count = 1;
+                uint32 count = (sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO) ? 0 : 1);
                 m_caster->ToPlayer()->DestroyItemCount(pItem, count, true);
             }
```
sogladev:
```suggestion
            else if (!sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO))
            {
                // decrease items amount for stackable throw weapon
                uint32 count = 1;
                m_caster->ToPlayer()->DestroyItemCount(pItem, count, true);
            }
```

src/server/game/Spells/Spell.cpp
```diff
@@ -5403,12 +5403,12 @@ void Spell::TakeAmmo()
...
         }
         else if (uint32 ammo = m_caster->ToPlayer()->GetUInt32Value(PLAYER_AMMO_ID))
-            m_caster->ToPlayer()->DestroyItemCount(ammo, 1, true);
+            m_caster->ToPlayer()->DestroyItemCount(ammo, sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO) ? 0 : 1, true);
```
sogladev:
```suggestion
        else if (!sWorld->getBoolConfig(CONFIG_ENABLE_INFINITEAMMO))
            if (uint32 ammo = m_caster->ToPlayer()->GetUInt32Value(PLAYER_AMMO_ID))
                m_caster->ToPlayer()->DestroyItemCount(ammo, 1, true);
```

sogladev (changes requested):
Thanks for clearing up my previous concern. Before approving, I suggest refactoring to split the ammo logic and the config check. There's also no need to call DestroyItemCount with a count of 0, so let's skip it if the config is enabled

src/server/game/Spells/Spell.cpp
aradep: Cant apply these automatically due to " Applying suggestions on deleted lines is currently not supported. " so i'll do a commit. Thanks for the improvement!

sogladev (approved):

Issue #22992, opened 2025-09-21
[Quest] Infernal Attackers not appearing for "Besieged!"


Teknishun7:
### Original CC Issue
https://github.com/chromiecraft/chromiecraft/issues/8176

### Triage notes
The https://github.com/azerothcore/azerothcore-wotlk/commit/5dbcdfcbd55adda4506fe2334a456abea79a7b90 commit seems to be the culprit. Upon commenting the lines and testing the Infernal Attackers now respawn properly when coming back to them after zoning out a while.

[Test Footage]
https://www.youtube.com/watch?v=V046h5P6V7c

Thanks to @amed80 for finding the potential issue.

### Current Behaviour
No [Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) are spawning outside Wildhammer Stronghold, preventing the quest from progressing.

### Expected Blizzlike Behaviour
[Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) should be appearing outside Wildhammer Stronghold

### Source
Horde version (I believe also broken): https://www.wowhead.com/tbc/quest=10595/besieged
Alliance version: https://www.wowhead.com/tbc/quest=10562/besieged

### Steps to reproduce the problem
.tele shadowmoon
.quest add 10562     (Alliance)
.quest add 10595     (Horde)
Head to Wildhammer Stronghold
The [Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) should be there.
Go to Terokkar Forest to zone out.
Wait like 2 to 5 mins
Come back to the [Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) and they will be now completely gone and not respawning

### AC rev. hash/commit
https://github.com/chromiecraft/azerothcore-wotlk/commit/2a3075504ebc856c23aba49da1bb8a6f2f33a22d

### Operating system
Ubuntu 22.04

### Modules
    [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
    [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
    [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
    [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
    [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
[...]

PR #23001, opened 2025-09-22
Fix infernal attacker respawn

Teknishun7:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/22992

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. .tele shadowmoon
2. .quest add 10562 (Alliance)
3. .quest add 10595 (Horde)
4. Head to Wildhammer Stronghold
5. The [Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) should be there.
6. Go to Terokkar Forest to zone out.
7. Wait like 2 to 5 mins
8. Come back to the [Infernal Attacker](https://www.wowhead.com/wotlk/npc=21419/infernal-attacker) and they will be now completely gone and not respawning

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Outland/zone_shadowmoon_valley.cpp
+++ b/src/server/scripts/Outland/zone_shadowmoon_valley.cpp
@@ -144,7 +144,7 @@ public:
 
         void SummonInfernal()
         {
-            Creature* infernal = me->SummonCreature(NPC_INFERNAL_ATTACKER, me->GetPositionX(), me->GetPositionY(), ground + 0.05f, 0.0f, TEMPSUMMON_TIMED_OR_DEAD_DESPAWN, 60000);
+            Creature* infernal = me->SummonCreature(NPC_INFERNAL_ATTACKER, me->GetPositionX(), me->GetPositionY(), ground + 0.05f, 0.0f, TEMPSUMMON_DEAD_DESPAWN, 0);
             infernalGUID = infernal->GetGUID();
         }
 


Teknishun7:
Tested extensively and works. Zoning out and coming back they will still respawn even after half an hour, staying near them for another half hour and I dont see any abnormality.

**PLEASE REVIEW CAREFULLY MY CODE AS I HAVE LITTLE EXPERIENCE.**

Nyeriah:
An issue I can see with this is infernal overpopulation, as they won't ever despawn now as opposed to the previous behaviour, as before they despawned hourly, this can lead to issues over time

Teknishun7:
I couldnt observe this in over an hour of testing. They only spawn 1 at the time at location. If theres no horde or alliance NPC to fight them they just stay still. They dont spawn on top of one another, but youre right that I removed that fail-safe. Do you have any idea how to tackle this then? @Nyeriah

Takenbacon:
Looks like to me the hook is only called when the inferno is killed (https://github.com/azerothcore/azerothcore-wotlk/pull/23001/files#diff-9f27842ed9abe62b865f2e941be9aed899e9dddf275730c6c8b5ec3c3524c038R213), it probably doesn't trigger a new inferno spawn when they despawn via timer instead which is probably causing the bug.

The PR seems like a reasonable fix assuming it works correctly, just removes the timed despawn and makes it only spawn a new inferno when killed, thus always calling the hook properly.

Nyeriah:
Looking into this, I believe the issue started happening because the infernals were not active before (but now they are, because tempsummons are always active). So, before, they wouldn't despawn unless there were players about, which effectively meant that they would in most cases only do so on death. But now the timer does run out without any player around to kill them, so they despawn and a new set never shows up.

So, with the considerations above in mind, this PR seems fine and can be merged

PR #23008, opened 2025-09-22
Creature: Fix crash due to recursive call for help

killerwife:
When testing PR 23005 the following crash occured:

Mob in hellfire peninsula spawned in. Called los check, then:

SMART_ACTION_CALL_FOR_HELP                      = 39

And then so many calls for help occured (20+) that the stack got smashed. The PR aims to prevent a cyclic call for help which chains like:

1 -> 2 -> 3 -> 4 ->5 ... potentially up to infinity based on npc proximity.


Nyeriah:
Nice catch! Btw, slightly OT, but what's the difference between CallForHelp and CallForAssistance?

killerwife:
Currently in acore, call for help is an explicit smart action

Call for assistance is the start of combat call for surrounding mobs to help the caller.

FrancescoBorzi:
thank you very much @killerwife and thanks @Rorschach91 for testing this PR

Issue #1430, opened 2019-02-08
Bug timer deeprum tram


comix1988:
<DESCRIPTION:

the wagons of the deeprum tram have incorrect timer

EXPECTED BLIZZLIKE BEHAVIOUR:

When a wagon is at the ironforge station, the other is at the stormwind station.The wagon meet halfway.

CURRENT BEHAVIOUR:

When a wagon starts from one station, the other is already on its way. This causes more waiting time in the station.

STEPS TO REPRODUCE THE PROBLEM:

.go -24.9 29.5 -4.3 369
And wait

BRANCH(ES):
staging

AC HASH/COMMIT:
commit [987095e](https://github.com/azerothcore/azerothcore-wotlk/commit/987095e2c9da04b6ac9efcd2639a012cc72552d7)

OPERATING SYSTEM:
linux

MODULES:
OTHER CUSTOMIZATIONS:

Viste:
can you test it also on master branch because of this https://github.com/azerothcore/azerothcore-wotlk/commit/cf2ffbe5612ea70a3b6bae961fd830b2933240e7

FrancescoBorzi:
@Deku the branch of the report is `staging`, have you tried to replicate it on the same branch?

comix1988:
I have also tested on the master branch but the bug is still there.

https://youtu.be/XIlkrm3iiy8

JamesGent:
Can confirm still a thing.

![Desktop 28-11-2020 17-09-31-722](https://user-images.githubusercontent.com/61194190/100521580-7dca4800-319c-11eb-8197-83fccec0043b.png)

Meltie2013:
This has been a known issue for years and no one found a good resolution yet. I'll poke around and see if I can port a change I did back in 2018 on a project I personally work on.

Rushor:
@Meltie2013 you found sth already?

Meltie2013:
> @Meltie2013 you found sth already?

I haven't had much time to view this. PL is getting in the way a lot. The problem I seen within my own project was the grids on the map (369) were not loaded all the way, so the objects were set to inactive state.

You can move the data from the database to the core to fix the problem, but the issues with the grids there would still remain. But both sides of the system will work once both ends of the system are triggered. e.g. Iron Forge and Stormwind entrances

Warglaive:
[...]

PR #23009, opened 2025-09-22
fix(Core): Deeprun Tram Subway timer/syncronization by loading grid

zeb139:
## Changes Proposed:

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/1430

## SOURCE:

https://www.youtube.com/watch?v=3Lo5x1S7s6E

## Tests Performed:
I'm able to build without errors from the latest azerothcore-wotlk master. Entering and exiting the tram from either side causes no issue. Riding both trams is also working fine.  

This issue can also be addressed by enabling IsPreloadedContinentTransport.Enabled in the worldserver.conf.   
I took some time to compare memory usages by the `worldserver` under a few configurations. Here were my results:  
**WITHOUT** my changes:  
IsContinentTransport.Enabled = 1  
IsPreloadedContinentTransport.Enabled = 0
The world server was reported as using ~1440 MB of memory   

**WITH** my changes:  
IsContinentTransport.Enabled = 1  
IsPreloadedContinentTransport.Enabled = 0
The world server was reported as using ~1620 MB  MB of memory  

I think this difference is likely tolerable by most servers.  
<br>

This PR has been:

## How to Test the Changes:
After building the changes, players can enter the Deeprun Tram from the Stormwind side and observe both sides of tram arrive and depart on a consistent timer (~1 min between sides) with all 3 Subway carts in sync.  
Previously, the Ironforge side tram would never load unless a player got close enough to load it, in which case 2 of the Ironforge-side carts would be in sync, with the 3rd trailing out of order.

1. `.go xyz -17 2516 -4 369 4.7`
2. Wait and observe both trams

### DIFF
--- a/src/server/game/Maps/TransportMgr.cpp
+++ b/src/server/game/Maps/TransportMgr.cpp
@@ -449,8 +449,8 @@ void TransportMgr::SpawnContinentTransports()
     }
 
     // Preloads Deeprun Tram to fix issues with Subway carts syncronization
+    /// @todo: This is a temporary workaround. Consider removing TransportMgr::PreloadGridsFromQuery() as part of fix.
     /**
-     @todo: This is a temporary workaround. Consider removing TransportMgr::PreloadGridsFromQuery() as part of fix.
      Takenbacon: "In the long run the most likely ideal fix would be to always spawn all transport types (and thus loading their grid) on map creation"
      See https://github.com/azerothcore/azerothcore-wotlk/pull/23009 for more details.
     */


src/server/game/Maps/TransportMgr.cpp
```diff
@@ -448,31 +448,40 @@ void TransportMgr::SpawnContinentTransports()
         LOG_INFO("server.loading", ">> Spawned {} continent motion transports in {} ms", count, GetMSTimeDiffToNow(oldMSTime));
     }
 
+    // Preloads Deeprun Tram to fix issues with Subway carts syncronization
+    PreloadGridsFromQuery("SELECT map, position_x, position_y FROM gameobject g JOIN gameobject_template t ON g.id = t.entry WHERE t.type = 11 AND g.map = 369", count);
```
heyitsbench, line 457: This screams hackfix to me.

zeb139: I agree! I'm happy to make changes, but I'm not sure what the best approach is.  
  
The logic here is that other transports still seem to work without IsPreloadedContinentTransport.Enabled (except Deeprun Tram), so by making an exception we can  
1. Fix the tram
2. Prevent loading all other grids

heyitsbench: I know @Takenbacon mentioned that he believes all transports should be loaded and spawned on startup, and not just motion transports as it currently is.

Pinged to make sure they're aware of this PR.

Takenbacon: In the long run the most likely ideal fix would be to always spawn all transport types (and thus loading their grid) on map creation, but last time I checked that isn't super straight forward if I recall right at the moment.

In the interim I don't see any real harm in a temporary fix like the PR until someone goes about trying to properly get transports to spawn at map load.

zeb139: From my (limited) testing, I found these differences in memory usage of worldserver:  
IsContinentTransport.Enabled = 0, IsPreloadedContinentTransport.Enabled = 0, without PR : ~840MB (tram broken)  
IsContinentTransport.Enabled = 1, IsPreloadedContinentTransport.Enabled = 0, without PR : ~1440 MB (tram broken)  
IsContinentTransport.Enabled = 1, IsPreloadedContinentTransport.Enabled = 0, WITH PR     : ~1620 MB (tram working)  
IsContinentTransport.Enabled = 1, IsPreloadedContinentTransport.Enabled = 1, without PR  : ~2020 MB (tram working)  

Maybe the Occam's razor approach of enabling IsPreloadedContinentTransport.Enabled by default is the most appropriate solution? I'm not sure what concerns there are about adding an extra ~600 MB of memory though.  
Could we document the Deeprun Tram issue in the worldserver.conf documentation in that case?

zeb139:
I tried digging around a bit to see where we might handle Transports, and found that if we move the "preload grids for continent static transports" logic into MapMgr::CreateBaseMap()  (in order to satisfy the condition of loading transports on map creation), it still works* , but the worldserver then had many errors in the console like:  
`WaypointMovementGenerator::LoadPath: creature <creature name> doesn't have waypoint path id: <path id>`  
So maybe there's something worth looking into there?   

I also spent some time looking at how other cores handle this and found this closed issue from cmangos: https://github.com/cmangos/issues/issues/2339 . One of [the relevant commits](https://github.com/cmangos/mangos-tbc/commit/a70612d429d3c64bcb9d758d4b7edeb6de28f659#diff-fa206079ffda1fff89d106f9a6275b3614c70024ece85e7ec1d8e5f1a34d3b08) seems to handle Transports in ObjectMgr::LoadGameObjects(), and the AzerothCore code already matches this implementation closely. Not sure if this is quite in line with what you had suggested though, @Takenbacon .  

@killerwife sorry to bother, but I was curious if I could solicit your input on this issue since it seems like you've already solved this problem once?

killerwife:
Ye i made deeprun tram respect visibility 
, simulate xyz properly and active

Teknishun7:
Let me know if this needs testing.

zeb139:
Would it be helpful to add some tag like "@hackfix" or "@todo" near my changes so that someone in the future can identify a need for improvement?

sudlud:
Todo would probably be good, so this would be a fix until the mechanics exist to fix this thoroughly?

zeb139:
> Todo would probably be good, so this would be a fix until the mechanics exist to fix this thoroughly?

That's right. I tried looking for a way to fix this properly, but I think that's probably beyond my capability.
According to Takenbacon:
> In the long run the most likely ideal fix would be to always spawn all transport types (and thus loading their grid) on map creation, but last time I checked that isn't super straight forward if I recall right at the moment.  

I'll add extra TODO comments soon

zeb139:
Added TODO. Let me know if you need anything changed/added

src/server/game/Maps/TransportMgr.cpp
```diff
@@ -448,31 +448,45 @@ void TransportMgr::SpawnContinentTransports()
         LOG_INFO("server.loading", ">> Spawned {} continent motion transports in {} ms", count, GetMSTimeDiffToNow(oldMSTime));
     }
 
+    // Preloads Deeprun Tram to fix issues with Subway carts syncronization
+    /*
+        TODO: This is a temporary workaround. Consider removing TransportMgr::PreloadGridsFromQuery() as part of fix.
```
sudlud: I think it should be @todo for doxygen

killerwife: doxygen is \todo

zeb139: Sorry about that. How does the latest change look? I can see doxygen syntax highlighting with `@todo`

sudlud: #14966 says it should be `/// @todo`

src/server/game/Maps/TransportMgr.cpp
```diff
@@ -448,31 +448,45 @@ void TransportMgr::SpawnContinentTransports()
         LOG_INFO("server.loading", ">> Spawned {} continent motion transports in {} ms", count, GetMSTimeDiffToNow(oldMSTime));
     }
 
+    // Preloads Deeprun Tram to fix issues with Subway carts syncronization
+    /**
+     @todo: This is a temporary workaround. Consider removing TransportMgr::PreloadGridsFromQuery() as part of fix.
```
sudlud:
```suggestion
    // Preloads Deeprun Tram to fix issues with Subway carts syncronization
    /// @todo: This is a temporary workaround. Consider removing TransportMgr::PreloadGridsFromQuery() as part of fix.
    /**
```

src/server/game/Maps/TransportMgr.cpp
zeb139: Good find!

Issue #6427, opened 2021-06-19
Mage's blink while swimming/falling


Stump876:
Original Chromiecraft report https://github.com/chromiecraft/chromiecraft/issues/893

#### WHAT CLIENT DO YOU PLAY ON?

- [ ] enGB
- [x] enUS
- [ ] other (specify)

##### FACTION

- [x] Alliance
- [x] Horde

##### CONTENT PHASE:

- [x] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59

##### CURRENT BEHAVIOUR:

Behaviour 1: Casting Blink while falling teleports the caster 20 yards straight up.

https://user-images.githubusercontent.com/68164076/122629815-a6e8f200-d074-11eb-84ab-3aaf81e95c80.mp4

Behaviour 2: Casting Blink while swimming teleports the caster to the bottom of the body of water.

https://user-images.githubusercontent.com/68164076/122629825-afd9c380-d074-11eb-8e69-dc4c6c0cb582.mp4

##### EXPECTED BLIZZLIKE BEHAVIOUR:

Issue 1: Casting Blink while falling should teleport the caster safely to the ground as long as they are within range. 
Issue 2: Casting Blink while swimming should only teleport the caster forward, not down to the bottom of the water.

##### SOURCE:

Issue 1: https://www.youtube.com/watch?v=nlrEsEKCU3s
Issue 2: https://youtu.be/_ddFox7jwn4?t=145

##### STEPS TO REPRODUCE THE PROBLEM:

Issue 1:
1. Jump off any high place (e.g. Stormwind Mage Tower)
2. Cast Blink just before hitting the ground

Issue 2:
1. Cast Blink while swimming in any deep body of water

##### EXTRA NOTES:

N/A

##### AC HASH/COMMIT:

https://github.com/chromiecraft/azerothcore-wotlk/commit/34da0cda5195dede48cb0406b23248330d6249cb

##### OPERATING SYSTEM:

Ubuntu 20.04

##### MODULES:

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
[...]

PR #23012, opened 2025-09-23
fix(Core/Map-Grids): Cylinder-plane/Upside-down-Pyramid-plane intersection (height calcs)

VG-prog:
Accurate ground height calculation (grid + VMAP + dynamic) — inspired by https://github.com/azerothcore/azerothcore-wotlk/discussions/3971 and @Pursche

## Changes Proposed

This PR introduces a new **accurate ground height calculation** pipeline for terrain, VMAP/static collision and dynamic objects.

The old server-side height logic treated the unit as a single point over a terrain triangle or collision ray. This PR keeps the legacy point-height behavior as the base plane, but adds **footprint-aware ground settlement** using the unit radius. This better matches the expected player/unit collider behavior on slopes, edges, caves, bridges, rocky terrain, WMOs and dynamic objects.

The system supports:

- **Circle / cylinder-plane** footprint settlement.
- **Square / oriented upside-down-pyramid** footprint settlement.
- Legacy fallback when accurate data is unavailable or invalid.
- Accurate grid height from legacy-exact terrain planes.
- Accurate VMAP/static height from the real collision triangle normal.
- Accurate dynamic-object height from the real collision triangle normal.
- Global accurate height usage for `Unit` callers through `WorldObject::GetMapHeight()`.
- Blink ground probing through the new accurate height API.
- Blink swept collision probing using the caster radius and collision height.

This makes the improvement directly testable in gameplay through Blink and through `.gps`.

---

### Core API additions

- `GridTerrainData::GetHeightAccurate(x, y, radius, shape, yaw, squareBlend, slopeClamp, gradientMode, normalEps)`
  - Supports **Circle** footprint: cylinder-plane settlement.
  - Supports **Square** footprint: oriented upside-down-pyramid settlement.
  - Square mode is oriented using the unit facing (`yaw`).
  - `SquareBlend` allows blending square and circle behavior.
  - `SlopeClamp` limits gradient magnitude to avoid extreme vertical lifts.
  - `GradientMode` supports plane-exact and least-squares slope calculation.
  - `NormalEps` handles invalid or near-vertical terrain faces safely.

- `GridTerrainData::SampleHeights()`
  - Unified accessor for float, uint8, uint16 and flat-height map data.
  - Flat grids now correctly return `gridHeight` instead of invalid height.

- `Map::GetGridHeightAccurate()`
  - Reads accurate grid height using the configured shape, blend, clamp and gradient mode.
  - Uses legacy-exact terrain triangle selection and plane coefficients.
  - Uses `radius = 0.0f` accurate height for point-height validation to keep coordinate handling consistent.

- `StaticMapTree::getHeightAndNormal()`
  - Returns both collision height and the real hit normal from VMAP/static geometry.

- `StaticVMapCollisionData::getHeightAndNormal()`
  - Exposes VMAP/static height + normal through map collision data.
  - Converts VMAP internal-space normals back to world-space normals.

- `Map::GetVMapHeightAccurate()`
  - Uses the real VMAP/static collision triangle normal instead of lateral finite-difference sampling.
  - Applies the same footprint support logic as grid terrain.
  - Respects the caller-provided `maxSearchDist`.

- `DynamicMapTree::getHeightAccurate(...)`
  - Uses the real dynamic-object hit normal.
  - Applies the same footprint support logic as grid and VMAP/static collision.

- `WorldModel`, `GroupModel`, `ModelInstance`, `GameObjectModel`
  - Collision ray APIs now optionally propagate the hit triangle normal.
  - Existing callers can keep using the old behavior without requesting a normal.
  - Normal-aware raycasts use detailed triangle intersection only when a normal is requested, preserving the legacy fast path for normal raycasts.

- `Map::GetHeightAccurate()`
  - Merges accurate grid + accurate VMAP/static collision with accurate dynamic-object height.
  - Preserves legacy fallback behavior.
  - Uses point-height validation for search visibility, then returns the footprint-aware lifted height.

- `WorldObject::GetGroundProbeRadius()`
  - Returns the effective unit footprint radius using `UNIT_FIELD_BOUNDINGRADIUS`.
  - Falls back to a safe default for non-units.

- `WorldObject::GetMapHeightAccurate()`
  - Uses the unit probe radius by default.
  - Applies `Height.Accurate.RadiusScale`.
  - Probes using the object's orientation.
  - Uses a small safe vertical search lift instead of large collision-height offsets, reducing false hits on ceilings or upper floors.
  - Falls back to legacy `GetHeight()` if the accurate query cannot resolve a valid surface.

- `WorldObject::GetMapHeight()`
  - Now routes `Unit` callers through `GetMapHeightAccurate()`.
  - Keeps legacy point-height behavior for non-unit callers.

---

### Blink

`TARGET_DEST_CASTER_FRONT_LEAP` now uses accurate height probing through `Map::GetHeightAccurate()`.

Blink ground samples now use:

- caster footprint radius,
- caster orientation,
- accurate grid height,
- accurate VMAP/static collision height,
- accurate dynamic-object height,
- legacy `GetHeight()` fallback when an accurate sample is invalid.

Blink collision probing was also improved. Instead of checking only a single point/ray path, Blink now performs a swept approximation using:

- center probe,
- cardinal radius probes,
- diagonal radius probes,
- lower/middle/upper body-height probes,
- static collision checks,
- dynamic collision checks.

This makes Blink much more robust near walls, slopes, edges, bridges, caves and VMAP/dynamic geometry.

Other terrain-sensitive movement/spell systems such as Charge, Shadowstep, Leap, short teleports or summon-position correction can use the same API.

---

## Global Unit Height Integration

`WorldObject::GetMapHeight()` now redirects `Unit` callers to `WorldObject::GetMapHeightAccurate()`.

This makes the accurate height system available globally for unit ground settlement while preserving legacy behavior for non-unit objects.

Additional direct integrations included in this PR:

- Arena height correction in Ring of Valor.
- Flying creature corpse/load adjustment.
- Unit hover enable/disable ground checks.
- Blink destination and path probing.

---

## Configuration Options (`worldserver.conf.dist`)

- `Height.Accurate.Shape`
  - `0` = circle/cylinder footprint, default and recommended for production.
  - `1` = square/upside-down-pyramid footprint.

- `Height.Accurate.RadiusScale`
  - Multiplier applied to the unit footprint radius.

- `Height.Accurate.SquareBlend`
  - `[0..1]` blend between square and circle behavior.
  - `0.00` = pure square.
  - `1.00` = pure circle.
  - Recommended QA range for square mode: `0.15` to `0.25`.

- `Height.Accurate.SlopeClamp`
  - Slope magnitude clamp.
  - Default: `4.0`.
  - `0.0` disables clamping.
  - Prevents extreme lifts on near-vertical faces, VMAP edges and dynamic-object surfaces.

- `Height.Accurate.GradientMode`
  - `0` = plane-exact grid gradient, default.
  - `1` = least-squares grid gradient.

- `Height.Accurate.NormalEps`
  - Degenerate/unsafe grid terrain threshold.

- `Height.Accurate.VMap.Enable`
  - `1` = enable accurate VMAP/static collision settlement.
  - `0` = legacy VMAP point ray.

- `Height.Accurate.VMap.Delta`
  - Kept for compatibility/future fallback tuning.
  - The current normal-based VMAP path does not require lateral gradient sampling.

- `Height.Accurate.Dynamic.Enable`
  - `1` = enable accurate dynamic-object settlement.
  - `0` = legacy dynamic point ray.

- `Height.Accurate.Dynamic.Delta`
  - Kept for compatibility/future fallback tuning.
  - The current normal-based dynamic path does not require lateral gradient sampling.

---

## Implementation Details

- Correct `floor()`-based terrain indexing for edge and negative coordinates.
- Terrain triangle selection matches legacy `getHeight*` diagonal splitting exactly.
- Grid accurate height uses the same legacy plane coefficients as `getHeightFromFloat`, `getHeightFromUint8` and `getHeightFromUint16`.
- Flat-height grids are supported.
- Grid code remains independent of `sWorld`; configuration is provided through `Map`.
- VMAP/static accurate height now uses the real hit triangle normal.
- Dynamic accurate height now uses the real hit triangle normal.
- VMAP/dynamic accurate height no longer relies on 4 lateral gradient samples in the normal-based path.
- Circle footprint uses L2 gradient support.
- Square footprint uses rotated L1 gradient support.
- Optional slope clamp rescales the gradient vector before applying the footprint offset.
- Accurate grid search validation uses point-height accurate mode (`radius = 0.0f`) to avoid false rejection on slopes.
- Fallbacks preserve legacy behavior when terrain faces, VMAP hits or dynamic hits are invalid.
- `WorldObject::GetMapHeightAccurate()` falls back to legacy `GetHeight()` when accurate height cannot resolve a valid result.
- The accurate height formula remains monotonic over the legacy base height: it does not lower units below the point-height plane.
- `.gps` still reports legacy height separately, so legacy vs accurate values can be compared directly.

---

## Commands

`.gps` was extended to print:

- legacy ground/floor height,
- accurate grid height,
- accurate map height,
- probe radius,
- shape,
- blend,
- radius scale,
- slope clamp,
- gradient mode,
- normal epsilon,
- VMAP accurate settings,
- dynamic accurate settings.

---

## Issues Addressed

- Reduces falling/sinking through terrain on slopes and edges.
- Improves ground settlement on terrain, VMAP/static collision and dynamic objects.
- Replaces point-only ground probing with footprint-aware height calculation.
- Removes the need for fragile fixed Z-offset workarounds in terrain-sensitive spell logic.
- Makes Blink use the new accurate height path directly.
- Improves Blink path collision by probing an approximate swept volume instead of a single center ray.
- Uses real VMAP/dynamic hit normals instead of approximating collision slopes with multiple nearby rays.
- Provides tunable QA/production configuration while keeping safe production defaults.

---

## Source / Background

- Based on NovusCore discussions and the reference work by @Pursche.
- Inspired by AzerothCore discussion https://github.com/azerothcore/azerothcore-wotlk/discussions/3971.
- Uses plane contact, gradient support functions, barycentric terrain interpolation and convex footprint support math.
- Keeps compatibility with AzerothCore/TrinityCore legacy terrain height behavior where fallback is required.

---

## Tests Performed / Validation

- `.gps` shows legacy vs accurate grid/map heights.
- Verified that flat terrain remains close to legacy height.
- Verified that slopes and terrain edges produce a slightly safer lifted settlement height.
- Verified caves, bridges, roads, rocky terrain and VMAP surfaces with accurate height enabled.
- Verified behavior with different `UNIT_FIELD_BOUNDINGRADIUS` values.
- Verified Grid vs Map vs Dynamic consistency.
- Verified that `SlopeClamp` prevents extreme vertical lifts.
- Migrated Blink ground probing to `GetHeightAccurate()` and tested Blink locally.
- Stress-tested Blink repeatedly on mountainous/steep terrain without reproducing falling below textures.
- Verified that flat terrain keeps the same practical height while steep terrain shows the expected accurate-height difference.

image
image

---

## How to Test the Changes

1. Compile and run the core with this PR.
2. Log in with a GM character.
3. Use `.gps` on different terrain:
   - flat ground,
   - slopes,
   - cave entrances,
   - bridges,
   - road edges,
   - rocks,
   - WMOs,
   - dynamic-object surfaces.
4. Compare legacy height vs accurate height.
   - Flat ground should remain very close to legacy.
   - Slopes and edges should report a slightly safer accurate height.
   - VMAP/dynamic surfaces should remain consistent and avoid excessive lifts.
5. Test Blink:
   - blink uphill,
   - blink downhill,
   - blink near road edges,
   - blink near cave/bridge edges,
   - blink near VMAP surfaces,
   - blink around water transitions,
   - blink near static and dynamic collision.
6. Optional production-like configuration:
   ```ini
   Height.Accurate.Shape = 0
   Height.Accurate.RadiusScale = 1.0
   Height.Accurate.SquareBlend = 0.2
   Height.Accurate.SlopeClamp = 4.0
   Height.Accurate.GradientMode = 0
   Height.Accurate.VMap.Enable = 1
   Height.Accurate.Dynamic.Enable = 1
   ```
7. Optional square-footprint QA:
   ```ini
   Height.Accurate.Shape = 1
   Height.Accurate.SquareBlend = 0.2
   Height.Accurate.SlopeClamp = 4.0
   ```

---

## Notes

- Default mode is Circle/cylinder footprint, intended as the safe production default.
- Square mode is available for QA and experimentation.
- `Height.Accurate.VMap.Delta` and `Height.Accurate.Dynamic.Delta` are currently compatibility/future fallback settings because VMAP/dynamic now use real hit normals.
- Non-unit `WorldObject::GetMapHeight()` callers keep legacy behavior.
- Unit `WorldObject::GetMapHeight()` callers use accurate height with legacy fallback.


heyitsbench:
CC: @walkline

src/server/game/Maps/Map.cpp
```diff
@@ -1175,6 +1175,47 @@ float Map::GetHeight(float x, float y, float z, bool checkVMap /*= true*/, float
...
+        const float tol = std::max(0.1f, 0.5f * radius); // tolerancia dinámica en función del tamaño del que pisa
+        if (G3D::fuzzyGe(z, gridHeight - tol))
+            mapHeight = gridHeight;
+    }
+
+    float vmapHeight = VMAP_INVALID_HEIGHT_VALUE;
+    if (checkVMap)
+    {
+        VMAP::IVMapManager* vmgr = VMAP::VMapFactory::createOrGetVMapManager();
+        vmapHeight = vmgr->getHeight(GetId(), x, y, z, maxSearchDist);   // VMAP: mismo criterio que GetHeight
```
heyitsbench: Non-English comments.

VG-prog: Resolved!

Pursche:
At an isolated glance, this LOOKS like the proper fix for collision testing + resolve to me. If this gets properly tested to not have any annoying side effects, this would be a major improvement.

src/server/game/World/WorldConfig.cpp
```diff
@@ -282,6 +282,14 @@ void WorldConfig::BuildConfigCache()
...
     SetConfigValue<float>(CONFIG_CHANCE_OF_GM_SURVEY, "GM.TicketSystem.ChanceOfGMSurvey", 50.0f);
 
+    SetConfigValue<uint32>(CONFIG_HEIGHT_ACCURATE_SHAPE, "Height.Accurate.Shape", 0);
+    float rawSc = sConfigMgr->GetOption<float>("Height.Accurate.RadiusScale", 1.0f);
+    float clampedSc = std::clamp(rawSc, 0.05f, 4.0f);
+    SetConfigValue<float>(CONFIG_HEIGHT_ACCURATE_RADIUS_SCALE, "Height.Accurate.RadiusScale", clampedSc);
+    float rawSq = sConfigMgr->GetOption<float>("Height.Accurate.SquareBlend", 0.20f);
+    float clampedSq = std::clamp(rawSq, 0.0f, 1.0f);
+    SetConfigValue<float>(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND, "Height.Accurate.SquareBlend", clampedSq);
+
```
Tereneckla:
```suggestion
    SetConfigValue<uint32>(CONFIG_HEIGHT_ACCURATE_SHAPE, "Height.Accurate.Shape", 0, ConfigValueCache::Reloadable::Yes, [](uint32 const& value) { return value <= 1; }, "<= 1");
    SetConfigValue<float>(CONFIG_HEIGHT_ACCURATE_RADIUS_SCALE, "Height.Accurate.RadiusScale", 1.0f, ConfigValueCache::Reloadable::Yes, [](float const& value) { return value >= 0.1f && value <= 4.0f; }, ">= 0.1 and <= 4.0");
    SetConfigValue<float>(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND, "Height.Accurate.SquareBlend", 0.20f, ConfigValueCache::Reloadable::Yes, [](float const& value) { return value >= 0.0f && value <= 1.0f; }, ">= 0.0 and <= 1.0");
```

The clamp like you want it here doesn't work, SetConfigValue reads the config again and would only use clampedSc/clampedSq when not present if you don't provide a function to check the validity.

src/server/game/Entities/Object/Object.cpp
```diff
@@ -3017,6 +3017,38 @@ float WorldObject::GetMapHeight(float x, float y, float z, bool vmap/* = true*/,
...
+            radius = std::max(0.25f, u->GetFloatValue(UNIT_FIELD_BOUNDINGRADIUS));
+        }
+        else
+            radius = 0.3f;
+    }
+
+    float rScale = 1.0f;
+    if (sWorld)
+        rScale = std::max(0.1f, sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_RADIUS_SCALE));
+    radius *= rScale;
```
Tereneckla:
```suggestion
    float rScale = sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_RADIUS_SCALE);
    radius *= rScale;
```

src/server/game/Grids/GridTerrainData.cpp
```diff
@@ -615,3 +616,160 @@ LiquidData const GridTerrainData::GetLiquidData(float x, float y, float z, float
...
+
+        const float d      = -(n.x * A.x + n.y * A.y + n.z * A.z);
+        const float zPlane = (-d - n.x * P.x - n.y * P.y) / n.z;
+        const float denom  = std::max(eps, nzAbs);
+
+        if (shape == GroundFootprintShape::Square)
+        {
+            float blend = 0.0f;
+            if (sWorld)
+                blend = std::clamp(sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND), 0.0f, 1.0f);
```
Tereneckla:
```suggestion
            float blend = sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND);
```

src/server/scripts/Commands/cs_misc.cpp
```diff
@@ -621,13 +621,30 @@ class misc_commandscript : public CommandScript
...
+            blend  = sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND);
+        }
+
+        float yaw = object->GetOrientation();
+        float probeRScaled = probeR * rScale;
+        float GridZAccurate = map->GetGridHeightAccurate(object->GetPositionX(), object->GetPositionY(), probeRScaled, yaw);
+        float MapZAccurate  = object->GetMapHeightAccurate(object->GetPositionX(), object->GetPositionY(), object->GetPositionZ(), /*vmap=*/true, DEFAULT_HEIGHT_SEARCH, probeR);
+        const char* shapeStr = "circle";
+        if (sWorld && sWorld->getIntConfig(CONFIG_HEIGHT_ACCURATE_SHAPE) == 1)
+            shapeStr = "square";
```
Tereneckla:
```suggestion
        float rScale = sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_RADIUS_SCALE);
        float blend  = sWorld->getFloatConfig(CONFIG_HEIGHT_ACCURATE_SQUARE_BLEND);
        float yaw = object->GetOrientation();
        float probeRScaled = probeR * rScale;
        float GridZAccurate = map->GetGridHeightAccurate(object->GetPositionX(), object->GetPositionY(), probeRScaled, yaw);
        float MapZAccurate  = object->GetMapHeightAccurate(object->GetPositionX(), object->GetPositionY(), object->GetPositionZ(), /*vmap=*/true, DEFAULT_HEIGHT_SEARCH, probeR);
        const char* shapeStr = "circle";
        if (sWorld->getIntConfig(CONFIG_HEIGHT_ACCURATE_SHAPE) == 1)
            shapeStr = "square";
```

src/server/game/Maps/Map.cpp
```diff
@@ -1183,6 +1229,24 @@ float Map::GetGridHeight(float x, float y) const
     return INVALID_HEIGHT;
 }
 
+float Map::GetGridHeightAccurate(float x, float y, float radius, float yaw) const
+{
+    if (GridTerrainData* gmap = const_cast<Map*>(this)->GetGridTerrainData(x, y))
+    {
+        GridTerrainData::GroundFootprintShape shape = GridTerrainData::GroundFootprintShape::Circle;
+        if (sWorld && sWorld->getIntConfig(CONFIG_HEIGHT_ACCURATE_SHAPE) == 1)
+            shape = GridTerrainData::GroundFootprintShape::Square;
```
Tereneckla:
```suggestion
        GridTerrainData::GroundFootprintShape shape = (GroundFootprintShapes)World->getIntConfig(CONFIG_HEIGHT_ACCURATE_SHAPE);
```

src/server/scripts/Commands/cs_misc.cpp
Tereneckla: Maybe add a to_string to the enum GroundFootprintShape

src/server/game/World/WorldConfig.cpp
VG-prog: Gotcha!

Tereneckla:
sorry for the typo in the suggestion

VG-prog:
Should be ready.

VG-prog:
> At an isolated glance, this LOOKS like the proper fix for collision testing + resolve to me. If this gets properly tested to not have any annoying side effects, this would be a major improvement.

Sorry to bother you, review to latest changes would be appreciated, thanks!

walkline:
Did I understand correctly that the new collision is only used in the .gps command for now, and in other places we are still relying on the old algorithm?
If so, I guess blink and similar spells wouldn’t be affected yet, but I appreciate you checking them anyway. It’s good to have confirmation that the current implementation works without noticeable issues.

From my perspective, it might not be necessary to expose all of the collision options in the config, I would prefer to keep only the one that gives the best results. I can see how having multiple options could be useful during testing of this PR to compare approaches, but ideally it would be resolved before merging. That’s just my preference, though.

Also, maybe we should specify in comments or somewhere that the suggested implementation is using old simple raycasting in vmaps.

Another minor point, I would consider using a different name instead of “accurate.” Maybe something like GetSafeGroundHeight or a name that reflects the implementation details, such as GetMapHeightForSphere or something. But that’s just a minor nitpick.

VG-prog:
> Did I understand correctly that the new collision is only used in the .gps command for now, and in other places we are still relying on the old algorithm? If so, I guess blink and similar spells wouldn’t be affected yet, but I appreciate you checking them anyway. It’s good to have confirmation that the current implementation works without noticeable issues.
> 
> From my perspective, it might not be necessary to expose all of the collision options in the config, I would prefer to keep only the one that gives the best results. I can see how having multiple options could be useful during testing of this PR to compare approaches, but ideally it would be resolved before merging. That’s just my preference, though.
> 
> Also, maybe we should specify in comments or somewhere that the suggested implementation is using old simple raycasting in vmaps.
> 
> Another minor point, I would consider using a different name instead of “accurate.” Maybe something like GetSafeGroundHeight or a name that reflects the implementation details, such as GetMapHeightForSphere or something. But that’s just a minor nitpick.

Yeah, you understood correctly, only used in the gps command for now, i still have to do the vmaps getHeight and dynamic getHeight to use the same algorithm, i will push them in couple minutes.

So yes, nothing is yet affected apart from .gps, i have implemented it on blink locally but, since blink still relies on old implementation which i don't want to touch for obvious reasons since it can lead to undesired results due to old code, this is a thing that the ones that modified blink in previous commits use the new functions accordingly and rewritte where it's needed using the new API.

Feel free to add suggestions on the code and i will change them if relevant and necessary.

Yehonal (approved):
Great job here

Kitzunu:
Status here? Merge conflicts.

@walkline any other inputs on this one?

VG-prog:
Update FYI: 

The old code computes the terrain height at a single point. The new code keeps the exact same legacy plane under that point, but adds the mathematically correct footprint support offset for the unit radius. On flat ground it returns the same height as legacy; on slopes and edges it raises the result just enough to prevent the unit footprint from penetrating the terrain. This removes the need for fixed Z hacks and makes Blink/terrain settlement depend on actual slope, radius and orientation instead of arbitrary offsets.

VG-prog:
Latest changes: VMAP/static and dynamic height now use real hit normals, unit GetMapHeight() routes through accurate height with legacy fallback, Blink adds swept footprint collision probing, and .gps compares legacy vs accurate heights clearly, this means for creatures and players, spells and all which relies on height calcs

image

Tested roughly 100-200 blinks, maybe more on very high slopes and all type of terrains, i haven't been able to clip a single time or abuse of blink to access unwanted zones for players, tests appreciated.

walkline (commented):
Hey, thanks for the contribution. First of all, this is great work. I see some lack of feedback, maybe because of my previous comment, so here are my another 50 cents. This is just my personal view.

I see that one of my previous feedback points was addressed (now you are using the new logic in some places like blink), but not all of them.

At the moment, I see a big issue from a practical point of view. As of now, we have the new implementation, and for example for vmaps it would be 4x slower. So if we enable this performance downgrade, we should gain some other benefits, right?
We are claiming here that it would improve the accuracy, but we don’t have a linked issue or any practical example where the improvement actually happens to make the claim valid. So until this claim has some evidence, I would keep it as a disabled feature. Then if some user reports an issue with blink, we can then enable this feature and see if it resolves the issue. And if it does, then we can reconsider it and make it the default behavior.

VG-prog:
> Hey, thanks for the contribution. First of all, this is great work. I see some lack of feedback, maybe because of my previous comment, so here are my another 50 cents. This is just my personal view.
> 
> I see that one of my previous feedback points was addressed (now you are using the new logic in some places like blink), but not all of them.
> 
> At the moment, I see a big issue from a practical point of view. As of now, we have the new implementation, and for example for vmaps it would be 4x slower. So if we enable this performance downgrade, we should gain some other benefits, right? We are claiming here that it would improve the accuracy, but we don’t have a linked issue or any practical example where the improvement actually happens to make the claim valid. So until this claim has some evidence, I would keep it as a disabled feature. Then if some user reports an issue with blink, we can then enable this feature and see if it resolves the issue. And if it does, then we can reconsider it and make it the default behavior.

Hello, thanks for you feedback, the logic is implemented in the places where the height was previously getting bad data becouse, actually, in the current code, (not this new implementation), you will have to continue doing mid-hacks to solve a foundational issue and there are hacks to avoid this to happen, such as arbitrary offsets and increased search dists just becouse you're treating a unit/colider like a point in a plane rather than a cylinder which has a volume.

I agree that performance matters, but the “4x slower VMAP” concern applied to the previous finite-difference version. The current implementation does not perform 4 extra VMAP height rays; it reuses the same ray hit and extracts the real triangle normal, while keeping the old fast path for callers that do not request a normal.

There are also practical reports related to this class of bug: #6232 has Charge clipping through terrain on a hill, #6427/#11721 cover Blink ground/water/falling height issues, TC #21255 mentions Blink/Shadowstep/Charge dropping through map geometry, TC #23706 covers Blink on hills, and AC #25126 reports platform/edge movement issues in BEM including Charge dropping through ground.

I do not claim this fixes all movement/pathing bugs, but it directly addresses the height-settlement problem described in the original @Pursche discussion: point-vs-plane server height checks are not equivalent to a unit footprint/cylinder settling on the surface. This PR provides that footprint-aware settlement for grid, VMAP and dynamic geometry, with legacy fallback and config control.

walkline:
> Hello, thanks for you feedback, the logic is implemented in the places where the height was previously getting bad data becouse, actually, in the current code, (not this new implementation), you will have to continue doing mid-hacks to solve a foundational issue and there are hacks to avoid this to happen, such as arbitrary offsets and increased search dists just becouse you're treating a unit/colider like a point in a plane rather than a cylinder which has a volume.

Are you aware of cases where we add an offset and pass it to the client? Last time I checked the VMAP implementation, I think we passed a raw point from triangle.

> I agree that performance matters, but the “4x slower VMAP” concern applied to the previous finite-difference version. The current implementation does not perform 4 extra VMAP height rays; it reuses the same ray hit and extracts the real triangle normal, while keeping the old fast path for callers that do not request a normal.

Ah, I think your previous implementation used an approach with four raycasts. But if you’re working with a single triangle, wouldn’t that be inaccurate on complex terrain? For example, this slope is made up of many triangles.
Screenshot 2026-04-28 at 19 01 42

> AC https://github.com/azerothcore/azerothcore-wotlk/issues/25126 reports platform/edge movement issues in BEM including Charge dropping through ground.

That’s a good and fresh report, thanks. Maybe someone could try to reproduce this charge issue https://youtu.be/Uj_vg37KYhU?t=130 on a clean core and then with these changes, this often happens due to incorrect height calculations.

VG-prog:
> > Hello, thanks for you feedback, the logic is implemented in the places where the height was previously getting bad data becouse, actually, in the current code, (not this new implementation), you will have to continue doing mid-hacks to solve a foundational issue and there are hacks to avoid this to happen, such as arbitrary offsets and increased search dists just becouse you're treating a unit/colider like a point in a plane rather than a cylinder which has a volume.
> 
> Are you aware of cases where we add an offset and pass it to the client? Last time I checked the VMAP implementation, I think we passed a raw point from triangle.
> 
> > I agree that performance matters, but the “4x slower VMAP” concern applied to the previous finite-difference version. The current implementation does not perform 4 extra VMAP height rays; it reuses the same ray hit and extracts the real triangle normal, while keeping the old fast path for callers that do not request a normal.
> 
> Ah, I think your previous implementation used an approach with four raycasts. But if you’re working with a single triangle, wouldn’t that be inaccurate on complex terrain? For example, this slope is made up of many triangles. Screenshot 2026-04-28 at 19 01 42
> 
> > AC #25126 reports platform/edge movement issues in BEM including Charge dropping through ground.
> 
> That’s a good and fresh report, thanks. Maybe someone could try to reproduce this charge issue https://youtu.be/Uj_vg37KYhU?t=130 on a clean core and then with these changes, this often happens due to incorrect height calculations.

Yes. The normal is local to the exact triangle hit by the same height ray, not a global approximation of the whole slope.

So complex terrain still resolves through the actual collision triangle selected at that point; the normal is only used to settle the unit footprint on that local plane. No 4 extra VMAP rays are used anymore.

I agree the BEM Charge issue is a good validation case. I’ll try to reproduce it on clean core vs this branch and provide before/after results.

VG-prog:
@walkline On latest clean Acore, only with my branch updates, i am not able to charge on circle of blood arena as on the video, can't find a valid path, but this relies on how is the mmap data extracted, which is non related to this PR:

image

Didn't tested without this branch changes.

walkline:
I think you’re using the wrong map. I assume you’re testing it on the outland map, but there’s a separate map for arena fights, and you should probably queue for an arena match to reproduce it.

VG-prog:
> I think you’re using the wrong map. I assume you’re testing it on the outland map, but there’s a separate map for arena fights, and you should probably queue for an arena match to reproduce it.

Yes, i was:

image

Actually, seems like a problem about mmaps and not height related, clip can mostly happen, doesn't seem like related at all.

VG-prog:
https://github.com/azerothcore/azerothcore-wotlk/pull/23012/changes/671c9a97e6a25c1960f8f76a6bf47f4395bbb342

On latest results, a great improvement has been made, from 10 results on previous scenario, i clipped through terrain 8 times, with latest changes, i only clipped 1 about 10 times, anyways, gonna do way more tests.

Edit: Outdated.

Pathfinding seems like a sepparate issue.

Nefertumm:
Re-requesting the review from @Yehonal, as there have been a few commits since the last one. 
Also, what's the status of this PR? Is it ready to be tested?

VG-prog:
Sorry, but unless you want this merged, i am going to close it, and re-do the scope again, needs a rewritte.

sudlud:
Thanks for all the work here, and for being upfront that this needs a re-scope. Agreed on closing it for now.

The line-level feedback has basically all been addressed, so the remaining blockers are the bigger ones from the discussion: the config surface is still 9 options, the "accurate" naming is unresolved, and we never got a concrete before/after on a reported bug to justify the cost.

For the follow-ups, splitting this into smaller PRs would make it much easier to land:

1. **VMAP/dynamic hit-normal plumbing** (the `intersectRay`/`getHeightAndNormal` changes). This is low-risk, self-contained, and useful on its own. Good first PR.
2. **Footprint settlement API** (`GetHeightAccurate` on grid + map), kept behind a single off-by-default toggle, with the config trimmed to just the one shape/mode you settle on. Drop the QA-only knobs once you've picked defaults, and consider a clearer name like `GetMapHeightForSphere`.
3. **Blink/movement integration**, last, with a before/after on a specific reported issue (the BEM Charge case is a good candidate) so the accuracy gain is demonstrable.

Also worth folding the duplicated `ComputeFootprintHeightFromNormal` helper into one shared place when it comes back.

Looking forward to the rewrite.

Issue #21729, opened 2025-03-18
[CORE]Critical behavior of the flight function


wefgame:
### Current Behaviour

[https://www.wowhead.com/wotlk/ru/spell=40120/облик-стремительной-птицы](https://www.wowhead.com/wotlk/ru/spell=40120/%D0%BE%D0%B1%D0%BB%D0%B8%D0%BA-%D1%81%D1%82%D1%80%D0%B5%D0%BC%D0%B8%D1%82%D0%B5%D0%BB%D1%8C%D0%BD%D0%BE%D0%B9-%D0%BF%D1%82%D0%B8%D1%86%D1%8B) - Druid form
All flying mounts

The problem is that when exiting the druid form or flying into a room on a mount, when the aura falls, the flight form remains until the relog. The flight functions are the same as with the GM command (you can arrange an invasion of owls :D )

https://youtu.be/5pVLPUxWAUA

### Expected Behaviour

the fly should not remain on the player.

### Steps to reproduce the problem

Approach the input for example: .go xyz -1511.8613 4765.3096 -8.653583 530
Take a position near the water
When leaving the water or entering, play the removal of the aura by flags
Fly

### AC rev. hash/commit

[15:31:27] AzerothCore rev. 3b90710e6487 2025-03-17 21:28:15 +0000 (master branch) (Win64, RelWithDebInfo, Static)
[15:31:27] Connected players: 1. Characters in world: 1.
[15:31:27] Connection peak: 1.
[15:31:27] Server uptime: 13 minute(s) 26 second(s)
[15:31:27] Update time diff: 16ms. Last 500 diffs summary:
[15:31:27] |- Mean: 9ms
[15:31:27] |- Median: 11ms
[15:31:27] |- Percentiles (95, 99, max): 16ms, 16ms, 16ms
[15:31:27] Using SSL version: OpenSSL 3.4.0 22 Oct 2024 (library: OpenSSL 3.4.0 22 Oct 2024)
[15:31:27] Using Boost version: 1.86.0
[15:31:27] Using CMake version: 3.31.1
[15:31:27] Using MySQL version: 80040
[15:31:27] Found MySQL Executable: C:/Program Files/MySQL/MySQL Server 8.0/bin/mysql.exe
[15:31:27] Compiled on: Windows 10.0.22631
[15:31:27] Worldserver listening connections on port 8085
[15:31:27] Realmlist (Realm Id: 1) configured in port 8085
[...]

PR #18902, opened 2024-05-15
fix(Unit/ProcessTerrainStatusUpdate): prevent removal swimming auras if swimming

sogladev:
## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

https://github.com/azerothcore/azerothcore-wotlk/blob/ef6db03e1351f4f0c0ddc81f98086c9d99ff04ed/src/server/game/Entities/Unit/Unit.cpp#L4223

Current logic to decide if swimming is based on `GetLiquidData().Status `. However, in the coilfang tunnel some parts are  `LIQUID_MAP_NO_WATER` which allows the player to breathe underwater. 

Should swimming/noswimming logic be based on `MOVEMENTFLAG_SWIMMING` flags? Similar to how `IsFlying()` is handled

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/14562

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

`.tele coilfang`
`.learn 1066` aquatic form

breath timer indicates which areas are breathable, no liquid

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

[...]

Issue #14562, opened 2023-01-08
Tunnel to Coilfang Reservoir removes Aquatic Form


Pyromanihack:
Original Issue: https://github.com/chromiecraft/chromiecraft/issues/4612

### What client do you play on?

enUS

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

When trying to swim through the tunnel to Coilfang Reservoir aquatic form of druids gets removed by the tiniest movement

### Expected Blizzlike Behaviour

Able to swim the whole way in aquatic form

### Source

It's water

### Steps to reproduce the problem

.learn 1066 (or .aura 1066)
.tele Zangarmarsh
go to the tunnel
try to swim

### AC rev. hash/commit

8d2dda24ed85 2023-01-06

### Operating system

Ubuntu 20.04 - Windows 10 x64

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
- [mod-queue-list-cache](https://github.com/azerothcore/mod-queue-list-cache)
- [mod-rdf-expansion](https://github.com/azerothcore/mod-rdf-expansion)
[...]

PR #23015, opened 2025-09-24
Movement: Implement order counters

killerwife:
Movement protocol upgrades that aim to fix:

https://github.com/azerothcore/azerothcore-wotlk/issues/21729

Affected systems after merge:

Anticheat

How to test(after first commit, will update):

Knockback, root, teleport, flight, waterwalk, gravity disable, any speed change, hover, feather fall, mount/dismount.

PS: Not done but opening PR to abuse CI :)

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -4330,10 +4330,10 @@ void Unit::ProcessTerrainStatusUpdate()
 
     LiquidData const& liquidData = GetLiquidData();
 
-    // remove appropriate auras if we are swimming/not swimming respectively
-    if (liquidData.Status & MAP_LIQUID_STATUS_SWIMMING)
+    // remove appropriate auras if we are swimming/not swimming respectively - exact mirror of client logic
+    if (liquidData.Status & MAP_LIQUID_STATUS_SWIMMING && (liquidData.Level - GetPositionZ()) > GetCollisionHeight() * 0.75f) // Shallow water at ~75% of collision height)
         RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_NOT_ABOVEWATER);
-    else if (!isSwimming())
+    else
         RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_NOT_UNDERWATER);
 
     // liquid aura handling


Apathyxia:
With this PR, I'm not able to perform any sort of flying without a flight form / mount. 
Tested at a body of water in Terokkar Forest and Shattrath bank (Aldor).
 
However, this PR now makes jumping into flight form from water a bit weird. It still works, but not smoothly. 
See: https://youtu.be/U3f4JrZOWoc

Basically, if trying to go into a flight form right away after jumping, it is automatically cancelled. It requires to wait until the top of the jump for it to be successful.

killerwife:
Ok thats actually the result of code i did not touch at all, and however weird may sound, everything is working according to core code.

Set can fly ack carries a pos with itself that was previously ignored. Client sends a pos still in water, core code immediately checks liquid level, finds player is still in water, and kicks him out of form.

That being said, if flight form is still possible at the peak (which i reproduced as well it is), I do not think any adjustments to 

GetMap()->GetFullTerrainStatusForPosition

are in order until more info is provided upon the functioning of movement code in client reversal. (i dont know the code with which client calculates height to enable UI buttons)

Apathyxia:
That makes sense. It is indeed a bit "weird" that you were able to shapeshift into flight form while still partly submerged, so it makes sense why it requires you to really be above water.

killerwife:
Ok so, in this PR I opted not to do an insanely big overhaul of all movement packets.

I tackled movement packet counters and the set of 4 packets: Hover, Can Fly, Water walk and Slow fall

These basically work the same, but had to integrate it with existing movement code since they do send movement info, which does show movement of client and has to be fully rechecked like anything else.

This has the side effect of better simulation of movement from client, but might brick some other hacks that were in place. (which was the main reason against full rework)

Please do note that this behaviours is from sniffs and client reversal, so if anyone has a bone to pick, outside of crashes ofc, there is a bit of a need for elevated proof versus wotlk specifically, cos the protocol changed almost every expansion.

killerwife:
PR ready with only downside fixed.

killerwife:
image

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -4330,10 +4330,10 @@ void Unit::ProcessTerrainStatusUpdate()
 
     LiquidData const& liquidData = GetLiquidData();
 
-    // remove appropriate auras if we are swimming/not swimming respectively
-    if (liquidData.Status & MAP_LIQUID_STATUS_SWIMMING)
+    // remove appropriate auras if we are swimming/not swimming respectively - exact mirror of client logic
+    if (liquidData.Status & MAP_LIQUID_STATUS_SWIMMING && (liquidData.Level - GetPositionZ()) > GetCollisionHeight() * 0.75f) // Shallow water at ~75% of collision height)
         RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_NOT_ABOVEWATER);
-    else if (!isSwimming())
```
sogladev, line 4338: This clause `!isSwimming()` exists to fix an issue with the server removing Aquatic Form when swimming through the Coilfang Tunnel. Is removing this required?

details:

- https://github.com/azerothcore/azerothcore-wotlk/issues/14562

- https://github.com/azerothcore/azerothcore-wotlk/pull/18902

to test
```
.cast 1066
.go xyz 621.83 6884.61 -64.54 530 2.4
```

killerwife: Yes.

Client only knows one type of logic, either swimming or not swimming. Thats what the 0.75 adjustment is there for.

image


image

Here are the receipts from client reversal. If there were an issue in client data, client wouldnt allow in said coilfang tunnel to engage swimming form. As such, if there is an issue, its not this piece of code.

Issue #23116, opened 2025-10-05
Client freezes after recent update (possible regression from latest commits)


r0m1ntik:
### Current Behaviour

After applying the latest update, many players (and myself) experienced a complete client freeze — the game becomes unresponsive and can only be closed via Task Manager.
I am not entirely sure how to reproduce it consistently, but it happened shortly after deploying the new core update.
Since these are the only commits recently applied that could affect it, I believe one of them introduced the issue.
The server had around 1000 players online during the incident.

I reverted both commits and the issue disappeared.

Commits possibly related:

https://github.com/azerothcore/azerothcore-wotlk/commit/86af3f302b2b1e464661ef4e80eb049f58738ccc
 (possibly unrelated, reverted for safety)

https://github.com/azerothcore/azerothcore-wotlk/commit/b80353d2887a41299d27f51a13108f18eb00e118
 (most likely related to the problem)

### Expected Behaviour

Client should not freeze; game should remain stable and responsive after the update.

### Source

No direct source; issue observed after applying recent commits.

### Steps to reproduce the problem

1. Update core including the commits above.
2. Run the server under heavy player load (~1000 online).
3. Wait for some time during gameplay — client eventually freezes.
4. Revert commits → issue disappears.

### Extra Notes

Happens randomly, not tied to a specific zone or spell. Could be threading or packet-handling related.

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/2afcd037c7a8fcef6d330c59b8bc4bc8a17a86d2
<img width="1027" height="692" alt="Image" src="https://github.com/user-attachments/assets/84c1f11f-6ac6-4813-b73c-206843a5e5fd" />

<img width="938" height="312" alt="Image" src="https://github.com/user-attachments/assets/0683c753-c0b3-4aff-ac01-52e86acdc764" /> 

### Operating system

Debian 12

### Custom changes or Modules

Cfbg
Transmog
Skeleton

Nyeriah:
cc @killerwife

killerwife:
Cheers, gonna check smth and offer quick triage.

killerwife:
Linked a PR which is meant to start triage of this.

digz6666:
I have same issue.

killerwife:
https://github.com/azerothcore/azerothcore-wotlk/pull/23147

Second pr as part of this issue

Also start posting client crash (has to be unmodded)

killerwife:
[...]

PR #23117, opened 2025-10-05
Movement: Add safeguards against root freezes

killerwife:
Linked issue:

https://github.com/azerothcore/azerothcore-wotlk/issues/23116

This PR aims to do one thing:

Check if aforementioned issues are with root, and if yes, there are two steps required:

1. examples of trace log have to be examined and if possible, forwarded to me through an issue (a script can do whatever, but it should never hit the added trace log)
2. if confirmed, will rework the root protocol into the same ack protocol as flight (flight revealed it, because it is now async, not sync)

### DIFF
--- a/src/server/game/Movement/Spline/MoveSplineInit.cpp
+++ b/src/server/game/Movement/Spline/MoveSplineInit.cpp
@@ -103,7 +103,13 @@ namespace Movement
 
         bool isOrientationOnly = args.path.size() == 2 && args.path[0] == args.path[1];
 
-        if ((moveFlags & MOVEMENTFLAG_ROOT) || isOrientationOnly)
+        if (moveFlags & MOVEMENTFLAG_ROOT) // This case should essentially never occur - hence the trace logging - hints to issues elsewhere
+        {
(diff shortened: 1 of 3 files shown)


src/server/game/Movement/Spline/MoveSplineInit.cpp
```diff
@@ -103,7 +104,13 @@ namespace Movement
 
         bool isOrientationOnly = args.path.size() == 2 && args.path[0] == args.path[1];
 
-        if ((moveFlags & MOVEMENTFLAG_ROOT) || isOrientationOnly)
+        if (moveFlags & MOVEMENTFLAG_ROOT) // This case should essentially never occur - hence the trace logging - hints to issues elsewhere
+        {
+            LOG_TRACE("movement", "Invalid movement during root. Entry: %u IsImmobilized %s, moveflags %u", unit->GetEntry(), unit->IsImmobilizedState() ? "true" : "false", moveFlags);
```
Nyeriah:
```suggestion
            LOG_TRACE("movement", "Invalid movement during root. Entry: {} IsImmobilized {}, moveflags {}", unit->GetEntry(), unit->IsImmobilizedState() ? "true" : "false", moveFlags);
```

killerwife: Yikes. Ty

Nyeriah (approved):

PR #23147, opened 2025-10-07
Movement: Rework root protocol with ack in mind

killerwife:
Related and followup: https://github.com/azerothcore/azerothcore-wotlk/pull/23117

Reworks root with ack protocol properly like flight previously

Should resolve client crashes that started happening with flight rework


Teknishun7:
I cannot test this PR specifically, but as of now:
image
I fiddled with many forms of flying, revisited spots that would crash and I cannot get the PTR to crash anymore.

PR #23160, opened 2025-10-09
fix(Core/Entities): Daily and other quests should grant XP when rewarded on subsequent completions.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes unreported issue.

## SOURCE:

## Tests Performed:

This PR:
- [X] Has not been tested.

## How to Test the Changes:

1. Complete a daily quest.
2. Time travel.
3. Complete the same daily quest.
4. Ensure that you receive experience for the second completion.

### DIFF
--- a/src/server/game/Entities/Player/PlayerQuest.cpp
+++ b/src/server/game/Entities/Player/PlayerQuest.cpp
@@ -740,7 +740,7 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver, 
     if (log_slot < MAX_QUEST_LOG_SIZE)
         SetQuestSlot(log_slot, 0);
 
-    const bool rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest() && !(quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly());
+    bool const rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest() && !(quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly());
 
     // Repeatable quests (not time-based reset ones) should not give XP on subsequent completions
     uint32 XP = rewarded ? 0 : CalculateQuestRewardXP(quest);


src/server/game/Entities/Player/PlayerQuest.cpp
```diff
@@ -742,6 +742,9 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
 
     bool rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest();
 
+    if (quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly())
+        rewarded = false;
```
sudlud: What about adding these conditions to the bool above, also make it const. Also please update the comment below. Aren't there repeatable quests also? That just have a repeatable flag. What about those

heyitsbench: Repeatable quests (that are not reset based on time) should not give XP on subsequent completions afaik. Will adjust the other requests when next I'm at my computer.

src/server/game/Entities/Player/PlayerQuest.cpp
```diff
@@ -740,9 +740,9 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
     if (log_slot < MAX_QUEST_LOG_SIZE)
         SetQuestSlot(log_slot, 0);
 
-    bool rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest();
+    const bool rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest() && !(quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly());
```
sudlud:
```suggestion
    bool const rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest() && !(quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly());
```
:P

src/server/game/Entities/Player/PlayerQuest.cpp
```diff
@@ -740,9 +740,9 @@ void Player::RewardQuest(Quest const* quest, uint32 reward, Object* questGiver,
     if (log_slot < MAX_QUEST_LOG_SIZE)
         SetQuestSlot(log_slot, 0);
 
-    bool rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest();
+    bool const rewarded = IsQuestRewarded(quest_id) && !quest->IsDFQuest() && !(quest->IsDaily() || quest->IsWeekly() || quest->IsMonthly());
```
sudlud, line 743: I would probably continue the style by doing && !() && !() etc all the way through for better readability

heyitsbench: Tbh eventually (IMO) the IsDaily/Weekly/Monthly should probably be made into its own helper function in checks also specific to these. I didn't see an existing helper function, nor did I spot any checks that already make use of these three checks that would be replaced, but sectioning them off like this would make it easier to swap in the future if/when.

sudlud: True, a helper function would probably make sense if needed more often

Nyeriah:
Hey were we sure about this one? From memory they only gave xp the first time you completed them in wrath, this does sound like a classic+ change

heyitsbench:
> Hey were we sure about this one?

I see no evidence anywhere saying that the previous behavior is correct. Various dailies on Wowhead have comments specifically mentioning the XP gain, nothing on any wiki that I see saying they only give that XP once.

heyitsbench:
https://www.wolf-hund.org/cataclysm-peparation-25-dailies-to-boost-xp/
This article also seems to quote a Blizzard dev chat mentioning that stacking 25 daily quests in the quest log to turn in at the start of an expansion as a way of getting a head start on experience, plus multiple other forum posts mentioning this (mostly in the transition from Wrath -> Cata that I see), I doubt they'd mention it if people could have only gotten that XP the first time they turned in.

wetbrownsauce:
I've also seen discussions from 15-16 years ago about completing dailies but not turning them in to prepare for cata launch in order to get some quick exp. 

Someone in this thread mentions they did it for wotlk launch as well:
https://www.reddit.com/r/wow/comments/d5jl9/cataclysm_leveling_scheme/

So I think bench is right on this one.

sudlud:
bench is never wrong

heyitsbench:
> bench is never wrong

I'm wrong pretty often, but occasionally I get lucky. 😛

PR #23162, opened 2025-10-09
fix(Scripts/Karazhan): Fix Tenris Mirkblood area trigger conditions.

GrenderG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Tenris Mirkblood script has 2 `if (instance->GetBossState(DATA_MIRKBLOOD) != DONE)` checks but the Boss State is never set to `DONE` ~~due to missing `_JustDied()` call~~. Because of this, the corpse of Tenris is still speaking whenever a player approaches the area trigger.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [X] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1. Engage with Tenris Mirkblood and defeat him.
2. See how `at_karazhan_mirkblood_approach` and `at_karazhan_mirkblood_entrance` no longer trigger when the corpse is still present.

- [ ]
- [ ]


heyitsbench:
Tested, [did not work for me](https://github.com/user-attachments/assets/5d806a1c-259e-46cf-9584-a86eb8732559). The issue seems to be that [instance](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/AI/ScriptedAI/ScriptedCreature.cpp#L642) does not exist in the _JustDied function, but the _JustDied function *is* getting hit with Mirkblood's death, even on master AC.

heyitsbench:
Also, if I understand correctly, if there is no override function provided for JustDied in a given BossAI, it will still run the _JustDied function, since it's not being overridden.

Tereneckla:
Yes, the default JustDied only calls _JustDied

GrenderG:
As a workaround, since this is a special boss tied to a game event, we could simply check if he's alive instead of relying on the Boss State. I implemented that now.

src/server/scripts/EasternKingdoms/Karazhan/boss_tenris_mirkblood.cpp
```diff
@@ -281,8 +281,8 @@ class at_karazhan_mirkblood_approach : public AreaTriggerScript
     bool OnTrigger(Player* player, AreaTrigger const* /*trigger*/) override
     {
         if (InstanceScript* instance = player->GetInstanceScript())
-            if (instance->GetBossState(DATA_MIRKBLOOD) != DONE)
-                if (Creature* mirkblood = instance->GetCreature(DATA_MIRKBLOOD))
+            if (Creature* mirkblood = instance->GetCreature(DATA_MIRKBLOOD))
```
heyitsbench, line 293: May want to check if in combat as well, else he may continue talking during the encounter.

GrenderG: Good idea.

GrenderG: I added `!mirkblood->IsInCombat()` check to the `SAY_APPROACH` area trigger but decided to use `mirkblood->IsImmuneToPC()` instead for the second one. I did this to avoid a rare (but probably possible) case of pulling the boss without entering the area trigger, leaving him in an immune state.

PR #23254, opened 2025-10-15
feat(Core/Network): Per-user togglable packet logging

killerwife:
Enables toggling packet logging per user rather than for whole server (even with 10 users on ptr that can make hundreds of megabytes)

For diagnosing client crashes. If player has this on while his client freezes/crashes, I can figure out why.


TheSCREWEDSoftware:
// `data\sql\updates\pending_db_world`
```sql
DELETE FROM `command` WHERE `name` = "packetlog";
INSERT INTO `command` (`name`, `security`, `help`) VALUES 
("packetlog", 2, "Syntax: .packetlog [on/off]\n Toggles to allow the character using the command to start to log their packets into the server, without needing PacketLogFile enabled in the worldserver config");
```

TheSCREWEDSoftware:
Be free to used a better sytanx, i'm horrible in writing for the command.

TheSCREWEDSoftware:
Will merge when the tests finish if no errors. (after the update branch)

TheSCREWEDSoftware:
Because i updated the branch I had the co-author, sighs

PR #23267, opened 2025-10-16
fix(Scripts/ICC): add null checks to Icecrown gunship scripts to prevent crashes

FrancescoBorzi:
Generated with Codex

## Summary
- guard Icecrown gunship AI logic against missing InstanceScript pointers before accessing encounter data
- ensure transport slot helpers bail out safely when no instance data is available
- avoid casting Burning Pitch when no instance context exists

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp
@@ -545,13 +545,22 @@ public:
 
     struct npc_gunshipAI : public NullCreatureAI
     {
-        npc_gunshipAI(Creature* creature) : NullCreatureAI(creature), _instance(creature->GetInstanceScript()), _teamIdInInstance(TeamId(creature->GetInstanceScript()->GetData(DATA_TEAMID_IN_INSTANCE))), _died(false), _summonedFirstMage(false)
+        npc_gunshipAI(Creature* creature) : NullCreatureAI(creature), _instance(creature->GetInstanceScript()), _teamIdInInstance(TEAM_NEUTRAL), _died(false), _summonedFirstMage(false)
         {
+            if (_instance)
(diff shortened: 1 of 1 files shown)


heyitsbench:
Why would the instance script be null?

Takenbacon:
> Why would the instance script be null?

Only if manually spawned outside of the instance for some reason. Highly unlikely but technically not impossible.

heyitsbench:
> Only if manually spawned outside of the instance for some reason. Highly unlikely but technically not impossible.

Iirc, scripts set up in instance script files don't actually attach to the entity unless they are in that instance. For example, spawning Orgrim's Hammer (NPC 37215, script name `npc_gunship`) in the `ScottTest` tele location doesn't seem to have that NPC hit any of their script functions, JustDied or DamageTaken for example.

Nyeriah:
The latest registry methods ensure scripts are not executed outside the creature respective instance but ancient stuff like this don’t make use of it

Nyeriah:
Actually, looking into the file, it seems that they do, but not the latest, which was implemented here: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/Northrend/IcecrownCitadel/icecrown_citadel.h#L605

So this: https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp#L766

prevents the script from running outside the instance, but the AI doesn't know about it 😄

FrancescoBorzi:
this is just me playing with AI :)

heyitsbench:
> this is just me playing with AI :)

That's fine, but sticking null checks in places where they're just not needed ain't something I'd like to see merged. Just want to make sure if they're actually warranted or not is all.

FrancescoBorzi:
> That's fine, but sticking null checks in places where they're just not needed ain't something I'd like to see merged. 

yes of course, closing this as not needed

PR #23268, opened 2025-10-16
fix(Scripts/ICC): guard gunship reset against missing transports

FrancescoBorzi:
## Summary
- restore the gunship script to its original logic while tightening ResetEncounterEvent
- check for missing transports before unloading passengers during encounter reset to avoid crashes

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp
@@ -520,12 +520,20 @@ public:
     bool Execute(uint64, uint32) override
     {
         _caster->CastSpell(_caster, _spellId, true);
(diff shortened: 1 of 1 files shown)


src/server/scripts/Northrend/IcecrownCitadel/boss_icecrown_gunship_battle.cpp
```diff
@@ -520,12 +520,20 @@ class ResetEncounterEvent : public BasicEvent
     bool Execute(uint64, uint32) override
     {
         _caster->CastSpell(_caster, _spellId, true);
-        _caster->GetTransport()->ToMotionTransport()->UnloadNonStaticPassengers();
-        _caster->GetTransport()->AddObjectToRemoveList();
+
+        if (TransportBase* casterTransport = _caster->GetTransport())
```
PkllonG: TransportBase To Transport

Issue #18474, opened 2024-03-01
Isfar incorrect dialogue options


BeefSteaker:
https://github.com/chromiecraft/chromiecraft/issues/5146
### What client do you play on?

enUS

### Faction

Alliance

### Content Phase:

Generic

### Current Behaviour

The NPC Isfar outside the Sethekk Walls has 4 questions you can ask him. clicking the 3rd or fourth does nothing

![image](https://user-images.githubusercontent.com/90989261/224418539-b85e6945-a0fa-4c9e-98fa-6074bbe77143.png)

### Expected Blizzlike Behaviour

3 dialogue options
![image](https://github.com/azerothcore/azerothcore-wotlk/assets/11332559/7fe1dfaf-971e-47ba-b1f4-68376caf9463)

### Source

https://wowpedia.fandom.com/wiki/Isfar

### Steps to reproduce the problem

.go c id 18933
click 3rd and 4th dialogue option

### Extra Notes

https://wowgaming.altervista.org/aowow/?npc=18933

He is outside the Sethekk Halls which is down in that big thunderdome area.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/0200a288cd89034a8b491253006a0b0055178deb

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
[...]

PR #18688, opened 2024-04-08
fix(DB/Creature): Isfar gossip options

bauxa:
## Changes Proposed:

This PR proposes changes to:
-  [x] Database (SAI, creatures, etc).

Removed dialogue option that shouldn't be here.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18474
- Closes https://github.com/chromiecraft/chromiecraft/issues/5146

## SOURCE:

The changes have been validated through:
- [x] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided in the linked issue

1. Go to Isfar
2. Click on the dialogue option you want

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #23321, opened 2025-10-20
fix(DB/Gossip): Fix Isfar incorrect dialogue options

pangolp:
## Changes Proposed

- Fixes gossip option "Where can I find the relics of Terokk?" for NPC Isfar (18933).
- Added conditions to show option only when player has quest 10098 and doesn't have item 32888.
- Implemented SmartAI to deliver item when option is selected.

## Issues Addressed

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18474
- Closes https://github.com/chromiecraft/chromiecraft/issues/5146

## Tests Performed

- [x] Removed all existing gossip options for menu 7866
- [x] Added the 3 correct dialogue options
- [x] Verified options match Wowpedia source

## How to Test the Changes

1. `.go c 67958` or `.go c id 18933`
2. Accept quest 10098, talk to Isfar, select the gossip option, verify item 32888 is received.

## Known Issues and TODO List

None

## Source

- https://wowpedia.fandom.com/wiki/Isfar
- https://www.wowhead.com/tbc/npc=18933/isfar

### DIFF
--- a/data/sql/updates/db_world/2025_10_28_05.sql
+++ b/data/sql/updates/db_world/2025_10_28_05.sql
@@ -1,0 +1,13 @@
+-- DB update 2025_10_28_04 -> 2025_10_28_05
+--
+UPDATE `creature_template` SET `AIName` = 'SmartAI' WHERE `entry` = 27971;
+DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 27971);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+(27971, 0, 0, 0, 0, 0, 100, 2, 12000, 18000, 12000, 18000, 0, 0, 11, 51842, 0, 0, 0, 0, 0, 6, 0, 0, 0, 0, 0, 0, 0, 0, 'Unrelenting Construct - In Combat - Cast \'Charge\' (No Repeat) (Dungeon/Normal)'),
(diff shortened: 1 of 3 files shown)


Nyeriah:
I suspect the relics one is related to this https://www.wowhead.com/tbc/quest=10098/terokks-legacy

FrancescoBorzi (approved):

pangolp:
> I suspect the relics one is related to this https://www.wowhead.com/tbc/quest=10098/terokks-legacy

I will try to find out, but if it is that way, then there are no conditions for the gossip to show everything while it has that mission active.

pangolp:
I haven't found any information on this. There was another PR open before, where the same change was attempted, but then the contributor closed it and I completely forgot about it. https://github.com/azerothcore/azerothcore-wotlk/pull/18688. I'm not sure how to proceed. I think I'll merge it, and if that's the case, wait for someone to report it. Then, I'll add the missing option and add a condition to prevent it from showing all the time and only have it appear when the player has the quest.

Nyeriah:
I would like this to be verified before this can be merged

> I suspect the relics one is related to this https://www.wowhead.com/tbc/quest=10098/terokks-legacy

pangolp:
> I would like this to be verified before this can be merged
> 
> > I suspect the relics one is related to this https://www.wowhead.com/tbc/quest=10098/terokks-legacy

I don't have any more information about it, so I guess it will remain open until someone can say something about it. It seems simpler to me to close the issues, which have been open for a long time, because I doubt that this gossip will really affect anyone. If it is verified that what you mention, the report is generated, and the corresponding fix is ​​quickly made. But it is not a decision that I should make. So this is where my work ends.

Nyeriah:
It is counterproductive to merge quick fixes for the sake of closing issues if they can't be verified/researched, we are, at least potentially, closing one bug report just to open another and, in worst case, data might be lost.

It's in no way critical though so I am sure we can wait a little longer to find evidence or conclude there is no evidence to be found.

pangolp:
> It is counterproductive to merge quick fixes for the sake of closing issues if they can't be verified/researched, we are, at least potentially, closing one bug report just to open another and, in worst case, data might be lost.
> 
> It's in no way critical though so I am sure we can wait a little longer to find evidence or conclude there is no evidence to be found.

Ok, I don't have more information unfortunately. We'll have to see if anyone has a sniff or additional information then.

pangolp:
image

image

According to the video I just watched, the gossip is apparently related to the quest, so the fix shouldn't actually be to remove it, but rather to hide it and show it only if the player has the quest active.

https://www.youtube.com/watch?v=MHMCPOgRAew

pangolp:
I managed to investigate the matter a little further, and upon taking the quest, the NPC gives us an item with information about it. What the gossip is supposed to do is give us the item, in case the player has taken the quest and for some reason has deleted it.

WoWScrnShot_102625_015849

WoWScrnShot_102625_015858

pangolp:
@Nyeriah @GrenderG

PR #23373, opened 2025-10-23
fix(Scripts/Karazhan): Barnes Missing Wizard of Oz Speech

brighton-chi:
Barnes was not saying his first line for the Wizard of Oz event ("Welcome ladies and gentlemen, to this evening's performance").
This is because the textid for the line is 0 so the if text checks were failing. I removed the if text checks--I don't think there is any adverse effect, but please correct me if this is not the way to do it. I made the same changes for the other Opera events too because it didn't make sense to me to have discrepancies between them.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. .go creature name Moroes
2. .die on Moroes
3. .go creature name Barnes
4. make sure GM mode is on; talk to Barnes and select the WoZ event
5. follow Barnes and watch his speech

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/EasternKingdoms/Karazhan/karazhan.cpp
+++ b/src/server/scripts/EasternKingdoms/Karazhan/karazhan.cpp
@@ -219,24 +219,21 @@ public:
             {
                 case EVENT_OZ:
                     text = OzDialogue[count].textid;
-                    if (OzDialogue[count].timer)
-                        TalkTimer = OzDialogue[count].timer;
+                    TalkTimer = OzDialogue[count].timer;
                     break;
-
                 case EVENT_HOOD:
                     text = HoodDialogue[count].textid;
-                    if (HoodDialogue[count].timer)
-                        TalkTimer = HoodDialogue[count].timer;
+                    TalkTimer = HoodDialogue[count].timer;
                     break;
-
                 case EVENT_RAJ:
                     text = RAJDialogue[count].textid;
-                    if (RAJDialogue[count].timer)
-                        TalkTimer = RAJDialogue[count].timer;
+                    TalkTimer = RAJDialogue[count].timer;
                     break;
+                default:
+                    return;
             }
 
-                CreatureAI::Talk(text);
+            CreatureAI::Talk(text);
         }
 
         void PrepareEncounter()


kissingers:
If miss some sql data, will this case crash?

brighton-chi:
I don't think so. CreatureAI::Talk can handle nullptr as an input.

Alternatively, it should be possible to change the textids so they don't start with zero, but that requires changing the sql in addition to the source.

src/server/scripts/EasternKingdoms/Karazhan/karazhan.cpp
```diff
@@ -218,28 +218,24 @@ class npc_barnes : public CreatureScript
...
-                    if (RAJDialogue[count].textid)
-                        text = RAJDialogue[count].textid;
+                    text = RAJDialogue[count].textid;
                     if (RAJDialogue[count].timer)
                         TalkTimer = RAJDialogue[count].timer;
                     break;
             }
 
-            if (text)
                 CreatureAI::Talk(text);
```
sogladev: I checked the script, and this is always called with valid parameters `count [0, 3]` and valid the struct data
```cpp
static Dialogue OzDialogue[] =
static Dialogue HoodDialogue[] =
static Dialogue RAJDialogue[] =
```
so we can get rid of the `ifs` here. Funnily the only thing needed to make this work was removing the `if (text)` check, but it can be cleaned up a little more

```suggestion
            switch (m_uiEventId)
            {
                case EVENT_OZ:
                    text = OzDialogue[count].textid;
                    TalkTimer = OzDialogue[count].timer;
                    break;
                case EVENT_HOOD:
                    text = HoodDialogue[count].textid;
                    TalkTimer = HoodDialogue[count].timer;
                    break;
                case EVENT_RAJ:
                    text = RAJDialogue[count].textid;
                    TalkTimer = RAJDialogue[count].timer;
                    break;
                default:
                    return;
            }

            CreatureAI::Talk(text);
```

rorshan:
I tested this, it works fine. Had I known that the first line of dialog was different depending on the boss, I would have made more gold betting on the outcome

brighton-chi:
Is there anything else I need to do here?

PR #23387, opened 2025-10-24
fix(CI): improve caching

FrancescoBorzi:
image

image


Yehonal:
what are you trying to do here?

FrancescoBorzi:
> what are you trying to do here?

@Yehonal I'm trying to improve the ccache configuration to make our cache great again, because lately it's not hitting very well and builds are taking ages to pass. Not sure what has changed compared to before (io so 4 parole).

Yehonal:
Ok so you think that 'ref' is the cause

IntelligentQuantum (approved):

Issue #23178, opened 2025-10-11
Druid can slowly rise to sky when teleported to the old world within swift flight form


kissingers:
### Current Behaviour

Druid can slowly rise when teleported to the old world within swift flight form, but can not see swift flight form

### Expected Behaviour

Can not fly at old world

### Source

<img width="486" height="540" alt="Image" src="https://github.com/user-attachments/assets/1d76519b-314e-4c7e-8d34-868c6c219bfb" />

### Steps to reproduce the problem

Druid 
.gm off
.go c 96666
.aura 40120
.go c 23455
You can not see aura 40120, but you can rise up.

If not gm, can use lua teleport test, the same result.

### AC rev. hash/commit

AzerothCore rev. 292f1a417d09 2025-10-10 20:07:56 -0300 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
服务器运行时间: 16 second(s)
Update time diff: 50ms. Last 178 diffs summary:
|- Mean: 50ms
|- Median: 50ms
|- Percentiles (95, 99, max): 51ms, 51ms, 67ms
Using SSL version: OpenSSL 3.5.1 1 Jul 2025 (library: OpenSSL 3.5.1 1 Jul 2025)
Using Boost version: 1.83.0
Using CMake version: 3.31.6
Using MySQL version: 80406
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.14.11-2-pve
Worldserver listening connections on port 54347
Realmlist (Realm Id: 3) configured in port 54347
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in /root/wow3/bin/maps. Total size: 291014339 bytes
vmaps directory located in /root/wow3/bin/vmaps. Total size: 658130553 bytes
mmaps directory located in /root/wow3/bin/mmaps. Total size: 2192911264 bytes
Default DBC locale: zhCN.
All available DBC locales: zhCN
Using World DB: ACDB 335.14-dev
Latest LoginDatabase update: 2025_07_24_00.sql
Latest CharacterDatabase update: 2025_09_03_00.sql
Latest WorldDatabase update: mod-zone-difficult.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-difficult
|- mod-eluna

[...]

PR #23390, opened 2025-10-25
Spell/GameObject: Fix flying upon teleport between map initiated from GO

killerwife:
Fixes flying upon using GO to travel to different map from lets say shattrath

Those teleporters are specifically marked as Usable as mounted per gameobject_template

image


Rorschach91:
Tested. It works.

r0m1ntik:
Tested. It works.

grace-of-hainaut:
This appears to resolve #23178 :)

Issue #23297, opened 2025-10-18
[WoTLK] Invisible passengers on player mounts


chauusie:
### Triage

Issue linked from CC:  https://github.com/chromiecraft/chromiecraft/issues/8381

Triage Notes: 

.additem 44234

Learn it. Mount. Eject passengers and have another char mount up.

I could replicate this on the left passenger side (Left from PoV of mammoth rider). I could see my alt in the right hand side. Alt swapping from right to left made alt invisible and came with massive FPS lag for me.

.additem 41508 (Horde bike thing)

Have alt sit in the vehicle, is invisible to rider. 

https://github.com/user-attachments/assets/dc453b09-9797-4f6e-8f17-74f496a1c594

Original Post Below: 

---

### Current Behaviour

While getting on mount for multiple players one player is becoming invisible (it lasts untill he change sit (in mamoths cases))

Traveler's tundra mammoth:
This is how its looking from pasager POV (there are total 3 characters on mount):
<img width="1917" height="1078" alt="Image" src="https://github.com/user-attachments/assets/8d886b4c-12f3-4e98-862c-e32cc990548f" />
This is how mount is looking from its ovner POV (there were total 3 characters on it):
<img width="187" height="231" alt="Image" src="https://github.com/user-attachments/assets/43464c5a-2fac-4bc4-beae-9e932b2ea304" />
How is it looking after sit change (only 2 total characters on mount):
<img width="1918" height="1078" alt="Image" src="https://github.com/user-attachments/assets/5b59aba7-4ae1-4211-8a4c-6e66459969fb" />
Vendors that appears at start are displayed correctly and can be seen by both player who ride and everyone around him.

Chopper:
While someone is getting on chopper first there is animation for additional sit:
<img width="1591" height="962" alt="Image" src="https://github.com/user-attachments/assets/50a7cb22-fd14-48ff-bcc8-677b88e036ec" />
but after short time it hides because there is "no player sitting":
<img width="1417" height="845" alt="Image" src="https://github.com/user-attachments/assets/58a72b75-aca6-45b9-8f49-8b904b65a3ee" />
On top of that after mount owner dismount player he was lifting become invisible untill he moves:
<img width="1419" height="619" alt="Image" src="https://github.com/user-attachments/assets/8c52e357-4fcc-48f4-82c3-b291e1a89bb4" />
[...]

PR #23406, opened 2025-10-25
Movement/Vehicle: Fix invisible vehicle passengers

killerwife:
Aims to fix:

https://github.com/azerothcore/azerothcore-wotlk/issues/23297

Please, I tested vehicles, but do test any vehicular combinations in the outdoors (pls no instance scripted ones, those might not be my fault) if they have any issues.


killerwife:
Just fyi, those copilot remarks are crap in this pr.

TheSCREWEDSoftware:
I don't think co-pilot reviews are always to be taken 100% it may mentioned certain things that are actually important, what's important is the human factor as always

lucafolloni:
Tested on a private test server. Seems to be working like a charm! 🚀

Issue #21691, opened 2025-03-11
[spell] thePaladin's Consecration range is very small. can not attack most of big boss


kissingers:
### Current Behaviour

When fighting large-model bosses like Felmyst, Brutallus, and Kil'jaeden, Consecration often doesn't register any damage statistics even if in the boss's red round. Only when you get extremely close to the boss does it start to show damage, but then the boss usually retreats, and the damage tracking is lost again. This indicates that the area on the ground where Consecration can actually hit the boss is extremely small. However, Druids' Swipe, which is also an AoE (Area of Effect) skill, works fine in these situations. I suspect there might be an issue with how the boss models' ground projection corresponds to the skill calculations.

### Expected Behaviour

 Consecration will same range take damage to boss such as Swipe and other range spell

### Steps to reproduce the problem

just fight any large boss and lookup spell Consecration damage

### AC rev. hash/commit

AzerothCore rev. d4b1c795c886 2025-03-11 06:13:10 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
服务器运行时间: 48 second(s)
Update time diff: 101ms. Last 119 diffs summary:
|- Mean: 100ms
|- Median: 100ms
|- Percentiles (95, 99, max): 101ms, 101ms, 157ms
Using SSL version: OpenSSL 3.0.15 3 Sep 2024 (library: OpenSSL 3.0.15 3 Sep 2024)
Using Boost version: 1.74.0
Using CMake version: 3.25.1
Using MySQL version: 80404
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.8.12-8-pve
Worldserver listening connections on port 54347
Realmlist (Realm Id: 3) configured in port 54347
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in /root/wow3/bin/maps. Total size: 291014339 bytes
vmaps directory located in /root/wow3/bin/vmaps. Total size: 658130553 bytes
mmaps directory located in /root/wow3/bin/mmaps. Total size: 2192911264 bytes
Default DBC locale: zhCN.
All available DBC locales: zhCN 
Using World DB: ACDB 335.13-dev
Latest LoginDatabase update: 2025_02_16_01.sql
Latest CharacterDatabase update: 2025_03_09_00.sql
[...]

PR #23452, opened 2025-10-28
Fix range spell effect range for big module object(most of Boss)

kissingers:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 
https://github.com/azerothcore/azerothcore-wotlk/issues/21691

## SOURCE:

## Tests Performed:

This PR has been:
- [X] Tested in-game by other community members/someone else other than the author/has been live on production servers.

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Tereneckla:
A real effort to solve the problem like here https://github.com/TrinityCore/TrinityCore/pull/16290 should be made.

We currently only can toggle using the object size of both source and target at once. This spell group probably should target size but not source size

kissingers:
> A real effort to solve the problem like here [TrinityCore/TrinityCore#16290](https://github.com/TrinityCore/TrinityCore/pull/16290) should be made.
> 
> We currently only can toggle using the object size of both source and target at once. This spell group probably should target size but not source size

Yes, but now not like blizz, because I remeber when year 2009 I use Paladin as a tank fight with Brutallus, my spell Consecration can hit the boss. and now if you want hit boss, must go ahead more close the boss, but boss will go back a little, so always can not hit boss. also other dps Paladin need very close to center of boss then can take effect of the spell, I dont think this is right way.
you can test all sunwell boss, they all have same problem. if you can fix this problem by other way may more good. such as fix only this spell range?

Tereneckla:
Can you try this one? https://github.com/Tereneckla/azerothcore-wotlk/commit/706c737dd8676d4a5d0103a6840ee733d00ea89b

Consecration hits and the fires on Archimonde still felt ok. But I would appreciate your input.

kissingers:
Close it as here will fix the problem
https://github.com/azerothcore/azerothcore-wotlk/pull/23478

PR #23465, opened 2025-10-29
Fix dying from fall damage and incorrect flying on map teleport

killerwife:
Aims to fix both dying from fall damage after teleport and flying after using loaned mounts

image


r0m1ntik:
Tested

Nyeriah (approved):

Issue #23468, opened 2025-10-29
[Zul'Aman] Timed event doesn't start.


Rorschach91:
### Triage

Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/8408

### Current Behaviour

1. Since the last CC restart on October 29th I heard reports of Zul'Aman functioning correctly EXCEPT that the NPC Harisson Jones doesn't die and the timer never starts. This means no extra loot is available, including the War Bear
2. When I test on PTR or my local install of AC, I can start the event and open the door. Harisson Jones goes stealth but then nothing happens. The enemies which are supposed to rush in never attack, and they're not attackable either. The rest of the instance appears to be fine but the timer never begins

### Expected Blizzlike Behaviour

The timer should begin normally, the enemies should rush in and kill Harrison Jones, and extra loot is possible if you beat the timer.

### Source

Worked just a few days ago

### Steps to reproduce the problem

1. get in a raid group
2. .tele za
3. Talk to Harisson Jones and bang on the gong

### Extra Notes

I also tested this on my local AC install, so this is not a generic CC bug

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/cbec99efb3d1c0d2c5a9428b0ae5641fc4ff54c0

### Operating system

Ubuntu 20.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-chromiecraft-smartstone](https://github.com/chromiecraft/mod-chromiecraft-smartstone)
- [mod-costumes](https://github.com/azerothcore/mod-costumes)
[...]

PR #23472, opened 2025-10-29
fix(Scripts/Misc): Adjust points from old wrong 0 based indexing to smart ai indexing

killerwife:
Fixes https://github.com/azerothcore/azerothcore-wotlk/issues/23468

Also moves all points previously that were DB -1, to DB numbering 1:1.

Hopefully everyone understands how having half of core be -1 is a landmine.

Pls check if i missed any, i tried to look everywhere thats impacted (smart scripts and c++ waypoint_movement_type)


Rorschach91:
Tested. Zul’Aman timer and Bloody Breakout seem to be working well now, and I don’t think the other edits will cause any problems.
I’ll let Nyeriah approve and merge it.

Nyeriah (approved):

PR #19687, opened 2024-08-19
fix(Core/Movement) Don't move for change in orientation only

Saqra1:
Changing only the orientation of a unit also flags it as if it were moving, which causes spells to be interrupted.

Checking if the unit is commanded to be moved to the same coordinates is an easy way to tell if it was an orientation only change. It is a workaround for the movement system but this way it won't require bigger changes.

## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:

- Closes #15913
- Closes https://github.com/chromiecraft/chromiecraft/issues/5385

## SOURCE:

The changes have been validated through:
- [x] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #23479, opened 2025-10-30
Fix: Distract spell stops target movement

DealsBeam:
The Distract spell was causing the target to stop moving because it was forcing the target to face the caster. This was happening because the `SetFacingTo` function in `Spell::EffectDistract` was clearing the movement stack.

This commit fixes the bug by removing the call to `SetFacingTo`, which prevents the movement stack from being cleared. A test case has been added to verify that the original movement is not interrupted.


Tereneckla:
Did you also test that Distract still forces the targets to face the caster?

sudlud:
Thanks for taking the time to dig into Distract and for adding a test, it's appreciated. After reviewing this against current master though, there are enough open issues that I'm going to close this for now. To be clear, please feel free to reopen or follow up at any time if you want to take another pass.

**Root cause looks incorrect.** The description says `SetFacingTo` was "clearing the movement stack." `SetFacingTo` (`Unit.cpp`) never touches the `MotionMaster` stack; it just launches an orientation-only movespline and returns. The MotionMaster generators are left intact, so the stated mechanism doesn't hold.

**Part of the original bug was already fixed.** Commit 62b825f8bd (#19687, "Don't move for change in orientation only") made orientation-only splines (exactly what `SetFacingTo` produces) stop flagging the unit as moving, and it updated this line's comment from "stops moving + interrupts spellcast" to just "stops moving." So the repro should be re-validated on current master first, since part of it is stale.

**The change removes intended behavior.** Facing the thrown location is the point of Distract (distracted mobs turn to look at it). This is what @Tereneckla asked about above and it wasn't resolved. Removing `SetFacingTo` deletes that turn entirely.

**It also doesn't actually stop the pause.** `MoveDistract` is still called and installs a `DistractMovementGenerator` at `MOTION_SLOT_CONTROLLED`, which suspends the active movement and adds `UNIT_STATE_DISTRACTED` for the duration. So the target still pauses its current movement regardless of this change; that part is by design.

**The test can't build or detect the bug.** A few blockers: `new Unit(false)` (Unit's only ctor takes no args and the class is abstract), `Unit` has no `Create()` method, and `sSpellMgr->GetSpellInfo(1725)` returns null in the gtest harness (no DBC/DB loaded), so `prepare()` would crash with no Map. The harness pattern is `TestMap` / `TestCreature` / `WorldMock` installed into `sWorld` (see `SnakeTrapEvadeTest`). Even setting that aside, the assertion checks `MotionMaster()->size()` goes 1 to 2, but since `SetFacingTo` never touches that stack, `size()` is 2 with or without the removed line, so the test passes on unpatched master too and guards nothing.

If you'd like to pursue this, the path forward would be: confirm the symptom still reproduces on current master and document the exact repro, fix it while preserving the face-the-distraction behavior, and add a test that compiles under the gtest harness and asserts something that actually differs between fixed and unfixed code. Thanks again, and feel free to reopen whenever.

Issue #23127, opened 2025-10-05
Feature: MMR-based matchmaking system for random battlegrounds


blinkysc:
### Describe your feature request or suggestion in detail

Currently, AzerothCore only matches battleground players based on level brackets and faction. This often results in severely imbalanced matches where one team significantly outgears or outskills the other, leading to poor player experience and discouraging participation in random battlegrounds.
        
        I propose implementing an MMR (Matchmaking Rating) system for random battlegrounds that considers both player skill and gear quality when forming teams. This system would track individual player performance across battleground matches and attempt to create balanced teams when assembling players from the queue.
        
        The goal is to provide fairer, more competitive matches for solo and random group players by ensuring teams are relatively balanced in terms of both player ability and character power level. This would make battlegrounds more engaging and rewarding for players of all skill levels.

### Describe a possible solution to your feature or suggestion in detail

The implementation would require several components working together:
        
        A battleground rating value stored per character that adjusts based on match outcomes, similar to how arena MMR currently functions. This rating would increase with wins and strong performance, and decrease with losses. The system should account for the team's overall performance rather than punishing individual players for unavoidable losses.
        
        A gear scoring calculation that evaluates a character's equipped item levels and overall power. This doesn't need to be complex—a simple average item level calculation would suffice, possibly weighted toward weapons and major armor pieces. This prevents newly dinged max-level characters from being matched against fully pvp-geared players.
        
        Database schema changes would include adding a new table or extending the existing character tables to store the battleground MMR value and potentially a cached gear score value. Something like adding `bg_mmr` and `bg_gear_score` columns to the characters table, with the gear score recalculated periodically or on equipment changes.
        
        The matchmaking algorithm would need modification to consider these new factors when assembling teams from the queue. When players queue for random battlegrounds, the system would attempt to balance teams so that the average MMR and gear score of each faction is roughly equivalent. This balancing would work within the existing level bracket system.
        
        For the initial rating, new characters or those without battleground history could start at a baseline MMR value (perhaps 1500, similar to arena starting ratings). The gear score would be calculated on-demand based on current equipment.
        
        The system should be configurable through worldserver.conf to allow server administrators to enable/disable the feature, adjust the weight given to MMR versus gear score in matchmaking, set the starting MMR value, and tune how quickly ratings adjust after matches.
        
        **Technical Implementation References:**
        
        Several proven rating algorithms could be adapted for this purpose:
        
[...]

PR #23513, opened 2025-11-02
feat(Core/Battlegrounds) Add skill-based matchmaking with Glicko-2 rating system

blinkysc:
- Zero synchronous DB queries during gameplay
- Minimal memory footprint (~60 bytes per player)
- No impact on BG completion performance
- Queue relaxation system for low-population servers
- Combined MMR + gear score weighting

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/23127

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Create 2+ characters
2. Queue for a battleground with both characters (one Alliance, one Horde)
3. .debug bg
4. Complete the battleground
5. Check ratings with .bgmmr info (requires GM level)
6. Verify database: SELECT * FROM character_battleground_rating;
7. Check history: SELECT * FROM character_battleground_rating_history;

- [ ]
- [ ]

### DIFF
--- a/src/test/server/game/Battlegrounds/Glicko2Test.cpp
+++ b/src/test/server/game/Battlegrounds/Glicko2Test.cpp
@@ -1,0 +1,310 @@
+/*
+ * This file is part of the AzerothCore Project. See AUTHORS file for Copyright information
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by the
+ * Free Software Foundation; either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "Glicko2.h"
+#include "gtest/gtest.h"
+#include <cmath>
+
+/**
+ * Glicko-2 Algorithm Unit Tests
+ *
+ * Tests the core Glicko-2 rating algorithm to ensure mathematical
+ * correctness and proper handling of edge cases.
+ */
+
+class Glicko2Test : public ::testing::Test
+{
+protected:
+    Glicko2System glicko{0.5f}; // Standard tau value
+
+    // Helper to compare floats with tolerance
+    bool FloatNear(float a, float b, float epsilon = 0.5f)
+    {
+        return std::abs(a - b) < epsilon;
+    }
+};
+
+// =============================================================================
+// Basic Rating Update Tests
+// =============================================================================
+
+TEST_F(Glicko2Test, RatingIncreasesAfterWin)
+{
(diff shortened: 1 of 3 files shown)


data/sql/updates/pending_db_characters/glicko2.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE `character_battleground_rating` (
+  `guid` INT UNSIGNED NOT NULL COMMENT 'Character GUID',
```
heyitsbench: I don't have the link handy, but there is a wiki page detailing style for DB naming. Mostly capitalization on column names is what I'm thinking of.

blinkysc: Fixed with https://www.azerothcore.org/wiki/sql-standards

data/sql/updates/pending_db_characters/glicko2.sql
```diff
@@ -0,0 +1,48 @@
...
+ROW_FORMAT = DEFAULT
+;
+
+DROP TABLE IF EXISTS `character_battleground_rating_history`;
+CREATE TABLE `character_battleground_rating_history` (
+  `Id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `Guid` INT UNSIGNED NOT NULL COMMENT 'Character GUID',
+  `OldRating` FLOAT NOT NULL,
+  `NewRating` FLOAT NOT NULL,
+  `OldRD` FLOAT NOT NULL,
```
heyitsbench: What is this, 'rating deviation'? IMO should be given a comment, or better yet un-acronymize it. Goal should be to have someone be able to know what these are for without consulting the wiki. 😛

blinkysc: Yeah sorry, you're right probably best not to have acronyms haha

heyitsbench: Some acronyms are fine, just because they're so common (ID/GUID in particular), but yea, things like this it doesn't hurt to have a little longer column name.

Nyeriah:
Does look like module stuff to me? I don't think rated bgs were a thing in wrath at all? Also curious about how it behaves with cf bg as that one queue in particular used to be pretty problematic

blinkysc:
> Does look like module stuff to me? I don't think rated bgs were a thing in wrath at all? Also curious about how it behaves with cf bg as that one queue in particular used to be pretty problematic

They had separate rated BG's later yeah. Not exactly sure they had system like this working silently but from my experience it definitely felt more balanced in the past. Some games certainly were steamrolls but overall could feel the balance (witness bias ftw). There is also blue posts about them looking at balancing based on gear and other factors too (this was more focused towards twinking). https://www.bluetracker.gg/wow/topic/us-en/19110336708-blizz-non-xp-battlegrounds/ All the posts about BG balancing was at the end of wrath going into Cata. Obviously this "system" would have been invisible to the client. For the cf bg not sure what bg you're referring too? Maybe isle? Might be worth just removing that and AV from this system entirely lol

blinkysc:
Ported to module https://github.com/blinkysc/mod-glicko2-mmr would love to still get any feedback =D. Also only real problem is the hooks needed for the current matchmaking system for the relaxing system

PR #23521, opened 2025-11-02
Fix root/stun client freeze

killerwife:
Attempts to fix root freeze reported by @r0m1ntik


r0m1ntik:
I’ve tested it 7–8 times and can no longer reproduce the crash after the fix.

Nyeriah (approved):

PR #23585, opened 2025-11-09
fix(DB/Loot): Fireproof Satchel will now always drop the Ritual of Torch 

TheSCREWEDSoftware:
Ritual Torch from 96.1% to 100%
Orb of Returning from 99.8% to 85%

Also Ritual Torch now only drops people welding the quest, I have not validated it, but you only can get these Satchel by having the same quest. The important part this fixes the CHANCE of a shamna player not getting the ritual torch.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes N/A

## SOURCE:

## Tests Performed:

This PR has been:
- [x] No tested just ran the SQL with no errors

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1762695074553179900.sql
+++ b/data/sql/updates/pending_db_world/rev_1762695074553179900.sql
@@ -1,3 +1,3 @@
 -- Makes sure the Shaman player gets the required ritual toch item and that requires the quest to drop
 UPDATE `item_loot_template` SET `Chance` = 0, `QuestRequired` = 1, `GroupId` = 1 WHERE `Entry` = 24336 AND `Item` = 23682;
-UPDATE `item_loot_template` SET `Chance` = 0 WHERE `Entry` = 24336, `GroupId` = 2 AND `Item` = 24335;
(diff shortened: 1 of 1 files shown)


Nyeriah:
I think the chance should be 0 and they should have different loot ids so both of them drop always. Whoever wrote this originally probably just copy pasted it off wowhead which we know can't be taken blindly when it comes to loot, or anything else really

TheSCREWEDSoftware:
When I get back to the pc I can re-check this, I would check mop but takes to long to level a dreanei.

The torch should be 100% (or 0 for the same result) the org looking at wowhead tbc, mop, cata or wrath it's 84-85 so is not always gurantee i will assume

TheSCREWEDSoftware:
@Nyeriah 

image

This is how it looks with the PR. 100% the quest item and 85% the orb of teleportation

TheSCREWEDSoftware:
I can't valid it this on officially servers as the torch doesn't drop on retail or mop, watching videos from tbc it seems they 100% gained from the bag, they are set 0 now, not grouped so they are 100% if not mistaken

PR #23627, opened 2025-11-11
Refactor UpdateSoulboundTradeItems method - crashfix

nl-saw:
Refactor UpdateSoulboundTradeItems to improve item checks and erase logic.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:
Crashfix:
[ae28c3e1dcc4_worldserver.exe_.11-11_15-57-46.txt](https://github.com/user-attachments/files/23482012/ae28c3e1dcc4_worldserver.exe_.11-11_15-57-46.txt)

Fixed a crash in Player::UpdateSoulboundTradeItems() caused by dereferencing invalid or deleted Item* pointers stored in m_itemSoulboundTradeable.

Cause

m_itemSoulboundTradeable could contain dangling Item* references if items were deleted, traded, or modified elsewhere (e.g., by Playerbots or inventory cleanup) without removing them from the list.
Although access to the list was protected by m_soulboundTradableLock, the lock did not prevent the function from accessing freed memory.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/PlayerStorage.cpp
+++ b/src/server/game/Entities/Player/PlayerStorage.cpp
@@ -4104,7 +4104,7 @@ void Player::TradeCancel(bool sendback, TradeStatus status /*= TRADE_STATUS_TRAD
     }
 }
 
-void Player::UpdateSoulboundTradeItems() // -- saw
+void Player::UpdateSoulboundTradeItems()
 {
     std::lock_guard<std::mutex> guard(m_soulboundTradableLock);
     if (m_itemSoulboundTradeable.empty())


TheSCREWEDSoftware:
This is your fix, a cherry pick or AI generated?

TheSCREWEDSoftware:
Just realised that this crash is a playerbots' crash, be free to submit a PR there.

nl-saw:
The crash was triggered by Playerbots, but the unsafe pointer access is in core code (Player::UpdateSoulboundTradeItems in PlayerStorage.cpp`).
This fix prevents a potential null/dangling pointer dereference in all cases, not just when bots are involved.
So even though Playerbots expose it more often, the root cause is in the core, and the fix properly belongs here.
Can you elaborate as to why proposed fix is not appropriate? The crashlog is provided; I would like some feedback as where to look if this isn't the right place.

walkline:
I think the more interesting question is why we have this dangling pointer in this data structure. 
Do you have any ASAN logs, or is the issue relatively easy to reproduce? 

From a dangling pointer perspective, the fix is not bulletproof, since we don't know what can be written in the place of the deleted object.
So, the most appropriate fix would be to remove the pointer from the data structure upon item deletion.

nl-saw:
Thanks @walkline for those pointers. It can very well be that it happens due to Playerbots - I'll further investigate in tracking down and removing the pointer on deletion. This will take me some time.

My patch was primarily meant as a safety net to prevent runtime crashes.

Cheers

Issue #23644, opened 2025-11-13
[Zul'Drak] Betrayal quest doesn't work correctly.


Rorschach91:
### Triage
Original CC Report: https://github.com/chromiecraft/chromiecraft/issues/8532

The event seems to have a reset issue.
To reproduce the bug, complete the event once, then try the quest again (it doesn’t matter whether you use the same character or another one).

### Current Behaviour

Quest ID: 12713 (https://wowgaming.altervista.org/aowow/?quest=12713)

This is the last quest in a long chain. After getting the quest the player has to use choker (ghoul costume) and talk to Drakuru who will teleport the player to the top platform where the player has to control a Blightblood troll to fight Drakuru. Drakuru DOES teleport the player to the platform as intended, but the (talking and fighting) event is not triggered so nothing happens while on the platform.

Note: The platform has a "Drakuru's Last Wish" object which teleports the user to ground level when used (FYI)

### Expected Blizzlike Behaviour

Drakuru event (fight) starts while on the platform.

### Source

https://www.wowhead.com/wotlk/quest=12713/betrayal

### Steps to reproduce the problem

1. (Do prequests)
2. Take quest Betrayal from Stefan at Ebon Watch (or you can summon him with an item)
3. Take the choker from the same NPC.
4. Wear your ghoul costume by using the choker.
5. Run into the portal and teleport inside the flying scourge structure where Drakuru resides.
6. Talk the Drakuru and get teleported to the top of the structure.
7. Observe nothing happening (event not triggering)

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/9429817e82f95eb285b2ac31fa706d464fd0b662

### Operating system

Ubuntu 20.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
[...]

PR #23650, opened 2025-11-13
fix(Scripts/Northrend): Betrayal quest

PkllonG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/23644

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Northrend/zone_zuldrak.cpp
+++ b/src/server/scripts/Northrend/zone_zuldrak.cpp
@@ -590,7 +590,6 @@ struct npc_overlord_drakuru_betrayal : public ScriptedAI
             case EVENT_BETRAYAL_EPILOGUE_10:
                 _state = BETRAYAL_EVADE;
                 EnterEvadeMode(EVADE_REASON_OTHER);
-                
                 break;
         }
 


src/server/scripts/Northrend/zone_zuldrak.cpp
```diff
@@ -375,7 +376,7 @@ struct npc_overlord_drakuru_betrayal : public ScriptedAI
     {
         if (Player* player = who->ToPlayer())
         {
-            bool shouldStartEvent = (_state == BETRAYAL_NOT_STARTED) && IsPlayerOnQuest(player) && player->HasAura(SPELL_SCOURGE_DISGUISE) && player->IsWithinDistInMap(me, 80.0f);
+            bool shouldStartEvent = (_state == BETRAYAL_NOT_STARTED) && IsPlayerOnQuest(player) && player->HasAura(SPELL_SCOURGE_DISGUISE) && player->IsWithinDistInMap(me, 80.0f) && !me->FindNearestGameObject(GO_DRAKURUS_BONE, 80.0f);
```
sogladev, line 379: good

src/server/scripts/Northrend/zone_zuldrak.cpp
```diff
@@ -587,7 +588,7 @@ struct npc_overlord_drakuru_betrayal : public ScriptedAI
                     lich->GetMotionMaster()->MovePoint(0, 6141.2393, -2011.2728, 589.8653);
                 break;
             case EVENT_BETRAYAL_EPILOGUE_10:
-                EnterEvadeMode(EVADE_REASON_OTHER);
+                Reset();
```
sogladev:
```suggestion
                _state = BETRAYAL_EVADE;
                EnterEvadeMode(EVADE_REASON_OTHER);
                
```

same behavior, but more consistent with manually handling Evade

Rorschach91:
Tested. The event seems to reset properly now.

rorshan:
I tested this. Here's the behavior I observed:
- If a player triggers the tp to the top while the fight is already ongoing or Drakuru's Skull is still on the ground, a new Drakuru won't appear. However the skull can be looted by everyone, whether in a party or not. This tracks with comments I could find on wowhead
- Apart from the first spawn after a restart, the subsequent spawns Drakuru is shown as already being tagged by someone (grey bar). This in itself is not an issue since the quest doesn't actually make you kill him, but the next part is
- Sometimes Drakuru can spawn without being full HP which makes the fight easier. I've seen him spawn with as low as 2% HP. He also doesn't regenerate at all in the start of the cutscene (which would be an easy fix for this issue). This is probably related to the previous point.
- The players tend to stay stuck in combat after the event. Using the portal TP does clear combat, but its still not ideal.
- The teleporting crystal never seems to despawn. This is minorly illogical if Drakuru is alive and the crystal exists since its name is "Drakuru's Last Wish"

Addionnal notes for further testing:

`.q a 12713`
`.add 38699` for choker
`.go c 103709`to tp to drakuru

sogladev:
@rorshan thanks for testing this. 

This PR can be merged, but for a future PR:

> * Apart from the first spawn after a restart, the subsequent spawns Drakuru is shown as already being tagged by someone (grey bar). This in itself is not an issue since the quest doesn't actually make you kill him, but the next part is

I think the grey tag for a second player is unavoidable. I'm not sure what you mean by the next part

> * Sometimes Drakuru can spawn without being full HP which makes the fight easier. I've seen him spawn with as low as 2% HP. He also doesn't regenerate at all in the start of the cutscene (which would be an easy fix for this issue). This is probably related to the previous point.

problematic. I expect he should regen combat once the epilogue begins. Maybe `me->CombatStop();` is not the correct method?

> * The players tend to stay stuck in combat after the event. Using the portal TP does clear combat, but its still not ideal.

This PR should fix this issue.

> * The teleporting crystal never seems to despawn. This is minorly illogical if Drakuru is alive and the crystal exists since its name is "Drakuru's Last Wish"

This behavior is observed in the sniff. Once the event starts, the crystal is destroyed. I can only deduce the crystal should be there if the event is not active

rorshan:
> I think the grey tag for a second player is unavoidable. I'm not sure what you mean by the next part

Next part was the HP not full issue


> problematic. I expect he should regen combat once the epilogue begins. Maybe `me->CombatStop();` is not the correct method?

Yes, I noticed that Drakuru stil shows that red glow (which i think means he's still aggressive towards me?) around his nameplate even though he turns allied for the part with Lich King.
Thematically he should not regen HP during the cutscene just before being killed. That's also not what happens in videos. Ideally there should be a way to put him out of combat but not let him regen HP until after the cutscene ends.
 
> > * The players tend to stay stuck in combat after the event. Using the portal TP does clear combat, but its still not ideal.
> 
> This PR should fix this issue.

I tested with this PR on AC, so I'm not sure what you mean.

> This behavior is observed in the sniff. Once the event starts, the crystal is destroyed. I can only deduce the crystal should be there if the event is not active

Okay, it's just a little weird since it's not there before the first person does the quest after a restart, but it's not an issue.

PR #23655, opened 2025-11-13
fix(docs/license): use GPLv2 as MaNGOS-based project

FrancescoBorzi:
This is an important transition for the AzerothCore project to fix a mistake we made in the very beginning.

As you know, the project started based on the source code of the [SunwellCore](https://www.azerothcore.org/pages/sunwell.pl/) project. Unfortunately, the source code was released without any git history or license.

We assumed we could apply the AGPL license to it, as the original MaNGOS files mentioned the possibility of redistributing under "either version 2 of the License, or (at your option) any later version". We later learned this assumption was a mistake and for that we take full responsibility.

It is very important for us to clarify that we applied such a stricter license to prevent anyone from improving the AzerothCore source code without contributing back, following the open-source collaboration spirit on which AzerothCore and its ancestors have been based.

Our intention has never been to hinder the cross-project collaboration with any other MaNGOS-based open-source project, which we have always explicitly encouraged, but we do acknowledge that because of the license issue, some people have felt discouraged from porting AzerothCore to other projects, and because of this, we offer our sincere apologies.

This issue has been brought to our attention a long time ago already, however, it took quite some time for us to understand the possibilities and take the right steps to fix this mistake properly. This is why it took so long. After a lot of thoughts, research and conversations, around one year ago we [announced the transition to GPLv2](https://github.com/azerothcore/azerothcore-wotlk/discussions/19868) and began to collect [consent from the majority of our contributors](https://github.com/azerothcore/azerothcore-wotlk/discussions/19869).

It is now time to complete this important process, hoping that this could build more bridges between the AzerothCore community and other MaNGOS-based open-source projects, following the true spirit of the free-software collaboration which has always been the main motivation behind working on our beloved project.

----------------------------

The license file has been copied from https://github.com/cmangos/mangos-wotlk/blob/master/LICENSE.md

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -81,8 +81,7 @@ You can check the [authors](https://github.com/azerothcore/azerothcore-wotlk/blo
 
 ## License
 
-- The new AzerothCore source components are released under the [GNU AGPL v3](https://www.gnu.org/licenses/agpl-3.0.en.html)
-- The old sources based on MaNGOS/TrinityCore are released under the [GNU GPL v2](https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html)
+- The AzerothCore source code is released under the [GNU GPL v2](https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html)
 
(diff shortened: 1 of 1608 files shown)


Nyeriah (approved):

pangolp:
Hopefully, this will set a precedent for the project, and we can learn from our mistakes. Apologies are valid, and a group of people who are no longer involved in the project cannot make decisions for those who are currently participating. Their opinions should be considered valid while they are active, and if they are no longer involved, they should support those who are currently working on and improving the project. A single mistake should never define a project, especially not an open-source one.

PR #23688, opened 2025-11-14
docs(README): Add BountyHub badge

omarsoufiane:
There are currently 11 bounties for the Azerothcore project on BountyHub, and this badge would make them more visible for people who would like to contribute


sudlud:
I think the placement in the build status section is not perfect

omarsoufiane:
@sudlud Yes you are right my mistake

PR #23759, opened 2025-11-18
Fix(Core/ScriptedCreature): SetInvincibility must be after ScheduleHealthCheckEvent

PkllonG:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


Nyeriah:
@PkllonG could you provide more information on what bug is this supposed to address?

PkllonG:
> @PkllonG could you provide more information on what bug is this supposed to address?

```
    if (IsInvincible() && damage >= me->GetHealth())
        damage = me->GetHealth() - 1;
```
Because 'damage' can be changed.
We should handle the HealthCheckEvent before 'damage' is changed, as it is more precise.

Nyeriah:
> “We should handle the HealthCheckEvent before ‘damage’ is changed, as it is more precise.”

That doesn’t make any sense to me — if anything, I’d say it’s quite the opposite.

Take this example:

A creature has 20,000 HP and an event set to trigger at 50%.  
It gets hit for 10,000 damage.  
That damage is calculated as 10000 for the HP trigger then reduced to 5,000.  

The event still fires, even though the actual damage taken was lower.  

So the trigger point becomes inaccurate rather than more precise. We are not only dealing with invincibility case here but also all other scripts that lower damage

If you have any bug reports or a concrete reproducible case that would justify this change, please link it. At this point I'm not convinced that we need this adjustment, and I'm concerned it may introduce unintended side effects, especially around how damage is processed and how thresholds are evaluated.

PR #23805, opened 2025-11-20
fix(Core): Clarify quest level hiding configuration comments

NoxMax:
## Changes Proposed:

Minor comment fix. Current comment implies that quests that are EQUAL TO Low/HighLevelHideDiff or more are hidden, when it's actually only quests that are MORE THAN Low/HighLevelHideDiff that are hidden. The logic can be seen at [PlayerQuest.cpp](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Entities/Player/PlayerQuest.cpp) with `CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF` and `CONFIG_QUEST_HIGH_LEVEL_HIDE_DIFF`.

Screenshot for example. If the comment was correct, then at default value of 4, this level 14 marker should be hidden from level 18:
Untitled

Second, `-1 - (Disabled, Show all available quest marks)` is removed as a possible option. Note the comment and code logic in WorldConfig.cpp:
```c++
// note: disable value (-1) will assigned as 0xFFFFFFF, to prevent overflow at calculations limit it to max possible player level MAX_LEVEL(100)
SetConfigValue<uint32>(CONFIG_QUEST_LOW_LEVEL_HIDE_DIFF, "Quests.LowLevelHideDiff", 4, ConfigValueCache::Reloadable::Yes, [](uint32 const& value) { return value <= MAX_LEVEL; }, "<= MAX_LEVEL");
SetConfigValue<uint32>(CONFIG_QUEST_HIGH_LEVEL_HIDE_DIFF, "Quests.HighLevelHideDiff", 7, ConfigValueCache::Reloadable::Yes, [](uint32 const& value) { return value <= MAX_LEVEL; }, "<= MAX_LEVEL"); 
```
A config of -1 is rejected as the bound is MAX_LEVEL. It errors out if you use it:
```
> Config: Bad value defined for name 'Quests.LowLevelHideDiff', going to use '4'
 instead
> Config: Bad value defined for name 'Quests.HighLevelHideDiff', going to use '7
' instead 
```

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


TheSCREWEDSoftware:
I dont recall now, does the comment for the Creature level is properly phrased?

E.g Level 9 Player would see a level 4 grey (a 5 green)
A level 8 player would see a level 4 and 5 green

NoxMax:
This config is about hiding lower level quests, as a static difference value. So with the default 4 value, at level 20 quests that are < 16 will be hidden, and at level 80 quests that are < 76 will be hidden.

If you are referring about when a green quest/creature becomes grey, that is not a static value. A level 80 will see a quest/creature as low as 72 (8 diff) as green, but a level 20 will see a quest/creature as low as 14 (6 diff) as green.

This table lists it all (and it's correct as far as WOTLK) https://warcraft.wiki.gg/wiki/Mob_difficulty_colors

NoxMax:
Also [reading more about it in old wikis](https://wowwiki-archive.fandom.com/wiki/Quest#Low_level_quests) and new ones, I am now pretty sure that the default value should be -1. This setting as a static value, is pretty helpful, but it's not aligned with WOTLK. WOTLK (and even modern retail) hides quest markers when the quest turned grey, not based on a fixed value difference. The -1 value does that, which also means the comment for -1 is misleading:

`#                     -1 - (Disabled, Show all available quest marks)`

Currently the comment for the -1 value implies that if you are level 21 for example, a level 11 quest should show its marker. That's not what happens as I just tested. It will be hidden because it's a grey quest. But, if you set LowLevelHideDiff = 10, then the level 11 marker will appear to a 21.

Need someone else to confirm my thought process. If true I will edit the values/comments here and the defaults at WorldConfig.cpp

PS: Testing was done without "Low Level Quests" checked on the minimap (default)

NoxMax:
Also removed the mention of -1 as an option. It's not valid.

PR #23857, opened 2025-11-22
fix(Scripts/UBRS): Remove SetInvincibility calls in Gyth script

sogladev:
## Changes Proposed:

This PR proposes changes to:

`.damage 50000` kills Gyth and summons Rend

related:
- https://github.com/azerothcore/azerothcore-wotlk/pull/23759

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_gyth.cpp
+++ b/src/server/scripts/EasternKingdoms/BlackrockMountain/BlackrockSpire/boss_gyth.cpp
@@ -69,12 +69,9 @@ public:
                 me->DespawnOrUnsummon();
             }
 
-            SetInvincibility(true); // Don't let boss die before summoning Rend.
-
             ScheduleHealthCheckEvent(25, [&] {
                 DoCastAOE(SPELL_SUMMON_REND, true);
                 me->RemoveAura(SPELL_REND_MOUNTS);
-                SetInvincibility(false);
             });
         }
 


Nyeriah:
I'm afraid this reintroduces the bug it was added to address (vide comment)

sogladev:
> I'm afraid this reintroduces the bug it was added to address (vide comment)

I had to check and this fixes the `boss is left at 1hp` issue mentioned. Updated the title/commit

Nyeriah:
I gotta check something

Nyeriah:
Okay, so the issue we had was that the dragon would die before using the spell that summons Rend. Back then I remember having done some research and it seems to me that creatures that got HP events that must always fire do get stuck on 1 hp in order to perform it, the same with creatures on cutscenes/scripted sequences

I just went back to check it again and confirmed on the MOP PTR 


https://github.com/user-attachments/assets/6e32c216-3fca-4f50-852c-78c67db746ab

You can see Gyth is stuck at 1 hp and even after a couple white hits it doesn't die until Rend is spawned, something like a 1-2 sec invulnarability time

So while this PR doesn't reintroduce the bug it is still not the intended behaviour

sogladev:
> HP events that must always fire do get stuck on 1 hp in order to perform it, the same with creatures on cutscenes/scripted sequences

This SetInvincibility is a hack to mimic this stuck at 1 hp behavior. I'll close this until we can do this more generally.

I feel retail HP checks are done on a 'tick', and creatures cannot die if there are unhandled HP events. On AC, hp events are checked every time damage is taken, which is more expensive. We could change the Boss HP events to be tick-based? 

One case I considered is that during Ulduar's Iron Council fight 32857 [`Stormcaller Brundir`](https://www.wowhead.com/wotlk/npc=32857/stormcaller-brundir#abilities;mode:n10) cannot die before casting Supercharge, which buffs the other bosses on death. If the boss is casting [Overload](https://www.wowhead.com/wotlk/spell=61869/overload), it cannot be killed and becomes stuck at 1hp and can only die after the cast is done. I guess they use a 1hp health check, queue it and execute once it is no longer casting

Nyeriah:
> One case I considered is that during Ulduar's Iron Council fight 32857 [Stormcaller Brundir](https://www.wowhead.com/wotlk/npc=32857/stormcaller-brundir#abilities;mode:n10) cannot die before casting Supercharge, which buffs the other bosses on death. If the boss is casting [Overload](https://www.wowhead.com/wotlk/spell=61869/overload), it cannot be killed and becomes stuck at 1hp and can only die after the cast is done. I guess they use a 1hp health check, queue it and execute once it is no longer casting

This is already the case on AC if a flag is set on the healthcheck event, it won't fire until they are done casting. While the check is done on every tick its down to checking if _nextHealthCheck isn't 0, which if we are talking about resource usage should be neglectable as about the same as a bool check

Issue #23897, opened 2025-11-25
ICC - problem with NPC flight animation


jorge990125:
### Current Behaviour

The flying units that I will mention below do not perform the correct animation:

37230 -- Spire Frostwyrm
37533 -- Rimefang
37534 -- Spinestalker
36853 -- Sindragosa

### Expected Behaviour

They will have to fly instead of walking on air.

### Source

https://github.com/user-attachments/assets/1428a721-0f8e-42ac-8414-7f4c0b378427

### Steps to reproduce the problem

1. Go to Icecrown Citadel.
2. Go to the aerial battle encounter where the first NPC, Spire Frostwyrm, will appear.
3. Go to the Sindragosa encounter where the remaining NPCs that have the Rimefang bug will appear: Spinestalker, Sindragosa.

### AC rev. hash/commit

[AzerothCore](https://github.com/azerothcore/azerothcore-wotlk/commit/10ec051c58b498976f32576ebc42690b325cc28f)

### Operating system

windows 10 pro

### Custom changes or Modules

|- mod-ale
|- mod-archmage-timear
|- mod-arena-replay
|- mod-breaking-news-override
|- mod-cfbg
|- mod-congrats-on-level
|- mod-duel-reset
|- mod-global-chat
|- mod-hardcore-makgora
|- mod-instance-reset
|- mod-learn-spells
|- mod-mythic-plus
|- mod-npc-buff
|- mod-npc-enchanter
|- mod-npc-promotion
|- mod-npc-services
|- mod-npc-spectator
|- mod-npc-talent-template
|- mod-premium
|- mod-profesiones
|- mod-pvp-titles
|- mod-raid-dashboard
|- mod-recruitafriend
|- mod-skip-dk-starting-area
|- mod-solo-lfg
|- mod-solocraft
|- mod-streams
|- mod-system-vip
|- mod-teleport
|- mod-transmog

Kacsacsont:
Same in Ulduar with Razorscale. She doesn’t fly, she walks in the air.

jorge990125:
It seems there are several flying NPCs that have issues with their flight animation. However, the solution is quite simple: in the creature_template_addons table, you just need to set bytes1 = 50331648 for the NPC. This fixes the flight animation. That said, if the NPC has a ScriptName, it might encounter some issues repositioning on the ground.

kadeshar:
EoE Malygos also affected

[...]

PR #23901, opened 2025-11-25
fix(DB/ICC) Solution to NPC flight animations

jorge990125:
37230 -- Spire Frostwyrm
37533 -- Rimefang
37534 -- Spinestalker
36853 -- Sindragosa

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:
https://github.com/azerothcore/azerothcore-wotlk/issues/23897

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1764102580140254300.sql
+++ b/data/sql/updates/pending_db_world/rev_1764102580140254300.sql
@@ -1,10 +1,7 @@
+--
 -- flight animation correction
 -- https://www.wowhead.com/npc=37230/spire-frostwyrm
 -- https://www.wowhead.com/npc=37533/rimefang
 -- https://www.wowhead.com/npc=37534/spinestalker
 -- https://www.wowhead.com/npc=36853/sindragosa
-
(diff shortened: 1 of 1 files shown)


Nyeriah:
I think this is a hack as it should be set dynamically at run time according to the movement flag updates

cc @killerwife

killerwife:
Fully agree with that. Its called animtier, and also changes upon special takeoff and landing motions. Prolly needs to be added to smartscripts if it isnt alrdy.

jorge990125:
I'm not really sure if this is the correct way or if there are other ways to correct the flight animation. I've done it this way: creature_template_addons bytes1 = 50331648. I've checked the other flying units and they have the same value in creature_template_addons as a reference.

killerwife:
The problem with that field is afaik it rewrites the field on each aggro/evade. Pretty sure on sindragosa at least you want to manage it manually.

jorge990125:
So, in the case of Sindragosa, would the solution be directly in the boss's code, so that when it appears it sets the animation and when it lands it deactivates

PR #23913, opened 2025-11-26
fix(Core/SmartAI): SMARTCAST_COMBAT_MOVE prevents movement on successful cast

blinkysc:
SMARTCAST_MAIN_SPELL System

  This adds support for ranged casters to maintain proper spell range distance instead of closing to melee.

  How it works:
  - At initialization, SmartAI scans for spells with SMARTCAST_MAIN_SPELL flag (0x400) to set the creature's attack distance
  - If no SMARTCAST_MAIN_SPELL is found, it falls back to the first spell with SMARTCAST_COMBAT_MOVE flag (0x40)
  - The attack distance is calculated as: spell max range - melee range

  Usage:
  - Existing creatures with SMARTCAST_COMBAT_MOVE will automatically work (seamless)
  - Scripters can use SMARTCAST_MAIN_SPELL to explicitly designate which spell determines chase distance
  - Useful for creatures with multiple ranged spells where the default pick isn't ideal

  Example:
  -- Add SMARTCAST_MAIN_SPELL (1024) to a spell that already has SMARTCAST_COMBAT_MOVE (64)
  -- Combined flags: 64 | 1024 = 1088
  UPDATE smart_scripts SET action_param2 = 1088 WHERE entryorguid = 203 AND action_type = 11 AND action_param1 = 20792;

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

Ok list of test NPC's

SELECT DISTINCT ct.entry, ct.name
  FROM creature_template ct
  JOIN smart_scripts ss ON ss.entryorguid = ct.entry AND ss.source_type = 0
  WHERE ss.action_type = 11
  AND (ss.action_param2 & 64) = 64
  LIMIT 15;

1. .go c id 203 
2. Make sure casts and stops appropriately
3. Should chase when out of range

- [ ]
- [ ]

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -942,7 +942,6 @@ void SmartAI::InitializeAI()
 {
     GetScript()->OnInitialize(me);
 
-    // Scan for SMARTCAST_MAIN_SPELL first (explicit)
     for (SmartScriptHolder const& event : GetScript()->GetEvents())
     {
         if (event.GetActionType() != SMART_ACTION_CAST)


src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -745,13 +748,18 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
...
-                            // if we are just out of range, we only chase until we are back in spell range.
+                        if (result == SPELL_FAILED_OUT_OF_RANGE)
+                        {
                             CAST_AI(SmartAI, me->AI())->SetCurrentRangeMode(true, std::max(spellMaxRange - NOMINAL_MELEE_RANGE, 0.0f));
-                        else // move into melee on any other fail
-                            // if spell fail for any other reason, we chase to melee range, or stay where we are if spellcast was successful.
+                        }
+                        else if (result == SPELL_CAST_OK)
+                        {
+                            CAST_AI(SmartAI, me->AI())->SetCombatMovement(false, false);
```
killerwife: So realistically, you should never do this. This is a command for scripters. Thats kinda what I removed from the AI alltogether. So that scripters, no matter what is going on automatically, have the last say if a mob moves or not.

That out of the way, what are you trying to accomplish.

src/server/game/AI/SmartScripts/SmartScript.cpp
```diff
@@ -715,6 +715,9 @@ void SmartScript::ProcessAction(SmartScriptHolder& e, Unit* unit, uint32 var0, u
                         if (me->IsRooted()) // Rooted inhabit type, never move/reposition
                             continue;
 
+                        // Re-enable combat movement so we can chase
+                        if (e.action.cast.castFlags & SMARTCAST_COMBAT_MOVE)
+                            CAST_AI(SmartAI, me->AI())->SetCombatMovement(true, false);
```
killerwife: Same here.

src/server/game/AI/SmartScripts/SmartScript.cpp
blinkysc: AttackStart keeps getting called which triggers MoveChase(who, 0) - constantly resetting the NPC to chase to melee. So even though the cast succeeds at range, the NPC immediately walks toward the target

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -1080,7 +1080,17 @@ void SmartAI::SetCurrentRangeMode(bool on, float range)
     _attackDistance = range;
 
     if (Unit* victim = me->GetVictim())
-        me->GetMotionMaster()->MoveChase(victim, _attackDistance);
+    {
+        float currentDist = me->GetDistance(victim);
```
killerwife: I think this code here is a bit misunderstood. What you are trying to do here, I think, is only turn on the movement if those conditions are met. Instead, you want MoveChase to always be there, and the change in attackDistance be as a quasi "message" to the class. And then do the decision making whether to move completely automatically in ChaseMovementGenerator

src/server/game/AI/SmartScripts/SmartScript.cpp
killerwife: That means the AttackStart must be fixed to MoveChase with proper distance. If I remember correctly, the system only knows the cast distance once first cast happens, so to fix this, the information must be known upon  the AttackStart call so that _attackDistance can be passed properly.

To explain the final desired result, blizzard has always 1 spell marked as main ability when its a caster, and the chase range is based upon that spell. For Defias Pillager that would be fireball for example.

blinkysc: Is it CAST_AI that is meant for scripters? and should be removed entirely?

blinkysc:
Don't mind all the nonsense atm pushed commit with some of my debug lol xD

blinkysc:
So right now essentially 1st spell found is selected as "main spell" where _attackDistance is set... I hate it... I will look at this more over little later but does work...

killerwife:
Nice. You can later look at this: https://github.com/cmangos/mangos-wotlk/blob/master/src/game/AI/BaseAI/UnitAI.cpp#L1176

TheSCREWEDSoftware:
Not sure if this issue can be related to a regression on this PR? https://github.com/azerothcore/azerothcore-wotlk/issues/24378 @blinkysc

PR #23965, opened 2025-11-29
fix(Core/Player): prevent null pointer dereference in SendLoot function

r0m1ntik:
- Added a null check for the game object before calling ForceValuesUpdateAtIndex to avoid potential crashes when the object is not valid.
In Player::SendLoot the condition checks !go, but then go->ForceValuesUpdateAtIndex is called without verifying that go is not NULL. If GetMap()->GetGameObject(guid) returns NULL, this causes a NULL pointer dereference.

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -7880,6 +7880,7 @@ void Player::SendLoot(ObjectGuid guid, LootType loot_type)
         {
             if (go)
                 go->ForceValuesUpdateAtIndex(GAMEOBJECT_BYTES_1);
+
             SendLootRelease(guid);
             return;
         }


sudlud:
i think this needs a deeper check to correct this - right before your changed part it actually enters that section for `if (!go)`

looks like this is there since the first commit of AC

r0m1ntik:
> i think this needs a deeper check to correct this - right before your changed part it actually enters that section for `if (!go)`
> 
> looks like this is there since the first commit of AC

Yes, that’s possible, the crash only started appearing after the recent changes.

Unfortunately I don’t have the full crash log anymore, but from what I kept, everything was pointing to this area:

The crash happens in Player::SendLoot (around line 8007), when loot_type = LOOT_FISHING.
go is 0x0 (the GameObject* is NULL), but ForceValuesUpdateAtIndex is still called on it.

The backtrace I had looked roughly like this:

```
#0  UpdateMask::SetBit (this=0x58, index=17) at UpdateMask.h:45
    → this=0x58 (invalid/corrupted pointer)

#1  Object::ForceValuesUpdateAtIndex (this=0x0, i=17) at Object.cpp:2082
    → this=0x0 (NULL pointer!)

#2  Player::SendLoot (this=0x7ffe9dd15040, guid=..., loot_type=LOOT_FISHING)
    at Player.cpp:8007
    → go = 0x0 (GameObject* was NULL)

#3  GameObject::Use (this=0x7ffecce09a00, user=0x7ffe9dd15040) at GameObject.cpp:1797
#4  WorldSession::HandleGameObjectUseOpcode at SpellHandler.cpp:344
```

So it does enter the if (!go) section, but later go->ForceValuesUpdateAtIndex is called without ensuring go is non-NULL, which is what seems to be causing the crash.

Nyeriah:
Not discrediting this PR, but the way the logic is set currently looks a bit funny, as per sudlud's comment

PR #24067, opened 2025-12-07
Introduce animtier to malygos to fix flying animations

killerwife:
Animtier is a part of the SMSG_MONSTER_MOVE packet used for animation transitions.

After executing spline its set in unit field bytes offset animtier. Core had that already.

What core was missing is use of AnimTier::Fly - 3. This is directly tied to animations for entities and cannot be guessed from videos , but is in specific motions.

I had the luck of researching malygos beforehand and knowing what he should use. I suspect sindragosa and and other dragons (NOT sapphiron, he doesnt use this, must not be used blindly) in wotlk use this. Must be used in combination with proper SetHover and SetGravityDisable order per sniffs.

Issue related: https://github.com/azerothcore/azerothcore-wotlk/issues/23897

Also the added AnimTier function should going forward be the only way of setting AnimTier on spawn. Likely need to add it to smart scripts.

### DIFF
--- a/src/server/scripts/Northrend/Nexus/EyeOfEternity/boss_malygos.cpp
+++ b/src/server/scripts/Northrend/Nexus/EyeOfEternity/boss_malygos.cpp
@@ -279,12 +279,6 @@ public:
                     case MI_POINT_SURGE_OF_POWER_CENTER:
                         events.RescheduleEvent(EVENT_SURGE_OF_POWER_WARNING, 0ms, 1);
                         break;
-                }
-            }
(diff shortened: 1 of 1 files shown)


PkllonG:
Malygos: This is truly an unfortunate matter.

Rorschach91:
@killerwife I retested the PR and noticed that when Malygos lands, he remains unattackable.

killerwife:
Done

blinkysc:
@Nyeriah Perhaps this is why Felmyst was such a battle?

killerwife:
Felmyst does not use this either.

Issue #23992, opened 2025-11-30
Auth server starts complains about missing property


silviu20092:
### Current Behaviour

Starting authserver throws:

> Config: Missing property Network.UseSocketActivation in config file /home/silviu/acore/config/authserver.conf, add "Network.UseSocketActivation = 0" to this file or define 'AC_NETWORK_USE_SOCKET_ACTIVATION' as an environment variable.

### Expected Behaviour

Should not show the error

### Source

Since  https://github.com/azerothcore/azerothcore-wotlk/commit/dc8318cf6a236322eea7a3ac0b3a37f6b756f1b3

### Steps to reproduce the problem

1. Open Authserver
2. See the warning

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/ff161bb82414817772ba01fa9bb9b81369cc1886

### Operating system

Ubuntu 22.04

sudlud:
the new config value needs to be added to the `authserver.conf.dist` so it can easily be copied over to the `authserver.conf`

@ovv can you please maybe prepare a pr for this? also is there anything else to consider or can be used socket activation for the authserver, too, without any issues?

ovv:
Sure I can do a follow up.

I don't think socket activation will work for the authserver. I haven't tried it. I'll look into it.

In the meantime is there a way to have a value only for the worldserver ?

Gravesly:
Not sure if correct way to fix it but i added Network.UseSocketActivation = 0 one line under the AllowLoggingIPAddressesInDatabase = 1 in the authserver.conf and seems to stop the error from happening

Dezaimasu:
This property `Network.UseSocketActivation` exists in `worldserver.conf.dist` though. Also there's no `Network.` namespace mentioned in `authserver.conf.dist`, at all. Maybe somthing got mixed up here?

PR #24075, opened 2025-12-08
fix(Core/Config) - Add UseSocketActivation to auth config

TheSCREWEDSoftware:
This PR
image

Current AC
image

## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/23992

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/apps/authserver/authserver.conf.dist
+++ b/src/server/apps/authserver/authserver.conf.dist
@@ -447,6 +447,9 @@ Logger.root=4,Console Auth
 #
 #    Network.UseSocketActivation
 #        Description: [LINUX ONLY FEATURE] Enable systemd socket activation support for the authserver.
+#                     When enabled and the process is started by systemd socket activation,
+#                     the server will use the socket passed by systemd instead of
+#                     creating and binding its own listening socket. Disabled by default.
 #
 #        Example:     1 - (Enabled)
 #        Default:     0 - (Disabled)


TheSCREWEDSoftware:
@ovv any correctiosn are needed?

sudlud:
issue here currently is that @ovv mentioned this functionality is probably not working at all on authserver as of now. so what about adding the value plus the comment that it's a worldserver feature, non-functional as of now for authserver and is just placed there to fix the startup warning for now.
let's see what @ovv has to say

ovv:
I haven't tested it with the authserver. Maybe it's working, maybe it's not.

If nobody does it first I plan to test it and make it work for the authserver when I find the time

In the meantime, if there is an easy way to skip this configuration for the authserver maybe that's the best first step. I'm not familiar enough with the codebase to know if something like that already exist. Otherwise what @sudlud proposed is fine with me, whatever is easier for the project.

src/server/apps/authserver/authserver.conf.dist
```diff
@@ -438,5 +438,19 @@ Appender.Auth=2,5,0,Auth.log,w
...
 Logger.root=4,Console Auth
 
+#
+#    Network.UseSocketActivation
+#        Description: [LINUX ONLY FEATURE] Enable systemd socket activation support for the authserver.
+#                     When enabled and the process is started by systemd socket activation,
+#                     the server will use the socket passed by systemd instead of
+#                     creating and binding its own listening socket. Disabled by default.
+#
+#                     When enabled the realm is not automatically set as offline on shutdown.
```
ovv: This isn't applicable to the authserver

Nyeriah:
In this case I think we should add it to that list of configs to suppress instead, since its not working, and also I don't think the same config being in both places is a good design

TheSCREWEDSoftware:
If everyone's is fine as it is, as temporarly solution to avoid the errors it's there a placeholder

TheSCREWEDSoftware:
Ups

sudlud:
Maybe also add a header for the new network section? Like in worldserver config

TheSCREWEDSoftware:
Since the this is just there to prevent the error, therefor the placeholder text, i'm unsure if it works, as far as i understand the only reason auth gets this error is because the implementation is using a shared file which both use

sudlud:
Yeah I know, I'd just like to have the section split off correctly in the config file when we add this new config for now. So could you please add something like

### network
########

Above, as this is the first network setting for now in the authserver config file

TheSCREWEDSoftware:
> Yeah I know, I'd just like to have the section split off correctly in the config file when we add this new config for now. So could you please add something like
> 
> ### network
> ########
> 
> Above, as this is the first network setting for now in the authserver config file

Done

ovv:
I was able to run some tests today. The authserver works fine with socket activation, no changes are needed.

To test, enable the option in the authserver.conf and then setup socket activation in systemd.

In `/etc/systemd/system/authserver.socket`
```
[Unit]
Description=AzerothCore Authserver socket

[Socket]
ListenStream=0.0.0.0:3724

[Install]
WantedBy=sockets.target
```

In `/etc/systemd/system/authserver.service`
```
[Unit]
Description=AzerothCore Authserver
After=network.target mysql.service worldserver.socket
Requires=authserver.socket

[Service]
Type=exec
Restart=always
RestartSec=60
User=azerothcore
WorkingDirectory=/opt/azerothcore
ExecStart=/usr/bin/stdbuf -o0 /opt/azerothcore/dist/bin/authserver --config /opt/azerothcore/dist/etc/authserver.conf
SyslogIdentifier=authserver
```

Adding the configuration option in the authserver.conf is all that's needed. It doesn't have to be a placeholder.

TheSCREWEDSoftware:
Any suggestion for the description?

ovv:
> Any suggestion for the description?
I think we can use the same one as the worldserver

TheSCREWEDSoftware:
```
When enabled and the process is started by systemd socket activation,
#                     the server will use the socket passed by systemd instead of
#                     creating and binding its own listening socket. Disabled by default.
#
#                     When enabled the realm is not automatically set as offline on shutdown.
```
fine then?

ovv:
> When enabled the realm is not automatically set as offline on shutdown.

We can remove this, it doesn't apply to the auth server. The rest is fine :+1:

sudlud:
Last minor nitpick would be that there's usually no empty line after the comment block

TheSCREWEDSoftware:
> Last minor nitpick would be that there's usually no empty line after the comment block

Which line

sudlud (approved):

PR #24081, opened 2025-12-11
feat(apps/cmake) - Implement Python config merger tool

TheSCREWEDSoftware:
Improvement / Addition to the existing PHP version: https://github.com/azerothcore/azerothcore-wotlk/pull/5779
and based on and modified from: https://github.com/Brian-Aldridge/update_module_confs

- With a quick CMAKE option allows you to have (or not by default) python script alongside your config files to allow merging of changes (from the dist) which you're missing from your config file.
- It's in python, so it's just plug and play with no need of e.g web server for the case of PHP implementation and is welcoming to new users (install python and run it and select the options)
- This was only tested on windows, would appreciate a lot if someone can test this on Linux and if even possible on MacOS before being merged.
- Supports Full CLI 

Video of how it works: https://www.youtube.com/watch?v=34lMbHCFhpc

- Divded https://github.com/azerothcore/azerothcore-wotlk/tree/master/apps/config-merger folder into 2 sub folders PHP and Python, added/updated a readme to the config-merger to brief explanition of both and one of the python.

- License added to the Python file due some code base being based on MIT License, can be used under GPL with no issues

### DIFF
--- a/apps/config-merger/python/README.md
+++ b/apps/config-merger/python/README.md
@@ -14,7 +14,7 @@ This tool compares your existing configuration files (`.conf`) with the latest d
 - **Interactive Menu System** - Easy-to-use numbered menu options
 - **Server Config Support** - Update authserver.conf and worldserver.conf
 - **Module Config Support** - Update all or selected module configurations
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
Tested in linux, and removed ready to review to see if i can do a full implemention for CLI without confirmations as suggested

PR #24141, opened 2025-12-17
fix(DB/Gameobject): Move Unreachable Mining Nodes in Swamp of Sorrows

Gultask:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Closes https://github.com/chromiecraft/chromiecraft/issues/7605

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1766002787364237500.sql
+++ b/data/sql/updates/pending_db_world/rev_1766002787364237500.sql
@@ -1,2 +1,2 @@
 --
-UPDATE `gameobject` SET `position_z` = -6.5780087, `comment` = 'Original position_z: -15.8641', `VerifiedBuild` = -52237 WHERE `guid` IN (208397, 208414, 208431);
+UPDATE `gameobject` SET `position_z` = -6.5780087, `comment` = 'Original position_z: -15.8641', `VerifiedBuild` = -52237 WHERE `guid` IN (208397, 208414, 208431) AND `id` IN (2040, 1734, 2047);


Gultask:
Authored by @Rorschach91 
Opened a new one cause the branch was deleted. This keeps the original position in a comment field, so best of both worlds, no?

data/sql/updates/pending_db_world/rev_1766002787364237500.sql
```diff
@@ -0,0 +1,2 @@
+--
+UPDATE `gameobject` SET `position_z` = -6.5780087, `comment` = 'Original position_z: -15.8641' WHERE `guid` IN (208397, 208414, 208431);
```
heyitsbench: Might want to specify that it is a sniffed position.

Gultask: I'd say it's implicit, but we can also change the VerifiedBuild to negative. I think that also implied modified data back in the day before we had a comment field.

sudlud:
So the z was sniffed before but underground?

Gultask:
Yep. Happens quite often, actually.

Nyeriah:
I'm a bit confused, was the previous value sniffed? And the new value is sniffed too? If its just the usual node being underground I think we shouldn't touch those anymore, its sort of a blizzlike bug it seems

Gultask:
The unreachable position is sniffed.
The new position is custom.
The comment field with the original position maintains the sniff data in the database in a way, and the VerifiedBuild being negative makes it clear it was changed.

If we keep the unreachable nodes, we'll keep getting reports/complaints/questions unless we remove them entirely. And this isn't uncommon, there are many unreachable nodes which I previously did the same process for the same reason :P 
I believe this is the best compromise.

Nyeriah:
It just sounds like a bit of clutter to me and now we know blizz indeed kept them underground to get hackers https://github.com/TrinityCore/TrinityCore/issues/31500#issuecomment-3657580639

blinkysc:
Maybe these nodes should have separate pools and anti cheat triggers assigned to them?

Gultask:
The anti-cheat system would be ancient and not worth it imo. Only Old World Azeroth has these. Eversong Woods has one peacebloom flying at character head height. Outland nodes are hand placed and I hadn't seen a case of unreachable. Northrend is a mix, although I haven't seen a case either. Borean Tundra is hand-placed, for example, every node is part of a group of 4 spawn points.
I have seen tons of unreachable nodes in WoD of all expansions, and Deepholm. In those, nodes look like they've been sprinkled in rather than hand-placed. There are thousands of spawn points.

The reason I believe this particular one was an accident is that it has an underground location beneath it, the Temple of Atal'Hakkar, so height level calculation might've gotten an error somewhere.

blinkysc:
Maybe then much like the HoS doors these should just be turned off?

Gultask:
I mean, it's a trivial problem, really. I see no downsides either way.

TheSCREWEDSoftware:
I understand us previously adjusting the cords for objects under the ground (as we didn't knew until recently their reason), since we do know their reason, probably the best solution if we want to stay true to the 3.3.5a (since we tend to emulate the bugs of the time) is that we keep these un-changed from the sniff values, but going with blinky's idea having these pooled, or flagged in the core (alongside with a DB comment for these specific objects) or at least a worldconfig option (to take or not action if a player interacts with an object that only hackers would, could be just a curious new user to AC, therefore a config to enable that behaviour) have these flag or not a player (either kick, ban, log whatever choice is taken)

data/sql/updates/pending_db_world/rev_1766002787364237500.sql
```diff
@@ -0,0 +1,2 @@
+--
+UPDATE `gameobject` SET `position_z` = -6.5780087, `comment` = 'Original position_z: -15.8641', `VerifiedBuild` = -52237 WHERE `guid` IN (208397, 208414, 208431);
```
sudlud: and id = ... :p

Gultask: image

sudlud: cat

sudlud (approved):

Issue #24122, opened 2025-12-16
[Cross-faction] Mounts with passengers can't be used in cross-faction


TheSCREWEDSoftware:
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/8420
Triage Notes: 
Original Post Below: 

---

### Current Behaviour

The NPCs created by Traveler's Tundra Mammoth (spellid 61425) are not crossfaction.
<img width="523" height="574" alt="Image" src="https://github.com/user-attachments/assets/52451135-b3f8-46ae-a042-d472cfbb1faf" />
Tested on PTR.

### Expected Blizzlike Behaviour

Crossfaction is not Blizzlike.

That said, the expectation would be for the NPCs to be crossfaction, as with repair bots.

### Source

No source to confirm.

### Steps to reproduce the problem

1. .additem 44234
2. learn Reins of the Traveler's Tundra Mammoth
3. /cast Traveler's Tundra Mammoth
4. invite second character of the opposing faction (Horde/Alliance) to party.
5. Confirm NPCs are hostile.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/b74d3dc62be1b21c4bcf86d454b5ef04fd140ee3

### Operating system

Ubuntu 20.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-chromiecraft-smartstone](https://github.com/chromiecraft/mod-chromiecraft-smartstone)
- [mod-costumes](https://github.com/azerothcore/mod-costumes)
[...]

PR #24142, opened 2025-12-17
fix(Scripts/NPC): Add Crossfaction functionality to Traveler Tundra's Mammoth vendors

Gultask:
## Changes Proposed:

This PR proposes changes to:

Moves the SmartAI script to a generic cpp script.
Changes the vehicle_template_accessory to avoid despawning immediately upon dismount

I'm just not sure if:
1. It's ok to put them under npc_special
2. It's ok to put the checks under Update()

I didn't quite understand how OnAuraRemoved would work, and this seemed like the next best option.
But overall it detects either the aura 61424 being removed, or not being on a vehicle, and proceeds with the timed script.

Timed script is sniffed from a build 4.4.2.59734 sniff

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/24122
- Closes https://github.com/chromiecraft/chromiecraft/issues/8420

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Set AllowTwoSide.Interaction.Group = 0 or 1
4. .cast 61447 if Alliance
5. .cast 61425 if Horde

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/World/npcs_special.cpp
+++ b/src/server/scripts/World/npcs_special.cpp
@@ -2696,7 +2696,7 @@ struct npc_traveler_mammoth_vendor : public ScriptedAI
         _hasEjected = false;
         _playerGuid.Clear();
 
-        if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_GROUP))
+        if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_GROUP) && !me->GetMap()->IsBattlegroundOrArena())
             me->SetFaction(FACTION_FRIENDLY);
 
         me->m_Events.KillAllEvents(false);


sogladev:
I wondered if this should be generic for vehicle passengers, but no other entries `vehicle_template_accessory` require cross-faction interaction.

For reference, TwoSideInteractionChecks is done here:
TwoSideIntereaction field flag update
- https://github.com/azerothcore/azerothcore-wotlk/blob/4fb7c0f9bf3444e7d08e5fe88ffe175907a72788/src/server/game/Entities/Unit/Unit.cpp#L21053
Reputation check on interact
- https://github.com/azerothcore/azerothcore-wotlk/blob/4fb7c0f9bf3444e7d08e5fe88ffe175907a72788/src/server/game/Entities/Player/Player.cpp#L2103

Adding more checks feels like a hack, so a script for this case is best.

Have you considered using a script on the mounts directly? You can set faction on passenger board, and keep existing SmartAI

```cpp
void PassengerBoarded(Unit* who, int8 /*seat*/, bool apply) override
    if (apply && who->IsCreature())
        who->SetFaction()
```
example usage:
- https://github.com/azerothcore/azerothcore-wotlk/blob/4fb7c0f9bf3444e7d08e5fe88ffe175907a72788/src/server/scripts/Northrend/zone_storm_peaks.cpp#L927

Gultask:
I think there is no mount, the player is the one acting as the vehicle (I believe), as there is not really an npc there. See the GetVehicleBase()? That's how we get the player guid for the despawn event :p . Previously the SAI checked if the vendors had the Aura at every update and caught any closest player (rare to cause issues since they occupy the same position as the player but it could)

As for the twoside interaction, I got from the chess event.

src/server/scripts/World/npcs_special.cpp
```diff
@@ -2679,6 +2679,60 @@ struct npc_controller : public PossessedAI
...
+
+    void Reset() override
+    {
+        _hasEjected = false;
+        _playerGuid.Clear();
+
+        if (sWorld->getBoolConfig(CONFIG_ALLOW_TWO_SIDE_INTERACTION_GROUP))
+            me->SetFaction(FACTION_FRIENDLY);
+
+        scheduler.CancelAll();
```
sogladev:
```suggestion
        me->m_Events.KillAllEvents(false);
```

Gultask:
> The only issue I found is that in PvP, if the player is flagged the vendors are friendly, but this is acceptable to me as it's a custom config

Oh, that is a good point I haven't thought about

Gultask:
I haven't been able to get the player by fetching the vehicle base at reset() without resulting in a nullptr crash, and I really don't like the idea of changing factions during the Update() loop :C 
At least I placed a BG and Arena exception there

PR #24242, opened 2025-12-29
fix(Scripts/Naxxramas): Build

Kitzunu:
Fix redefinition error


elthehablo:
Need to get my eyes checked. Thanks

Issue #24291, opened 2026-01-01
Lunar Festival Active and Many Holidays Missing from Calendar


LannyE:
### Current Behaviour

The Lunar Event is currently active on New Year's Eve and many of the holidays are missing from the 2026 Calendar.  I even updated to the newest clean build and the Lunar Event is still active.

I noticed these events missing from the Calendar:
Lunar Event
Love is in the Air
Noblegarden
Children's Week
Harvest Festival
Pilgrim's Bounty

It seems basically any event that shows up in the Holiday_Dates table except for the Darkmoon Faire events are missing from the Calendar.  Not sure why the Lunar Event would be active right now though. The game_event table says it shouldn't be on but if I stop it using GM commands and then restart the server it comes back.  

A Lunar Event NPC is present and gm command says it is active. 
![Image](https://github.com/user-attachments/assets/09efac66-e2ad-4197-95ac-2241d4b0e58c)

Missing Calendar Events
![Image](https://github.com/user-attachments/assets/a5c038aa-16c9-4dad-a647-d8a8f4d6ad28)

![Image](https://github.com/user-attachments/assets/650af89b-2390-4724-bc7e-ddc705678cb1)

### Expected Behaviour

Only the Winterveil and its events should be running at this time and the calendar should show the dates for the upcoming events.  
 

### Steps to reproduce the problem

1. Log in to server
2. Look for a Lunar Event Commoner or quest giver or gm command ".event activelist" and notice Lunar is running. 
3. Look at Calendar and notice all those events I mentioned are missing in the future.  

### AC rev. hash/commit

AzerothCore rev. 9785d5cc1620 2025-12-31 22:34:45 +0000 (master branch) (Win64, RelWithDebInfo, Static) (worldserver-daemon)

### Operating system

Windows 11

### Custom changes or Modules

None

Aladanth:
Have been investigating this on my own Windows 11 instance after noticing Lunar Festival activated yesterday. Modifying the start date in game_event table resolves the event being active too early after restarting worldserver, but the next state change is still incorrect. 

On further poking around, all annual holiday events tagged as **holidayStage > 0** have an occurrence of 360 days when pulling `.event info #` All annual events tagged as **holidayStage 0** correctly return an occurrence of 365 days with the next state change date being reported correctly. 

[...]

PR #2890, opened 2020-04-14
fix(Core/DB/Events): Game Event Start Date & Events on Calendar

Kitzunu:
https://github.com/TrinityCore/TrinityCore/commit/299d6ebbffd11a8fa92327c5a9296602a1596202
https://github.com/TrinityCore/TrinityCore/commit/87c79c9485cbb9536912e1566174c717946c1537
https://github.com/TrinityCore/TrinityCore/commit/4913edf81bfbd505e3218161b45705962876552d
https://github.com/TrinityCore/TrinityCore/commit/f41dfa0089a7ba8719f28816ae75ba91313df5c4

## CHANGES PROPOSED:
-  Implement holiday_duration in holiday_dates (holiday_duration is the number of hours of how long the holiday may last (visible in the calendar))
-  Update Start and End dates of events

## ISSUES ADDRESSED:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2550
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/542

## TESTS PERFORMED:
- Builds
- Checked in-game calendar and start dates. All bueno

## HOW TO TEST THE CHANGES:

1. Check calendar that all event frames are shown
2. .ev #event_id (To check start times)

## Target branch(es):
- [x] Master

 
## How to test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here in the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

PR #24293, opened 2026-01-01
fix(DB/Misc) Added 2026 holiday_dates for Calendar

afdz1:
## Changes Proposed:

This PR proposes changes to:
- Adds missing calendar dates for 2026

## Issues Addressed:

- Progresses https://github.com/azerothcore/azerothcore-wotlk/issues/24291

## SOURCE:

- [ ] https://www.wowhead.com/events#holidays;0-2+4

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [ ] Run SQL script and review calendar for holiday events

- [ ] This does not address the server having lunar festival running at incorrect times. It simply a calendar update for players.


sudlud:
Thanks for the pr! but I am preferring #24038 to dynamically calculate these dates in the future so we don't have to add them to the database anymore.

afdz1:
> Thanks for the pr! but I am preferring #24038 to dynamically calculate these dates in the future so we don't have to add them to the database anymore.

I agree that is a better approach. Some dates do float around though, so we may want to be cautious on how we address those. I.E Love is in the Air is not on a fixed day but on Feb 9 this year. Opposed to last year's Feb 11.

sudlud:
Yes this will be fixed by that PR also, correctly moving the events every year.

Up until now, AC always had them as fixed dates, so even if you e.g. move them in your PR here to a different date, the actual game event itself is beeing started by the configuration in the database table game_event, so calendar and actual event occurring could probably be misaligned.

sudlud:
so we could merge this as an intermediate solution until the dynamic calculation is ready.
the max index for date_id is 25 so you have to overwrite e.g. the oldest existing entry instead of adding index 26.

have you tested this pr yourself? because i would expect it not to be working in the current state.

edit:
nvm for some of these the index is even not up to 25 used yet if i've seen it correctly

sudlud:
see also #2890

afdz1:
> so we could merge this as an intermediate solution until the dynamic calculation is ready. the max index for date_id is 25 so you have to overwrite e.g. the oldest existing entry instead of adding index 26.
> 
> have you tested this pr yourself? because i would expect it not to be working in the current state.
> 
> edit: nvm for some of these the index is even not up to 25 used yet if i've seen it correctly

Thank you. Not sure how I missed the date_id limit, I thought I tested but it was late last night. Maybe I missed something (HNY). I noticed date_id 6 is not used for any of these events. Considering this is a intermediate solution, I believe it'll be safe here.

sudlud:
Thanks for the update. Did you test your changes in game yourself?

sudlud:
please check in game calendar if the dates there are correct, at least brewfest should be incorrect as you have to consider the extra 7 days for the building event

sudlud:
thanks again for this pr, but it has been superceeded now by the other pr beeing ready for merge.

PR #24373, opened 2026-01-10
fix(Core/Player): Fix Player::GetSpec talent ranks for inactive spec

TRPB:
## Changes Proposed:

This PR proposes changes to:

`Player::GetSpec` fetches the talents and ranks for a given spec index using this code

```cpp
                // find max talent rank (0~4)
                int8 curtalent_maxrank = -1;
                for (int8 rank = MAX_TALENT_RANK - 1; rank >= 0; --rank)
                {
                    if (talentInfo->RankID[rank] && HasTalent(talentInfo->RankID[rank], specIdx))
                    {
                        curtalent_maxrank = rank;
                        break;
                    }
                }
```

source: https://github.com/azerothcore/azerothcore-wotlk/blob/226591a02e3f9c4c18007776a3b31d20906ba7a3/src/server/game/Entities/Player/Player.cpp#L15983

However, it is fundamentally flawed because `HasTalent` has a second argument which is never used and `HasTalent(talentInfo->RankID[rank], specIdx))` always uses the active spec, the `specIdx` variable is irrelevant because it's ignored.

The result is that calling `GetSpec` makes all talents in the inactive spec (that aren't also in the active spec) appear as rank 0.

This PR fixes it by reinstating that second argument so that `HasTalent` can be sent the spec.

I have verified that all calls to `HasTalent` outside of `GetSpec` correctly send the active spec. E.g.

`src/server/game/Spells/Auras/SpellAuraEffects.cpp`:

```
            if (player->HasTalent(63410, player->GetActiveSpec()) || player->HasTalent(63411, player->GetActiveSpec()))

```

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## SOURCE:

## Tests Performed:

This PR has been:


src/server/game/Entities/Player/Player.cpp
```diff
@@ -3883,10 +3883,10 @@ bool Player::HasSpell(uint32 spell) const
     return (itr != m_spells.end() && itr->second->State != PLAYERSPELL_REMOVED && itr->second->IsInSpec(m_activeSpec));
 }
 
-bool Player::HasTalent(uint32 spell, uint8  /*spec*/) const
+bool Player::HasTalent(uint32 spell, uint8  spec) const
```
Kitzunu: remove the extra space between uint8 and spec

TRPB: good spot, I just removed the comment from the existing code

sudlud (approved):

sudlud (approved):
Approve. `HasTalent`'s `spec` parameter was commented out and the function always queried `m_activeSpec`, so `GetSpec(spec)` misreported talent ranks for the inactive spec. The fix wires the parameter through. Verified against current master: every other `HasTalent` caller passes `GetActiveSpec()`, so their behavior is unchanged and the only effect is fixing the inactive-spec path. No regression risk.

PR #24397, opened 2026-01-12
fix(Scripts/Spells): Shadowfiend should not restore mana while priest…

### LINKED ISSUE #22536: [Druid/Priest] Shadowfiend's attacks restore priest's mana while the priest is cycloned
### Current Behaviour

Currently, if [Cyclone](https://wowgaming.altervista.org/aowow/?spell=33786) is cast on a priest that has an active shadowfiend, cyclone is not preventing the priest from gaining mana. 

### Expected Behaviour

Cyclone should prevent all actions on the target.

### Source

I can't find proper sources for this, just how I have observed cyclone to work in general. 

1. Have a druid and a priest
2. Duel with them
3. Spend some mana on the priest, and cast shadowfiend on the druid.
4. Cast cyclone on the priest
5. Observer the priest keeps getting mana every time shadowfiend attacks.

### Extra Notes

When, for example, druid uses innervate and they are cycloned, they do not gain any mana from it.

### AC rev. hash/commit

AzerothCore rev. 50ff7ef67ba5 2025-07-22 06:24:30 +0000 (master branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 11 x64

### Custom changes or Modules

_No response_

xDevICCI:
Cyclone should prevent all actions on the target, including mana restoration from Shadowfiend attacks. Added check for `UNIT_STATE_ISOLATED` in `spell_pri_mana_leech::CheckProc`.

Closes #22536

## Changes Proposed:

### AI-assisted Pull Requests

Claude was used to help analyze the code and identify the fix.

## Issues Addressed:
- Closes #22536

## SOURCE:

The fix follows the same pattern used by `SPELL_AURA_PERIODIC_ENERGIZE` (Innervate) in `SpellAuraEffects.cpp:7162-7166`:
```cpp
if (target->HasUnitState(UNIT_STATE_ISOLATED))
{
    SendTickImmune(target, caster);
    return;
}
```

Innervate correctly stops restoring mana when the target is cycloned because it uses this check. Shadowfiend's mana proc should behave the same way.

## Tests Performed:

## How to Test the Changes:

1. Create a priest and a druid
2. Priest casts Shadowfiend on a target and starts getting mana
3. Druid casts Cyclone on the priest
4. Verify priest does NOT receive mana while cycloned
5. After Cyclone ends, verify Shadowfiend resumes restoring mana

- [ ] None known

### DIFF
--- a/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
+++ b/src/server/scripts/Northrend/ChamberOfAspects/RubySanctum/boss_halion.cpp
@@ -598,7 +598,7 @@ public:
 
         void SetData(uint32 id, uint32 value) override
         {
-            if (!events.HasTimeUntilEvent(EVENT_CHECK_CORPOREALITY))
+            if (!_events.HasTimeUntilEvent(EVENT_CHECK_CORPOREALITY))
                 return;
 
             if (id == DATA_MATERIAL_DAMAGE_TAKEN)


Nyeriah:
> The fix follows the same pattern used by SPELL_AURA_PERIODIC_ENERGIZE (Innervate) in SpellAuraEffects.cpp:7162-7166:

But it doesn't, it's leading to a completely different outcome

xDevICCI:
You're right, but I found something interesting:

The check for `UNIT_STATE_ISOLATED` already exists in `EffectEnergize` (SpellEffects.cpp:1888-1892):
```cpp
if (unitTarget->HasUnitState(UNIT_STATE_ISOLATED))
{
    m_caster->SendSpellDamageImmune(unitTarget, GetSpellInfo()->Id);
    return;
}
```

However, `UNIT_STATE_ISOLATED` is only applied when a spell has `MECHANIC_BANISH` (SpellAuraEffects.cpp:4255-4258).

Could someone verify if Cyclone (33786) has the correct mechanic in the DBC/DB? If it doesn't have `MECHANIC_BANISH`, that would explain why the existing check doesn't work.

TheSCREWEDSoftware:
image

It does have mechanic for Banished, you can also view the same info in: https://wowgaming.altervista.org/aowow/?spell=33786 which uses AC

Nyeriah:
It works for every other spell except shadowfiend, I don't believe it is sound to assume cyclone is the issue here

xDevICCI:
@Nyeriah Thanks for the feedback.  (if is possible, please review my last commit of this issue)

If the `UNIT_STATE_ISOLATED` check in `EffectEnergize` (SpellEffects.cpp:1888-1892) works for other spells but not Shadowfiend, there must be something specific about how Shadowfiend's mana proc works.

Looking at the code, the Shadowfiend uses `CastSpell(_procTarget, SPELL_PRIEST_MANA_LEECH_PROC, true, ...)` with triggered=true. Could the triggered flag be bypassing the immunity check? Or is there something else specific to spell 34650 that I'm missing?

Nyeriah:
> @Nyeriah Thanks for the feedback. (if is possible, please review my last commit of this issue)

Your latest commit doesn't compile, but even if it did, it's still wrong. You're preventing the proc altogether which to my knowledge is not what should be happening.

> Could the triggered flag be bypassing the immunity check? 

No

> Or is there something else specific to spell 34650 that I'm missing?

Check its spell effect and if its being handled at all by the energize handler

Issue #19658, opened 2024-08-17
Shaman Totem elementals are always hostile to opposite faction


BeefSteaker:
https://github.com/chromiecraft/chromiecraft/issues/7284
### What client do you play on?

enGB

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

Shaman Fire elemental and Earth elemental totems are hostile and attackable by the opposite faction to the Shaman that summoned the totem. This includes non PvP flagged players and those in a crossfaction Raid/Party. It can also attack non targetable enemy faction NPC's like the  PvP call to arms ones in shattrath

### Expected Blizzlike Behaviour

They should not automatically be hostile

### Steps to reproduce the problem

.learn 2894 or 2062
test with character of opposite faction

### Extra Notes

possibly related to https://github.com/azerothcore/azerothcore-wotlk/pull/19599

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/ee8c103d3d9b0473e8014f11bba910f2e8ee84ec

### Operating system

Windows 10

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-BreakingNewsOverride](https://github.com/AnchyDev/BreakingNewsOverride)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna](https://github.com/azerothcore/mod-eluna)
- [mod-instanced-worldbosses](https://github.com/nyeriah/mod-instanced-worldbosses)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-low-level-rbg](https://github.com/azerothcore/mod-low-level-rbg)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-progression-system](https://github.com/azerothcore/mod-progression-system)
[...]

PR #19599, opened 2024-08-12
fix(Core/Spells): Adjust ownership and follow logic for guardian summons.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:
-  [X] Core (units, players, creatures, game systems).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19514
- Closes https://github.com/chromiecraft/chromiecraft/issues/6583

## SOURCE:

The changes have been validated through:
- [X] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [X] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.

## How to Test the Changes:

- [X] This pull request can be tested by following the reproduction steps provided in the linked issue
- [X] This pull request requires further testing.
1. Observe summoned guardians follow target.
2. Confirm that they are correct.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #24402, opened 2026-01-12
fix(Scripts/Pets): Shaman elementals follow totem instead of player

### LINKED ISSUE #19514: Shaman Elementals follow the shaman instead of the totem.
### Current Behaviour

Currently when creating either a fire or an earth elemental totem, the elemental that is summoned will follow the player.

The elemental should stay with the totem.

### Source

Wrath Classic, [Wrath retail](https://www.youtube.com/watch?v=U9bm6I9G1_4&t=459s)

1. `.learn 2894`
2. Cast Fire Elemental Totem.
3. Observe the Elemental as it follows you instead of stands by the totem.

### Extra Notes

Originally reported at https://github.com/chromiecraft/chromiecraft/issues/6583.

### AC rev. hash/commit

d5c472b25d3197d8b474768144139c09406f057d

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

xDevICCI:
Closes #19514

Fire and Earth elementals now follow their respective totems instead of the shaman player. This is achieved by:
- Adding MoveToTotem() that finds the nearby totem and follows it
- Calling MoveToTotem() in InitializeAI() when elemental spawns
- Overriding EnterEvadeMode() to return to totem after combat

This approach keeps the elemental's ownership with the player (avoiding faction issues from reverted PR #19599) while fixing the movement behavior.

## Changes Proposed:
PR #19599 previously attempted to fix this by changing elemental ownership to the totem. However, this caused a regression (#19658) where elementals became hostile to opposite faction players, so it was reverted in #19664.

The root cause is that many pet-related checks assume the owner is a player. Changing ownership to the totem broke faction handling.

This PR takes a different approach: keep the elemental owned by the player but only change its follow behavior to target the totem. This is done in the elemental's AI script (`pet_shaman.cpp`) by:
- Finding the nearby totem using `FindNearestCreature()`
- Using `MoveFollow()` to follow the totem in `InitializeAI()`
- Overriding `EnterEvadeMode()` to return to the totem after combat

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [X] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.

Claude was used to analyze the codebase and help develop the solution.

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/19514

## SOURCE:
  - Video: https://www.youtube.com/watch?v=U9bm6I9G1_4&t=459s (Wrath retail)

## Tests Performed:
This PR has been:

## How to Test the Changes:

1. `.learn 2894` (Fire Elemental Totem) or `.learn 2062` (Earth Elemental Totem)
2. Cast the totem
3. Observe the elemental stays near the totem instead of following the player
4. Move away from the totem - elemental should remain near totem
5. Enter combat and exit - elemental should return to totem, not to the player

- None

### DIFF
--- a/src/server/scripts/Pet/pet_shaman.cpp
+++ b/src/server/scripts/Pet/pet_shaman.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "CreatureScript.h"
+#include "MotionMaster.h"
 #include "Player.h"
 #include "ScriptedCreature.h"
 
@@ -46,13 +47,34 @@ struct npc_pet_shaman_earth_elemental : public ScriptedAI
 {
     npc_pet_shaman_earth_elemental(Creature* creature) : ScriptedAI(creature), _initAttack(true) { }
 
+    void InitializeAI() override
+    {
+        MoveToTotem();
+    }
+
     void JustEngagedWith(Unit*) override
     {
         _events.Reset();
         _events.ScheduleEvent(EVENT_SHAMAN_ANGEREDEARTH, 0ms);
     }
 
-    void InitializeAI() override { }
+    void EnterEvadeMode(EvadeReason why) override
+    {
+        if (!_EnterEvadeMode(why))
+            return;
+
+        MoveToTotem();
+    }
+
+    void MoveToTotem()
+    {
+        if (Unit* owner = me->GetCharmerOrOwner())
+            if (Creature* totem = ObjectAccessor::GetCreature(*me, owner->m_SummonSlot[SUMMON_SLOT_TOTEM_EARTH]))
+            {
+                me->GetMotionMaster()->Clear(false);
+                me->GetMotionMaster()->MoveFollow(totem, PET_FOLLOW_DIST, me->GetFollowAngle(), MOTION_SLOT_ACTIVE);
+            }
+    }
 
     void UpdateAI(uint32 diff) override
     {
@@ -89,7 +111,10 @@ struct npc_pet_shaman_fire_elemental : public ScriptedAI
 {
     npc_pet_shaman_fire_elemental(Creature* creature) : ScriptedAI(creature), _initAttack(true) { }
 
-    void InitializeAI() override { }
+    void InitializeAI() override
+    {
+        MoveToTotem();
+    }
 
     void JustEngagedWith(Unit*) override
     {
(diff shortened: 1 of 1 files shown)


heyitsbench:
This is a hack (as the linked PR mentioned, the totem was confirmed to be the creator of the elemental), and I'm not fond of the "find nearest totem" solution, as obviously if another Shaman creates a totem, the elemental could easily decide to follow that one instead.

xDevICCI:
I recently pushed commit, updated to verify same owner before following:
```cpp
if (totem->GetOwnerGUID() == me->GetOwnerGUID())
```

This ensures the elemental only follows its own shaman's totem.

Nyeriah:
Now if the first totem it finds isn't owned by the player, it won't do anything

xDevICCI:
Fixed. Please review last commit pushed.

Now iterates all totems in range and follows the one with matching owner using `GetCreaturesWithEntryInRange()`.

src/server/scripts/Pet/pet_shaman.cpp
```diff
@@ -46,13 +53,37 @@ struct npc_pet_shaman_earth_elemental : public ScriptedAI
 {
     npc_pet_shaman_earth_elemental(Creature* creature) : ScriptedAI(creature), _initAttack(true) { }
 
+    void InitializeAI() override
+    {
+        MoveToTotem();
```
blinkysc, line 52: Probably could move to first update tick using _initAttack so follow doesn't override

src/server/scripts/Pet/pet_shaman.cpp
```diff
@@ -46,13 +53,37 @@ struct npc_pet_shaman_earth_elemental : public ScriptedAI
...
+        if (!_EnterEvadeMode(why))
+            return;
+
+        MoveToTotem();
+    }
+
+    void MoveToTotem()
+    {
+        std::list<Creature*> totemList;
+        me->GetCreaturesWithEntryInRange(totemList, 100.0f, NPC_EARTH_ELEMENTAL_TOTEM);
```
blinkysc: Instead of searching for creatures in range, you can get the totem directly from the owner's summon slot - it's an O(1) GUID lookup instead of an O(n) range search:

Something like

  if (Unit* owner = me->GetCharmerOrOwner())
      if (Creature* totem = ObjectAccessor::GetCreature(*me, owner->m_SummonSlot[SUMMON_SLOT_TOTEM_FIRE]))
          me->GetMotionMaster()->MoveFollow(totem, PET_FOLLOW_DIST, PET_FOLLOW_ANGLE);

blinkysc:
Might be worth adding more directives for your system prompt for claude if you're overly reliant on it. In addition I assume you're using the clang LSP with /plugin, if not, I recommend. In addition https://github.com/blinkysc/azerothMCP might help too.

Claude is very good, but still have to make sure it's not looking for weird or inefficient solutions. Be skeptical of it's answers

src/server/scripts/Pet/pet_shaman.cpp
Nyeriah: That is a much better approach

sudlud:
Also no need for force pushing if possible please, it makes tracking reviews and following changes much easier. All the commits get squashed into a single commit on merge anyways.

sudlud:
Thanks for taking this on, @xDevICCI, and for the writeup tracing the #19599 / #19658 / #19664 history.

I'm going to close this one, though. As @heyitsbench and @Nyeriah noted, the totem is confirmed to be the elemental's actual owner on retail, so the Blizzlike fix is to model that ownership correctly rather than keep the player as owner and patch follow behavior. "Find the nearest totem and follow it" is a workaround for the symptom: fragile around multiple shamans/totems, and it leaves the underlying model wrong. The real blocker is the one #19599 hit, where many pet checks (faction especially) assume a player owner.

This has also been inactive ~5 months with the core review concerns unaddressed. I'll leave #19514 open so it can be tackled along these lines. Don't let it discourage you, the investigation was solid, and a PR fixing the ownership/faction handling would be very welcome. Thanks again!

PR #24568, opened 2026-01-31
refactor: [SUGGESTION] Changed SpellMgr::instance to return a hard reference

SmashingQuasar:
## Changes Proposed:

This PR proposes changes to:

> [!IMPORTANT]
> This PR is a suggestion to introduce some changes to the way singletons are being managed in the Azerothcore source code.
> The goal is to open a discussion about this.

I am suggesting to change the way singleton instances are being returned.
In the current code, singleton instances are passed as a pointer, but stored as a value. I know that a long time ago, some developers used a trick by creating leaky singletons on purpose to avoid shutdown issues. I don't think this trick applies here because the pointer is not stored, only returned.
I think going for a reference is semantically more correct, and more importantly, it makes it easier to understand for contributors who are new to the code base.
It also has the benefit to remove raw pointers, which is never a loss.

I understand if this is not something the team is willing to do right now. It is obviously fine to reject this PR if it does not meet the current objectives of the project.
I tried to respect the commit message guidelines as best as I could. I simply added a `[SUGGESTION]` tag in uppercase to make the purpose of this PR very clear and avoid creating unnecessary friction.

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

> [!NOTE]
> This PR was **NOT** generated using any AI tool.

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/AI/CoreAI/CombatAI.cpp
+++ b/src/server/game/AI/CoreAI/CombatAI.cpp
@@ -50,7 +50,7 @@ void AggressorAI::UpdateAI(uint32 /*diff*/)
 void CombatAI::InitializeAI()
 {
     for (uint32 i = 0; i < MAX_CREATURE_SPELLS; ++i)
-        if (me->m_spells[i] && sSpellMgr->GetSpellInfo(me->m_spells[i]))
+        if (me->m_spells[i] && sSpellMgr.GetSpellInfo(me->m_spells[i]))
             spells.push_back(me->m_spells[i]);
 
     CreatureAI::InitializeAI();
@@ -181,7 +181,7 @@ ArcherAI::ArcherAI(Creature* c) : CreatureAI(c)
     if (!me->m_spells[0])
         LOG_ERROR("entities.unit.ai", "ArcherAI set for creature (entry = {}) with spell1=0. AI will do nothing", me->GetEntry());
 
-    SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(me->m_spells[0]);
+    SpellInfo const* spellInfo = sSpellMgr.GetSpellInfo(me->m_spells[0]);
     m_minRange = spellInfo ? spellInfo->GetMinRange(false) : 0;
 
     if (!m_minRange)
@@ -230,7 +230,7 @@ TurretAI::TurretAI(Creature* c) : CreatureAI(c)
     if (!me->m_spells[0])
         LOG_ERROR("entities.unit.ai", "TurretAI set for creature (entry = {}) with spell1=0. AI will do nothing", me->GetEntry());
 
-    SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(me->m_spells[0]);
+    SpellInfo const* spellInfo = sSpellMgr.GetSpellInfo(me->m_spells[0]);
     m_minRange = spellInfo ? spellInfo->GetMinRange(false) : 0;
     me->m_CombatDistance = spellInfo ? spellInfo->GetMaxRange(false) : 0;
     me->m_SightDistance = me->m_CombatDistance;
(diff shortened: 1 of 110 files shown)


Takenbacon:
Not that my opinion matters much but I do see what you're going for however I think that ship had sailed long ago to make such a refactor. Nearly all singleton "manager" classes return a pointer, modifying SpellMgr alone and leaving the dozens of others the same would be even more problematic and confusing. To change all of them would be significant and cause problems with most, if not all modules and custom changes for no added benefit. I don't think it's worth it.

SmashingQuasar:
> Not that my opinion matters much but I do see what you're going for however I think that ship had sailed long ago to make such a refactor. Nearly all singleton "manager" classes return a pointer, modifying SpellMgr alone and leaving the dozens of others the same would be even more problematic and confusing. To change all of them would be significant and cause problems with most, if not all modules and custom changes for no added benefit. I don't think it's worth it.

I only changed one here to open the discussion, it would be quite a lot of work to change them all, even with a replace all, and it was not necessary for the discussion.

It would indeed break modules, but AC breaks module on a regular basis. Many modules need to be updated to the new core when changes are being made, even changes that are not impacting the core behaviour. For example the naming change on hooks (`PlayerScript::OnUpdate => PlayerScript::OnPlayerUpdate`) was purely a cosmetic change, yet it broke all modules using those scripts.

I am suggesting this change because it increases cognitive load and breaks a fundamental invariant of Meyer's singletons. I am not hellbent on it, but it would be nice to do.
I understand if people are unwilling to invest any energy on this.

sudlud:
Thanks for taking the time to put this together, and especially for framing it so clearly as a suggestion to open discussion. That's appreciated.

Speaking as one of the maintainers (so this is my view, and others on the team may see it differently), the reasoning here is sound on its own terms. A reference is semantically more honest for a Meyers singleton, and we don't rely on the leaky-singleton trick on the return path. So this isn't a "you're wrong" rejection.

The blocker for me is ownership. A change like this only makes sense if it's carried through to all the manager singletons consistently. Doing SpellMgr alone, as you noted yourself, leaves the codebase in a more confusing half-and-half state than it's in today. That means someone on the maintainer side has to own the full sweep: drive it to completion, keep it rebased against a fast-moving core, handle the inevitable module breakage, and review the whole thing. Right now none of us has the bandwidth to take that on. Content and script fixes plus the existing review backlog are simply more demanding priorities, and a large cosmetic-but-invasive refactor with no behavioural payoff isn't something I can responsibly commit to seeing through at the moment.

To be clear about what "owning" means here, it's more than being willing to do the work, which I don't doubt you are. A sweep like this needs someone who can also carry it long-term on the maintainer side: merge authority over the full conversion, staying on top of the module fallout as it surfaces over the following weeks, and being the point of contact when something downstream breaks because of it. That's a level of standing commitment that really sits with the established maintainer group, and it's built up over time through smaller contributions and review history rather than a single large PR. That's not a knock on you at all, it's just where the bar naturally sits for a change this invasive. If this is an area you're interested in, the best path is to keep contributing on smaller, self-contained fixes, and that trust and familiarity builds naturally from there.

So I'm going to close this. Not because the idea lacks merit, but because we can't currently give it the ownership it would need to land properly. If priorities free up and someone's willing to champion the full conversion, this is a reasonable starting point to revisit. Thanks again for the contribution, and for the considerate tone.

SmashingQuasar:
Thanks for the detailed and respectful answer. It's very appreciated.
Honestly I did expect this to be closed for these reasons. I completely understand the reasoning.

There are many - smaller - things that I believed could be improved. Right now I am working on a forked version of `mod-playerbots` to attempt to improve the code base and modernise it. As you imagine it's quite a large commitment so it would be unreasonable to me to try to commit to regular contributions to this repository as well.

Again, great answer that makes a lot sense. Thanks!

PR #24926, opened 2026-02-27
fix(Core/Spells): Fix item use macro breaking melee auto-attack

blinkysc:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Claude was used for Ghidra reverse engineering analysis of the WoW 3.3.5a client to understand the root cause.

## Description

When a player uses an item spell (e.g. healing potion) via macro while the spell is on cooldown, `CastItemUseSpell` silently skips the cast without notifying the client. The client never receives `SMSG_CAST_FAILED`, so it orphans the pending spell cast state, which permanently blocks melee auto-attack until the player manually casts another spell.

### Root Cause

AzerothCore's `CastItemUseSpell` has a `HasSpellCooldown` optimization that `continue`s without sending any packet to the client. TrinityCore doesn't have this optimization — the spell always goes through `prepare()` → `CheckCast()` → `SendCastResult()`, which naturally sends `SMSG_CAST_FAILED` and allows the client to clean up its pending cast state.

### Fix

Send `SMSG_CAST_FAILED` with `SPELL_FAILED_NOT_READY` when skipping cooldown-blocked item spells

## SOURCE:

- [x] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). Original TrinityCore behavior by QAston credited via Co-Authored-By.

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Equip a melee weapon and engage a target dummy with auto-attack
2. Create a macro: `/use Super Healing Potion`
3. Spam the macro twice quickly (second use hits cooldown)
4. Verify melee auto-attack continues swinging without interruption

- None

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -7459,6 +7459,15 @@ void Player::CastItemUseSpell(Item* item, SpellCastTargets const& targets, uint8
 
         if (HasSpellCooldown(spellInfo->Id))
         {
+            // Client-side bug: silently skipping here orphans a pending spell cast that
+            // permanently blocks melee auto-attack. The client's SMSG_CAST_FAILED handler
+            // won't clean it up (broken lookup), so send a no-op SMSG_SPELL_GO instead.
+            if (cast_count
+                && (spellInfo->InterruptFlags
+                    & SPELL_INTERRUPT_FLAG_INTERRUPT))
+                Spell::SendEmptySpellGo(item->GetGUID(),
+                    GetGUID(), cast_count,
+                    spellInfo->Id, this);
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


src/server/game/Entities/Player/Player.cpp
```diff
@@ -7459,6 +7459,30 @@ void Player::CastItemUseSpell(Item* item, SpellCastTargets const& targets, uint8
 
         if (HasSpellCooldown(spellInfo->Id))
         {
+            // Client-side bug: silently skipping here leaves
+            // an orphaned pending spell cast that permanently
+            // convinces the client you're still casting,
+            // which blocks melee auto-attack forever. The
+            // client won't clean it up on its own (its
+            // SMSG_CAST_FAILED handler is broken), so we
+            // send a no-op SMSG_SPELL_GO to do the job.
```
heyitsbench: Not honestly sure the general opinion, but I would prefer these on less rows and using more columns. This looks formatted for a vertical screen. 😆

blinkysc: Haha ok

Gultask (approved):

PR #24989, opened 2026-03-03
fix(DB/SmartAI): Fix quest 11280 Draconis Gastritis completion

Pokeuu:
## Changes Proposed:

Fix the SmartAI script for Proto-Drake (NPC 23689) used in quest 11280 "Draconis Gastritis".

**Problem:** When the player uses Tillinghast's Plagued Meat, the Proto-Drake attacks the player on sight instead of following the Draconis Gastritis Bunny (NPC 24170) to eat the meat. The Out-of-Combat follow event never fires because the drake enters combat.

**Fix:** Added `Set React State Passive` when the Proto-Drake detects the quest bunny. This prevents aggro during the quest sequence. Also added an eating emote for visual feedback.

### AI-assisted Pull Requests
- [x] AI tools were used partially in preparing this pull request. **Tool: Google Antigravity (Gemini)**. I understand the changes and can explain them.

## Issues Addressed:
- Quest 11280 "Draconis Gastritis" cannot be completed

## SOURCE:
- [x] Video evidence, knowledge databases or other public sources
  - [Wowhead Quest 11280](https://www.wowhead.com/wotlk/quest=11280/draconis-gastritis)

## Tests Performed:

SAI logic reviewed: the Proto-Drake creature_template confirms hostile faction. The existing SAI had no mechanism to suppress aggro during the quest sequence. Fix uses Set React State Passive, the standard SmartAI approach for scripted sequences.

## How to Test the Changes:
1. Log in as Horde character level 68+
2. Pick up quest "Draconis Gastritis" from Plaguebringer Tillinghast in New Agamand
3. Go to Ember Clutch and use Tillinghast's Plagued Meat
4. Verify Proto-Drake flies down, eats the meat, gets sick, grants quest credit
5. Verify Proto-Drakes not involved in quest still attack normally

- [ ] In-game testing needed to confirm the fix


data/sql/updates/pending_db_world/fix_quest_11280_draconis_gastritis.sql
```diff
@@ -0,0 +1,40 @@
...
+(23689, 0, 4, 5, 61, 0, 100, 512, 0, 0, 0, 0, 0, 0, 51, 0, 0, 0, 0, 0, 0, 19, 24170, 10, 0, 0, 0, 0, 0, 0, 'Proto-Drake - On Follow Complete - Kill Closest ''Draconis Gastritis Bunny'''),
+-- Move to phase 1 so the despawn timer starts
+(23689, 0, 5, 0, 61, 0, 100, 512, 0, 0, 0, 0, 0, 0, 22, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Proto-Drake - On Follow Complete - Set Event Phase 1'),
+-- Despawn the drake 45 seconds after quest completion to clean up
+(23689, 0, 6, 0, 1, 1, 100, 512, 45000, 45000, 45000, 45000, 0, 0, 41, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Proto-Drake - Out of Combat (Phase 1) - Despawn Instant'),
+-- Separate quest: Malister Frost Wand (spell 40969) also attracts the Proto-Drake to the caster
+(23689, 0, 8, 0, 8, 0, 100, 0, 40969, 0, 120000, 120000, 0, 0, 69, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 0, 0, 'Proto-Drake - On Spellhit ''Malister''s Frost Wand'' - Move To Invoker'),
+-- Combat abilities below only trigger when not in quest mode (normal combat behavior)
+(23689, 0, 9, 0, 9, 0, 100, 513, 0, 0, 0, 0, 0, 20, 101, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Proto-Drake - Within 0-20 Range - Set Home Position (No Repeat)'),
```
Gultask, line 37: Why is Set Home Position needed?

data/sql/updates/pending_db_world/fix_quest_11280_draconis_gastritis.sql
```diff
@@ -0,0 +1,40 @@
...
+DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 23689);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+-- Quest sequence: When the quest bunny is nearby, drake goes passive and follows it
+(23689, 0, 0, 1, 1, 0, 100, 512, 10000, 10000, 10000, 10000, 0, 0, 29, 0, 0, 24170, 0, 0, 0, 19, 24170, 75, 0, 0, 0, 0, 0, 0, 'Proto-Drake - Out of Combat - Start Follow Closest Creature ''Draconis Gastritis Bunny'''),
+-- Set react state to passive so the drake does not attack the player during the quest sequence
+(23689, 0, 1, 0, 61, 0, 100, 512, 0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Proto-Drake - On Link - Set React State Passive'),
```
Gultask, line 23: Ok, from what I'm seeing in some videos (there are so many private server videos omg), it seems more like they're using Immune to All (Add unit flags 768)

I'd use that otherwise player can fight the drakes and either kill them or make them enter some evade mode. Just remember to remove the unit flags when they respawn, otherwise it won't clear

data/sql/updates/pending_db_world/fix_quest_11280_draconis_gastritis.sql
```diff
@@ -0,0 +1,40 @@
...
+
+-- ============================================================================
+-- Proto-Drake (NPC 23689) - SmartAI fix for quest "Draconis Gastritis" (11280)
+-- ============================================================================
+DELETE FROM `smart_scripts` WHERE (`source_type` = 0 AND `entryorguid` = 23689);
+INSERT INTO `smart_scripts` (`entryorguid`, `source_type`, `id`, `link`, `event_type`, `event_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `event_param5`, `event_param6`, `action_type`, `action_param1`, `action_param2`, `action_param3`, `action_param4`, `action_param5`, `action_param6`, `target_type`, `target_param1`, `target_param2`, `target_param3`, `target_param4`, `target_x`, `target_y`, `target_z`, `target_o`, `comment`) VALUES
+-- Quest sequence: When the quest bunny is nearby, drake goes passive and follows it
+(23689, 0, 0, 1, 1, 0, 100, 512, 10000, 10000, 10000, 10000, 0, 0, 29, 0, 0, 24170, 0, 0, 0, 19, 24170, 75, 0, 0, 0, 0, 0, 0, 'Proto-Drake - Out of Combat - Start Follow Closest Creature ''Draconis Gastritis Bunny'''),
```
Gultask: Doing this may cause more than one to follow the bunny. I'd do it differently, make the bunny or the object DO ACTION to the nearest drake, then have the Drake on Event ACTION_DONE to go to the nearest Bunny. That way the scan only happens once instead of every 10s

Gultask: Also, might want to add a fallback to the bunny (if there isn't one already) where it despawns in a timer. Cause if there's no dragons nearby to listen to the call it might stay there forever

data/sql/updates/pending_db_world/fix_quest_11280_draconis_gastritis.sql
Gultask: If they go Immune when the quest scripting starts, it shouldn't be needed. Cause I'm assuming it's here in case the player attacks while it's passive and it stops following the bunny?

Gultask:
Also, aside from the comments, I just changed '' to \' to match our standards

sudlud:
Hi @Pokeuu, thank you for taking the time to dig into this quest and put together a fix with clear notes and testing steps. It's appreciated.

After reviewing, I'm going to close this for now, for a couple of reasons:

- Master already has a working SAI for this quest (merged in `2025_09_04_03.sql`), covering both the Proto-Drake (23689) and the Draconis Gastritis Bunny (24170) follow / kill / quest-credit chain, plus conditions in `2025_11_08_01.sql`. So the effective delta here is mainly the passive react state and the eating emote on top of what's already in place.

- The core review feedback is still open. As Copilot noted, setting passive via a link off the out-of-combat follow event is circular: the reported symptom is the drake aggroing the player (in combat), so the OOC event never fires and the passive state is never applied. Gultask's suggestion to use immune-to-all unit flags (768), set when the sequence starts and removed on respawn, addresses that directly. His points about avoiding multiple drakes following one bunny (event-driven trigger instead of a 10s scan) and adding a fallback despawn on the bunny are also still unaddressed.

Please feel free to reopen this anytime in the future once it's updated for the latest master and addresses the review comments. Thanks again for contributing.

PR #24991, opened 2026-03-03
chore(Core/Spell): add null-checks

Helias:
## Changes Proposed:
Add some null-checks potentially to prevent crashes


Nyeriah:
I have to be honest: to this day, I’ve yet to see AI successfully land a proper crash fix. More often than not, when someone prompts it to “make the code safer”, it results in a flurry of redundant checks. In many cases, those paths are already guarded elsewhere or are simply impossible to hit, something the AI fails to grasp.

The outcome is usually the same: it becomes overzealous and introduces redundant or dead code rather than addressing the underlying issue properly. This can be merged, but ultimately I don't personally believe it does anything one way or another, were it a risk we would have noticed it sooner (10+ years old code after all)

Having looked into this, for caster to be null during spell preparation it would have to occur within milliseconds of the cast itself, realistically, a one-in-a-billion scenario.

src/server/game/Spells/Spell.cpp
```diff
@@ -8137,8 +8137,9 @@ SpellEvent::~SpellEvent()
     }
     else
     {
+        Unit* caster = m_Spell->GetCaster();
         LOG_ERROR("spells", "~SpellEvent: {} {} tried to delete non-deletable spell {}. Was not deleted, causes memory leak.",
-                       (m_Spell->GetCaster()->IsPlayer() ? "Player" : "Creature"), m_Spell->GetCaster()->GetGUID().ToString(), m_Spell->m_spellInfo->Id);
```
Nyeriah, line 8141: no, very redundant, see the ABORT() right underneath this, the server already crashed at this point, for other reasons

Helias:
Then I think we can avoid to merge tbh, it was more a try of adding some safer null-checks, but I get that we are already covering this good enough

Takenbacon:
> I have to be honest: to this day, I’ve yet to see AI successfully land a proper crash fix. More often than not, when someone prompts it to “make the code safer”, it results in a flurry of redundant checks. In many cases, those paths are already guarded elsewhere or are simply impossible to hit, something the AI fails to grasp.
> 
> The outcome is usually the same: it becomes overzealous and introduces redundant or dead code rather than addressing the underlying issue properly. This can be merged, but ultimately I don't personally believe it does anything one way or another, were it a risk we would have noticed it sooner (10+ years old code after all)
> 
> Having looked into this, for caster to be null during spell preparation it would have to occur within milliseconds of the cast itself, realistically, a one-in-a-billion scenario.

From what I've been able to tell so far with AI giving it a problem to solve in a large codebase like AC just doesn't work often, or at least correctly. It doesn't have the computational power or context to have a full understanding of what is going on at a higher level. I think AI is actually great for something like a rough peer-review of PR changes to check for simple human mistakes but yeah, definitely not something anyone should realistically be using to say "fix this crash".

Helias:
the crashlog usually are not so big and an AI can easily understand if there was a missing nullpointer that needs to be checked, mostly of the crashes that I have prevented in the past had this as reason, and they were easy to fix, so I gave it a try.

Issue #22498, opened 2025-07-16
[Ulduar][Thorim] Chain Lightning hits targets beyond intended 8 yard range


blacknightt1:
### Current Behaviour

During the Thorim encounter in Ulduar, the spell Chain Lightning (25-man) (SPELL_CHAIN_LIGHTNING_25) incorrectly jumps to players who are more than 8 yards away from each other. This causes the spell to chain further than it should.

Expected behavior:
Chain Lightning should only jump to players within 8 yards of each other. Any player beyond 8 yards should not be affected by the chain.

How to reproduce:

Enter Ulduar on 25-man difficulty.

Reach Thorim and let him cast Chain Lightning.

Spread players >8 yards apart.

Notice that the chain still hits players beyond the 8-yard intended range.

Spell ID:
SPELL_CHAIN_LIGHTNING_25 – (check definition in boss_thorim.cpp)

![Image](https://github.com/user-attachments/assets/64465fa6-ab1a-4be9-b391-cabfa6da3006)
![Image](https://github.com/user-attachments/assets/438bc92f-1b52-4ec1-aafa-b06a42a939ef)
![Image](https://github.com/user-attachments/assets/8904433d-29ad-4a5c-8163-746453e92424)

### Expected Behaviour

The spell targeting logic likely does not correctly filter by distance. A distance check (<= 8.0f) should be enforced when determining valid jump targets for the chain.

### Source

none

### Steps to reproduce the problem

none

### Extra Notes

What we tested does not share damage at a distance of 13 yards, and this should be at a distance of 8 yards or less.

### AC rev. hash/commit

AzerothCore rev. 9f5b56ac86af+ 2025-07-16 10:22:00 +0200 (master branch)

### Operating system

Windows 11 x64

### Custom changes or Modules

none

sogladev:
confirmed. The jumpRadius is set to the default 10.0f, which is too large in this case

- https://github.com/azerothcore/azerothcore-wotlk/blob/d40be8c1d4a4d99ae14be8c4211b84681c8c554f/src/server/game/Spells/Spell.cpp#L2208

blacknightt1:
I tested a change to Spell.cpp (line 2208) to fix the incorrect jump distance for chain spells like Chain Lightning and Chain Heal.
Previously, the default jump radius for magic chain spells was set to 10 yards, which caused the spell to sometimes jump too far or target unintended units.

I changed the code from:



// 10y as default for magic chain spells
else
[...]

PR #25023, opened 2026-03-07
fix(Scripts/Spells): Implement spell_jump_distance

Nyeriah:
This pull request introduces support for per-spell chain jump distance overrides, allowing certain spells to have custom jump distances defined in the database. The main changes involve adding a new table for these overrides, updating the spell loading and initialization logic, and applying the custom jump distance in spell mechanics.

Database and data loading:

* Added a new `spell_jump_distance` table to the database to store per-spell jump distance overrides, and populated it with initial values for spell IDs 62131 and 64390.
* Implemented the `SpellMgr::LoadSpellJumpDistances()` method to load custom jump distances from the database and assign them to the corresponding `SpellInfo` objects during server startup.
* Registered the loading of spell jump distances in `World::SetInitialWorldSettings()` to ensure these values are loaded at startup.

Spell system enhancements:

* Added a `JumpDistance` field to the `SpellInfo` class, initialized to `0.0f`, to store the custom jump distance. [[1]](diffhunk://#diff-6ccfbebc48a19c053095eed2f566f58aa3afb0bd31843e35fec42873792b8140R404) [[2]](diffhunk://#diff-b85d08735de1a07454ea9d244c37ce7ac6ece09993b475d6ba455237606488cfR864)
* Modified `Spell::SearchChainTargets()` to use the per-spell jump distance override if set, affecting how far chain spells can jump between targets.

These changes collectively enable more flexible tuning of chain spell mechanics via database configuration.

Thorim's Chain Lightning chains to targets up to 10y apart due to hardcoded default jump radius for magic chain spells. Add a reusable spell script that trims chain hop targets exceeding a configurable distance. Applied to spells 62131/64390 with a 5y limit.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1772885471981.sql
+++ b/data/sql/updates/pending_db_world/rev_1772885471981.sql
@@ -1,0 +1,11 @@
+--
+CREATE TABLE IF NOT EXISTS `spell_jump_distance` (
+  `ID` int unsigned NOT NULL COMMENT 'spell id',
+  `JumpDistance` float NOT NULL DEFAULT '0' COMMENT 'max hop distance in yards',
+  PRIMARY KEY (`ID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci COMMENT='Per-spell chain jump distance override';
+
+DELETE FROM `spell_jump_distance` WHERE `ID` IN (62131, 64390);
+INSERT INTO `spell_jump_distance` (`ID`, `JumpDistance`) VALUES
+(62131, 5.0),
+(64390, 5.0);
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -2126,6 +2126,10 @@ void Spell::SearchChainTargets(std::list<WorldObject*>& targets, uint32 chainTar
             break;
     }
 
+    // per-spell override from spell_jump_distance table
+    if (m_spellInfo->JumpDistance > 0.0f)
+        jumpRadius = m_spellInfo->JumpDistance;
+
     // chain lightning/heal spells and similar - allow to jump at larger distance and go out of los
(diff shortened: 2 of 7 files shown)


sogladev:
Does hard-coding values make more sense than a spellscript just for Thorim? This could be a `spell_jump_radius` table too

See my comment in: 
https://github.com/azerothcore/azerothcore-wotlk/issues/22498#issuecomment-3113738470

issue:
- https://github.com/azerothcore/azerothcore-wotlk/issues/22498

Nyeriah:
> Does hard-coding values make more sense than a spellscript just for Thorim? This could be a spell_jump_radius table too

I've considered it, but does it make sense for something used by at best 5 spells? I'm not sure
Spellscripts are better than hardcoded values imo.

sogladev:
These are blizzlike spells, so it makes more sense to hard-code. You're achieving the same behavior with a spellscript, but patching the selection instead of fixing the initial selection. An override table is overkill for now, but I was thinking more of modding. A spellscript like this won't allow for a larger radius?

Nyeriah:
Might be a matter of personal opinion but hardcoded values feel very odd to me and I think we should avoid them, random ids here and there. This is a generic spell so it just need the params changed @ the register so it can be used with other spells/radius

Nyeriah:
We can move it to a table if you want. I just feel it to be very much an overkill considering then we have to handle loading, storage, everything else for what might have very few spells

sogladev:
A table isn't required; it's just an alternative to hard-coding. I looked through the list from cmangos-wotlk, and Spellscripts like this wouldn't work for distance increases. I can't say if these are correct for 335. 

335 DBC `EffectChainTarget = 5`  with override:
```cpp
45664: // Legion Lightning - KJ
            return 18.f;
```

A switch + porting these overrides may be the way to go for this issue?

Gultask (approved):

sogladev (approved):

PR #25087, opened 2026-03-14
fix(Core/Unit): Fix infinite loop in RemoveAllControlled

Shard-MW:
The iterator was not incremented when a TempSummon was not unsummoned, causing an infinite loop.

Introduced in fda5093e5.

## Changes Proposed:

This PR proposes changes to:

In `Unit::RemoveAllControlled`, when iterating over `m_Controlled`, the branch handling `TempSummon` that are guardians during `onDeath` (where `!(onDeath && !IsPlayer() && target->IsGuardian())` is false) did not increment the iterator, causing the loop to spin forever on the same element.

Added the missing `else ++it;` to advance the iterator when the summon is kept alive.

This is purely a code logic fix — every other branch in the loop either calls `it = m_Controlled.erase(it)` or `++it`, this one was simply missing. No in-game testing is strictly necessary to validate it, but steps are provided below for completeness.

### AI-assisted Pull Requests

## Issues Addressed:

- Closes fda5093e5 regression (infinite loop in `RemoveAllControlled` when a guardian TempSummon is not unsummoned on owner death)

## SOURCE:

Code review of the iterator logic in the while loop. The bug is apparent from reading the code: every other branch either calls `it = m_Controlled.erase(it)` or `++it`, except this one.

## Tests Performed:

This PR has been:

## How to Test the Changes:

I recommend starting the server with waypoints and creature activity enabled to stress test the summon/guardian paths, but given this is a straightforward iterator logic fix, simple reproduction should suffice:

1. Have a player with a guardian-type summon active (e.g. a Warlock pet flagged as guardian).
2. Kill the player (`.kill` GM command).
3. Verify the server does not hang/freeze in `RemoveAllControlled`.
4. Also test with a Lightwell active on owner death to ensure it still stays alive as intended by fda5093e5.

- [ ] Regression test with all summon types (Lightwell, guardians, totems, standard pets) on owner death to ensure no other iterator issue exists.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -8102,6 +8102,8 @@ void Unit::RemoveAllControlled(bool onDeath /*= false*/)
                 target->ToTempSummon()->UnSummon();
                 it = m_Controlled.erase(it);
             }
+            else
+                ++it;
         }
         else
         {


brighton-chi:
This fixed the crash where killing the summoner before the guardian caused an infinite loop, but now I get a crash if the guardian is killed before the summoner. I added some logs to debug, and it appears that UnSummon() is being called on dead player guardians guardians--for example, if I kill a Spirit Wolf, the server crashes, and the last log is in this block:

```
  if (!(onDeath && !IsPlayer() && target->IsGuardian()))
  {
      target->ToTempSummon()->UnSummon();
      it = m_Controlled.erase(it);
  }
```

Separately, it seems dead NPC guardians are being repeatedly processed repeatedly in the else block following the above if block--should they be erased from m_Controlled also?

LannyE:
I can confirm this crash is happening.  On my DEV server it crashes about every 4 hours.  While commit #24675 caused the server to freeze.  This commit #25087 fixes the freeze and replaces it with a crash.  I had AI analyze my Call Stack info and it led me to this commit by its answer.  Here is part of the analysis response:
```
"This call stack is incredibly helpful. It tells a very specific story: the server didn't crash because of a random Windows error; it crashed while a Creature using SmartAI was trying to die and clean up its "controlled" units (like summons, pets, or possessed targets).

The "Smoking Gun"
The crash happens at Unit::RemoveAllControlled. Specifically, it’s failing inside a std::set (a "Tree" structure in C++) while trying to deallocate memory.

C++
worldserver.exe!Unit::RemoveAllControlled(bool onDeath) Line 8522"
```

Here is a Github Gist of "The Fix" section after feeding the AI the commits, code, and changes made.  It also answers @brighton-chi's question:

https://gist.github.com/LannyE/bd5a2544b3e1111e10e5c51f3b016008

Hopefully this helps.

PR #21652, opened 2025-03-04
fix(scripts/invisible): boss Captain Skarloc

grimgravy:
## Changes Proposed:

This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [x] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/17146

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #25122, opened 2026-03-19
fix(scripts/OldHilldBradFootHills): fix Captain Skarloc reacting to stealthed players"

grimgravy:
Reverts azerothcore/azerothcore-wotlk#21652
I think this fixes the MoveInLineOfSight

### DIFF
--- a/src/server/game/Grids/Notifiers/GridNotifiers.cpp
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.cpp
@@ -132,7 +132,7 @@ inline void CreatureUnitRelocationWorker(Creature* c, Unit* u)
     }
 
     // Prevents AI reaction when immunity flags are active
-    if (c->GetEntry() == 17862 && c->IsImmuneToNPC())
+    if (c->GetZoneId() == 2367 && c->IsImmuneToNPC())
     {
         return;
     }


src/server/game/Grids/Notifiers/GridNotifiers.cpp
```diff
@@ -131,6 +131,12 @@ inline void CreatureUnitRelocationWorker(Creature* c, Unit* u)
         return;
     }
 
+    // Prevents AI reaction when immunity flags are active
+    if (c->GetEntry() == 17862 && c->IsImmuneToNPC())
```
blinkysc: this is a hack with the magic number for the npc. Maybe there is a better way to handle this?

sudlud:
Please update the title, the linked pr has already been reverted

src/server/game/Grids/Notifiers/GridNotifiers.cpp
TheSCREWEDSoftware: > if (c->GetEntry() == 17862 && c->IsImmuneToNPC())

This would be properly betterr added within either a script of the creature for that event / instance or smart scripts

grimgravy: @TheSCREWEDSoftware, That line of code is outdated

src/server/game/Grids/Notifiers/GridNotifiers.cpp
```diff
@@ -131,6 +131,12 @@ inline void CreatureUnitRelocationWorker(Creature* c, Unit* u)
         return;
     }
 
+    // Prevents AI reaction when immunity flags are active
+    if (c->GetZoneId() == 2367 && c->IsImmuneToNPC())
+    {
+        return;
+    }
```
sudlud, line 138: No braces in single line if

sudlud: Also if we should accept the magic number here, there must be a comment above to explain the reason for the line.
Also there are enum values available for all zones / areas that must be used.

kissingers:
Why only Zone 2367 need check?

src/server/game/Grids/Notifiers/GridNotifiers.cpp
grimgravy: > No braces in single line if

This is to prevent a possible collision, should it occur

Nyeriah:
Zone hard code is not an acceptable solution, this is a widespread issue also present in other places e.g violet hold

PR #25161, opened 2026-03-21
fix(Core/XP): Reduce XP when ungrouped high-level player helps kill

### LINKED ISSUE #24191: [Core] - Experience doesn't get properly reduced when a higher level player kills a lower level player's mob
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/7725

Triage Notes: Currently in AC it seems we have the Experience gain for non-grouped members (where a high level kills a mob that a low level, still grants the low level full experience, this would be true for 2.4.0 or 2.4.1 not for anything 2.4.2 and onwards

People acknowledge this not working on 2.4.2: https://www.youtube.com/watch?v=5-cDnRJt9i4 (and you can see how is done there):

> The experience amounts given above assume that the solo character and their pets have done 100% of the damage to kill the mob. Experience will be reduced when:

> Someone else (ungrouped) helps to damage the mob. The XP you receive depends on if that someone else will receive xp for killing that mob. If yes, you get full XP. If not, you get a tiny fraction. It doesn't matter how many other ungrouped help damage nor how much damage you done.
Example: You're 6th level fighting a 6th level mob. If the ungrouped helper is level 10 or lower damage the mob, you will get full XP because a character level 10 or lower can receive XP from a 6th level mob. If the ungrouped helper is level 11 or higher, you get a tiny fraction of the XP.

> Someone else (ungrouped) helps you by healing you in mid fight. This will cost you only a few XP points per heal.
Your damage shields do significant damage to the mob. This is usually a small loss, 5-10 XP points at worst.
The mob's maximum health is reduced below its current health
The effect of this is that power-leveling low levels is much less effective than at higher levels. It is faster for low levels to kill without help, than it is to powerlevel them. It also makes it hard to determine the solo XP for high level mobs, since you won't get an accurate number unless the character does 100% of the damage.

> The way to power-level is to be near the character level you're power-leveling. Thus, a 49th character will be much more effective than a 70th character in power-leveling a 40th level character. More effective would be multiple 49th characters power-leveling the 40th level character.

From: https://wowpedia.fandom.com/wiki/Mob_experience#Solo_Experience_Modifiers

How I did my testing, tried 3 but didnt make a difference so i went back to 2 same results, so we will use 2 for the steps.
- Make 2 alliance characters
- One of them `.level 7` to be level 8 and other `.level 79` to be level 80, i recommend making the low level a mage because frost armour always tags the mob (if it should, that's not topic for this issue)
- `.cheat god` for both
- Get in a group with both characters and both `.tele rfc`
- If you did a mage for the low level, use frost armour
- Kill a level 13 Earthborer with your level 80, level 8 character gains 12 XP, correct behaviour
- Leave the group, attack a level 13 earthborer with your level 8 (or let it hit your frost armour) and kill it with high level, your level 8 just gained 255 experience, not the intended behaviour, it should give you a fraction of XP not full XP
- Kill a level 13 earthborer with your level 8 (`.damage 99999`) and yo see getting 255 (excepted behaviour, killed alone with no help or group to share xp)

Original Post Below: 

---

### What client do you play on?

enUS

### Faction

Both

### Content Phase:

Generic

### Current Behaviour

Tagging a mob on a low level character, and then having the mob killed by high level character not in group grants full experience.

The exp gained by the low level character should be heavily reduced.
If someone else not in group kills the mob after the tag:
"The XP you receive depends on if that someone else will receive xp for killing that mob. If yes, you get full XP. If not, you get a tiny fraction. "

### Source

https://wowpedia.fandom.com/wiki/Mob_experience#Solo_Experience_Modifiers

the main part of the section from this wiki is from a 2007 edit with no change in mechanics explained after

from edit history:
https://wowpedia.fandom.com/wiki/Mob_experience?diff=prev&oldid=557499

have two accounts / test with two players

create a character on each account of the same race 
.level 69 or whatever
on one of the characters

kill a mob solo on level 1 should be around 50 exp
also grants 50 exp if the higher level character finishes off the mob

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/90566589b0349fc31941af6e6ed660d30a0a761e

### Operating system

Ubuntu 22.04

### Modules

### Customizations

None

### Server

ChromieCraft

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Claude Opus 4.6 was used.

### Description

When an ungrouped player whose level makes the mob gray assists in killing a creature tagged by a lower-level player, the tagger now receives halved XP instead of the full amount. Currently the tagger receives full XP, which enables a power-leveling exploit where a high-level player leaves the group to grant full XP to a lower-level character.

**Implementation:**
- Tracks the highest player attacker level on each `Creature` during combat via `_highestPlayerAttackerLevel`, updated in `LowerPlayerDamageReq()` alongside existing damage tracking
- On creature death, `KillRewarder::_RewardXP()` checks whether the mob is gray for the highest attacker (using existing `GetGrayLevel()`) and if so, halves the XP using the same formula as the group gray-member penalty (`xp / 2 + 1`)
- The field resets on evade/respawn in `ResetPlayerDamageReq()`
- Only applies to the solo (non-group) reward path — group XP already handles level differences

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/24191

## SOURCE:

Source: https://wowpedia.fandom.com/wiki/Mob_experience#Solo_Experience_Modifiers

> "The XP you receive depends on if that someone else will receive xp for killing that mob. If yes, you get full XP. If not, you get a tiny fraction."

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Create two characters: one level 8, one level 80 (`.level 7` and `.level 79`)
2. `.cheat god` for both, `.tele rfc` for both
3. **Grouped test (baseline):** Group both characters, kill a level 13 Earthborer with the level 80 → level 8 gets ~12 XP (existing behavior, unchanged)
4. **Ungrouped test (fix):** Leave group. Tag a level 13 Earthborer with the level 8, kill it with the level 80 → level 8 should now get ~128 XP instead of 255
5. **Solo test (no regression):** Kill a level 13 Earthborer solo with the level 8 → should still get 255 XP
6. **Same-level helper test:** Use a level 10 ungrouped helper instead of level 80 (mob is NOT gray for level 10) → level 8 should get full 255 XP

- [ ] The "tiny fraction" described by Wowpedia is implemented as the group gray-member formula (half XP). The exact Blizzlike fraction is not precisely documented — this may need tuning based on further research or sniff data.

## Summary by CodeRabbit

* **Game Balance Updates**
  * Creatures now track the highest player attacker level to dynamically adjust damage contribution requirements.
  * XP rewards for creature encounters are now scaled based on the highest player level engaging in combat.


src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -160,6 +161,19 @@ void KillRewarder::_RewardXP(Player* player, float rate)
...
+    else if (Creature* creature = _victim->ToCreature())
+    {
+        // If an ungrouped helper attacked this creature and the mob
+        // is gray for that helper, reduce XP — consistent with the
+        // group formula (half XP when a gray-level member is present).
+        uint8 highestLevel = creature->GetHighestPlayerAttackerLevel();
+        if (highestLevel > player->GetLevel())
+        {
+            uint8 grayLevel = Acore::XP::GetGrayLevel(highestLevel);
+            if (creature->GetLevel() <= grayLevel)
```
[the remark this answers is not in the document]
Nyeriah: This can't happen in practice. The solo path (`else if`) is mutually exclusive with the `if (_group)` branch, which is the only place `xp` gets zeroed. In the solo path, `xp` starts as `_xp` and nothing sets it to 0 before reaching this code.

Additionally, the caller already guards against this — `_RewardXP` is only invoked when `_xp > 0` (line 214: `if (_xp)`), and `_xp` is a member that doesn't change during the reward flow.

Nyeriah:
Tested but could use more testing

TheSCREWEDSoftware:
> Tested but could use more testing

The testing steps could even applied in world of having a level 60 helping farming mobs south of elywirn forest for full XP capped to your level ranged (prior to this, have not tested this and wont be able most likely during the weekend)

Nyeriah:
Its enough to pull a mob and hit it with a higher level player, I think

Kitzunu (approved):

blinkysc:
Test Before:
Mob level 2   1 + 80 = 53xp 

Test After
Mob level 2   1 + 80 = 1xp

Good to go I think

src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -160,6 +161,27 @@ void KillRewarder::_RewardXP(Player* player, float rate)
         else
             xp = 0;
     }
+    else if (Creature* creature = _victim->ToCreature())
```
r-o-b-o-t-o: Doing this in an `else` branch after `if (_group)` means that the helper calculation is not applied when one of the rewarded player is in a group.

For instance, player A and B are level 1 and grouped together. Player C is level 80 and ungrouped. Player C attacks a creature engaged by A and B. Then player A and B get the normal XP amount (albeit being divided by 2 for group sharing), instead of the lowered rate.

As it is now, the code prevents powerleveling solo players but you can still boost two or more players grouped together that are roughly the same level.

TheSCREWEDSoftware: From the issue: https://github.com/azerothcore/azerothcore-wotlk/issues/24191

As i understand nyeriahs pr and your comment i think it does suit the sources

> The way to power-level is to be near the character level you're power-leveling. Thus, a 49th character will be much more effective than a 70th character in power-leveling a 40th level character. More effective would be multiple 49th characters power-leveling the 40th level character.

r-o-b-o-t-o: What they mean is you lose less exp the closer you are in levels, so in order to maximize you'd want your powerleveler to be only a few levels above. However in that case you'd need multiple boosters to match the power of a single max level character. The sentence you quoted says nothing about the boosted char being in a group.

sudlud: I agree with roboto that it would make sense to also fix this if the low levelers are grouped together. Otherwise we would just put a patch on the exploit and still not fully fix it

Nyeriah: What I understand of this is that if the receiving character is grouped you can still bypass it. I can fix that. If there's some problem with the formula used it won't be fixed in this PR though as I haven't researched it (it currently applies the same as grouped characters).

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -1149,7 +1149,10 @@ uint32 Unit::DealDamage(Unit* attacker, Unit* victim, uint32 damage, CleanDamage
         {
             uint32 unDamage = health < damage ? health : damage;
             bool damagedByPlayer = unDamage && attacker && (attacker->IsPlayer() || attacker->m_movedByPlayer != nullptr);
-            victim->ToCreature()->LowerPlayerDamageReq(unDamage, damagedByPlayer);
+            uint8 attackerLevel = 0;
+            if (Player* attackerPlayer = attacker ? attacker->GetCharmerOrOwnerPlayerOrPlayerItself() : nullptr)
+                attackerLevel = attackerPlayer->GetLevel();
+            victim->ToCreature()->LowerPlayerDamageReq(unDamage, damagedByPlayer, attackerLevel);
```
sudlud: Please add an empty line between these lines

src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -161,6 +162,27 @@ void KillRewarder::_RewardXP(Player* player, float rate)
...
+        {
+            // Level ratio: same as group formula (playerLevel / sumOfLevels)
+            uint32 sumLevel = uint32(player->GetLevel()) + uint32(highestLevel);
+            float levelRate = float(player->GetLevel()) / float(sumLevel);
+
+            uint8 grayLevel = Acore::XP::GetGrayLevel(highestLevel);
+            if (creature->GetLevel() <= grayLevel)
+                xp = uint32(xp * levelRate / 2) + 1;
+            else
+                xp = uint32(xp * levelRate);
```
sudlud: The `else` branch (mob not gray for the helper) shouldn't reduce XP. Per the source you cite, if the helper earns XP the tagger gets full XP. As-is, receiver L30 + mob L35 + helper L40 (not gray) gives ~43% instead of full. Penalize only in the gray branch; leave `xp` untouched here.

sudlud (commented):
The gray-case fraction changed: old `xp/2+1` gave the "~128 XP" the description still cites; the new formula gives ~12 (matching the grouped amount). Worth updating the description and Known Issues. I merged master to clear the conflict and added the missing blank line in DealDamage.

src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -161,6 +162,27 @@ void KillRewarder::_RewardXP(Player* player, float rate)
         else
             xp = 0;
     }
+    else if (Creature* creature = _victim->ToCreature())
```
sudlud: The grouped-receiver case roboto/sudlud raised is still open. This penalty only runs when `_group` is null, so two low-levels grouped together with an ungrouped high-level helper still get full XP via the `if (_group)` path. The check needs to apply there too.

src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -161,6 +162,27 @@ void KillRewarder::_RewardXP(Player* player, float rate)
...
+    {
+        // If an ungrouped helper attacked this creature and the mob
+        // is gray for that helper, reduce XP — consistent with the
+        // group formula (half XP when a gray-level member is present).
+        // If an ungrouped higher-level player helped kill this creature,
+        // apply group-like XP scaling: level ratio + gray penalty.
+        uint8 highestLevel = creature->GetHighestPlayerAttackerLevel();
```
sudlud: Tracking only the highest attacker over-nerfs mixed helpers: receiver L8, mob L13, helpers L80 (gray) + L12 (not gray). The L12 earns XP so it should be full XP, but the code sees 80 and reduces. Worth a known-issue note.

src/server/game/Entities/Player/KillRewarder.cpp
```diff
@@ -161,6 +162,27 @@ void KillRewarder::_RewardXP(Player* player, float rate)
         else
             xp = 0;
     }
+    else if (Creature* creature = _victim->ToCreature())
+    {
+        // If an ungrouped helper attacked this creature and the mob
```
sudlud: This comment stacks two contradictory versions ("half XP" vs "level ratio + gray penalty"). Collapse to one accurate line.

Nyeriah:
Addressed reviews, needs testing again.

TheSCREWEDSoftware:
Same steps and testing as the original reported.

- Kill a level 13 Earthborer with your level 80, level 8 character gains 12 XP, correct behaviour
No regression here, still the same behaviour

- Leave the group, attack a level 13 earthborer with your level 8 (or let it hit your frost armour) and kill it with high level, your level 8 just gained 255 experience, not the intended behaviour, it should give you a fraction of XP not full XP
Fixed, it used to give you 255 XP to level 8, now it only gave 12 XP

Kill a level 13 earthborer with your level 8 (.damage 99999) and yo see getting 255 (excepted behaviour, killed alone with no help or group to share xp)
- No regression here, still ge the 255 XP, since you killed it alone with no help.

The above, are working and fix the reported issue, the below requirements research.

Not part of the issue but also tested higher level being 15 (so he also gets XP) alongside the level 8, grouped
- Unsure if this is the correct behaviour, Level 13 earthborer that gives you 255 XP, for level 8, 262 for level 15 when killed solo (There seems to have mising XP when group/share XP)
255 XP Level 8 | 262 XP Level 15 (Solo XP)
78 XP Level 8 | 146 XP Level 15 (Group XP) Total of 224 XP (to my understanding it's misisng beteen 31-38 XP?)

Nyeriah:
@
Thanks for testing! Glad the ungrouped fix and the no-regression cases check out.

The grouped lvl 8 + lvl 15 XP difference is the existing group-share formula (base XP is seeded from the highest non-gray member, then split per member by `level / _aliveSumLevel` times the party rate, which is 1.0 for a 2-person party), so it is not something this PR touches. The new penalty only fires for an ungrouped helper: in a group the reference level is `_maxLevel`, and the highest attacker is the highest group member, so `highestLevel > _maxLevel` is false and the halving never runs. The total differing from "solo + solo" is just how Blizzlike group sharing works, plus integer truncation, so no XP is actually lost.

Worth a separate issue if we want to dig into the group-share rounding, but it is outside the scope of this fix.
@

Issue #11433, opened 2022-04-18
[Arathi Highlands] Myzrael's Prismatic Exiles


Gultask:
https://github.com/chromiecraft/chromiecraft/issues/3373

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [X] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [X] 30-39
- [X] 40-49
- [ ] 50-59
- [ ] 60

### Current Behaviour

[Prismatic Exiles](https://wowgaming.altervista.org/aowow/?npc=2887) summoned by [Myzrael](https://wowgaming.altervista.org/aowow/?npc=2755#abilities) during [Summoning the Princess](https://wowgaming.altervista.org/aowow/?quest=656):
1. Don't engage combat with player when summoned, mindlessly following Myzrael
2. Are not a random of four elements (all are earth elementals)
3. Don't have their element cycled over time
4. Don't despawn after Myrael's death*
5. Don't cast Frost Nova and Lightning Shield should be recasted, interrupting Fireball. For some reason stunning them at start prevents it.*

________________________________________________________________________________________________________________
**Recording here:** https://www.youtube.com/watch?v=5YpzMaWHsnY

________________________________________________________________________________________________________________
![WoWScrnShot_041822_024238](https://user-images.githubusercontent.com/79346412/163741814-47083e26-5f4e-47a2-9dd1-8f54910cdfc9.jpg)

![WoWScrnShot_041822_024233](https://user-images.githubusercontent.com/79346412/163741832-8f895f9e-1c21-4f97-afc9-b29271a32efc.jpg)

### Expected Blizzlike Behaviour

Prismatic Exiles should:
1. Engage combat with player when summoned
2. Be one of four element each, chosen randomly
3. Have their element cycled periodically
4. Despawn after boss's death _(Probably)_*
5. Except Fireball they should also cast Frost Nova and upkeep Lightning Shield*

### Source

Wowhead comments are from 2008-2010 (TBC/WotlK patch), so I guess it's a correct source

https://www.wowhead.com/npc=2755/myzrael#comments
https://www.wowhead.com/npc=2887/prismatic-exile#comments

### Steps to reproduce the problem

1. Do the respective quest chain up to [Summoning the Princess](https://wowgaming.altervista.org/aowow/?quest=656)
2. Engage Myzrael
3. Observe her summoned guardians

### Extra Notes

_Myzrael_:
[...]

PR #25169, opened 2026-03-21
fix(Script): Implement correct behaviour for Prismatic Exiles summone…

MokonaNico:
## Changes Proposed:

This PR proposes changes to:

I implemented the expected behaviour of the Prismatic Exiles as explained in the link issue. I made a script in c++ code because I don't think it's possible to do the correct behaviour with smart script, especially the fact that the spell cast depend on the current DisplayId of the creature.

### AI-assisted Pull Requests

## Issues Addressed:

- https://github.com/azerothcore/azerothcore-wotlk/issues/11433

## SOURCE:

I used both link that are in the link issue to gather basic information. And I record timing for changing element in those video :

 - https://www.youtube.com/watch?v=ywK3rwR-12s
 - https://www.youtube.com/watch?v=8cwrgQffPEM&t=77s

Both video are record of wow classic but I think the timing for changing element didn't change between version.

## Tests Performed:

This PR has been:

## How to Test the Changes:


src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+        hasLightningShield = false;
+        frostNovaTimer = urand(1400, 7300); // Got those values from smart_script of 2761
+        changeElementTimer = urand(6 * IN_MILLISECONDS, 10 * IN_MILLISECONDS);
+    }
+
+    void InitializeAI() override
+    {
+        ScriptedAI::InitializeAI();
+
+        // Get the target from Myzrael
```
sudlud, line 158: The summoner is not available here yet, so `GetOwner()` returns `nullptr` and this whole block (including the `SetGazeOn`) is skipped. `InitializeAI()` runs while the creature is being added to the map, but the summon spell only attaches the owner a moment later (`SetOwnerGUID` is called after the creature is already in the world). In practice the exiles still fight because, as a hostile summon, they aggro on their own, but the intended "focus Myzrael's current target and otherwise stay passive" never actually takes effect.

The clean fix is the `IsSummonedBy` hook, which the summon system calls once the summoner is known:
```cpp
void IsSummonedBy(WorldObject* summoner) override
{
    if (Unit* unitSummoner = summoner->ToUnit())
        if (Unit* target = unitSummoner->GetVictim())
            SetGazeOn(target);
}
```
You can then drop the `InitializeAI` override entirely.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+                }
+                break;
+            case 2: // WATER
+                // Can only case frostnova if is in range and timer is done
+                Unit * target = me->GetVictim();
+                if (target && me->IsInRange(target, 0.0f, 10.0f) && frostNovaTimer <= diff)
+                {
+                    me->CastSpell((Unit*)nullptr, SPELL_FROSTNOVA, false);
+                    frostNovaTimer = urand(1400, 7300);
+                }
```
sudlud, line 205: Small bug here: when the timer is ready (`frostNovaTimer <= diff`) but the target is out of the 10y range (or there is no victim), control lands in this `else` and does `frostNovaTimer -= diff` on a `uint32`, which underflows to a huge value. After that Frost Nova will not fire again for a very long time. Better to always tick the timer down and only gate the cast itself, for example tick it once near the top of `UpdateAI` and check `!frostNovaTimer` before casting.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+        if (!me->HasUnitState(UNIT_STATE_CASTING))
+        {
+            // Cast spell depending on the current element
+            switch (currentElement) {
+            case 0: // FIRE
+                // Always cast fire ball
+                me->CastSpell(me->GetVictim(), SPELL_FIREBALL, false);
+                return;
+            case 1: // AIR
+                // Can only cast lightning shield once
```
sudlud, line 191: The linked issue asks for Lightning Shield to be kept up (recast), but `hasLightningShield` latches to `true` permanently after the first cast, so it is only ever cast once and never refreshed on a later air phase. If you want it to behave like the issue describes, consider recasting when the aura is missing instead of latching a bool, for example `if (!me->HasAura(SPELL_LIGHTNING_SHIELD))`.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+            changeElementTimer = urand(6 * IN_MILLISECONDS, 10 * IN_MILLISECONDS);
+        }
+        else changeElementTimer -= diff;
+
+        if (!me->HasUnitState(UNIT_STATE_CASTING))
+        {
+            // Cast spell depending on the current element
+            switch (currentElement) {
+            case 0: // FIRE
+                // Always cast fire ball
```
sudlud, line 187: Two small things: `GetVictim()` can be `nullptr` here, so it is worth guarding. Also there is no repeat timer, so Fireball gets recast immediately each time the previous cast finishes, which is more frequent than the old script (which cast it every 6s). A small cast timer would keep it closer to the original cadence.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
 
+enum Spells
+{
+    SPELL_FIREBALL = 34083,
+    SPELL_FROSTNOVA = 38033,
+    SPELL_LIGHTNING_SHIELD = 12550,
+    SPELL_VISUAL_TRANSFORMATION = 24085
+};
+
+// Fire, Air, Water, earth model displayId (same as the elemental in the zone)
```
sudlud, line 136: Couple of convention notes: please use the project type `uint32` rather than `uint32_t`, and since this array is never modified it should be `constexpr` (a mutable file scope global is best avoided). Wrapping it in an anonymous namespace is the usual pattern.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+            me->CastSpell(me, SPELL_VISUAL_TRANSFORMATION, true);
+            currentElement = (currentElement + _rand) % 4; // Change to a new element different from actual one
+            me->SetDisplayId(elements_display[currentElement]);
+            changeElementTimer = urand(6 * IN_MILLISECONDS, 10 * IN_MILLISECONDS);
+        }
+        else changeElementTimer -= diff;
+
+        if (!me->HasUnitState(UNIT_STATE_CASTING))
+        {
+            // Cast spell depending on the current element
```
sudlud, line 184: A few style points the CI lint (`-Werror`) will flag:
- We use Allman braces, so the `{` goes on its own line here (same for the `else` one-liners at the timer decrements).
- In the `case 2:` block, `Unit * target` should be `Unit* target`, and please wrap that case body in `{ }` since it declares a local under a `case` label.
- `me->CastSpell((Unit*)nullptr, SPELL_FROSTNOVA, false)` uses a C style cast; for a self centered AoE you can just cast on `me`: `me->CastSpell(me, SPELL_FROSTNOVA, false)`.

sudlud (commented):
Hi @MokonaNico, thanks for the PR, and sorry it sat untouched for so long. That is on us. The approach is right: this does need C++ since the element switching drives the spell choice.

I left a few inline notes. Most are minor style points the CI lint enforces, but two affect behaviour:

1. The combat engagement in `InitializeAI` does not actually run, because the summoner is not attached to the creature yet at that point. The `IsSummonedBy` hook fixes it cleanly (details inline).
2. A small unsigned timer underflow that can stop Frost Nova from firing again (details inline).

If you have time to pick this back up we are happy to help you finish it, and if you would rather hand it off that is fine too. Either way, thanks again.

MokonaNico:
I'll do that this weekend ;)
Thank you for the feedback and don't worry for the time

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+enum Spells
+{
+    SPELL_FIREBALL = 34083,
+    SPELL_FROSTNOVA = 38033,
+    SPELL_LIGHTNING_SHIELD = 12550,
+    SPELL_VISUAL_TRANSFORMATION = 24085
+};
+
+// Fire, Air, Water, earth model displayId (same as the elemental in the zone)
+uint32_t elements_display[] = {2172, 5490, 5561, 9587};
```
Kitzunu, line 137:
```suggestion
uint32 elements_display[] = {2172, 5490, 5561, 9587};
```

Kitzunu:
"This weekend", sounds like something I would say Kek

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+        if (!owner)
+            return;
+
+        Unit* target = owner->GetVictim();
+        if (target)
+            SetGazeOn(target);
+    }
+
+    void UpdateAI(uint32 diff) override
+    {
```
sudlud, line 169: Missing `if (!UpdateVictim()) return;`. Without it the element cycles and casts run out of combat with a null victim.

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
```diff
@@ -125,7 +125,96 @@ class npc_professor_phizzlethorpe : public CreatureScript
...
+struct npc_prismatic_exile : public ScriptedAI
+{
+    uint32 changeElementTimer;
+    uint32 currentElement;
+    uint32 frostNovaTimer;
+    bool hasLightningShield;
+
+    npc_prismatic_exile(Creature* creature) : ScriptedAI(creature)
+    {
+        currentElement = 3;
```
sudlud, line 148: Issue point 2 asks for a random element per exile, but every one spawns as earth and only randomizes after 6-10s. Consider rolling `currentElement` here and applying the display id. Also, these timers belong in `Reset()`, and AGENTS.md prefers `TaskScheduler`/`EventMap` over manual `-= diff` counters (which also removes the underflow noted above).

src/server/scripts/EasternKingdoms/zone_arathi_highlands.cpp
sudlud, line 137: `creature_template_model` for 2887 is display **9588**, not 9587 (that is Rumbling Exile 2592). Spawn model and the earth phase currently differ.

sudlud (commented):
Still applies cleanly to master, and all earlier inline notes are unaddressed. Two corrections and a few additions:

On the `InitializeAI` note: these are guardian summons (spell 10388 goes through `Spell::SummonGuardian`), so `SetOwnerGUID` runs in `Minion::InitStats` before `AddToMap` and `GetOwner()` is not null there. The real issue is `SummonGuardian` doing `MotionMaster::Clear` + `MoveFollow(caster)` right after the summon returns, gated only on `!IsInCombat()`, which `Unit::Attack` does not set. `IsSummonedBy` runs before that too, so it does not fix it on its own. `SetGazeOn` additionally forces `REACT_PASSIVE`, undoing the `REACT_AGGRESSIVE` guardians get by default. Worth re-testing engagement in-game.

Also: the title is truncated (`summone…`) and should use the `fix(Scripts/ArathiHighlands):` scope.

sudlud:
PR abandoned, closing it for now. please feel free to reopen it anytime in the future addressing the reviews. Thanks!

PR #25215, opened 2026-03-25
feat(Core/Unit): Add OnUnitExitCombat hook

DoubledaCoder:
A hook to allow for custom logic to apply when combat ends for all unit types.

## Changes Proposed:

This PR proposes changes to:
The UnitScript to include a new hook for custom logic to be applied when combat finishes. The hook is for a custom module i've adapted and expanded from a patch which synchronises the creature level to the player level. My issue is that there was no appropriate unit script hook for when combat finishes that allows for custom logic onto creatures.

OnEnterCombat is the initial trigger hook to sync the level, OnDamage hook is the trigger I've used which limits griefing and facilitates multi-level combat, however when combat ends e.g. leash, evade, the creature would return to the initial coordinates at whatever level it was set to. So if a Lv80 ran through a starting zone and aggroed 10 mobs, a lv5 entering that area would be swarmed (even though OnEnterCombat means they will sync to the correct level) death would be hard to avoid. With this hook, regardless of the reason, when combat is being cleared, the creature resets to within its level template bounds.

### AI-assisted Pull Requests

Copilot was only used as an assistant. All code has been tested in the context of the Playerbot fork.

## Tests Performed:

Playerbots were running for 1 hour with Free for all PVP active and the mod-mob-sync-level active doing quests etc. I also ran the below test to specifically check whether the hook is successfully triggered using the mod.

## How to Test the Changes:

1. Install the mod-mob-sync-level
2. Using a GM character attack a creature that is 10 or so lower levels than the character so it switches it level then:
    a.  run away from the creature until it resets
    b.  kill it and wait to see what occurs on respawn
    c.  fly up into the air so it cannot attack you
    d. go into an out of bounds area the creature cannot reach


src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -7489,6 +7489,7 @@ void Unit::CombatStop(bool includingCast, bool mutualPvP)
     // xinef: just in case
     if (IsPetInCombat() && !IsPlayer())
         ClearInPetCombat();
+    sScriptMgr->OnUnitStopCombat(this);
```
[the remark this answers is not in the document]
DoubledaCoder: Based on what Copilot is talking about the ClearInCombat may be a better location to place the hook because it is called from the unit::update function as well as in the CombatStop function. The player version of this OnUnitStopCombat hook exists there too. However, I am unsure how many of those scenarios/issues Copilot raises apply to non-player entities. Modules typically check for valid types prior to proceeding to their logic and regardless of whether the hook is at the start or end of the                                                               function, the end result is a unit no longer in combat so any hooked action should be assuming a non-combat state upon completion of their logic.

sudlud:
Thanks for the contribution. Plumbing looks good and the earlier Copilot points are addressed. One thing before merge: the hook fires from `CombatStop()`, which misses the natural combat-end path. When the last combat ref is purged (kill, out-of-range, PvP timeout), combat ends via `CombatReference::EndCombat()` -> `UpdateOwnerCombatState()` -> `AtExitCombat()` without ever calling `CombatStop()`, so the hook won't fire. It also fires on units that were never in combat, since `CombatStop()` runs as cleanup in many places.

The player equivalent `OnPlayerLeaveCombat` already fires from the real transition in `UpdateOwnerCombatState()`. Suggest moving the call to `Unit::AtExitCombat()` instead:

```cpp
void Unit::AtExitCombat()
{
    RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_LEAVE_COMBAT);
    sScriptMgr->OnUnitStopCombat(this);
}
```

That fires once on the actual transition, for all unit types, covering every exit path. `ClearInCombat()` would also miss the `EndCombat` path, so `AtExitCombat()` is the right spot.

DoubledaCoder:
> Thanks for the contribution. Plumbing looks good and the earlier Copilot points are addressed. One thing before merge: the hook fires from `CombatStop()`, which misses the natural combat-end path. When the last combat ref is purged (kill, out-of-range, PvP timeout), combat ends via `CombatReference::EndCombat()` -> `UpdateOwnerCombatState()` -> `AtExitCombat()` without ever calling `CombatStop()`, so the hook won't fire. It also fires on units that were never in combat, since `CombatStop()` runs as cleanup in many places.
> 
> The player equivalent `OnPlayerLeaveCombat` already fires from the real transition in `UpdateOwnerCombatState()`. Suggest moving the call to `Unit::AtExitCombat()` instead:
> 
> ```c++
> void Unit::AtExitCombat()
> {
>     RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_LEAVE_COMBAT);
>     sScriptMgr->OnUnitStopCombat(this);
> }
> ```
> 
> That fires once on the actual transition, for all unit types, covering every exit path. `ClearInCombat()` would also miss the `EndCombat` path, so `AtExitCombat()` is the right spot.

Makes sense and thanks for reviewing my work. I didn't know CombatStop was run outside of combat!

I agree and will amend when i have time.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -7584,6 +7584,7 @@ void Unit::UpdatePetCombatState()
 void Unit::AtExitCombat()
 {
     RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_LEAVE_COMBAT);
+    sScriptMgr->OnUnitStopCombat(this);
```
Kitzunu:
```suggestion
    sScriptMgr->OnUnitExitCombat(this);
```

src/server/game/Scripting/ScriptDefines/UnitScript.cpp
```diff
@@ -124,6 +124,11 @@ void ScriptMgr::OnUnitEnterCombat(Unit* unit, Unit* victim)
     CALL_ENABLED_HOOKS(UnitScript, UNITHOOK_ON_UNIT_ENTER_COMBAT, script->OnUnitEnterCombat(unit, victim));
 }
 
+void ScriptMgr::OnUnitStopCombat(Unit* unit)
+{
+    CALL_ENABLED_HOOKS(UnitScript, UNITHOOK_ON_UNIT_STOP_COMBAT, script->OnUnitStopCombat(unit));
+}
+
```
Kitzunu:
```suggestion
void ScriptMgr::OnUnitExitCombat(Unit* unit)
{
    CALL_ENABLED_HOOKS(UnitScript, UNITHOOK_ON_UNIT_EXIT_COMBAT, script->OnUnitExitCombat(unit));
}

```

src/server/game/Scripting/ScriptDefines/UnitScript.h
```diff
@@ -41,6 +41,7 @@ enum UnitHook
     UNITHOOK_ON_DISPLAYID_CHANGE,
     UNITHOOK_ON_UNIT_ENTER_EVADE_MODE,
     UNITHOOK_ON_UNIT_ENTER_COMBAT,
+    UNITHOOK_ON_UNIT_STOP_COMBAT,
```
Kitzunu:
```suggestion
    UNITHOOK_ON_UNIT_EXIT_COMBAT,
```

src/server/game/Scripting/ScriptDefines/UnitScript.h
```diff
@@ -106,6 +107,7 @@ class UnitScript : public ScriptObject
 
     virtual void OnUnitEnterEvadeMode(Unit* /*unit*/, uint8 /*evadeReason*/) { }
     virtual void OnUnitEnterCombat(Unit* /*unit*/, Unit* /*victim*/) { }
+    virtual void OnUnitStopCombat(Unit* /*unit*/) { }
```
Kitzunu:
```suggestion
    virtual void OnUnitExitCombat(Unit* /*unit*/) { }
```

src/server/game/Scripting/ScriptMgr.h
```diff
@@ -563,6 +563,7 @@ class ScriptMgr
     void OnDisplayIdChange(Unit* unit, uint32 displayId);
     void OnUnitEnterEvadeMode(Unit* unit, uint8 why);
     void OnUnitEnterCombat(Unit* unit, Unit* victim);
+    void OnUnitStopCombat(Unit* unit);
```
Kitzunu:
```suggestion
    void OnUnitExitCombat(Unit* unit);
```

Kitzunu (approved):

Kitzunu:
Thank you for the PR!

PR #25272, opened 2026-03-27
Fix: Change Faction Issue

ArashDevelop:
Fix: Change Faction Issue:
this changes fix character service: faction change, before that when try to change faction worldserver.exe crashed.


Nyeriah:
Please clean up your pull request by reverting all the style changes

sudlud:
Please address nyeriahs review and reduce the pr changes to the minimum required. There's e.g. lots of whitespace changes

Also using the PR template is highly appreciated.

sudlud:
Thanks for the contribution, @ArashDevelop, and for taking the time to dig into the faction-change crash.

We're going to close this PR for now. Please feel free to reopen or submit a fresh PR anytime.

To be clear on why it can't be merged as-is, and what a follow-up would need:

**Blocking issues**
- The review feedback from @Nyeriah and @sudlud has not been addressed: the PR still reformats the entire file (pointer style, 80-col wrapping, indentation), which is unrelated to the fix and fails our code style.
- Underneath the reformatting, the change is effectively a revert of `CharacterHandler.cpp` to an older revision, and it reintroduces real regressions:
  - Spell conversion swaps the prepared-statement params for `CHAR_DEL_CHAR_SPELL_BY_SPELL` (`WHERE guid = ? AND spell = ?`), so faction-specific spells are no longer removed.
  - The title-conversion `else` branch uses `htitleInfo->bit_index` instead of `atitleInfo->bit_index`, clearing the wrong title bit.
  - Taxi-mask handling reverts to old byte-indexed logic that doesn't match the current `TaxiMask` representation.
  - Named constants (`MAP_EASTERN_KINGDOMS`, `AREA_STORMWIND_CITY`, ...) replaced with magic numbers, and C++17 range loops reverted to iterator loops.

**What a mergeable follow-up should contain**
Only the minimal real fix against current `master`:
1. The genuine crash fix: the bounds check `if (index >= knownTitles.size()) continue;` before indexing `knownTitles[index]` in the title conversion.
2. If blocking DK faction changes is intended, gate it on the starting-zone/campaign state rather than a flat `level <= 70`, which also blocks legitimate level 70 Death Knights.

Plus the PR template filled in with a clear reproduction case. Thanks again.

PR #25349, opened 2026-04-02
fix(Scripts/ICC): make Blood Queen Lana'thel IsVampire check difficulty-aware

### LINKED ISSUE #25343: [ICC] Blood-Queen Lana'thel Fight error
### Current Behaviour

In ICC, during the battle against Blood-Queen Lana'thel, you can get bitten more than once because the spell checks are misconfigured and only take into account those from the 25 Normal mode.

### Expected Behaviour

You should only be able to be bitten 1 time

### Source

_No response_

Go to ICC in mode 25H
Got bitten
Try to Got bitter again

### Extra Notes

_No response_

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/e144326f02f3c25533f66a8579a0bfbc91ceb0db

### Operating system

Windows

### Custom changes or Modules

Image

Those spells are only from normal mode

Image

So this isn't taking in count in this part of code

EricksOliveira:
Players can be affected by Vampiric Bite multiple times during the Blood Queen Lana'thel encounter, especially in heroic difficulties.  
This happens because the current IsVampire() check only validates auras from a single difficulty, ignoring other difficulty-specific spell IDs.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/25343

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


src/server/scripts/Northrend/IcecrownCitadel/boss_blood_queen_lana_thel.cpp
```diff
@@ -81,10 +81,20 @@ enum Shadowmourne
...
+};
+
 bool IsVampire(Unit const* unit)
 {
-    if (unit->HasAnyAuras(SPELL_ESSENCE_OF_BLOOD_QUEEN, SPELL_ESSENCE_OF_THE_BLOOD_QUEEN_PLR, SPELL_FRENZIED_BLOODTHIRST))
-        return true;
+    uint8 difficulty = unit->GetMap()->GetSpawnMode();
+
+    for (uint8 i = 0; i < 3; ++i)
+        if (unit->HasAura(vampireAuras[i][difficulty]))
```
sudlud: I would say theres a function like hasAnyAura(..) taking multiple spell ids, also if these have entries in spell difficulty, the core probably automatically handles the different spell ids depending on raid difficulty

sudlud:
Looks correct to me. The vampire auras get stored under their difficulty-resolved IDs (e.g. Essence `71525`/`71531` in 25-man), so the old `HasAnyAuras(70867, 70879, 70877)` only matched in 10N and let already-bitten players be re-targeted in every other mode. Wrapping each base ID in `GetSpellIdForDifficulty` fixes that, and it matches the pattern already used at the end of `vampiric_bite`'s `OnCast`.

This also addresses the earlier feedback: the hardcoded `vampireAuras` table is gone, so there's no second source of truth to drift from `spelldifficulty_dbc`.

Out of scope here, but a possible follow-up: `CleanAuras()` removes the same auras by base ID and has the identical difficulty mismatch, leaving stale Essence buffs after a 25-man wipe.

sudlud (approved):
The updated version addresses my comment. Using `GetSpellIdForDifficulty` over the base IDs is the right call and matches the existing pattern in the script.

The separate `CleanAuras()` difficulty mismatch noted above is a follow-up, not a blocker here.

PR #25432, opened 2026-04-12
fix(Core/Spells): Fix proc spell type classification for cast and finish

### LINKED ISSUE #25371: [Core] - Crafting, equipping or taking damage can proc trinkets, it shouldn't (that dont have it stated)
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/9250#issuecomment-4186867662

Triage Notes: 

- Make a character
- `.level 79`
- `.learn 27028`
- `.additem 2589 100`
- `.additem 37657`
- Equip `Spark of Life` and wait 1 minute.
- Start crafting `Linen bandage` from the `First Aid` proffesion.
- Wait until it procs `Spark of Life`

> Craft something, watch a trinket like spark of life, mirror of truth etc.

Image

---

You also can proc it from equipping / de-equipping

- `.additem 44255`
- `.additem 44253`
- Equip both `Darkmoon Card: Greatness`
- Keep equipping / swapping with `Spark of Life`
- It will proc `Spark of Life` eventually

Image

https://streamable.com/z4kxvb from @amed80 

--- 

Taking damage also procs it

- `.cheat god`
- `.quest add 10877`
- `.go ga 26109`
- Open `Massive Teasure Chest``
- Get hit and see the trinket proc `Spark of Life`

Image

Thanks for @amed80 bringing it up and @youdeoo indirectly sharing a good testing steps for this issue

Original Post Below: 

---

### Current Behaviour

Crafting anything including creating spell stones, mana stones or picking up health stones will cause trinkets to proc.

Image

### Expected Behaviour

Should not proc trinekts.

### Source

NA

On the trinket descriptions.

Craft something, watch a trinket like spark of life, mirror of truth etc.

### Extra Notes

NA

### Customizations

NA

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/8ea9252d764767dba2d9ddd7fd5c794a13fdc054

### Operating system

Ubuntu 22.04

### Modules

### Server

ChromieCraft

youdeoo:
Cache static proc spell type masks on SpellInfo and refresh them during spell loading so spells can be classified as damage, heal, or no-damage-heal from static data when reliable.

Use the cached classification for PROC_SPELL_PHASE_CAST and PROC_SPELL_PHASE_FINISH, while keeping PROC_SPELL_TYPE_MASK_ALL as a fallback for ambiguous dummy and trigger-based spells. Apply the same logic to zero-result outcomes so absorb and immune cases keep the correct damage or heal spell type.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [X] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/25371

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1..additem 34657
2.Take damage from an enemy — proc should NOT trigger
3.Use an enchant scroll — proc should NOT trigger
4.Apply a buff to yourself — proc should NOT trigger
5.Perform a Life skill (gathering, fishing, etc.) — proc should NOT trigger
6.Perform a Professional craft — proc should NOT trigger

1..additem 40432
2.Use a DOT (Damage over Time) skill — verify each tick does NOT trigger the proc
3.Use a HOT (Heal over Time) skill — verify each tick does NOT trigger the proc

- [ ]
- [ ]

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1775946528519437800.sql
+++ b/data/sql/updates/pending_db_world/rev_1775946528519437800.sql
@@ -3,4 +3,3 @@ DELETE FROM `spell_proc` WHERE `SpellId` IN (60485, 60519);
 INSERT INTO `spell_proc` (`SpellId`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `ProcFlags`, `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`, `DisableEffectsMask`, `ProcsPerMinute`, `Chance`, `Cooldown`, `Charges`) VALUES
 (60485, 0, 0, 0, 0, 0, 81920, 3, 4, 0, 0, 0, 0, 0, 0, 0),
 (60519, 0, 0, 0, 0, 0, 82944, 3, 2, 0, 0, 0, 0, 0, 50000, 0);
-
--- a/data/sql/updates/pending_db_world/rev_1775946528519437800.sql
+++ b/data/sql/updates/pending_db_world/rev_1775946528519437800.sql
@@ -1,0 +1,6 @@
+-- Fix spell_proc: Add missing proc entry for spell 60485 and fix SpellTypeMask for spell 60519
+DELETE FROM `spell_proc` WHERE `SpellId` IN (60485, 60519);
+INSERT INTO `spell_proc` (`SpellId`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `ProcFlags`, `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`, `DisableEffectsMask`, `ProcsPerMinute`, `Chance`, `Cooldown`, `Charges`) VALUES
+(60485, 0, 0, 0, 0, 0, 81920, 3, 4, 0, 0, 0, 0, 0, 0, 0),
+(60519, 0, 0, 0, 0, 0, 82944, 3, 2, 0, 0, 0, 0, 0, 50000, 0);
+
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -6792,14 +6792,28 @@ void Unit::ProcSkillsAndAuras(Unit* actor, Unit* victim, uint32 procAttacker, ui
             // procs that check for damage/heal type based on spell info (like Backlash)
             // At FINISH phase, damageInfo may be null but spell did do damage - use MASK_ALL
             // to match TrinityCore behavior (see TC Spell.cpp PROC_SPELL_PHASE_FINISH call)
-            spellTypeMask = PROC_SPELL_TYPE_MASK_ALL;
+            bool reliable = false;
+
+            if (procSpellInfo)
+                spellTypeMask = procSpellInfo->GetStaticProcSpellTypeMask(reliable);
+
+            if (!reliable)
+                spellTypeMask = PROC_SPELL_TYPE_MASK_ALL;
         }
         else if (healInfo && healInfo->GetHeal())
             spellTypeMask = PROC_SPELL_TYPE_HEAL;
         else if (damageInfo && (damageInfo->GetDamage() || damageInfo->GetAbsorb()))
             spellTypeMask = PROC_SPELL_TYPE_DAMAGE;
         else if (procSpellInfo)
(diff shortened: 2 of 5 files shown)


blinkysc:
Isn't the cpp not even needed for this? You could solve the problem with just the override?

youdeoo:
> Isn't the cpp not even needed for this? You could solve the problem with just the override?

I do not think the SQL override alone is sufficient here.

The `spell_proc` entry fixes one side of the issue: using `SpellTypeMask = 3` and `SpellPhaseMask = FINISH` prevents periodic DoT/HoT ticks from proccing it, which matches the intended “cast a damaging or healing spell” behavior.

However, that override by itself still leaves another problem: self-cast buff spells can also proc it. The reason is that the SQL override can only choose the proc phase and the requested spell type mask, but it still depends on the core correctly classifying the triggering spell at that phase.

At `PROC_SPELL_PHASE_FINISH`, the proc system may not have `damageInfo` / `healInfo` available even for a damaging spell, so relying on the existing runtime classification is not enough. That is what the C++ part fixes: it gives CAST/FINISH (and zero-result cases such as absorb/immune) a more reliable spell type classification from static spell data, so damaging/healing spells stay DAMAGE/HEAL while pure buffs remain NO_DMG_HEAL.

So in my view, the override is needed, but it only fixes the specific database configuration. The C++ change is still needed to fix the generic spell type classification issue behind it.

blinkysc:
The big problem with this is this is essentially duplicate logic for the proc system. Additionally, it blanket stops all trinkets from procing based on things like trade skill when we know there is trinkets that have this behavior (good example from the linked issue https://www.wowhead.com/wotlk/item=40255/dying-curse#comments:id=498530). I think the better approach might be to be more selective and just override (though exhausting I know) so that you preserve the blizzlike behavior for the trinkets that for whatever reason blizzard decided should proc on nonsense.

blinkysc:
Here is some examples of approaches you "might take", but you should be very very cautious because the goal should be to preserve the blizzlike behavior as much as possible... Even the ridiculous stuff

https://github.com/azerothcore/azerothcore-wotlk/pull/25061
https://github.com/azerothcore/azerothcore-wotlk/pull/24903
https://github.com/azerothcore/azerothcore-wotlk/pull/24890

youdeoo:
For `Illustration of the Dragon Soul` ([[item 40432](https://www.wowhead.com/wotlk/item=40432/illustration-of-the-dragon-soul#comments)](https://www.wowhead.com/wotlk/item=40432/illustration-of-the-dragon-soul#comments)) specifically, the issue I am running into is that the item text is not just “on spellcast”, it is:

“Each time you cast a damaging or healing spell…”

So the expected behavior here is:

* direct damage / direct heal spells should proc
* pure DoT / HoT application spells should also proc on the initial cast
* periodic ticks themselves should not proc
* pure buffs should not proc

The problem is that I could not get all of those conditions at the same time with `spell_proc` alone.

If I use `SpellPhaseMask = 2` (`HIT`) together with `SpellTypeMask = 3`, it filters buffs better, but then pure DoT / HoT application spells do not proc, because at `HIT` they do not have direct damage / heal information. That makes the behavior inconsistent with the item text.

If I try `SpellPhaseMask = 3` (`CAST | HIT`), then DoT / HoT application can proc, but direct damage / direct heal spells can proc twice: once on cast and once on hit.

If I use `SpellPhaseMask = 4` (`FINISH`), it avoids periodic ticks proccing every tick, but then pure buffs can also proc, which is also inconsistent with “damaging or healing spell”.

So from what I can see, for this item specifically, I cannot reproduce the tooltip behavior correctly with the current `spell_proc` fields alone without introducing another inconsistency somewhere else.

If you have a more selective approach in mind for this specific case, I would be happy to rework it in that direction.

blinkysc:
That trinket has this in DBC
PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_POS (0x4000)                                                                                                  
PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_NEG (0x10000)                                                                                                 
PROC_FLAG_DONE_PERIODIC (0x40000) 

So it should proc from hots?

blinkysc:
The best approach is to keep it Blizzlike, tooltips are not always what it seems and procs can be a bit weird. Even reading through the comments of that wowhead for Illustration of the Dragon Soul you can see the current behavior without this PR seems correct.

youdeoo:
Thanks for the clarification and the additional references. I am going to close this PR for now.
Instead, I will prepare a separate PR with a targeted spell_proc fix for Spark of Life.

PR #25591, opened 2026-04-25
fix(Core/Vehicle): no longer eject and overwrite an existing passenger

sogladev:
## Changes Proposed:

This PR proposes changes to:

followup on:
- https://github.com/azerothcore/azerothcore-wotlk/pull/25587

> TODO: Using the harpoon can overwrite / mount an already occupied player's seat.

Current vehicle logic allows a player to board a vehicle by ejecting an existing passenger. It's more sensible to eject a passenger before overwriting their seat.

Unsure which core logic or scripts rely on this force-eject behavior.

After this PR, dismounts must be done explicitly before mounting.

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

GPT-5.4 to triage

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

2 or more characters
```
.tele storm
.add q 12886
```

Mount an already occupied vehicle

1.
2.
3.

- [ ]
- [ ]


sudlud (approved):
Reviewed the final diff against current upstream/master. The change is correct and well-scoped. When a caller requests a specific occupied seat, refusing the board (`return false`) is the right behavior and matches retail.

I checked both core callers: `ChangeSeat` already guards with `HasEmptySeat`/`GetNextEmptySeat` before calling `AddPassenger`, so its `ABORT()` path can't be hit; `_EnterVehicle` resets `m_vehicle` on failure and the CONTROL_VEHICLE aura unapply path no-ops when `m_vehicle` is null, so a failed board leaves no stuck state. Only explicit positive seat IDs reach the changed branch; auto-seat selection is untouched.

The earlier Copilot comments targeted an intermediate `GetNextEmptySeat` revision that was dropped. The final `return false` actually implements what that last comment asked for (reject instead of falling through to another seat).

Minor, non-blocking: the trailing `ASSERT(seat->second.IsEmpty())` is now redundant since the preceding check guarantees it. LGTM.

PR #25600, opened 2026-04-26
fix(Scripts/IcecrownCitadel): fix Valithria Dreamwalker portals not spawning correctly

mingzi120:
### Description
Fixes a regression introduced in PR #23477 where Valithria Dreamwalker fails to spawn portals correctly in all difficulties (10/25/Heroic/Normal).

**Root cause:**
- Using false for CastSpell caused subsequent portals in the for loop to fail to cast due to the recently added strict cast validation in Core/Spells, preventing the remaining portals from spawning in 10-man mode, and completely breaking them in 25-man mode.

This fix sets the triggered flag to true to ensure all portals spawn correctly in a single loop, and relies on the core's spelldifficulty_dbc system to handle the spell ID automatically.


Kitzunu:
Use spelldifficulty, that eliminates the need to check for heroic

mpfans:
https://github.com/azerothcore/azerothcore-wotlk/pull/25516  
This PR has fixed the bug where the door does not refresh

Kitzunu (approved):

Issue #25375, opened 2026-04-05
[Magtheridon] Encounter does not properly start when released by killing all Channelers


b0ngface:
### Current Behaviour

When all 5 Hellfire Channelers are killed without directly engaging Magtheridon:
- Magtheridon remains in a banished/passive state
- Instance doors do not close
- DBM / boss mods do not detect the fight
- Boss does not properly enter combat state

<img width="1763" height="1153" alt="Image" src="https://github.com/user-attachments/assets/1e34aae5-f982-4945-8cb8-3c1b035c219d" />

### Expected Behaviour

After all Hellfire Channelers are killed Magtheridon should:
- Become fully active (remove Shadow Cage / banish)
- Enter combat properly
- Close instance doors

### Steps to reproduce the problem

Enter Magtheridon's Lair:

1. Kill all 5 Hellfire Channelers
2. Magtheridon yells and appears to be released
3. Observe:
 - Magtheridon yells and appears to be released
 - However:  Doors remain open
 - Encounter is not properly started
 - Boss is not fully engaged

### AC rev. hash/commit

[89b810b](https://github.com/mod-playerbots/azerothcore-wotlk/commit/89b810b2fb92b21667df6f7a1c4692232eb17728)

### Operating system

Ubuntu 24.04

### Custom changes or Modules

[mod-playerbots](https://github.com/mod-playerbots/mod-playerbots)
[mod-ah-bot-plus](https://github.com/NathanHandley/mod-ah-bot-plus)

Rorschach91:
Confirmed. @Nyeriah another regression?

Nyeriah:
I'm looking into it

b0ngface:
Fix is working. Thanks a lot!

Praetak:
This bug happened to me as well.  After the fix, Magtheridon does get released but the Hellfire Channelers also all respawn which doesnt end well.

Rorschach91:
> This bug happened to me as well. After the fix, Magtheridon does get released but the Hellfire Channelers also all respawn which doesnt end well.

Confirmed. CC @Nyeriah

PR #25603, opened 2026-04-26
fix(Scripts/MagtheridonsLair): prevent Hellfire Channelers from respawning during boss fight

TazzerMAN:
## Changes Proposed:

This PR fixes a bug in **Magtheridon's Lair** where killing the 5 Hellfire Channelers in phase 1 caused them to respawn and rejoin the boss fight as soon as Magtheridon himself engaged the raid.

**Root cause:** `SetBossState(DATA_MAGTHERIDON, IN_PROGRESS)` was triggered late in the encounter (when Magtheridon engaged), at which point `InstanceScript::UpdateMinionState()` iterated over the registered minions (the Channelers) and called `Respawn()` on every dead one. The boss script never excluded that path.

**Fix:** transition the encounter to `IN_PROGRESS` as soon as the first Channeler enters combat. By the time Magtheridon's own `JustEngagedWith` runs, the state is already `IN_PROGRESS` and `UpdateMinionState` becomes a no-op for the dead Channelers.

**Side benefit (intentional):** room doors close and boss mods (DBM, etc.) are notified the moment a Channeler is engaged, rather than waiting for Magtheridon to attack a player. This matches the way the encounter behaves on retail.

**Phase-1 wipe handling:** since Magtheridon stays passive (immune to PC) until the last Channeler dies, his `BossAI` never triggers `_EnterEvadeMode`. The instance script now resets the encounter to `NOT_STARTED` when all Channelers have evaded; otherwise, after a phase-1 wipe, the room doors would stay locked. A new SmartAI hook on `SMART_EVENT_EVADE` for entry 17256 sends `DATA_CHANNELER_COMBAT` with `data=0` to trigger the reset.

This PR proposes changes to:
- [x] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request.

I used Claude (Anthropic) to help analyze the AzerothCore source, identify the root cause, and draft the patch. I tested all scenarios in-game on a current master build (which already includes #25379) and I understand the changes I'm submitting.

## Issues Addressed:

Related to #25375 (closed on 2026-04-05 by #25379). #25379 fixed the scheduler-vs-`UpdateVictim` deadlock in `boss_magtheridon.cpp` so Magtheridon now becomes engageable properly, but it did not touch the Channeler-respawn flow described above — that flow is what this PR addresses.

No open issue currently tracks the respawn bug. I'm happy to open one if a maintainer prefers having a dedicated issue reference.

## SOURCE:

- [x] Knowledge databases or other public sources

On retail/Classic, killing the Hellfire Channelers permanently removes them for the duration of the encounter — they never respawn during phase 2/3. Confirmed via Wowhead, player guides and video evidence.

## Tests Performed:

I tested on a current master build (commit `295e45b0a`, which includes #25379).

## How to Test the Changes:

- [x] This pull request requires further testing.

### Test scenarios

1. **Normal kill** — Enter Magtheridon's Lair, attack one of the Hellfire Channelers. **Expect:** doors close immediately, encounter is `IN_PROGRESS`. Kill all 5 Channelers; wait for Magtheridon to be released; defeat him. **Expect:** no Channeler ever respawns during phase 2/3.

2. **Phase-1 wipe** — Engage the Channelers, then wipe the raid. **Expect:** surviving Channelers evade and reset; once the last one is back home, the encounter resets to `NOT_STARTED`, doors open, and the raid can re-enter.

3. **Phase-1 leave** — Engage a Channeler, then have everyone leave the room (Feign Death / vanish / out of range). **Expect:** same as wipe — encounter resets when Channelers evade.

4. **Phase 2/3 wipe** — Kill all Channelers, engage Magtheridon, wipe. **Expect:** standard `BossAI::_EnterEvadeMode` reset; doors open. (Verifies the new code path doesn't interfere with the existing one.)

5. **Victory** — Defeat Magtheridon. **Expect:** doors open, encounter state `DONE`.

- None known. The fix is contained to the Magtheridon's Lair instance script and SmartAI for entry 17256.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1778152540751174200.sql
+++ b/data/sql/updates/pending_db_world/rev_1778152540751174200.sql
@@ -8,13 +8,13 @@
 --
 DELETE FROM `creature_formations` WHERE `leaderGUID`=90978;
 INSERT INTO `creature_formations`
-    (`leaderGUID`,`memberGUID`,`dist`,`angle`,`groupAI`,`point_1`,`point_2`)
+(`leaderGUID`,`memberGUID`,`dist`,`angle`,`groupAI`,`point_1`,`point_2`)
 VALUES
-    (90978, 90978, 0, 0, 8, 0, 0),
-    (90978, 90979, 0, 0, 8, 0, 0),
-    (90978, 90980, 0, 0, 8, 0, 0),
-    (90978, 90981, 0, 0, 8, 0, 0),
-    (90978, 90982, 0, 0, 8, 0, 0);
+(90978, 90978, 0, 0, 8, 0, 0),
+(90978, 90979, 0, 0, 8, 0, 0),
+(90978, 90980, 0, 0, 8, 0, 0),
+(90978, 90981, 0, 0, 8, 0, 0),
+(90978, 90982, 0, 0, 8, 0, 0);
 
 --
 -- On evade:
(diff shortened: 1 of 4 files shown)


Nyeriah:
I'm not sure I'm seeing this one, the issue was that the boss resets/reenters combat switching from IN_COMBAT to NONE then IN_COMBAT again. Since the ads are registered as minions they would respawn when it changed to IN_COMBAT. I think the solution would rather be use formations to handle the respawn instead of the minion system and avoid the respawn measure in place

TazzerMAN:
Thanks for the feedback, @Nyeriah :)

Formations make sense. In the case I ran into, it wasn’t really an oscillation, but rather the first transition to `IN_PROGRESS` when Mag engages after the 5th Channeler dies. That said, I think it still comes from the same root cause.
For v2 if you want, I’ll remove them from `minionData[]` and add a formation with `RESPAWN_ON_EVADE`.

TazzerMAN:
Hey @Nyeriah, Pushed v2. 

Channelers are out of the minion system, replaced by a creature_formations row with `RESPAWN_ON_EVADE` like you suggested.

Couple of small things on top:

- Smart AI hook on evade tells the instance to reset Magtheridon's `_channelersKilled` counter, otherwise `SetInCombatWithZone()` keeps him in combat and his Reset never runs, so the counter stays stale between pulls.
- Another one strips the Soul Transfer stacks from the survivors so they don't keep the buffs through the wipe.
- If the raid wipes once all 5 are already dead (phase 2/3 or during the 3s release window), nobody's left alive to trigger the formation, so I respawn them manually from `SetBossState` as a fallback.
Tested full kill, 3/5 wipe + repull, and phase 2/3 wipe. All good on my end.

sogladev (changes requested):
Hi @TazzerMAN 

Firstly, I cleaned up the sql and script. 

> Normal kill — Enter Magtheridon's Lair, attack one of the Hellfire Channelers. Expect: doors close immediately, encounter is IN_PROGRESS

During testing, the Hellfire Channelers no longer aggro as a group. The door does not close either. Could you have another look at this?

TazzerMAN:
Hi @sogladev, thanks for the cleanup and the catch!

You were right, sorry, I had the formation behavior wrong. 
Fixed in the latest push, should be good now: attacking one Channeler pulls all 5, door closes immediately, and partial wipes properly respawn the dead members. 
Let me know if anything else looks off.

sogladev:
@TazzerMAN thanks for the update!

The Magtheridon encounter does not hard reset correctly; If the party wipes, the entire room needs to despawn.

Currently, Magtheridon only hard resets after all channelers are dead, but the Channelers never despawn.

Here's a visual/clip of the expected behavior: https://youtu.be/ald7dLbqN5E&t=6084

TazzerMAN:
Hi @sogladev,

Thanks for the video, it helped a lot to understand the intended behavior.

Sorry for the delay on this one. I ended up reworking quite a bit because I initially took the wrong approach by trying to preserve state across wipes instead of fully resetting the encounter.

The latest changes now use a single reset path through `SetBossState(NOT_STARTED)`:

* Any wipe now properly resets the encounter, respawns the Channelers, and cleans up lingering Burning Abyssals.
* Phase 1 wipe detection was moved to `OnCreatureEvade` since Magtheridon's evade logic doesn't fire while he's passive.
* Channeler death tracking was moved from SAI to the instance script (`OnUnitDeath`).
* Removed the remaining SAI hooks (`id=6`, `id=9`, `id=10`), so everything is now handled in C++ except the aggro trigger.

It should now behave much closer to what your video shows. Let me know if you spot anything odd.

sogladev (approved):
Thank you for continuing your work on this!

> Any wipe now properly resets the encounter, respawns the Channelers, and cleans up lingering Burning Abyssals.

If you wipe after clearing the channelers and before engaging Magtheridon, the encounter becomes stuck.

I consider it a TODO as the main reset issues are fixed

Nyeriah (approved):

PR #25641, opened 2026-04-28
feat(Core/Scripting): add spell and dispel script hooks

scarecr0w12:
## Changes Proposed:

This PR proposes changes
to:

This PR adds new script hooks for spell-related lifecycle events and dispel handling in the core scripting layer.

Included changes:
- Add new hooks for spell and damage-related events in the script manager interfaces.
- Wire cast and aura/spell execution paths so scripts can observe relevant spell activity.
- Add `OnSpellDispel` handling so scripts can react to dispel events with the participating units and spell context.

Touched areas include `ScriptMgr`, `GlobalScript`, `PlayerScript`, `UnitScript`, `Player`, `Unit`, `Spell`, and aura handling paths.

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any. GitHub Copilot with GPT-5.4.

## Issues Addressed:

- Closes

## SOURCE:

**Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:

Additional verification performed:
- Built successfully on Linux in this workspace (`./acore.sh compiler build`).

## How to Test the Changes:

1. Build the core and enable a minimal test script or module that implements the newly exposed spell hooks on `GlobalScript`, `PlayerScript`, or `UnitScript` and logs when each callback fires.
2. In-game, cast normal damage and healing spells and verify the corresponding cast/damage-related hooks fire with the expected caster, target, and spell information.
3. Use one or more dispel-capable spells (for example, `Dispel Magic`, `Cleanse`, or `Purge`) against units with removable auras and verify `OnSpellDispel` receives the expected dispeller, target, and dispel context.

- [ ] Broader in-game validation across more dispel variants and scripted spell consumers would still be valuable.
- [ ] Reviewers should pay extra attention to generic spell-flow regressions because this touches shared core paths.


blinkysc:
@scarecr0w12 if you could address the CI and the AI comments (correct or not) I can find some time to check this out fully. I would say you should also make sure the systems are working without your module utilizing the hooks to make sure nothing broken too.

scarecr0w12:
> @scarecr0w12 if you could address the CI and the AI comments (correct or not) I can find some time to check this out fully. I would say you should also make sure the systems are working without your module utilizing the hooks to make sure nothing broken too.

Will do

scarecr0w12:
Addressed the AI review items in commit `4649529ab`.

Summary:
1. `OnEnvironmentalDamage(...)` now receives the original raw input damage before mitigation.
2. `OnSpellSendSpellGo(...)` now fires only when `SMSG_SPELL_GO` is actually sent to clients.
3. `OnDealDamageShieldDamage(...)` now receives synchronized hook data after the post-processing that affects the final absorb state.
4. Removed the duplicate `OnPlayerBeforeGetLevelForXPGain(...)` declaration from `ScriptMgr.h` as cleanup.
5. Hook docs were aligned with the resulting behavior.

Notes:
- The duplicate declaration cleanup was still worth doing, but it was redundant rather than a real redefinition/ambiguity compile failure.
- Rebuilt on Linux after rebasing onto the latest PR branch tip; `worldserver` builds successfully on this state.

scarecr0w12:
@blinkysc Did a clean test of the PR code with no modules included. Server launches, is joinable, and shows no obvious signs of issues.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2190,6 +2190,11 @@ void Unit::DealDamageShieldDamage(Unit* victim)
         data << uint32(i_spellProto->GetSchoolMask());
         victim->SendMessageToSet(&data, true);
 
+        if (absorb > dmgInfo.GetAbsorb())
+            dmgInfo.AbsorbDamage(absorb - dmgInfo.GetAbsorb());
+
+        sScriptMgr->OnDealDamageShieldDamage(&dmgInfo, overkill > 0 ? overkill : 0);
```
blinkysc: Hidden behavior change. `DamageInfo::AbsorbDamage` (Unit.cpp:262) bumps `m_absorb` **and** subtracts from `m_damage`. After this sync, `dmgInfo.GetDamage()` no longer matches the local `damage` actually passed to `Unit::DealDamage` on the next line — scripts inspecting `GetDamage()` see `0` while gameplay deals whatever `DealDamageMods` left.

If the goal is to hand the script the post-`DealDamageMods` absorb, please pass the values to the hook explicitly (`OnDealDamageShieldDamage(attacker, victim, damage, absorb, ...)`) rather than mutating `dmgInfo` in-place. `AbsorbDamage` is the wrong primitive for a "report final absorb" sync.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6737,6 +6752,9 @@ void Unit::SendSpellNonMeleeDamageLog(SpellNonMeleeDamage* log)
     //    data << float(log->GlanceChance);
     //    data << float(log->CrushChance);
     //}
+
+    sScriptMgr->OnSendSpellNonMeleeDamageLog(log, overkill < 0 ? 0 : overkill);
```
blinkysc, line 6756: Hook fires **after** `data` has been fully serialized from `*log` (lines 6707–6740) and immediately before `SendMessageToSet`. The signature takes a non-const `SpellNonMeleeDamage*`, but any mutation a subscriber makes to `*log` is silently dropped because the packet is already built.

Move the call above the packet construction (so mutations propagate) **or** change the parameter to `SpellNonMeleeDamage const* log` to remove the false promise of mutability. As written, it's a contract that doesn't hold.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -6916,6 +6942,9 @@ void Unit::SendAttackStateUpdate(CalcDamageInfo* damageInfo)
     data << uint32(tmpDamage[0] + tmpDamage[1]);                    // Full damage
     int32 overkill = tmpDamage[0] + tmpDamage[1] - damageInfo->target->GetHealth();
     data << uint32(overkill < 0 ? 0 : overkill);                    // Overkill
+
+    sScriptMgr->OnSendAttackStateUpdate(damageInfo, overkill < 0 ? 0 : overkill);
```
blinkysc, line 6946: This hook fires **between** packet writes — after `overkill` is serialized but before `count` and the per-damage-index loop on line 6948 onward. Mutations to fields already written are silently dropped; mutations to fields written later **do** take effect. The behavior depends on which field a subscriber touches relative to the call site, which is brittle and not something a reasonable script author would expect.

Fix: move the call above all of the `data << ...` writes, or const-qualify the parameter. Same shape of issue as `OnSendSpellNonMeleeDamageLog` above.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -2466,6 +2471,9 @@ void Unit::CalcAbsorbResist(DamageInfo& dmgInfo, bool Splited)
         tempAbsorb = currentAbsorb;
         absorbAurEff->GetBase()->CallScriptEffectAfterAbsorbHandlers(absorbAurEff, aurApp, dmgInfo, tempAbsorb);
 
+        if (currentAbsorb > 0)
+            sScriptMgr->OnDamageAbsorbed(dmgInfo, absorbAurEff->GetSpellInfo(), absorbAurEff->GetCaster(), currentAbsorb);
```
blinkysc: `OnDamageAbsorbed` is wired into the `SPELL_AURA_SCHOOL_ABSORB` block here and the `SPELL_AURA_MANA_SHIELD` block at line 2540 — but `CalcAbsorbResist` calls `dmgInfo.AbsorbDamage(...)` four more times in the split-damage paths (~2566, 2578, 2637, 2651) without firing the hook. A subscriber summing absorbed damage will undercount any encounter involving split-damage auras (paladin Blessing of Sacrifice, warlock Soul Link, hunter Misdirection-style splits, etc).

Either wire those four call sites or rename the hook so the partial coverage is explicit (`OnSchoolAbsorbApplied`?).

src/server/game/Spells/Auras/SpellAuraEffects.cpp
```diff
@@ -3152,8 +3152,13 @@ void AuraEffect::HandleAuraModSilence(AuraApplication const* aurApp, uint8 mode,
         for (uint32 i = CURRENT_MELEE_SPELL; i < CURRENT_MAX_SPELL; ++i)
             if (Spell* spell = target->GetCurrentSpell(CurrentSpellTypes(i)))
                 if (spell->m_spellInfo->PreventionType == SPELL_PREVENTION_TYPE_SILENCE)
+                {
+                    if (Unit* caster = GetCaster())
+                        sScriptMgr->OnSpellInterrupt(caster, target, GetSpellInfo()->Id, spell->m_spellInfo->Id);
```
blinkysc: The hook is inside `if (Unit* caster = GetCaster())`, but the next statement (`target->InterruptSpell(...)`) is outside the conditional. The interrupt happens unconditionally; the hook fires only when the silence aura's caster is still in world. Subscribers that count interrupts will undercount whenever the caster despawned, left the area, or died between aura application and this tick.

The `EffectInterruptCast` path at SpellEffects.cpp:3766 doesn't have this asymmetry, which makes the divergence here especially surprising. Either pass `nullptr` through and document that `interrupter` may be null, or pre-resolve a stable identity at aura-application time so the hook can always fire.

blinkysc (commented):
I had mcp create unit tests for each and this was some of its observations

Emyrk:
Just to comment, this is also missing any hooks that despawn entities without their deaths right now. Some boss fights have units just "disappear" from the client's perspective.

I'm not sure what this looks like on the server, but those events would also be helpful :+1:

scarecr0w12:
Addressed all five @blinkysc review comments plus @Emyrk's request in commit `74f16c320`. Summary of changes:

**1. `OnDealDamageShieldDamage` — hidden `AbsorbDamage` mutation removed**
Removed the `dmgInfo.AbsorbDamage()` sync hack. The hook now takes explicit post-`DealDamageMods` values:
```cpp
OnDealDamageShieldDamage(Unit* shieldOwner, Unit* attacker, SpellInfo const*, uint32 damage, uint32 absorb, uint32 overkill)
```
Scripts always see the values that were actually used for the packet and `DealDamage` call.

**2. `OnSendSpellNonMeleeDamageLog` — false mutability removed**
`SpellNonMeleeDamage*` → `SpellNonMeleeDamage const*`. Packet is already serialised at the call site; const makes the observation-only contract explicit.

**3. `OnSendAttackStateUpdate` — false mutability removed**
`CalcDamageInfo*` → `CalcDamageInfo const*`. Same rationale — hook fires mid-packet, const prevents misleading mutations.

**4. `OnDamageAbsorbed` → renamed `OnSchoolAbsorbApplied`**
Renamed hook and enum value (`UNITHOOK_ON_SCHOOL_ABSORB_APPLIED`) to make the partial coverage explicit. The hook only covers `SPELL_AURA_SCHOOL_ABSORB` and `SPELL_AURA_MANA_SHIELD` aura paths inside `CalcAbsorbResist`; split-damage redirections are not covered and the name no longer implies otherwise.

**5. `OnSpellInterrupt` (silence-aura path) — fires unconditionally**
Removed the `if (Unit* caster = GetCaster())` guard in `HandleAuraModSilence`. The hook now always fires, passing `GetCaster()` which may be `nullptr` when the silence caster left the world. Doc updated to document the nullable `interrupter` parameter.

**6. New hook: `AllCreatureScript::OnBeforeCreatureDespawn`** (addressing @Emyrk)
Fires from `Creature::DespawnOrUnsummon()` before the actual despawn, covering both `TempSummon::UnSummon()` (summons vanishing alive) and `Creature::ForcedDespawn()` (regular creatures removed without going through the death state). Scripts can use `creature->IsAlive()` to distinguish the client-visible "disappearing alive" case from post-death cleanup.

blinkysc:
Just curious what does perf reveal when you're grabbing the data from the hooks?

scarecr0w12:
> Just curious what does perf reveal when you're grabbing the data from the hooks?

If we're talking about straight performance impact, negligible. I see almost no increase in my percentiles on diff and run an avg of 1-2ms while the module we have grabs the data and uploads it.

Emyrk:
@blinkysc any updates on the review?

blinkysc:
> @blinkysc any updates on the review?

Been very busy with work. I would like to build something that saturates the hooks and see whats up with perf. Likely some of the other hooks could use const

Emyrk:
> > @blinkysc any updates on the review?
> 
> Been very busy with work. I would like to build something that saturates the hooks and see whats up with perf. Likely some of the other hooks could use const

I don't do any AZ development, but lmk if I can provide any support :+1:

src/server/game/Entities/Creature/Creature.cpp
```diff
@@ -2185,6 +2185,8 @@ void Creature::ForcedDespawn(Milliseconds timeMSToDespawn, Seconds forceRespawnT
 
 void Creature::DespawnOrUnsummon(Milliseconds msTimeToDespawn /*= 0ms*/, Seconds forcedRespawnTimer /*= 0s*/)
 {
+    sScriptMgr->OnBeforeCreatureDespawn(this);
```
blinkysc, line 2188: **Perf: this hook uses the `ExecuteScript` + `std::function` pattern, which is materially more expensive than the `CALL_ENABLED_HOOKS` path the other hooks in this PR use.**

`ScriptMgr::OnBeforeCreatureDespawn` constructs a `std::function` on every `DespawnOrUnsummon` call and then iterates the **entire, unfiltered** `AllCreatureScript` list, invoking the empty virtual on every registered script rather than only on subscribers. `DespawnOrUnsummon` fires in bursts (encounter resets, grid/instance unload, AoE-summon expiry).

I benchmarked the two dispatch patterns in isolation — a faithful replica of `ScriptMgrMacros.h` (`vector<vector<T*>>` `EnabledHooks` vs `ExecuteScript`), clang 19 `-O2`, dispatchers in a separate TU so the calls stay out-of-line like the real `ScriptMgr::OnXxx`, min of 5×300M iters:

| dispatch path | ns/call |
|---|---|
| `ExecuteScript`, **empty** list | **1.71** |
| `ExecuteScript`, 5 scripts | 7.45 |
| `ExecuteScript`, 20 scripts | 26.5 |
| filtered (`CALL_ENABLED_HOOKS`), **empty** | **0.76** |

Two takeaways: (1) even with **zero** subscribers it costs ~2.5× the filtered pattern, because the `std::function` is built unconditionally; (2) it scales linearly with the *total* number of `AllCreatureScript`s, not with subscribers (26 ns at 20 scripts). The other hooks added in this PR don't have this problem — they sit at ~0.6 ns when unused.

Suggestion: route this through an `EnabledHooks`-filtered dispatch like the other hooks in the PR, or drop it — most "creature about to despawn" needs are already covered by existing death/summon hooks.

_(Numbers are mechanism-level; in absolute terms all of this is sub-microsecond, so the broader send-log/absorb hooks are a non-issue perf-wise — this `ExecuteScript` outlier is the one worth changing.)_

blinkysc:
> > > @blinkysc any updates on the review?
> > 
> > 
> > Been very busy with work. I would like to build something that saturates the hooks and see whats up with perf. Likely some of the other hooks could use const
> 
> I don't do any AZ development, but lmk if I can provide any support 👍

No problem, just know these all in very hot paths, and auras/procs is essentially most the cycles in the server process so performance should be a big concern.

Emyrk:
@blinkys, @scarecr0w12, I fear this could eventually become too stale to consider valid anymore.

Is there someone we can find to push it over the edge into a merge? Personally I have such little experience with AZ core, I don't think I have the chops.

sudlud:
Went through the whole PR against current master.

**Verified fixed:** all five Copilot items and all five of @blinkysc's 2026-05-07 items hold up in the code, not just in the summaries.

**Still open:**

1. @blinkysc's perf thread on `OnBeforeCreatureDespawn`. Worth flagging that the suggested fix is not a one-liner: `AllCreatureScript` has no `EnabledHooks` infrastructure at all and all five of its existing dispatchers use `ExecuteScript`, so switching this one hook means converting the whole family first.

2. `OnBeforeCreatureDespawn` also misses most despawns. It only sits in `Creature::DespawnOrUnsummon()`, but `TempSummon::UnSummon()` is called directly from 41 sites including the entire `TempSummon::Update()` expiry ladder (TemporarySummon.cpp:72-194), which is exactly the "summon disappears alive" case @Emyrk asked about. And `DespawnOrUnsummon(5000ms)` fires the hook 5s before the creature is gone. `AllCreatureScript::OnCreatureRemoveWorld` already covers real removals on every path, so dropping this hook may be cleanest.

3. Hot path cost the const pass missed: `absorbAurEff->GetCaster()` at both `OnSchoolAbsorbApplied` call sites. `CALL_ENABLED_HOOKS` guards the loop body, not argument evaluation, so `Aura::GetCaster()` (GUID compare, `GetApplicationOfTarget` map lookup, possible `ObjectAccessor::GetUnit`) runs on every absorb event with zero subscribers. `GetCasterGUID()` is a field read.

4. Const is still inconsistent. `OnSendSpellNonMeleeReflectLog` takes `SpellNonMeleeDamage*` while `OnSendSpellNonMeleeDamageLog` takes `SpellNonMeleeDamage const*`, and the reflect one fires before the packet is built so mutations do land. Same for `OnSendPeriodicAuraLog`. `OnSchoolAbsorbApplied` takes `DamageInfo&` mid-iteration over the absorb list, so a script can silently change damage, and the doc calls it "incoming damage context".

5. `OnSpellDispel`, `OnSpellInterrupt` and `OnAuraApplicationClientUpdate` are per-unit events sitting on `GlobalScript` while every other unit event here is on `UnitScript`. Hook classes are module API, so moving them after merge is breaking.

6. `OnEnvironmentalDamage` should be `OnPlayerEnvironmentalDamage`, every other PlayerScript hook carries the prefix. `PLAYERHOOK_ON_ENVIRONMENTAL_DAMAGE` is also inserted mid-enum instead of appended.

7. `#include "Player.h"` in PlayerScript.h is only there for `EnviromentalDamage`. PlayerScript.h reaches nearly every TU via `AllScriptsObjects.h` and `ScriptPCH.h`, so that is a 3084 line header added broadly. Moving the enum to SharedDefines.h avoids it.

The branch also conflicts with master (GlobalScript.cpp, PlayerScript.cpp, UnitScript.cpp, all from the `const char*` to `char const*` rename). Maintainer edits do not work on org-owned forks so I could not push to your branch, but I resolved it here and it fast-forwards cleanly:

```
git pull https://github.com/sudlud/azerothcore-wotlk.git pr25641-sync-master
```

Finally, the description still describes the original two-commit version rather than the 18 hooks now in it.

scarecr0w12:
Sorry guys, I havn't been able to work on this, Life issues keeping me from getting the outstanding sorted.

sudlud:
> Sorry guys, I havn't been able to work on this, Life issues keeping me from getting the outstanding sorted.

no worries, life always comes first. just pick it up whenever you have time

sudlud:
Went back over the hook call sites in more detail. A few correctness issues on top of my earlier comment.

**1. `OnAuraApplicationClientUpdate` can invalidate a live iterator**

`Unit::_UpdateSpells` walks `m_visibleAuras` with a raw iterator and calls `ClientUpdate()` (Unit.cpp:4037-4039). The hook is now the first statement of `AuraApplication::ClientUpdate`. `RemoveVisibleAura` is a plain `m_visibleAuras.erase(slot)`, so a module that removes an aura from inside the hook erases the node the loop is standing on and `++itr` walks freed memory. Nothing reachable from `ClientUpdate` could touch that map before this PR.

It also fires from `AuraApplication::_Remove` right after `RemoveVisibleAura(slot)`, so modules get an application that is mid-teardown.

**2. `OnSchoolAbsorbApplied` passes a null-capable pointer documented as non-null**

`absorbAurEff->GetCaster()` goes through `Aura::GetCaster()`, which returns `nullptr` when the caster is offline or on another map. The doc says `@param absorbCaster Unit that applied the absorb aura` with no nullability note. Shield someone, log out, shield keeps absorbing, module dereferences null. Passing `GetCasterGUID()` instead fixes this and the hot path cost from my earlier comment in one go.

Same call sites are followed by `absorbAurEff->GetAmount()` / `SetAmount()` / `GetBase()->Remove()`. The effect list is copied because handlers mutate it, but the copy holds raw `AuraEffect*`, so a module removing the shield from inside the hook leaves that pointer dangling for the next three statements.

**3. Two hooks fire when the thing they report did not happen**

`OnSendPeriodicAuraLog` is the first statement of `Unit::SendPeriodicAuraLog`, but the `switch` has a `default:` branch that logs an error and returns without `SendMessageToSet`. While you are there, that switch only handles `SPELL_AURA_PERIODIC_DAMAGE` and `SPELL_AURA_PERIODIC_MANA_LEECH`, so the doc claim that it covers healing and energize ticks does not hold.

`OnSpellInterrupt` on the silence path fires before `InterruptSpell(type, false)`, which does nothing when the spell is `SPELL_STATE_DELAYED` or `!IsInterruptable()`. A Frostbolt already in flight when a silence lands gets logged as interrupted and then resolves normally. `EffectInterruptCast` pre-checks the state but still not `IsInterruptable()`.

**4. Two more things on `OnBeforeCreatureDespawn`**

Beyond the coverage gap: on the delayed non-summon path `ForcedDespawnDelayEvent::Execute` calls `DespawnOrUnsummon(0ms)` again, so the hook fires twice for one despawn. And `DespawnOnEvade` fires it for a plain creature but not for a `TempSummon`, same call, two behaviours.

**5. Payload accuracy**

- `OnAuraApplicationClientUpdate` with `remove == false` does not mean "applied". `ClientUpdate` also runs on every stack, charge and duration refresh, so anything counting it over-counts stacking DoTs. It also never fires for auras without a visible slot.
- `OnEnvironmentalDamage` passes the pre-mitigation value, while the packet carries post-mitigation damage plus absorb and resist. None of those three are exposed, so for a combat log module this is the one hook whose payload cannot be reconciled with what the client shows.
- `OnSpellDispel` fires before `RemoveAurasDueToSpellBy*`, which is useful, but the doc does not say so and it does not pass the `Aura*`.
- `OnSpellExecuteLogSummonObject` drops `effIndex` and fires when the log entry is buffered, not when `SendLogExecute` sends it.

**Two corrections to my earlier comment**

Converting `AllCreatureScript` to `EnabledHooks` is more work than I said. It is 13 virtuals dispatched from `AllCreatureScript.cpp` and `CreatureScript.cpp`, not five.

And I lumped all five `GlobalScript` hooks together. `OnSpellDispel`, `OnSpellInterrupt` and `OnAuraApplicationClientUpdate` are unit events and fit `UnitScript`, but `OnSpellSendSpellGo` and `OnSpellExecuteLogSummonObject` take `Spell*` and match `AllSpellScript`, which already has `OnSpellCast`, `OnSpellPrepare` and `OnSpellCastCancel`. It should be a split rather than one destination.

Two smaller things: the title still says "spell and dispel hooks" for what is now 18 hooks spanning combat log, absorb, energize, periodic, environmental damage and creature despawn, and this needs the `Requires WIKI Update` label.

sudlud:
closing this for now. please feel free to reopen anytime in the future addressing the review comments. thanks!

Issue #13842, opened 2022-11-16
Quest: Draconis Gastritis can't be completed


whiteblue8823:
### Current Behaviour

The quest is: Draconis Gastritis id:11280

You can accept the quest from the NPC, but when you use the item, the dragon comes and just attacks and doesn't allow you to complete the quest. I've tested it several times.

### Expected Blizzlike Behaviour

after using the quest item, the target dragon should react with the item been used and let the quest be done.

### Steps to reproduce the problem

1. take the quest
2. use the quest item
3. bug happen

### AC rev. hash/commit

AzerothCore rev. 5cf7d3d198f6 2022-11-13 18:41:10 -0300 (master branch) (Win64, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1s  1 Nov 2022 (library: OpenSSL 1.1.1s  1 Nov 2022)
Using Boost version: 1.79.0
Using MySQL version: 80031
Using CMake version: 3.24.3
Compiled on: Windows 10.0.22000
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: 1, getHeight: 1, indoorCheck: 1
MMAPs status: Enabled
maps directory located in C:/App/AzerothCore/RelWithDebInfo/Data/maps. Total size: 291014951 bytes
vmaps directory located in C:/App/AzerothCore/RelWithDebInfo/Data/vmaps. Total size: 658130721 bytes
mmaps directory located in C:/App/AzerothCore/RelWithDebInfo/Data/mmaps. Total size: 2320900676 bytes
Using enUS DBC Locale as default. All available DBC locales: enUS
Using World DB: ACDB 335.8-dev
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
No modules enabled
AC> Update time diff: 17. Players

### Operating system

Windows 11

### Custom changes or Modules

None

zylok520:
The normal situation is that after using the task item, the dragon will go down to the ground, eat the food, and then the changes will be made to complete the task; now the dragon has come down, but there is nothing happened.

saligin:
I used this method, and it worked for me!

```sql
UPDATE creature_template SET AIName='SmartAI' WHERE entry IN (23689,24170);

DELETE FROM smart_scripts WHERE entryorguid IN (23689,24170) AND source_type=0;

[...]

PR #25672, opened 2026-05-01
fix(scripts): [Howling Fjord] Quest 11280 Draconis Gastritis - add NP…

srikar6152:
## Fix: [Howling Fjord] Quest 11280 - Draconis Gastritis

**Fixes:** #13842

### Problem
NPC 24170 (Draconis Gastritis Bunny), summoned by item 33441 (Tillinghast's Plagued Meat 
via spell 43172), had no C++ script. The Proto-Drake (NPC 23689) never reacted to 
the meat, so the quest objective was never fulfilled.

### Fix
- Added `npc_draconis_gastritis_bunny` AI to `zone_howling_fjord.cpp`
- The bunny polls every 1s for a nearby Proto-Drake (NPC 23689) within 30 yards
- When found: drake faces the meat, performs eat emote, player receives KilledMonsterCredit
- SQL assigns ScriptName to NPC 24170

### Testing
- Accept quest from Plaguebringer Tillinghast
- Use Tillinghast's Plagued Meat in the Ember Clutch
- Proto-Drake reacts, quest objective completes

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1746123456789.sql.sql
+++ b/data/sql/updates/pending_db_world/rev_1746123456789.sql.sql
@@ -1,0 +1,6 @@
+-- [Howling Fjord] Quest 11280: Draconis Gastritis
+-- Assign C++ script to NPC 24170 (Draconis Gastritis Bunny)
+UPDATE `creature_template`
+SET `AIName` = '',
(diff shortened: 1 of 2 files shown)


Rorschach91:
This bug has been solved (the report is closed as completed). Otherwise this is not the right approach to fix it.

Issue #22331, opened 2025-06-19
[Warrior] Charge is not climbing terrain in wsg


Apathyxia:
### Triage

Original CC issue: https://github.com/chromiecraft/chromiecraft/issues/8087

This should be possible as a druid in bear form too when using Feral Charge - Bear.
Additional source from original WotLK that this was indeed possible (as a warrior): https://youtu.be/HrGcvxggyrM?t=212
This should be possible on both graveyards in WSG. 

### Current Behaviour

You cannot charge people in spawn zone in WSG with charge

### Expected Blizzlike Behaviour

Should be able to

### Source

https://wowwiki-archive.fandom.com/wiki/Charge

![Image](https://github.com/user-attachments/assets/b5950448-2fd2-436f-b1ab-864c43d5870d)

https://youtu.be/Z4174ce_62M?t=94

### Steps to reproduce the problem

* 

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/a20576dc8c4916791e7830f18ef6c3ca528e29a9

### Operating system

Ubuntu 22.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-costumes](https://github.com/azerothcore/mod-costumes)
- [mod-cta-switch](https://github.com/azerothcore/mod-cta-switch)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-dmf-switch](https://github.com/azerothcore/mod-dmf-switch)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
[...]

PR #25720, opened 2026-05-05
fix(Core/MMaps): fix Blade's Edge Arena falling/edge pathing (bump mmap version to 20)

### LINKED ISSUE #25126: [BEM arena] Lots of issues with pathing / behaviour on edges
Later, reported also on ChromieCraft repository with some extra footage: https://github.com/chromiecraft/chromiecraft/issues/9433

### Current Behaviour

Currently, in Blade's Edge Arena, a lot of pet pathing is horrible, making classes like warlock unplayable. 

See my video from live ChromieCraft testing:
https://www.youtube.com/watch?v=Uj_vg37KYhU

Few issues:
- Pets don't want to move around easily
- Pets dropping down through the platforms
- Fear / horrify / disorient effects can drop you from the platform / rope
- Charging opponent on the platform may drop you down or even through the ground

### Expected Behaviour

- You should not be randomly dropping through the platform when charging
- Fear/horrify/disorient should not make you fall off the platform
- Pets should be able to pursue their targets normally

### Source

These issues are almost exclusive to this map. On other maps the behaviour is normal. 

1. Have a druid and a warlock
2. Get into BEM arena (might need 4 characters)
3. Check the video above about positioning, and then sending pet / charging / fearing / death coiling around.
4. Observe "broken" map textures that result in weird behaviours

### Extra Notes

Also tested on local quickly to verify behaviour

### AC rev. hash/commit

AzerothCore rev. 8f5900b36b2c 2026-03-19 08:33:38 -0300 (master branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 11 x64

### Custom changes or Modules

_No response_

walkline:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/25126

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


VG-prog:
Tested, it fixes the addressed issue.

Nyeriah:
Hey 👋 

Since we are generating new maps would it be possible to include this one? https://github.com/azerothcore/azerothcore-wotlk/issues/22331

walkline:
Hey @Nyeriah. I don't think that [this](https://github.com/azerothcore/azerothcore-wotlk/issues/22331) issue can or should be fixed with mmaps. It's most likely an issue with the charge implementation.

src/tools/mmaps_generator/Config.cpp
```diff
@@ -225,8 +235,10 @@ namespace MMAP
                     override.walkableHeight = mapNode["walkableHeight"].get_value<int>();
                 if (mapNode.contains("walkableClimb"))
                     override.walkableClimb = mapNode["walkableClimb"].get_value<int>();
-                if (mapNode.contains("vertexPerMapEdge"))
-                    override.vertexPerMapEdge = mapNode["vertexPerMapEdge"].get_value<int>();
+                if (mapNode.contains("verticesPerMapEdge"))
+                    override.vertexPerMapEdge = mapNode["verticesPerMapEdge"].get_value<int>();
+                if (mapNode.contains("verticesPerTileEdge"))
+                    override.vertexPerMapEdge = mapNode["verticesPerTileEdge"].get_value<int>();
```
sudlud: This assigns the tile-edge value into `vertexPerMapEdge`; it should be `override.vertexPerTileEdge`. `GetConfigForTile` reads `m->vertexPerTileEdge` (Config.cpp:112), so as written the tile-edge override is dropped and the map-edge value gets clobbered instead.

src/tools/mmaps_generator/Config.cpp
```diff
@@ -191,6 +191,16 @@ namespace MMAP
         tryBoolean(mmapsNode, "skipBattlegrounds", _skipBattlegrounds);
         tryBoolean(mmapsNode, "debugOutput", _debugOutput);
 
+        if (mmapsNode.contains("offmeshConnections") &&
+    mmapsNode["offmeshConnections"].is_sequence())
```
sudlud: Align the continuation with the start of the condition instead of 4 spaces.

src/tools/mmaps_generator/Config.cpp
```diff
@@ -202,8 +212,8 @@ namespace MMAP
         tryInt(mmapsNode, "walkableHeight", _global.walkableHeight);
         tryInt(mmapsNode, "walkableClimb", _global.walkableClimb);
         tryInt(mmapsNode, "walkableRadius", _global.walkableRadius);
-        tryInt(mmapsNode, "vertexPerMapEdge", _global.vertexPerMapEdge);
-        tryInt(mmapsNode, "vertexPerTileEdge", _global.vertexPerTileEdge);
+        tryInt(mmapsNode, "verticesPerMapEdge", _global.vertexPerMapEdge);
```
sudlud, line 214: Renaming `vertexPer*Edge` to `verticesPer*Edge` breaks existing user configs that set these keys, and it's unrelated to the BEM fix. Could be its own PR, or keep a backward-compatible lookup of the old key. This is also where the override bug below slipped in.

src/server/game/Movement/MovementGenerators/PathGenerator.cpp
```diff
@@ -24,6 +24,105 @@
 #include "Map.h"
 #include "Metric.h"
 
+// Blades Edge Arena Ropes normalization
+namespace
+{
+    constexpr float BLADE_EDGE_ROPE_SNAP_DIST = 1.5f;
+    constexpr float BLADE_EDGE_ROPE_SNAP_DIST2 = BLADE_EDGE_ROPE_SNAP_DIST * BLADE_EDGE_ROPE_SNAP_DIST;
+    constexpr float BLADE_EDGE_ROPE_PI = 3.14159265358979323846f;
```
sudlud: `M_PI` is already used across `src/server/game/Movement/`; prefer it over a local pi constant.

src/server/game/Movement/MovementGenerators/PathGenerator.cpp
```diff
@@ -24,6 +24,105 @@
...
+
+    struct BladeEdgeArenaRope
+    {
+        G3D::Vector3 Start;
+        G3D::Vector3 End;
+        float Sag;
+    };
+
+    static const std::array<BladeEdgeArenaRope, 2> BladeEdgeArenaRopes =
```
sudlud, line 40: These ropes use different coordinates than the off-mesh connections in `mmaps-config.yaml`. Same rope: this one starts at (6234.32, 256.30, 11.002), the off-mesh entry at (6234.47, 256.56, 11.064), and the off-mesh Z sits above this curve before the sag is even subtracted. A snapped point can land below the off-mesh link. Worth reconciling the two sets, or at least confirming the Z lines up in-game.

src/tools/mmaps_generator/Config.h
```diff
@@ -76,6 +76,8 @@ namespace MMAP
         std::string MMapsPath() const { return (_dataDir / "mmaps").string(); }
         std::string DataDirPath() const { return _dataDir.string(); }
 
+        std::vector<std::string> OffMeshConnections() const { return _offmeshConnection; }
```
sudlud: Returns the vector by value, copied on each tile build. Return `std::vector<std::string> const&`.

src/tools/mmaps_generator/Config.h
```diff
@@ -153,6 +155,8 @@ namespace MMAP
         bool _debugOutput;
 
         std::filesystem::path _dataDir;
+
+        std::vector<std::string> _offmeshConnection;
```
sudlud: Holds multiple connections; `_offmeshConnections` (plural) reads better.

sudlud (commented):
Direction looks good and VG-prog confirmed the fix. Blocking item: the tile-edge override parsing writes into the wrong field (inline). Also worth addressing: the rope coordinates in NormalizePath don't match the off-mesh entries in the yaml and the Z sits above the snapped curve, so a snap could drop a unit below the link (inline). Lastly the verticesPer* key rename is an unrelated breaking change for existing configs and would be cleaner split out.

src/server/game/Movement/MovementGenerators/PathGenerator.cpp
walkline: I don't think it's that big of a deal. For an off-mesh connection, the connection should start a little bit before the actual rope to create a valid connection. As for the normalization, I wanted a point a little after the start of the rope. But anyway, I got those points empirically and tested them ingame.

src/tools/mmaps_generator/Config.cpp
walkline: Those are leftovers from the initial implementation of the config system. Currently, the config file doesn't match the logic that parses it. Since the config examples should be the source of truth for users, this fixes the issue on the logic side.

src/tools/mmaps_generator/Config.cpp
```diff
@@ -202,8 +211,8 @@ namespace MMAP
...
+        tryInt(mmapsNode, "verticesPerTileEdge", _global.vertexPerTileEdge);
```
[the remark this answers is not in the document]
walkline: Already addressed that comment.

walkline:
@sogladev thanks for the review! Most comments should be addressed now.

Kitzunu (approved):

PR #24641, opened 2026-02-07
feat(Core/RBAC): Role Based Access Controls

blinkysc:
## Changes Proposed:
This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).

Ports TrinityCore's Role-Based Access Control (RBAC) permission system to AzerothCore, replacing the legacy security-level-only approach for command and gameplay permission gating.

**Core changes:**
- New `RBAC.h` / `RBAC.cpp` implementing `RBACPermission`, `RBACData` with grant/deny/revoke/expand logic
- `AccountMgr` extended with `LoadRBAC()` to load permissions, linked permissions, and defaults at startup
- `WorldSession` gains `HasPermission()`, `LoadPermissions()`, `InvalidateRBACData()` for live RBAC checks
- `ChatCommand.cpp` updated: `IsInvokerVisible()` routes RBAC perms (>=200) through `HasPermission()`, legacy security levels (0-4) through `IsAvailable()`; DB `command.security` column no longer overrides RBAC perms
- Removed legacy `AllowPlayerCommands` config in favor of RBAC perm 33 (matching TrinityCore)
- Fixed crash in `RevokePermission()` when called with invalid permission ID

**Permissions implemented (53 gameplay + 700+ command):**
- Role hierarchy: Admin (192) → GM (193) → Mod (194) → Player (195)
- Command roles: Admin Commands (196), GM Commands (197), Mod Commands (198), Player Commands (199)
- All 53 core gameplay permissions (instant logout, BG/arena join, cross-faction interaction, GM features, etc.)
- All command scripts (`cs_*.cpp`) updated with proper RBAC permission constants

**New command:** `.rbac account grant/deny/revoke/list` for live permission management

**Module RBAC permission system:**
- New `module_rbac_permissions` table lets modules register permissions with local IDs
- Auto-incrementing `global_id` (starting at 100000) prevents collisions between modules and core
- `AccountMgr::GetModulePermission(module, localId)` resolves local→global at runtime
- Syncs module permissions into `rbac_permissions` for FK compatibility

**SQL migrations:** `rbac_permissions`, `rbac_linked_permissions`, `rbac_default_permissions`, `rbac_account_permissions`, `module_rbac_permissions` tables with full permission/role data

### AI-assisted Pull Requests

> [!IMPORTANT]
[...]

PR #25723, opened 2026-05-05
fix(Core/Trade): fix use-after-free in GM trade log on stack merge

### LINKED ISSUE #25716: Unable to trade stacking items
### Current Behaviour

Trading another player an item that would combine into a stack in their inventory is not possible. The trade window stays open, and the trade just does not go through.

### Expected Behaviour

Items should be able to be traded between players, even if it would go onto a recipient's stack.

### Source

_No response_

1. Make sure the trade recipient has at least one of a stackable item
2. Open a trade window and attempt to trade the recipient such item

### Extra Notes

I had GPT-5.4 look into the issue. This is its summary:

>WorldSession::HandleAcceptTradeOpcode, the trade code stores the traded items in local Item* arrays (myItems / hisItems) and then completes the transfer with moveItems(...). If the received item merges into an existing stack, the original Item* can stop being valid during that move. The problem is that later code still uses those same cached pointers instead of only using data saved before the move.

>This was already addressed by commit f1a2b8304 (fix(Core/Trade): fix crash from use-after-free in trade logging), which moved item logging before moveItems(...) because the pointers could become invalid afterward. The likely regression is commit 779331fc4 (feat(Core/RBAC): Role Based Access Controls), which added new GM trade logging after moveItems(...) and dereferences myItems[i] / hisItems[i] again. That can lead to stale-pointer reads after a merge, which explains the trade failure and, when the GM logging path is hit, the crash. 

I tested between two players with the main AC repo. However, if this trade is attempted between a player and a bot, it actually crashes the server.

### AC rev. hash/commit

AzerothCore rev. 0f6755b0303f 2026-05-05 00:01:19 -0500 (master branch) (Win64, RelWithDebInfo, Static)
Connected players: 2. Characters in world: 2.
Connection peak: 2.
Server uptime: 1 minute(s) 45 second(s)
|- Mean: 2ms
|- Median: 2ms
|- Percentiles (95, 99, max): 7ms, 8ms, 9ms
Using SSL version: OpenSSL 3.4.0 22 Oct 2024 (library: OpenSSL 3.4.0 22 Oct 2024)
Using Boost version: 1.81.0
Using CMake version: 3.31.4
Using MySQL version: 80404
Found MySQL Executable: C:/Program Files/MySQL/MySQL Server 8.4/bin/mysql.exe
Compiled on: Windows 10.0.26200
MMAPs status: Enabled
maps directory located in C:/build-azerothcore/bin/RelWithDebInfo/data/maps. Total size: 291014951 bytes
vmaps directory located in C:/build-azerothcore/bin/RelWithDebInfo/data/vmaps. Total size: 657797442 bytes
mmaps directory located in C:/build-azerothcore/bin/RelWithDebInfo/data/mmaps. Total size: 2192760616 bytes
Default DBC locale: enUS.
Using World DB: ACDB 335.16-dev
Latest LoginDatabase update: acore_cms_subscriptions.sql
Latest CharacterDatabase update: u_MIB_2024_04_18_00.sql
Latest WorldDatabase update: zz_optional_vanilla_transports.sql
LoginDatabase queue size: 0
WorldDatabase queue size: 0

### Operating system

Windows 11 x64

### Custom changes or Modules

None

blinkysc:
## Changes Proposed:

The GM trade logging block in `WorldSession::HandleAcceptTradeOpcode` (added by the RBAC PR #24641) dereferences `myItems[i]` / `hisItems[i]` *after* `moveItems()` runs. When a traded item merges into an existing stack on the recipient side, `Player::MoveItemToInventory` destroys the source `Item`, leaving those pointers dangling. Any account with `rbac::RBAC_PERM_LOG_GM_TRADE` (granted by Moderator role 194 and inherited by Administrator) hits a use-after-free, which manifests as a crash in some setups and as the trade window hanging in others.

This is the same class of bug that #25452 already fixed for the trade-completion log line. This PR extends the existing pre-`moveItems()` cache loop to also capture each traded item's name, entry, and count into a small `TradedItemInfo` struct, and the GM logging blocks now read from that struct instead of from `myItems[i]` / `hisItems[i]`.

### AI-assisted Pull Requests

> [!IMPORTANT]
>

Tool used: Claude Code with AzerothMCP.

## Issues Addressed:
- Closes #25716

## SOURCE:

Reproduction steps and root-cause confirmation provided in #25716 by @TheSCREWEDSoftware (revoking RBAC perm 11 makes the crash disappear, which exactly matches this code path).

## Tests Performed:

Verified locally with two Administrator-level accounts (both inherit `RBAC_PERM_LOG_GM_TRADE`, so both sides hit the patched logging path). Trade of a 2-stack of Frostweave Cloth (33470) into an existing 18-stack on the recipient now completes cleanly: recipient ends with a single 20-stack, sender's stack is gone, and the GM trade log emits with the correct pre-merge `Count: 2` value. No crash; server uptime preserved.

## How to Test the Changes:

1. Two accounts where at least one has `RBAC_PERM_LOG_GM_TRADE` (granted by Moderator role 194 / inherited by Administrator).
2. `.additem <gm_char> 33470 2` and `.additem <other_char> 33470 18` (or any stackable item the recipient already partially holds).
3. Open a trade from the GM character to the other, drop the 2-stack into a trade slot, both Accept.
4. Pre-fix: server crashes (or trade window hangs). Post-fix: trade completes, recipient has a merged 20-stack, GM trade log line is emitted with the correct pre-merge count.

- [ ]

### DIFF
--- a/src/server/game/Handlers/TradeHandler.cpp
+++ b/src/server/game/Handlers/TradeHandler.cpp
@@ -459,14 +459,35 @@ void WorldSession::HandleAcceptTradeOpcode(WorldPacket& /*recvPacket*/)
             return;
         }
 
-        // log traded items before moving (pointers become invalid after moveItems)
+        // cache traded item info before moving (pointers become invalid after moveItems if items merge into existing stacks)
+        struct TradedItemInfo
+        {
+            bool present = false;
+            std::string name;
+            uint32 entry = 0;
+            uint32 count = 0;
+        };
+        TradedItemInfo myItemsInfo[TRADE_SLOT_TRADED_COUNT];
+        TradedItemInfo hisItemsInfo[TRADE_SLOT_TRADED_COUNT];
         std::string myItemsStr, hisItemsStr;
         for (uint8 i = 0; i < TRADE_SLOT_TRADED_COUNT; ++i)
         {
             if (myItems[i])
-                myItemsStr += Acore::StringFormat("{} (Entry:{}) x{}, ", myItems[i]->GetTemplate()->Name1, myItems[i]->GetEntry(), myItems[i]->GetCount());
+            {
+                myItemsInfo[i].present = true;
+                myItemsInfo[i].name = myItems[i]->GetTemplate()->Name1;
+                myItemsInfo[i].entry = myItems[i]->GetEntry();
+                myItemsInfo[i].count = myItems[i]->GetCount();
+                myItemsStr += Acore::StringFormat("{} (Entry:{}) x{}, ", myItemsInfo[i].name, myItemsInfo[i].entry, myItemsInfo[i].count);
+            }
             if (hisItems[i])
-                hisItemsStr += Acore::StringFormat("{} (Entry:{}) x{}, ", hisItems[i]->GetTemplate()->Name1, hisItems[i]->GetEntry(), hisItems[i]->GetCount());
+            {
+                hisItemsInfo[i].present = true;
+                hisItemsInfo[i].name = hisItems[i]->GetTemplate()->Name1;
+                hisItemsInfo[i].entry = hisItems[i]->GetEntry();
+                hisItemsInfo[i].count = hisItems[i]->GetCount();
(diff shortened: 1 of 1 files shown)


Nyeriah (approved):

TheSCREWEDSoftware (approved):
No longer crashes!

PR #25867, opened 2026-05-16
RFC: Observability v1

espkk:
This is RFC, not to be merged

### DIFF
--- a/docs/observability/rfc.md
+++ b/docs/observability/rfc.md
@@ -125,26 +125,38 @@ Label values should be emitted safely for Prometheus text exposition and should 
 
 ## Producer API Direction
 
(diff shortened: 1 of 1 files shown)


heyitsbench:
CC: @Takenbacon

Takenbacon:
Seems good to me. I was personally more of a fan of victoriametrics but prometheus is overall more widely used and supported.

espkk:
> Seems good to me. I was personally more of a fan of victoriametrics but prometheus is overall more widely used and supported.

Sounds like we can switch later if needed. It's almost a drop-in replacement as far as I understand

espkk:
implemented in #25898

PR #25885, opened 2026-05-17
perf(Core/Unit): Replace red-black tree with flat_multimap for aura containers

blinkysc:
## Changes Proposed:

This PR proposes changes to:

Replaces three `std::multimap` typedefs (`AuraMap`, `AuraApplicationMap`, `AuraStateAurasMap`) with `boost::container::flat_multimap` in `Unit.h`.

`std::multimap` uses a red-black tree internally — each element is a separately heap-allocated node with left/right/parent pointers, which scatters aura data across memory and causes frequent cache misses during iteration. `boost::container::flat_multimap` stores elements contiguously in a sorted vector, giving much better cache locality for the iteration-heavy aura update loops that run every server tick for every unit.

The tradeoff is that `flat_multimap` invalidates all iterators on insert/erase (unlike red-black trees which only invalidate the erased node's iterator). The `_UpdateSpells` loop previously relied on a persistent `m_auraUpdateIterator` — this is replaced with a snapshot pattern (member scratch buffer, no per-tick allocation). Direct erase loops use reset-to-begin/lower_bound patterns.

Review of the iterator contracts additionally surfaced and fixed (second commit):
- `Unit::RemoveAura(iterator&)`: `Aura::Remove` runs **after** `_UnapplyAura` reset the caller's iterator and can cascade (linked spells, procs) into further `m_appliedAuras` mutations, leaving the caller's iterator dangling. It is now reset again after the cascades. The bug is reproduced by a dedicated ASAN test (`DISABLED_StaleIteratorAfterCascade_AsanOnly` — deliberate UB, excluded from normal runs): `./unit_tests --gtest_also_run_disabled_tests --gtest_filter='*StaleIteratorAfterCascade*'` under `-fsanitize=address` reports a deterministic heap-use-after-free.
- Both `RemoveAppliedAuras` overloads skip applications already mid-removal — `RemoveAura` no-ops on those without advancing and the loops would never terminate.

**Measured impact** (microbenchmark over aura-shaped workloads — realistic aura-count distribution, deliberately fragmented heap, identical work checksums on both sides):
- tick iteration (`_UpdateSpells` shape, including the snapshot copy): **~6-7× faster**
- `equal_range` lookups (HasAura-style): ~1.3-1.5× faster
- apply/expire churn: ~2× faster (node alloc/free costs more than flat memmove at aura sizes)
- mixed realistic workload where churn interleaves with iteration: **~8× faster** — node containers degrade as the heap ages, flat storage is immune

Whole-server impact scales with the aura share of total CPU (~7% overall reduction was observed profiling with 5000 bots under load on earlier revisions of this change).

### AI-assisted Pull Requests

> [!IMPORTANT]
>

- [x] AI tools were used partially in preparing this pull request. **Tools used: Claude Code with azerothMCP**

## Issues Addressed:

Performance optimization — no linked issue.

## SOURCE:

- [x] Profiling with `perf record`/`perf script` before and after the change under load

## Tests Performed:

This PR has been:

Microbenchmarked (numbers above) and tested in-game. 15 unit tests added covering the container iteration patterns, including an ASAN-only crash repro for the iterator-cascade bug and its fixed-contract counterpart.

## How to Test the Changes:

1. Build with `-DBUILD_TESTING=ON` and run `./unit_tests --gtest_filter="FlatMultimapAuraPattern*"` — all 14 enabled tests should pass (one additional test is ASAN-only, see above)
2. Log in and verify auras apply/expire/remove correctly (buffs, debuffs, area auras)
3. Test aura-heavy scenarios: multiple players in combat, AoE spells, channeled spells with caster leaving map
4. Verify no crashes or aura-related issues during extended play


Nefertumm:
Really interesting read, nicely done. Any plans on using the standard implementation of flat_multimap instead of the boost one whenever the core supports C++23?  IIRC it _could_ bring even more performance because of the way it's implemented, using SoA instead of AoS, although it would require a little bit more groundwork.

blinkysc:
> Really interesting read, nicely done. Any plans on using the standard implementation of flat_multimap instead of the boost one whenever the core supports C++23? IIRC it _could_ bring even more performance because of the way it's implemented, using SoA instead of AoS, although it would require a little bit more groundwork.

I think so, I’m not really familiar with the difference but if perf reveals performance differences I think it’s worth it

Nefertumm (approved):

TheSCREWEDSoftware:
For the few class, item and quest buffs they seemed to work correctly for thoes that expired (that should), phasing auras from wrath areas also seemed to properly be added or removed depending on what quests i had completed before (no change here).

I tried warlock and mage aoe spells, didn't see any issue here, not much I can figure out during testing, perfomance or crashes i don't see a difference.

PkllonG:
```.go  160 442 -6 553``` Crash

Kitzunu:
@PkllonG https://github.com/azerothcore/azerothcore-wotlk/pull/26644 should fix it

PR #25904, opened 2026-05-19
fix(Scripts/PitOfSaron): Fix Scourgelord Tyrannus bug combat after Ick's death

EricksOliveira:
After Ick's death, the Scourgelord Tyrannus fly event begins, but players were previously stuck in combat. Now they are no longer stuck in combat.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Enter the Pit of Saron
2. Kill Boss Ick
3. Note that the player is no longer locked into combat.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Northrend/FrozenHalls/PitOfSaron/boss_scourgelord_tyrannus.cpp
+++ b/src/server/scripts/Northrend/FrozenHalls/PitOfSaron/boss_scourgelord_tyrannus.cpp
@@ -117,7 +117,7 @@ public:
                 me->RemoveAllAuras();
                 me->RemoveUnitFlag(UNIT_FLAG_NON_ATTACKABLE);
                 me->SetImmuneToPC(false);
-                DoZoneInCombat();
+                me->SetInCombatWithZone();
                 me->CastSpell(me, 43979, true);
                 Talk(SAY_AGGRO);
                 events.Reset();


blinkysc:
I actually think this isn't right this just calls DoZineInCombat anyways? But would need to dig deeper. I won't have much time until the weekend though

EricksOliveira:
I tested it this way and it worked for me. With DoZoneInCombat(), everyone would get stuck in combat and wouldn't be able to get out.

blinkysc:
https://github.com/azerothcore/azerothcore-wotlk/blob/e944605b9b3e5325d90b6ee26eba513f32ec66aa/src/server/game/Entities/Creature/Creature.cpp#L2850

sudlud:
Thanks for digging into this, @EricksOliveira.

Unfortunately this change is a no-op. `Creature::SetInCombatWithZone()` just calls `AI()->DoZoneInCombat()`, and since `boss_tyrannusAI` doesn't override `DoZoneInCombat`, the new line resolves to the exact same `CreatureAI::DoZoneInCombat()` with the same default arguments as the original. So the compiled behavior is identical and it can't be what fixed the combat issue in your test (that was likely a clean restart or some other state difference). This was also what @blinkysc pointed out.

The real "stuck in combat after Ick's death" cause is somewhere else in the Pit of Saron flow, not in this line, so I'm going to close this for now.

Please don't let that discourage you. If you want to revisit the topic and track down the actual cause, a proper fix would be very welcome anytime. Thanks again for the effort.

Issue #2900, opened 2020-04-16
(DB/Gameobject): Head of Onyxia


wowissues:
After killing **Onyxia** in a 25-player raid, the head did not appear in the capital, **Stormwind** (https://wowhead.com/item=18423/head-of-onyxia)

**STEPS TO REPRODUCE THE PROBLEM**:

1. Complete the 25 player raid 
2. Check StormWind or Orgrimmar

**AC HASH/COMMIT**: AzerothCore rev. f7562f8+ 2020-04-12 21:59:00 +0000

**OPERATING SYSTEM:** Debian 9

**MODULES**: Anticheat, ResetDuel, CFBG, Transmog, RewardPlayedTime, PvPTitles

Nefertumm:
Seems very obvious but did you complete the quest from the head?

wowissues:
> Seems very obvious but did you complete the quest from the head?

yes

merdlin:
UPDATE `item_template` SET `flags` = flags|4096 WHERE `entry` IN (18422,18423,49644,49643);
 test this correction

FrancescoBorzi:
@merdlin it would be better to create a PR: 

http://www.azerothcore.org/wiki/How-to-create-a-PR

heyitsbench:
Would this not be a "blizzlike" bug then, as it's a different quest entirely?

PR #25913, opened 2026-05-20
chore(deps/jemalloc): Replace jemalloc with mimalloc

### LINKED ISSUE #26734: jemalloc compile failure
### Current Behaviour

I'm going through the Arch install instructions and revamping them. Some of the description text isnt accurate, but the shell steps I followed to get my VM ready to do an install (https://github.com/mrbass21/azerothcore-wiki/blob/archinstall/docs/arch-linux.md).

```
[  4%] Building CXX object deps/jemalloc/CMakeFiles/jemalloc.dir/src/jemalloc_cpp.cpp.o
[  4%] Built target revision.h
[  8%] Built target g3dlib
/home/beck/azerothcore/deps/jemalloc/src/jemalloc_cpp.cpp:70:8: error: no member named '__throw_bad_alloc' in namespace 'std'; did you mean '__throw_bad_cast'?
   70 |                 std::__throw_bad_alloc();
      |                      ^~~~~~~~~~~~~~~~~
      |                      __throw_bad_cast
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/16/../../../../include/c++/16/bits/functexcept.h:52:3: note: '__throw_bad_cast' declared here
   52 |   __throw_bad_cast(void) __attribute__((__noreturn__,__cold__));
      |   ^
1 error generated.
make[2]: *** [deps/jemalloc/CMakeFiles/jemalloc.dir/build.make:275: deps/jemalloc/CMakeFiles/jemalloc.dir/src/jemalloc_cpp.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:970: deps/jemalloc/CMakeFiles/jemalloc.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

I was able to compile by applying the following patch:

```
From 58d3dd7d2611255a9ea24f46434c88378b3418f3 Mon Sep 17 00:00:00 2001
From: mrbass21 <mrbass21@gmail.com>
Date: Wed, 22 Jul 2026 15:49:06 +0000
Subject: [PATCH] patch jemalloc

---
 deps/jemalloc/src/jemalloc_cpp.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/deps/jemalloc/src/jemalloc_cpp.cpp b/deps/jemalloc/src/jemalloc_cpp.cpp
index da0441a7c..05d5c3a7c 100644
--- a/deps/jemalloc/src/jemalloc_cpp.cpp
+++ b/deps/jemalloc/src/jemalloc_cpp.cpp
@@ -67,7 +67,7 @@ handleOOM(std::size_t size, bool nothrow) {
        }
 
        if (ptr == nullptr && !nothrow)
-               std::__throw_bad_alloc();
+               throw std::bad_alloc();
        return ptr;
 }
 

base-commit: 007f866cbf44d49af6f2c3955d3677bca8ad0a55
-- 
2.55.0

```
Apply this patch by saving it to a file (jemalloc.patch) in the root of AzerothCore. Then run `git am --3way jemalloc.patch`, replacing the file name with whatever you named the patch

Make changes without patch:

Linux: 
```
sed -i 's/std::__throw_bad_alloc();/throw std::bad_alloc();/g' deps/jemalloc/src/jemalloc_cpp.cpp
```

Windows:
```
powershell -Command "(Get-Content deps\jemalloc\src\jemalloc_cpp.cpp) -replace 'std::__throw_bad_alloc\(\);', 'throw std::bad_alloc();' | Set-Content deps\jemalloc\src\jemalloc_cpp.cpp"
```

### Logs and Screenshots

_No response_

### Operating System

Arch 7.1.4-arch1-1

### Dependencies & versions

-- CMake version: 4.4.0
CMake Warning (deprecated) at CMakeLists.txt:22 (cmake_policy):
  The OLD behavior for policy CMP0153 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
This warning is for project developers.  Use -Wno-author or -Wno-deprecated
to suppress it.

-- Running cmake hook: AFTER_LOAD_CONF
-- No hooks registered for AFTER_LOAD_CONF
-- Enabled С++20 standard
-- Detected 64-bit platform
-- Detected amd64 processor architecture
-- UNIX: Using jemalloc
-- UNIX: Using default configuration directory
-- UNIX: Using default library directory
-- UNIX: Configuring uninstall target
-- UNIX: Created uninstall target
-- UNIX: Detected compiler: /usr/bin/clang
-- Clang: Minimum version required is 10.0.0, found 22.1.8 - ok!
-- Clang: All warnings enabled
-- Running cmake hook: AFTER_LOAD_CMAKE_MODULES
-- No hooks registered for AFTER_LOAD_CMAKE_MODULES
-- Using mysql-config: /usr/bin/mysql_config
-- Found MySQL version: 9.7.1
-- Found git binary : /usr/bin/git

* AzerothCore revision            : 7402f1461a26+ 2026-07-21 22:38:03 -0300 (master branch)
* AzerothCore buildtype           : RelWithDebInfo

* Install core to                 : /env/dist
* Install libraries to            : /env/dist/lib
* Install configs to              : /env/dist/etc

* Build applications              : Yes (all)
* Build tools                     : Yes (all)
* Build with scripts              : Yes (static)
* Build with modules              : Yes (static)
* Build unit tests                : No  (default)
* Build core w/PCH                : Yes (default)
* Build scripts w/PCH             : Yes (default)
* Show all warnings               : Yes
* Use coreside debug              : No  (default)
* Use unix gperftools             : No  (default)
* Use GIT revision hash           : Yes (default)
* Enable vmap DisableMgr checks   : Yes (default)
* Show source tree                : No (For UNIX default)

CMake Warning (policy) at deps/boost/CMakeLists.txt:35 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-author or -Wno-policy to
suppress it.

-- Found Readline library: /usr/lib/libreadline.so
-- Include dir is: /usr/include
-- Running cmake hook: BEFORE_SRC_LOAD
-- No hooks registered for BEFORE_SRC_LOAD

* Apps build list (all):
  |
  +- apps
  |   +- authserver
  |   +- worldserver
  |

-- Running cmake hook: BEFORE_GAME_LIBRARY
-- No hooks registered for BEFORE_GAME_LIBRARY
-- Running cmake hook: AFTER_GAME_LIBRARY
-- No hooks registered for AFTER_GAME_LIBRARY
-- Running cmake hook: BEFORE_SCRIPTS_LIBRARY
-- No hooks registered for BEFORE_SCRIPTS_LIBRARY
* Script configuration (static):
  |
  +- worldserver
  |   +- Commands
  |   +- Custom
  |   +- EasternKingdoms
  |   +- Events
  |   +- Kalimdor
  |   +- Northrend
  |   +- OutdoorPvP
  |   +- Outland
  |   +- Pet
  |   +- Spells
  |   +- World
  |

* Tools build list (all):
  |
  +- tools
  |   +- dbimport
  |   +- map_extractor
  |   +- mmaps_generator
  |   +- vmap4_assembler
  |   +- vmap4_extractor
  |

* Modules configuration (static):
  |

-- * Modules config list:
  |

-- Running cmake hook: AFTER_SRC_LOAD
-- No hooks registered for AFTER_SRC_LOAD
-- Configuring done (0.5s)
-- Generating done (0.7s)
-- Build files have been written to:

### Commit

7402f1461a26a5e7bee3548fd2e73e3d40ea3a46

### Additional Context

_No response_

Nefertumm:
Time for jemalloc to go. It has been a long way and used for memory allocation, but its usage was strictly locked to UNIX systems. We tried to ignore this, but we can't anymore because of building issues with newer compilers. We grasped the possibility of bumping its version to a newer one, but our implementation was messy enough to make the work way more difficult than it should be.

Reiterating on jemalloc being locked to UNIX systems specifically, this caused an environmental disparity, where UNIX systems benefited from a high-performance custom allocator, while Windows builds were stuck using the default MSVC system allocator. This lack of parity makes memory profiling, reproducibility, and leak tracking inconsistent across different system builds.

This PR drops the jemalloc dependency and introduces mimalloc (v2.3.2 - stable) as a universal, drop-in replacement for both Linux and Windows. Developed by Microsoft, mimalloc provides first-class native support for Windows while delivering exceptional performance characteristics (see https://github.com/microsoft/mimalloc#performance)

## Changes Proposed:

- Removed all references to jemalloc and the NOJEM flag across CMake configurations.
- Vendored mimalloc v2.3.2 into deps/mimalloc.
- Configured CMake to enforce static building and global allocation overrides (MI_OVERRIDE ON), enabling a zero-touch C++ integration.
- Maintained Valgrind tracking support for UNIX environments via MI_TRACK_VALGRIND.

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

Claude Sonnet 4.6

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/26734

## SOURCE:

## Tests Performed:

This PR has been:
Compiled and ran worldserver and authserver on arch linux.

## How to Test the Changes:

1. Compile with every configuration possible.
2. Everything should work as before


sudlud:
that's a lot of whitespace changes, are those necessarty for this pr?

Nefertumm:
Whitespaces gone, copilot suggestion accepted.

walkline:
Do we apply changes to dependencies directly?

I think previously, whenever we needed custom changes for a dependency, we added a patch file alongside with the changes itself.

Maybe we could apply this official PR instead: [#2900](https://github.com/jemalloc/jemalloc/pull/2900)
Unfortunately, it hasn't been included in a release yet. We could also mention in [TC_NOTE.txt](https://github.com/azerothcore/azerothcore-wotlk/blob/master/deps/jemalloc/TC_NOTE.txt) (or maybe rename it to AC_NOTE?) that this PR has been applied.

Nefertumm:
Yes, I'm more inclined to bump the version / apply official commits than make custom patches for deps. I'll see what I can do later today

LuckyLuc96:
Will have it idle for an hour or so and play some tomorrow, but it compiled and is running for me. I do use several mods. 



AC> server debug
AC> AzerothCore rev. 5955f52f0447 2026-05-22 00:18:49 -0300 (jemalloc_build branch) (Unix, RelWithDebInfo, Static)
Connected players: 0. Characters in world: 0.
Connection peak: 0.
Server uptime: 41 second(s)
|- Mean: 1ms
|- Median: 1ms
|- Percentiles (95, 99, max): 4ms, 5ms, 6ms
Using SSL version: OpenSSL 3.5.5 27 Jan 2026 (library: OpenSSL 3.5.5 27 Jan 2026)
Using Boost version: 1.90.0
Using CMake version: 4.3.0
Using MySQL version: 80409
Found MySQL Executable: /usr/sbin/mysql
Compiled on: Linux 7.0.9-205.fc44.x86_64
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: false
MMAPs status: Enabled
maps directory located in /home/luc/server cores/azerothcore/env/dist/data/maps. Total size: 291014951 bytes
vmaps directory located in /home/luc/server cores/azerothcore/env/dist/data/vmaps. Total size: 662146712 bytes
mmaps directory located in /home/luc/server cores/azerothcore/env/dist/data/mmaps. Total size: 2213624496 bytes
Default DBC locale: enUS.
Using World DB: ACDB 335.16-dev
Latest LoginDatabase update: acore_cms_subscriptions.sql
Latest CharacterDatabase update: u_mp_2025_06_30_00.sql
Latest WorldDatabase update: u_mp_2025_07_03_00.sql
LoginDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-1v1-arena
|- mod-account-achievements
|- mod-account-mounts
|- mod-ale
|- mod-aoe-loot
|- mod-autobalance
|- mod-custom-login
|- mod-fireworks-on-level
|- mod-gain-honor-guard
|- mod-individual-xp
|- mod-learn-spells
|- mod-low-level-arena
|- mod-morph-all-players
|- mod-pocket-portal
|- mod-profession-experience
|- mod-random-enchants
|- mod-skip-dk-starting-area
|- mod-solo-lfg
|- mod-transmog
|- mod-who-logged


____ System info

image


(Ignore diskdrive space)

espkk:
Since you mentioned windows, this won't work the way you expect, but I need some time to recall all the specifics
Also, just curious - why not v3 instead? It's been stable and production-ready for a while. IMO, v2 is overly conservative for AC :)

Nefertumm:
Hey, thanks for the feedback! 
Well damn, please take your time to recall everything you need. I would greatly appreciate help on this matter.
About the version, yeah, I was pretty conservative about it, and I went for their 'stable' versión, but it can be bumped easily if needed.

Nefertumm:
@espkk I see what you meant now by not working as I expected. It's just the way MSVC works. I need to add a lot of things to CMake and make sure the dlls from mimalloc are used by the compiler at runtime, while also adding a piece of code to worldserver's main cpp file. (https://github.com/microsoft/mimalloc#dynamic-override-on-windows)

espkk:
> @espkk I see what you meant now by not working as I expected. It's just the way MSVC works. I need to add a lot of things to CMake and make sure the dlls from mimalloc are used by the compiler at runtime, while also adding a piece of code to worldserver's main cpp file. (https://github.com/microsoft/mimalloc#dynamic-override-on-windows)

I think it's not that bad! mimalloc actually recently got static override support on windows, but there are still many nuances in how it works depending on configuration, so this is probably the right thing to do if CRT is also linked statically (which, however, is not great by itself)

In this particular case, the main problem is that you don't use any of mimalloc's interface functions, which (I haven't tested yet) likely means the whole library will be stripped on link stage. At least on windows. On linux, iirc it was recommended to link .o file instead (don't remember the details unfortunately), but maybe .a works too without being stripped out

Second thing is that I think it's much better to link to a TU with `#include <mimalloc-new-delete.h>` - this way they will be overloaded on a language level avoiding malloc path, losing extra context and missing optimizations. This would also work if CRT intercept failed. This also automatically solves the issue of missing calls to mimalloc lib - `mimalloc-new-delete.h` itself contains them

Then finally, yes there is `redirect` mechanism, that not only overrides the allocator for the built module, but also for all early-stage and late-stage linked dynamic dlls. But in my experience this is fragile as even windows sometimes ships updates with misaligned allocations that leads to crashes with `redirect` enabled.
But if we keep it disabled, we'll need to keep in mind that cross-module deallocations (if any) will become UB if the corresponding modules will be linked without mimalloc. e.g. some boost or mysql dynamic lib that allocates and returns unique_ptr back to own and delete on destroy

I tried to be concise but happy to go into details if anything is unclear. Also happy to test the changes on windows, but maybe not in the next few days

Nefertumm:
This is incredibly helpful, thank you so much for the explanation!

Yes, on Windows, the linker would optimize out the whole thing from the link stage, thinking it was some kind of mistake.

Also, knowing that the redirect mechanism can be fragile with Windows updates makes me completely agree with your assessment: sticking to the static approach and avoiding the dynamic redirect might be the safest bet for server stability.

Regarding the stripping issue and the TU approach: that makes perfect sense. During my initial Linux tests, I actually tried including <mimalloc-new-delete.h> directly in both worldserver and authserver's main.cpp files. However, because I also had MI_OVERRIDE=ON set in CMake, the linker yelled at me with multiple definitions of 'operator delete' since the library was trying to override it simultaneously at the C and C++ levels.
Based on what you said, the cleanest path seems to be:

-  Turn off MI_OVERRIDE.

-  Explicitly include <mimalloc-new-delete.h> in the Main.cpp for both apps to anchor the library, prevent MSVC from stripping it, and benefit from the C++ level optimizations.

-  Keep the redirect disabled and trust AC's current memory ownership boundaries (which, as you mentioned, generally handle cross-module unique_ptr deletions safely, though it's something to keep an eye on and we should be very careful about).

I will prepare a new commit bumping the version to v3, adjusting the CMake flags, and re-introducing the header to each file as stated previously. Then, we can test it back on Windows!
Thank you for the insights!

espkk:
> This is incredibly helpful, thank you so much for the explanation!
> 
> Yes, on Windows, the linker would optimize out the whole thing from the link stage, thinking it was some kind of mistake.
> 
> Also, knowing that the redirect mechanism can be fragile with Windows updates makes me completely agree with your assessment: sticking to the static approach and avoiding the dynamic redirect might be the safest bet for server stability.
> 
> Regarding the stripping issue and the TU approach: that makes perfect sense. During my initial Linux tests, I actually tried including <mimalloc-new-delete.h> directly in both worldserver and authserver's main.cpp files. However, because I also had MI_OVERRIDE=ON set in CMake, the linker yelled at me with multiple definitions of 'operator delete' since the library was trying to override it simultaneously at the C and C++ levels.
> Based on what you said, the cleanest path seems to be:
> 
> -  Turn off MI_OVERRIDE.
> 
> -  Explicitly include <mimalloc-new-delete.h> in the Main.cpp for both apps to anchor the library, prevent MSVC from stripping it, and benefit from the C++ level optimizations.
> 
> -  Keep the redirect disabled and trust AC's current memory ownership boundaries (which, as you mentioned, generally handle cross-module unique_ptr deletions safely, though it's something to keep an eye on and we should be very careful about).
> 
> I will prepare a new commit bumping the version to v3, adjusting the CMake flags, and re-introducing the header to each file as stated previously. Then, we can test it back on Windows!
> Thank you for the insights!

I think MI_OVERLOAD is fine, it's just that on windows it also needs 'mimalloc-new-delete.h' to be included somewhere once to directly overload the operators (on Linux it should already be able override them automatically, hence the linkage error with MI_OVERLOAD)
Otherwise, if we don't enable MI_OVERLOAD it can become a performance regression compared to jemalloc, because I assume a number of allocations are still implicitly going through direct malloc/free

Cross-module deallocation can still probably become an issue, but if it does, there should be a few ways to solve that, and also they should be pretty resilient (unless we leak too much memory - because mismatched allocator can ignore, but cannot fulfil our deallocations). I'd recommend to test debug build of AC and all dependencies, where mismatched deallocations should be properly asserted in both mimalloc and CRT

The rest sounds good!

sudlud:
Thanks for tackling this, the jemalloc situation has needed attention for a while.

The main issue is that this PR is stale relative to the discussion it ended on. Back in late May, @espkk explained that the static override won't actually take effect (especially on Windows, where the linker strips the unreferenced mimalloc archive), and you agreed to a follow-up: bump to v3, include `<mimalloc-new-delete.h>` in worldserver and authserver `Main.cpp`, and adjust the `MI_OVERRIDE` flags. None of that made it in. The current head still ships v2.3.2, no `Main.cpp` is touched, and nothing references a mimalloc symbol, so the override is very likely a no-op on Windows and unverified on Linux.

Specific points:

1. Add `#include <mimalloc-new-delete.h>` to both Main.cpp files. This anchors the library and overloads new/delete at the language level, which is what fixes the Windows stripping.
2. Please prove the override is actually live: boot with `MIMALLOC_VERBOSE=1` (or log `mi_version()`) on both Linux and Windows and confirm allocations route through mimalloc. That is the one test that validates the whole PR, and it hasn't been shown yet. "Compiles and runs" doesn't confirm the allocator is in use.
3. `MI_TRACK_VALGRIND` is forced ON for every UNIX build. mimalloc documents this as adding overhead even outside Valgrind, so it's a perf regression on production Linux builds. It should be opt-in (default OFF), like the old NOJEM gate.
4. `MIMALLOC_LIBRARY` in `settings.cmake` is set but never consumed (same as the old `JEMALLOC_LIBRARY`); better to drop it than rename it.
5. Minor: `common` guards the link with `if(TARGET mimalloc-static)` but `apps` links it unconditionally. Worth making consistent, and the apps-level link may be redundant since `common` links it PUBLIC.

Happy to help test the Windows path once the new-delete header is in.

airbreather:
In the meantime, I've opened #26438 with just the author's first few commits on this PR, up to the last ones before the scope of this PR changed to be a full jemalloc --> mimalloc conversion.

I'm cleaning up a different branch for a PR and noticed that I still had my own version of that change lying around, and I saw several people on Discord also run into the same issue since my May 9 heads-up message on #dev-core, so I figured there's some room for a standalone quick ~~uncontroversial~~ fix.

Edit: eh, the "patch or edit inline?" question I guess makes it not quite so "uncontroversial", but I think it's the right way to go because jemalloc is hopefully not going to be around in AzerothCore much longer anyway.

sudlud:
closing this for now. please feel free to reopen anytime in the future addressing the review comments and the merge conflict. thanks!

PR #25915, opened 2026-05-21
fix(Core/Packets): Correct data source for attack stop packet, rewrite to modern class.

heyitsbench:
## Changes Proposed:

This PR proposes changes to:

## Issues Addressed:

- Might address the permanent no auto attack state, but probably not.

## SOURCE:

  - https://github.com/TrinityCore/TrinityCore/commit/a006e66058a26d9514e6c8b78f903038fb30e1d2
  - https://github.com/TrinityCore/TrinityCore/commit/345e0586114c2f8b9d2d35b89b6a66f213558274

## Tests Performed:

This PR has been:
- [X] Compiled.


PkllonG:
The test failed. Once the character performs an attack swing action, it cannot stop.

Nefertumm (approved):

Nefertumm:
So... Is the behavior still the same as before?

src/server/game/Server/Packets/CombatPackets.cpp
```diff
@@ -21,3 +21,14 @@ void WorldPackets::Combat::SetSheathed::Read()
 {
     _worldPacket >> CurrentSheathState;
 }
+
+WorldPacket const* WorldPackets::Combat::SAttackStop::Write()
+{
+    _worldPacket << Attacker;
+    _worldPacket << Victim;
```
Kitzunu:
```suggestion
```
In AC, << ObjectGuid writes the full unpacked uint64 (ObjectGuid.cpp:71), so this prepends 16 junk bytes before the packed GUIDs the client expects. The cherry-pick kept TC's << Attacker/Victim (packed by default in TC) and added explicit WriteAsPacked(). The client misparses the packet → auto-attack state won't clear — the very bug this PR targets.

Fix: drop the two unpacked lines, keep only the WriteAsPacked() writes + uint32.

sudlud (approved):

sudlud:
Pushed some fixes to the branch (packed GUID writes were duplicated, plus cleanup and a merge of master). Two behaviour changes worth calling out, both checked against sources outside TC:

**`NowDead` now reads the attacker instead of the victim.** A 1.10 retail sniff confirms this is right: a dying player gets `SMSG_ATTACK_STOP` with his own GUID, victim 0, NowDead 1 ([sniff](https://github.com/ratkosrb/VanillaSniffs/blob/ff119e685c0945833b76ab139e3d6b72dd4df898/1.10/gnome_warrior_dun_morogh_1.10.0.5195_2006-03-31_09-59-00_parsed.txt#L4645)). vmangos landed the same change independently a day after TC ([ef0b5bcb59](https://github.com/vmangos/core/commit/ef0b5bcb59b9ac844d2999d8bb93a64b78fe5dcd)). WowPacketParser documents the field as blocking clientside facing when set, so the old victim-based value was blocking facing on a live attacker whose target had died.

**The three error paths in `HandleAttackSwingOpcode` went from unicast to broadcast** when `SendAttackStop` was folded into `SendMeleeAttackStop`. No new amplification: the success path of the same opcode already broadcasts through `SendMeleeAttackStart`, and the packet is inert for other clients in these cases.

@PkllonG for context on the failed test: the packet was malformed at that point, 16 junk bytes in front of the packed GUIDs, so the client could not parse the attacker and auto attack never stopped. That has since been fixed.

PR #18244, opened 2024-01-31
chore(Docker): update mysql image to 8.3 version

playerx:
## Changes Proposed:

It's an enhancement for Docker setup. Updated mysql image to the latest 8.3 version. After updating tested it locally with an existing data and worked perfectly.

## Issues Addressed:
none

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.

## How to Test the Changes:

To tested this change you will need to:
1. Create a brand new local azerothcore server setup using docker
2. Make sure that game is playable by creating a character and grabbing some items in-game
3. Update mysql database version up to 8.3
4. Deploy the updated image by running `docker compose up -d`
5. Test that game is playable, your character and items are in place

## Known Issues and TODO List:

none

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #25968, opened 2026-05-25
feat(CI): publish AC DB Docker image

FrancescoBorzi:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

Claude Code with Opus 4.7

## Issues Addressed:

Adds Docker image publishing mechanism

## Tests Performed:
None yet

## How to Test the Changes:

N/A

N/A


sudlud:
The DB-image gate only watches `data/sql/**`, but the seeded datadir also bakes in mod-ale SQL, checked out from its own master at build time. `db-import:master` is rebuilt on every master push, so a push that bumps mod-ale without touching `data/sql/**` updates `db-import:master` while leaving `ac-wotlk-db:master` stale. No commit signal in this repo can catch that.

Suggest dropping the gate and always rebuilding the DB image on master, like `db-import` does.

Minor: `workflow_dispatch` also re-pushes the six sibling images on a manual run, not just the DB image. Intended?

TheSCREWEDSoftware:
@FrancescoBorzi what's the status of this PR :)?

Yehonal (changes requested):
Thanks for putting this together. A pre-imported DB image would save people a lot of time, and most of the scripting here is careful.

Two things I'd like to sort out before this goes in.

## The workflow has never completed a run

All the `docker-build` runs on this branch are `skipped`, so as far as I can tell the pipeline has never actually built or pushed anything. The PR also still says "Tests Performed: None yet".

For a change that publishes artifacts other people will run, I think we need at least one successful end to end run before merge. Push to a staging repo if you prefer, then pull the image and boot it. Right now we'd be merging a publishing pipeline that nobody has seen finish.

Worth covering arm64 there too. We build and push it, but the smoke test only boots amd64.

## The DB gate can lose an SQL change for good

Say push A changes `data/sql/**`. It publishes the six sibling images, then push B arrives and cancels it before the DB step, since `cancel-in-progress` is true on the master ref.

B has no SQL changes, and the gate only compares `github.event.before` with `github.sha`, so `build=false`. A's SQL never makes it into the DB image, and no later push looks back past its own `before`.

`ac-wotlk-db:master` then sits behind while the sibling tags move forward. Could we base the gate on the last DB commit we actually published, instead of on `github.event.before`?

## Smaller things, but I'd fix them here

The workflow doesn't declare `permissions`, and the repo default is `write`. On a labeled PR the job runs code from the merge ref with a write token. It isn't a secret exposure, since fork PRs are skipped and the Docker login is gated on master, but I'd still set `contents: read` and give write only to the push job.

The actions are all on moving tags (`@v7`, `@v3`, `@v6`), and `mysql:8.4`, the importer image and the `mod-ale` checkout are moving too. So rebuilding the same commit doesn't give you the same image. Could we pin the actions to SHAs and the images to digests?

We only publish `:master` and `:${VERSION}`, and the version is `17.0.0-dev`, so both get overwritten on every build. That leaves no way to pull the image for a given commit or to roll back, while the README promises a matching commit. Maybe add a `${{ github.sha }}` tag and the OCI source and revision labels?

## Already flagged, still open

The known `acore`/`acore` credentials with `GRANT OPTION` end up in the published datadir, and the README shows `-p 3306:3306`. CodeRabbit raised this one.

The gate only watches `data/sql/**`, but the importer pulls `mod-ale` from its default branch with no ref. Copilot and sudlud raised this one.

## What I like

You didn't use `pull_request_target`, and both the push and the Docker login are limited to master on the canonical repo. That's the part that matters most and it's right.

The error handling is careful too: `set -euo pipefail`, explicit timeouts, cleanup traps, and MySQL stopped cleanly before the datadir is captured. Restricting the build context to the datadir instead of the whole repo is a nice touch.

Yehonal:
One more thing, and I think it comes before my other comments.

I went looking for who would actually use this image, and I couldn't find them. The PR doesn't touch either compose file. `acore-docker` still runs `mysql` plus the `ac-wotlk-db-import` job, and the compose here does the same. So after the merge we'd be publishing an image that nothing consumes, while the README asks people to wire it up by hand and turn the updater off.

That's two DB bootstrap paths to maintain, with their own docs and their own ways to break.

To be clear, this isn't a duplicate image. I checked and `acore/ac-wotlk-db` doesn't exist on Docker Hub yet, and the snapshot idea has a real use case for CI and throwaway fixtures on a fresh volume. What worries me is it becoming a second official path instead of an optimization inside the current one.

## About the MySQL 8.4 base

I was ready to call this a problem, but it's the opposite. We moved to 8.3 in #18244 and to 8.4 LTS in #20976, so this PR is consistent with us. `acore-docker` is the one still on 8.0, roughly 21 months behind, and `mysql:8.0` hasn't been rebuilt since 2026-05-05 while `8.4` was rebuilt on 2026-07-28.

It does force an order though. A MySQL datadir can't be downgraded, so a snapshot baked on 8.4 won't open on an 8.0 server. `acore-docker` needs to move to 8.4 before or together with this, otherwise we publish an image its main audience can't run. I'll open a separate issue there, since that bump is worth doing on its own.

## What I'd like to settle first

Who is the first consumer, and where is the compose change that uses it?

Is the snapshot a baseline that `db-import` still applies deltas on top of, or a replacement for the importer? Those are two different proposals and the PR reads like both.

My earlier review still stands, but I'd answer these first. Fixing eight details on an artifact we haven't decided to build is the wrong order.

FrancescoBorzi:
# Self-review — docker-image → master

**Outcome** 16 findings — 12 fixed, 2 deferred, 2 dismissed. Round 2 found a step-fatal bug that
round 1's own fix had put on the publish path; round 3 confirmed the repair and raised 3 more.

**Caveat** Round 3's two fixes were not re-reviewed (3-round cap reached).

**Caveat** Unpinned — round 3's fixes are still uncommitted on top of `96d523791`.

**Caveat** Nothing in this changeset has ever executed in CI. The smoke test carried a `local`
unbound-variable bug from the first commit onward, so it could never have passed; the first green
run will also be the first real exercise of `buildx build --load` on a multi-GB image, the amd64
boot, the emulated arm64 boot inside its 600s budget, and `buildx prune` between arches.

**Reviewed** working tree on `96d523791` (`docker-image` vs `master`, merge-base diff — 5 files,
+273 −1), unpinned

**By** Claude Code, claude-opus-5 — self-review v1.1, 2026-09-07

**Testing performed** None. No end-to-end publish run, no labeled-PR run, no manual compose
opt-in. Author confirmed.

**Verification during review** Both branches of the build step were executed against stubbed
`docker`/`jq`/`sleep`/`sudo`, covering the master path (amd64 smoke then multi-arch push), the PR
path (amd64, prune, arm64), and every failure path: readiness timeout, boot-poll timeout, unclean
mysqld shutdown (exit 137), and `USE <db>` failure. All abort non-zero, and no failure case reaches
the `--push`. YAML parses; `.editorconfig` conformance and `git diff --check` clean.

**Before merging** A labeled `run-build` PR run needs to go green — that is the only thing that
exercises any of this. Beyond the happy path, worth checking: a compose bring-up on a host that
already has an `ac-database` volume (the snapshot is silently skipped, now documented); a run with
`DOCKER_DB_ROOT_PASSWORD` set to confirm the stack hangs the way the README says; and whether the
`:<sha>` tag accumulation in finding 14 is acceptable to the maintainers.

<details>

**Intent** Publish a pre-imported AzerothCore database Docker image (`acore/ac-wotlk-db`) from the
existing `docker_build` workflow: a MySQL 8.4 image whose datadir already holds the imported
`acore_auth`/`acore_world`/`acore_characters` so it boots without the multi-minute first-run
import. Pushed only on master of the canonical repo, smoke-tested on labeled PRs, and offered as an
opt-in baseline for this repo's docker-compose stack via a new `DOCKER_DB_IMAGE` variable.

**Project rules** `.agents/docs/self-review-rules.md` found and applied — pinned review (held for
rounds 2 and 3, not for the final state), regression-risk depth on shared infrastructure, upstream
attribution (N/A, nothing mirrored), in-game testing asked rather than guessed.

**Diff** `277d6994a30a961323a44d6a2b926a1951670246`

**Caveat** Each round ran in a fresh context with `.agents/plans/**` and the report excluded, so no
author rationale reached any reviewer.

## Round 1 — `50181bc2f`, 10 findings

1. `docker_build.yml:232` — master pushes a multi-arch build that can never be `--load`ed, so the
   only publishing path is the only one with no boot test; the smoke test ran solely on labeled
   PRs → **fixed**: smoke hoisted out of the `else`, master boots amd64 before pushing
2. `docker_build.yml:207` — `docker stop` exits 0 even when the grace period expires and mysqld is
   SIGKILLed, capturing a torn datadir → **fixed**: exit code checked
3. `README.md:68` — `:master` recommended alongside `--build` from a local checkout, so an older
   branch gets a newer schema that `ac-db-import` can only roll forward → **fixed**: drift warning
4. `docker_build.yml:242` — comment claims two images never coexist, but `docker rmi` leaves the
   buildx store populated → **fixed**: `buildx prune -f` between arches
5. `README.md:50` vs `docker_build.yml:242` — "a few hundred MB" contradicts "multi-GB" → **fixed**:
   download and on-disk sizes separated
6. `docker_build.yml:232` — 126 chars, over `.editorconfig`'s 120 → **fixed**
7. `docker_build.yml:10` — `workflow_dispatch` re-pushes all seven images and fires the downstream
   dispatch → **dismissed**: intentional, it is the lever to refresh every image in lockstep when
   `mod-ale` moves with no AzerothCore commit to trigger a build
8. `docker_build.yml:75` — QEMU set up unconditionally though only the PR path boots arm64 →
   **dismissed**: a few seconds, and an `if:` would duplicate the branch condition
9. `Dockerfile.db:14` — hardcoded uid 999 needs a comment and can drift → **fixed**:
   `--chown=mysql:mysql`
10. PR #25968 description — "Tests Performed: None yet", "How to Test: N/A", disclosed model
    "Opus 4.7", and the body predates master's current template (which added a `SOURCE:` section
    and a second AI checkbox) → **deferred**: author updates the PR before merge, and it cannot be
    honestly filled in until a run exists

## Round 2 — `3c4318214`, 3 findings

11. `docker_build.yml:241` — bash expands every word of `local` before assigning any, so
    `tag="...-$arch"` on the same line read the unset outer `arch` and `set -u` killed the step.
    Present since the first commit, so the smoke test never ran; round 1's hoist moved it in front
    of the publish, turning a dead check into a dead publish → **fixed**: split declarations,
    reproduced and re-verified by simulation
12. `README.md:69` — advised pinning `:<version>` on an older branch, but that tag moves with
    master exactly like `:master`, causing the drift the same sentence warns about → **fixed**:
    `:<sha>` only
13. `docker_build.yml:75,277` — comments assert "PR validation boots both arches", untrue on the
    push path and on `workflow_dispatch` → **fixed**: reworded to say arm64 is covered only by the
    labeled-PR check

## Round 3 — `96d523791`, 3 findings

14. `docker_build.yml:284` — a per-commit `:<sha>` tag is pushed on every master build and nothing
    reclaims it; layers never dedupe because each run regenerates the datadir (fresh server UUID,
    LSNs, timestamps), so every merge permanently adds a few hundred MB to the org's Docker Hub
    → **deferred**: the tag was added at a reviewer's request and the README documents it as the
    only immutable one, so a retention window vs. dropping it is a maintainer call; author raises
    it on the PR
15. `docker_build.yml:227` — `VERSION` re-derived with `jq` though the job already publishes
    `steps.version.outputs.version`, which all six sibling steps use → **fixed**
16. `README.md:52` — the quick-start silently no-ops on a host that has already run the stack,
    since Docker seeds a volume only when it is new and empty; user gets their old database with no
    warning → **fixed**: `docker compose down -v` documented in the quick-start

</details>

Yehonal:
Thanks for the rework. This is a much better shape than what I reviewed, so let me drop what no longer applies first.

The consumer question is answered. `DOCKER_DB_IMAGE` in the main compose is a real opt-in consumer, and keeping `ac-db-import` for the deltas makes the image a baseline instead of a second updater. That was my main worry and it's gone.

Point 6 on the dispatch order I withdraw. `acore-docker` doesn't consume this image, the main compose does, and the importer covers the deltas anyway.

Point 2 on pinning I'd narrow to a follow-up. The moving action tags are worth fixing, but they're repo wide and this PR didn't introduce them.

And on MySQL I was half wrong. `acore-docker` does need to move to 8.4 before it could adopt the seed, but that isn't a prerequisite for publishing here, since the main compose is already on 8.4. I opened azerothcore/acore-docker#52 for that bump, it stands on its own.

Permissions, the gate race, `GRANT OPTION` and the mod-ale blind spot all look fixed to me in the diff.

## Three things left

**The run.** We agree here, you said it yourself. Stubs cover the control flow, not Buildx, arm64 emulation, disk headroom or a real boot. I'd like a green `run-build` on `eb7bb714`, with the observed times and image size, before this goes in.

**The SHA tag isn't immutable yet.** `workflow_dispatch` is enabled and the push includes `acore/ac-wotlk-db:${{ github.sha }}`, so a second dispatch on the same commit rebuilds from `mod-ale` HEAD, `db-import:master` and `mysql:8.4`, then pushes the same tag again. Same SHA, different bytes. Either record those inputs in a manifest, or drop the immutability and deterministic rollback wording from the README.

**One thing the volume fix introduced.** The README tells people with an existing stack to run `docker compose down -v`. That removes every named volume in the project, so `ac-client-data`, `ac-build-dev` and `ac-ccache-dev` go with it, and the instruction is aimed exactly at users who already have real data. Could we point at the DB volume specifically and add a line about backing up first?

## Two calls for the maintainers, not for you

Whether we keep every SHA tag forever or prune them. That's release policy and it shouldn't land on you.

And the lockstep republish of all seven images on manual dispatch. You explained why and the reason holds, someone just needs to accept the cost.

I'll clear the block as soon as the run is green and the tag wording matches what the workflow actually does.

FrancescoBorzi:
@Yehonal, thanks for your feedback. I've just updated this PR again. Can you please check?

PR #26058, opened 2026-06-01
feat(Core): Diagnostics buffer (and wiring to dump WG data)

espkk:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


espkk:
CI errors doesn't seem to be related

UPD:
OK I guess it's because I added `acore-core-interface`  to tests to match deps and compiler flags. Without it windows tests don't even compile. I'd rather fix the new errors

sudlud:
Framework is clean and well tested, but I have a design question and a blocker.

**Why a custom framework over logging?**

This adds a parallel observability system (ring buffer, `StringLiteralView`, `OverwritingRingBuffer`, `ScopeExit`, variant record, custom dump format) next to the logging we already have. A ring buffer earns its keep as a flight recorder: always-on capture you dump after an unpredictable incident. But this is toggled on with `.bf diagnostics on` before reproducing, which is exactly the workflow of raising a log category to TRACE. In that mode plain logging gives the same data with none of the new core surface.

Can this be a dedicated category (e.g. `bg.battlefield`) with `LOG_TRACE` carrying the same fields? That keeps the value, reuses existing appenders/rotation, and drops ~2000 lines of core to maintain. If there is a real need for bounded always-on in-memory capture that logging can't do, please state it with a linked issue for the WG problem being chased, since that is the only thing that justifies the extra machinery. As written, the inline `trace.Arg(...)` calls also hurt readability of the core paths more than log lines would.

**Blocker: does not build on the `-Werror` CI configs**

Linking tests against `acore-core-interface` propagates the project warning-as-error flags onto the tests, making pre-existing unused parameters fatal:

```
src/test/mocks/UnitStub.h:47: fatal error: unused parameter 'unit' [-Wunused-parameter]
```

All clang/gcc/macOS/windows jobs fail on this. Needs fixing before the PR can be evaluated on green CI.

**Minor**
- The two earlier bot comments (SQL placeholder, console-only help text) look already addressed.
- Trace args are still evaluated when diagnostics are off (the enabled check is inside `Arg`). Negligible here, but not zero-cost-when-off.
- `friend class BfCapturePoint/BfGraveyard` look unnecessary; they only use the public `Trace()`.
- `dump` without ever enabling allocates the full 1MB buffer to dump nothing.

PR #26069, opened 2026-06-03
feat(Scripting): add OnPlayerCreatureLootOpened hook

pangolp:
## Changes Proposed:

Adds scripting infrastructure for modules that need to react when a player opens a creature's loot window or auto-collect items from nearby corpses (AOE loot use-case).

### Details

**`OnPlayerCreatureLootOpened` (PlayerScript hook)**
Fires inside `Player::SendLoot` after the permission check passes and the `SMSG_LOOT_RESPONSE` packet has been sent to the client. At that point the loot state is fully initialised: `roundRobinPlayer` is set, `FillNotNormalLootFor` has run for all nearby group members, and any group loot / NeedBeforeGreed / MasterLoot roll setup is complete. Only fires for creature GUIDs (not items, GOs, or corpses).

**`Player::AutoTakeCreatureLoot(Creature*)`**
Takes all loot items a player is eligible for from a creature directly into their inventory, without opening a loot window on the client.
1. Validates `UNIT_DYNFLAG_LOOTABLE` and `isAllowedToLoot`.
2. Calls `FillNotNormalLootFor` (no-op if already done at creature death).
3. Temporarily sets the player's active loot GUID to this creature so `StoreLootItem` internal checks work correctly.
4. Iterates all loot slots via `StoreLootItem` — items are stored, notifications sent, and loot state updated exactly as if the player had manually taken each item.
5. Restores the original loot GUID.
6. If loot is fully depleted, calls `AllLootRemovedFromCorpse` and removes `UNIT_DYNFLAG_LOOTABLE`.

Returns `true` if at least one item was taken.

### AI-assisted Pull Requests

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. **Claude Sonnet 4.6** was used as a co-author for implementation guidance and code review.

## Issues Addressed:

- Closes

## SOURCE:

> This is scripting infrastructure; it does not change any game behaviour by itself. Behaviour is implemented by modules consuming the hook and the new method.

## Tests Performed:

## How to Test the Changes:

- [ ] This pull request requires further testing. Provide steps to test your changes.

1. Apply the patch and compile.
2. Load the world server and confirm no errors on startup.
3. Create or use a module that registers a `PlayerScript` and implements `OnPlayerCreatureLootOpened`. Kill a creature and open its loot window — verify the hook fires with the correct `Creature*` and `Loot*` pointers.
4. From the same module, call `player->AutoTakeCreatureLoot(creature)` on a lootable corpse. Verify the items land in the player's inventory without a loot window appearing, and that `UNIT_DYNFLAG_LOOTABLE` is removed from the corpse once all eligible items are taken.
5. Verify normal looting (opening the loot window manually) still works correctly and that no regression exists for group loot modes.

- [ ] `AutoTakeCreatureLoot` currently takes all eligible items without respecting the player's free bag space gracefully — a future improvement could return the remaining items to the corpse or notify the player.

## Summary by CodeRabbit

* **New Features**
  * Added a new script event hook that notifies when a player opens loot from a creature, enabling customization of creature looting behavior through scripts.


src/server/game/Entities/Player/Player.cpp
```diff
@@ -13709,6 +13713,50 @@ LootItem* Player::StoreLootItem(uint8 lootSlot, Loot* loot, InventoryResult& msg
     return item;
 }
 
+bool Player::AutoTakeCreatureLoot(Creature* creature)
```
Nyeriah: I believe this code was meant to be part of the module. It is dead code as it is, not used anywhere

pangolp: > I believe this code was meant to be part of the module. It is dead code as it is, not used anywhere

You're right, it's code for the mod-aoe-loot, I should only include the hook, and for the rest, see if I can write it inside the mod itself. I'll see if I can write it another way. Although, I think I need the loot anyway, because with the current hooks, I can't solve the loot problem as a group.

pangolp: I'll move that part of the code to the module, and leave only the hook. I don't think I can write it any other way, because I've been trying to figure out how to solve this group loot issue for a while now.

PR #26074, opened 2026-06-04
fix(CI/Windows): fix MySQL installation

pangolp:
## Changes Proposed:

This PR fixes the Windows CI pipeline, which was installing MySQL 9.x from the
Chocolatey CDN instead of using the MySQL 8.x already pre-installed on the
GitHub Actions runner.

-  [x] CI / build infrastructure (Windows installer script, compiler helper).

**Details:**

- `apps/installer/includes/os_configs/windows.sh`: The CI branch was still
  calling `choco install mysql` without a version pin, which caused Chocolatey
  to download and install MySQL 9.x. The fix skips the MySQL installation
  entirely when `CONTINUOUS_INTEGRATION` is set (the runner image already ships
  MySQL 8.x at `C:\tools\mysql\current\`). The same skip is triggered by
  `SKIP_MYSQL_INSTALL=1`, mirroring the pattern already used in `ubuntu.sh`.

- `apps/compiler/includes/functions.sh`: The MySQL CMake flags for `msys*` were
  unconditionally hardcoded to the Chocolatey path. They are now conditional on
  that path actually existing, so `FindMySQL.cmake` auto-detection via the
  Windows registry and `Program Files` glob is used as a fallback when MySQL was
  not installed through Chocolatey.

### AI-assisted Pull Requests

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or
  partially in preparing this pull request. **Claude (claude-sonnet-4-6)** was
  used to analyse the root cause and draft the fix.

## Issues Addressed:

- Closes 

The Windows CI was installing MySQL 9.x from the Chocolatey CDN due to a
missing version pin, breaking or potentially breaking the build against a MySQL
version not validated by the project.

## SOURCE:

- [x] The root cause is directly observable in the CI logs (MySQL 9.x reported
  as installed) and traceable to the missing version pin introduced in commit
  `0ec122510`.

## Tests Performed:

This PR has been:
- [x] This pull request requires further testing and may have edge cases to be
  tested. The fix must be validated by a successful CI run on the `windows-latest`
  runner to confirm that MySQL 8.x from the base image is picked up correctly by
  CMake.

## How to Test the Changes:

- [x] This pull request requires further testing. Steps:

1. Push the branch or open the PR to trigger the `windows-build` workflow.
2. Inspect the "Configure OS" step log and confirm that the MySQL Chocolatey
   install is skipped.
3. Inspect the "Build" step log and confirm that CMake finds MySQL 8.x
   (look for `Found MySQL version: 8.x.x` or the path resolved under
   `C:/tools/mysql/current`).
4. Confirm the build completes without errors.

- [ ] The `ccache-action` step is present in `windows_build.yml` but
  `AC_CCACHE` is never set to `true`, so compiler caching is silently disabled
  on Windows. This can be addressed in a follow-up PR.
- [ ] OpenSSL and Boost are still installed via Chocolatey CDN with pinned
  versions; those pins should be reviewed periodically.

### DIFF
--- a/apps/compiler/includes/functions.sh
+++ b/apps/compiler/includes/functions.sh
@@ -90,7 +90,9 @@ function comp_configure() {
         OSOPTIONS=" -DMYSQL_ADD_INCLUDE_PATH=/usr/local/include -DMYSQL_LIBRARY=/usr/local/lib/libmysqlclient.dylib -DREADLINE_INCLUDE_DIR=/usr/local/opt/readline/include -DREADLINE_LIBRARY=/usr/local/opt/readline/lib/libreadline.dylib -DOPENSSL_INCLUDE_DIR=/usr/local/opt/openssl@3/include -DOPENSSL_SSL_LIBRARIES=/usr/local/opt/openssl@3/lib/libssl.dylib -DOPENSSL_CRYPTO_LIBRARIES=/usr/local/opt/openssl@3/lib/libcrypto.dylib "
         ;;
       msys*)
-        OSOPTIONS=" -DMYSQL_INCLUDE_DIR=C:\tools\mysql\current\include -DMYSQL_LIBRARY=C:\tools\mysql\current\lib\mysqlclient.lib "
+        if [[ -d "C:/tools/mysql/current/include" ]]; then
+            OSOPTIONS=" -DMYSQL_INCLUDE_DIR=C:/tools/mysql/current/include -DMYSQL_LIBRARY=C:/tools/mysql/current/lib/mysqlclient.lib "
+        fi
         ;;
     esac
 
--- a/apps/installer/includes/os_configs/windows.sh
+++ b/apps/installer/includes/os_configs/windows.sh
@@ -26,4 +26,4 @@ choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  cmake.install -y --insta
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  visualstudio2022-workload-nativedesktop
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  openssl --force --version=3.6.2
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  boost-msvc-14.3 --force --version=1.87.0
-choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  mysql --force --version=8.4.9
+choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  mysql --force --version=8.4.6
--- a/apps/installer/includes/os_configs/windows.sh
+++ b/apps/installer/includes/os_configs/windows.sh
@@ -26,8 +26,6 @@ choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  cmake.install -y --insta
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  visualstudio2022-workload-nativedesktop
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  openssl --force --version=3.6.2
 choco install -y --skip-checksums "${INSTALL_ARGS[@]}"  boost-msvc-14.3 --force --version=1.87.0
-if [[ $CONTINUOUS_INTEGRATION ]]; then
-    choco install -y --skip-checksums --no-progress mysql
-else
-    choco install -y --skip-checksums mysql --force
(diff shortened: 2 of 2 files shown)


juancarlosdiazpiaggio (commented):
Hi, how are you!

This change is upgrading the database to 9.6 since you are letting chocolatey decide which version install.

Image

I think changing the db version is a decision to make.

pangolp:
> I think changing the db version is a decision to make.

That wasn't the intention, therefore, it's not finding an integrated version. The else clause of the condition is being executed. We'll have to keep looking for another alternative then...

juancarlosdiazpiaggio:
Both do the same with the progress or no-progress, which is "let chocolatey decide which mysql version install".

if [[ $CONTINUOUS_INTEGRATION ]]; then
    choco install -y --skip-checksums --no-progress mysql
else
    choco install -y --skip-checksums mysql --force
fi

The problem I see here is that the provisioning on the windows vm depends on an external source (chocolatey package manager). Imo it will be nice to look for alternatives to avoid that dependency and have a more controled way on managing the mysql version.

pangolp:
> Both do the same with the progress or no-progress, which is "let chocolatey decide which mysql version install".
> 
> if [[ $CONTINUOUS_INTEGRATION ]]; then choco install -y --skip-checksums --no-progress mysql else choco install -y --skip-checksums mysql --force fi
> 
> The problem I see here is that the provisioning on the windows vm depends on an external source (chocolatey package manager). Imo it will be nice to look for alternatives to avoid that dependency and have a more controled way on managing the mysql version.

We'll continue investigating then...

pangolp:
From what I see in the log, there is currently no mention of a specific version, so now yes, you should use the one that is pre-installed, as previously intended, before confirming that you were using version 9.x

juancarlosdiazpiaggio (commented):
The initial problem is the package not being found when using chocolatey package manager. 

The solution that you just implemented is to completely ignore the chocolatey installation and use the mysql previously installed when the virtual machine was provisioned, right? This is a check to see if I get it correctly.

Where is the virtual machine configuration that installs mysql? What happens if the mysql installation is not there? It will use chocolatey again and go to 9.6 version, right? Can we ensure that the previous installation of mysql will be there?

Thanks for your time!

pangolp:
> chocolatey

It would be good to see if anyone has more information, but initially, I think so. At least, checking the condition, if it's not found, I should install the one Chocolatey recommends in this case, at least until the CDN is accessible again, and the version can be manually reset.

juancarlosdiazpiaggio:
What gets me thinking is what version is previously installed in the machine, right? Because I don't know how we can say that. And for the look of the code, it seems like it's also a chocolatey installation, which means that it will be an old installation. It's like postponing the real issue.

If that configuration depends on a previous installed version, why we are trying to install it again? The provisioning should happen only once, and the configuration should be in only one file.

pangolp:
> What gets me thinking is what version is previously installed in the machine, right? Because I don't know how we can say that. And for the look of the code, it seems like it's also a chocolatey installation, which means that it will be an old installation. It's like postponing the real issue.
> 
> If that configuration depends on a previous installed version, why we are trying to install it again? The provisioning should happen only once, and the configuration should be in only one file.

This is where my possibility of providing a solution ends. Outside of this change, I wouldn't know what other solution to offer at the moment. Since there are few Windows users and at least I, in particular, install the dependencies manually.

sudlud:
Thanks for digging into this, @pangolp, and for the detailed write-up.

The catch is that the Windows CI has moved on since this was opened. The passing check here is from June 4 on an older runner image. On the current image, master fails earlier at the `visualstudio2022-workload-nativedesktop` Chocolatey step, before the MySQL line is ever reached, so this change on its own no longer turns the build green.

Because Windows CI is currently red for that other reason, we also can't actually verify this PR right now. A green run isn't possible until the build is fixed upstream of the MySQL step. So realistically step 1 is a separate PR that gets Windows CI passing again (the VS workload step first), and this PR would just sit stale until then.

There's also a side effect worth avoiding: dropping `--version=8.4.9` means non-CI users running `acore.sh install-deps` would get whatever Chocolatey serves (currently 9.x), which we don't validate.

As @juancarlosdiazpiaggio pointed out, the real problem is the fragile Chocolatey/CDN provisioning, and that's bigger than this PR. Given all of the above I'm going to close this one for now. Would you be up for a fresh PR against current master that gets Windows CI green again and keeps a pinned, known-good MySQL 8.4.x for the non-CI install? Happy to help review it.

Thanks again for the effort here.

PR #26012, opened 2026-05-28
fix(Core/Player): keep spectral gryphon mount when relogging as a ghost

itsFranc:
## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> While the use of AI tools when preparing pull requests is not prohibited, contributors must clearly disclose when such tools have been used and specify the model involved.
> 
> Contributors are also expected to fully understand the changes they are submitting and must be able to explain and justify those changes when requested by maintainers.

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.
Claude, ask mainly debugging and advising.

## Issues Addressed:

- Closes #26007 

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [x] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

[...]

PR #26094, opened 2026-06-07
fix: ghost gryphon disappears after relog when dead in Northrend (#26007)

### LINKED ISSUE #26007: missing gryphon when you die and then reload
### Current Behaviour

when u release in icecrown or any other places like stormpeaks then reload you will lose the gryphon mount but u can fly like .gm fly on 
test vid: https://dl.dropboxusercontent.com/scl/fi/8i6l855v2vfk8debfhdb4/Wow_6spKHSSLdN.mp4?rlkey=napvgms6txelqk1t7jk2v08pl&dl=0

### Expected Behaviour

you should never lose the gryphon

### Source

none

- .tele icecrown
- .die on ur self 
- release
- reload while in release
- observe 

### Extra Notes

nope

### AC rev. hash/commit

AzerothCore rev. 1cccafbe0c1c 2026-05-28 08:17:39 +0200 (master branch) (Unix, Release, Static)

### Operating system

win 10

### Custom changes or Modules

_No response_

Chinoske:
## Changes Proposed

Fixes #26007 — spectral gryphon (spell 55164) disappears after relogging while dead in Northrend.

This PR proposes changes to:
- [x] Core ()

### Update

The mount-visual part of this fix is now obsolete: #26012 (already merged into master) fixed the same root cause by deleting the redundant re-cast of spells 8326 (Ghost) / 20584 (Wisp transform) in `HandlePlayerLoginFromDB` — that re-cast was what cleared the client's mount display field on relog. The aforementioned aura is already correctly persisted/reloaded via the standard `character_aura` save/reload mechanism, so no explicit re-cast is needed at all.

I confirmed live that current master, with only #26012's fix, now displays the spectral gryphon mount correctly on relog with no extra code required. I've dropped the deferred remove/reapply/teleport workaround from this PR accordingly and rebased onto master.

### Remaining fix (CharacterHandler.cpp)

**Corpse reclaim delay re-send**
`SMSG_CORPSE_RECLAIM_DELAY` is sent on death but not on relog, so the penalty countdown timer disappears after logging back in while dead. Uses `CalculateCorpseReclaimDelay(true)`, which returns the **remaining** time in milliseconds (accounting for ghost time + tier delay).

**AI-assisted Pull Requests**

## Issues Addressed

Fixes #26007 (corpse reclaim delay portion; the mount-visual portion is fixed by #26012)

## Tests Performed

## How to Test the Changes

1. Die in Northrend (Icecrown or Storm Peaks).
2. Relog while dead — spectral gryphon should still be visible (fixed by #26012, verified still working after this rebase).
3. Die twice (penalty stacked) → relog → corpse reclaim countdown shows correct remaining time.
4. Resurrect at corpse after delay expires — works normally.


src/server/game/Handlers/CharacterHandler.cpp
```diff
@@ -958,6 +959,40 @@ void WorldSession::HandlePlayerLoginFromDB(LoginQueryHolder const& holder)
             pCurrChar->CastSpell(pCurrChar, 20584, true, 0); // auras SPELL_AURA_INCREASE_SPEED(+speed in wisp form), SPELL_AURA_INCREASE_SWIM_SPEED(+swim speed in wisp form), SPELL_AURA_TRANSFORM (to wisp form)
 
         pCurrChar->CastSpell(pCurrChar, 8326, true, 0);     // auras SPELL_AURA_GHOST, SPELL_AURA_INCREASE_SPEED(why?), SPELL_AURA_INCREASE_SWIM_SPEED(why?)
+
+        // The ghost re-cast clears the mount display on the client (e.g. spectral gryphon in
+        // Northrend). Two nested deferred events replicate the sequence from the Lua workaround
+        // that confirmed to work: remove at 300ms, then recast+teleport 100ms later. The
+        // teleport to the same position forces the client to refresh the mount visual.
+        pCurrChar->m_Events.AddEventAtOffset([pCurrChar]
```
sogladev: This checks if there's a `SPELL_AURA_MOUNTED`, and recasts it a bit later to ensure it shows up properly?

Is the correct fix to write the initial application of the mount at the correct time, rather than this remove + re-apply hack?

Chinoske: I even tried it before uploading it, besides it doesn't let you fly without a mount.

PkllonG:
image
This is the correct way to do it; the PR is very hack.

Chinoske:
I tested the `spell_area` approach suggested by @PkllonG — removing the deferred events and relying solely on `HandleAuraSpecificMods` to auto-cast spell 55164 when 8326 is applied. The entries for spell 55164 already exist in the base DB (`spell_area` with `aura_spell = 8326` for areas 67, 210, 4161, 4197), so in theory the mechanism should handle it. However, after testing in-game the spectral gryphon stopped appearing entirely after relog while dead in Northrend — the bug came back.

The current implementation with the deferred remove + recast + teleport sequence is the approach that actually works in practice. I'm open to a cleaner solution if someone can pinpoint exactly why the `spell_area` path doesn't fire correctly on relog, but until then this hack produces the correct behavior.

sogladev:
> I tested the `spell_area` approach suggested by @PkllonG — removing the deferred events and relying solely on `HandleAuraSpecificMods` to auto-cast spell 55164 when 8326 is applied. The entries for spell 55164 already exist in the base DB (`spell_area` with `aura_spell = 8326` for areas 67, 210, 4161, 4197), so in theory the mechanism should handle it. However, after testing in-game the spectral gryphon stopped appearing entirely after relog while dead in Northrend — the bug came back.
> 
> The current implementation with the deferred remove + recast + teleport sequence is the approach that actually works in practice. I'm open to a cleaner solution if someone can pinpoint exactly why the `spell_area` path doesn't fire correctly on relog, but until then this hack produces the correct behavior.

The most robust method would be to look at a sniff and compare if there's a mismatch with packets. AC may be sending the packet too early? Have you tried sending the aura at a later time?

Chinoske:
Thanks for the suggestion @sogladev — tested sending the aura at a later time and it pointed in the right direction.

After investigation: `HandleAuraSpecificMods` correctly applies spell 55164 (Spectral Gryphon) when 8326 is cast at login, but the client discards the mount visual during its initial load sequence. The key insight is that `UNIT_FIELD_MOUNTDISPLAYID` needs to be resynced to the client after it finishes loading — a same-position teleport forces `HandleMoveWorldportAckOpcode` which sends the full `SMSG_UPDATE_OBJECT` with the correct display ID.

The previous nested events have been replaced with a single deferred event at 400ms that:
1. Removes any existing mount aura (so `HasAura` = false and spell_area can reapply)
2. Calls `UpdateAreaDependentAuras` — this uses the existing `spell_area` entries (spell 55164/55173 tied to `aura_spell = 8326` for areas 67, 210, 4161, 4197), no hardcoded spell IDs
3. Teleports to the same position to force the client visual resync

No more nested lambdas, no hardcoded mount spell IDs.

Chinoske:
Did a full review of the current diff (post the single-deferred-event refactor). No further changes needed from me — a few things I checked carefully before concluding that:

1. **The deferred lambda capturing \`pCurrChar\` by raw pointer** — this pattern (long-lived raw \`Player*\`) is normally something to flag, but verified \`m_Events\` is a member of the \`Player\`/\`WorldObject\` itself, and \`EventProcessor::~EventProcessor()\` calls \`KillAllEvents(true)\`, destroying pending events without executing them. So if the player is destroyed before the 400ms elapse, the event is destroyed alongside it and never fires — no dangling-pointer risk here, this is the standard safe pattern for "schedule something on myself later."
2. **Checked for interaction with #26430** (my own open PR, also touching \`UpdateAreaDependentAuras\` for the Dalaran no-fly-zone fix): traced the order of operations — this PR's \`RemoveAurasByType(SPELL_AURA_MOUNTED)\` runs *before* calling \`UpdateAreaDependentAuras\`, so by the time my PR's added mount-strict-check runs (early in that function), there's no mount aura present to act on; the \`spell_area\`-based reapply (later in the same function) is unaffected. No conflict if both get merged.
3. **\`CalculateCorpseReclaimDelay(true)\`** — confirmed in \`Player.cpp\` that \`load=true\` returns the actual remaining delay (\`expected_time - now\`, in ms), and that \`LoadCorpse()\` already ran earlier in \`HandlePlayerLoginFromDB\` by the time this runs, so the corpse data it depends on is available.

@sogladev's inline comment about the "remove + re-apply hack" looks like it's from before the single-deferred-event refactor (the reply right below it, about trying \`spell_area\` alone and it not working, plus your later comment about "sending the aura at a later time" leading to the current approach) — probably just a stale thread GitHub didn't auto-resolve when the diff changed underneath it.

Chinoske:
Rebased onto master and dropped the mount-recovery workaround — #26012 (already merged) independently fixed the same gryphon-mount bug at the root, by deleting the redundant Ghost/Wisp spell re-cast in `HandlePlayerLoginFromDB`. Confirmed live that master alone now fixes the mount visual with no extra code. This PR now only carries the corpse reclaim delay re-send fix, which #26012 doesn't touch.

PR #26096, opened 2026-06-07
fix(Core/Pet): restore Fel Intelligence autocast state across mount/dismount

### LINKED ISSUE #26079: [Warlock] - Felhunter's Fel Intelligence inconstancy on the auto-cast behaviour
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/9631

Triage Notes: 

Be a warlock level 80
- `.learn all my class`
- `.additem 6265 20`
- `.cheat cast`
- `.cheat power`
- `.cast 691`
- See if it's on or off the auto cast for `Fel Intelligence` if is off, turn it on, now mount and dismount and see the inconstancy 
- `.cast 17481` (a mount spell)

Original Post Below: 

---

### Current Behaviour

Fel intelligence is no longer auto cast when ever the pet is summoned, reappears after dismounting
this means you have to manually click the ability every time

### Expected Behaviour

Auto cast being enabled should auto cast Fel Intelligence every time the pet comes out

Expected behaviour from ability having auto cast

1. Summon felhunter
2. right click Fel intelligence so its on auto cast and left click to enable it
3. mount up
4. dismount
5. see that fel intelligence is no longer active

### Extra Notes

_No response_

### Customizations

_No response_

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/f32f6d422f125cdebe10201f00941ea579d98a5a

### Operating system

Ubuntu 22.04

### Modules

### Server

ChromieCraft

Chinoske:
## Changes Proposed

Fixes #26079 — Fel Intelligence autocast is reset to OFF every time the Felhunter is resummoned (e.g. after a mount/dismount cycle).

`Pet::CleanupActionBar()` determines whether to enable autocast by reading `ab->GetType()` from the action bar entry. However, `LoadPetActionBar()` may force a slot to `ACT_PASSIVE` during its validation pass, overwriting the correct active state that `_LoadSpells()` had just loaded from the `pet_spell` table. The result is that every resummon causes `CleanupActionBar()` to call `ToggleAutocast(false)` for Fel Intelligence.

**`Pet::CleanupActionBar()`** — use the active state from `m_spells` (populated by `_LoadSpells()`) instead of the action bar type. Falls back to the action bar type for spells not in `m_spells`.

**`Pet::addSpell()`** — when a Felhunter learns Fel Intelligence for the first time (state `ACT_DECIDE`), default to `ACT_ENABLED` instead of `ACT_DISABLED`, matching retail behavior. Uses `GetFirstSpellInChain` to cover all 5 ranks (54424–57567).

This PR proposes changes to:
- [x] Core (`Pet.cpp`)

**AI-assisted Pull Requests**

## Issues Addressed

Fixes #26079

## Tests Performed

## How to Test the Changes

1. Summon Felhunter — Fel Intelligence autocast should be ON by default without any manual setup.
2. Enable Fel Intelligence autocast, mount, dismount — autocast remains ON.
3. Disable Fel Intelligence autocast, mount, dismount — autocast remains OFF.
4. Relog while pet is active — autocast state is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pet spell autocast management. The pet action bar now correctly preserves autocast states for spells the pet actually knows, preventing autocast settings from being incorrectly reset during routine maintenance operations.


src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -1804,7 +1804,13 @@ bool Pet::addSpell(uint32 spellId, ActiveStates active /*= ACT_DECIDE*/, PetSpel
     if (active == ACT_DECIDE)                               // active was not used before, so we save it's autocast/passive state here
     {
         if (spellInfo->IsAutocastable())
-            newspell.active = ACT_DISABLED;
+        {
+            // Fel Intelligence should default to autocast on Felhunter
+            if (GetEntry() == NPC_FELHUNTER && sSpellMgr->GetFirstSpellInChain(spellId) == 54424)
```
heyitsbench: This feels very hackfix-y. I know that's not very helpful feedback, but I don't know a better solution than this offhand.

sogladev: Is enabling autocast by default correct? CMangos sets these values to ENABLED, so that may be the right way to fix this.

Though there may be side effects as seen in the 2 fixup commits for this

https://github.com/cmangos/mangos-wotlk/commit/a30c3490b492bbd0418cf6ac937ba6491b942131

https://github.com/cmangos/mangos-wotlk/commit/056f57286ea4503345f85483ab450d969f2f596f

https://github.com/cmangos/mangos-wotlk/commit/03394d9f78fd1e5fa2592b98d3b32157a44e28a6

Chinoske:
Removed the `addSpell` change based on reviewer feedback.

The hardcoded `NPC_FELHUNTER + spell 54424` check was too specific and would need to be replicated for every other spell with a non-default autocast preference. The CMangos commits linked by @sogladev show that making the default broader (all autocastable spells → `ACT_ENABLED`) introduced side effects requiring two follow-up fixes, so that direction is risky without more research.

The `CleanupActionBar` fix alone fully addresses the reported bug (#26079): autocast state is now read from `m_spells` (populated by `_LoadSpells()` from the DB) instead of the action bar slot type, which was being overwritten by `LoadPetActionBar()`'s validation pass. Mount/dismount and relog now preserve the player's chosen autocast state correctly.

The "first summon defaults to ON" behaviour is a separate concern that can be addressed in a dedicated PR with proper research into which spells should default to enabled and why.

Chinoske:
@heyitsbench fair feedback — took a closer look at what specifically felt off. Two things were contributing:

1. \`HasSpell(ab->GetAction())\` and the separate \`m_spells.find(...)\` right after were two lookups against the same map for the same key.
2. Once the spell is confirmed present, the "not found in \`m_spells\`" fallback (\`ab->GetType() == ACT_ENABLED\`) is dead code — a spell that's on the action bar and passes the \`hasSpell\` check is always in \`m_spells\` by construction, so that ternary always looked like it had two sources of truth when it really only had one.

Merged into a single lookup and dropped the fallback, plus added a comment on *why* \`m_spells\` should win over the action bar entry's own type here: checked the actual load order (\`Pet::LoadPetFromDB\`) and confirmed \`CharmInfo::LoadPetActionBar()\` runs before \`_LoadSpells()\`, so the action bar's copy of the autocast flag can genuinely be stale relative to \`pet_spell\` by the time \`CleanupActionBar()\` runs later in that same sequence. That ordering is also *why* the fix has to live here rather than in \`LoadPetActionBar()\` itself — \`m_spells\` isn't populated yet at that point.

No functional change, just consolidating the duplicate lookup and making the reasoning explicit instead of implicit. Let me know if that addresses the concern or if there's something more specific still bugging you about the approach.

PR #26051, opened 2026-06-01
feat(Core/Chat): Add chat filter with Aho-Corasick matching

Kitzunu:
## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [x] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).
image

### AI-assisted Pull Requests

> [!IMPORTANT]
> While the use of AI tools when preparing pull requests is not prohibited, contributors must clearly disclose when such tools have been used and specify the model involved.
> 
> Contributors are also expected to fully understand the changes they are submitting and must be able to explain and justify those changes when requested by maintainers.

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any. Claude Opus 4.7

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/26028

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [x] Tested in-game by the author.

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided in the linked issue

1. Enable config
2. .filteredwords add $word
3. say said $word

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #26099, opened 2026-06-07
fix(Conf): improve ChatFilter config documentation

Chinoske:
## Changes

Follow-up documentation fixes for #26051 and #26053.

**ChatFilter** (`ChatFilter.Whisper / .Say / .Yell / .Emote`):
- Fix typo `Disbaled` → `Disabled`
- Rename header comment to list all four keys explicitly
- Document the default value of each key individually instead of a generic `0`/`1` description
- Reword "Blizzlike is only whispers are enabled." → "Blizzlike: only Whisper filtering is enabled."

**Achievement.RealmFirstRaceLimitOnePerCharacter**:
- Expand the description to clarify it limits race-specific 'Realm First!' achievements and prevents abuse via race/faction change service

### DIFF
--- a/src/server/apps/worldserver/worldserver.conf.dist
+++ b/src/server/apps/worldserver/worldserver.conf.dist
@@ -2262,9 +2262,8 @@ Achievement.RealmFirstKillWindow = 60
 
 #
 #    Achievement.RealmFirstRaceLimitOnePerCharacter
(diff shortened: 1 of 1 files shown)


sogladev (approved):
Thank you for fixing this!

PR #26106, opened 2026-06-08
fix(Scripts/Ulduar): prevent Mimiron from evading during outro and fix chest despawn

Chinoske:
https://github.com/user-attachments/assets/88026554-f3c3-408f-a146-38b6acf3e38f

## Changes Proposed

Three bugs in the Mimiron encounter in Ulduar:

**1. Outro evade bug (`boss_mimiron.cpp`)**

After Mimiron is defeated, `SetFaction(FRIENDLY)` is called as part of the RP sequence. On the next `UpdateAI` tick, `SelectTargetFromPlayerList` returns `nullptr` (no hostile players), causing `EnterEvadeMode` to fire mid-outro — resetting the boss back to hostile and interrupting the RP.

Fix: introduce an `_outroStarted` bool flag set to `true` when `EVENT_FINISH` triggers. Both the out-of-combat evade check in `UpdateAI` and `EnterEvadeMode` are guarded with `!_outroStarted`, allowing the outro to complete without interruption.

---

**2. Chest spawn / loot loop fix (`instance_ulduar.cpp` + SQL)**

The chest was previously spawned via `Map::SummonGameObject()`. That path does not register the GO in any summon list and doesn't properly handle the loot → `GO_JUST_DEACTIVATED` → despawn lifecycle, causing items to regenerate infinitely on every open.

After multiple rounds of testing, the correct fix required two changes working together:

**C++ (`instance_ulduar.cpp`):**
- Changed from `instance->SummonGameObject()` to `mimiron->SummonGameObject()` with `respawnTime = DAY` and `GO_SUMMON_TIMED_DESPAWN`.
- Using `GO_SUMMON_TIMED_DESPAWN` keeps the chest independent of Mimiron's lifetime (no `Unit::AddGameObject()` call), so it persists through the outro.
- Using `respawnTime = DAY` triggers `go->SetSpellId(1)` inside `WorldObject::SummonGameObject`, which is required for the chest to reach `Delete()` in `GO_JUST_DEACTIVATED` (line 863 of `GameObject.cpp`: `else if (GetOwnerGUID() || GetSpellId())`).

**SQL (`data/sql/updates/pending_db_world/2026_06_17_00_gameobject_mimiron_chest_consumable.sql`):**
- Sets `data3 = 1` (consumable) on all four Mimiron chest entries (194789, 194956, 194957, 194958).
- Without this, `GameObject::Update()` in `GO_ACTIVATED` state calls `SetLootState(GO_READY)` for non-consumable chests with no restock timer (line 780 of `GameObject.cpp`), regenerating loot indefinitely.

The two fixes are interdependent:
- SQL alone: `GetSpellId() = 0` → `GO_JUST_DEACTIVATED` falls through to `SetLootState(GO_READY)` → still loops.
- C++ alone (with `respawnTime = DAY`): `GetSpellId() = 1` → `Delete()` triggered, but loot also regenerates in `GO_ACTIVATED` if `consumable = 0`.

This PR proposes changes to:
- [x] Scripts (`boss_mimiron.cpp`, `instance_ulduar.cpp`)
- [x] DB/world (`gameobject_template` consumable flag for Mimiron chests)

**AI-assisted Pull Requests**

## Issues Addressed

Related to #26107

## Tests Performed

## How to Test the Changes

1. Enter Ulduar and engage Mimiron (normal or hard mode, 10 or 25 man).
2. Defeat Mimiron — the outro RP sequence should play completely without him resetting or turning hostile again.
3. Verify the loot chest spawns and stays in the world after the outro ends.
4. Loot all items from the chest — the chest should despawn immediately once fully looted.
5. Verify the chest does not respawn and items do not regenerate.

### DIFF
--- a/data/sql/updates/pending_db_world/2026_06_17_00_gameobject_mimiron_chest_consumable.sql
+++ b/data/sql/updates/pending_db_world/2026_06_17_00_gameobject_mimiron_chest_consumable.sql
@@ -1,4 +1,0 @@
--- Mimiron chest entries (194789/194956/194957/194958) have chest.consumable (data3) = 0,
--- which causes the chest to restock loot indefinitely after being opened.
--- Set data3 = 1 so the chest is consumed (deleted) once fully looted.
-UPDATE `gameobject_template` SET `data3` = 1 WHERE `entry` IN (194789, 194956, 194957, 194958);
--- a/data/sql/updates/pending_db_world/2026_06_17_00_gameobject_mimiron_chest_consumable.sql
+++ b/data/sql/updates/pending_db_world/2026_06_17_00_gameobject_mimiron_chest_consumable.sql
@@ -1,0 +1,4 @@
+-- Mimiron chest entries (194789/194956/194957/194958) have chest.consumable (data3) = 0,
+-- which causes the chest to restock loot indefinitely after being opened.
+-- Set data3 = 1 so the chest is consumed (deleted) once fully looted.
+UPDATE `gameobject_template` SET `data3` = 1 WHERE `entry` IN (194789, 194956, 194957, 194958);
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_mimiron.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_mimiron.cpp
@@ -270,10 +270,12 @@ struct boss_mimiron : public BossAI
             instance->SetBossState(BOSS_MIMIRON, DONE);
 
         _isEvading = false;
+        _outroStarted = false;
     }
 
     void Reset() override
     {
+        _outroStarted = false;
         _hardmode = false;
         _berserk = false;
         _achievProximityMine = false;
@@ -362,7 +364,7 @@ struct boss_mimiron : public BossAI
 
     void UpdateAI(uint32 diff) override
     {
-        if (!me->IsInCombat())
+        if (!me->IsInCombat() && !_outroStarted)
         {
             _outOfCombatTimer += diff;
             if (_outOfCombatTimer >= 10000)
@@ -374,11 +376,14 @@ struct boss_mimiron : public BossAI
             return;
         }
 
-        Position p = me->GetHomePosition();
-        if (me->GetExactDist(&p) > 80.0f || !SelectTargetFromPlayerList(150.0f))
+        if (!_outroStarted)
         {
-            EnterEvadeMode(EVADE_REASON_OTHER);
-            return;
+            Position p = me->GetHomePosition();
+            if (me->GetExactDist(&p) > 80.0f || !SelectTargetFromPlayerList(150.0f))
+            {
+                EnterEvadeMode(EVADE_REASON_OTHER);
+                return;
+            }
         }
 
         events.Update(diff);
@@ -735,6 +740,7 @@ struct boss_mimiron : public BossAI
                         if (Creature* computer = me->SummonCreature(NPC_COMPUTER, 2746.7f, 2569.44f, 410.39f, 0.0f, TEMPSUMMON_TIMED_DESPAWN, 1000))
                             computer->AI()->Talk(TALK_COMPUTER_TERMINATED);
 
+                    _outroStarted = true;
                     events.Reset();
                     events.ScheduleEvent(EVENT_STAND_UP_FRIENDLY, 6s);
                 }
@@ -752,7 +758,7 @@ struct boss_mimiron : public BossAI
                 // spawn chest
                 if (uint32 chestId = (_hardmode ? RAID_MODE(GO_MIMIRON_CHEST_HARD, GO_MIMIRON_CHEST_HERO_HARD) : RAID_MODE(GO_MIMIRON_CHEST, GO_MIMIRON_CHEST_HERO)))
                 {
-                    if (GameObject* go = me->SummonGameObject(chestId, 2744.65f, 2569.46f, 364.397f, 0, 0, 0, 0, 0, 0))
+                    if (GameObject* go = me->SummonGameObject(chestId, 2744.65f, 2569.46f, 364.397f, 0, 0, 0, 0, 0, 0, true, GO_SUMMON_TIMED_DESPAWN))
                     {
                         go->ReplaceAllGameObjectFlags((GameObjectFlags)0);
                         go->SetLootRecipient(me->GetMap());
@@ -780,6 +786,8 @@ struct boss_mimiron : public BossAI
 
     void EnterEvadeMode(EvadeReason why) override
     {
+        if (_outroStarted)
+            return;
         if (_isEvading)
             return;
         _isEvading = true;
@@ -907,6 +915,7 @@ struct boss_mimiron : public BossAI
 
 private:
     bool _isEvading;
+    bool _outroStarted;
     bool _hardmode;
     bool _berserk;
     bool _achievProximityMine;
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_mimiron.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_mimiron.cpp
@@ -755,15 +755,6 @@ struct boss_mimiron : public BossAI
                 Talk(SAY_V07TRON_DEATH);
                 me->HandleEmoteCommand(EMOTE_ONESHOT_TALK);
                 instance->SetBossState(BOSS_MIMIRON, DONE);
-                // spawn chest
-                if (uint32 chestId = (_hardmode ? RAID_MODE(GO_MIMIRON_CHEST_HARD, GO_MIMIRON_CHEST_HERO_HARD) : RAID_MODE(GO_MIMIRON_CHEST, GO_MIMIRON_CHEST_HERO)))
-                {
-                    if (GameObject* go = me->SummonGameObject(chestId, 2744.65f, 2569.46f, 364.397f, 0, 0, 0, 0, 0, 0, true, GO_SUMMON_TIMED_DESPAWN))
-                    {
-                        go->ReplaceAllGameObjectFlags((GameObjectFlags)0);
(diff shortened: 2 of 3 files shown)


Fazzul:
Can confirm this worked:
image

Server Debug print:
```
AC> .server debug
AC> AzerothCore rev. 98e7a05caa68+ 2026-06-16 15:06:56 -0500 (master branch) (Unix, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 1.
Connection peak: 1.
Server uptime: 4 minute(s) 55 second(s)
|- Mean: 6ms
|- Median: 2ms
|- Percentiles (95, 99, max): 17ms, 50ms, 334ms
Using SSL version: OpenSSL 3.0.13 30 Jan 2024 (library: OpenSSL 3.0.13 30 Jan 2024)
Using Boost version: 1.83.0
Using CMake version: 3.28.3
Using MySQL version: 80046
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.17.0-35-generic
MMAPs status: Enabled
maps directory located in ./maps. Total size: 291014951 bytes
vmaps directory located in ./vmaps. Total size: 657797442 bytes
mmaps directory located in ./mmaps. Total size: 2192760616 bytes
Default DBC locale: enUS.
Latest CharacterDatabase update: 2026_06_01_00.sql
Latest WorldDatabase update: 2026_06_16_07.sql
LoginDatabase queue size: 0
WorldDatabase queue size: 0
No modules are enabled
```
Only changes were the ones in the MR

Nyeriah:
The instance should spawn the chest instead so it doesn't go away if the boss evades, resets or despawns (as objects get removed from world when the creature object is destroyed)

Chinoske:
Good point @Nyeriah. Updated — the chest is now spawned by the InstanceScript via `Map::SummonGameObject` in the `SetBossState(BOSS_MIMIRON, DONE)` handler, completely decoupled from the creature lifetime. Hardmode is queried from the boss AI (`GetData(1)`) at the moment DONE fires, while the creature is still alive.

```cpp
// instance_ulduar.cpp — SetBossState handler
if (type == BOSS_MIMIRON && state == DONE)
{
    bool hardmode = false;
    if (Creature* mimiron = GetCreature(BOSS_MIMIRON))
        hardmode = mimiron->AI()->GetData(1) != 0;
    uint32 chestId = hardmode
        ? RAID_MODE(GO_MIMIRON_CHEST_HARD, GO_MIMIRON_CHEST_HERO_HARD)
        : RAID_MODE(GO_MIMIRON_CHEST, GO_MIMIRON_CHEST_HERO);
    if (GameObject* go = instance->SummonGameObject(chestId, 2744.65f, 2569.46f, 364.397f, 0, 0, 0, 0, 0, 0))
    {
        go->ReplaceAllGameObjectFlags((GameObjectFlags)0);
        go->SetLootRecipient(instance);
    }
}
```

The chest spawn is removed from `boss_mimiron.cpp` entirely.

Fazzul:
Should all the keepers chests be done like this? I think I saw a similar thing when Freya de-spawned

Chinoske:
## Additional bug found: chest infinite loot / no despawn

During in-game testing a second bug was discovered: after killing Mimiron and looting the spawned chest, the chest **did not despawn** and items (including Emblems of Triumph) kept regenerating on every open — effectively granting infinite loot.

**Root cause:** The previous code called `Map::SummonGameObject()` with `respawnTime=0`. That variant does not register the GO in the creature's summon list and does not wire up the loot → `GO_JUST_DEACTIVATED` → despawn lifecycle that boss chests rely on.

**Fix applied:**
- Replaced `instance->SummonGameObject(...)` with `mimiron->SummonGameObject(...)` (`WorldObject` variant), which correctly tracks the GO in the summoner's list.
- Changed the summon type from `0` to `GO_SUMMON_TIMED_DESPAWN` so the GO despawns once fully looted.
- Since `mimiron` is now required for both the hardmode check and the summon call, the entire block is nested inside the `if (Creature* mimiron = GetCreature(BOSS_MIMIRON))` guard.

Hope this fix resolves it — will update after further testing.

Chinoske:
## Fix confirmed working ✓

After several rounds of debugging, the chest loot loop is now resolved. Summary of what was found:

### Root cause (two-part)

**Part 1 — Non-consumable chest (`gameobject_template.data3 = 0`):**
In `GameObject::Update()` during `GO_ACTIVATED` state (line 780):
```cpp
if (GetGOInfo()->chest.consumable == 0 && GameTime::GetGameTime() >= m_restockTime)
{
    m_lootState = GO_READY; // restock loot infinitely
}
```
All four Mimiron chest entries had `consumable = 0`, so loot regenerated on every open.

**Part 2 — Missing `SetSpellId` on the summoned GO:**
In `GameObject::Update()` during `GO_JUST_DEACTIVATED` (line 863):
```cpp
else if (GetOwnerGUID() || GetSpellId())
{
    SetRespawnTime(0);
    Delete(); // ← only path that removes the summoned chest
}
```
With `respawnTime = 0`, `WorldObject::SummonGameObject` never calls `go->SetSpellId(1)`, so `GetSpellId()` returns 0 and this block is skipped — the chest falls through to `SetLootState(GO_READY)` instead of being deleted.

### Final fix

- **C++**: `mimiron->SummonGameObject(..., DAY, true, GO_SUMMON_TIMED_DESPAWN)` — `respawnTime = DAY` ensures `SetSpellId(1)` is called; `GO_SUMMON_TIMED_DESPAWN` keeps the chest independent of Mimiron's lifetime.
- **SQL**: `UPDATE gameobject_template SET data3 = 1 WHERE entry IN (194789, 194956, 194957, 194958)` — marks chests as consumable so the restock path is never entered.

In-game test confirmed: chest spawns after kill, items are lootable once, chest despawns immediately when fully looted. No loot regeneration.

Chinoske:
@coderabbitai full review

Chinoske:
Refactor: replaced the manual `is25man = instance->GetDifficulty() == RAID_DIFFICULTY_25MAN_NORMAL` check with `instance->Is25ManRaid()` (the exact pattern this same file uses at line 1037, and what `boss_thorim.cpp` uses for its identical hard/normal chest-selection case). Small correctness bonus: `Is25ManRaid()` internally checks `_difficulty & RAID_DIFFICULTY_MASK_25MAN`, so it correctly covers 25-man heroic too — the original `==` comparison against only `RAID_DIFFICULTY_25MAN_NORMAL` would have silently treated that enum value as 10-man.

(First pushed a version using `RAID_MODE(...)` since that's what the original pre-move code in `boss_mimiron.cpp` used — turns out that macro lives on `ScriptedAI`/`BossAI`, not `InstanceScript`, so it wouldn't have compiled in this context. Caught it locally and pushed the corrected version instead.)

Compile-verified locally (both versions, so the mistake never made it further than my own build). Didn't do a full Mimiron kill retest since the change is a value-equivalent substitution for all currently reachable difficulties (10N/25N/10H — Ulduar doesn't have a 25H equivalent in WotLK) — happy to do a full encounter retest if preferred before merge.

Rorschach91:
CC: @Nyeriah

Nyeriah:
Can't update gameobject_template fields, they are sniffed
Instead set respawn time when the chest spawns to 7 days (like similar chests in other raids)

Chinoske:
@Nyeriah thanks — makes sense, fixed. Removed the SQL entirely and replaced it with the same pattern used by other raid boss chests in the codebase (Four Horsemen in Naxxramas, several ICC spots, Oculus): plain \`Map::SummonGameObject\` with \`respawnTime=0\`, then explicit \`SetRespawnTime(7 * DAY)\` right after spawning. Also switched back to \`instance->SummonGameObject\` (Map variant) since the \`WorldObject\`-variant \`GO_SUMMON_TIMED_DESPAWN\` summon type isn't needed anymore.

Will retest the full kill in-game to confirm the chest no longer restocks with this approach (haven't re-tested live yet after this change).

Chinoske:
Update: live-tested the \`SetRespawnTime(7 * DAY)\` version and it did **not** actually fix the restock bug — the chest still restocked after being fully looted. Traced through \`GameObject::Update()\`'s \`GO_JUST_DEACTIVATED\` handler to find out why:

\`\`\`cpp
bool isSummonedAndExpired = (GetOwner() || GetSpellId()) && m_respawnTime == 0;
if ((chest || goober) && !IsDespawnAtAction() && !isSummonedAndExpired)
    { /* restock path — the bug */ }
else if (GetOwnerGUID() || GetSpellId())
    { SetRespawnTime(0); Delete(); return; } // what we actually want
\`\`\`

Reaching the `Delete()` branch requires `GetSpellId() != 0` **and** `m_respawnTime == 0` at the same time. `SetRespawnTime(7 * DAY)` sets `m_respawnTime` to a nonzero value, which keeps `isSummonedAndExpired` false and routes right back into the restock branch — no respawn duration would ever reach `Delete()` this way.

Checked why Four Horsemen's chest (the reference pattern) actually works: its `gameobject_template.chest.consumable` is natively sniffed as `1` (confirmed via DB query on both `181366`/`193426`), so `IsDespawnAtAction()` is already true for it regardless of `m_respawnTime` — the `SetRespawnTime` call there isn't what's preventing the restock. Mimiron's 4 chest entries are natively sniffed `consumable=0`, so that reference pattern doesn't transfer over even though the code shape looks identical.

Fixed by calling `go->SetSpellId(1)` after summoning (a public setter, no DB touch) instead of `SetRespawnTime`, and not calling `SetRespawnTime` at all so `m_respawnTime` stays `0` — satisfies both `isSummonedAndExpired` conditions and reaches `Delete()`. Live-tested and confirmed: the chest now properly disappears once fully looted instead of restocking.

Nyeriah:
Sorry, the AI is getting too noisy. I think the context got too polluted. I have no such issue you described, maybe you have GM mode on when looting the chest. Anyway, this is what I have been trying to say: https://github.com/azerothcore/azerothcore-wotlk/pull/26498

PR #26189, opened 2026-06-14
fix(Core/Combat): After evade NPC not Attack

Unkn87:
## Changes Proposed:

This PR proposes changes to:
-  [ -] Core (units, players, creatures, game systems).

## Issues Addressed:
After evade NPC not Attack. Not all. I notice this bug in Darnassus NPC (guards and Ancien of War).
Image
and if i am attack one other (after evade) not attack:
Image

Revert some missing code from https://github.com/azerothcore/azerothcore-wotlk/pull/25516

### DIFF
--- a/src/server/game/AI/CreatureAI.cpp
+++ b/src/server/game/AI/CreatureAI.cpp
@@ -309,9 +309,12 @@ void CreatureAI::EngagementOver()
 
 void CreatureAI::JustExitedCombat()
 {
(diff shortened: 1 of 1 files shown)


blinkysc:
JustExitedCombat() must be no-op. If scripted then you can override, otherwise if smartAI just needs deliberate behavior.

EDIT: Its because GuardAI

blinkysc:
@Unkn87 please see https://github.com/azerothcore/azerothcore-wotlk/pull/26196 can give it a test too ;)

Issue #26193, opened 2026-06-14
[MAGE]Frostfire Bolt should deal damage against Crazed Water Spirit


LoturCC:
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9710

Triage Notes: 

reproduce:
- make a mage
- .learn all my class
- .go c i 16570
- cast frostfire bolt on the mob and observe
test video: https://dl.dropboxusercontent.com/scl/fi/krfglo9y20yaiurw7za8v/Wow_gZr7aujqrV.mp4?rlkey=elqoje5hhcthkuegjn0ex0ru6&dl=0

off source: https://www.wowhead.com/wotlk/spell=47610/frostfire-bolt

```Launches a bolt of frostfire at the enemy, causing 722 to 838 Frostfire damage, slowing movement speed by 40% and causing an additional 90 Frostfire damage over 9 sec. This spell will be checked against the lower of the target's Frost and Fire resists.```

Original Post Below: 

---
### Current Behaviour

Frostfire Bolt does not deal damage to Crazed Water Spirit (Frost-immune-NPC).

Video: https://youtu.be/cYoO4Erz7Bk

### Expected Behaviour

Frostfire Bolt should deal damage to Crazed Water Spirit.

Overall, it would be good to check other Frost- or Fire-immune NPCs, as the bug may not be limited to this particular NPC.

So against a water elemental that is Frost-immune but not Fire-immune:

Frostbolt → Immune
Fireball → Deals damage
Frostfire Bolt → Should deal damage

Against a Fire-immune NPC that is not Frost-immune:

Fireball → Immune
Frostbolt → Deals damage
Frostfire Bolt → Should deal damage

### Official Sources (that prove or disprove) the issue being reported

https://www.wowhead.com/wotlk/spell=47610/frostfire-bolt

Launches a bolt of frostfire at the enemy, causing 722 to 838 Frostfire damage, slowing movement speed by 40% and causing an additional 90 Frostfire damage over 9 sec. **_This spell will be checked against the lower of the target's Frost and Fire resists._**

### Steps to reproduce the problem

1. Go to https://www.wowhead.com/npc=16570/crazed-water-spirit
2. Attack Crazed Water Spirit with Frostfire bolt.
3. Observe.

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/df82d1be97ec25e909f581c1460990d893113ca1

### Operating system

[...]

PR #26206, opened 2026-06-15
fix(Spells): dual-school spells bypass single-school immunity incorrectly

Chinoske:
https://github.com/user-attachments/assets/23412b11-ffd7-4395-b829-707ae42edf64

## Background

Spells with a dual school mask (e.g. Frostfire Bolt — `FROST|FIRE`) should only be fully blocked by school immunity if the target is immune to **all** of the spell's schools. If the target is immune to only one of the schools, the spell should still deal damage via the other component.

> Retail reference: https://www.wowhead.com/wotlk/spell=47610/frostfire-bolt — *"Checked against the lower of the target's Frost and Fire resistances"*

## Bug

**Frostfire Bolt** (spell 47610, school mask `FROST|FIRE`) deals no damage against frost-immune NPCs (e.g. Crazed Water Spirit, NPC 16570), despite the fire component being unresisted.

Reported in: #26193

## Root cause

In `Unit::IsImmunedToSpell(SpellInfo const*, Unit const*, SpellSchoolMask)`, the school immunity check used an **ANY-overlap** condition:

```cpp
if (!(itr->first & spellSchoolMask))
    continue;
```

For Frostfire Bolt (`FROST|FIRE`) against a FROST-only immune target:
- `!(FROST & FROST|FIRE)` → `false` → does **not** skip → returns `true` (immune) ❌

This is inconsistent with every other immunity check in the codebase (`IsImmunedToSpell` line 960, `IsImmunedToDamage`, `IsImmunedToSchool`), which all require the immunity mask to cover **all** schools in the spell.

## Fix

The `creature_immunities` table applies school immunities as **one entry per school bit** (e.g. a fire+frost immune creature gets two separate entries: key=`FIRE` and key=`FROST`, not one combined key=`FIRE|FROST`). A naive "all-schools-covered" single-entry check would therefore break creatures that are immune to both schools via separate entries.

Instead of checking each entry individually, the fix **accumulates** all valid (non-pierceable) school immunity entries that overlap with the spell's schools, then checks whether their **combined mask** covers the full spell school mask:

```cpp
// Before — ANY overlap → immediate immune (too permissive for multi-school spells)
if (!(itr->first & spellSchoolMask))
    continue;
return true;

// After — accumulate, then check combined coverage
if (!(itr->first & spellSchoolMask))
    continue;
// ... IgnoresSchoolImmunityFromFriendlyCaster / CanPierceImmuneAura checks ...
coveredSchoolMask |= itr->first;
// end of loop:
if ((coveredSchoolMask & spellSchoolMask) == spellSchoolMask)
    return true;
```

This correctly handles all cases:
| Scenario | Result |
|---|---|
| Frost-only immune vs Frostfire Bolt (`FROST\|FIRE`) | covered=`FROST`, `FROST != FROST\|FIRE` → **not immune** ✓ |
| Fire+Frost immune (separate entries) vs Frostfire Bolt | covered=`FROST\|FIRE` → **immune** ✓ |
| Fire+Frost immune (combined entry) vs Frostfire Bolt | covered=`FROST\|FIRE` → **immune** ✓ |
| Fire-only immune vs Fireball (`FIRE`) | covered=`FIRE`, `FIRE == FIRE` → **immune** ✓ |
| Anti-Magic Shell (all magic) vs any magic spell | covered covers all → **immune** ✓ |

## Test plan

1. Find a frost-immune NPC (e.g. Crazed Water Spirit, NPC 16570)
2. Cast Frostfire Bolt at it
3. Verify it now deals damage (fire component) instead of showing **Immune**
4. Verify single-school fire spells (Fireball) are still fully blocked by fire-immune NPCs
5. Verify Anti-Magic Shell still blocks all magic spells including Frostfire Bolt
6. Verify a creature immune to both Fire and Frost (e.g. Emperor Vek'nilash, entry 15275) is still fully immune to Frostfire Bolt

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spell immunity logic for multi-school spells. Units now correctly evaluate immunity by combining overlapping immunity sources, so protection is applied only when the combined coverage fully includes every spell school (while preserving existing filtering behavior like exclusions and pierce-immune-aura handling).

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10092,7 +10092,7 @@ bool Unit::IsImmunedToSpell(SpellInfo const* spellInfo, Unit const* caster, Spel
                     continue;
 
                 SpellInfo const* immuneSpellInfo = sSpellMgr->GetSpellInfo(itr->second);
-                if (!(itr->first & spellSchoolMask))
+                if ((itr->first & spellSchoolMask) != spellSchoolMask)
                     continue;
 
                 if (IgnoresSchoolImmunityFromFriendlyCaster(caster, itr->second, immuneSpellInfo))
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10085,30 +10085,26 @@ bool Unit::IsImmunedToSpell(SpellInfo const* spellInfo, Unit const* caster, Spel
     {
         if (spellSchoolMask != SPELL_SCHOOL_MASK_NONE)
         {
+            // Accumulate valid school-immunity bits that overlap the spell's
+            // schools. A multi-school spell (e.g. Frostfire Bolt) can be
+            // covered by separate per-school entries (as produced by
+            // creature_immunities), so the combined mask must be checked
+            // instead of requiring any single entry to cover every school.
             SpellImmuneContainer const& schoolList = m_spellImmune[IMMUNITY_SCHOOL];
-            // Accumulate all valid school immunity bits that overlap with the spell's schools.
-            // A multi-school spell (e.g. Frostfire Bolt) may have its schools covered by
-            // separate per-school entries (as produced by creature_immunities), so we must
-            // check the combined mask rather than requiring any single entry to cover all schools.
             uint32 coveredSchoolMask = 0;
-            for (auto itr = schoolList.begin(); itr != schoolList.end(); ++itr)
+            for (auto const& [immunitySchoolMask, immunityAuraId] : schoolList)
             {
-                if (itr->second == spellInfo->Id)
+                if (immunityAuraId == spellInfo->Id || !(immunitySchoolMask & spellSchoolMask))
                     continue;
 
-                if (!(itr->first & spellSchoolMask))
+                SpellInfo const* immuneSpellInfo = sSpellMgr->GetSpellInfo(immunityAuraId);
+                if (IgnoresSchoolImmunityFromFriendlyCaster(caster, immunityAuraId, immuneSpellInfo) ||
+                    spellInfo->CanPierceImmuneAura(immuneSpellInfo))
                     continue;
 
-                SpellInfo const* immuneSpellInfo = sSpellMgr->GetSpellInfo(itr->second);
-
-                if (IgnoresSchoolImmunityFromFriendlyCaster(caster, itr->second, immuneSpellInfo))
-                    continue;
-
-                if (spellInfo->CanPierceImmuneAura(immuneSpellInfo))
-                    continue;
-
-                coveredSchoolMask |= itr->first;
+                coveredSchoolMask |= immunitySchoolMask;
             }
+
             if ((coveredSchoolMask & spellSchoolMask) == spellSchoolMask)
                 return true;
         }
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10086,14 +10086,20 @@ bool Unit::IsImmunedToSpell(SpellInfo const* spellInfo, Unit const* caster, Spel
         if (spellSchoolMask != SPELL_SCHOOL_MASK_NONE)
         {
             SpellImmuneContainer const& schoolList = m_spellImmune[IMMUNITY_SCHOOL];
+            // Accumulate all valid school immunity bits that overlap with the spell's schools.
+            // A multi-school spell (e.g. Frostfire Bolt) may have its schools covered by
+            // separate per-school entries (as produced by creature_immunities), so we must
+            // check the combined mask rather than requiring any single entry to cover all schools.
+            uint32 coveredSchoolMask = 0;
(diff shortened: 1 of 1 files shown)


blinkysc:
Possible regression in the other direction: creatures immune to *both* fire and frost (via `creature_immunities`, stored as separate per-school entries) would no longer be immune to Frostfire Bolt, since the new check requires a single entry to cover all schools.

These creatures are immune to both fire and frost and would now take Frostfire Bolt damage:

| Entry | Name |
|---|---|
| 7734 | Ilifar |
| 7735 | Felcular |
| 10485 | Risen Aberration |
| 12457 | Blackwing Spellbinder |
| 15275 | Emperor Vek'nilash |
| 16491 | Mana Feeder |

Is that correct / intended?

Chinoske:
Is creature_immunities a new table that was added?

heyitsbench:
> Is creature_immunities a new table that was added?

Relatively new, yes.

Chinoske:
Good catch @blinkysc. You're right — the `creature_immunities` table applies school immunities as **one entry per school bit** (see `applyCreatureImmunities` in `Creature.cpp`), so a fire+frost immune creature has two separate entries (key=`FIRE`, key=`FROST`) rather than one combined entry (key=`FIRE|FROST`).

Updated the fix to **accumulate** all valid school immunity bits across all matching entries, then check if the combined mask covers the spell's full school mask:

```cpp
uint32 coveredSchoolMask = 0;
for (auto itr = schoolList.begin(); itr != schoolList.end(); ++itr)
{
    if (itr->second == spellInfo->Id)    continue;
    if (!(itr->first & spellSchoolMask)) continue;  // no overlap — skip
    // ... pierceable checks ...
    coveredSchoolMask |= itr->first;
}
if ((coveredSchoolMask & spellSchoolMask) == spellSchoolMask)
    return true;
```

Result:
- **Frost-only immune** vs Frostfire Bolt: `coveredSchoolMask = FROST`, `FROST ≠ FROST|FIRE` → **not immune** ✓
- **Fire+Frost immune** (separate entries) vs Frostfire Bolt: `coveredSchoolMask = FROST|FIRE` → **immune** ✓
- **Fire-only immune** vs Fireball: `coveredSchoolMask = FIRE == FIRE` → **immune** ✓

Chinoske:
The `ubuntu-24.04-clang-18-nopch-modules` check failure is pre-existing and unrelated to this PR's changes.

The errors originate from third-party modules included in the CI modules build:
- `mod-guildhouse`: `Unknown column 'id1' in 'where clause'`
- `mod-guild-zone-system`: `Unknown column 'id1' in 'where clause'`
- `mod-zone-difficulty`: column count mismatch

These modules have SQL files that still reference the old `id1` column, which was renamed to `id` as part of the `creature_multispawn` schema migration. This PR's code compiled successfully and all other checks passed.

Chinoske:
@coderabbitai full review

Chinoske:
Refactored the accumulation loop to use structured bindings and match the style of the equivalent immunity-accumulation logic already in this file (`IsImmunedToDamage`, `IsImmunedToAuraPeriodicTick`) — no behavior change, same filters/order/result. This also resolves the line-length nit above (moot either way since `.coderabbit.yml`/CLAUDE.md now allow 120 columns, but tightened it regardless).

Also looked into whether a genuinely different fix (not just restyled) made sense here — it doesn't. The "accumulate matching entries, then require full coverage of the requested mask" approach is already the established pattern this codebase uses for the same class of problem (see the two functions above), so this PR's fix now just applies that same pattern rather than reinventing it locally.

One further cleanup is possible but out of scope for this fix: extracting a single shared private helper (e.g. `AccumulateSchoolImmunityMask(...)`) that all three functions call, to eliminate the duplication entirely. Happy to open that as a separate follow-up cleanup PR if maintainers want it — didn't want to widen this bug-fix PR's diff.

Chinoske:
Live-tested on a local build with the refactored version applied. Confirmed with real spell casts:

| Test | NPC | Spell | Result |
|---|---|---|---|
| Frost-only immune vs pure Frost | Crazed Water Spirit (16570) | Frost Bolt | Immune ✓ |
| Frost-only immune vs dual-school | Crazed Water Spirit (16570) | Frostfire Bolt | Damage dealt (fire component) ✓ |
| Fire-only immune vs pure Fire | Fire Elemental (575) | Fireball | Immune ✓ |
| Fire-only immune vs dual-school | Fire Elemental (575) | Frostfire Bolt | Damage dealt (frost component) ✓ |
| Fire+Frost immune (5 separate school entries via `creature_immunities`) vs Fire/Frost/Frostfire | Emperor Vek'nilash (15275) | Fire Bolt, Frost Bolt, Frostfire Bolt | Immune to all three ✓ |

The last case is the exact regression @blinkysc flagged — a creature immune to multiple schools via separate per-school entries (confirmed in `Creature::LoadTemplateImmunities`, which splits a combined `creature_immunities.SchoolMask` into one `m_spellImmune[IMMUNITY_SCHOOL]` entry per set bit) stays fully immune via accumulated coverage, while a single-school-immune target now correctly takes partial damage from a dual-school spell instead of being fully immune.

PR #9725, opened 2021-12-15
fix(Core/Spells): Implemented spell proc phases.

UltraNix:
Fixes #9634

## Changes Proposed:
-  Set proper proc phase to some spells (ported from TrinityCore)

## Issues Addressed:

- Closes #9634
- Closes https://github.com/chromiecraft/chromiecraft/issues/2537

## Tests Performed:

- Not tested

## How to Test the Changes:

.go c 48335
.aura 44545
.cast 116

For Druid's Nature's Grace:
Pickup Nature's Grace: talent
Set spell critical chance to 100
Cast Wrath at some target - Nature's Grace should proc off on spell cast.
Cast another Wrath immediately - Nature's Grace should affect that cast and be consumed before the 1st Wrath hits the target.

For Shaman's Elemental Mastery:
Pickup Elemental Mastery talent
Cast Elemental Mastery 
Cast Lightning Bolt at some target - you should get spell haste buff on cast.
Cast another Lightning Bolt - spell haste buff should affect that cast  before the 1st Lightning Bolt hits the target.

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #26210, opened 2026-06-15
fix(Mage): Fingers of Frost charges consumed on spell hit instead of cast

### LINKED ISSUE #25117: (Mage]  Fingers of Frost interactions
### Current Behaviour

Currently Fingers of Frost is consumed on spell cast. Meaning when it procs

1st charge for Frostbolt
2nd charge for Frostbolt

https://github.com/user-attachments/assets/a6d74889-379e-4eda-8661-e65bf585b841

### Expected Behaviour

1st charge for Frostbolt
2nd charge for Frostbolt + Deep Freeze combo

### Source

Sources are hard to dig up since youtube removed filtering via date... but;

https://www.wowhead.com/wotlk/spell=44545/fingers-of-frost#comments:id=436859

Comment chain going to 3.3.2 patch mentioning this interaction

Get a frost mage with points in FoF and Deep Freeze

Play around until FoF procs

Try do FB, FB+Deep Freeze combo

### Extra Notes

Note that this is contrary to the spell tooltip on fingers of frost so may have been a blizzard bug. Seems that in OG wrath charges were used up on spell hit, instead of spell cast, or something which caused a slight delay in the FoF procs being used

### AC rev. hash/commit

c80a0f1fad62 

### Operating system

windows

### Custom changes or Modules

_No response_

Chinoske:
https://github.com/user-attachments/assets/21a16ecc-c09e-4cea-a069-ea7020d42a0c

## Current behavior

**Fingers of Frost** (spell `74396`) charges are consumed at **spell cast time** (`PROC_SPELL_PHASE_CAST`), which prevents the classic Frost Mage double-FoF combo. When the player casts Frostbolt #1, one charge is consumed immediately — by the time Frostbolt #2 or Deep Freeze is cast, no charges remain.

Fixes #25117

## Expected behavior

Per retail WotLK behavior, FoF charges should be consumed at **spell hit time** (`PROC_SPELL_PHASE_HIT`). Because Frostbolt has travel time, both charges remain active during the cast → cast → hit → hit sequence, allowing the intended combo:

1. FoF procs → 2 charges
2. Cast Frostbolt #1 → charge **not yet consumed** (projectile in flight)
3. Cast Frostbolt #2 (or Deep Freeze) → charge **not yet consumed**
4. Frostbolt #1 hits → 1st charge consumed
5. Frostbolt #2 / Deep Freeze hits → 2nd charge consumed

> Retail source: [WoWHead comment #436859 on spell 44545](https://www.wowhead.com/wotlk/spell=44545)

## Changes introduced

**`src/server/scripts/Spells/spell_mage.cpp`** — `spell_mage_fingers_of_frost::PrepareProc`

```cpp
// Before
bool isCastPhase = (eventInfo.GetSpellPhaseMask() & PROC_SPELL_PHASE_CAST) != 0;
// ...
else if (!isCastPhase)
    prevent = true;

// After
bool isHitPhase = (eventInfo.GetSpellPhaseMask() & PROC_SPELL_PHASE_HIT) != 0;
// ...
else if (!isHitPhase)
    prevent = true;
```

Two-line change. The rest of the logic (allow triggered spells, block channeled spells) is unchanged.

## How to test

1. Frost Mage with Fingers of Frost talented
2. Wait for a FoF proc (2 charges on buff bar)
3. Immediately cast Frostbolt → then cast a second Frostbolt or Deep Freeze before the first lands
4. **Before fix**: second spell gets no FoF benefit (charge already gone)
5. **After fix**: both spells benefit from FoF and each consumes one charge on hit

## Extra notes

- No SQL changes required
- Channeled spells (e.g. Blizzard) correctly continue to not consume FoF charges
- Triggered spells continue to be allowed as before
- This was previously fixed in PR #9725 and regressed with the QAston spell overhaul

### DIFF
--- a/data/sql/updates/db_world/2026_06_11_02.sql
+++ b/data/sql/updates/db_world/2026_06_11_02.sql
@@ -1,3 +1,4 @@
+-- DB update 2026_06_11_01 -> 2026_06_11_02
 --
 DELETE FROM `spell_cone` WHERE `Id` > 0;
 INSERT INTO `spell_cone` (`ID`, `ConeDegrees`) VALUES


bonfax4:
This is incorrect. It was on cast. What allowed the ghost charge was something else. Prior to qaston it was still on cast, but the ghost charge worked. 

https://github.com/chromiecraft/chromiecraft/issues/3123

https://github.com/chromiecraft/chromiecraft/issues/2537

Chinoske:
Thank you for the clarification. After reviewing the Patch 3.3 note ("*Fingers of Frost: This talent now triggers immediately on casting a spell rather than being delayed until the spell strikes the target*") and the ChromieCraft issues you linked, I agree this PR is incorrect — the origin/master code already has the right behavior (`PROC_SPELL_PHASE_CAST`), and changing it to `PROC_SPELL_PHASE_HIT` would actually revert to the pre-3.3 behavior.

I would like to close this PR. However, since the ghost charge still does not work in AzerothCore — which was the original motivation behind issue #25117 — would you be able to point toward what the correct fix should be? Based on the ChromieCraft discussion it sounds like there was a separate mechanism that allowed the ghost charge to work on cast-based consumption. Was it related to the spell queue, or something specific in how the charges were checked/removed?

bonfax4:
I believe it was mentioned on the issue that the fix they did was not possible with the new proc system. All i can say is there was a delay from cast to buff removal in retail that ppl attributed to spell batching, so spell queue allowed 2 spells to benefit from the small delay. 

One way I can see this working is adding a 3rd charge that lasts around 0-400ms maybe to emulate batching ?

PR #26038, opened 2026-05-31
fix(Ulduar): Archivum RP missing after Iron Council defeat

Chinoske:
## Changes Proposed

Fixes #25464 — After defeating the Iron Council in Ulduar, Brann Bronzebeard (entry 33235) remains invisible and the Archivum RP sequence never triggers. Brann and Prospector Doren stay invisible, making quests 13604–13607 (and Algalon access) impossible to complete.

`SetBossState(BOSS_ASSEMBLY, DONE)` had no `case` in the instance script switch, so the Archivum RP relied entirely on a SmartAI OOC event (`event_type=1`, `event_flags=513`, NOT_REPEATABLE) that called `SetData(1,1)` on Brann. That event fires at most once per creature lifetime. On server restart with a saved DONE state, Brann respawned and the OOC event could never re-fire.

**Fix:**
- Register Brann (entry 33235) in the Ulduar instance script
- Add `case BOSS_ASSEMBLY` in `SetBossState()`: when state reaches DONE, directly call `brann->AI()->SetData(1, 1)` to trigger the Archivum RP
- Add `case NPC_BRANN_ARCHIVUM` in `OnCreatureCreate()`: if instance is already saved as DONE (server restart scenario), call `SetData(1, 1)` immediately on spawn — same pattern used for NPC_KOLOGARN
- SQL: Remove `SMART_EVENT_FLAG_NOT_REPEATABLE` (0x001) from Brann's SAI OOC self-signal event (513 → 512)

This PR proposes changes to:
- [x] Scripts (`instance_ulduar.cpp`, `ulduar.h`)
- [x] Database (`pending_db_world` — SmartAI flag update for Brann entry 33235)

**AI-assisted Pull Requests**
- [x] AI tools were used partially in preparing this pull request.

## Issues Addressed

Fixes #25464

## Tests Performed

- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members.
- [x] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes

1. Reset a 10 or 25-man Ulduar instance.
2. Kill the Iron Council — verify Brann Bronzebeard (33235) becomes visible near the Archivum and plays the full dialogue RP.
3. Verify Prospector Doren (33956/33957) becomes visible and offers quest 13607 `The Celestial Planetarium`.
4. Save the instance, restart the server, re-enter — verify Brann and Doren are still visible in their correct positions.
[...]

PR #26212, opened 2026-06-15
fix(DB/Ulduar): Brann Bronzebeard Archivum event condition uses wrong INSTANCE_INFO type

Chinoske:
## Description

Brann Bronzebeard (NPC 33235) fails to appear in the Ulduar Archivum after a server restart even when Assembly of Iron has already been defeated. The SmartAI OOC event that triggers the Archivum data-recovery RP sequence is gated by a `CONDITION_INSTANCE_INFO` row that permanently evaluates to false due to a wrong subtype value, making the condition unable to detect that the encounter is complete.

Fixes #26038.

### Root Cause

In `data/sql/base/db_world/conditions.sql`, the condition row for Brann 33235 (`SourceTypeOrReferenceId=22` / `SourceGroup=13` / `SourceEntry=33235`) had `ConditionValue3=0` (`INSTANCE_INFO_DATA`).

In `ConditionMgr.cpp`, `CONDITION_INSTANCE_INFO` dispatches on `ConditionValue3`:

```
case INSTANCE_INFO_DATA (0):      condMeets = instance->GetData(ConditionValue1)      == ConditionValue2;
case INSTANCE_INFO_BOSS_STATE(2): condMeets = instance->GetBossState(ConditionValue1) == EncounterState(ConditionValue2);
```

Ulduar's `GetData()` override only handles tower events, `DATA_MAGE_BARRIER`, and `DATA_CALL_TRAM`. It has no case for boss indices, so `GetData(BOSS_ASSEMBLY_OF_IRON)` (index 4) falls through and returns 0. The condition `0 == 3 (DONE)` is permanently false. `GetBossState()` is the correct path used by the entire Ulduar instance script for tracking encounter completion.

### Fix

Changed `ConditionValue3` from `0` (`INSTANCE_INFO_DATA`) to `2` (`INSTANCE_INFO_BOSS_STATE`) on the Brann 33235 condition row. This routes the check through `instance->GetBossState(4) == DONE(3)`, which is how every other boss-completion check in the Ulduar instance works and correctly returns `DONE` after a server restart.

### Sources (WotLK 3.3.5a)

- https://www.wowhead.com/wotlk/npc=33235 — Brann Bronzebeard in Ulduar; comments confirm he appears at the Archivum after Assembly of Iron is killed and initiates the data-recovery event
- https://www.wowhead.com/wotlk/npc=32867 — Assembly of Iron (Steelbreaker); the encounter whose completion gates the Archivum RP

### How to test

1. Enter Ulduar (603) on a fresh instance and defeat the Assembly of Iron encounter fully.
2. Log out and back in (or restart the server) while still in the instance.
3. **Before fix:** Brann Bronzebeard does not appear in the Archivum; the data-recovery RP never fires.
4. **After fix:** Brann spawns at the Archivum console and begins the RP sequence as expected.
5. Verify that entering a saved instance where Assembly of Iron is already `DONE` also shows Brann present immediately.

### Extra notes

- No C++ change is required. The core condition evaluation logic is correct; only the DB value selecting which instance accessor to call was wrong.
- The same condition row correctly keeps `ConditionValue1=4` (boss index) and `ConditionValue2=3` (`DONE`); only `ConditionValue3` was wrong.
- A `pending_db_world` update file (`2026_06_15_00_fix_brann_archivum_condition.sql`) is included alongside the `conditions.sql` base file correction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the script condition for Brann Bronzebeard to properly detect Assembly of Iron completion, ensuring the character's interaction triggers correctly after defeating the boss encounter.

### DIFF
--- a/data/sql/updates/pending_db_world/2026_06_15_00_fix_brann_archivum_condition.sql
+++ b/data/sql/updates/pending_db_world/2026_06_15_00_fix_brann_archivum_condition.sql
@@ -1,0 +1,9 @@
+-- DB update 2026_06_11_02 -> 2026_06_15_00
+--
+-- Fix Brann Bronzebeard (33235) SAI condition for Assembly of Iron completion check.
+-- ConditionValue3=0 (INSTANCE_INFO_DATA) calls GetData() which has no handler for
+-- BOSS_ASSEMBLY_OF_IRON (4) in Ulduar, always returning 0. The correct type is
+-- ConditionValue3=2 (INSTANCE_INFO_BOSS_STATE) which calls GetBossState(4) == DONE(3).
+DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId` = 22 AND `SourceGroup` = 13 AND `SourceEntry` = 33235;
+INSERT INTO `conditions` (`SourceTypeOrReferenceId`, `SourceGroup`, `SourceEntry`, `SourceId`, `ElseGroup`, `ConditionType`, `ConditionTarget`, `ConditionValue1`, `ConditionValue2`, `ConditionValue3`, `NegativeCondition`, `ErrorType`, `ErrorTextId`, `ScriptName`, `Comment`) VALUES
+(22, 13, 33235, 0, 0, 13, 1, 4, 3, 2, 0, 0, 0, '', 'Execute SAI only if Assembly of Iron Done');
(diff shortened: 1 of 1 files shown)


PkllonG:
Tested

Nyeriah (approved):

PR #24233, opened 2025-12-28
refactor(Core/Spells): QAston proc system

blinkysc:
## Changes Proposed:

This PR ports the QAston proc system from TrinityCore, which is a major refactor of how spell procs are handled. The key changes include:

**Core Changes:**
- New `spell_proc` table with enhanced filtering: `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`
- Proc phase separation: CAST phase (1) vs HIT phase (2) for proper timing
- **CAST phase HitMask filtering for travel-time spells** - Allows procs like Nature's Grace and Elemental Focus to trigger at cast time using pre-calculated crit results, enabling buffs to affect spells cast during projectile flight
- Moved 4400+ lines of `HandleDummyAuraProc` switch statements to individual spell scripts
- Added 174 new spell proc scripts across all classes

**Benefits:**
- More accurate proc behavior matching retail WotLK
- Easier to maintain - each spell has its own script instead of giant switch statements
- Better proc filtering prevents incorrect triggers
- Unit test framework for proc system validation

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [x] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> While the use of AI tools when preparing pull requests is not prohibited, contributors must clearly disclose when such tools have been used and specify the model involved.
> 
> Contributors are also expected to fully understand the changes they are submitting and must be able to explain and justify those changes when requested by maintainers.

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.

**AI Tools Used:** Claude Code (claude-opus-4-5-20251101) thinking with [azerothMCP](https://github.com/blinkysc/azerothMCP)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9177
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5967
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/9400
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10993
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/11462
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/12115
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/13330
[...]

PR #26216, opened 2026-06-15
fix(Mage): Fingers of Frost ghost charge — prevent self-consumption and add ghost window (#25117)

Chinoske:
## Description

Fingers of Frost ghost charge is non-functional after the QAston proc overhaul (#24233). Two independent bugs prevent it: the proccing spell immediately consumes one of the two fresh charges, and the aura state is removed instantly when the last charge is consumed, leaving no window for a queued instant spell.

### Root Cause

The QAston overhaul replaced the previous two-script system (`spell_mage_fingers_of_frost_proc_aura` + `spell_mage_fingers_of_frost_proc`) with a minimal single-script that delegates charge consumption to `PrepareProcToTrigger`. This introduced two regressions:

**1 — Self-consumption:** The old `spell_mage_fingers_of_frost_proc` script (on spell 44543) blocked the spell that triggered FoF from also consuming a charge from the newly-applied buff (74396). Without this guard, the proccing Frostbolt consumes one of the two charges in the same proc pass, effectively leaving only one usable charge instead of two.

**2 — No ghost charge window:** When the last charge is consumed, `OnRemove` called `RemoveAurasDueToSpell(44544)` immediately. The aura state (44544) was gone before the spell queue could fire a queued instant cast, so no ghost charge was possible. In retail WotLK, a small latency/batching window allowed a queued instant cast to still see the FoF aura state active.

Reference: https://github.com/azerothcore/azerothcore-wotlk/issues/25117

### Fix

**Fix 1 — Anti-self-consumption (`_freshlyApplied` flag):**
`HandleApply` sets `_freshlyApplied = true` when the aura is applied. `PrepareProc` blocks the first non-triggered, non-channeled CAST-phase proc after fresh application and clears the flag. Triggered procs (e.g. Blizzard ticks) also clear the flag so the guard doesn't persist into unrelated casts.

**Fix 2 — Ghost charge window (`SetDuration(400)`):**
Instead of immediately removing aura state 44544 in `OnRemove`, its duration is set to 400ms — matching AC's spell queue window (PR #20797). A queued instant cast (Ice Lance, Deep Freeze, Brain Freeze proc) that fires within that window still sees 44544 active and benefits from Shatter crit.

**Expected combo sequence:**
```
FB1 → procs FoF (2 charges), self-consume blocked
FB2 → consumes charge 1 (1 remaining)
FB3 → consumes charge 2 (0) → 44544.SetDuration(400ms)
IL  → queued, fires ~10ms later → sees 44544 → ghost charge ✓
```

### Sources (WotLK 3.3.5a)

- https://www.wowhead.com/wotlk/spell=44545/fingers-of-frost#comments — comment chain confirming ghost charge behavior in 3.3.5a retail
- https://web.archive.org/web/20101204052143/http://www.worldofwarcraft.com/patchnotes/patch3p3.html — Patch 3.3: *"Fingers of Frost: This talent now triggers immediately on casting a spell rather than being delayed until the spell strikes the target"* (confirms CAST consumption is correct, ghost charge is a separate mechanic)
- https://www.mmo-champion.com/threads/677129-Fingers-of-Frost-Change — *"Ghost charge is all latency. The buff fades on spell cast, not hit."*
- https://wowpedia.fandom.com/wiki/Fingers_of_Frost?oldid=2301949 — documents ghost charge behavior in 3.3.5a

### How to test

1. Frost Mage with 2/2 Fingers of Frost + Frostbolt + Ice Lance + Deep Freeze
2. Spam Frostbolt on a target dummy until FoF procs (2 stacks appear)
3. **Fix 1 check:** Stop casting immediately — buff should show 2 stacks without auto-decrementing to 1
4. **Ghost charge check:** Cast Frostbolt (1 stack consumed) → Frostbolt again (last stack consumed) → immediately cast Ice Lance
5. Ice Lance should receive the Shatter crit bonus even though FoF stacks already hit 0
6. **Regression check:** Cast Blizzard on mobs with FoF — each tick should consume exactly 1 charge normally

cc @bonfax4 — you identified the root cause in https://github.com/azerothcore/azerothcore-wotlk/issues/25117 and https://github.com/chromiecraft/chromiecraft/issues/3123, wanted to get your eyes on this approach.

### DIFF
--- a/src/server/scripts/Spells/spell_mage.cpp
+++ b/src/server/scripts/Spells/spell_mage.cpp
@@ -956,6 +956,11 @@ class spell_mage_fingers_of_frost : public AuraScript
         return ValidateSpellInfo({ SPELL_MAGE_FINGERS_OF_FROST_AURASTATE_AURA });
     }
 
+    void HandleApply(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
+    {
+        _freshlyApplied = true;
+    }
+
     void PrepareProc(ProcEventInfo& eventInfo)
     {
         if (Spell const* spell = eventInfo.GetProcSpell())
@@ -966,11 +971,21 @@ class spell_mage_fingers_of_frost : public AuraScript
             bool prevent = false;
 
             if (isTriggered)
+            {
+                // Triggered procs (e.g. Blizzard ticks) consume charges normally and clear the guard
+                _freshlyApplied = false;
                 prevent = false;
+            }
             else if (isChanneled)
                 prevent = true;
             else if (!isCastPhase)
                 prevent = true;
+            else if (_freshlyApplied)
+            {
+                // The spell that procced FoF into existence must not also consume a charge
+                prevent = true;
+                _freshlyApplied = false;
+            }
 
             if (prevent)
                 PreventDefaultAction();
@@ -979,14 +994,22 @@ class spell_mage_fingers_of_frost : public AuraScript
 
     void OnRemove(AuraEffect const* /*aurEff*/, AuraEffectHandleModes /*mode*/)
     {
-        GetTarget()->RemoveAurasDueToSpell(SPELL_MAGE_FINGERS_OF_FROST_AURASTATE_AURA);
+        // Keep the aura state active briefly so a spell queued before the last charge
+        // was consumed can still benefit (ghost charge window, ~400ms spell queue window)
+        if (Aura* auraState = GetTarget()->GetAura(SPELL_MAGE_FINGERS_OF_FROST_AURASTATE_AURA))
+            auraState->SetDuration(400);
+        else
+            GetTarget()->RemoveAurasDueToSpell(SPELL_MAGE_FINGERS_OF_FROST_AURASTATE_AURA);
     }
 
     void Register() override
     {
+        AfterEffectApply += AuraEffectApplyFn(spell_mage_fingers_of_frost::HandleApply, EFFECT_0, SPELL_AURA_DUMMY, AURA_EFFECT_HANDLE_REAL);
         DoPrepareProc += AuraProcFn(spell_mage_fingers_of_frost::PrepareProc);
         AfterEffectRemove += AuraEffectRemoveFn(spell_mage_fingers_of_frost::OnRemove, EFFECT_0, SPELL_AURA_DUMMY, AURA_EFFECT_HANDLE_REAL);
     }
+
+    bool _freshlyApplied = false;
 };
 
 // -31571 - Arcane Potency


Chinoske:
### Test video — Fingers of Frost ghost charge

https://github.com/user-attachments/assets/5747af40-8d0c-48a7-b453-eaf4d0cea5ce

**Sequence shown:**
1. Frostbolt procs FoF (2 stacks appear, no self-consume)
2. Frostbolt consumes stack 1 → 1 remaining
3. Frostbolt consumes stack 2 → 0 stacks
4. Ice Lance queued fires within the 400ms window → benefits from ghost charge ✓

bonfax4:
> ### Test video — Fingers of Frost ghost charge
>  Grabacion.de.pantalla.2026-06-15.193611.mp4 
> **Sequence shown:**
> 
> 1. Frostbolt procs FoF (2 stacks appear, no self-consume)
> 2. Frostbolt consumes stack 1 → 1 remaining
> 3. Frostbolt consumes stack 2 → 0 stacks
> 4. Ice Lance queued fires within the 400ms window → benefits from ghost charge ✓

Doesnt seem to be working correctly in your test, the stack didn't go from 2 to 1 on the cast after the proc ? 

Best way to test is actually frostbolt(proc) > frost bolt (2>1) > frostbolt (1>0) + deep freeze. Deep freeze should use the ghost charge

Chinoske:
https://github.com/user-attachments/assets/220a1c98-74ef-4771-9bf8-75bdb90cbc55



Thanks for the review! You're right that the video wasn't the clearest test — here's what actually happened:

In my recording, **Frostbolt 2 also proc'd FoF** (which is normal — it has a percentage chance every cast). Because of the `_freshlyApplied` guard in the script, any spell that triggers a new FoF proc is also blocked from consuming an existing charge at the same time. So the stack stayed at 2/2 instead of going 2→1, which made the video hard to read.

The fix itself is correct — the `_freshlyApplied` flag is cleared as soon as a cast that **doesn't** proc FoF hits, and that cast consumes a charge normally.

Chinoske:
Here's the exact sequence visible in the video, with the fix applied:

| Cast | Stacks | What happened |
|------|--------|---------------|
| Frostbolt 1 | 2/2 | FoF procs — **no self-consume** (the spell that creates the charge can't also burn it) |
| Frostbolt 2 | 2/2 | FoF procs again — **no consume on refresh** (same guard applies) |
| Frostbolt 3 | 1/1 | FoF procs — **1 charge consumed** (2→1), proc refreshes remaining stack |
| Frostbolt 4 (queued) | 0 | Fires inside the ghost window — **last charge consumed**, benefits from FoF ✓ |

The 2/2→2/2 transition on FB2 looks like nothing happened, but it's correct: both FB1 and FB2 triggered fresh procs, which intentionally prevents self-consumption. Charges only drain once a cast doesn't create a new proc.

I can re-record following your Deep Freeze suggestion for a cleaner visual if needed.

bonfax4:
~~4th shouldn't be able to benefit from it since the cast is > 400 ms, if it does that should not happen. Spells snapshot stuff when cast ends~~

Edit: Actually in your example a 5th cast would be needed to check for the ghost charge

Chinoske:
Hey @bonfax4, could you be a bit more specific about what you'd like to see?

A few things that would help clarify:

1. When you say "4th shouldn't be able to benefit from it" — in the sequence I described, **Frostbolt 4 consumed the last charge** (1→0) and triggered the ghost window. Are you saying FB4 itself critted unexpectedly, or that something after FB4 critted when it shouldn't have?

2. What exact spell do you want to see use the ghost window in the test? (Ice Lance, Holy Shock, Deep Freeze?)

3. What's the exact sequence you'd consider a clean proof? Something like:
   - Frostbolt (proc 2/2)
   - Frostbolt (2→1)
   - Frostbolt (1→0) + immediately queue [spell X]
   - [spell X] crits inside the ghost window ✓

Just want to make sure I'm recording the right thing so we don't go back and forth on the video.

bonfax4:
Yeah just record a deep freeze to make sure. I edited the comment,I was confused because you said the 4th consumed the ghost charge

Chinoske:
To clarify the sequence — **Frostbolt 4 is not benefiting from the ghost window, it's creating it**:

- **FB1** → FoF procs → 2/2 active (no self-consume)
- **FB2** → FoF procs again → 2/2 **refreshed** (timer resets, still no self-consume)
- **FB3** → consumes 1 charge → **1/1** remaining
- **FB4** → consumes the last charge → **1→0**, ghost window opens (400 ms)
- **Next queued spell** (Ice Lance / instant) → fires inside the 400 ms window → crits ✓

So when I wrote "Frostbolt 4 ghost charge" I meant FB4 consumed the final real charge and *triggered* the ghost window — not that FB4 itself critted from a ghost. FB4 used a legitimate charge. A cast-time spell queued after FB4 would not benefit from the ghost window since it takes >400 ms to land.

Chinoske:
Correction to my previous comment — there was no extra spell in the test. The full sequence is:

- **FB1** → FoF procs → 2/2 active (no self-consume)
- **FB2** → FoF procs again → 2/2 refreshed (no self-consume)
- **FB3** → consumes 1 charge → 1/1 remaining
- **FB4** → consumes the last charge → **1→0, ghost window opens** ✓

FB4 used a real charge, not a ghost. "Ghost charge" here means FB4 is the cast that drove charges to zero and triggered the 400 ms aurastate window — any instant spell already in the queue at that moment (e.g. Ice Lance) would benefit, but the test itself only showed the mechanism activating.

Chinoske:
To summarize the full picture for clarity:

The sequence in the test is:

- **FB1** → FoF procs → **2/2** (no self-consume —  guard blocks it)
- **FB2** → FoF procs again → **2/2 refreshed** (same guard, still no self-consume)
- **FB3** → no proc → consumes 1 charge → **1/1 remaining**
- **FB4** → no proc → consumes last charge → **1→0**, ghost window opens (400 ms)
- **Ice Lance** queued just before FB4 lands → fires instantly after → still sees FoF active → **crits ✓**

The reason it takes 4 Frostbolts instead of 3 is the fix itself: FB1 and FB2 both proc FoF, and the  guard correctly prevents the spell that *created* a charge from also *consuming* one in the same instant. Without this fix, FB1 would proc and consume simultaneously — the proc would appear to do nothing.

The ghost window (400 ms on the internal aurastate after the last charge drops to 0) exists for instant spells like Ice Lance that are already queued at the moment the final charge is consumed.

blinkysc:
I'm wondering if this is the best way; iterate through possible procs that have "batching feature" and implement in spell script. I'm wondering if this should be generalized in some way and then that prescribed through the db spell_proc?

The best way is to really emulate the batching... But this is difficult...

Chinoske:
The batching approach would be the theoretically correct solution, but it's a significant core rework that goes well beyond the scope of a spell-specific fix. Spell batching requires changes to how the entire spell processing pipeline handles concurrent events within the same server tick, which affects a huge surface area of the codebase.

This script targets the specific observable bug: a spell queued within the spell queue window (~400 ms) loses its FoF benefit because the aurastate is removed before the queued spell resolves. The ghost window reproduces that behavior in a contained, spell-specific way without touching core systems.

If a generalized batching system is ever implemented, this script can be removed. Until then, the individual script approach is the practical path to correct gameplay behavior.

PR #26262, opened 2026-06-18
fix(Scripts/Ulduar): Assembly of Iron no longer revives dead council members on a partial evade

### LINKED ISSUE #25559: Regression: Assembly of Iron revives dead council members when one survivor evades
### Current Behaviour

On upstream 72591ee8415ad93aa47fddb0dc7bda57ffe9e06b, a dead Steel Council boss can respawn during the same attempt if another surviving council member enters evade/reset before the encounter is actually over.

The problematic path is:
1. A surviving boss calls UpdateVictim().
2. Creature::SelectVictim() gets no acceptable current victim and calls AI()->EnterEvadeMode().
3. In boss_assembly_of_iron.cpp, each council member Reset() unconditionally calls RespawnAssemblyOfIron() and sets BOSS_ASSEMBLY back to NOT_STARTED.
4. RespawnAssemblyOfIron() respawns every dead council member, so a partial reset by one survivor revives bosses that were already killed.

This is not a normal full-wipe reset. It happens when combat desynchronizes and one living council member resets while another living council member is still in combat.

### Expected Behaviour

Already killed council members should stay dead for the current attempt.

Only a full encounter reset, after all surviving council members have disengaged, should respawn the council and set the encounter back to NOT_STARTED.

A partial evade/reset by one surviving council member must not revive already killed bosses.

### Source

Related historical report: #16670

Wrath encounter behavior: the council progresses through Supercharge after each kill; already killed bosses should not revive unless the whole encounter fully resets after a wipe.

1. Start Assembly of Iron on latest upstream.
2. Kill one council member, but leave at least one other council member alive.
3. Before the encounter fully ends, make one surviving council member lose all acceptable targets and enter evade/reset while another surviving council member is still in combat.
4. Observe that the dead council member respawns immediately because the resetting boss runs RespawnAssemblyOfIron() from Reset().

Examples of step 3 include any combat desync where one surviving boss temporarily has no acceptable victim while another survivor is still engaged.

This is why the bug is not guaranteed on every pull, but it is deterministic once a surviving council member hits Reset() before the whole encounter has reset.

### Extra Notes

- The script bug itself is old: boss_assembly_of_iron.cpp has long unconditionally respawned dead council members from each boss Reset().
- The most likely regression trigger is d3f1be6d2223fc1a1d08751316885b609b693a20 (fix(Core/Scripts): Fix boss scripts resetting during death/defeat RP after threat system port), because it changed threat/reset behavior so a surviving council member can reach the no-valid-victim -> EnterEvadeMode() path more easily.
- The later core change b00b424d63b391ffdbd2c07eda3b2dfff844b966 (fix(Core/Unit): Prevent creature evade when on another unit's threat list) only narrows one class of false evades; it does not fix the Assembly of Iron script path.
- A minimal script-side fix is to guard RespawnAssemblyOfIron() and SetBossState(BOSS_ASSEMBLY, NOT_STARTED) in each council member Reset(), and only run them when no other alive council member is still in combat.
- This looks related to #16670, but this report adds the likely regression commit, the exact control flow, and a concrete fix direction.

### AC rev. hash/commit

72591ee8415ad93aa47fddb0dc7bda57ffe9e06b
Likely regression trigger: d3f1be6d2223fc1a1d08751316885b609b693a20
Related partial mitigation: b00b424d63b391ffdbd2c07eda3b2dfff844b966

### Operating system

Windows x64

### Custom changes or Modules

Local workspace contains custom modules, but the affected path is in upstream core/script files only:
- src/server/game/Entities/Unit/Unit.cpp
- src/server/game/AI/CreatureAI.cpp
- src/server/scripts/Northrend/Ulduar/Ulduar/boss_assembly_of_iron.cpp

The regression analysis above is based on upstream source state 72591ee8415ad93aa47fddb0dc7bda57ffe9e06b.

Nyeriah:
## Changes Proposed:
This PR proposes changes to:

Fixes a regression in the Ulduar **Assembly of Iron** (Steel Council) encounter where an already-killed council member could respawn mid-attempt.

Each council member's `Reset()` unconditionally called `RespawnAssemblyOfIron()` (which revives every dead member) and reset `BOSS_ASSEMBLY` to `NOT_STARTED`. Since `Reset()` runs on every per-creature `EnterEvadeMode()`, a single *surviving* boss losing all valid targets (a combat desync) revived bosses the raid had already killed.

The fix guards `RespawnAssemblyOfIron()` + `SetBossState(NOT_STARTED)` behind a new helper so they run only when no *other* council member is still alive and engaged. On a real wipe, members disengage one by one and only the last one performs the full reset; a partial/false evade by one survivor no longer revives the dead.

The guard keys on `Creature::IsEngaged()` (the AI engagement flag) rather than `IsInCombat()`, because Brundir intentionally calls `CombatStop()` for ~16s during Lightning Tendrils while the fight is still active — `IsInCombat()` would be `false` in that window and could let the revive slip through, whereas the engagement flag survives the brief combat gap (`ScriptedAI::JustExitedCombat` is a no-op).

### AI-assisted Pull Requests

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. **Tool used: Claude Code (Claude Opus 4.8).** The author reviewed and understands the change.

## Issues Addressed:
- Closes #25559

## SOURCE:

This is a logic/regression fix. The intended behavior — that already-killed Steel Council members stay dead and only revive on a full encounter reset after a wipe — is standard WotLK encounter behavior (the council progresses through Supercharge after each kill). See the linked issue for the control-flow analysis and likely regression commit (`d3f1be6d2`).

## Tests Performed:
This PR has been:

Builds cleanly (script-only change, no core/SQL edits) and passes `apps/codestyle/codestyle-cpp.py`.

## How to Test the Changes:

1. Start Assembly of Iron. Kill one council member, leaving at least one other alive.
2. Before the encounter ends, make one surviving member lose all acceptable targets and evade while another surviving member is still engaged.
3. Verify the already-dead member does **not** respawn, and the encounter continues.
4. Wipe the raid fully and confirm the council still respawns and the encounter resets to `NOT_STARTED` (doors reopen) as before.

- [ ] None.

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_assembly_of_iron.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_assembly_of_iron.cpp
@@ -135,6 +135,22 @@ static uint8 CountAliveBosses(InstanceScript* instance)
     return count;
 }
 
+// True while any council member other than `me` is still alive and engaged.
+// IsEngaged() (not IsInCombat()) survives Brundir's ~16s CombatStop during
+// Lightning Tendrils, so a lone survivor's reset cannot revive dead members.
+static bool IsAnyAssemblyMemberEngaged(InstanceScript* pInstance, Creature* me)
+{
+    if (!pInstance || !me)
+        return false;
+
+    for (uint8 i = 0; i < 3; ++i)
+        if (Creature* boss = pInstance->GetCreature(DATA_STEELBREAKER + i))
+            if (boss != me && boss->IsAlive() && boss->IsEngaged())
+                return true;
+
+    return false;
+}
+
 bool IsEncounterComplete(InstanceScript* pInstance, Creature* me)
 {
     if (!pInstance || !me)
@@ -186,12 +202,16 @@ struct boss_steelbreaker : public ScriptedAI
     void Reset() override
     {
         me->SetLootMode(0);
-        RespawnAssemblyOfIron(pInstance, me);
 
         _phase = 0;
         events.Reset();
-        if (pInstance)
-            pInstance->SetBossState(BOSS_ASSEMBLY, NOT_STARTED);
+
+        if (!IsAnyAssemblyMemberEngaged(pInstance, me))
+        {
+            RespawnAssemblyOfIron(pInstance, me);
+            if (pInstance)
+                pInstance->SetBossState(BOSS_ASSEMBLY, NOT_STARTED);
+        }
     }
 
     void JustReachedHome() override
@@ -376,14 +396,17 @@ struct boss_runemaster_molgeim : public ScriptedAI
     void Reset() override
     {
         me->SetLootMode(0);
-        RespawnAssemblyOfIron(pInstance, me);
 
         _phase = 0;
         events.Reset();
         summons.DespawnAll();
 
-        if (pInstance)
-            pInstance->SetBossState(BOSS_ASSEMBLY, NOT_STARTED);
+        if (!IsAnyAssemblyMemberEngaged(pInstance, me))
+        {
+            RespawnAssemblyOfIron(pInstance, me);
+            if (pInstance)
+                pInstance->SetBossState(BOSS_ASSEMBLY, NOT_STARTED);
+        }
 
         me->m_Events.AddEventAtOffset(new CastRunesEvent(*me), 8s);
     }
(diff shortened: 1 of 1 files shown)


PkllonG:
Tested

PR #26295, opened 2026-06-21
fix(Scripts/BlackTemple): Shade of Akama, Illidan encounter, and passage doors

wukodlak21:
## Summary

Three separate fixes for Black Temple:

### 1. Shade of Akama — `SetWalk(true)` causes movement failure

`boss_shade_of_akama.cpp` called `SetWalk(true)` in `Reset()`. The channeler auras applied to the Shade reduce its walk speed to 0, which causes a MoveSpline velocity failure. The Shade stops moving entirely during the encounter.

**Fix:** Remove `SetWalk(true)` from `Reset()`.

### 2. Akama/Illidan encounter — `MovePoint` not triggering `MovementInform`

In `npc_akama_illidan`, the gossip select handler called `MovePoint(POINT_FACE_ILLIDAN, FaceIllidan)`. When Akama is summoned at a position very close to the target (within the same tick), `MovePoint` completes instantly and `MovementInform` is never called. As a result, `ACTION_START_EVENT` never reaches Illidan and the encounter does not start.

**Fix:** Replace `MovePoint` with `NearTeleportTo(FaceIllidan)`, which guarantees positioning. Since `NearTeleportTo` also does not trigger `MovementInform`, the Illidan encounter action and all Akama dialogue/emote events are now scheduled inline immediately after the teleport.

### 3. Passage doors — boss order not enforced in Black Temple (blizzlike)

Black Temple in original TBC did not require killing all bosses in sequence — players could skip directly to Illidan. The current `DOOR_TYPE_PASSAGE` entries in `doorData` enforced a strict kill order that is not blizzlike.

**Fix:** Remove `DOOR_TYPE_PASSAGE` entries from `doorData`. All passage doors (`GO_NAJENTUS_GATE`, `GO_SUPREMUS_GATE`, `GO_GURTOGG_DOOR`, `GO_TEMPLE_DOOR`, `GO_MOTHER_SHAHRAZ_DOOR`, `GO_ILLIDAN_GATE`) are now force-opened at instance creation via `OnGameObjectCreate`. `DOOR_TYPE_ROOM` entries (lockout zones during fights) are preserved.

## How to test

1. Enter Black Temple — verify all passage doors are open from the start
2. Engage Shade of Akama — verify the Shade moves during the encounter
3. Complete up to Illidan — verify Akama's gossip triggers the cinematic and Illidan becomes attackable

### DIFF
--- a/src/server/scripts/Outland/BlackTemple/boss_illidan.cpp
+++ b/src/server/scripts/Outland/BlackTemple/boss_illidan.cpp
@@ -799,7 +799,26 @@ struct npc_akama_illidan : public ScriptedAI
         else
         {
             me->SetSheath(SHEATH_STATE_UNARMED);
-            me->GetMotionMaster()->MovePoint(POINT_FACE_ILLIDAN, FaceIllidan);
+            me->NearTeleportTo(FaceIllidan);
+            // NearTeleportTo does not trigger MovementInform, so we invoke the Illidan encounter directly
+            if (Creature* illidan = instance->GetCreature(DATA_ILLIDAN_STORMRAGE))
+            {
+                me->SetFacingToObject(illidan);
+                illidan->AI()->DoAction(ACTION_START_EVENT);
+                me->SetHomePosition(me->GetPosition());
+            }
+            me->m_Events.AddEventAtOffset([&] { Talk(SAY_AKAMA_FREE); }, 15400ms);
+            me->m_Events.AddEventAtOffset([&] { me->HandleEmoteCommand(EMOTE_ONESHOT_TALK); }, 19440ms);
+            me->m_Events.AddEventAtOffset([&] { me->HandleEmoteCommand(EMOTE_ONESHOT_SALUTE); }, 23080ms);
+            me->m_Events.AddEventAtOffset([&] { Talk(SAY_AKAMA_TIME_HAS_COME); }, 33840ms);
+            me->m_Events.AddEventAtOffset([&] {
+                me->HandleEmoteCommand(EMOTE_ONESHOT_ROAR);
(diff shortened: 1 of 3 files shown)


Nyeriah:
We encourage the responsible use of AI tools. Please use them to assist your work, but ensure that any factual, technical, or consequential information is independently reviewed and validated before being relied upon or shared.

I'm afraid this doesn't appear to pass muster. The description shows the AI has made a number of assumptions that don't hold up, leading it to tackle the wrong issue entirely.

brighton-chi:
I'm confused. All bosses in BT are required. That was in the case in TBC and is still the case in retail.

Nyeriah:
Just AI-generated nonsense, I'm afraid. I'll be closing this, as it is quite clear the output hasn't undergone even minimal review.

heyitsbench:
> That was in the case in TBC and is still the case in retail.

It was not the case during the TBC Classic run, as (iirc) doing Shade of Akama, Teron Gorefiend, Gurtogg Bloodboil, *or* Reliquary of Souls would open the doors to the upper bosses, rather than needing all four to be defeated. Being able to skip directly to Illidan was not possible however.

brighton-chi:
> > That was in the case in TBC and is still the case in retail.
> 
> 
> 
> It was not the case during the TBC Classic run, as (iirc) doing Shade of Akama, Teron Gorefiend, Gurtogg Bloodboil, *or* Reliquary of Souls would open the doors to the upper bosses, rather than needing all four to be defeated. Being able to skip directly to Illidan was not possible however.

From what I can tell, that was a TBC Classic special that has not been implemented in retail to this day, which is kind of interesting. I wonder if there are other instances of this sort of thing in the Classic versions.

PR #26296, opened 2026-06-21
fix(Scripts/Karazhan): Attumen the Huntsman (Mounted) spawn corrections

wukodlak21:
## Summary

Two fixes for the Attumen the Huntsman encounter in Karazhan:

### 1. Mounted Attumen spawns with copied HP instead of 100%

When Attumen mounts Midnight, the mounted form was spawning with `max(Midnight HP, Attumen HP)` instead of full health. Blizzlike behaviour (confirmed via live sniff data) is that the mounted Attumen always spawns at 100% HP.

**Fix:** Replace the conditional HP assignment with `summon->SetHealth(summon->GetMaxHealth())`.

### 2. `IsSummonedBy` phase assignment logic

The original code used two separate `if` blocks to assign `_phase`:
```cpp
if (summoner->GetEntry() == NPC_MIDNIGHT)
    _phase = PHASE_ATTUMEN_ENGAGES;
if (summoner->GetEntry() == NPC_ATTUMEN_THE_HUNTSMAN)
{
    _phase = PHASE_MOUNTED;
    DoCastSelf(SPELL_SPAWN_SMOKE);
}
```
The second block had no `else`, so if both conditions could ever be true `PHASE_ATTUMEN_ENGAGES` would immediately be overwritten. Additionally, `summoner` could theoretically be null.

**Fix:** Rewrite as `if/else` checking `me->GetEntry()` for the mounted NPC entry (which is unambiguous) with an explicit `else` for `PHASE_ATTUMEN_ENGAGES`. Added null check for `summoner`.

## How to test

1. Enter Karazhan and engage Midnight/Attumen
2. Bring Midnight to ~25% HP — Attumen should mount Midnight
3. The mounted Attumen should spawn with 100% HP (not a copy of either's current HP)
4. Complete the encounter normally

### DIFF
--- a/src/server/scripts/EasternKingdoms/Karazhan/boss_midnight.cpp
+++ b/src/server/scripts/EasternKingdoms/Karazhan/boss_midnight.cpp
@@ -146,14 +146,8 @@ struct boss_attumen : public BossAI
         {
             if (Creature* midnight = instance->GetCreature(DATA_MIDNIGHT))
             {
-                if (midnight->GetHealth() > me->GetHealth())
-                {
-                    summon->SetHealth(midnight->GetHealth());
-                }
-                else
-                {
-                    summon->SetHealth(me->GetHealth());
-                }
+                // Mounted Attumen spawns at 100% HP (blizzlike - confirmed via live sniffs)
+                summon->SetHealth(summon->GetMaxHealth());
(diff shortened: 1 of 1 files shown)


Nyeriah:
Well this is just wrong

wukodlak21:

Issue #16256, opened 2023-05-14
(Events): Zombie Infestation unimplemented.


heyitsbench:
### Current Behaviour

Not yet implemented.

### Expected Blizzlike Behaviour

An event during the pre-patch leading into WotLK. Conspicuous crates spawn in various cities that, when interacted with, give you a debuff that eventually turns you into a zombie. The time on the debuff depends on the phase of the event. There are various Argent Dawn NPCs spawned as well that can aid the player to either kill zombies or cure the living of the plague. Over time the amount of healers are reduced until they are removed entirely. At a certain point the [Scourge Invasion](https://wowpedia.fandom.com/wiki/Scourge_Invasion) event also kicked off as a part of the infestation, as well as a couple of quests to cure the plague once and for all.

### Source

[Wowpedia](https://wowpedia.fandom.com/wiki/Zombie_Infestation), [Wowhead](https://www.wowhead.com/wotlk/news/wrath-classic-surviving-the-zombie-infestation-and-scourge-invasion-328412)

### Steps to reproduce the problem

1. Exist.
2. Observe.

### AC rev. hash/commit

89acad2145cb555148f01101001a9460b4066bef

### Operating system

Windows 10 x64

### Custom changes or Modules

[mod-ptr-template](https://github.com/heyitsbench/mod-ptr-template)

heyitsbench:
Using [`.aura 43869`](https://wowgaming.altervista.org/aowow/?spell=43869) (I believe the actual zombie status aura) does not seem to change your faction, something it should do. Based on an added [Wrath Classic spell](https://www.wowhead.com/wotlk/spell=394653), it seems the faction it's supposed to apply is 2075.

PR #26333, opened 2026-06-24
feat(WorldEvent): implement base Zombie Infestation event

vibecoder99-cmd:
## Changes

Implements a base version of the missing Wrath pre-patch Zombie Infestation world event.

This addresses #16256.

Implemented:
- `48953 Infect!` script effect applies infection to living players
- Timed `43958 You're Infected!` infection countdown (60 seconds)
- Natural infection expiry transforms player into `43869 You're a Zombie!`
- Zombie faction hostility using faction 1945 and `UNIT_FLAG2_IGNORE_REPUTATION`
- Safe cleanup on aura removal, death, logout, and login
- Conspicuous Crate (186799) infection source
- Plagued Roach (27845) / Plagued Vermin (27855) infection on death via SmartAI
- Argent Healer (27305) cures infected and zombie players within 6 yards
- Minimal Booty Bay `game_event` staging (event 200) with clean start/stop

## Scope

This PR implements the base event loop only.

Follow-up work may include:
- Broader historical city spawn coverage
- Phase-dependent infection durations
- Staged phase progression
- NPC zombification
- Scourge Invasion integration

## Testing

Build:
- [x] worldserver builds
- [x] worldserver starts
- [x] C++ codestyle passes
- [x] SQL codestyle passes

Manual:
- [ ] `.event start 200` spawns crates, roaches/vermin, and Argent Healer in Booty Bay
- [ ] Conspicuous Crate interaction applies infection
- [ ] Plagued Roach / Vermin death applies infection to killer
- [ ] Infection countdown (60s) naturally transforms player into zombie
- [ ] Manual infection removal (`.unaura 43958`) does not transform player
- [ ] Zombie player remains player-controlled
- [ ] Guards attack zombie player on sight
- [ ] Argent Healer cures infected player (Cure Disease)
- [ ] Argent Healer cures zombie player (removes aura)
- [ ] `.unaura 43869` restores normal state
- [ ] Death while zombie restores normal state
- [ ] Logout/relog while infected or zombie restores normal state
- [ ] `.event stop 200` despawns all event spawns
- [ ] Server restart with event off produces no event spawns
- [ ] Re-starting event produces no duplicate spawns

## Notes

- `UNIT_FLAG_PLAYER_CONTROLLED` is never removed from zombie players.
- `UNIT_FLAG2_IGNORE_REPUTATION` bypasses the reputation lookup in `GetFactionReactionTo` so NPCs use faction template hostility checks instead of the player's standing. This flag is set on zombie aura apply and cleared on all removal paths.
- Infection spell 43958 has DurationIndex 0 in the client DBC. The 60-second countdown is set via `SetMaxDuration`/`SetDuration` in the AuraScript `AfterEffectApply` handler.
- The Conspicuous Crate (186799) already has `data10 = 48953` in its gameobject template. No template modification is needed for crate infection behavior.
- Linked trap gameobject 193194 is auto-spawned by the core via `linkedTrapId`.


sudlud:
Thanks for taking a swing at this, it has been open since 2023 and nobody had touched it.

I am closing this, and the reason is the approach rather than any individual bug, so let me be direct about it. For a world event like this we only want a real implementation, built on the actual event data. A smaller first slice is completely fine, phases and full city coverage can come later, but the slice itself has to be the real mechanic rather than an approximation of it. This is an approximation, and working through review comments would not change that.

Concretely, the client data already contains most of the event, and this goes around it instead of through it.

`43869 You're a Zombie!` applies shapeshift form 21, which is `FORM_ZOMBIE` and already handled in `SpellAuraEffects.cpp`. That form has a full ability set in Spell.dbc gated on its stance mask:

- 43949 Lurch!
- 44426 / 56528 Zombie Explosion!
- 51230 Retch!
- 56560 Beckoning Groan
- 57596 Mangle!
- 51244 Water Breathing

There is also `48309 Cure Infection`, which is the event's own cure spell, plus `45320 You're Re-infected!` and `61454 Zombie Dies`. 43869 brings the screen effect (aura 260) and a mechanic immunity mask along on its other two effects for free.

None of that is used here. A zombie in this PR is a player wearing the zombie model on faction template 1945 with no abilities. 1945 is not a zombie faction either, its faction is 1071 "Craig's Squirrels", a Blizzard test faction shared by templates 1936 to 1947. It happens to be hostile to everything, which is presumably why it looked like it worked.

And nothing sets a PvP or FFA flag, so `Unit::_IsValidAttackTarget` (Unit.cpp:10867-10878) still gates player versus player on `IsPvP()`. Guards attack zombies, living players cannot, and zombies cannot attack back. That interaction is the whole point of the event.

So what is here is the shape of the event without the event, and rebuilding it on form 21 and its spell set is both more correct and less work than fixing this up.

For completeness, the CI is red on something unrelated and small: the SQL inserts into `creature (guid, id1, ...)` and `id1` was renamed to `id` in `2026_06_16_00.sql`. The import aborts on that line, which also means the `spell_script_names`, `ScriptName` and `AIName` rows never applied, so none of this has been exercised by a running server yet.

Other notes worth carrying into a future attempt:

- 43958 is not `DurationIndex 0`, it is index 21, which resolves to -1. Either way the fix belongs in `SpellInfoCorrections.cpp` as a `DurationEntry` override rather than `SetDuration` in the aura script. Retail also varied this by event phase.
- The Argent Healer walks `GetMap()->GetPlayers()`, every player on Eastern Kingdoms, every 3 to 5 seconds, to find someone within 6 yards. `Acore::PlayerListSearcher<Acore::AnyPlayerInObjectRangeCheck>` is the usual approach, see `Events/hallows_end.cpp:809`.
- Stripping both auras `OnPlayerLogin` means a relog cures you.
- The spawns are hand placed in Booty Bay with `VerifiedBuild = 0`. The event ran in the capitals, and spawn positions should come from sniffs.

#16256 stays open. If you want to come back to it, starting from form 21 and its spell set is a much shorter path than what is here, and I am happy to look again.

PR #26404, opened 2026-06-30
feat(Core): randomized respawn delays for creatures and gameobjects

### LINKED ISSUE #5751: Feature: divide spawntimesecs into spawntimesecsmin and spawntimesminmax
**Describe your feature request or suggestion in detail**

Some creatures and gameobjects should have variable respawntimes, for example some Water Barrels that can be found all around Azeroth.
vmangos (not sure if other cores too) divided `spawntimesecs` into `spawntimesecsmin` and `spawntimesecsmax`, we should implement this too, to recreate this mechanic.

**Describe a possible solution to your Feature or Suggestion in detail**

Divide the field into two and make the according changes to the core, we can probably take a look at how vmangos implemented this and do something similiar.
For the DB data we should just use the same values for both fields that we have in `respawntimesecs` right now and implement the new DB data afterwards.

**Additional context**

https://github1s.com/vmangos/core search for "m_respawnTime" and "GetRandomRespawnTime()" for an idea how to implement this.

Kitzunu:
## Changes Proposed:

This PR proposes changes to:
- [x] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [x] Opus 4.7

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/5751

## SOURCE:

## Tests Performed:

This PR has been:

Tested by changing respawn min/max in the database, killing a creature and checking remaining respawn time with .npc info

## How to Test the Changes:

1. Check that no respawn times are broken on creatures and gobs
2. Check that respawn timers work with a random range
3. On NPC just do .npc info on a dead creature and see respawn time, and just kill/revive repeat and you will see different respawn times

- [x] Breaking modules


Kitzunu:
https://github.com/azerothcore/mod-zone-difficulty/pull/82
https://github.com/azerothcore/mod-guild-zone-system/pull/23
https://github.com/azerothcore/mod-instanced-worldbosses/pull/15

sudlud (changes requested):
Merged current master into the branch to clear the conflicts, `ObjectMgr` field indices needed reconciling with #27001 and `ForcedDespawn` with #26827.

On `0a55ee4e80`, first four look blocking.

**`GameObject.cpp:895`** unconditional re-roll wipes the forced respawn time `DespawnOrUnsummon` just set. It writes the forced value into `m_respawnDelayTime`, then the next `Update()` tick hits `GO_JUST_DEACTIVATED` and overwrites it from `m_respawnDelayTimeMin/Max`, still at the DB values. Before this PR `SetRespawnTime(forced)` set `m_respawnDelayTime` and this branch read it verbatim.

Affects `instance_molten_core.cpp:157`/`:289` (`Seconds(WEEK)`), `boss_urok_doomhowl.cpp:58` (`Seconds(MONTH)`), `instance_ulduar.cpp:908` (`7_days`). All return after the `gameobject` row's spawn time instead, and MC persists the wrong value via `SaveRespawnTime()`. Can the forced delay survive one cycle, flag set in `DespawnOrUnsummon` and cleared here?

**`Creature.h:297`** `SetRespawnDelay` now writes min and max, and `GetRespawnDelay` returns the rolled sample. So the save/override/restore in `icecrown_citadel.cpp:403`/`:412`, `boss_valithria_dreamwalker.cpp:256`/`:271` and `zone_western_plaguelands.cpp:50-52` pins the creature at whatever it rolled: read 700 out of `[300, 900]`, set 2, despawn, restore 700, min == max == 700 until the grid reloads. `Get`/`SetRespawnDelayRange` pair, or save min/max in those scripts?

**`cs_npc.cpp:1173`** `.npc setspawntime` doesn't refresh the cached `CreatureData`. Harmless before, but `Creature.cpp:2203` now reads `m_creatureData->SpawnTimeSecMin/Max` back in on every forced despawn, so the first evade reverts it (`DespawnOnEvade` passes a positive timer) and a later `.npc move` writes the stale value to the DB via `SaveToDB`. Can we update the cached data next to the DB write?

**`rev_1751212800000000000.sql:13`** only `content_default` moves to three placeholders. `acore_string` 582 also has `locale_deDE` (`SpawnTime: Gesamt:{} Verbleibend:{}`) and `locale_zhCN`, both still on two, and `fmt` ignores the extra arg, so those locales show max where "remaining" was. Can we update both locale columns in the same file?

**modules** mod-progression-system and mod-ptr-template are both in the module CI, but their SQL sits outside what the updater scans (`src/Bracket_*/sql/world/` and `data/sql/optional/db-world/`), so the build stays green while their spawn SQL is still on `spawntimesecs`. Worth a PR each.

**keira3** `spawntimesecs` is in the spawn model types (`creature-spawn.type.ts`, `gameobject-spawn.type.ts`) and both spawn editors, no PR yet.

Non-blocking:

**`ObjectMgr.cpp:2411` and `:2694`** raid guard checks `SpawnTimeSecMax`, so `Min=8d, Max=20d` slips past and `Min=3d, Max=8d` bumps min from 3d to 14d. Key on `SpawnTimeSecMin`?

**`GameObject.cpp:895` and `:955`** `Min=0, Max=600` passes both new checks, and a roll of 0 leaves `m_respawnDelayTime` at 0. Next despawn returns at `:883` before the re-roll, and `isSpawned()` reads 0 as spawned, so the node stays permanently lootable. Reject `Min == 0 && Max > 0` in the loader?

**`GameObject.h:379-380`** comments have the negative case backwards, `:1190` sets min from `SpawnTimeSecMax`. Worth documenting the sign convention for DB authors, `-100 / -300` currently logs "swapping" even though the swap is right.

**`GameObject.cpp:955` and `:961`** `m_respawnDelayTime` assigned the same value twice.

**`cs_npc.cpp:689` and `cs_gobject.cpp:341`** expired case prints a bare `"0"`, everything else goes through `secsToTimeString`.

**`Battleground.cpp:1503`** that block is commented out (still references `data.rotation0`), so the change is a no-op.

PR #26422, opened 2026-07-02
fix(Core/BG): invited count leak when a BG teleport fails + group join checks

FrancescoBorzi:
## Changes Proposed:

When a player accepts a battleground invite but the teleport into the BG map fails (map could not be created, map is full on arrival, or `AddPlayerToMap` fails), the player is sent to their homebind. On that worldport ack the core only wiped the player-`HasFreeSlots` for the whole lifetime of that instance, and the player keeps a dead queue slot occupied until logout. The 60 second queue-side invite expiry cannot catch this case because the queue entry is already erased when the invite is accepted.

This adds the missing cleanup to the failed-landing branch of `HandleMoveWorldportAck`: if the player still holds an invite for that exact instance, decrement the invited count for the team they were invited as and remove the matching queue slot, then wipe the BG data as before. The guard mirrors what `RemovePlayerAtLeave` does for participants, so normal leaves, declined invites and GM `.appear` are unaffected. Any player removed from a live BG map goes through `BattlegroundMap::RemovePlayerFromMap`, which already reconciles the count and clears the invite slot before landing anywhere else, so this new branch can only fire for players who never entered the BG map at all.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any. **Claude Code was used to help analyze the invite bookkeeping paths and implement the fix.**

## Issues Addressed:

- None filed. Found while auditing the battleground queue bookkeeping.

## SOURCE:

This is a server-side bookkeeping bug, so none of the above apply. It was found and verified by tracing every path that increments and decrements the per-team invited count (`InviteGroupToBG`, `RemovePlayerAtLeave`, the queue-side expiry and the worldport ack handler).

## Tests Performed:

This PR has been:

Builds without errors or warnings (macOS arm64, clang). Not yet tested in-game.

## How to Test the Changes:

The failure branch is hard to hit naturally (it needs map creation to fail or the map to be full at arrival), so reproduce it with a temporary debug hack:

1. In `WorldSession::HandleMoveWorldportAck`, temporarily force the `AddPlayerToMap` failure branch when the destination is a BG map, so every BG port fails and sends the player to homebind.
2. Use `.debug bg` so battlegrounds start without the minimum player requirement.
3. Queue for a battleground, accept the invite. You land at your homebind instead of the BG.
4. Without this fix the BG instance is never deleted, because its invited count never returns to zero (easiest to observe with a temporary log line printing `GetInvitedCount` in `Battleground::Update`). With the fix the emptied instance is cleaned up on the next BattlegroundMgr update and the player's queue slot is freed.
5. Remove the debug hack and check for regressions: play a few normal matches (join, fight, leave, no premature instance deletion mid-match), let an invite expire without accepting (it should still be removed after 60 seconds), and have a GM `.appear` into a running BG and out again (counts must not change).

- [ ] A related but separate leak exists when a player queued for two BGs accepts the second invite while already in transit to the first. That one happens in the accept handler itself and needs a change in `RemovePlayerAtLeave` for non-participants, which is deliberately out of scope here to keep this PR minimal.

If you do run the in-game validation before opening the PR, tick "Tested in-game by the author" and replace "Not yet tested in-game" with what you actually observed.

### DIFF
--- a/src/server/game/Handlers/BattleGroundHandler.cpp
+++ b/src/server/game/Handlers/BattleGroundHandler.cpp
@@ -252,7 +252,7 @@ void WorldSession::HandleBattlemasterJoinOpcode(WorldPacket& recvData)
             }
         });
 
-        if (err)
+        if (err > 0)
         {
             err = grp->CanJoinBattlegroundQueue(bg, bgQueueTypeId, 0, bg->GetMaxPlayersPerTeam(), false, 0);
         }
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -154,7 +154,7 @@ void WorldSession::HandleMoveWorldportAck()
         if (!mEntry->IsBattlegroundOrArena())
         {
             // release the unconsumed invite, otherwise the BG never satisfies its empty + uninvited deletion gate
-            if (Battleground* bg = _player->GetBattleground())
+            if (Battleground* bg = _player->GetBattleground(true))
                 if (_player->IsInvitedForBattlegroundInstance(bg->GetInstanceID()))
                 {
                     bg->DecreaseInvitedCount(_player->GetBgTeamId());
--- a/src/server/game/Handlers/MovementHandler.cpp
+++ b/src/server/game/Handlers/MovementHandler.cpp
@@ -152,7 +152,17 @@ void WorldSession::HandleMoveWorldportAck()
     {
         // but landed on another map, cleanup data
         if (!mEntry->IsBattlegroundOrArena())
+        {
+            // release the unconsumed invite, otherwise the BG never satisfies its empty + uninvited deletion gate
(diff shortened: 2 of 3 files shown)


Nyeriah (approved):

sudlud:
claude pointed out that this bundles 3 unrelated changes: the invited-count leak fix, the `err > 0` group-join fix, and the `Battleground.Random.ResetHour` change (the last one isn't in the title or description at all). All three look correct, but worth splitting or at least documenting all three. Copilot's `GetBattleground()` null concern is already handled by the `(true)` arg.

Francesco-Borzi:
@sudlud that is correct, I bundled 3 fixes here because they were very small (one of them was just 1-line change)

PR #26428, opened 2026-07-02
fix(Scripts/VaultOfArchavon): VoA bosses now turn to stone and players are kicked before Wintergrasp

### LINKED ISSUE #26411: [Vault of Archavon] Mechanics close to Wintergrasp start not implemented
### Current Behaviour

Currently, you can fight bosses normally even if Wintergrasp would be about to start.
Also, you are not removed from the instance if the battle for Wintergrasp starts while inside the raid. 

### Expected Behaviour

When The Battle for Wintergrasp is <= 5-10 minutes away, all the bosses should turn to stone unless engaged. 

When The Battle for Wintergrasp starts, any group inside the instance should be kicked out.

### Source

https://www.wowhead.com/wotlk/spell=63080/stoned#comments:id=739108

https://www.wowhead.com/wotlk/zone=4603/vault-of-archavon#comments:id=617416

https://www.wowhead.com/wotlk/zone=4603/vault-of-archavon#comments:id=1042144

1. Use the following commands to manipulate the state of WG:
.bf timer 15m / 10m / 5m / 10s (15 minutes, 10 minutes, 5 minutes, 10 seconds)
2. Observe you can enter at 15 minute mark, but not at <10min
3. Go inside the raid with 15m remaining.
4. Manipulate the clock to say, for example, 10m 1s and wait few seconds, do the the same for lower amounts
5. Manipulate the clock to 10s and wait that time. You can type .bf queue to see war is in progress.
6. Observe all bosses are still possible to engage and your group has not been teleported out of the instance.

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. be01c9fc48af 2026-06-30 22:48:14 +0200 (master branch) (Win64, RelWithDebInfo, Static)

### Operating system

Windows 11 x64

### Custom changes or Modules

_No response_

Chinoske:
## Changes Proposed:

Issue title claims the pre-Wintergrasp VoA mechanics (bosses turn to stone, players kicked out) aren't implemented - they actually are (`SPELL_STONED_AURA = 63080`, `Wintergrasp.KickVoAPlayers` config, defaults to `true`), but they don't reliably *fire*.

Root cause: the checks used 60-second-wide time **windows** (e.g. `timer <= 10min && timer >= 9min`) evaluated on a fixed **60-second poll**. If the Wintergrasp timer moves through a window faster than the next poll (a GM-set jump via `.bf timer`, but this can in principle also happen from ordinary phase misalignment), the window is skipped entirely and none of the mechanics fire for that whole pre-war cycle - matching the issue's repro steps (`.bf timer 15m` -> `10m` -> `5m` -> `10s` with only a few seconds between each).

Fix: replaced the window checks with **threshold-crossing checks gated by one-shot flags** (`warned15`, `stoned`, `warned2`, `kicked`), polled every 5s. Each stage fires exactly once as soon as the timer is observed at or below its threshold, regardless of how it got there - no window to miss.

Also handles the edge case where the timer jumps so far that Wintergrasp's war has *already started* by the next poll: an active war is treated as "timer at 0", so any stage not yet fired gets fired retroactively in the same tick, instead of silently doing nothing because `IsWarTime()` is already true.

Flags reset for a fresh countdown once the timer is next observed back above 15 minutes (previous war cycle fully concluded).

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #26411

## SOURCE:

Code inspection: Verified `Battlefield::Timer` semantics (`Battlefield.cpp`): counts down continuously in both pre-war and war states; `.bf timer` (`cs_bf.cpp`) calls `SetTimer()` directly on the same `Timer` field the fix reads.

## Tests Performed:
This PR has been:

Applied to a local test server and reproduced the original bug (stepped `.bf timer` calls worked, but a direct 15m->10s jump did nothing) with the first version of the fix, then fixed the war-already-started edge case and confirmed both the stepped case and the direct-jump case now correctly fire the 15-min/2-min warnings, boss stoning, and player kick-to-homebind.

## How to Test the Changes:

1. In Vault of Archavon's zone, use `.bf timer <value>` to move the Wintergrasp pre-war timer through the warning/stoning/kick thresholds, both in small steps (e.g. 15m -> 10m -> 5m -> 10s) and as a single direct jump (15m -> 10s).
2. Confirm the 15-minute and 2-minute warnings, boss stoning, and player kick-to-homebind all fire in both cases.
3. Let the timer reach 0 and Wintergrasp's war start - confirm any stage that hadn't fired yet fires retroactively instead of being skipped.

- None

🤖 Generated with Claude Code (Sonnet 5)+AzerothMCP

### DIFF
--- a/src/server/scripts/Northrend/VaultOfArchavon/instance_vault_of_archavon.cpp
+++ b/src/server/scripts/Northrend/VaultOfArchavon/instance_vault_of_archavon.cpp
@@ -95,6 +95,12 @@ public:
             // Far from the next war again: previous cycle is fully over, reset for a fresh countdown.
             if (!warTime && timer > 15 * MINUTE * IN_MILLISECONDS)
             {
+                if (stoned)
+                    for (uint8 i = 0; i < MAX_ENCOUNTER; ++i)
+                        if (Creature* cr = instance->GetCreature(bossGUIDs[i]))
+                            if (!cr->IsInCombat())
+                                cr->RemoveAllAuras();
+
                 warned15 = false;
                 stoned = false;
                 warned2 = false;
--- a/src/server/scripts/Northrend/VaultOfArchavon/instance_vault_of_archavon.cpp
+++ b/src/server/scripts/Northrend/VaultOfArchavon/instance_vault_of_archavon.cpp
@@ -53,6 +53,9 @@ public:
             KoralonDeath = 0;
             checkTimer = 0;
             stoned = false;
+            warned15 = false;
+            warned2 = false;
+            kicked = false;
         }
 
         void OnPlayerEnter(Player* ) override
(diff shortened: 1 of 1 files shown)


Chinoske:
Pushed a follow-up: the timer-based flag reset cleared `stoned` but never removed the actual aura from bosses (only `OnPlayerEnter` did that, and it fires too late since the timer reset happens within 5s of `.bf stop`/war ending). Now the aura is cleaned up in the same place the flag resets. Confirmed on a local test server: `.bf start` → bosses get Stoned → `.bf stop` → aura clears within ~5-10s without needing to re-enter the instance.

Chinoske:
Addressed both CodeRabbit findings:
- Reset now triggers on the war-active→inactive transition (new `wasWarTime` bool) instead of relying on `GetTimer() > 15min`, which would never be true again if `Wintergrasp.NoBattleTimer` is configured below 15 minutes — good catch, that was a real latent bug.
- Wrapped the `TeleportTo` call that exceeded 120 columns.

Chinoske:
Re-reviewed the reworked timer logic end to end. The flag-based thresholds re-checked: each of warned15/stoned/warned2/kicked fires exactly once per countdown regardless of tick spacing, the war-active state maps to timer=0 so a jump straight past every threshold still fires them retroactively, and the wasWarTime transition is what resets the cycle - so a short Wintergrasp.NoBattleTimer config can't leave the flags stuck true, per the earlier review round. The stoned-aura cleanup on war end also re-verified (removes from bosses out of combat only). No changes.

Chinoske:

PR #26438, opened 2026-07-03
chore(deps/jemalloc): Fix jemalloc compilation

airbreather:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- ~~[ ] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.~~

No AI tools were used in preparing this pull request, though some of the changes were prompted by https://github.com/azerothcore/azerothcore-wotlk/pull/25913#discussion_r3278818390 and https://github.com/azerothcore/azerothcore-wotlk/pull/25913#discussion_r3278818412 from Copilot's review of the PR they came from.

## Issues Addressed:

- Fixes compiling on certain systems like Arch with libstdc++ 16 or higher.

## SOURCE:

- [x] #25913 before it changed direction. The commits were untouched from those first few.

## Tests Performed:

This PR has been:

## How to Test the Changes:

- [x] This pull request requires further testing. Compiled and ran worldserver and authserver on arch linux.

Exactly the same as #25913 but safer.

- [ ] This does not replace the need for #25913, it only makes it feel less urgent. I would like nothing more than to abandon this PR because jemalloc has been replaced by something more modern, I just don't like seeing what could be a one-line spot-fix blocked behind that entire effort.


airbreather:
fun fact that I don't know where else to put: in both [clang](https://github.com/llvm/llvm-project/blob/22e5273cc3334906fc18fce2aebb74986259e31d/libcxx/src/new_helpers.cpp#L22) and [gcc](https://github.com/gcc-mirror/gcc/blob/a8072f31bc90000b3b8a2c0c7819710cf924727a/libstdc%2B%2B-v3/include/bits/new_throw.h#L55), `std::__throw_bad_alloc()` is implemented by literally running `throw std::bad_alloc()` unless the project is using `-fno-exceptions`

Kitzunu:
I want you to take https://github.com/azerothcore/azerothcore-wotlk/pull/25913#issuecomment-4505582596 into concideration

airbreather:
> I want you to take [#25913 (comment)](https://github.com/azerothcore/azerothcore-wotlk/pull/25913#issuecomment-4505582596) into concideration

I did mention it in https://github.com/azerothcore/azerothcore-wotlk/pull/25913#issuecomment-4877020154

if the plan is to keep jemalloc indefinitely rather than replace it like in that other PR, then IMO either patching or editing it like this would make sense for different reasons:
- with patching, the original file stays intact, so a future maintainer won't have to take the time to look through another change in the history for any project-specific tweaks that would need to be re-applied.
- with just editing it inline, as long as it's clear enough from the history that this happened and why, it's one less extra task that the future maintainer will have to do during the update on top of pulling in the new files from upstream

either way can work IMO - if the patch variant is needed, it'll take me a bit longer, but it's not a dealbreaker

Kitzunu:
Im happy if you add a `.diff` file to the jemalloc root folder with the changes.

CheckYourFax:
I think this should be high-prio now ubuntu 26.04 is officially supported.

Because clang is the default compiler used by the build config and it requires libstdc++6 and libstdc++-16-dev

This means currently with everything on default on ubuntu 26.04, builds fail.

For now only https://github.com/azerothcore/azerothcore-wotlk/pull/26438/changes/80f737b364cc7c80a4dbb83c11ce9bb836101fc9#diff-4e1bf59578589e4707acb1c92215bc7d082ab5b7ba779dc71da9140fdd260ae4L70-R70

is required to fix compilation on ubuntu 26.04

GCC still compiles fine because the default (15.2) uses the older libstdc++

Kitzunu:
CI 26.04 builds without issue.

But this PR is just waiting for the author and then it is good to go :)

airbreather:
> Im happy if you add a `.diff` file to the jemalloc root folder with the changes.

Does this just mean that the `.diff` file would serve as documentation of what the change was, rather than that the patching happens when the build happens? If so, we probably also ought to remove `TC_NOTE.txt` which looks like it was originally added to serve the same purpose and has since become obsolete - and it seems a bit unnecessary considering that this is going to be in the Git history (doubly unnecessary since there's reason to believe that #25913 will make anything we do here irrelevant in the not-too-distant future, now that this issue has brought the topic up for discussion).

Otherwise, I'm not seeing a straightforward way to patch the file just at build time. There don't seem to be any current examples, and I'm not scanning through the history to see how it would have been done in the past. The options I can see:
1. put .patch file in deps/jemalloc, apply patch at build or configure time, leave the dirty edit in the working area (easy enough, we can use `git apply`), do `$something` to ensure that we don't try to apply it a second time.
2. put the patch there, apply the patch when we go to build, but undo the patch after compiling that one file so that it can be re-applied later and doesn't sit around in the working tree as a pending change (same basic idea as before, except then a poorly timed cancellation could leave your local build broken without manual intervention that goes beyond just "delete `build` --> re-run CMake")
3. rename `jemalloc_cpp.cpp` to something like `jemalloc_cpp.cpp.in.cmake` or something, have CMake run the patch to make the actual `jemalloc_cpp.cpp` file, and `.gitignore` the `jemalloc_cpp.cpp`. but this feels like it has an even worse version of the same flavor of problem that the one-line edit would have.
4. make the change directly to the file

Do you have any suggestions here? To reiterate, I don't actually see an overt problem making a trivial one-line change to a dependency directly, especially (but not exclusively) because the dependency is likely to be replaced by another, especially (but not exclusively) because the problem would be fixed just as well by pulling in a newer version of the dependency that will include the fix once such a version is tagged, but I'm of course not the one calling the shots here.

Kitzunu:
I will apologize in advance because I am lazy and that is too much text.

What we will do is, you keep the changes in this PR, but also add a .diff file just to somewhat hold a changelog of custom changes to a dep.

So all that is missing is a diff file called `Hotfix throw bad_alloc.diff` or something similar, then we merge it.

airbreather:
should we also delete `TC_NOTE.txt` which looks like it was intended to do the same thing but has become out of date?

Kitzunu:
Eh, keep it. Doesnt hurt

Kitzunu (approved):

CheckYourFax:
> CI 26.04 builds without issue.
> 
> But this PR is just waiting for the author and then it is good to go :)

Edit: I found what the issue is. Clang requires libstdc++6 installed which is provided by default by gcc-16. clang will then force use the libstdc++6 provided by gcc-16, which dropped the ancient helpers from the dev headers.

I don't know how your CI somehow still uses libstdc++-15-dev. On my install clang fails to compile a test program without libstdc++-16-dev. Idk man. Ubuntu package management is a mess.

No matter. This fixes it by using standard C++.

PR #26451, opened 2026-07-04
fix(Core/Unit): require contested guard attacker for friendly-target attack bypass

### LINKED ISSUE #24561: (Battleground) Teammates in Wintergrasp take damage from AoE attacks
### Current Behaviour

Aoe Attacks and CC AoE (like hunter trap) works against allies and own pets.

https://youtu.be/MuECAIf1u58

### Expected Behaviour

AoE Attacks and CC should have no effect on teammates

### Source

_No response_

1. Queu into Wintergrasp
2. Use an AoE effect close or aimed at a teammate
3. It will do damage

### Extra Notes

Right now I'm thinking the culprit behind the issue could be mod-cfbg but I don't know why it causes it.

### AC rev. hash/commit

Image

### Operating system

Ubuntu 24.04

### Custom changes or Modules

Modules are:
mod-cfbg
mod-ale
mod-individual-xp
mod-transmog

Nyeriah:
## Changes Proposed:

e6e6c6289 (#23935) added a bypass to the friendly-target rejection in `Unit::_IsValidAttackTarget` so contested guards (Booty Bay/Gadgetzan/Area 52/K3 bruisers) can attack players flagged with `PLAYER_FLAGS_CONTESTED_PVP` even though the reaction between them is friendly.

The bypass was granted when *either* the attacker is a contested guard *or* the target's owner carries the contested flag:

```cpp
if (!isContestedGuard && !isContestedPvp)
    return false;
```

The second grant makes any contested-flagged player attackable by everyone, teammates included. This is most visible in Wintergrasp:

- Wintergrasp has 40 spawns of Wintergrasp Detection Unit (entry 27869, faction template 2135, which carries `FACTION_TEMPLATE_FLAG_ATTACK_PVP_ACTIVE_PLAYERS`). They satisfy the contested-guard proximity check in `Unit::SetContestedPvP`, so nearly every battle participant is flagged after their first hit on an enemy player.
- The 30 s contested timer does not tick while in combat (`Player::UpdateContestedPvP`), and a WG battle is near continuous combat, so the flag sticks for the whole battle.
- Everyone is force PvP-flagged during the war, so after the bypass the final PvP block returns true via `target->IsPvP()`.
- Only server-side area target selection (`TARGET_CHECK_ENEMY`) is affected; the client still blocks direct hostile casts on allies. Hence enemy-only AoE: Shadowfury, Earthgrab Totem, hunter traps. Own pets are hit too, because the check reads the pet owner's flag.
- Battlegrounds are unaffected because `SetContestedPvP` early returns for players in a BG; Wintergrasp is a Battlefield.

This PR requires both conditions, so the bypass only applies to actual contested guards attacking flagged players. Player attackers never have a contested-guard faction template, which restores the pre-#23935 behavior for players. The scenario #23935 fixed still works: the guard grants both conditions, and its own reaction to the flagged player is already hostile via `GetFactionReactionTo`, while the flagged player's friendly reputation reaction to the guard no longer blocks the attack.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Claude Code (Claude Fable 5) was used to investigate the root cause and prepare the fix.

## Issues Addressed:
- Closes #24561

## SOURCE:

Shadowfury and similar spells are enemy-only per Wowhead; AoE friendly fire never existed in WotLK outside FFA situations. Also reported on ChromieCraft: https://github.com/chromiecraft/chromiecraft/issues/9715 (analysis in the issue comments), where a two day test with crossfaction Wintergrasp disabled ruled out crossfaction modules.

## Tests Performed:
This PR has been:

Builds without errors. Root cause and fix verified by code analysis (see reproduction chain above); not yet runtime tested.

## How to Test the Changes:

Friendly fire (needs 3 players, two on the same faction):
1. Start a Wintergrasp battle with same-faction players A and B and one enemy player.
2. Near a tower or fortress wall (within 45 yd of a Wintergrasp Detection Unit), have A hit the enemy once. A now has `PLAYER_FLAGS_CONTESTED_PVP` (0x100).
3. While A is still in combat, have B cast Shadowfury or drop an Earthgrab Totem on A's position.
4. Before this PR: A takes the damage/stun/root. After: A is unaffected.

Contested guard regression check (behavior from #23935 must be kept):
1. Flag two players of opposite factions for PvP and have them fight next to a Booty Bay bruiser (or Area 52 / K3 bruiser).
2. The bruiser must still attack the contested-flagged players.
3. Unflagged players of either faction must not be attacked by the bruiser.

- [ ] Runtime testing per the steps above.

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10835,7 +10835,7 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
         if (Player const* player = target->GetCharmerOrOwnerPlayerOrPlayerItself())
             isContestedPvp = player->HasPlayerFlag(PLAYER_FLAGS_CONTESTED_PVP);
 
-        if (!isContestedGuard && !isContestedPvp)
+        if (!isContestedGuard || !isContestedPvp)
             return false;
     }
 


Nyeriah:
Tested both scenarios fixed by https://github.com/azerothcore/azerothcore-wotlk/pull/23935

Area52 mobs and nesingwary guards still work as they did before this PR

This is ready to be shipped

PR #26461, opened 2026-07-04
feat(Core/Battlefield): defer server restart during Wintergrasp

Nyeriah:
## Changes Proposed:

Adds a new config option `PreventWintergraspShutdownTimer` (default `0`, disabled). When a scheduled server shutdown or restart countdown reaches zero while a Wintergrasp battle is in progress, the restart is deferred instead of stopping the world. It is rescheduled to the battle's remaining time plus the configured number of buffer minutes (so `PreventWintergraspShutdownTimer = 5` gives "remaining battle time + 5 minutes").

Players online at the moment of deferral receive a localized message explaining that the restart was postponed because a Wintergrasp battle is running. The follow-up countdown then reflects the new time through the normal restart/shutdown server message.

- New config `PreventWintergraspShutdownTimer` in `worldserver.conf.dist` and `WorldConfig`.
- Deferral hook in `World::_UpdateGameTime` via a new `World::RescheduleShutdownForWintergrasp()` helper. When the feature is disabled the existing shutdown code path is unchanged (the helper returns early and control falls through to the original branches).
- New `acore_string` entry (35455) with translations for all supported locales, sent per-player in each player's own session locale. Idle shutdowns are not affected.

## Issues Addressed:

- N/A

## SOURCE:

Local development.

## Tests Performed:

Builds locally. C++ and SQL codestyle checks pass (`apps/codestyle/codestyle-cpp.py`, `apps/codestyle/codestyle-sql.py`).

## How to Test the Changes:

1. Set `PreventWintergraspShutdownTimer = 5` in `worldserver.conf` and ensure Wintergrasp is enabled.
2. During an active Wintergrasp battle, issue a `.server restart <short-time>` (e.g. 60 seconds) so the countdown elapses mid-battle.
3. When the countdown reaches zero the server does not restart. It reschedules to the battle's remaining time + 5 minutes, and online players receive the postponement message.
4. With `PreventWintergraspShutdownTimer = 0` (default), confirm the restart proceeds normally regardless of Wintergrasp state.


src/server/game/World/World.cpp
```diff
@@ -1447,6 +1448,11 @@ void World::_UpdateGameTime()
     ///- if there is a shutdown timer
     if (!IsStopped() && _shutdownTimer > 0 && elapsed > 0s)
     {
+        ///- ... if it is overdue but a Wintergrasp battle is running and deferral is enabled,
+        ///  push the restart past the end of the current battle and keep the world running
+        if (_shutdownTimer <= elapsed.count() && RescheduleShutdownForWintergrasp())
+            return;
+
         ///- ... and it is overdue, stop the world (set m_stopEvent)
         if (_shutdownTimer <= elapsed.count())
```
Kitzunu: the first check could be merged under the second to prevent double `_shutdownTimer <= elapsed.count()`, but cosmetic

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -3748,6 +3748,17 @@ Wintergrasp.KickVoAPlayers = 1
...
+#
+#     PreventWintergraspShutdownTimer
+#         Description: Defer a scheduled server shutdown/restart if a Wintergrasp battle
+#                      is still in progress when the countdown elapses. The restart is
+#                      rescheduled to the battle's remaining time plus this many minutes
+#                      of buffer. Idle shutdowns are not affected.
+#         Default:     0 - (Disabled)
+#                      N - (Enabled, defer to remaining battle time + N minutes)
+
+PreventWintergraspShutdownTimer = 0
```
Kitzunu: I would rename it to something like `Wintergrasp.DeferShutdownTimer` just to keep with the `Wintergrasp.` config names

Kitzunu (approved):
Approving but leaving the reviews up to you

PR #25836, opened 2026-05-14
ci(build): support Ubuntu 26.04

FrancescoBorzi:
Onhold until GH Actions adds 26.04

PR #26481, opened 2026-07-05
fix(CI): stop ccache from freezing after the first run on a branch

Kitzunu:
## Changes Proposed:

- **ccache no longer freezes** — the primary cache key now includes `github.run_id`, with `restore-keys` falling back from the newest cache on the branch to progressively broader prefixes. A static key hits on itself from the second run onward, and `actions/cache` then skips the upload, so the cache froze after the first run on a branch and permanently on master. Applies to the Linux and macOS builds.
- **`CCACHE_COMPRESSLEVEL` 9 → 3** — level 9 burns CPU for marginal size gains.
- **Stale runs cancelled** — `concurrency` added to the codestyle and sql-codestyle workflows, matching what the build workflows already do.
- **macOS concurrency group** fixed; it was a malformed expression that never evaluated.
- **Docker actions bumped** — `build-push-action` v5 → v6, `login-action` v1 → v3, `repository-dispatch` v2 → v3.

### Evidence for the freeze

Repo cache listing today: master ccache entries sit at 44-58 MB, having been saved once and never refreshed, while active PR entries are 99-372 MB for the same compilers. Master has been building against a near-empty cache.

### Removed from this PR

- **Single build leg for DB-only PRs** — superseded by #26716, which uses a `paths-ignore: data/**` filter on pch/nopch. That skips the core builds entirely instead of running one leg, needs no duplicated bash in two workflows, and relies on Dashboard CI, which has no paths filter and already imports the SQL and starts a real worldserver.
- **Docker GHA layer caching** — `cache-to: type=gha` writes into the same 10 GB repo-wide Actions cache that ccache uses. That pool is already over its limit (10.7 GB active), so GitHub evicts LRU continuously, and `mode=max` across six components would evict the master ccache entries this PR exists to keep warm. Worth revisiting once the pool has headroom. The v6 bump is kept.
- **dashboard-ci ccache changes** — split out for separate review, since `dashboard-ci.yml` is under CODEOWNERS (@Yehonal). The same keying fix applies there and can follow in its own PR.
- **`CCACHE_MAXSIZE` 5G → 2G** — a no-op. The largest ccache entry in the repo today is 372 MB, so 5G was never the binding limit.

### AI-assisted Pull Requests

> [!IMPORTANT]
>

- [X] Opus 4.8

## Tests Performed:

The earlier fork tests covered the dynamic build matrix, which is no longer part of this PR. What remains is cache keying and workflow metadata, best judged on master over the next several commits by watching whether master ccache entries start growing past the ~50 MB they have been stuck at.

## How to Test the Changes:

After merge, check the repo cache listing. Master ccache entries should refresh on every push and grow toward the 100-370 MB range that PR branches already show, instead of staying frozen at their first-run size.


sudlud:
Synced the branch with master. The `full` matrix in both setup jobs predated #25836, so as-is it would have re-added ubuntu-22.04/clang-15 and dropped ubuntu-26.04/clang-20. The merge commit fixes that: matrices now match current master (clang-18 / gcc-14 / clang-20 for nopch, clang-18 / clang-20 for pch), just wrapped in the dynamic `fromJson` output.

Notes on the bot findings:
- dashboard-ci "dead restore-keys" (CodeRabbit, flagged Major): not dead. The `:true:pch=false:` etc. entries are valid literal prefixes of the linux-build action's ccache keys, and the GHA cache is repo-scoped, so they act as intentional cross-workflow fallbacks. Leaving them.
- `github.base_ref` template injection: base_ref can only be an existing upstream branch name, not attacker-controlled, so real exploitability is nil. Moving it to `env:` is fine as tidy-up but not blocking.
- persist-credentials: false on the setup checkout: valid minor hardening, optional.
- docker per-component `mode=max`: reasonable; worth watching cache eviction once it lands.

sudlud:
Note on the `if: always()` ccache save (dashboard-ci, and linux-build's post-save): combined with the run-id key it now saves on every run, including cancelled/failed ones. An early-cancelled run can leave a thin cache that wins the `-<ref>` prefix match, so the next run restores less. ccache stays correct, it self-heals after one run, and caching a failed build is actually useful, so not blocking. Flagging it as a conscious choice. `if: success()` avoids it, at the cost of not caching failed builds.

sudlud (approved):
Synced with master and fixed the compiler-matrix drift (would have re-added clang-15/22.04 and dropped clang-20/26.04). Logic reviewed, no blocking issues. DB-only single-leg is the intended behavior. Minor cache note above is non-blocking.

sudlud:
Holding off on merge, I think there's a much simpler way to get the DB-only saving.

`dashboard-ci` already validates DB-only PRs end to end: same repo/draft gate, no paths filter so it runs on every PR, and `./acore.sh init` does the full SQL import followed by `worldserver -dry-run` and a real worldserver start. More coverage than a core-build leg gives, and it took 22m here vs ~50m for the core legs.

So instead of the `setup` job, pch/nopch can opt out natively:

```yaml
  pull_request:
    types: [opened, reopened, synchronize]
    paths-ignore:
      - 'data/**'
```

`data/` contains only `sql/`, so that is exactly "DB-only PR". `push: master` stays unfiltered. DB-only PRs then use 0 core legs instead of 1, and both files lose the duplicated bash, the `fromJson` and the `needs:` (also resolves Copilot's duplication comment).

One thing to settle either way: the repo cache pool is already at 9.94/10 GB across 65 entries. Docker `cache-to: mode=max` for six components writes into that same pool and will evict the ccache entries this PR is trying to keep warm. `mode=min`, or leaving docker cache out, is safer.

The ccache freeze diagnosis itself is confirmed by the cache listing: master entries sit at 40-50 MB (saved once, never refreshed), fresh PR entries are 100-320 MB. Side note, `MAXSIZE 5G -> 2G` is a no-op, real ccache dirs never exceed ~0.3 GB.

sudlud (changes requested):
see review comment

sudlud:
this PR actually contains quite a few changes unrelated to what the title suggests.

I just opened #26716 with the db-only approach for the dashboard CIs, so this PR here could now be reduced to contain the remaining optimizations only

sudlud (approved):

PR #26486, opened 2026-07-05
fix(Core/Mail): load undelivered mails at login

Nyeriah:
## Changes Proposed:

Player reports: "If you open the mailbox before the mail arrives, then you basically have to relog after it has really arrived to see it."

The login query `CHAR_SEL_MAIL` filtered `deliver_time <= now`, so mails still inside their delivery delay (cross-account mail with items or money, COD) were never loaded into the session of a player who logged in during the delay window. After the delivery time passed, nothing in that session could surface the mail: the arrival tick in `Player::Update` never fired (`m_nextMailDelivereTime` was never armed), `HandleGetMailList` only iterates in-memory mails, and `MSG_QUERY_NEXT_MAIL_TIME` reported nothing incoming. The only workaround was logging out and back in after the delivery time.

If the recipient was online when the mail was sent, everything worked, because `SendMailTo` puts the pending mail into memory directly. The bug only hit sessions that started inside the delay window, which is why it looked intermittent.

Fix: load all mails at login regardless of delivery time, as TrinityCore does, and drop the now-unused timestamp binding. Every consumer of the in-memory mail list already filters on `deliver_time` (`HandleGetMailList`, `HandleQueryNextMailTime`, `UpdateNextMailTimeAndUnreads`, and the take/read/return/delete handlers), so pending mails only become visible once due. `UpdateNextMailTimeAndUnreads` runs at the end of `_LoadMail` and arms the delivery notification for them.

This also removes an inconsistency where `CHAR_SEL_MAILITEMS` (which has no delivery filter) loaded the item objects of pending mails into the player's mail item map while the mails themselves were missing from the mail list. That divergence is one of the producers of the empty-return bug addressed in #26485.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

- [x] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any. **Claude Code (Claude Fable 5).**

## Issues Addressed:
- Based on live server reports of delayed mails not appearing until relog. No open issue on the tracker.

## SOURCE:

Ported from TrinityCore, committed with `--author` crediting Shauren: TrinityCore/TrinityCore@57eaab80d9e1f6f25274f376da16e1f2f67ab4bc ("Core/Mail: Load mails at login instead of on demand when queried by packets").

## Tests Performed:
This PR has been:

Passes `codestyle-cpp.py`.

## How to Test the Changes:

1. From character A (account 1), send character B (account 2) a mail with an item attached, so the mail gets a delivery delay (`MailDeliveryDelay`, default 1h; lower it in worldserver.conf to test faster).
2. Log character B in while the mail is still pending and stay online until the delivery time passes.
3. Without this PR: no new-mail notification appears, the mailbox stays empty, and the mail only shows after relogging. With this PR: the new-mail icon appears when the delay expires and the mailbox lists the mail.
4. Regression checks: the pending mail must not be visible or takeable before its delivery time (mailbox list, `.mail list`), and normal instant mails must behave as before.

- [ ] `CHAR_SEL_NEXT_MAIL_DELIVERYTIME` is prepared but never executed anywhere; it looks like a leftover from an earlier approach to this problem and could be removed in a follow-up.
- [ ] `Player::_LoadMail` skips expired mails but leaves the rows in the DB (and leaks the `Mail` object), keeping a small DB/memory divergence window until the startup cleanup runs.

### DIFF
--- a/src/server/database/Database/Implementation/CharacterDatabase.cpp
+++ b/src/server/database/Database/Implementation/CharacterDatabase.cpp
@@ -423,8 +423,6 @@ void CharacterDatabaseConnection::DoPrepareStatements()
     PrepareStatement(CHAR_SEL_CHAR_SOCIAL, "SELECT DISTINCT guid FROM character_social WHERE friend = ?", CONNECTION_SYNCH);
     PrepareStatement(CHAR_SEL_CHAR_OLD_CHARS, "SELECT guid, deleteInfos_Account FROM characters WHERE deleteDate IS NOT NULL AND deleteDate < ?", CONNECTION_SYNCH);
     PrepareStatement(CHAR_SEL_ARENA_TEAM_ID_BY_PLAYER_GUID, "SELECT arena_team_member.arenateamid FROM arena_team_member JOIN arena_team ON arena_team_member.arenateamid = arena_team.arenateamid WHERE guid = ? AND type = ? LIMIT 1", CONNECTION_SYNCH);
-    // No deliver_time filter: mails still in their delivery delay must be loaded so the
-    // session can announce and list them once the delay expires without requiring a relog
     PrepareStatement(CHAR_SEL_MAIL, "SELECT id, messageType, sender, receiver, subject, body, expire_time, deliver_time, money, cod, checked, stationery, mailTemplateId FROM mail WHERE receiver = ? ORDER BY id DESC", CONNECTION_ASYNC);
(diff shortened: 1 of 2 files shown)


PkllonG:
Tested
You can get notifications and receive mails with items without logging in again

Nyeriah:
Thank you very much

Rorschach91 (approved):

PR #26492, opened 2026-07-05
refactor(Core/Misc): enforce east-const style and add codestyle check

Kitzunu:
## Changes Proposed:
Aligns the entire `src/` tree to **east-const** (`T const&` / `T const*`) as required by the [C++ Code Standards](https://www.azerothcore.org/wiki/cpp-code-standards), and adds a codestyle check so it can't regress.

Two parts:

1. **Mechanical east-const sweep** — `const T&` → `T const&`, `const T*` → `T const*` across `src/` (game, common, test, tools). This is a purely syntactic reorder — the two forms are identical to the compiler, so there is **no behavioural change**. The diff is exactly balanced (equal insertions/deletions per source file): nothing but token reordering.

2. **New "Qualifier alignment check" in `apps/codestyle/codestyle-cpp.py`** — flags any newly introduced `const T&` / `const T*` and prints the east-const form to use. It is string/comment/raw-string aware (literals and comments are blanked before matching), so it never fires inside a string or comment. This **replaces** the two older, weaker const checks (the `const auto&` substring check and a pointer regex that only matched `const Foo*bar` with no space — which is why hundreds of west-const declarations had slipped through over the years).

No CI workflow change is needed: `codestyle.yml` already runs this script on every PR.

This PR proposes changes to:

### AI-assisted Pull Requests

The east-const sweep and the codestyle check were prepared with **Claude Code (Claude Opus 4.8)**. The mechanical transform is semantically inert (compiler-identical), and the check was reviewed and stress-tested against adversarial input by the author.

## Issues Addressed:
- Enforces the documented `Type const&` / `Type const*` standard in CI (previously unenforced).

N/A — style-only change. Correctness rests on east-const and west-const being semantically identical to the compiler; verified as a pure reordering (balanced diff, no `const const` artifacts, no string-literal edits, no auto-generated files touched).

## Tests Performed:
This PR has been:
- [x] Tested by the author.

- `python apps/codestyle/codestyle-cpp.py` passes green on the full tree.
- The new check was verified against adversarial valid C++ (member-function `const`/`const&`/`const&&` qualifiers, `const_cast`, `constexpr`, digit separators `1'000'000`, function pointers, already-east forms, `a < b & c`) — **zero false positives**.
- Masking verified against `//` inside strings, `const` inside string/block comments, and raw-string literals `R"(...)"` including embedded quotes.

> Note: a full multi-platform build has not been run locally. Since the transform is compiler-identical, CI's build jobs are the authoritative compile check.

## How to Test the Changes:

1. Run `python apps/codestyle/codestyle-cpp.py` from the repo root — it should pass on the whole tree.
2. Add a `const Foo& x;` (or `const Foo* p;`) anywhere under `src/` and re-run — it should be flagged with the east-const suggestion.
3. Confirm CI's build jobs compile cleanly (the reorder is semantically inert).


sudlud (approved):
Verified the sweep is a pure reorder (every changed line differs only in `const` placement/whitespace, no `const T*` -> `T* const` flips). The check has zero false positives across all 1698 `src/` files and passes an adversarial suite; correctly supersedes the two weaker `const` checks.

Merged master, resolved the 5 conflicts (kept master's changes + east-const), fixed the one new west-const line and the glued `const&x` refs. `codestyle-cpp.py` green. LGTM.

PR #26509, opened 2026-07-06
fix(Scripts/ShadowLabyrinth): stop Cabal reinforcements respawning after Murmur dies

### LINKED ISSUE #26489: Shadow Labyrinth's Cabal Spellbinders and Cabal Summoners continue to spawn after Murmur is dead
### Current Behaviour

In the final room of Shadow Labyrinth, a series of Shadow Council mobs are shown containing Murmur (NPC ID 18708) in several ranks (lines of troops); mobs that are scripted to be slain by Murmur are replaced by new ones, specifically Cabal Spellbinders (NPC ID 18639) and Cabal Summoners (NPC ID 18634). However, even after Murmur is defeated by players, the spellbinders and summoners continue to spawn in and populate the hallway leading up to Murmur. This will only happen if the corpses of the previous spellbinders and summoners have despawned.

### Expected Behaviour

The spellbinders and summoners should stop appearing upon Murmur's defeat; the being they're there to contain is dead, and so they should stop spawning.

### Source

https://youtu.be/ht9781OYx6M?t=2320

Unfortunately, all videos I could find from the TBC/WotLK era are either too low-quality to submit as sources, or do not show the hallway leading up to Murmur after its defeat. This video from TBC Classic _does_ show that hallway, at 38:40 on. No additional mobs appear to contain Murmur, since it is already dead.

From what I recall of playing TBC myself way back in the day, the spellbinders and summoners are not supposed to spawn in if their associated rank of mobs is slain by the players, rather than by Mumur. However, I can find no proof of this in text or video, so it's ultimately just my memory, which could easily be flawed, that asserts this.

1. Create a character and ``.character level 70``
2. ``.tele shadowlabyrinth`` and go inside.
3. Either progress through the instance normally, or use cheats or whatever until you arrive at the final room of the instance where Murmur is
4. Defeat all cabalist mobs and then defeat Murmur, ensuring that the cabalist corpses have all despawned
5. Observe the preceding hallway over time and notice how it continues to populate with spellbinders and summoners

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. https://github.com/azerothcore/azerothcore-wotlk/commit/c62dc737741fc24dd0e7d2c9cfa7a1eb7377f312+ 2026-07-03 09:02:05 -0700 (Playerbot branch) (Win64, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 2001.
Connection peak: 1.
Server uptime: 2 hour(s) 10 minute(s) 51 second(s)
Update time diff: 16ms. Last 500 diffs summary:
|- Mean: 60ms
|- Median: 20ms
|- Percentiles (95, 99, max): 165ms, 178ms, 230ms
Using SSL version: OpenSSL 3.6.0 1 Oct 2025 (library: OpenSSL 3.6.0 1 Oct 2025)
Using Boost version: 1.81.0
Using CMake version: 4.1.2
Using MySQL version: 80407
Found MySQL Executable: C:/Program Files/MySQL/ACServer/bin/mysql.exe
Compiled on: Windows 10.0.26100
MMAPs status: Enabled
maps directory located in D:/Blizzard/WoW/AC Build/bin/RelWithDebInfo/Data/maps. Total size: 291014951 bytes
vmaps directory located in D:/Blizzard/WoW/AC Build/bin/RelWithDebInfo/Data/vmaps. Total size: 657797442 bytes
mmaps directory located in D:/Blizzard/WoW/AC Build/bin/RelWithDebInfo/Data/mmaps. Total size: 2192760616 bytes
Default DBC locale: enUS.
All available DBC locales: enUS frFR deDE zhTW
Using Playerbots DB Revision:
Latest CharacterDatabase update: playerbots_names.sql
Latest WorldDatabase update: zzz_tcg_vendors_setup.sql
LoginDatabase queue size: 0
WorldDatabase queue size: 0
PlayerbotsDatabase queue size: 0
List of enabled modules:
|- mod-ah-bot-plus
|- mod-ale
|- mod-auto-gather-tracking
|- mod-autobalance
|- mod-autofish
|- mod-dynamic-loot-rates
|- mod-fly-anywhere
|- mod-guild-levels
|- mod-individual-progression
|- mod-junk-to-gold
|- mod-multibot-bridge
|- mod-player-bot-level-brackets
|- mod-playerbots
|- mod-rdf-expansion
|- mod-tcg-vendors
|- mod_weather_vibe

### Operating system

Windows 11 x64

### Custom changes or Modules

[AH Bot Plus](https://github.com/NathanHandley/mod-ah-bot-plus)
[AIO](https://github.com/Rochet2/AIO)
[ALE](https://github.com/azerothcore/mod-ale)
[Autobalance](https://github.com/azerothcore/mod-autobalance)
[Autofish](https://github.com/Flerp/mod-autofish)
[Auto-Gather Tracking](https://github.com/Day36512/mod-auto-gather-tracking)
[Dynamic Loot Rates](https://github.com/hallgaeuer/mod-dynamic-loot-rates)
[Fly Anywhere](https://github.com/abracadaniel22/mod-fly-anywhere)
[Guild Levels](https://github.com/Old-Man-Warcraft/mod-guild-levels)
[Individual Progression](https://github.com/ZhengPeiRu21/mod-individual-progression/)
[Junk to Gold](https://github.com/noisiver/mod-junk-to-gold)
[MultiBot Standalone](https://github.com/Wishmaster117/MultiBot-Standalone)
[Playerbots](https://github.com/mod-playerbots/mod-playerbots)
[Playerbot Level Brackets](https://github.com/DustinHendrickson/mod-player-bot-level-brackets)
[RDF Expansion](https://github.com/azerothcore/mod-rdf-expansion)
[Transmog](https://github.com/Thiesant/MOD-TRANSMOG-SYSTEM_LUA-IMPLEMENTATION)
[TCG Vendors](https://github.com/lightninjay/mod-tcg-vendors)
[Weather Vibe](https://github.com/hermensbas/mod_weather_vibe)

lipeter1982:
## Changes Proposed:

The five corridor reinforcements (Cabal Spellbinders/Summoners with a 5s respawn timer, used by Murmur's intro event) kept respawning endlessly after Murmur was killed.

- Track the reinforcements' spawn ids in the instance script and push their respawn out a week once the Murmur encounter is DONE.
- Reinforcements are handled in whatever state they are in when the boss dies: alive (respawn delay extended), dead with corpse (respawn time extended and saved), or dead and already despawned (pending map-level respawn overridden via `Map::SaveCreatureRespawnTime`, since no creature object exists at that point — this was the trickiest case, they despawn 5s after death via SmartAI `FORCE_DESPAWN`).
- The DONE branch in `OnCreatureCreate` covers instances reloaded after the kill, including expired respawn times inherited from an older save.
- Also fixes Murmur's encounter state never being stored: `DATA_MURMUR` was 5 while `SetBossNumber(4)` only allocates ids 0-3, so `SetBossState` silently failed and the DONE state was never persisted. Boss ids renumbered to 0-3.

## Issues Addressed:

- Closes #26489

## SOURCE:

- Video evidence below; respawn mechanics verified against the smart_scripts rows for spawn ids 146225-146229 (5s `spawntimesecs` + on-death `FORCE_DESPAWN`).

## Tests Performed:

- Built on Windows (MSVC, Release) and tested on a live server.
- Before: reinforcements respawn ~10s after Murmur's death, endlessly. https://streamable.com/2i1tt6
- After: killed Murmur with some reinforcements alive, some as corpses and some already despawned (worst case) — no further respawns, including after leaving and re-entering the instance. https://streamable.com/0cgpu8
- Verified with temporary core logging that the pending map-level respawns of already-despawned reinforcements are overridden at the moment the encounter is set DONE.

## How to Test the Changes:

1. Enter Shadow Labyrinth and clear up to Murmur (or GM-kill your way there).
2. While Murmur is alive, note the corridor reinforcements respawning every few seconds (intended intro behaviour).
3. Kill Murmur — ideally right after killing a couple of reinforcements, so some are despawned at that moment.
4. Wait in the corridor: nothing should respawn. Kill any leftover reinforcements: they should not respawn either, including after re-entering the instance.

- None


Nyeriah:
I think linked_respawns need to be used

Nyeriah:
*Edited: I originally suggested re-pointing the `linked_respawn` rows at Murmur. That was wrong and I have withdrawn it below, the original rows look deliberate.*

Following up on my `linked_respawn` comment, since I went and looked at the data.

Those five spawns already have `linked_respawn` rows, pointed at Cabal Executioners:

| guid | entry | linkedGuid | master |
|---|---|---|---|
| 146225 | 18639 Cabal Spellbinder | 146211 | Cabal Executioner |
| 146226 | 18634 Cabal Summoner | 146213 | Cabal Executioner |
| 146227 | 18639 Cabal Spellbinder | 146217 | Cabal Executioner |
| 146228 | 18634 Cabal Summoner | 146221 | Cabal Executioner |
| 146229 | 18639 Cabal Spellbinder | 146221 | Cabal Executioner |

My first read was that these were simply pointed at the wrong master. Checking `creature_formations` says otherwise: all four of those Executioners are formation leaders of the corridor trash groups (146211 leads 146211-146212, 146213 leads 146213-146216, 146217 leads 146217-146220, 146221 leads 146221-146224). The reinforcements are not members of those formations, they are separate spawns hung off each group's leader. That reads as deliberate: every corridor pull has its own endless reinforcement feed, and clearing the group shuts that feed off.

Which means the endless respawning in #26489 is most likely people running past the corridor trash to Murmur. Under the existing design that is working as intended, it just looks broken.

The part that matters for this PR, and that I got wrong in my first pass: `linked_respawn` cannot express what you want here. The primary key is `(guid, linkType)` and `ObjectMgr` keeps it as a single guid to guid map, so there is exactly one master per slave. There is no way to say "stop when the Executioner is dead **or** when Murmur is dead". Re-pointing the rows at Murmur would have traded one gate for the other and lost the clear-the-trash behaviour, so ignore that suggestion.

So the real question before reviewing the implementation is a design one: is Murmur's death supposed to stop the feed at all, or is the Executioner gate the whole intent? If it is the latter, this is not a bug and the PR should shrink to the enum fix. If we do want Murmur's death as a second gate, then it has to live outside `linked_respawn`, and doing it in the instance script is fair enough.

If we go that way, two notes on the current code:

- `GetRespawnDelay() < MINUTE` as the discriminator for the corridor spawns is a hidden dependency on `spawntimesecs` staying at 5. Retune those spawns and the fix silently stops applying, or starts applying to spawns it was never meant to touch. The spawn ids are known and stable, keying off those would be clearer.
- The three-way handling of alive, corpse and already despawned can be collapsed a fair bit, and the tracking set is not really needed if you drive it off the `DONE` state.

Either way, the enum fix is a genuine bug and worth keeping regardless of how the respawn side lands: `EncounterCount` is 4 while `DATA_MURMUR` is 5, so `SetBossState(DATA_MURMUR, DONE)` fails the bounds check and Murmur's DONE state is never stored. Boss ids in use are 1, 2, 3 and 5, with 0 unused. That one is worth splitting out on its own.

lipeter1982:
Thanks for digging into the spawn/formation data, this is really helpful context.

On the design question: I think Murmur's death should stop the feed too, not just the Executioner gate. The report in #26489 is exactly the case you describe — people skip the corridor trash and pull straight to Murmur — and from a player's perspective that reads as a bug regardless of whether the Executioner-gated design was intentional. An endless stream of adds spawning in a corridor nobody is fighting in anymore, after the boss is already dead, is confusing and has no gameplay purpose. So I'd rather keep both gates: Executioner death for players who clear normally, Murmur death as a backstop for players who don't.

I've pushed a commit that addresses your two implementation notes:

- Replaced the `GetRespawnDelay() < MINUTE` discriminator with the five spawn ids you identified (146225–146229), declared as a named constant in `shadow_labyrinth.h`. No more hidden dependency on `spawntimesecs` staying at 5s.
- Extracted the WEEK respawn-suppression logic into a single `SuppressReinforcementRespawn()` helper used by both `OnCreatureCreate` and `SetBossState`, removing the duplication (also addresses CodeRabbit's note on this). The tracking `std::set` is gone too, since we no longer need to discover the spawn ids at runtime — we key off the static list directly.

The enum fix (`DATA_MURMUR`/`EncounterCount` bounds) was already correct in the original commit, so no changes needed there.

Let me know if you'd still rather see this scoped down to just the enum fix — happy to split it if the team prefers the Executioner-only gate as the intended design.

ArchasTL:
Just to be clear, when I opened the issue, I had manually cleared the whole instance, trash included. All enemies leading up to Murmur were slain. I didn't even know there was a way to skip that trash.

lipeter1982:
Never a rouge i see😅

Gultask:
Oh hey I remember doing Shadow Labyrinth ages ago.

The respawning individuals are tied to the horizontal groups, yes.
However, there is something that can happen. If they are hit as they start running, they will be stunned and block the movement. I believe it's because it's not a proper path, only a single point.
If that happens, it's easier to skip past the trash, making the other ones keep respawning.

I can think of delaying the suppression bolts down a bit. ONLY the one that hits 5 NPCs, not the ones that kill.

The part where they respawn because they are linked is, as far as I know, very blizzlike. I spent ages working on those dungeons, believe me.

Gultask:
Or if they ARE all killed but still spawn, might be something deeper in linked_respawn

sudlud:
Renumbering shifts existing instance saves. `ReadSaveDataBossStates` reads positionally with no header guard, so a saved heroic of `"0 3 0 0"` (Hellmaw killed) reloads as Blackheart DONE and opens the Refectory door. Slot 0 was never written, so `DATA_MURMUR = 0` with the others left at 1/2/3 fixes the same bug with no migration.

On the respawn side, 146225-146229 link to Executioners with `spawntimesecs = 86400` and `Creature::Respawn` holds a slave while its master is dead, so clearing a group does stop its feed. That does not explain @ArchasTL clearing all trash and still seeing respawns. Which Executioners were alive?

Gultask:
yeah they have GUIDs, it's not summons. A .npc guid on target would clear some questions

sudlud:
@lipeter1982 can you please address the reviews

sudlud:
closing this for now. please feel free to reopen anytime in the future addressing the review comments. thanks!

Issue #120, opened 2016-08-29
Visual Stuido 2013 Compile error


talamortis:
**Description**:

**Current behaviour**: Tell us what happens
Getting 3 errors and 1 warning.

**Errors**:

1. error C2664: 'std::atomic_long::atomic_long(const std::atomic_long &)' : cannot convert argument 1 from 'int' to 'const std::atomic_long &'	E:\AzerothCore\src\game\Server\WorldSocketMgr.cpp	50	1	game

2. error C2664: 'std::atomic_long::atomic_long(const std::atomic_long &)' : cannot convert argument 1 from 'int' to 'const std::atomic_long &'	E:\AzerothCore\src\game\Scripting\ScriptMgr.cpp	154	1	game

3. error LNK1104: cannot open file '..\game\Debug\game.lib'	E:\AzerothCore\build\src\worldserver\LINK	worldserver

Warning
1. Warning	1	warning C4018: '>=' : signed/unsigned mismatch	E:\AzerothCore\src\game\Handlers\MiscHandler.cpp	401	1	game

**Operating system**:  Windows 10 x64 Bit

**Program Used to compile**: Visual Stuido 2013

DantestyleXD:
Same errors here

----------------------------
Server: Local (Repo AzerothCore)
Last commit hash tested:  2407208eb6288db753a5611fb0a81205aee03410
O.S.: Windows 10
Compiler: MS Visual Studio 2013

ShinDarth:
@talamortis @DantestyleXD  can you please tell us which is the latest commit that works for you?

you can change commit using:

`git checkout COMMIT_HASH`

hero2880:
I also like this, the first version can be compiled successfully; later versions have failed. Also used VS2013. However, the recent compilation of TC and SINGLECORE succeed.

ShinDarth:
@hero2880 what is the LATEST commit that you can successfully compile?

hero2880:
@ShinDarth ，it‘s’ Version July 28, I was successful compilation at July 30.

ShinDarth:
@talamortis  @DantestyleXD  @hero2880 can you guys please test it using Visual Studio 2015?

By the way, Visual Studio 2013 is deprecated now: https://community.trinitycore.org/topic/12805-visual-studio-2013-end-of-life/

Chrisjiajun:
Me too, I hope to fix. 
[...]

PR #26513, opened 2026-07-06
Core/Maps: Implement Multi-threaded Map Partitioning

### LINKED ISSUE #22571: Feature: Map Partitioning Support for AzerothCore

Hi,

I’d like to propose implementing map partitioning for large continent maps, inspired by the following TrinityCore PR:

- https://github.com/Project-Epoch/TrinityCore/pull/120

### The Problem

Currently, all continent maps (like Eastern Kingdoms or Kalimdor) are processed on a single thread. With thousands of entities and players, this becomes a performance bottleneck. While instances benefit from parallel processing via separate map threads, continents remain monolithic, even under heavy load.

### The Proposed Approach

The PR from TrinityCore introduces a `MapPartitioned` class that allows continents to be split into smaller logical partitions. These partitions share the base map data and are processed in parallel, significantly reducing per-tick load.

Notable points from the implementation:

- Seamless player and entity movement across partitions (no loading screens)
- Entities are assigned dynamically to the correct partition on startup
- GridUnload removed to simplify logic and improve performance
- Shared systems (e.g., weather) remain consistent via the base map
- Cleaner structure for future improvements and easier debugging

### Benefits for AzerothCore

- Better performance on high-population servers
- Reduced lag in major cities or crowded quest zones
- Scales better for events like world PvP or Wintergrasp
- Lays groundwork for more advanced systems (e.g. phasing, regional load balancing)

I don't know if it's possible to implement a system similat to this, but this could be a significant architectural upgrade and help AzerothCore scale further as adoption grows. I’d love to hear your thoughts on feasibility and whether a similar system could be integrated.

Thanks for your time.

_No response_

### Additional context

_No response_

grunkerZ:
## Changes Proposed:
This PR proposes changes to:
-  [ X ] Core (units, players, creatures, game systems).

The PR introduces a multithreaded map architecture that divides larger maps into logical partitions (currently 4 partitions per map) which can be processed in parallel using a Fork-Join model. This will reduce single-thread CPU bottlenecks in highly populated areas.
- Implemented LockedQueue<WorldObject*> for transferring entities across partition boundaries.
- Made DynamicTreeMap thread safe with std::shared_mutex for concurrent checks while protecting transport updates
- Refactored map object update lists to safely transition between partitions
- Resolves BountyHub Bounty:  https://www.bountyhub.dev/en/bounty/view/ae3d1ab6-8f15-4d72-b39c-a7153339fa44/feature-map-partitioning-support-for-azerothcore

### AI-assisted Pull Requests

> [!IMPORTANT]

- [ X ] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.

>Google Antigravity CLI was used in identifying race conditions and debugging call stacks
>All code was written by a human

## Issues Addressed:

- Closes #22571

## SOURCE:

- [X ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). 
> The architectural approach was inspired by Project-Epoch TrinityCore PR #120, and adapted to fit AzerothCore's map update sequence and collision trees.

## Tests Performed:

This PR has been:
- [ X ] Tested in-game by the author.

> Tested cross partition movement with a boundary at X=0, Y=0 with both Player and non player map objects, cross partition spell casting, teleporting across maps and partitions, and instance map generations. Benchmarks showed a large reduction in main thread diff time under heavy spawn loads since worker threads now take over partition processing.

## How to Test the Changes:

- [ X ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

Map Partitions divide the map into 4 quadrants intersecting at X=0 and Y=0. Testing the partition hand offs can be done as following:
1. Teleport slightly inside of a partition near a boundary (e.g. x=5, y=5, .gm fly may need to be on to avoid dying from fall damage)
2. Spawn or find an aggressive monster and have it aggro to you (I used the naturally spawning Moss Creepers on Map 0)
3. Run across the partition boundary (X=0 or Y=0)
4. Confirm that the monster seamlessly continues pathfinding as it transfers partitions without any server stutter and continues to pathfind/attack.
5. Additionally spawn a monster on one side of a partition boundary (e.g. x=5, y=5) and then stand across the partition boundary in another partition (e.g. x=5 y=-5).
6. Target the monster with a spell and attack it to confirm cross-partition spell casting and targeting
7. Additionally spawn hundreds of NPCs in a partition (such as Booty Bay in map 0), and then teleport to another partition and check the server info to confirm that the update time diff remains low, since multiple CPU cores now process the grids in parallel.

- [ ]
- [ ]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability for map updates, object movement, and collision queries under heavy concurrent load.
  * Reduced the chance of crashes or corrupted world state during grid loading, unloading, and relocation.
  * Improved pathfinding reliability by making navigation data access more resilient during updates.
  * Updated delayed visibility handling to avoid timing-related issues during object updates.

### DIFF
--- a/src/common/Collision/DynamicTree.cpp
+++ b/src/common/Collision/DynamicTree.cpp
@@ -30,6 +30,7 @@
 #include <G3D/AABox.h>
 #include <G3D/Ray.h>
 #include <G3D/Vector3.h>
+#include <mutex>
 
 using VMAP::ModelInstance;
 
--- a/src/common/Collision/DynamicTree.cpp
+++ b/src/common/Collision/DynamicTree.cpp
@@ -116,31 +116,40 @@ DynamicMapTree::~DynamicMapTree()
 
 void DynamicMapTree::insert(const GameObjectModel& mdl)
 {
+    // Exclusive lock because Transports can move concurrently across map partitions
+    std::unique_lock<std::shared_mutex> lock(_lock);
     impl->insert(mdl);
 }
 
 void DynamicMapTree::remove(const GameObjectModel& mdl)
 {
+    // Exclusive lock because Transports can move concurrently across map partitions
+    std::unique_lock<std::shared_mutex> lock(_lock);
     impl->remove(mdl);
 }
 
 bool DynamicMapTree::contains(const GameObjectModel& mdl) const
 {
+    std::shared_lock<std::shared_mutex> lock(_lock);
     return impl->contains(mdl);
 }
 
 void DynamicMapTree::balance()
 {
+    // Exclusive lock to rebalance collision tree safely
+    std::unique_lock<std::shared_mutex> lock(_lock);
     impl->balance();
 }
 
 int DynamicMapTree::size() const
 {
+    std::shared_lock<std::shared_mutex> lock(_lock);
     return impl->size();
 }
 
 void DynamicMapTree::update(uint32 t_diff)
 {
+    std::unique_lock<std::shared_mutex> lock(_lock);
     impl->update(t_diff);
 }
 
@@ -198,6 +207,8 @@ private:
 
 bool DynamicMapTree::GetIntersectionTime(const uint32 phasemask, const G3D::Ray& ray, const G3D::Vector3& endPos, float& maxDist) const
 {
+    // Shared lock to allow concurrent reads across map partitions
+    std::shared_lock<std::shared_mutex> lock(_lock);
     float distance = maxDist;
     DynamicTreeIntersectionCallback callback(phasemask, VMAP::ModelIgnoreFlags::Nothing);
     impl->intersectRay(ray, callback, distance, endPos, false);
@@ -212,6 +223,8 @@ bool DynamicMapTree::GetObjectHitPos(const uint32 phasemask, const G3D::Vector3&
                                      const G3D::Vector3& endPos, G3D::Vector3& resultHit,
                                      float modifyDist) const
 {
+    // no Map Partition lock here as shared_mutex is not recursive
+    // this will call GetIntersectionTime internally which will get the lock for us
     bool result = false;
     float maxDist = (endPos - startPos).magnitude();
     // valid map coords should *never ever* produce float overflow, but this would produce NaNs too
@@ -256,6 +269,8 @@ bool DynamicMapTree::GetObjectHitPos(const uint32 phasemask, const G3D::Vector3&
 
 bool DynamicMapTree::isInLineOfSight(float x1, float y1, float z1, float x2, float y2, float z2, uint32 phasemask, VMAP::ModelIgnoreFlags ignoreFlags) const
 {
+    // Shared lock to allow concurrent reads across map partitions
+    std::shared_lock<std::shared_mutex> lock(_lock);
     G3D::Vector3 v1(x1, y1, z1), v2(x2, y2, z2);
 
     float maxDist = (v2 - v1).magnitude();
@@ -274,6 +289,8 @@ bool DynamicMapTree::isInLineOfSight(float x1, float y1, float z1, float x2, flo
 
 float DynamicMapTree::getHeight(float x, float y, float z, float maxSearchDist, uint32 phasemask) const
 {
+    // Shared lock to allow concurrent reads across map partitions
+    std::shared_lock<std::shared_mutex> lock(_lock);
     G3D::Vector3 v(x, y, z);
     G3D::Ray r(v, G3D::Vector3(0, 0, -1));
     DynamicTreeIntersectionCallback callback(phasemask, VMAP::ModelIgnoreFlags::Nothing);
@@ -291,6 +308,8 @@ float DynamicMapTree::getHeight(float x, float y, float z, float maxSearchDist, 
 
 bool DynamicMapTree::GetAreaAndLiquidData(float x, float y, float z, uint32 phasemask, Optional<uint8> reqLiquidType, VMAP::AreaAndLiquidData& data) const
 {
+    // Shared lock to allow concurrent reads across map partitions
+    std::shared_lock<std::shared_mutex> lock(_lock);
     G3D::Vector3 v(x, y, z + 0.5f);
     DynamicTreeLocationInfoCallback intersectionCallBack(phasemask);
     impl->intersectPoint(v, intersectionCallBack);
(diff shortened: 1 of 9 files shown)


Kitzunu:
So how much does this tank resources and have you tested this at any large scale

grunkerZ:
> So how much does this tank resources and have you tested this at any large scale

@Kitzunu The architecture that I chose for the Map Partitioning actually has very low processing overhead and memory consumption. The overhead of launching parallel partition updates using std::async is negligible compared to the previous continent update loops. It only dispatches as many tasks as there are partitions, rather than something like a thread per entity. The memory consumption of this is basically identical to the single threaded map architecture (maybe a tiny bit more, but nothing noticeable). I haven't created any pseudo maps or duplicated any navmeshes, only grouping the grids logically. The biggest risk this has for tanking any resources is thread contention. I fixed  ##this issue by making objects operate independently during the tick. Locking only occurs at the partition boundary using _transferQueue, which is only hit when a object physically passes the partition boundary. Since almost all of the movement in a tick is within a partition, the risk of contention is near 0. I also guarded the DynamicMapTree with std::shared_mutex to prevent this issue, meaning that common operations such as spatial queries remain concurrent and unlocked for readers.

Testing wise, I made sure of the lock stability and ensured there were no deadlocks or dropped objects during any of the cross boundary hand offs by running some stress tests with huge clusters of creatures pathing across partition boundaries simultaneously. I kept an eye on the CPU utilization and server info and they all looked much better than before. My computers CPU utilization scaled very well with the tests, and there was no mass drop in server TPS across the continent's partitions during the test. Even in the partition with all the entities, the server TPS was much better than my baseline tests with no modifications (My baseline test had the highest tick be well over 100ms, maybe 140ms, and my stress test had the highest tick take 50ms).

Overall:
Memory and CPU - Overhead is practically zero. All I've done is just parallelize the update loops.
No Lock Bottlenecks - Almost all of the movement are not crossing partition boundaries and are lock free. Thread locks only occur when entities cross a partition boundary.
Tested at Scale - I stress tested with a huge cluster of creatures pathing across partition boundaries. CPU utilization scaled well and peak tick times dropped from 100+ms (baseline test) to ~50ms.

Kitzunu:
See if you think any of this is valid :)

- **Conflicts with the existing `MapUpdater` pool.** `Map::Update` already runs on a worker-thread pool (`MapMgr::m_updater.schedule_update`, sized by `MapUpdate.Threads`). Adding `std::async` partitions *inside* each map nests a second threading layer, which means thread explosion and heavy contention on global singletons (`ObjectAccessor`, `sObjectMgr`, grid stores). A viable version needs to build on `MapUpdater`, not around it.

- **`obj->Update()` races on unprotected shared state.** WoW updates are cross-object and cross-partition: a creature in one quadrant attacks/casts on a target in another, mutating its auras, `ThreatMgr`, `HostileRefMgr`, health/power, summons, packet queues, etc. The added locks cover ~5 structures; `Unit`/spell/threat/grid state is otherwise unsynchronized. Players are also updated on the main thread while worker threads mutate them. This will race under exactly the crowded load the PR targets.

- **Static quadrants don't load-balance.** The hot case is crowding in one spot (raid boss, Dalaran, Orgrimmar) — that all lands in a single partition (one thread busy, three idle), while every small instance/BG/arena still spawns 4 threads per tick for a handful of objects. Likely a net regression where it matters most.

- **`std::async(launch::async)` per partition per tick** spawns fresh threads (not pooled) and joins via `future.get()`; creation/teardown outweighs the work. It also defeats the `thread_local` navmesh cache, since those threads don't persist. And `_gridLock` held across `PathGenerator::CalculatePath` serializes all pathfinding across partitions — removing the main benefit.

- **Worker-vs-worker vector race:** when one partition's `obj->Update()` despawns/removes an object in another partition, `RemoveObjectFromMapUpdateList` loops all partitions and mutates another running partition's `_updatableObjects` with no lock.
- **Possible infinite loop:** in `MapPartition::Update`, a null slot → `RemoveObject(nullptr)` finds nothing, `continue` re-reads the same index → spin.
- **Dead code:** `AddObjectToPendingUpdateList` no longer sets `PendingAdd`, so `_pendingAddUpdatableObjectList` / `_updatableObjectList` / `_AddObjectToUpdateList` are now unreachable — the update path is half-migrated.
- Removing `EnsureGridLoaded(new_cell)` on grid-crossing relocation changes load ordering and needs justification/testing.

**Minor:** raw `new`/`delete` for partitions (prefer `unique_ptr`); reuse project threading infra instead of `std::async`/`std::mutex`; several comment typos.

**Suggestion:** this needs to integrate with `MapUpdater`, define a real thread-safety contract for `Unit`/spell/threat state (or defer cross-partition interactions), drop per-tick `std::async`, and be validated under ThreadSanitizer with automated stress tests. Happy to help sketch a `MapUpdater`-integrated approach if useful.

grunkerZ:
@Kitzunu Thank you for the review. I do think you are right on your points. I was trying to keep my modifications as an isolated layer on top of the existing map structure to minimize the footprint of the PR, and avoid rewriting the core threading model or heavily refactoring the MapUpdater. I feel a bit silly for missing some of the more obvious points you brought up though. 

I treated the partition updates as localized tasks, but I didn't consider that nesting std::async inside the already threaded map updater would cause thread explosion. The creation and teardown does vastly outweigh the work it does and wiping the thread_local navmesh defeats the purpose. I overlooked the deep cross object combat states, and I assume my tests showed improvements because the entities were somewhat evenly distributed, which would hide the load balancing issue, and since they weren't fighting with each other, they didn't cause any issues with the combat states that you mentioned. Thank you for finding the continue infinite loop and the unreachable pending update list. As for removing EnsureGridLoaded, I had assumed the grid would be naturally paged in by the player's presence rather than needing explicit ensuring on relocation, but I see now how that breaks the load ordering.

I would absolutely love to take you up on your offer to help sketch a MapUpdater-integrated approach, as having someone with much deeper knowledge of this codebase than me will be a huge help architecturally. I do want to get this right and build something that actually benefits the core architecture. How would you recommend I approach the cross object state safely without locking up the whole Map?

Kitzunu:
Superseded by https://github.com/azerothcore/azerothcore-wotlk/pull/16832

Closing as map partitioning is nothing we will pursue on AC. If anything we will take the path of clustering which Walkline is activley working on.

PR #25647, opened 2026-04-29
fix(Core/Spells): Fire CAST proc before HIT/FINISH for instant spells

blinkysc:
## Changes Proposed:
This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).

For instant spells, HIT and FINISH happen inside `handle_immediate()`, so CAST procs were firing last - the actual order was `HIT -> FINISH -> CAST`. An aura applied during HIT (e.g. Arcane Potency from a Clearcasting that procs mid-AoE on Arcane Explosion) was then consumed by the same cast's trailing CAST proc, never benefiting the player.

This moves the CAST proc block to before the immediate/delayed branch in `Spell::_cast()` so the spec-correct order `CAST -> HIT -> FINISH` is restored. `m_appliedMods` is already populated by `HandleLaunchPhase` (via `ApplyModToSpell` and `SpellDoneCritChance`), so `PROC_ATTR_REQ_SPELLMOD`-based consumption (Hot Streak, Brain Freeze, Arcane Potency, etc.) still drops charges correctly.

Missile Barrage (44401) is the one outlier: its three modifiers (`SPELLMOD_DURATION` effect 0, `SPELLMOD_ACTIVATION_TIME` effect 1, `SPELLMOD_COST` effect 2) are all applied inside `handle_immediate` (channel start and periodic-aura setup), so consumption on CAST would fail. The accompanying SQL update moves its `SpellPhaseMask` from CAST (1) to FINISH (4) so the charge drops once `m_appliedMods` is fully populated.

I audited every other `spell_proc` row with `SpellPhaseMask = CAST` and `AttributesMask & PROC_ATTR_REQ_SPELLMOD` (12043, 12536, 16166, 16246, 16870, 17116, 17941, 18708, 34936, 47383, 48108, 53817, 54274/76/77, 57529/31, 57761, 71162/65) - all use `SPELLMOD_CASTING_TIME`, `SPELLMOD_COST`, `SPELLMOD_GLOBAL_COOLDOWN`, `SPELLMOD_DAMAGE`, or `SPELLMOD_CRITICAL_CHANCE`, all of which are applied either in `Spell::prepare()` or in `HandleLaunchPhase` (i.e. before the new CAST proc location). Missile Barrage is the only entry that needed a phase change.

### AI-assisted Pull Requests

> [!IMPORTANT]
[...]

PR #26535, opened 2026-07-09
fix(Core/Spells): clear stale spellmod tracking spell

Yehonal:
## Changes Proposed

This PR proposes changes to:

Clear `Player::m_spellModTakingSpell` when the tracked `Spell` finishes, with a destructor fallback if the pointer still references the spell being destroyed.

## Crash Symptom

A crash report showed a player cast of spell `53595` crashing while calculating power cost:

`WorldSession::HandleCastSpellOpcode -> Spell::prepare -> SpellInfo::CalcPowerCost -> Player::ApplySpellMod<int> -> Player::ApplyModToSpell -> std::set<Aura*>::insert`

The failing `std::set` operation was on `spell->m_appliedMods`.

## Root Cause

`Player::ApplySpellMod` redirects charge tracking to `m_spellModTakingSpell` when it is set. If that pointer outlives the `Spell` it points to, later spell modifier calculations can insert into a stale `Spell::m_appliedMods` set.

The crash evidence matches that pattern: the spell modifier logic was using a `Spell*` whose `m_appliedMods` storage no longer looked like a valid `Spell` set.

## Fix Summary

- Clear `m_spellModTakingSpell` in `Spell::finish()` when the finishing spell is the tracked spell.
- Add a matching destructor fallback to clear the pointer if a spell is destroyed while still tracked.
- Leave normal spell modifier charge application/removal behavior unchanged.

## Validation

- `git diff --check`
- `cmake --build var/build/obj --target game -j2`

I also attempted `cmake --build var/build/obj --target worldserver -j2`; it built the core `game` target but failed in modules because `mod-eluna` could not find `lua.h`, which appears to be an environment/module dependency issue unrelated to this patch.

## Notes

Related but separate from #25647, which changes spell proc phase ordering and touches `m_appliedMods` behavior but does not clear stale `m_spellModTakingSpell` pointers.


sudlud (approved):

sudlud:
Reviewed against current master. Correct and safe.

Both hunks only clear when the tracked pointer equals `this`, so they can't wipe another spell's tracking. No normal-path change: `cast()` already clears at `Spell.cpp:4047` before `finish()`, and charge drop/restore work on `this->m_appliedMods` directly, so this is a pure safety net. Destructor deref of `m_caster` matches `~SpellEvent` and the null guard covers the `UpdatePointers()`-failed case.

Non-blocking: `cast()` saves the previous tracking spell into local `lastMod` and restores it after `_cast()` (`Spell.cpp:3763-3775`) — if that spell dies during `_cast()`, the restore re-installs a dangling pointer neither hunk covers. Low risk since it's up the call stack.

PkllonG:
The crash happens when shutting down the server, randomly.

Nyeriah:
This introduced a crash on unit-summoned casters. The new `GetSpellModOwner()` call in `Spell::~Spell` resolves a creature-owned caster's owner through `ObjectAccessor::GetCreature`, which calls `GetMap()` on the caster. When the spell is destroyed during the caster's own deletion (`Map::RemoveAllObjectsInRemoveList` -> `DeleteFromWorld` -> `~WorldObject` -> `EventProcessor::KillAllEvents` -> `~SpellEvent`), the caster has already been removed from the map and `ASSERT(m_currMap)` fires.

Caught it in production with a Scorching Totem (entry 9637, creature-summoned) despawning with a pending spell event:

```
#1  WorldObject::GetMap () at Object.h:630 (ASSERT m_currMap)
#2  ObjectAccessor::GetCreature
#3  Unit::GetOwner
#4  Unit::GetSpellModOwner
#5  Spell::~Spell () at Spell.cpp:698
#8  EventProcessor::KillAllEvents
#10 WorldObject::~WorldObject
#12 Totem::~Totem
#15 Map::RemoveAllObjectsInRemoveList
```

Player casters are fine since `GetSpellModOwner` returns the player without a map lookup. Fix in #26780.

PR #26439, opened 2026-07-03
fix(DB/GameObject): correct Higher Learning pooled respawn cycle

Gazeshooter:
Fix the Higher Learning book pools so a book despawns after being read and its location stays empty until the respawn timer expires, instead of the same object lingering until server restart.

## Changes Proposed:
This PR proposes changes to:
- [x] Core (units, players, creatures, game systems).
- [x] Database (SAI, creatures, etc).

The 35 Higher Learning books and decoys are pooled, one active per location across 8 pools. Their timed SmartAI event only set the loot state to `GO_JUST_DEACTIVATED` 240 seconds after a book was read, which never actually despawned it, so a location stayed occupied by the same object until restart.

Two changes:
1. **SQL**: the timed event (`SMART_EVENT_TIMED_EVENT_TRIGGERED`) now uses `SMART_ACTION_FORCE_DESPAWN` instead of setting the loot state, so the book despawns 240 seconds after being read.
2. **Core**: `GameObject::DespawnOrUnsummon` now only re-rolls the pool immediately when no respawn timer is set. When a respawn timer is set, the current pool member stays reserved and `GameObject::Update` re-rolls the pool once the timer expires. Without this, a force-despawned pooled object was re-rolled instantly and a new book appeared immediately instead of after the respawn time.

The 240-second linger and the 10800-second respawn time are unchanged. Root cause was diagnosed by @sudlud in the linked issue.

## Issues Addressed:
- Closes #22632

## SOURCE:
Validated against AzerothCore gameobject, SmartAI and pool behaviour, and through testing on a local AzerothCore server.

### AI-assisted Pull Requests
- [x] AI tools (ChatGPT, GPT-5.5 Thinking) assisted with source analysis, test planning, and drafting the description. I manually implemented the change, reproduced the issue, performed in-game testing, and verified the resulting behaviour.

## Tests Performed:
- [x] Tested in-game by the author.

### Unpatched behaviour
- Interacted with Higher Learning decoy entry `192872`.
- Confirmed the book remained spawned after the existing 240-second SmartAI event, with unchanged entry and GUID.

### Patched behaviour
- Applied the SAI change.
- Interacted with a pooled Higher Learning book, confirmed the page displayed and the book stayed readable during the linger period, then despawned after roughly 240 seconds and did not respawn during the normal 10800-second window.

### Accelerated pool testing
For testing only, all four members of pool `5694` were temporarily set to a 30-second respawn time.
[...]

PR #26546, opened 2026-07-10
fix(DB/Gameobject): correct Higher Learning book respawn

anonwow175:
## Changes Proposed
- Mark all Dalaran Higher Learning book pool candidates as consumable goobers (`Data5 = 1`).
- This lets the existing SmartAI read timer's `GO_JUST_DEACTIVATED` action schedule the normal `spawntimesecs` respawn instead of immediately resetting the active book.
- Covers both the achievement books and decoy books in pools 5691-5698, so any read book can despawn and later reroll its pool.

## Source reasoning
- The book SmartAI already starts a 240000 ms timed event on gossip/use and then sets loot state 3.
- In core, goober `Data5` maps to `goober.consumable`, which drives `IsDespawnAtAction()`.
- Without `Data5 = 1`, `GO_JUST_DEACTIVATED` returns the object to `GO_READY`; with it, the object uses its configured 10800 second respawn timer.

## Tests Performed
- Verified by source/SQL inspection.
- Verified branch diff against upstream `master` contains only `data/sql/updates/db_world/2026_07_10_00.sql`.
- No local worldserver runtime test performed.

### DIFF
--- a/data/sql/updates/db_world/2026_07_10_00.sql
+++ b/data/sql/updates/db_world/2026_07_10_00.sql
@@ -1,0 +1,12 @@
+-- DB update 2026_07_06_00 -> 2026_07_10_00
+--
+-- Higher Learning Dalaran books should despawn after the SmartAI read timer and then use their gameobject respawn timer.
+UPDATE `gameobject_template`
+SET `Data5` = 1
+WHERE `entry` IN (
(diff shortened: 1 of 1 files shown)


heyitsbench:
Same thing as https://github.com/azerothcore/azerothcore-wotlk/pull/26439, this fix is incorrect.

Kitzunu:
Closing as these are sniffed fields and can't be updated.
Needs to be fixed via script/SAI

PR #23251, opened 2025-10-15
feature(SmartAI/Movement) - Unify waypoint systems

killerwife:
Unifies waypoints and waypoint_data so its interchangably used

Renames

SMART_EVENT_WAYPOINT_START --> SMART_EVENT_ESCORT_START
SMART_EVENT_WAYPOINT_REACHED --> SMART_EVENT_ESCORT_REACHED
SMART_EVENT_WAYPOINT_PAUSED --> SMART_EVENT_ESCORT_PAUSED 
SMART_EVENT_WAYPOINT_RESUMED --> SMART_EVENT_ESCORT_RESUMED 
SMART_EVENT_WAYPOINT_STOPPED --> SMART_EVENT_ESCORT_STOPPED 
SMART_EVENT_WAYPOINT_ENDED --> SMART_EVENT_ESCORT_ENDED
SMART_EVENT_WAYPOINT_DATA_REACHED --> SMART_EVENT_WAYPOINT_REACHED
SMART_EVENT_WAYPOINT_DATA_ENDED --> SMART_EVENT_WAYPOINT_ENDED

Renames 

SMART_ACTION_WP_START --> SMART_ACTION_ESCORT_START                       
SMART_ACTION_WP_PAUSE --> SMART_ACTION_ESCORT_PAUSE                       
SMART_ACTION_WP_STOP --> SMART_ACTION_ESCORT_STOP                        
SMART_ACTION_WP_RESUME --> SMART_ACTION_ESCORT_RESUME                      

To properly reflect escort behaviour

Renamed 

SMART_ACTION_WAYPOINT_DATA_START --> SMART_ACTION_WAYPOINT_START

to reflect it now supports all types of waypoints with a newly added parameter:

param3 = 'pathSource': If it is setted on 0 the action will take the path from waypoint_data table. If it is on 1 the action will take the path from waypoints table.

Also refactored some scripts that basically emulated this functionality.

Renamed

MovePath to MoveWaypoint per official naming (uses pathfinding)

MoveSplinePath to MovePath (doesnt use pathfinding)

PR #26609, opened 2026-07-14
fix(core/SAI): remove duplicate WAYPOINT_REACHED dispatch in MovementInform

Voxstrasza:
#23251 reworked SAI WP/movement system and added dedicated Waypoint_Reached hook but did not remove the old dispatch inside of Movement_Inform. This removes the redundancy and NPCs who talk on waypoint_reached won't double fire their text lines.

## Changes Proposed:
This PR proposes changes to:

### AI-assisted Pull Requests

## Tests Performed:
This PR has been:

I was working on #26608 - can use it to test: .go c 125712

### DIFF
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -1317,13 +1317,11 @@ void SmartAI::WaypointReached(uint32 nodeId, uint32 pathId)
     }
 }
 
-void SmartAI::WaypointPathEnded(uint32 nodeId, uint32 pathId)
+void SmartAI::WaypointPathEnded(uint32 /*nodeId*/, uint32 /*pathId*/)
 {
-    if (!HasEscortState(SMART_ESCORT_ESCORTING))
-    {
-        GetScript()->ProcessEventsFor(SMART_EVENT_WAYPOINT_ENDED, nullptr, nodeId, pathId);
-        return;
-    }
+    // WAYPOINT_ENDED is fired from SmartAI::PathEndReached (the waypoint generator calls both
+    // hooks at path end); firing it here too made the event trigger twice. PathEndReached also
+    // runs first, while me->GetWaypointPath() is still valid for the event's pathId filter.
 }
 
 void SmartAI::DistancingEnded()
--- a/src/server/game/AI/SmartScripts/SmartAI.cpp
+++ b/src/server/game/AI/SmartScripts/SmartAI.cpp
@@ -687,8 +687,7 @@ void SmartAI::MovementInform(uint32 MovementType, uint32 Data)
     if (MovementType == POINT_MOTION_TYPE && Data == SMART_ESCORT_LAST_OOC_POINT)
         me->ClearUnitState(UNIT_STATE_EVADE);
 
-    // WAYPOINT_REACHED is fired from SmartAI::WaypointReached (the waypoint generator calls it
-    // alongside this and passes the pathId); firing it here too made the event trigger twice.
+    // WAYPOINT_REACHED is fired from WaypointReached() instead, to avoid a double trigger
     GetScript()->ProcessEventsFor(SMART_EVENT_MOVEMENTINFORM, nullptr, MovementType, Data);
     if (!HasEscortState(SMART_ESCORT_ESCORTING))
         return;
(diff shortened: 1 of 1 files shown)


Voxstrasza:
@killerwife

killerwife:
As long as this was tested against both waypoint tables, (i dont remember both names off the top of my head rn) this is fine.

I am just wondering why you are exposing two functions to Movement generators, which should view the AI API as simple as possible, and moving the logic there, instead of doing it in MovementInform, which was designed as said AI API so that movement generators have as little AI logic as possible. But thats a question for the already merged code, not this one, so what is done is done in my book.

Voxstrasza:
> As long as this was tested against both waypoint tables, (i dont remember both names off the top of my head rn) this is fine.
> 
> 
> 
> I am just wondering why you are exposing two functions to Movement generators, which should view the AI API as simple as possible, and moving the logic there, instead of doing it in MovementInform, which was designed as said AI API so that movement generators have as little AI logic as possible. But thats a question for the already merged code, not this one, so what is done is done in my book.

could you explain what you mean? if you're referring to my Lord Thorval script, it wasn't merged when you made this comment

killerwife:
```
// WAYPOINT_REACHED is fired from SmartAI::WaypointReached (the waypoint generator calls it
// alongside this and passes the pathId); firing it here too made the event trigger twice.
```

Waypoint reached is simply a special case of MovementInform where the movegen is the waypoint movegen and the AI state is escorting.

MovementInform and WaypointReached are currently functions of the external AI API (inherited public function in this case)

The logic for determining if an npc is escorting with a waypoint movement generator has to be obviously somewhere.

My opinion is that Movement generators should not care if you are escorting, and as such the logic junction should be in AI function (i.e. inside movementinform) rather than in a movement generator.

This issue stops being theoretical at the moment where you need a second movement generator (lets say catmullrom flying spline) to work with escort.

Voxstrasza:
you nudged me in the right direction. at present, it does execute twice. but I noticed this:

> void SmartAI::WaypointPathEnded(uint32 nodeId, uint32 pathId)
> {
>     if (!HasEscortState(SMART_ESCORT_ESCORTING))
>     {
>         GetScript()->ProcessEventsFor(SMART_EVENT_WAYPOINT_ENDED, nullptr, nodeId, pathId);
>         return;
>     }

a potential fix might be 

> void SmartAI::WaypointPathEnded(uint32 /*nodeId*/, uint32 /*pathId*/)

i'm hoping someone smarter than me comes along soon

Voxstrasza:
@Kitzunu help

src/server/game/AI/SmartScripts/SmartAI.cpp
```diff
@@ -1318,13 +1316,9 @@ void SmartAI::WaypointReached(uint32 nodeId, uint32 pathId)
...
 
-void SmartAI::WaypointPathEnded(uint32 nodeId, uint32 pathId)
+void SmartAI::WaypointPathEnded(uint32 /*nodeId*/, uint32 /*pathId*/)
 {
-    if (!HasEscortState(SMART_ESCORT_ESCORTING))
-    {
-        GetScript()->ProcessEventsFor(SMART_EVENT_WAYPOINT_ENDED, nullptr, nodeId, pathId);
-        return;
-    }
+    // // WAYPOINT_ENDED is fired from PathEndReached() instead, to avoid a double trigger
```
Kitzunu, line 1321: You are aware that PathEndReached() does not pass nodeId, but instead hardcodes it to 0. I am worried that it could potentially cause some issue.

Voxstrasza: I will do some more testing. 

I tested my PR, and it stops double texts. I can only think of testing out escort quests and more text on waypoints. 

if I can't figure it out I'll open an issue, hopefully someone smarter than me can figure it out

Kitzunu: SMART_EVENT_WAYPOINT_ENDED and SMART_EVENT_WAYPOINT_REACHED provides the ability to specify PointId and PathId. If we make sure nothing breaks with those events. Then I think this is good

Voxstrasza:
@Kitzunu I am marking this WIP/needing to be tested. as far as I'm aware, need to test to ensure this doesn't break escort events. I'll post back with a list of escort quests, a list of NPCs that do stuff on waypoints, etc.

so far, this PR **does** fix the double text from the Lord Thorval event I implemented in #26608 - so that is good

Issue #22334, opened 2025-06-19
List of unimplemented Spell Attributes


Tereneckla:
Roughly sorted by their comments.

- [ ] SPELL_ATTR0_HEARTBEAT_RESIST
- [ ] SPELL_ATTR1_PREVENTS_ANIM
- [ ] SPELL_ATTR1_COMBO_ON_BLOCK
- [ ] SPELL_ATTR2_NO_ACTIVE_PETS
- [ ] SPELL_ATTR2_NO_JUMP_WHILE_CAST_PENDING
- [ ] SPELL_ATTR2_INITIATE_COMBAT_POST_CAST
- [ ] SPELL_ATTR2_PROC_COOLDOWN_ON_FAILURE
- [ ] SPELL_ATTR3_HIDE_IN_RAID_FILTER
- [ ] SPELL_ATTR4_NO_HARMFUL_THREAT
- [ ] SPELL_ATTR4_FORCE_DISPLAY_CASTBAR
- [ ] SPELL_ATTR5_NOT_AVALIABLE_WHILE_CHARMED
- [x] SPELL_ATTR5_NOT_ON_PLAYER #22332 
- [x] SPELL_ATTR5_NOT_ON_PLAYER_CONTROLLED_NPC #22332
- [ ] SPELL_ATTR6_NO_COOLDOWN_ON_TOOLTIP
- [ ] SPELL_ATTR7_HORDE_SPECIFIC_SPELL
- [ ] SPELL_ATTR7_ALLIANCE_SPECIFIC_SPELL
- [ ] SPELL_ATTR7_REPORT_SPELL_FAILURE_TO_UNIT_TARGET

unknown
- [ ] SPELL_ATTR0_PROC_FAILURE_BURNS_CHARGE
- [ ] SPELL_ATTR0_WEARER_CASTS_PROC_TRIGGER
- [ ] SPELL_ATTR1_NO_SKILL_INCREASE
- [ ] SPELL_ATTR1_THREAT_ONLY_ON_MISS
- [ ] SPELL_ATTR1_IGNORE_OWNERS_DEATH
- [ ] SPELL_ATTR1_REQUIRE_ALL_TARGETS
- [ ] SPELL_ATTR1_DISCOUNT_POWER_ON_MISS
- [ ] SPELL_ATTR1_CAST_WHEN_LEARNED
- [ ] SPELL_ATTR2_NO_SHAPESHIFT_UI
[...]

PR #26616, opened 2026-07-14
fix(Core/Spell): Implement SPELL_ATTR5_NOT_AVALIABLE_WHILE_CHARMED

Kitzunu:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
> 

## Issues Addressed:

- part of https://github.com/azerothcore/azerothcore-wotlk/issues/22334

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


sogladev (dismissed):
Is this attr fully implemented like this?

There's only a few spells that have this attribute.

This may be used for mind control, so players can't cast this spell by MCing a handler:

https://www.wowhead.com/wotlk/spell=59694/unholy-rage

`creature_template_spell` is empty for this creature so it doesn't show up as of now anyway

Kitzunu:
So I havent done extensive research more than that it felt self explanatory through the attribute name.
But you're saying it could be the same restrictions for both MC and charm?

sogladev:
> So I havent done extensive research more than that it felt self explanatory through the attribute name. But you're saying it could be the same restrictions for both MC and charm?

Yes, I think it should be both

Kitzunu:
Do you know if it would be IsControlledByPlayer() or isPossessedByPlayer()? Otherwise I will try to figure it out

sogladev:
`IsCharmed()` to check if the target is charmed. For attr checks, I think it should be somewhere in `CharmInfo::InitCharmCreateSpells()` that sets the pet action bar spells.

Kitzunu:
Ahh we dont want it to be able to create the spell for the actionbar at all. Got you

Kitzunu:
I might look at it tomorrow. Im way to tired today

sudlud:
Two things on top of what's above.

`creature_template_spell` isn't empty for it: 24082 has no rows, but its difficulty_entry_1 31675 does, incl. (31675,2,59694). Repro: MC a Proto-Drake Handler in heroic UK, Unholy Rage sits on the pet bar. 13 spells carry the attribute, none passive, and two more are reachable via Enslave Demon: 33804 on 21249 and 38607 on 19740.

Worth keeping the CheckCast check alongside the CharmInfo filter. HandlePetCastSpellOpcode only validates Creature::HasSpell (m_spells), not the action bar, so a crafted CMSG_PET_CAST_SPELL still gets through if we only filter the bar.

sudlud:
the three hunks here are byte-identical to TrinityCore 076c63c766, but that commit had a fourth one in `Player::VehicleSpellInitialize` filtering the vehicle bar the same way. maybe we can pull that in too? no vehicle carries one of the 13 attr spells today, so the button would just show and fail on click, but it's the same hole.

heads up if you do: our loop writes 4 bytes per slot inline, so the skip branch needs `data << uint16(0) << uint8(0) << uint8(i + 8);` before the `continue`, like the two branches above it. TC's bare `continue` would desync the packet.

could we also add a follow-up commit crediting Shauren for the cherry-pick?

```
```

PR #26634, opened 2026-07-15
Revert "fix(Core/Spells): Fire CAST proc before HIT/FINISH for instant spells"

blinkysc:
Reverts azerothcore/azerothcore-wotlk#25647

### DIFF
--- a/data/sql/updates/pending_db_world/revert.sql
+++ b/data/sql/updates/pending_db_world/revert.sql
@@ -1,0 +1,1 @@
+UPDATE `spell_proc` SET `SpellPhaseMask` = 1 WHERE `SpellId` = 44401;
--- a/src/server/game/Spells/Spell.cpp
+++ b/src/server/game/Spells/Spell.cpp
@@ -3951,44 +3951,6 @@ void Spell::_cast(bool skipCheck)
         }
     }
 
-    // CAST -> HIT -> FINISH ordering: fire CAST before handle_immediate so an aura
-    // applied during HIT (e.g. Arcane Potency from Clearcasting) isn't consumed by
-    // the same cast. Triggered spells skip this so periodic ticks (Blizzard etc.)
-    // don't burn cast-charge buffs.
-    if (m_originalCaster && !IsTriggered())
-    {
-        uint32 procAttacker = m_procAttacker;
-        if (!procAttacker)
-        {
-            bool IsPositive = m_spellInfo->IsPositive();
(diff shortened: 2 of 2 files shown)


Nyeriah:
Reverting #25647 most likely won't stop the crashes this is aimed at — the evidence points at a different root cause.

**Crash signature** (15 dumps, one server, all identical modulo where the stale pointer finally derefs):
```
SIGSEGV in the aura proc path:
Unit::GetProcAurasTriggeredOnEvent → Aura::GetProcEffectMask → Aura::GetId
always the same proc target: creature entry 18587 (Frayer), map 553 (The Botanica)
```

**Why the #25647 revert doesn't address it.** #25647 only changes *when* the CAST-phase proc fires relative to `handle_immediate`. The faulting frame is the HIT-phase target proc, and the actual fault is iterator/pointer invalidation during the live-map walk in `Unit::GetProcAurasTriggeredOnEvent` (the `procAuras == nullptr` branch that iterates `GetAppliedAuras()` directly). #25647 doesn't touch that loop or the container, so reverting it leaves the fault in place.

**Root cause.** `spell_botanica_shift_form_aura` (Frayer's spell 34201) registers on `DoCheckProc`, but `CheckProc` isn't a pure predicate — it mutates the owner's aura set:
```cpp
GetUnitOwner()->RemoveAurasDueToSpell(_lastForm);      // erase
GetUnitOwner()->CastSpell(GetUnitOwner(), _lastForm, true);  // insert
```
`CheckProc` runs via `CallScriptCheckProcHandlers` inside `Aura::GetProcEffectMask`, which the proc collector calls *while iterating that same unit's `m_appliedAuras`*. So the handler erases/inserts into the container being walked.

**Why it only became fatal now.** This script has done this since 2016. It survived because `m_appliedAuras` was a node-based `std::multimap` — insert never invalidates other iterators, erase only invalidates the erased node, so the collector's iterator (positioned on 34201) stayed valid. #25885 replaced the aura containers with `boost::container::flat_multimap` (vector-backed): insert can reallocate and erase shifts subsequent elements, so the reentrant mutation now invalidates the active iterator and every collected `AuraApplication*`. Next deref → the `GetId()` SIGSEGV. Timeline: #25885 merged 2026-07-14, crashes started the next day.

**Suggested direction.** The fix is the script, not this revert: make `CheckProc` side-effect-free and move the remove+cast into `OnProc`/`AfterProc` (or defer it off the proc path). Worth noting #25885 likely makes the same latent pattern fatal in *other* proc-check handlers that mutate the owner's auras — probably worth a sweep.

Happy to open a PR for the Frayer script fix.

PR #26644, opened 2026-07-15
fix(Scripts/Spells): defer aura mutations out of proc-check handlers

Nyeriah:
## Changes Proposed:

Fixes a class of use-after-free crash where a proc **check** handler (`DoCheckProc`) mutates the owner's aura containers while the proc engine is iterating them.

`DoCheckProc` runs from `Aura::GetProcEffectMask`, which `Unit::GetProcAurasTriggeredOnEvent` calls while iterating the owner's applied-aura map with a live iterator. Removing or applying an aura on that unit from inside the check reenters and mutates the container mid-walk. This was survivable while aura maps were node-based `std::multimap`s, but #25885 switched them to `boost::container::flat_multimap` (vector-backed): insert can reallocate and erase shifts elements, invalidating the live iterator and every collected `AuraApplication*`. The next dereference reads freed/moved storage and crashes (`SIGSEGV` in `GetProcEffectMask`/`GetId`).

Three handlers mutate the walked unit's own auras. Each is fixed by deferring the mutation to the owner's event queue so it runs outside the proc walk (behavior preserved; the mutation lands one tick later):

- `spell_botanica_shift_form_aura` (Frayer, spell 34201) — the form swap (`RemoveAura` + `CastSpell`) runs on every proc, so the erase reliably corrupts the collection walk. **This is the observed crash** (always creature 18587 / Frayer, map 553 Botanica).
- `spell_mage_burning_determination` — `RemoveAurasDueToSpell(54748)` on the proc target (the aura owner) when the buff is already up.
- `spell_mage_combustion` — `CastSpell(stack)` on the actor (the aura owner).

This PR proposes changes to:
- [ ] Core (units, players, creatures, game systems).

### AI-assisted Pull Requests

   - Tools/models used: Claude Code (Claude Opus 4.8)

## Issues Addressed:

- Related: #26634 (revert of #25647). That revert targets the wrong change; the crash root cause is the interaction described above with #25885, not #25647.

## SOURCE:

Diagnosed from GDB crash backtraces (15 dumps, all identical: `Unit::GetProcAurasTriggeredOnEvent` -> `Aura::GetProcEffectMask` -> `Aura::GetId`, always proc target 18587) combined with code analysis of the `#25885` container change. Not a game-behavior change, so the checkboxes below do not apply.

## Tests Performed:

`apps/codestyle/codestyle-cpp.py` passes. **Not yet compiled or tested in-game** — needs a build + in-game verification before merge.

## How to Test the Changes:

1. Enter The Botanica (map 553) and get a Frayer (entry 18587) to proc its shift-form aura by exposing it to incoming spells of different schools in quick succession. On an affected build (post-#25885) this crashes the worldserver in the aura proc path; with this PR it should not.
2. Mage `Burning Determination`: with the buff (54748) active, take another interrupt/silence effect so the `CheckProc` removal path runs.
3. Mage `Combustion`: activate Combustion and land non-crit fire damage so the crit-stack is applied.

- [ ] Defensive follow-up worth considering separately: have `Unit::GetProcAurasTriggeredOnEvent` iterate a snapshot instead of the live `flat_multimap`, so this whole class of script reentrancy is non-fatal.

### DIFF
--- a/src/server/scripts/Outland/TempestKeep/botanica/instance_the_botanica.cpp
+++ b/src/server/scripts/Outland/TempestKeep/botanica/instance_the_botanica.cpp
@@ -127,9 +127,19 @@ class spell_botanica_shift_form_aura : public AuraScript
             {
                 _swapTime = GameTime::GetGameTime().count() + 6;
                 _lastSchool = spellInfo->GetSchoolMask();
-                GetUnitOwner()->RemoveAurasDueToSpell(_lastForm);
+
+                // Runs from Aura::GetProcEffectMask while the proc engine iterates the owner's
+                // applied-aura map. Swapping forms mutates that map (RemoveAura + CastSpell),
+                // which invalidates the live iterator now that aura containers are flat_multimaps.
+                // Defer the swap to the owner's event queue so it runs outside the proc walk.
+                Unit* owner = GetUnitOwner();
+                uint32 const oldForm = _lastForm;
                 _lastForm = form;
-                GetUnitOwner()->CastSpell(GetUnitOwner(), _lastForm, true);
+                owner->m_Events.AddEventAtOffset([owner, oldForm, form]()
+                {
+                    owner->RemoveAurasDueToSpell(oldForm);
+                    owner->CastSpell(owner, form, true);
+                }, 1ms);
             }
         }
 
--- a/src/server/game/Spells/SpellScript.h
+++ b/src/server/game/Spells/SpellScript.h
@@ -823,12 +823,14 @@ public:
 #define AuraEffectSplitFn(F, I) EffectSplitFunction(&F, I)
 
     // executed when aura checks if it can proc
+    // do not use this hook for triggering spellcasts/removing auras etc - runs mid proc-iteration and may be unsafe
     // example: DoCheckProc += AuraCheckProcFn(class::function);
     // where function is: bool function (ProcEventInfo& eventInfo);
     HookList<CheckProcHandler> DoCheckProc;
 #define AuraCheckProcFn(F) CheckProcHandlerFunction(&F)
 
     // executed when aura effect checks if it can proc the aura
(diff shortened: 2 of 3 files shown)


src/server/scripts/Outland/TempestKeep/botanica/instance_the_botanica.cpp
```diff
@@ -127,9 +127,19 @@ class spell_botanica_shift_form_aura : public AuraScript
             {
                 _swapTime = GameTime::GetGameTime().count() + 6;
                 _lastSchool = spellInfo->GetSchoolMask();
-                GetUnitOwner()->RemoveAurasDueToSpell(_lastForm);
+
+                // Runs from Aura::GetProcEffectMask while the proc engine iterates the owner's
```
sogladev, line 131: Probably needs a doc note in `src/server/game/Spells/SpellScript.h` to forbid modifying aura map in the handler.

Language like existing 

> do not use this hook for triggering spellcasts/removing auras etc

```cpp
    // executed when aura checks if it can proc
    // example: DoCheckProc += AuraCheckProcFn(class::function);
    // where function is: bool function (ProcEventInfo& eventInfo);
    HookList<CheckProcHandler> DoCheckProc;
#define AuraCheckProcFn(F) CheckProcHandlerFunction(&F)
```

sogladev (approved):

src/server/scripts/Outland/TempestKeep/botanica/instance_the_botanica.cpp
Nyeriah: Done in 2b7e59589d. Added the warning to both `DoCheckProc` and `DoCheckEffectProc` (both run during proc-collection iteration over the applied-aura map), mirroring the existing "may be unsafe" wording on the effect-apply/remove hooks.

PR #26648, opened 2026-07-16
refactor(Core): expose a few values more publicly

airbreather:
## Changes Proposed:

This PR proposes changes to:

While working on [mod-nodejs](https://github.com/airbreather/mod-nodejs), I found some places where it would be useful to have a bit more access to a few more values on domain objects.

The unusual nature of this module has me paying extra attention to the declared types of various values throughout the system, so I remember seeing several imperfections and mismatches throughout the past few months. I've mostly been ignoring trivial mismatches that don't seem like they could ever come up, but I couldn't let `LootItem::count` go.

Sure enough, when actually testing that one, the client was just fine with stacks FAR exceeding 255, though some combination of the client and server seem to disapprove of exactly 2^31-1.

all-the-items

I have a feeling that some modules that add area looting would appreciate this, especially because the client actually lets you EXCEED the item's max stack limit and loot the items **IF** you have enough bag space for them once they split into their respective stacks properly:

|before clicking it|after clicking it|
|-|-|
|overstuffed|but-still-ok|

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: pi.dev agent, Qwen 3.6-27B-Q5_K_M model hosted locally.
   - Only tangentially while working on the early stages of the module that prompted this change, and I don't think any of its code changes survived to this particular PR (if it even touched Core at all), so it's almost a "no", but I think it still counts enough.

## Issues Addressed:

- Nothing logged, just (apart from one specific line) exposing some APIs to read/write data to/from some domain objects.
- The one exception to that is widening a field from `uint8_t` to `uint32_t`, see [this comment in #dev-core](https://discord.com/channels/217589275766685707/284323424032129024/1521470265245302874) discussing it.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

For the most part, if it compiles, it works - most of it is just trivial accessors for things that already exist. There are two things that actually change existing code and structures, though, so if I were a maintainer, I would expect a contributor to AT LEAST give them proper regression testing.

Easy part:
1. have privileges
2. select yourself
3. `.guild create "Grownups"`
4. change Guild Message of the Day
5. change Guild Information
6. relog
7. make sure the changes stuck

Harder part is proving that the loot item count change is correct, because nothing blizzlike will even come close to hitting that cap (otherwise it wouldn't have taken so many years, lol). I came upon it in the context of developing [mod-nodejs](https://github.com/airbreather/mod-nodejs), so it's trivial for me to just throw this into some script:

```typescript
Acore.hooks.on('misc:after-loot-template-process', (args) => {
	for (const entry of [2589, 9358]) {
		args.loot.addItem(entry, 2147483646, 2147483646, 1, 1 /* LOOT_MODE_DEFAULT */, false, true, 2147483646);
	}
	args.loot.addItem(20384, 1500, 1500, 1, 1 /* LOOT_MODE_DEFAULT */, false, true, 1500);
});
```

to see the last screenshots above after killing a lootable creature and going to loot it. But you could get the same effect by adding code to the gap between these two lines that adds something with a stack size exceeding 255 and then looting a corpse: https://github.com/azerothcore/azerothcore-wotlk/blob/70da2f25d67cebb0670a84ccf685abb923ca4549/src/server/game/Loot/LootMgr.cpp#L563-L565

For example, putting this in there will add a single stack of 1500 carapaces that you can loot if you have at least 8 free bag slots (we can't do it in one go because `LootStoreItem`'s min/max counts ACTUALLY look risky to go over 255):

```cpp
this->AddItem({20384, 0, 1, false, 1, 0, 1, 1});
this->items.back().count = 1500;
```

- [ ]
- [ ]


airbreather:
~~:page_facing_up: the branch is intentionally based on a commit from a couple of weeks ago (at the time of writing) so that this same change can cleanly merge both here and into the mod-playerbots fork.~~

~~If merge conflicts arise in the meantime, I'll abandon that idea, but for now - with all the different branches I'm looking at - I want to be able to see which branches have X commit or Y commit, which I can't do with cherry-picking or rebasing.~~

edit: nah, I'll just keep it up to date.

NoxMax:
You don't have to worry about Playerbots. The Playerbots AC fork is updated once a week anyway.

airbreather:
:thinking: for some reason I remember seeing it get pretty far behind while working on my current project, but it does look like updates have been reasonably consistent. All right, I'll go ahead and keep this up to date.

src/server/game/Loot/LootMgr.h
```diff
@@ -159,7 +159,7 @@ struct LootItem
     ConditionList conditions;                               // additional loot condition
     AllowedLooterSet allowedGUIDs;
     ObjectGuid rollWinnerGUID;                              // Stores the guid of person who won loot, if his bags are full only he can see the item in loot list!
-    uint8   count             : 8;
+    uint32  count;
```
sudlud, line 162: why

airbreather: might get reverted before the PR leaves draft depending on if it actually makes things worse

but it's essentially because of things like this https://github.com/azerothcore/mod-ale/blob/9e5b8c66efeb383871ec58b925e47094c92cc8d5/src/LuaEngine/methods/LootMethods.h#L30-L72 that have to delicately tiptoe around this limit even for items that stack high enough.

everything that touches this value directly is already converting to or from a 32-bit value immediately (or it doesn't matter because it's a literal `0`), so **unless there's a client limitation with using values >255** (which I still need to check), all it does is make module developers have to split stacks of certain items earlier than where their max stack size would otherwise suggest that they need to

airbreather: yep, the client is perfectly fine with it:

all-the-items

not 100% sure why 2147483647 is getting split into 2147483646 and 1, but that could be a mistake in [this code](https://github.com/airbreather/mod-nodejs/blob/94767403c911a18f72b4b19d186be1ae76cca1e5/src/templates/LootTemplate.cpp#L127-L168) since that's at least slightly complicated and this is the first time I'm testing it like this.

airbreather:
> `1294-1294`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> 
> **Apply Allman braces to the new Guild function definitions.**
> 
> The three newly added definitions use same-line opening braces, violating the repository’s C++ formatting convention.
> 
> - `src/server/game/Guilds/Guild.cpp#L1294-L1294`: move the `SetMOTD` opening brace to the next line.
> - `src/server/game/Guilds/Guild.cpp#L1319-L1319`: move the `SetInfo` opening brace to the next line.
> - `src/server/game/Guilds/Guild.h#L783-L783`: format the inline `GetUnorderedMembers` body with Allman braces.

Agreed on the two `Guild.cpp` bits. `Guild.h` is more consistent with the code immediately around it, so I'll wait for a human to confirm that this is specifically something they want to have happen.

PR #26691, opened 2026-07-19
fix(Scripts/Ulduar): XT-002 Searing Light and Gravity Bomb target one non-tank player

### LINKED ISSUE #26687: [Uludar]XT-002 "Gravity Bomb" and "Searing Light" are incorrectly cast on all players.
### Current Behaviour

After commit c09afb080 refactored XT‑002, Gravity Bomb and Searing Light are erroneously applied to all players. Initial investigation suggests that the refactoring removed the old SpellScript logic and replaced it with the current AuraScript structure, but the filtering hook originally attached to OnObjectAreaTargetSelect was not preserved. As a result, the target selection now expands to all alive players.

### Expected Behaviour

Both abilities are intended to affect only a single random player per cast, and they should exclude the current tank.

### Source

_No response_

1..go creature 33350.
2.Engage XT‑002 with a tank character to build threat; ensure at least two other non‑tank players are alive and within the boss’s spell range.
3.During the fight, wait for the boss to cast Gravity Bomb (spell ID 64238) and Searing Light (spell ID 63018). These abilities are typically used after a few seconds and periodically.

### Extra Notes

_No response_

### AC rev. hash/commit

92bc564

### Operating system

win 11 X64

### Custom changes or Modules

_No response_

Nyeriah:
## Changes Proposed:
This PR proposes changes to:

XT-002's Searing Light (63018/65121) and Gravity Bomb (63024/64234) are area-enemy spells with no target cap in our spell data. Before the recent rework they were kept to a single random non-tank target by the boss casting them with `SPELLVALUE_MAX_TARGETS, 1` plus a `SpellScript` that removed the current victim from the target list.

The rework (c09afb0, #26345) ported the TrinityCore AI, which casts them with a bare `DoCastSelf` and relies on TC applying `MaxAffectedTargets = 1` to these spells in its `SpellMgr` corrections. AzerothCore has no equivalent correction, and the rework also removed the old `SpellScript`, so nothing constrained the target and both debuffs landed on every player in range.

This restores the intended behavior with a single `SpellScript` bound to all four spell IDs. In `OnObjectAreaTargetSelect` it removes the boss's current victim (the tank) and then resizes the list to one random target. If the tank is the only candidate the list ends up empty and the cast simply fizzles, which is the desired "never the tank" outcome.

### AI-assisted Pull Requests

   - Tools/models used: Claude Code (Opus 4.8)

## Issues Addressed:
- Closes #26687

## SOURCE:

Cross-checked against reference implementations: TrinityCore caps these spells to one target via `SpellMgr` (`MaxAffectedTargets = 1`), and CMaNGOS selects a single random attacker through its spell-list targeting. Both limit the abilities to one player per cast. This restores AzerothCore's own prior behavior, which additionally excludes the current tank.

## Tests Performed:
This PR has been:

## How to Test the Changes:

1. Engage XT-002 (creature 33350) with a tank and at least two other players in range.
2. Observe periodic Searing Light and Gravity Bomb casts.
3. Each cast applies to a single random player other than the current tank, instead of all players in range.

### DIFF
--- a/src/server/scripts/Northrend/Ulduar/Ulduar/boss_xt002.cpp
+++ b/src/server/scripts/Northrend/Ulduar/Ulduar/boss_xt002.cpp
@@ -709,6 +709,29 @@ private:
     TaskScheduler _scheduler;
 };
 
+// 63018, 65121 - Searing Light / 63024, 64234 - Gravity Bomb
+// Both hit a single random player, never the current tank
+class spell_xt002_searing_light_gravity_bomb : public SpellScript
+{
+    PrepareSpellScript(spell_xt002_searing_light_gravity_bomb);
+
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        if (Unit* victim = GetCaster()->GetVictim())
+        {
+            ObjectGuid victimGuid = victim->GetGUID();
+            targets.remove_if([victimGuid](WorldObject* target) { return target->GetGUID() == victimGuid; });
+        }
+
+        Acore::Containers::RandomResize(targets, 1);
+    }
+
+    void Register() override
+    {
+        OnObjectAreaTargetSelect += SpellObjectAreaTargetSelectFn(spell_xt002_searing_light_gravity_bomb::FilterTargets, EFFECT_ALL, TARGET_UNIT_DEST_AREA_ENEMY);
+    }
+};
+
 // 63018, 65121 - Searing Light
 class spell_xt002_searing_light_spawn_life_spark : public AuraScript
 {
@@ -1057,6 +1080,7 @@ void AddSC_boss_xt002()
     RegisterUlduarCreatureAI(npc_boombot);
     RegisterUlduarCreatureAI(npc_life_spark);
     RegisterUlduarCreatureAI(npc_xt_void_zone);
+    RegisterSpellScript(spell_xt002_searing_light_gravity_bomb);
     RegisterSpellScript(spell_xt002_searing_light_spawn_life_spark);
     RegisterSpellScript(spell_xt002_gravity_bomb_aura);
     RegisterSpellScript(spell_xt002_gravity_bomb_damage);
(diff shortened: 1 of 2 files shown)


src/server/scripts/Northrend/Ulduar/Ulduar/boss_xt002.cpp
```diff
@@ -709,6 +709,29 @@ struct npc_xt_void_zone : public PassiveAI
...
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        if (Unit* victim = GetCaster()->GetVictim())
+        {
+            ObjectGuid victimGuid = victim->GetGUID();
+            targets.remove_if([victimGuid](WorldObject* target) { return target->GetGUID() == victimGuid; });
+        }
+
+        Acore::Containers::RandomResize(targets, 1);
+    }
```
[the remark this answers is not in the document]
Nyeriah: SPELL_ATTR3_ONLY_TARGET_PLAYERS

src/server/scripts/Northrend/Ulduar/Ulduar/boss_xt002.cpp
```diff
@@ -709,6 +709,29 @@ struct npc_xt_void_zone : public PassiveAI
...
+class spell_xt002_searing_light_gravity_bomb : public SpellScript
+{
+    PrepareSpellScript(spell_xt002_searing_light_gravity_bomb);
+
+    void FilterTargets(std::list<WorldObject*>& targets)
+    {
+        if (Unit* victim = GetCaster()->GetVictim())
+        {
+            ObjectGuid victimGuid = victim->GetGUID();
+            targets.remove_if([victimGuid](WorldObject* target) { return target->GetGUID() == victimGuid; });
```
Kitzunu: why not use `Acore::ObjectGUIDCheck()` instead?

Nyeriah: Good call, done in 80d6cf1. Switched to Acore::ObjectGUIDCheck, which is also what the pre-rework script used.

Kitzunu (approved):

PR #26692, opened 2026-07-19
feat(Core/Creature): add creature_hostile_override table to allow scripted combat between neutral creatures

### LINKED ISSUE #26659: [NPCs] Neutral-faction creatures can chase endlessly without fighting
### Current Behaviour

When a creature is directed to attack a target of a mutually-neutral faction (neither side flagged hostile to the other in FactionTemplate.dbc), combat initiates but never resolves: the attacker repeatedly acquires and drops the target, chasing it without landing sustained damage (I.E. Lady Alistra (faction 2050) --> Cenarion Scout (faction 994)

### Expected Behaviour

These NPCs should be able to attack each other.

### Source

_No response_

1. Two creatures whose faction templates are mutually neutral (no hostileMask/enemyFaction linking them).
2. Force one to attack the other (AttackStart / SmartAI / script).
3. Attacker chases but combat never executes

### Extra Notes

Potentially targeted files:

src/server/game/Entities/Creature/Creature.cpp
Location: _IsTargetAcceptable (2594–2623), CanCreatureAttack (2662)
────────────────────────────────────────
src/server/game/Combat/ThreatManager.cpp
Location: ThreatReference::ShouldBeOffline (112–121)
────────────────────────────────────────
src/server/game/Entities/Unit/Unit.cpp
Location: IsValidAttackTarget (10763), _IsValidAttackTarget (10769)
────────────────────────────────────────
src/server/game/Entities/Unit/Unit.h
Location: IsEngagedBy (933)

issue found when trying to implement ambient combat event: 
https://warcraft.wiki.gg/wiki/Lady_Alistra
https://warcraft.wiki.gg/wiki/Shadow_Construct
https://warcraft.wiki.gg/wiki/Forgotten_Servant
https://warcraft.wiki.gg/wiki/Cenarion_Scout

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/448f07f06e6b

### Operating system

win11

### Custom changes or Modules

_No response_

Voxstrasza:
Create a new world DB table to whitelist creatures that have mutually neutral factions to allow scripted combat without changing their blizz-like factions.
## Changes Proposed:

### AI-assisted Pull Requests
   - Tools/models used: Opus 4.8 was used to diagnose the issue and search for an addition to the core to implement feature without causing regressions in other areas

## Issues Addressed:
- Closes #26659

## Tests Performed:

I am able to build this and have started work on implementing the Lady Alistra ambient event (described below). Lady Alistra attacks the Cenarion Scout and it is able to play out as seen on retail.

## Testing
1. Lady Alistra (29195) has a blizzlike combat sequence with Cenarion Scout (29220), Forgotten Servant (29232), Shadow Construct (29231) which blizzlike factions cannot naturally attack each other. This PR will allow this ambient event to take place in-game.
2. Amal'thazad (29194) has a blizzlike combat sequence involving Frigid Bones (29210) and Disciple of Frost (28490) which blizzlike factions cannot naturally attack each other. This PR will allow this ambient event to take place in-game.
3. During the quest [Last Rites (12019)](https://warcraft.wiki.gg/wiki/Last_Rites), Thassarian (26170) has a blizzlike combat sequence with Prince Valanar (28189) which blizzlike factions cannot naturally attack each other. This PR will allow Thassarian to assist players.
4. There are possibly more cases, especially in WotLK, where this table will get use.


Voxstrasza:
cc @blinkysc

blinkysc:
I feel like this likely like using a large backhoe to dig a small hole? For the original issue that we’re trying to solve, must it be that it’s a full combat lifecycle? Or is it more or less a visual “sparring”. Because if that’s the case likely the infrastructure exists?

Voxstrasza:
> I feel like this likely like using a large backhoe to dig a small hole? For the original issue that we’re trying to solve, must it be that it’s a full combat lifecycle? Or is it more or less a visual “sparring”. Because if that’s the case likely the infrastructure exists?

i looked into that and it's not sparring, it's actual combat data that would get sent to a player's combat log if they hang around. the event and subsequent events are driven by summon -> kill -> talk cycle -> repeat

it's also backed by sniffs: lady alistra -> casts -> plague strike -> on -> cenarion scout

retail combat log (had to play with some settings):
image

sniffs organized by claude:

> LADY ALISTRA (29195) — SNIFF-VERIFIED COMBAT PROOF
> Alistra 0x203CAC00001C82C0000025000058EA98 (Entry 29195) vs Cenarion Scout (Entry 29220)
> SMSG_SPELL_NON_MELEE_DAMAGE_LOG (real server damage packets)
> 
> CENARION SCOUT DEMO (16:40) — interleaved cast + damage timeline
> 16:40:23.742  SCOUT 29220  cast 53708 Shadow Summon Visual (spawn)
> 16:40:26.823  ALISTRA      cast 53765 Open Shadow Portal
> 16:40:34.730  ALISTRA      cast 53694 Plague Strike   -> DMG 955  on 29220
> 16:40:34.730  SCOUT 29220  cast 53691 Tree of Life (morph)
> 16:40:37.197  SCOUT 29220  cast 53608 Lifebloom (HoT)
> 16:40:38.429  ALISTRA      cast 53605 Degeneration    -> DMG 654  on 29220
> 16:40:38.429  SCOUT 29220  cast 53692 Lifebloom (heal)
> 16:40:42.057  ALISTRA      cast 53605 Degeneration    -> DMG 485  on 29220
> 16:40:42.057  SCOUT 29220  cast 53607 Rejuvenation
> 16:40:45.711  ALISTRA      cast 53605 Degeneration    -> DMG 690  on 29220
> 16:40:47.295  SCOUT 29220  cast 53608 Lifebloom (HoT)
> 16:40:49.368  ALISTRA      cast 53605 Degeneration    -> DMG 523  on 29220
> 16:40:49.368  SCOUT 29220  cast 53692 Lifebloom (heal)
> 16:40:52.968  ALISTRA      cast 53605 Degeneration    -> DMG 606  on 29220
> 16:40:56.569  ALISTRA      cast 53605 Degeneration    -> DMG 678  on 29220  OverKill=534  ***KILL***
>

Kitzunu:
An override like this feels weird. Are we sure there is no faction flag, or that we should do it via SAI?
Maybe it is right, but it feels odd

Voxstrasza:
> An override like this feels weird. Are we sure there is no faction flag, or that we should do it via SAI? Maybe it is right, but it feels odd

hi kitzunu

there is no faction flag, no faction swap according to sniffs, no hidden server-side spell to enable combat or anything that would suggest enabling combat a different way (i scanned the sniffs multiple times over).  without an override, you _can_ implement the combat sequence via SAI but the aggressor (alistra) won't attack, will just repeatedly target in combat idle and the sequence is stuck

**possible workarounds** if you genuinely don't want this: script the NPCs to die on a timer, or have them cast debuffs on themselves to die, or change the factions to something that is mutually hostile but these are all hacks and directly go against sniff data

blinkysc:
I honestly think that the override approach with the database isn’t the correct approach. This feels way too niche as you said in the issue too. Typically having the database settings or overrides is for things we typically have lots of changes/maintenance for, and I don’t think this scenario is applicable. I propose closing this and finding possible other solutions =(

Gultask:
Doesn't Trinity have this in a sparring table? From what I can recall, it depends on the person that scripted the area. I know that sparring in Shattered Halls flip their factions manually, for example. So does the sparring in Black Temple gladiators. But later on in WotLK I think they added some new sparring method.

Rule of thumb, sparring tables have a floor where a npc cannot get damaged below a certain percentage, usually 85%, and only the player can kill them.

Voxstrasza:
> Doesn't Trinity have this in a sparring table? From what I can recall, it depends on the person that scripted the area. I know that sparring in Shattered Halls flip their factions manually, for example. So does the sparring in Black Temple gladiators. But later on in WotLK I think they added some new sparring method.
> 
> 
> 
> Rule of thumb, sparring tables have a floor where a npc cannot get damaged below a certain percentage, usually 85%, and only the player can kill them.

which is why I didn't use sparring, this is an actual combat event where NPCs kill NPCs with zero player involvement. I've said this. 

https://warcraft.wiki.gg/wiki/Lady_Alistra#Dialogue

https://warcraft.wiki.gg/wiki/Amal%27thazad#Quotes

Gultask:
Oh, my bad hehehe.

My guess it's a faction thing deeper in. From what I can recall a lot of faction flags (DBC or whatever) aren't implemented.
Like, for example, some early level creatures run away slightly when combat occurs near them. The consensus was that it was faction related.

My opinion is that a table is overkill for a single case.

Voxstrasza:
> Oh, my bad hehehe.
> 
> 
> 
> My guess it's a faction thing deeper in. From what I can recall a lot of faction flags (DBC or whatever) aren't implemented.
> 
> Like, for example, some early level creatures run away slightly when combat occurs near them. The consensus was that it was faction related.
> 
> 
> 
> My opinion is that a table is overkill for a single case.

it wouldn't be a single case;

lady alistra event 1
event 2
event 3
amal'thazad event 1
event 2
last rites quest

6 total so far. there are potentially more. need to do some more research. off the top of my head, potentially the Vrykul next to the shadow vault that fight.

Kitzunu:
What if you change the CvC check in unit.cpp? If you add `IsInCombatWith(target)` to the return, shouldn't it theoretically work with CvC combat if it is forced in script/SAI?

It would be less of a hack since it wouldn't define unique creatures for every event you find that should have this behavior, but instead it would work with just the script. As well as `CanBeginCombat` allows neutral targets to create combat, but `_IsValidAttackTarget` doesnt at the moment. But with the change I proposed above it would allow it.

Any objections to this approach? @Gultask @blinkysc @Voxstrasza

Voxstrasza:
> What if you change the CvC check in unit.cpp? If you add `IsInCombatWith(target)` to the return, shouldn't it theoretically work with CvC combat if it is forced in script/SAI?
> 
> It would be less of a hack since it wouldn't define unique creatures for every event you find that should have this behavior, but instead it would work with just the script. As well as `CanBeginCombat` allows neutral targets to create combat, but `_IsValidAttackTarget` doesnt at the moment. But with the change I proposed above it would allow it.
> 
> Any objections to this approach? @Gultask @blinkysc @Voxstrasza

no objections here, I'll give it a try and report my findings

Kitzunu:
Here's the diff I have if you want. lmk how it works

```diff
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -10741,9 +10741,11 @@ bool Unit::_IsValidAttackTarget(Unit const* target, SpellInfo const* bySpell, Wo
-    // CvC case - can attack each other only when one of them is hostile
+    // CvC case - can attack each other only when one of them is hostile,
+    // or when combat between them is already established through script or SAI
     if (!HasUnitFlag(UNIT_FLAG_PLAYER_CONTROLLED) && !target->HasUnitFlag(UNIT_FLAG_PLAYER_CONTROLLED))
-        return GetReactionTo(target) <= REP_HOSTILE || target->GetReactionTo(this) <= REP_HOSTILE;
+        return GetReactionTo(target) <= REP_HOSTILE || target->GetReactionTo(this) <= REP_HOSTILE
+            || IsInCombatWith(target);
```

Kitzunu:
Be ware if there are any RP cases though existing that for some reason use SetInCombatWith() that shouldnt kill eachother, because that would make them valid combatants with this I assume

Voxstrasza:
> Be ware if there are any RP cases though existing that for some reason use SetInCombatWith() that shouldnt kill eachother, because that would make them valid combatants with this I assume

yes, this change will require an investigation of any and all scripted combat to confirm if the result is death or player intervention

Voxstrasza:
@Kitzunu 
image

that gets us there, but i am concerned about regressions elsewhere

Kitzunu:
Shouldnt the real "RP" events use sparring instead them, for those that may break due to this but that shouldnt kill eachother?

Voxstrasza:
> Shouldnt the real "RP" events use sparring instead them, for those that may break due to this but that shouldnt kill eachother?

that depends on what we define as sparring - currently, what's defined as sparring are two NPCs 'fighting', bringing each other to a certain % and not moving beyond that without player involvement

if NPCs are supposed to kill each other, or are to assist in player combat (Thassarian quest) then this fix is necessary to bypass faction limitations

PR #1416, opened 2019-02-05
Core/Instances: Finally fix instance_reset

Nefertumm:
##### CHANGES PROPOSED:

-  Change the if condition.
-  

###### ISSUES ADDRESSED:

Hopefully Closes #966

##### TESTS PERFORMED:

Compiled, debugged and it worked. 

##### HOW TO TEST THE CHANGES:

Just play with instances.

##### KNOWN ISSUES AND TODO LIST:

- [ A lot of test. ]

##### Target branch(es):

Master

PR #26711, opened 2026-07-20
fix(Core/Instances): day-align expired reset time

CarloHagens:
## Changes Proposed:

In `InstanceSaveMgr::LoadResetTimes` the expression `t = (t * DAY) / DAY;` is a no-op. Looking at #1416, where this block was reworked, the intent was `t = (t / DAY) * DAY;`, rounding the expired reset time down to midnight before computing the next one.

Since the old hour of day survives, a global reset time that expires while the server is offline gets `Instance.ResetTimeHour` added on top of its old hour at startup, so the reset hour drifts later on every such recalculation. Resets that happen while the server is running go through `_ResetOrWarnAll`, which aligns correctly. On my server all heroic rows in `instance_reset` were at 04:00 while every raid row had drifted to 12:00 with the default `Instance.ResetTimeHour = 4`.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code (Claude Fable 5). The code change, this description and the tests were produced by the AI agent, working with the author.

## SOURCE:

## Tests Performed:

A/B tested on my server (Docker on a Windows host) with the steps below: same planted reset time (20:00 the previous day), buggy build recalculated it to 00:00, fixed build to 04:00 (`Instance.ResetTimeHour = 4`). Builds without errors.

This PR has been:

## How to Test the Changes:

1. Stop the worldserver and pick a heroic row in the characters DB table `instance_reset`.
2. Set its `resettime` to yesterday at an hour different from `Instance.ResetTimeHour`, for example 20:00.
3. Start the worldserver and check the row. Broken: the new time keeps the wrong hour and adds `Instance.ResetTimeHour` on top (20:00 + 4h gives 00:00). Fixed: it lands on `Instance.ResetTimeHour` (04:00 by default).

Note: the wrong hour plus `Instance.ResetTimeHour` must still be ahead of the current time of day, otherwise the reset fires immediately at startup and `_ResetOrWarnAll` corrects the row before you can observe it.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -354,7 +354,7 @@ void InstanceSaveMgr::LoadResetTimes()
         {
             // assume that expired instances have already been cleaned
             // calculate the next reset time
-            t = (t * DAY) / DAY;
+            t = (t / DAY) * DAY;
             t += ((today - t) / period + 1) * period + diff;
             CharacterDatabase.DirectExecute("UPDATE instance_reset SET resettime = '{}' WHERE mapid = '{}' AND difficulty = '{}'", (uint32)t, mapid, difficulty);
         }


Nyeriah (approved):
Reviewed and verified the analysis:

- The old line is provably a no-op: with 64-bit time_t, `(t * DAY) / DAY == t` exactly, no overflow involved.
- Traced the line to 6dfb40b446 (#1416), which introduced this block in 2019. The day-floor idiom `(x / DAY) * DAY` appears in the same function at lines 302 and 340 and in `_ResetOrWarnAll` (https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Instances/InstanceSaveMgr.cpp#L586), so the operands were simply transposed.
- TrinityCore's equivalent block does the same day alignment before computing the next reset: https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Instances/InstanceSaveMgr.cpp#L414-L419
- With `t` floored to midnight, `today - t` and `period` are whole-day multiples, so the result lands exactly on `Instance.ResetTimeHour` and is guaranteed to be in the future (>= today + DAY + diff), no immediate re-fire at startup.
- Rows that already drifted get realigned on their next expiry, so existing databases self-heal without manual cleanup.

Startup and runtime recalculation now agree on the reset hour, which matches the drift symptom described (heroics correct at 04:00, raids drifted to 12:00).

PR #26715, opened 2026-07-21
fix(Commands/GM): .learn all my talents now learns every talent

Chinoske:
## Changes Proposed:

`.learn all my talents` could silently skip talents that have a cross-tree `DependsOn` prerequisite. `Player::LearnTalent` still enforces that dependency check even when called with `command=true` (only the free-point and tier-row checks are skipped for that flag), and `sTalentStore` isn't ordered by dependency - so a dependent talent processed before its prerequisite in the same pass would just get silently skipped and never learned. The command now retries in passes until nothing new gets learned, so those catch up instead of being missed.

`.learn all my talents` also takes an optional `1`/`2`/`3` argument to force-learn just one talent tree instead of all three, matching each tree's position in-game (mapped through the DBC's `tabpage` field). Picking a tab force-learns that tree's talents first, then learns every trainer spell afterwards (same as `.learn all my class` grants) - so any trainer spell gated behind one of those talents is already teachable by then. The plain no-arg command stays talents-only. The free talent point pool is zeroed out afterwards either way (full command or single tab) - otherwise those points would stay spendable on top of the tree that was just force-filled, letting the player go over their level's talent point cap.

`.learn all my class` keeps its original behavior of force-learning every talent as part of teaching the class (trainer spells + all talents + quest-reward spells), now benefiting from the same multi-pass fix.

A new `LANG_COMMAND_LEARN_CLASS_TALENTS_TAB` string was added for the per-tab success message; the existing `LANG_COMMAND_LEARN_CLASS_TALENTS` string is reused for the no-argument case since it's still accurate.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- None

## SOURCE:

## Tests Performed:
This PR has been:

Confirmed `.learn all my talents` fills out every talent (including ones with cross-tree dependencies that previously got skipped), the `1`/`2`/`3` tab argument fills the chosen tree before learning trainer spells while still zeroing free points, and `.learn all my class` still learns every talent as before.

## How to Test the Changes:

1. On a level 80 character, run `.learn all my talents` - confirm every talent in all 3 trees ends up learned, including any with a cross-tree prerequisite, and no trainer spells are touched.
2. On the same character (reset talents first), run `.learn all my talents 1` (then `2`, then `3`) - confirm only the corresponding tree gets filled, trainer spells are learned afterwards, and free talent points end up at 0.
3. On a fresh/low-talent character, run `.learn all my class` - confirm trainer spells, all talents, and quest-reward spells are learned, same as before this PR.

- None

🤖 Generated with Claude Code (Sonnet 5)+AzerothMCP


Kitzunu:
1. .learn all my class must still learn all talents. It's always meant "learn everything for my class" (trainer + all talents + quest). Swapping to InitTalentForLevel() redefines the command, please keep it learning talents (reuse your improved talents routine). Note InitTalentForLevel() can even resetTalents() on some characters, wiping talents instead.
2. Dont hardcode text.
3. .learn all my talents <tab> leaves the full point pool intact while maxing a tree, putting the char over cap

Chinoske:
@Kitzunu good catches, thanks.

1. Reverted `.learn all my class` back to force-learning everything through the fixed talents routine - you're right the command's always meant "learn everything for my class". Bonus reason to drop `InitTalentForLevel()` here: it can call `resetTalents()` on a char whose used talent count is above what their level allows, which would've wiped talents instead of just granting points.
2. Moved the new message to a proper `LANG_COMMAND_LEARN_CLASS_TALENTS_TAB` string instead of hardcoding it. The no-arg case just reuses the existing `LANG_COMMAND_LEARN_CLASS_TALENTS` string again, no new entry needed there.
3. Free talent points now get zeroed after `.learn all my talents <tab>` too, same as the full command, so there's no leftover pool to spend on another tree and go over the level cap.

Pushed.

Chinoske:
Small update: `.learn all my talents <tab>` now also learns every trainer spell (talents first, trainer spells after, so anything gated behind one of those talents is teachable by then) - matches what `.learn all my class` grants, just scoped to one tree instead of all three. Description updated to match.

Chinoske:
Re-reviewed the final shape of this after the earlier feedback rounds. The dependency-ordering loop re-checked against Player::LearnTalent: with command=true only the free-point and tier-row checks are skipped, cross-tree DependsOn still applies, and sTalentStore isn't dependency-ordered - so the repeat-until-stable pass is genuinely needed, and it terminates because each pass either learns something new or stops. Tab filtering via talentTabInfo->tabpage (0-2 UI order) re-verified, and .learn all my class still force-learns everything as agreed. Nothing further to change.

Chinoske:

PR #26724, opened 2026-07-21
fix(Core/Player): .reset spells restores exact character-creation state

Chinoske:
## Changes Proposed:
This PR proposes changes to:

`Player::resetSpells()` (used by `.reset spells`) removed every spell in the player's spell map - quest rewards, custom start spells, default class/race spells, all of it - then tried to restore a subset of that (`LearnDefaultSkills()`, `LearnCustomSpells()`, `learnQuestRewardedSpells()`). That restore doesn't cover everything that was removed, so an online player ended up with spells missing from their spellbook until they relogged.

The fix now targets exactly what `.learn all my class` grants:
- Trainer-taught spells (built from `sObjectMgr->GetClassTrainers()`).
- Quest-reward spells for quests the character's class can complete (mirrors the matching logic `.learn all my class` itself uses).
- Talents, reset separately via `resetTalents()` in the command handler (they live in their own map, not the spell map).

This race/class's starting-kit spells (`PlayerInfo::customSpells`, e.g. a Warrior's Heroic Strike and Battle Stance) are excluded from removal entirely, even though a trainer also teaches them - removing and re-adding them has side effects (via `removeSpell()`'s rank-chain cascade) that can leave things like stance-switching broken until the character relogs, so the safest fix is to never touch them in the first place. The command also forces a `SendInitialSpells()` resync so the client picks up the change without needing to relog.

## Known limitation
In testing, the server-side state after `.reset spells` is always correct (verified directly against the database), but the client's spellbook window can still show stale entries in the same session for spells removed via chain cascades - a full relog reliably shows the correct state. This appears to be a client-side display quirk rather than a data or logic issue, and is out of scope for what a server-side fix can address.

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code (Claude Sonnet 5)

## Issues Addressed:
- No linked issue.

## SOURCE:

## Tests Performed:
This PR has been:

Confirmed via direct database inspection that running `.learn all my class` followed by `.reset spells` on a Warrior leaves only starting-kit/racial/quest spells in `character_spell` - trainer-taught combat abilities (Whirlwind, Shield Wall, Disarm, etc.) are correctly removed, while Heroic Strike, Battle Stance, weapon skills, and racials remain untouched.

## How to Test the Changes:

- [x] This pull request can be tested by following the reproduction steps provided below.

1. On an online character, run `.learn all my class` to pick up trainer spells, all talents, and quest-reward spells.
2. Run `.reset spells` on the same character.
3. Check the character's `character_spell` table (or relog and open the spellbook) - confirm trainer-taught and quest-reward spells are gone, talents are reset, while starting-kit spells (e.g. Heroic Strike/Battle Stance for a Warrior) and quest-reward spells outside the class's own quests remain.

### DIFF
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -12055,36 +12055,24 @@ void Player::resetSpells()
     if (HasAtLoginFlag(AT_LOGIN_RESET_SPELLS))
         RemoveAtLoginFlag(AT_LOGIN_RESET_SPELLS, true);
 
-    // Only remove active (non-passive) spells this character could re-learn from a trainer, plus
-    // the active spells granted for free at creation (e.g. a Paladin's starting Seal of
-    // Righteousness/Holy Light) - weapon skills, languages, and other passive grants from that
-    // same creation list are left alone, along with quest-reward spells. The player never ends up
-    // with an empty spellbook: re-learning happens the normal way, by visiting a trainer again (or
-    // a GM re-teaching them).
-    std::unordered_set<uint32> resettableSpellIds;
-    for (Trainer::Trainer const* trainer : sObjectMgr->GetClassTrainers(getClass()))
-        for (Trainer::Spell const& trainerSpell : trainer->GetSpells())
-            resettableSpellIds.insert(trainerSpell.SpellId);
-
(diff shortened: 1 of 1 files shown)


Chinoske:
Closing - ended up reverting the change back to stock `Player::resetSpells()` after testing, nothing left to propose here.

PR #26758, opened 2026-07-23
fix(Core/Spells): Skip charge pre-path for oversized targets

### LINKED ISSUE #26266: [Ulduar] Kologarn - Charging Kologarn causes the character to rush to his feet and dying
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9741

Triage Notes: 

reproduce:
- make a war
- .levelup 79
- .learn all my class
- .go xyz 1797.15 -24.4027 448.741 603
- jump and charge kologarn
- observe
test video: https://dl.dropboxusercontent.com/scl/fi/z51sxkxm1hl7n59ujoukv/Wow_shg9aP0Z3e.mp4?rlkey=tzcvdy9cp9lm8rpebs90k9x2g&dl=0

Original Post Below: 

---
### Current Behaviour

Charging Kologarn causes the character to rush to his feet and dying

https://github.com/user-attachments/assets/2dd99b3d-094b-4e17-bd9e-f3429be1f3dd

### Expected Behaviour

Charge should end in front of the boss

https://youtu.be/OaIWgbV3OSQ?si=J8FgS_HP1x49qUFO&t=1

.learn 2457 (Battle Stance)
.learn 100 (Charge)
Charge the Boss
Fall through floor

### Extra Notes

_No response_

### Customizations

_No response_

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/a6f5f98a3101bc47ad89070f562774a3aaccde13

### Operating system

Ubuntu 24.04

### Modules

### Server

ChromieCraft

sogladev:
## Changes Proposed:

> When charging oversized targets (e.g. Kologarn), the charge pre-check pathfinds to the target's exact position. For targets partially over unwalkable geometry, this fails with PATHFIND_INCOMPLETE and returns SPELL_FAILED_NOPATH ("No path available") before EffectCharge ever runs.

> Instead of aborting, skip the pre-generated path for oversized targets and let EffectCharge compute the destination normally via GetFirstCollisionPosition, which produces a valid bridge-level position.

You can jump and charge to the bottom, but that may be bizzlike.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: DeepSeek V4 Pro + AzerothMCP

## Issues Addressed:

- Closes #26266

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

```cpp
- [ ] `.learn 100` (Charge) + `.go xyz 1797.15 -24.4027 448.741 603` → charge Kologarn → player stops on the bridge in front of Kologarn, does not fall through floor
- [ ] `.learn 20252` (Intercept) → same test → player stops on bridge
```

1.
2.
3.

- [ ] 
- [ ]


EricksOliveira:
I ran a test. I only found one problem: when you jump and use the charge, you land right back on the boss's legs.

src/server/game/Spells/Spell.cpp
```diff
@@ -6243,14 +6243,23 @@ SpellCastResult Spell::CheckCast(bool strict, uint32* /*param1*/, uint32* /*para
...
-                        else if (!result || m_preGeneratedPath->GetPathType() & (PATHFIND_NOPATH | PATHFIND_INCOMPLETE))
-                            return SPELL_FAILED_NOPATH;
-                        else if (m_preGeneratedPath->IsInvalidDestinationZ(target)) // Check position z, if not in a straight line
-                            return SPELL_FAILED_NOPATH;
-
-                        m_preGeneratedPath->ShortenPathUntilDist(G3D::Vector3(target->GetPositionX(), target->GetPositionY(), target->GetPositionZ()), objSize); // move back
+
+                        bool oversizedTarget = target->GetCombatReach() > NOMINAL_MELEE_RANGE && !target->IsCharmedOwnedByPlayerOrPlayer();
+
+                        bool pathFailed = (m_preGeneratedPath->GetPathType() & PATHFIND_SHORT) || !result ||
```
EricksOliveira, line 6260: // first try with raycast, if it fails fall back to normal path
bool result = m_preGeneratedPath->CalculatePath(target->GetPositionX(), target->GetPositionY(), target->GetPositionZ(), false);

bool oversizedTarget = target->GetCombatReach() > NOMINAL_MELEE_RANGE && !target->IsCharmedOwnedByPlayerOrPlayer();

bool hardFailure = (m_preGeneratedPath->GetPathType() & PATHFIND_SHORT) || !result
                || (m_preGeneratedPath->GetPathType() & PATHFIND_NOPATH);

if (hardFailure)
    return SPELL_FAILED_NOPATH;

bool pathFailed = (m_preGeneratedPath->GetPathType() & (PATHFIND_INCOMPLETE | PATHFIND_FARFROMPOLY_END))
                || m_preGeneratedPath->IsInvalidDestinationZ(target);

if (pathFailed)
{
    if (oversizedTarget)
        m_preGeneratedPath.reset();
    else
        return SPELL_FAILED_NOPATH;
}

sudlud:
Root cause checks out: the nearest navmesh poly to Kologarn's center is >7yd away, so `BuildPolyPath` ends at `PATHFIND_INCOMPLETE | PATHFIND_FARFROMPOLY_END` for a grounded caster and the pre-check bails.

Three points:

1. The repro in #26266 is jump + charge, and this doesn't cover it. While falling, `BuildPolyPath` takes the `IsFalling() && endPos.z < startPos.z` shortcut branch and returns `PATHFIND_NORMAL | PATHFIND_NOT_USING_PATH | PATHFIND_FARFROMPOLY_END`, which trips none of the `pathFailed` conditions, so the new code never runs. That matches @EricksOliveira's test. Adding `PATHFIND_FARFROMPOLY_END` to the soft set as he suggests does catch it, since that branch calls `AddFarFromPolyFlags()` too, but note it would also start failing normal-sized targets standing off-mesh.

2. Resetting on `PATHFIND_SHORT` / `PATHFIND_NOPATH` / `!result` drops path validation for every NPC with combat reach > 5 (452 of 24143 display models, before scale). When the spline's own path lookup then comes back NOPATH, `MoveSplineInit::MoveTo` falls back to a 2-point straight line, so a large mob across a gap with clear LoS becomes chargeable.

3. The fallback destination isn't safe on its own. `GetFirstCollisionPosition` is called on the *target*, so `CheckCollisionAndGetValidCoords` raycasts from the target's own off-mesh position and picks up `PATHFIND_FARFROMPOLY_START`, which isn't in the whitelist at `Map.cpp:3400`. It returns false, `MovePositionToFirstCollision` leaves the position untouched, and the charge destination becomes the target's exact center. Kologarn escapes this only because he has gravity disabled, so `notOnGround` skips that gate.

#26726 hit the same navmesh problem for pets with the same guard, but retries `CalculatePath` against the melee ring point and still fails if the retry fails. The same shape here would fix the grounded case and keep NOPATH/SHORT enforcement.

sogladev:
This could use some more testing

Rorschach91:
If the PR goal is obtaining this specific effect, then it works well.
https://github.com/user-attachments/assets/492e875b-5546-40a9-a4bb-e5e98ee54a03

Issue #26753, opened 2026-07-23
[NPC] "Orgrim's Hammer Scout" not working as intended


LoturCC:
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9928

Triage Notes: 
rep:
- be alliance 
- .go c i  33201 
- try to pull one of them and observe how they dismount on the air 
test vid: https://dl.dropboxusercontent.com/scl/fi/s1lzwnwiz8qfndnhtswr2/PI2rnCMjfN.mp4?rlkey=x7kcjlkbt733s2t4prki0jrp0&dl=0

Original Post Below: 

---
### Current Behaviour

The NPCs "Orgrim's Hammer Scout" in Icecrown behave as aggressive rather than neutral, albeit with a small pull range. On top of that they also drop their mounts and 'swim' on air while attacking you.

Also I don't know how that happened, but one of them also pulled on me while I was on the ground fighting unrelated mobs. Maybe their pull range only detects the horizontal axis?

### Expected Behaviour

Mobs should be neutral in this zone and should not swim in the air

### Official Sources (that prove or disprove) the issue being reported

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bbcad44f-4ff3-47e3-a0f4-41774b2d60d6" />

### Steps to reproduce the problem

Just walk into them or let them walk into you.
Mobs are related to the "Get the Message" below, but the quest isn't necessary to test the bugs.

https://www.wowhead.com/wotlk/quest=13314/get-the-message

### Extra Notes

Only tested on Ally side

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/ec52a3294ae653a507ead3680f6286ef8d16eafd

### Operating system

Ubuntu 24.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
- [mod-arena-replay](https://github.com/azerothcore/mod-arena-replay)
- [mod-bg-item-reward](https://github.com/azerothcore/mod-bg-item-reward)
- [mod-breaking-news-override](https://github.com/azerothcore/mod-breaking-news-override)
[...]

PR #26767, opened 2026-07-24
fix(DB/Creature): Orgrim's Hammer Scout no longer starts fights on its own

Chinoske:
## Changes Proposed:

"Orgrim's Hammer Scout" (creature 32201) attacks players on approach, even though it should be neutral (confirmed against retail).

Faction 714 is already non-hostile to Alliance - `FactionTemplate.dbc` entry 714 has `HostileMask = 0` - so this is not proximity aggro from a wrong faction.

Correcting my own earlier comment on this PR, which named the wrong function. The path is `Creature::_IsTargetAcceptable` (`Creature.cpp:2625`), the branch @sudlud pointed at from the start:

```cpp
// if the target's victim is friendly, and the target is neutral, the target is acceptable
if (targetVictim && !IsNeutralToAll() && IsFriendlyTo(targetVictim))
    return true;
```

A player fighting anything the Scout is friendly to becomes an acceptable target, which is why the original report has the player getting pulled while already in combat with unrelated mobs rather than on approach.

`SmartAI::AssistPlayerInCombatAgainst` is not involved. It is the opposite behaviour - the creature helping a player, requiring `who` to be non-player-controlled and its victim to be player-controlled - so it can never make the Scout attack a player.

That branch is reached through `CanStartAttack` from `SmartAI::MoveInLineOfSight:809`, which is behind `HasReactState(REACT_AGGRESSIVE)`, so Defensive closes it. Defensive rather than Passive because Passive never fights back: `CreatureAI::UpdateVictim` calls `AttackStop()` every tick while passive and in combat.

An earlier revision of this PR switched the faction to 1978 to match the other Orgrim's Hammer crew. That was wrong in the other direction - 1978 has `HostileMask = 2` (Alliance) - and has been reverted.

The react state applies from world load, not only after the first evade: `SmartAI::InitializeAI` ends in `JustReachedHome()`, which reaches `SmartScript::OnReset` and fires `SMART_EVENT_RESET`.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Related to #26753 (covers the unwanted aggression only - see Known Issues below for what's still open)
- Original report on the ChromieCraft tracker: chromiecraft/chromiecraft#9928

## SOURCE:

## Tests Performed:
This PR has been:

Live-tested with `REACT_DEFENSIVE`, all three cases:
- Approaching an idle Scout: no longer initiates combat.
- Fighting an unrelated mob near a Scout (the actual original repro): the Scout doesn't join in.
- Attacking a Scout directly: it retaliates normally, same as before the fix.

## How to Test the Changes:

- The original report (chromiecraft/chromiecraft#9928) also mentions mounted NPCs sometimes showing a "swimming in the air" animation glitch during combat, and a related quest mechanic (dart gun / "Get the Message") where killing them should drop them to the ground for looting. Not addressed here - this PR is scoped to the unwanted aggression only, which was independently confirmed against retail.


sudlud:
the title should always have the fix and not repeat the bug please.

sudlud:
Faction 714 is already neutral to Alliance: hostileMask 0, no enemy factions, and its faction
(66, "Horde Generic") has ReputationIndex -1 so the reputation path never applies. Friendly to
Horde, neutral to Alliance, which matches the yellow nameplates in the screenshot on the issue.

That matters because proximity aggro needs hostility, not just react state:
`Creature::CanStartAttack` (Creature.cpp:1902) -> `_IsTargetAcceptable` (Creature.cpp:2594),
true only on `IsEngagedBy || IsHostileTo` (:2614) or the assist branch (:2618). So a lone
Alliance player shouldn't be able to pull these at all, and 1978 has hostileMask 2 (Alliance),
so that experiment made the scout hostile rather than ruling faction out.

The assist branch is the one legit way they open combat: they join in on a player already
fighting something friendly to them, e.g. another scout. Can you post the exact repro? Mainly
whether you were already in combat with anything when they engaged.

Also, REACT_PASSIVE means it never fights back: `CreatureAI::UpdateVictim`
(src/server/game/AI/CreatureAI.cpp:350) calls `AttackStop()` every tick while passive and in
combat. REACT_DEFENSIVE (action_param1 = 1) is the "doesn't initiate, does retaliate" state.

Minor: `Closes #26753` would auto-close an issue this only partly covers.

Chinoske:
Good catch, fixed - title now describes the fix instead of restating the bug.

Chinoske:
You're right on both counts, thanks - dug into it more and updated the PR.

Confirmed 714 has `HostileMask=0` against Alliance in the DBC, so faction was never the issue. And found the actual mechanism: `SmartAI::MoveInLineOfSight` (`SmartAI.cpp:786-789`) gates plain proximity aggro *and* the assist branch behind the same `HasReactState(REACT_AGGRESSIVE)` check - so it's one switch, not two separate things to rule out. Went back to the original issue and the reporter actually says it themselves: "pulled on me while I was on the ground fighting unrelated mobs" - that's the assist path, not unprovoked proximity aggro.

Switched `REACT_PASSIVE` to `REACT_DEFENSIVE` per your point - same aggro-blocking effect through that code path, but it'll still retaliate if attacked directly instead of just eating hits.

Also swapped `Closes` for `Related to` on the issue link, since the mount/swim glitch and the dart-gun mechanic in that issue are still open.

Chinoske:
Closing the loop on this one after the rework: re-verified the final state with a fresh pass. Faction 714 confirmed non-hostile to Alliance in FactionTemplate.dbc (HostileMask 0), so the real trigger was the assist branch in SmartAI::MoveInLineOfSight - and both that branch and plain proximity aggro sit behind the same HasReactState(REACT_AGGRESSIVE) gate, which is why a single REACT_DEFENSIVE override closes both paths while keeping retaliation intact. All three scenarios re-tested live after the switch from PASSIVE: no aggro on approach, no joining a nearby fight, still fights back when attacked directly. Nothing further planned here.

Chinoske:
Correcting myself on this one, @sudlud, and the correction lands on your side.

My earlier comment said `SmartAI::MoveInLineOfSight` gates plain proximity aggro and the assist branch behind the same `HasReactState(REACT_AGGRESSIVE)` check. It does not. Those are two separate lines with different guards, and `AssistPlayerInCombatAgainst` is the opposite behaviour anyway: it needs `who` to be non-player-controlled and its victim to be player-controlled, so it is the creature helping a player, never attacking one.

The real path is the one you pointed at from the start, `Creature::_IsTargetAcceptable`, now at line 2625:

```cpp
if (targetVictim && !IsNeutralToAll() && IsFriendlyTo(targetVictim))
    return true;
```

It is reached through `CanStartAttack` from `MoveInLineOfSight:809`, which is behind `REACT_AGGRESSIVE`, so Defensive still closes it and the fix stands. I just had the wrong function in the writeup.

Also in this push: the file was still named `..._passive.sql` after the switch to Defensive, and its comment had grown into the entire analysis. Renamed, and the comment cut to the three lines that say what the fix does. Description updated with the correct mechanism, and the ChromieCraft report is now linked alongside #26753.

One thing I checked while I was in there, since it is the obvious question about a Reset event: the state does apply from world load, not only after the first evade. `SmartAI::InitializeAI` ends in `JustReachedHome()`, which reaches `SmartScript::OnReset` and fires `SMART_EVENT_RESET`.

Chinoske:

PR #23233, opened 2025-10-14
fix(Core/Vmaps): Fix inconsistency of hitInstance and hitModel to cause wrong area ids

PkllonG:
https://github.com/TrinityCore/TrinityCore/pull/28632
https://github.com/TrinityCore/TrinityCore/commit/b4d6ca277aef81a49e3654a3eec131ea621da991
https://github.com/TrinityCore/TrinityCore/commit/5eed7506a253fea65cef7dfb8560667defe4a564
https://github.com/TrinityCore/TrinityCore/commit/c0d8dba15a1a25756f4297eaa0647797ca13e376
Co-Authored-By: ModoX <moardox@gmail.com>
Co-Authored-By: Shauren <shauren.trinity@gmail.com>

## Changes Proposed:

This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

## Issues Addressed:

- Closes 

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [ ✔ ] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [ ] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

[...]

PR #7898, opened 2021-09-15
fix(Core/Movement): Disabled invalid code about getting area info fro…

UltraNix:
…m gameobjects.

Fixes #7335

## Issues Addressed:

- Closes #7335

## Tests Performed:

- Tested ingame on boat from BB to Rachet.

## How to Test the Changes:

More info [here](https://github.com/chromiecraft/chromiecraft/issues/1615)

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

Issue #7335, opened 2021-08-14
Unmount\Mount when traveling by ship or zeppelin


Mitradis:
### Current Behaviour

Unmount\Mount when traveling by ship or zeppelin. When you traveling by ship or zeppelin you can hear or even see how you character unmount and mount again after loading screen. Sometimes **mounting doesn't happen** and after loading screen you unmount. And **Pet sometimes is gone** after traveling.

on screen i was mounted. After traveling dismounted.
![WoWScrnShot_081521_090657](https://user-images.githubusercontent.com/21084600/129468932-bb55dbc7-3b37-409d-b966-1cd92b27ba9a.jpg)

I think it's a some hack for someting, i dont know.

### Expected Blizzlike Behaviour

No unmount\mount when traveling by ship or zeppelin. No sound. No visible effects. The same on warname.

### Steps to reproduce the problem

traveling by ship or zeppelin

### AC rev. hash/commit

https://github.com/azerothcore/azerothcore-wotlk/commit/6ff9b21b736c5bbf506a1ffa439caadb42c5dbdf

### Operating system

Windows 10 x64

lineagedr:
Pretty sure it's blizzlike

Mitradis:
"Sometimes mounting doesn't happen and after loading screen you unmount. And Pet sometimes is gone after traveling."

Azcobu:
@VFaleiro comments in https://github.com/chromiecraft/chromiecraft/issues/1615:

When running around (while mounted) ships/zeppelins, or even standing by (as soon as the ship/zeppelin begin to move), you can get randomly dismounted as if just walking through a door, but still outdoors. You can mount right after it without any problems.

Zeppelins that got me dismounted:

Tirisfal-Durotar
Tirisfal-Stranglethorn Vale
Durotar-Stranglethorn Vale
Ships that got me dismounted

Ratchet-Booty Bay
Did not test the zeppelin Durotar-Mulgore.

I was playing horde, but for Alliance it probably behaves the same way since most Alliance ships (E.G. Darkshore-Stormwind or Wetlands-Dustwallow) use the exact same model as the Ratchet-Booty Bay ship.

Have no clue about the behaviour of Alliance ships that have no indoors (E.G. Teldrassil-Darkshore or Azuremyst-Darkshore), and evidently Northrend ships/zeppelins cannot be tested (but they also have no indoors).

@Shineslippers adds:

[...]

PR #26790, opened 2026-07-25
fix(Core/Collision): make spawned WMO gameobjects provide indoor/area info

### LINKED ISSUE #5545: Wintergrasp: Player doesn't get dismounted inside the towers
## CURRENT BEHAVIOUR:
When you ride your mount inside the tower(s) in wintergrasp you don't get dismounted.
You should be dismounted.

## EXPECTED BLIZZLIKE BEHAVIOUR:
Inside the buildings of wintergrasp you should get dismounted.

Please fix this, thank you

## SOURCE:
WoWScrnShot_042921_193917

## STEPS TO REPRODUCE THE PROBLEM:

1. You go to the Wintergrasp map
2. You were standing in front of a tower
3. Ride your mount into the tower

## AC HASH/COMMIT:

rev. 7fa1cb1c1ad2+ 2021-04-29 15:07:03 +0200

## OPERATING SYSTEM:
Windows Server 2019 Standard

## MODULES:
mod-transmog
mod-cfbg
mod-duel-reset
mod-guildhouse
mod-mall-teleport
mod-dynamic-xp

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/98247820-wintergrasp-player-doesn-t-get-dismounted-inside-the-towers?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F40032087&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nyeriah:
## Changes Proposed:

Players are never dismounted inside spawned WMO gameobjects such as the Wintergrasp towers, because the server only ever computed the indoor/outdoor state from static vmaps and terrain.

Two separate problems:

1. `GameObjectModel::GetLocationInfo` never populated `hitModel` and `rootId` in the result, so `DynamicMapTree::GetAreaAndLiquidData` always returned no area info. Before #23233 the dynamic query in `Map::GetAreaInfo` went through the old `IntersectPoint` path and worked; #23233 unified everything on the `LocationInfo` path and gameobject-based area ids silently regressed, since the copy-back of `GroupLocationInfo` was only added to `ModelInstance::GetLocationInfo` (the static path) and not to the gameobject path. So this part is a regression fix. Note that the gap comes from upstream: TrinityCore's `GameObjectModel::GetLocationInfo` is missing the same assignments on both master and 3.3.5, so this PR deliberately diverges from TC here.
2. The dynamic tree query in `Map::GetFullTerrainStatusForPosition` (the function that feeds `WorldObject::IsOutdoors()` and therefore the indoor dismount in `Player::CheckAreaExploreAndOutdoor`) was commented out by #7898 because it caused wrong area data on moving transports (#7335).

This PR restores the missing assignments, re-enables the dynamic tree query, and excludes transports (type 11 and 15) from serving area info via a new `IsTransport()` on `GameObjectModelOwnerBase`. The exclusion is a deliberate trade-off, not a fix of #7335 at the source: transport WMO data is unreliable while they move, so they are kept out of area/indoor detection entirely. Boat and zeppelin cabins therefore still do not count as indoors. That matches the behavior without this PR (verified in-game: no cabin dismount on master either), so nothing changes for transports. A real cabin fix would need transport-local position handling and is out of scope.

With this, riding into a WG tower flips `IsOutdoors()` to false and the existing dismount logic fires, with no Wintergrasp-specific code. Requires `vmap.enableIndoorCheck = 1`.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code (Claude Fable 5)

## Issues Addressed:
- Closes #5545

## SOURCE:

On retail 3.3.5 players are dismounted inside the Wintergrasp tower and keep interiors (see the linked issue).

## Tests Performed:
This PR has been:

Builds without errors (MSVC 2022, Release). Tested in-game: dismount when riding into WG towers, mounting blocked inside, mounting works again outside. Boat cabins behave the same as master (no dismount).

## How to Test the Changes:

1. Make sure `vmap.enableIndoorCheck = 1` in worldserver.conf.
2. Go to Wintergrasp, mount up and ride into one of the towers or the keep buildings: you should be dismounted inside.
3. Try to mount while inside: it should fail with the "can't use that here" error.
4. Step outside and mount again: it should work immediately.

A full regression checklist is posted in the comments below. Since this touches a generic code path (area/indoor detection near all spawned WMO gameobjects), testing beyond Wintergrasp is very welcome.

- [ ] A WMO gameobject despawning from under a player does not retrigger position data until the player moves (pre-existing note in the code). In practice: after a tower is destroyed, players inside flip back to outdoors on their next movement.
- [ ] The dynamic path does not honour `VMAP_DISABLE_AREAFLAG` / `VMAP_DISABLE_LIQUIDSTATUS` while the static path does (pre-existing gap, only consequential now that the dynamic path returns data). Follow-up candidate.
- [ ] Boat/zeppelin cabins still do not count as indoors (unchanged from master, see trade-off above).

### DIFF
--- a/src/common/Collision/Models/GameObjectModel.cpp
+++ b/src/common/Collision/Models/GameObjectModel.cpp
@@ -198,7 +198,8 @@ bool GameObjectModel::intersectRay(G3D::Ray const& ray, float& MaxDist, bool Sto
 
 bool GameObjectModel::GetLocationInfo(G3D::Vector3 const& point, VMAP::LocationInfo& info, uint32 ph_mask) const
 {
-    if (!(phasemask & ph_mask) || !owner->IsSpawned() || !IsMapObject())
+    // transports provide inaccurate area info while moving (e.g. Booty Bay - Ratchet boat, see #7335)
+    if (!(phasemask & ph_mask) || !owner->IsSpawned() || !IsMapObject() || owner->IsTransport())
         return false;
 
     if (!iBound.contains(point))
@@ -216,6 +217,8 @@ bool GameObjectModel::GetLocationInfo(G3D::Vector3 const& point, VMAP::LocationI
         float world_Z = ((modelGround * iInvRot) * iScale + iPos).z;
         if (info.ground_Z < world_Z)
         {
+            info.rootId = groupInfo.rootId;
+            info.hitModel = groupInfo.hitModel;
             info.ground_Z = world_Z;
             return true;
         }
--- a/src/common/Collision/Models/GameObjectModel.h
+++ b/src/common/Collision/Models/GameObjectModel.h
@@ -42,6 +42,7 @@ public:
     virtual ~GameObjectModelOwnerBase() = default;
 
     [[nodiscard]] virtual bool IsSpawned() const = 0;
+    [[nodiscard]] virtual bool IsTransport() const = 0;
     [[nodiscard]] virtual uint32 GetDisplayId() const = 0;
     [[nodiscard]] virtual uint32 GetPhaseMask() const = 0;
     [[nodiscard]] virtual G3D::Vector3 GetPosition() const = 0;
--- a/src/common/Collision/Models/GameObjectModel.h
+++ b/src/common/Collision/Models/GameObjectModel.h
@@ -42,7 +42,7 @@ public:
     virtual ~GameObjectModelOwnerBase() = default;
 
     [[nodiscard]] virtual bool IsSpawned() const = 0;
-    [[nodiscard]] virtual bool IsTransport() const = 0;
+    [[nodiscard]] virtual bool IsTransport() const { return false; }
     [[nodiscard]] virtual uint32 GetDisplayId() const = 0;
     [[nodiscard]] virtual uint32 GetPhaseMask() const = 0;
     [[nodiscard]] virtual G3D::Vector3 GetPosition() const = 0;
--- a/src/server/game/Entities/GameObject/GameObject.cpp
+++ b/src/server/game/Entities/GameObject/GameObject.cpp
@@ -2968,6 +2968,7 @@ public:
     explicit GameObjectModelOwnerImpl(GameObject* owner) : _owner(owner) { }
 
     bool IsSpawned() const override { return _owner->isSpawned(); }
+    bool IsTransport() const override { return _owner->IsTransport(); }
     uint32 GetDisplayId() const override { return _owner->GetDisplayId(); }
     uint32 GetPhaseMask() const override { return (_owner->GetGoState() == GO_STATE_READY || _owner->IsTransport()) ? _owner->GetPhaseMask() : 0; }
     G3D::Vector3 GetPosition() const override { return G3D::Vector3(_owner->GetPositionX(), _owner->GetPositionY(), _owner->GetPositionZ()); }
--- a/src/server/game/Maps/Map.cpp
+++ b/src/server/game/Maps/Map.cpp
@@ -1389,15 +1389,15 @@ LiquidData const Map::GetLiquidData(uint32 phaseMask, float x, float y, float z,
    return liquidData;
 }
 
-void Map::GetFullTerrainStatusForPosition(uint32 /*phaseMask*/, float x, float y, float z, float collisionHeight, PositionFullTerrainStatus& data, Optional<uint8> reqLiquidType)
+void Map::GetFullTerrainStatusForPosition(uint32 phaseMask, float x, float y, float z, float collisionHeight, PositionFullTerrainStatus& data, Optional<uint8> reqLiquidType)
 {
     GridTerrainData* gmap = GetGridTerrainData(x, y);
 
     VMAP::AreaAndLiquidData vmapData;
-    // VMAP::AreaAndLiquidData dynData;
+    VMAP::AreaAndLiquidData dynData;
     VMAP::AreaAndLiquidData* wmoData = nullptr;
     _mapCollisionData.GetStaticTree().GetAreaAndLiquidData(x, y, z, reqLiquidType, vmapData);
-    // _dynamicTree.GetAreaAndLiquidData(x, y, z, phaseMask, reqLiquidType, dynData);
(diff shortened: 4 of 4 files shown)


Nyeriah:
Test checklist for anyone who wants to help. The PR has been tested in-game, but it touches a generic code path (area/indoor/liquid detection near every spawned WMO gameobject), so broader testing is appreciated.

**Core fix (Wintergrasp)**

- [x] Riding into a WG tower or keep building dismounts you a step or two inside
- [x] Mounting while inside fails with the "can't use that here" error
- [x] Mounting works again immediately after stepping outside
- [x] `.gps` inside a tower still reports the Wintergrasp zone (4197), so Tenacity/WG buffs and OutdoorPvP logic keep applying to players inside buildings
- [x] Stand inside a tower while it gets destroyed: after moving, you are flagged outdoors again and can mount
- [x] Walk tower doorways and open rooftops mounted: no dismount flapping at spots that are visually outdoors

**Transports (excluded from area info by this PR)**

- [x] Booty Bay - Ratchet boat: no dismount mid-ride, area/zone display stays correct the whole trip (this was the #7335 regression that got the old code disabled)
- [x] Zeppelins: same
- [x] Undercity and Thousand Needles elevators: ride mounted, no dismount, no area weirdness
- [x] ICC gunship encounter behaves as before

**Other spawned WMO gameobjects (this path now works for the first time)**

- [x] ICC: Lich King platform: correct area in `.gps`, not flagged indoors, no phantom swimming state (the dynamic tree can now also report liquid from gameobject models)
- [ ] Strand of the Ancients destructible buildings: dismount inside, mountable outside
- [ ] Druid Travel Form / shaman Ghost Wolf are removed inside WG towers like mounts, and unaffected outdoors next to them

**General**

- [ ] Mount/dismount behavior in static buildings (Dalaran, Orgrimmar, instances) is unchanged
- [ ] Worldserver update diffs during a full Wintergrasp battle look normal (this adds a dynamic tree point lookup to position updates, so a busy WG battle is the worst case)

Requires `vmap.enableIndoorCheck = 1`.

sudlud:
The provenance in the description is wrong: those two assignments do not exist in TrinityCore. TC 6ce6665992 introduced GroupLocationInfo and added the copy-back only to ModelInstance::GetLocationInfo, missing the gameobject path, and TC master and 3.3.5 both still have the bug. AC ported it faithfully in #23233 (6d64ecd, 2025-11-11). Nothing was lost in the port, and this fixes something still broken upstream.

"Never had any observable effect" is wrong too: before #23233 the query in Map::GetAreaInfo went through the old IntersectPoint path and worked. GO-based area ids regressed in November, so this is a regression fix, worth naming as one.

The transport exclusion is not a fix of #7335 at the source, it makes the exclusion permanent, so boat and zeppelin cabins will never be indoors while chromiecraft/chromiecraft#1615 expects a dismount when actually inside. Fine as a trade-off, just word it that way.

Checked and fine: Arthas Platform (202161) and Precipice (202078) are type 33, so the exclusion does not touch them; destroyed buildings zero the model phasemask via EnableCollision(false), so rubble stops reporting indoors.

Minor: a defaulted IsTransport() { return false; } avoids breaking out-of-tree implementers of GameObjectModelOwnerBase. And, pre-existing but only now consequential, DynamicVMapCollisionData ignores VMAP_DISABLE_AREAFLAG/LIQUIDSTATUS while the static one honours them. CodeRabbit's 120 column comment is a false positive, master's line is 177 chars and yours is 173.

What decides this for me is .gps inside the towers. The dynamic path passes adtId 0 into GetWMOAreaTableEntryByTripple, so it either misses and falls back to the grid area or matches an unrelated adtId 0 row. Wintergrasp has to keep zone 4197 for everyone inside.

Nyeriah:
Thanks for digging into this, you're right on all counts.

Provenance: confirmed against TC 3.3.5 and master, the assignments are indeed missing in their gameobject path too, and the old `IntersectPoint` path here worked until #23233 unified everything on `LocationInfo`. I rewrote the description to frame that part as a regression fix and to note the deliberate divergence from TC.

Transports: reworded as a permanent trade-off rather than a fix of #7335. I also checked in-game on master without this PR: there is no dismount inside the Booty Bay boat cabin today either, so the exclusion just preserves current behavior. A proper cabin fix needs transport-local position handling and stays out of scope, noted in the TODO list along with the `VMAP_DISABLE_AREAFLAG`/`LIQUIDSTATUS` gap.

`IsTransport()` is now a defaulted virtual returning false, so out-of-tree owners keep compiling.

Re the WMOAreaTable lookup with adtId 0: agreed this is the deciding check. I'll test `.gps` inside the towers and post the zone/area values here.

Nyeriah:
As promised, `.gps` results.

Inside a WG tower with this PR:

```
You are indoors
Map: 571 (Northrend) Zone: 4197 (Wintergrasp) Area: 4581 (Flamewatch Tower)
X: 4459.013 Y: 1945.6403 Z: 440.05728
GroundZ: 434.61557 FloorZ: 440.05353 Have height data (Map: 1 VMap: 1 MMap: 1)
```

Zone stays 4197, so Tenacity/OutdoorPvP are unaffected. The adtId 0 lookup does not miss or hit an unrelated row: it resolves the tower's own WMOAreaTable entry (nameset 0) and we get the proper named subarea, Flamewatch Tower, which links back to zone 4197. FloorZ comes from the tower model as expected.

Inside the Booty Bay boat cabin on master (without the PR):

```
You are outdoors
Map: 0 (Eastern Kingdoms) Zone: 33 (Stranglethorn Vale) Area: 1578 (Southern Savage Coast)
GroundZ: -24.288818 FloorZ: 6.0986805
```

Outdoors in the cabin, area from the terrain grid, deck height still served by the collision path. Since this PR excludes transports from area info, this is exactly what it preserves, no behavior change on boats.

That should cover the open points. Thanks again for the thorough review.

Gultask (approved):

PR #26763, opened 2026-07-23
fix(DB/Creature): The Avalanche sub-Zone Improvements Part 1: Bythius the Flesh-Shaper, Cerberon and Glonn

Voxstrasza:
1. apply waypoints for Bythius the Flesh-Shaper (Entry 28212, GUID 100735) and mark as patrol leader
2. remove Cerberon (Entry 28207, GUID 100431) & Glonn (Entry 28211, GUID 100704) wander
3. apply formations for plague-dogs
4. remove summon Cerberon and Glonn from Bythius SAI due to lack of evidence that Bythius summons on retail and to avoid fighting 4 during the quest now that they are linked via formation

## Changes Proposed:
-  [ ] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).

### AI-assisted Pull Requests
- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Claude Opus 4.8 used to parse sniffs and consolidate/clean-up SQL file
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## SOURCE:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [x] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:
.go c 100735

PR #26795, opened 2026-07-25
fix(Core/Updater): fail the dry run when a SQL update fails to apply

sudlud:
## Changes Proposed:

This PR proposes changes to:

Plus the Dashboard CI workflow.

On #26763, a `data/`-only PR, the world DB import failed:

```
ERROR 1054 (42S22) at line 8: Unknown column 'Entry' in 'where clause'
Applying of file '.../pending_db_world/rev_1784847243540513500.sql' to database 'acore_world' failed!
```

Every check still passed, so a broken SQL file was mergeable:
https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30093358328/job/89481482864

`DBUpdater<T>::ApplyFile` deliberately does not throw during a dry run, so one run reports every
broken file instead of stopping at the first. That part is intentional (#22610). The problem is that
nothing recorded the failure, so `UpdateFetcher` wrote the broken file into the `updates` table as
applied, and `worldserver -dry-run`, `authserver -dry-run` and `dbimport` all still exited 0. Every
job that validates SQL through a dry run was blind to it.

Changes:

- `DBUpdaterUtil` gets a process-wide failure counter. `ApplyFile` marks a failure and returns
  whether the file applied. The `isDryRun()` suppression stays exactly where #22610 put it, so
  `Populate` and both update loops keep walking their file lists, across all three databases, and
  every failure still gets its own `LOG_FATAL`.
- `UpdateFetcher` skips `UpdateEntry` for a file that failed, so a later start in the same job
  retries it instead of treating it as done.
- The three dry-run exit points report the count and return non-zero.
- Dashboard CI is currently the only SQL validation on a `data/`-only PR and it trusts a single exit
  code, so the dry-run steps also grep their output for a failed import. A final step feeds the
  updater a deliberately broken file and requires exit code 1, so a regression here cannot silently
  reopen the hole.

Note for downstream: any pipeline with steps sequenced after a `worldserver --dry-run` will now stop
at the dry run when an import fails, instead of running on.

### AI-assisted Pull Requests

## Issues Addressed:

- Closes

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Put a deliberately broken file in `data/sql/updates/pending_db_world/`, run `worldserver -dry-run`,
   confirm exit code 1 and that `SELECT * FROM acore_world.updates WHERE name = '<file>'` returns no row.
2. Add a second broken file in `pending_db_characters` and confirm both `LOG_FATAL` lines appear in one
   run and the terminating line reports 2 failed updates. This is the #22610 check: the run must
   continue past a failure both to the next file and to the next database.
3. Break a base file under `data/sql/base/db_world/` and confirm the dry run still walks the rest of
   `Populate`'s file list and still exits non-zero.
4. Remove the broken files and confirm exit code 0 and unchanged normal behaviour.
5. Confirm a non-dry-run start with a broken file still aborts as before, including
   `Updates.ExceptionShutdownDelay`.

- [ ] Dashboard CI runs zero-conf, so it has no `Logger` config and writes no `Errors.log`. Any future
      `ci-error-check.sh`-style step there would be a no-op. Separate problem, worth a follow-up.


sudlud:
@coderabbitai review

Gultask (approved):

sudlud:
Claude found new issues on this PR, will look into it tomorrow.

1. `DBUpdater.cpp:288` and `:438-450`: base file failures escape the per-database
   reporting. `DatabaseLoader::Load()` populates all three pools before it updates them,
   and in a dry run `Populate()` no longer stops on a bad base file, so its failures are
   counted before the `failedBefore` snapshot in `Update()` and get subtracted out again.
   A dry run against an empty `acore_world` where one `data/sql/base/db_world/*.sql` file
   fails still logs ">> World database is up-to-date!" for a database that was never
   populated. The exit code is still 1, so the gate holds, but the summary contradicts
   the failure right above it. `CheckUpdateTable()`'s base file apply has the same problem.

2. `UpdateFetcher.cpp:381-382`: not recording a failed file means a partially applied file
   fails on a different error the next run. The mysql CLI runs without `--force`, so the
   statements before the error are already committed. A file that adds a column and then
   hits a bad statement will fail on "Duplicate column name" the second time, which hides
   the real defect and cannot succeed without fixing the database by hand. Still better
   than recording it as applied, but it should be called out.

3. `DBUpdater.cpp:324-382`: the `setDirectories` overload of `Update()` takes the new
   callback but has none of the new reporting, so a failing module SQL file exits non-zero
   without a summary saying which database it came from. The two overloads now differ.

4. `DBUpdater.h:96-97`: the comment says the counter is only touched in a dry run, but
   `MarkUpdateFailed()` is called before the `isDryRun()` check, so it also runs on the
   throwing path. Harmless today, wrong as documentation.

5. `dashboard-ci.yml:309-339`: the probe step leaves `rev_9999999999999999999.sql` behind
   and only a comment keeps it last in the job. An `rm` or a `trap` after the assertions
   would make that self-enforcing.

Helias (approved):

PR #26801, opened 2026-07-25
fix(Core/Instances): instances now reset after the server was down

### LINKED ISSUE #26741: Instance lockouts are not reset when a global reset elapses while the server is offline
## Current Behaviour

When a heroic dungeon or raid global reset falls due while the worldserver is offline (for example the server is shut down over the nightly reset hour), the missed reset is never applied. On the next startup the stored reset time is advanced to the future, but the existing player lockouts (`character_instance` binds) and their `instance` rows are left in place. Affected players stay locked to those instances past the intended reset: a daily heroic lockout can persist well over 24 hours, and a weekly raid lockout up to a full extra week.

A visible symptom of this: after such an offline reset, a player who logs in sees the calendar / raid lockout showing a reset time that has effectively already passed, so it reads as expiring almost immediately, but the lockout never actually clears. Only a later event that re-derives the reset time (a server restart, or a relog) shifts the displayed expiry forward to the next cycle, roughly a day after the original.

This only happens in the offline case. When a reset falls due while the server is running it is handled correctly by the runtime path (`InstanceSaveMgr::Update` -> `_ResetOrWarnAll` -> `_ResetSave`).

## Expected Behaviour

A global reset that falls due while the server is offline should be applied on the next startup: the affected instance lockouts should be cleared (with extended lockouts surviving one extra cycle, as they do during a normal reset), exactly as if the reset had happened while the server was up. The calendar should show the correct next reset time rather than a past one that never triggers.

## Source

Standard WotLK instance lockout behaviour: heroic and raid lockouts expire on their fixed daily/weekly schedule at the configured reset hour, independent of server uptime. A lockout that expires during downtime is gone when the realm comes back.

## Steps to reproduce the problem

1. Bind a character to a heroic (or raid) instance so a `character_instance` row exists.
2. Stop the worldserver.
3. In the characters database, set that map/difficulty's `instance_reset.resettime` to a time in the past, simulating a reset that fell due while the server was down.
4. Start the worldserver.
5. Observe that the character is still locked to the instance: the `instance` and `character_instance` rows remain and the lockout persists past the reset that should have cleared it. In game, the calendar shows the lockout as expiring immediately but it does not clear.

## Extra Notes

Root cause: in `InstanceSaveMgr::LoadResetTimes()`, the branch that handles an already-expired global reset time carries the comment "assume that expired instances have already been cleaned" and only advances the reset time forward. That assumption holds for normal dungeons but not for heroics and raids:

- The only startup cleanup of expired instances is in `LoadInstances()`: `DELETE FROM instance WHERE resettime > 0 AND resettime < UNIX_TIMESTAMP()`.
- Heroic and raid saves store `resettime = 0` in the `instance` table (`InstanceSave::GetResetTimeForDB()` returns 0 for raids and heroic difficulty), so that DELETE never touches them.
- As a result, nothing at startup clears heroic/raid lockouts whose global reset lapsed while offline, and the runtime `_ResetOrWarnAll`/`_ResetSave` path never ran for them because the server was down when the reset was due.

The calendar symptom has the same origin: the expired branch advances `instance_reset` but does not refresh the in-memory reset time, so a loaded save keeps its old (past) reset time. The client shows that past time as an imminent expiry, and because no reset fires it never clears; a later restart re-reads the advanced time and the displayed expiry jumps forward a cycle.

Prior report and history: this was reported before, in #966 (2018), and closed as completed by #1416 in February 2019. That fix only addressed part of the problem. Before #1416, the expired branch used a `while (t < now)` loop; #1416 replaced it, which resolved the symptom where the reset time appeared to extend to login time on every login or restart. But it left the underlying offline case unhandled (this is where the "assume that expired instances have already been cleaned" comment was added), so heroic/raid lockouts whose reset lapsed offline are still not cleared. In the same change it also introduced a separate reset-hour drift bug in that branch (the no-op `(t * DAY) / DAY`), which is addressed independently in #26711. Notably, a commenter on #966 (`a4501150`) diagnosed this exact root cause at the time: that `LoadResetTimes` assumes the server is running when the reset time is reached and does not handle the server starting up after a missed reset.

Possible approaches (raising this for discussion, would appreciate input on the preferred direction before opening a PR):

1. Clean the lapsed lockouts directly in the expired branch of `LoadResetTimes` with SQL (delete non-extended binds, clear the extended flag on the rest, delete now-unbound instances and their respawn rows). Simplest and most local, but it duplicates the teardown that `_ResetSave` already performs and adds raw SQL.
2. Reuse `_ResetSave` after saves and binds are loaded: have `LoadResetTimes` record which map/difficulties lapsed (or detect loaded saves whose reset time is now in the past) and run `_ResetSave` over them later in `LoadInstances`. Avoids duplicating the teardown, but the hand-off is a bit indirect.
3. Treat the missed reset as simply overdue: in `LoadResetTimes`, schedule an immediate reset event for the lapsed map/difficulty and let the normal `Update` -> `_ResetOrWarnAll` -> `_ResetSave` path apply it on the first world update, by which point the saves are loaded. No duplication and it reuses the exact runtime machinery. The reset then happens a tick after startup rather than synchronously, and it wants a small change to `_ResetOrWarnAll` so a gap spanning several reset periods catches up in one step.

Each has trade-offs, and I would rather agree on a direction here before writing it up as a PR. Happy to put one together whichever way the maintainers prefer.

## AC rev. hash/commit

Output of `.server debug`:

```
AzerothCore rev. 62a991e783b3+ 2026-06-26 11:50:23 -0700 (Playerbot branch) (Unix, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 151.
Connection peak: 1.
Server uptime: 22 minute(s) 41 second(s)
|- Mean: 7ms
|- Median: 2ms
|- Percentiles (95, 99, max): 22ms, 29ms, 33ms
Using SSL version: OpenSSL 3.0.13 30 Jan 2024 (library: OpenSSL 3.0.13 30 Jan 2024)
Using Boost version: 1.83.0
Using CMake version: 3.28.3
Using MySQL version: 80046
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 5.15.153.1-microsoft-standard-WSL2
MMAPs status: Enabled
maps directory located in /azerothcore/env/dist/data/maps. Total size: 291014951 bytes
vmaps directory located in /azerothcore/env/dist/data/vmaps. Total size: 657797442 bytes
mmaps directory located in /azerothcore/env/dist/data/mmaps. Total size: 2192760616 bytes
Default DBC locale: enUS.
Using Playerbots DB Revision:
Latest CharacterDatabase update: playerbots_names.sql
Latest WorldDatabase update: rev_1784047766052542100.sql
LoginDatabase queue size: 0
WorldDatabase queue size: 0
PlayerbotsDatabase queue size: 0
List of enabled modules:
|- mod-ah-bot-plus
|- mod-playerbots
```

This is from a server running the Playerbot branch (rev `62a991e783b3`). The affected code in `InstanceSaveMgr::LoadResetTimes` is identical on upstream master (currently `9a9924ed4`) and the bug reproduces there with no modules loaded.

## Operating system

Platform independent (logic bug in core). Observed on Windows 11 x64 running the server in Docker (WSL2), but not OS specific.

## Custom changes or Modules

`mod-playerbots` and `mod-ah-bot-plus` are loaded on the server the debug output above is from, but neither is related. The bug is in core instance-save handling (`InstanceSaveMgr`) and reproduces on a clean AzerothCore with no modules.

Chinoske:
## Changes Proposed:

When a heroic dungeon or raid global reset falls due while the worldserver is offline, the missed reset was never applied. On the next startup the stored reset time got silently advanced to the future, but existing player lockouts (`character_instance` binds and their `instance` rows) stayed in place - affected players remained locked to those instances past the intended reset, sometimes for a full extra cycle.

Root cause: heroic and raid saves store `resettime = 0` in the `instance` table (by design - only normal-difficulty instances track a per-save reset time), so the startup cleanup query (`DELETE FROM instance WHERE resettime > 0 AND resettime < UNIX_TIMESTAMP()`) never touches them. Meanwhile `InstanceSaveMgr::LoadResetTimes()`, on finding an already-expired global reset time, only recalculated the *next* reset time and moved on - nothing ever actually applied the lapsed reset.

Fix: when `LoadResetTimes()` finds a global reset time that's already in the past, it now queues an immediate, non-warning reset event (`InstResetEvent` type 5 - the same type already used for a live in-session reset, just never triggered from the startup path before) instead of silently recalculating past it. This fires on the very first `Update()` tick after saves and binds are loaded, so `_ResetOrWarnAll`/`_ResetSave` run the exact same teardown a live reset does - unbinding non-extended players, deleting now-unbound instances, and correctly refreshing both the DB and in-memory reset time (which also fixes the related calendar symptom, where the client showed a lockout as expiring immediately but it never actually cleared).

Also hardened `_ResetOrWarnAll`'s next-reset-time calculation to advance by more than one period when needed, so a server that was offline across multiple reset cycles catches up to a genuinely future time in one step instead of landing on a time that's still in the past.

No changes to the runtime reset path (`_ResetOrWarnAll`/`_ResetSave` themselves are only extended, not altered) - this only changes what happens at startup for a reset that was missed while offline.

**Not a regression of #26711:** that PR fixed a day-alignment no-op (`t = (t * DAY) / DAY;` -> `t = (t / DAY) * DAY;`) inside the same expired-branch block this PR removes entirely, replacing the whole inline recalculation with the `InstResetEvent(5, ...)` hand-off described above. `_ResetOrWarnAll`'s own reset-time formula (`(resetTime + MINUTE) / DAY * DAY`) is a separate, already-correct calculation that #26711 never touched, so #26711's fix isn't lost - the branch it applied to just no longer exists in this version.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #26741

## SOURCE:

Code inspection: Traced the flow: `InstanceSaveMgr::LoadInstances` -> `LoadResetTimes` -> (new) queues an immediate `InstResetEvent(5, ...)` for any lapsed map/difficulty -> `LoadInstanceSaves`/`LoadCharacterBinds` load the affected saves/binds -> first `InstanceSaveMgr::Update()` tick (called every world update from `World::Update`) processes the queued event since its key time is in the past -> `_ResetOrWarnAll(warn=false)` -> `_ResetSave` per matching instance save, same path a live reset takes. Confirmed `InstResetEvent::type` already documents "5 - reset" in its own header comment, so this reuses an existing, already-load-bearing mechanism rather than introducing a new one.

## Tests Performed:
This PR has been:

Reproduced the bug live pre-fix, following the issue's own repro steps: bound a character to a heroic instance, stopped worldserver, set that map/difficulty's `instance_reset.resettime` to 1 hour in the past, restarted. Pre-fix, the bind stayed. Rebuilt with the fix and repeated the exact same steps: after restart, the `character_instance`/`instance` rows for that bind were gone (query returned empty), and `instance_reset.resettime` had advanced to a correct future date (one full period ahead, landing after "now") instead of staying in the past or jumping incorrectly.

Also live-tested the extended-lock edge case sudlud flagged: bound a character to Utgarde Keep heroic (map 574, difficulty 1) with `extended=1`, forced that map/difficulty's `instance_reset.resettime` ~8 days into the past (more than 2 reset periods), rebuilt with the two-pass fix, restarted. After startup, both `character_instance` and `instance` rows for that bind were completely gone (not just `extended` cleared) - confirms the double-pass teardown correctly carries an extended lock through both stages in one startup, and `instance_reset.resettime` landed on a real future date.

## How to Test the Changes:

1. Bind a character to a heroic (or raid) instance so a `character_instance` row exists for it.
2. Stop the worldserver.
3. In the characters database, set that map/difficulty's `instance_reset.resettime` to a time in the past (simulating a reset that fell due while offline): `UPDATE instance_reset SET resettime = UNIX_TIMESTAMP() - 3600 WHERE mapid = <mapid> AND difficulty = <difficulty>`.
4. Start the worldserver.
5. Confirm the character's `character_instance`/`instance` rows for that bind are gone (or, for an extended bind, that its lockout correctly advanced instead of persisting), and that `instance_reset.resettime` for that map/difficulty is now a real future date.

- None

🤖 Generated with Claude Code (Opus 5)+AzerothMCP

### DIFF
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -589,11 +589,12 @@ void InstanceSaveMgr::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, bool 
 
         uint32 next_reset = uint32(((resetTime + MINUTE) / DAY * DAY) + period + diff);
         // catch up in one step if the server was offline across more than one reset period.
-        // _ResetSave() below still only needs to run once regardless of how many periods were
-        // skipped: `extended` is a single-shot flag, not a per-period counter, so one pass
-        // still correctly drops it before the actual unbind on the following normal reset.
+        bool missedMultiplePeriods = false;
         while (next_reset <= uint32(now))
+        {
             next_reset += period;
+            missedMultiplePeriods = true;
+        }
         SetResetTimeFor(mapid, difficulty, next_reset);
         SetExtendedResetTimeFor(mapid, difficulty, next_reset + period);
         ScheduleReset(time_t(next_reset - 3600), InstResetEvent(1, mapid, difficulty));
@@ -607,11 +608,21 @@ void InstanceSaveMgr::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, bool 
 
         // remove all binds to instances of the given map and delete from db (delete per instance id, no mass deletion!)
         // do this after new reset time is calculated
-        for (InstanceSaveHashMap::iterator itr = m_instanceSaveById.begin(), itr2; itr != m_instanceSaveById.end(); )
+        // extended locks need a second pass to fully clear: PlayerUnbindInstanceNotExtended
+        // (called from _ResetSave) only flips `extended` to false on the first call and
+        // unbinds on the next one, matching a real reset one period later. When more than
+        // one period was missed while offline, run the teardown twice so extended players
+        // end up unbound here too, instead of just losing the extension - `extended` is a
+        // single-shot flag, so two passes are enough no matter how many periods were skipped.
+        uint8 resetPasses = missedMultiplePeriods ? 2 : 1;
+        for (uint8 pass = 0; pass < resetPasses; ++pass)
         {
-            itr2 = itr++;
-            if (itr2->second->GetMapId() == mapid && itr2->second->GetDifficulty() == difficulty)
-                _ResetSave(itr2);
+            for (InstanceSaveHashMap::iterator itr = m_instanceSaveById.begin(), itr2; itr != m_instanceSaveById.end(); )
+            {
+                itr2 = itr++;
+                if (itr2->second->GetMapId() == mapid && itr2->second->GetDifficulty() == difficulty)
+                    _ResetSave(itr2);
+            }
         }
     }
 
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -588,7 +588,10 @@ void InstanceSaveMgr::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, bool 
             period = DAY;
 
         uint32 next_reset = uint32(((resetTime + MINUTE) / DAY * DAY) + period + diff);
-        // catch up in one step if the server was offline across more than one reset period
+        // catch up in one step if the server was offline across more than one reset period.
+        // _ResetSave() below still only needs to run once regardless of how many periods were
+        // skipped: `extended` is a single-shot flag, not a per-period counter, so one pass
+        // still correctly drops it before the actual unbind on the following normal reset.
         while (next_reset <= uint32(now))
             next_reset += period;
         SetResetTimeFor(mapid, difficulty, next_reset);
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -352,11 +352,15 @@ void InstanceSaveMgr::LoadResetTimes()
 
         if (t < now)
         {
-            // assume that expired instances have already been cleaned
-            // calculate the next reset time
-            t = (t / DAY) * DAY;
-            t += ((today - t) / period + 1) * period + diff;
-            CharacterDatabase.DirectExecute("UPDATE instance_reset SET resettime = '{}' WHERE mapid = '{}' AND difficulty = '{}'", (uint32)t, mapid, difficulty);
+            // The reset fell due while the server was offline. Queue an immediate, non-warning
+            // reset event (type 5) using the lapsed time instead of silently advancing past it -
+            // heroic/raid saves store resettime = 0 in `instance`, so nothing else ever clears
+            // their lockouts for a reset that was missed (see issue #26741). This fires on the
+            // very first Update() tick, by which point LoadInstanceSaves()/LoadCharacterBinds()
+            // have loaded the affected saves and binds, so _ResetOrWarnAll runs the same teardown
+            // a live reset does, including refreshing the in-memory/DB reset time correctly.
+            ScheduleReset(t, InstResetEvent(5, mapid, difficulty));
+            continue;
         }
 
         SetExtendedResetTimeFor(mapid, difficulty, t);
@@ -584,6 +588,9 @@ void InstanceSaveMgr::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, bool 
             period = DAY;
 
         uint32 next_reset = uint32(((resetTime + MINUTE) / DAY * DAY) + period + diff);
+        // catch up in one step if the server was offline across more than one reset period
+        while (next_reset <= uint32(now))
+            next_reset += period;
         SetResetTimeFor(mapid, difficulty, next_reset);
         SetExtendedResetTimeFor(mapid, difficulty, next_reset + period);
         ScheduleReset(time_t(next_reset - 3600), InstResetEvent(1, mapid, difficulty));
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -609,7 +609,8 @@ void InstanceSaveMgr::_ResetOrWarnAll(uint32 mapid, Difficulty difficulty, bool 
         uint8 resetPasses = missedMultiplePeriods ? 2 : 1;
         for (uint8 pass = 0; pass < resetPasses; ++pass)
         {
-            for (InstanceSaveHashMap::iterator itr = m_instanceSaveById.begin(), itr2; itr != m_instanceSaveById.end(); )
+            for (InstanceSaveHashMap::iterator itr = m_instanceSaveById.begin(), itr2;
+                 itr != m_instanceSaveById.end(); )
             {
                 itr2 = itr++;
                 if (itr2->second->GetMapId() == mapid && itr2->second->GetDifficulty() == difficulty)
--- a/src/server/game/Instances/InstanceSaveMgr.cpp
+++ b/src/server/game/Instances/InstanceSaveMgr.cpp
@@ -352,13 +352,9 @@ void InstanceSaveMgr::LoadResetTimes()
 
         if (t < now)
         {
-            // The reset fell due while the server was offline. Queue an immediate, non-warning
-            // reset event (type 5) using the lapsed time instead of silently advancing past it -
-            // heroic/raid saves store resettime = 0 in `instance`, so nothing else ever clears
-            // their lockouts for a reset that was missed (see issue #26741). This fires on the
-            // very first Update() tick, by which point LoadInstanceSaves()/LoadCharacterBinds()
-            // have loaded the affected saves and binds, so _ResetOrWarnAll runs the same teardown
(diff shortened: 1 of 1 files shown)


Chinoske:
The failing Linux CI checks here (clang/gcc build jobs) aren't caused by this PR - they're a pre-existing compile error on `master` in `boss_skeram.cpp` from #26789 (unrelated to this change). Fix proposed in #26803. These checks should pass once that's merged and re-run.

LucidTaZ:
As a solo player that turns off the server when I'm not using it: thank you! This fix really helps me. 🙏 

One concern I'd like to raise is around extended locks: if I'm not mistaken, a player can extend their instance lock, and rather than resetting, only the extension is removed during a reset round. Only in the next reset round, the instance actually resets.

This means the multiple reset catch up code should be adjusted to accommodate this, or at least a comment should be added because admittedly it's an edge case.

Chinoske:
Good catch, but I don't think it needs adjusting - `_ResetSave()` only runs once per `_ResetOrWarnAll()` call regardless of how many periods the catch-up loop skips, and that's actually correct for extended locks: the `extended` flag is a single-shot transition (extended -> not-extended on one reset, then actually unbound on the *next* reset), not something that accumulates per missed period. So even if the server was offline across 3 reset periods, this catch-up call represents crossing exactly one meaningful boundary for bind state - extended players go to not-extended here, and unbind on the following normal reset, same as if only one period had elapsed.

Added a comment near the `while` loop explaining this for future readers: 4d222d53

sudlud:
The extended-lock point LucidTaZ raised still stands, and the comment at `InstanceSaveMgr.cpp:591-596` states the opposite.

`PlayerUnbindInstanceNotExtended` (`InstanceSaveMgr.cpp:748`) clears `extended` and keeps the bind, so dropping an extended lock always takes two reset passes. The catch-up loop advances `next_reset` by N periods but still runs `_ResetSave` once, so after a downtime spanning two or more cycles an extended player only loses the extension and stays bound until `T0 + N * period`, where a server that stayed up would have unbound them at `T0 + 2 * period`.

Since `extended` is single-shot, two passes are always enough. Running the teardown a second time when more than one period lapsed covers it, no unbounded replay needed.

Chinoske:
@sudlud you're right, I got this wrong the first time. Traced it through again: `_ResetSave` -> `PlayerUnbindInstanceNotExtended` only clears `extended` on the first call, the actual unbind (in-memory erase + DB delete) only happens on a second call once it's already not-extended. So one pass after a multi-period gap only drops the extension, it doesn't unbind.

Pushed a fix: when the catch-up loop advances `next_reset` by more than one period, the reset teardown now runs twice instead of once. Since `extended` is single-shot, two passes fully resolve it no matter how many periods were actually missed - matches what you described. Also swapped the wrong comment for one that actually explains it: 7456c142

sudlud:
you  might want to try opus instead of sonnet and also a higher effort setting - this might ensure that PRs will be right on the first try. it is also useful to let the AI review your PR again before posting it :)

sudlud:
i can also recommend this toolkit https://github.com/eai-org/agent-toolkit

so right after you've prepare a PR like you're now doing, you could e.g. use the skill `/fresh-eyes-review` of the toolkit.
This usually leads to better results in the end and can also be repeated multiple times until you've got a solid implementation for the PR.

So all in all this could lead to PRs beeing merge-ready when you send them.

Chinoske:
@sudlud took both suggestions - this PR (and the rest of my open ones) just went through another full pass with the upgraded setup, including an isolated fresh-context review over the final diff.

That re-review confirmed the current state end to end: the two-pass teardown fully unbinds an extended lock after a multi-period gap (traced through the actual prepared statements - pass 1 clears `extended` in memory and DB, pass 2 unbinds and deletes the now-empty `character_instance`/`instance` rows), the catch-up loop's period counting has no off-by-one against the `t < now` detection in `LoadResetTimes`, and single-period gaps still correctly do one pass only.

One asymmetry it flagged, for the record: the new `continue` in `LoadResetTimes` skips the `SetExtendedResetTimeFor` call the non-expired branch does, so between load and the first `Update()` tick the in-memory extended reset time for that map/difficulty is stale. It's not reachable - the queued reset fires on the very first tick, before any session can log in and query it, and every surviving save gets the value refreshed inside that same call - so I left it alone rather than adding code for a window that can't be observed. Flagging it here in case a reviewer disagrees with that call.

Chinoske:
Wrapped the reset loop, it was one column over the limit after the extra nesting.

Chinoske:

PR #26811, opened 2026-07-25
fix(Spells): generic ghost-charge window for charge-based procs

### LINKED ISSUE #25117: (Mage]  Fingers of Frost interactions

Chinoske:
## Changes Proposed:

Fingers of Frost's last charge gets consumed by the same Frostbolt that used it, so a queued Deep Freeze right after loses the guaranteed-crit/frozen-target bonus even though it should still count as "frozen" for that split second.

Two prior attempts at this, both closed:
- #26210 changed the proc's consumption phase (`PROC_SPELL_PHASE_CAST` -> `HIT`), which was the wrong direction - CAST-phase consumption is correct per 3.3 patch notes, closed same-day.
- #26216 fixed it for Fingers of Frost specifically with a spell-only hack (extend the buff's duration by 400ms on removal). Reviewed and video-confirmed working, but stalled: the maintainer asked whether this should be generalized through `spell_proc` instead of hardcoded per spell, since the same timing gap affects other charge-based procs (Divine Favor + Holy Shock is the same story). Closed unresolved after no follow-up.

This PR does that. Added a `ChargeDropDelay` column to `spell_proc` (ms). When a charge-based proc aura runs out of charges, instead of removing it immediately, the aura now stays up for `ChargeDropDelay` ms if that value is set - long enough for an instant cast issued right as the last charge lands to still catch it. It can't proc again in that window (out-of-charges already blocks new procs), so nothing double-dips.

Applied it to two unrelated cases as proof it's genuinely reusable:
- Fingers of Frost (74396): already had a `spell_proc` row, just added the delay.
- Divine Favor (20216): had no row at all before this (ran on the DBC-derived auto-generated proc entry). Added one that mirrors exactly what the engine currently auto-generates for it, plus the new delay - so this is a pure DB addition, zero C++.

Reverted the spell-specific 400ms hack from `spell_mage.cpp`'s Fingers of Frost script, keeping only the unrelated self-consumption guard (a Frostbolt that procs FoF shouldn't also consume the charge it just created - that's a separate bug from the ghost-charge one).

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #25117

## SOURCE:

## Tests Performed:
This PR has been:

## How to Test the Changes:

Frost Mage: Frostbolt-spam until Fingers of Frost is down to its last charge, then queue Deep Freeze right as that last-charge Frostbolt lands - Deep Freeze should still guarantee the frozen-target effect.

Paladin: pop Divine Favor, cast Holy Shock immediately - both effects should get the guaranteed crit, same as before this PR (this case has zero C++ changes, it's here to prove the DB mechanism generalizes).

Not personally verified in-client yet - built and deployed cleanly (no spell_proc load errors), opening this now for others to test both combos since that's hard to verify solo.

### DIFF
--- a/data/sql/updates/pending_db_world/2026_07_25_02_fingers_of_frost_charge_drop_delay.sql
+++ b/data/sql/updates/pending_db_world/2026_07_25_02_fingers_of_frost_charge_drop_delay.sql
@@ -1,6 +1,7 @@
--- Adds a generic "ghost charge" window for charge-based proc auras: keeps the aura
--- alive for ChargeDropDelay ms after its last charge is consumed instead of removing
--- it immediately, so an instant cast already queued in the same batch can still use it.
+-- Adds a generic "ghost charge" grace window for charge-based proc auras: keeps the
+-- aura alive for ChargeDropDelay ms after its last charge is consumed instead of
+-- removing it immediately, so an instant cast issued right as the last charge lands
+-- can still benefit from it.
 ALTER TABLE `spell_proc` ADD COLUMN `ChargeDropDelay` INT UNSIGNED NOT NULL DEFAULT 0 AFTER `Charges`;
 
 -- 74396 - Fingers of Frost (Mage): already has a row, just add the delay.
--- a/data/sql/updates/pending_db_world/2026_07_25_02_fingers_of_frost_charge_drop_delay.sql
+++ b/data/sql/updates/pending_db_world/2026_07_25_02_fingers_of_frost_charge_drop_delay.sql
@@ -1,0 +1,16 @@
+-- Adds a generic "ghost charge" window for charge-based proc auras: keeps the aura
+-- alive for ChargeDropDelay ms after its last charge is consumed instead of removing
+-- it immediately, so an instant cast already queued in the same batch can still use it.
+ALTER TABLE `spell_proc` ADD COLUMN `ChargeDropDelay` INT UNSIGNED NOT NULL DEFAULT 0 AFTER `Charges`;
+
+-- 74396 - Fingers of Frost (Mage): already has a row, just add the delay.
+UPDATE `spell_proc` SET `ChargeDropDelay` = 400 WHERE `SpellId` = 74396;
+
+-- 20216 - Divine Favor (Paladin): had no row (ran purely on the DBC-derived auto-generated
+-- proc entry). Values below mirror exactly what SpellMgr::LoadSpellProcs() auto-generates
+-- for it today, so this only adds ChargeDropDelay without changing existing proc behavior.
+DELETE FROM `spell_proc` WHERE `SpellId` = 20216;
+INSERT INTO `spell_proc`
+    (`SpellId`, `SchoolMask`, `SpellFamilyName`, `SpellFamilyMask0`, `SpellFamilyMask1`, `SpellFamilyMask2`, `ProcFlags`, `SpellTypeMask`, `SpellPhaseMask`, `HitMask`, `AttributesMask`, `DisableEffectsMask`, `ProcsPerMinute`, `Chance`, `Cooldown`, `Charges`, `ChargeDropDelay`)
+VALUES
+    (20216, 0, 10, 3223322624, 0, 0, 81920, 7, 2, 0, 8, 0, 0, 100, 0, 1, 400);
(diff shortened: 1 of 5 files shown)


Chinoske:
Update:

Thanks for the review, went through all of it.

Fixed:
- The uninitialized `ChargeDropDelay` catch was real and the important one - `SpellProcEntry` has no default member initializers, and the auto-generated branch (spells with no explicit `spell_proc` row) never set that field, so it was reading indeterminate stack memory for every one of those ~1300 entries. Zeroed it explicitly now.
- Wrapped the lines that went over 120 columns (`SpellMgr.h`, the new validation log line, and the `Register()` call in `spell_mage.cpp`).

Pushing back on two:
- The `PROC_ATTR_USE_STACKS_FOR_CHARGES` path (`ModStackAmount(-1)`) still removes immediately regardless of `ChargeDropDelay` - true, and deliberate for this PR. Neither Fingers of Frost nor Divine Favor use that attribute, so nothing currently configured hits it. Extending it safely needs its own guard against re-proc while the aura is artificially held at "spent," which the charges path gets for free from an existing check (`IsUsingCharges() && !GetCharges()`) that has no equivalent for stacks yet - didn't want to bolt that on unverified in the same PR.
- The prepared-statement suggestion for the `LoadSpellProcs` query - this file has 18 other `Load*` functions using raw `WorldDatabase.Query()` the same way, all one-time startup config loaders with no user input involved. Converting just this one would be inconsistent with everything around it for no real benefit; happy to revisit if there's an appetite for converting the whole file.

On the `_freshlyApplied` gating in `spell_mage.cpp`: that state machine isn't new here, it's carried over unchanged from #26216 (already reviewed and video-confirmed working by @bonfax4 in that thread). This PR only touched `OnRemove` in that class. Worth a fresh look on its own if anyone hits a concrete repro, but treating it as out of scope for this diff.

Kitzunu:
I will close this as it is a hackfix and just slop from Sonnet.

Adding one column for one spell is the peak definition of a hackfix.
You're burning many credits on PRs using Sonnet, which is a way worse model to use for coding. My recomendation to you -> Use Opus.

Chinoske:
im using opus no sonnet

Kitzunu:
image

Nontheless is this a hackfix

Chinoske:
All the fixes I've had with Sonnet have been fine, and you apparently consider it a hackfix. I've had the same results as everyone else. With Opus, I wouldn't change anything; besides, I'm using AzerothMCP with added tools, not just the basics that the repo has.

So the other fixes that got merged were hackfixes too, then?

Chinoske:
You should revert the ones that were merged then if they are hackfixes, since they were done by Sonnet

Kitzunu:
I didnt say the other ones was? This PR is a hackfix for the reason I said above

> Adding one column for one spell is the peak definition of a hackfix

Kitzunu:
You have other PRs that are great, and your contributions are appreciated. Dont get me wrong :)

Chinoske:
I think you should have tested it before closing it, but oh well...

Kitzunu:
Im sure it works, but the fix is not accepted as it is a hackfix.
There needs to be a broader or more generic fix. Because as I said we can't add a new column to the database just to fix one spell, becuase that means there is something else that should be done instead.

We had the exact same thing with another guy who submitted a similar hackfix for something else.

bonfax4:
I'm sorry but wouldnt the previous fix before the proc system overhaul also be considered an hackfix ? The proper fix for this would require for spellbatching mechanics to be added which will never happen. Ghost charge is a major mechanic for frost play style in pvp and pve, some exceptions should be considered.

Or you know remove all the artificial batching mechanics added to ignite and deep wounds as well to be consistent

PR #26816, opened 2026-07-26
fix(Core/Player): .learn all my class no longer grants the other faction's spells

Chinoske:
## Changes Proposed:

`.learn all my class` hands a character spells restricted to the other faction. An Alliance paladin ends up with the Blood Elf warhorse (60%) in their mount list, though not the Blood Elf charger (100%) - that asymmetry is the tell.

`Player::IsSpellFitByClassAndRace` returns true for any spell with no `SkillLineAbility` entry of its own. Spells that exist purely to teach another spell have no skill line, so the race mask on what they actually teach was never looked at:

- 34767 (Blood Elf charger) is a trainer spell in its own right and carries `RaceMask 0x200`, so it was correctly skipped.
- 34768 is only a wrapper whose `SPELL_EFFECT_LEARN_SPELL` teaches 34769 (Blood Elf warhorse). The wrapper has no skill line, so it sailed through - and the restriction on 34769 never got a chance to apply.

Fixed by following the learn chain when a spell has no skill line of its own, so the restriction on what it teaches is what decides. Depth-capped at 3 hops since nothing in the data prevents a cycle. Only `Trainer::SendSpells` / `Trainer::GetSpellState` call this, so the blast radius is the trainer list and trainer teaching - both of which should have been filtering this way already.

The quest path needed its own fix: `HandleLearnAllMyQuestSpells` only checked `SatisfyQuestClass`. Adding `SatisfyQuestRace` would not have helped, because the paladin mount quests (9712, 9737) are set to `AllowableRaces` 1791 - every race. In normal play nobody completes the wrong faction's version because the quest giver sits in the other faction's city, not because the data forbids it. So that path now checks the reward spell through the same helper.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- None

## SOURCE:

Code inspection: Traced the two paths that teach spells here against the DBC data - `SkillLineAbility.dbc` has no entry for the wrapper spells (34766, 34768, 13820, 23215) while the spells they teach carry proper race masks (0x200 for Blood Elf, 0x405 for Human/Dwarf/Draenei).

## Tests Performed:
This PR has been:

Ran `.learn all my class` on a fresh Alliance paladin and a fresh Blood Elf paladin. Each now gets only their own faction's 60% and 100% mounts; before the change the Alliance one also got the Blood Elf 60%.

## How to Test the Changes:

On a fresh Alliance paladin: `.char level 80`, then `.learn all my class`, then open the mount tab. Only the Alliance warhorse and charger should be there - no "Thalassian Warhorse".

Same on a Blood Elf paladin: they should get their own two and none of the Alliance ones.

Worth a regression check on a normal (non-GM) paladin buying mounts from their own class trainer, since the same helper backs the trainer list.

- None


Chinoske:
Re-reviewed from scratch. The interesting half is the IsSpellFitByClassAndRace change: spells with no skill-line entry used to be waved through unconditionally, and the pure-teaching spells (paladin mount rewards) are exactly that shape - their race restriction lives on the taught spell, so the recursion (depth-capped at 3 against data loops) judges them by what they teach. Re-checked the callers of IsSpellFitByClassAndRace to make sure the stricter answer doesn't regress normal learning paths: everything else that calls it passes spells that DO have skill-line entries, so the new branch only ever fires for the teach-only shells. The cs_learn guard on GetRewSpellCast() (> 0 check, since the field can carry sentinel values) closes the actual reported hole - Alliance paladins collecting Blood Elf mounts. No further changes.

Chinoske:

PR #26818, opened 2026-07-26
fix(Core/Combat): taunt now puts the taunter on top of the threat list

### LINKED ISSUE #26775: [Class][Paladin] Hand of Protection + any Taunt - delay on receiving the threat
### Current Behaviour

Taunting with Hand of Protection (BOP) does not instantly aggro.

https://www.youtube.com/watch?v=2TGeFDlNzIY

Strangely, it seems to work instantly if you taunt with a 2nd taunt simultaneously.

https://www.youtube.com/watch?v=tMwPR0jdWXQ&list=PLKU0gMWd1c4ejijBB61Zpwdjna8w9J9_l&index=19

### Expected Behaviour

Taunting with Hand of Protection (BOP) should instantly aggro without delay.

### Source

Videos above

- be level 80 with 1 Paladin and 1 other class with 2 Taunt-Abilities (2nd Paladin works e.g.)
- `.cheat god`
- engage any creature with the 1st paladin
- use BOP on the not tanking person 
- use Taunt as the BOPed person and observe having a delay on receiving the aggro
- repeat the process, but now use 2 taunts (e.g. as paladin Righteous Defense + Hand of Reckoning) simultanously and observer receiving INSTANTLY the threat/aggro

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. 172877a35d1b 2026-07-23 21:52:15 +0200 (master branch) (Unix, RelWithDebInfo, Static)

### Operating system

Windows 11

### Custom changes or Modules

Chinoske:
## Changes Proposed:

Taunting a mob while immune to its damage (Hand of Protection is the reported case, but any damage immunity applies) has a noticeable delay before the mob actually switches target, even though `ThreatManager::TauntUpdate()` is supposed to force an immediate reselect.

Root cause: `ThreatManager::EvaluateSuppressed()` iterates `_threatenedByMe` (who the mob itself threatens - normally empty) instead of `_myThreatListEntries` (who threatens the mob - where the taunter's reference actually lives). `TauntUpdate()` calls it specifically to un-suppress a reference the instant it starts taunting, per `ShouldBeSuppressed()`'s own rule that "a taunting victim can never be suppressed" - but scanning the wrong list means that never happens.

The full sequence for issue #26775: taunting a mob you've never engaged creates a brand-new `ThreatReference` during the taunt spell's `SPELL_EFFECT_TAUNT` effect (which runs before the `SPELL_AURA_MOD_TAUNT` aura effect that actually marks it as taunting). At creation, `ShouldBeSuppressed()` sees the caster is immune to the mob's melee damage school and marks the reference `SUPPRESSED`. The taunt aura then applies and `TauntUpdate()` runs, correctly marking the taunt state - but `EvaluateSuppressed()`'s wrong-list bug means the `SUPPRESSED` flag is never cleared. Since online/suppressed state is compared *before* taunt state when picking a victim, the suppressed-but-taunting reference loses to any other online, non-taunting entry (the original tank) until something unrelated (LOS change, another combat event) forces a correct re-evaluation later - that's the reported delay.

Confirmed this isn't an AzerothCore porting bug: TrinityCore's own `3.3.5` branch has the identical `_threatenedByMe` in `EvaluateSuppressed()`, so it's inherited from upstream.

Fix is a one-line list swap plus a comment explaining why, so the same mistake doesn't happen again.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #26775

## SOURCE:

Code inspection: Traced the exact sequence through Spell::EffectTaunt / ThreatManager::MatchUnitThreatToHighestThreat / AuraEffect::HandleModTaunt / ThreatManager::TauntUpdate / EvaluateSuppressed / ThreatManager::ReselectVictim's CompareReferencesLT ordering (online state compared before taunt state). Cross-checked against TrinityCore's 3.3.5 branch source for the same function - identical bug present there.

## Tests Performed:
This PR has been:

Reproduced the exact steps from the issue: paladin tanking a `.cheat god` mob, Hand of Protection on a second paladin who had not engaged that mob, then Hand of Reckoning from the protected paladin. Before the fix: delayed target switch, matching the report. After the fix: instant switch.

## How to Test the Changes:

Two level 80 characters, one able to tank and one with a taunt ability (a second Paladin works well, per the issue). `.cheat god` on a training dummy or any creature. Tank engages it. Cast Hand of Protection on the other character (make sure they haven't attacked that creature yet, so they have no existing threat reference). Have them Taunt (Hand of Reckoning / Righteous Defense / any class's taunt) - the creature should switch to them instantly, no delay.

- None

### DIFF
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -532,12 +532,9 @@ void ThreatManager::TauntUpdate()
         else
             pair.second->UpdateTauntState();
 
-        // A reference that just started taunting can never be suppressed
-        // (ShouldBeSuppressed() special-cases IsTaunting()) - but EvaluateSuppressed()
-        // below only refreshes _threatenedByMe (what WE threaten), not this list (what
-        // threatens US), so a reference that was created SUPPRESSED (e.g. the taunter was
-        // already immune to our damage before taunting) never gets cleared there. Do it
-        // here directly instead, right after its taunt state is set.
+        // A taunting reference can never be suppressed, but EvaluateSuppressed below only
+        // refreshes what we threaten, not what threatens us, so one created suppressed would
+        // never be cleared. Do it here instead.
         if (pair.second->IsSuppressed() && !pair.second->ShouldBeSuppressed())
         {
             pair.second->_online = ThreatReference::ONLINE_STATE_ONLINE;
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -373,7 +373,10 @@ bool ThreatManager::IsThreateningTo(Unit const* who, bool includeOffline) const 
 
 void ThreatManager::EvaluateSuppressed(bool canExpire)
 {
-    for (auto const& pair : _threatenedByMe)
+    // Re-evaluates entries on OUR OWN threat list (who is threatening us), not _threatenedByMe
+    // (who we threaten - unrelated). TauntUpdate() relies on this to un-suppress a reference the
+    // instant it starts taunting, e.g. a target that was immune to our damage before taunting.
+    for (auto const& pair : _myThreatListEntries)
     {
         bool const shouldBeSuppressed = pair.second->ShouldBeSuppressed();
         if (pair.second->IsOnline() && shouldBeSuppressed)
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -373,10 +373,7 @@ bool ThreatManager::IsThreateningTo(Unit const* who, bool includeOffline) const 
 
 void ThreatManager::EvaluateSuppressed(bool canExpire)
 {
-    // Re-evaluates entries on OUR OWN threat list (who is threatening us), not _threatenedByMe
-    // (who we threaten - unrelated). TauntUpdate() relies on this to un-suppress a reference the
-    // instant it starts taunting, e.g. a target that was immune to our damage before taunting.
-    for (auto const& pair : _myThreatListEntries)
+    for (auto const& pair : _threatenedByMe)
     {
         bool const shouldBeSuppressed = pair.second->ShouldBeSuppressed();
         if (pair.second->IsOnline() && shouldBeSuppressed)
@@ -534,6 +531,18 @@ void ThreatManager::TauntUpdate()
             pair.second->UpdateTauntState(it->second);
         else
             pair.second->UpdateTauntState();
+
+        // A reference that just started taunting can never be suppressed
+        // (ShouldBeSuppressed() special-cases IsTaunting()) - but EvaluateSuppressed()
+        // below only refreshes _threatenedByMe (what WE threaten), not this list (what
+        // threatens US), so a reference that was created SUPPRESSED (e.g. the taunter was
+        // already immune to our damage before taunting) never gets cleared there. Do it
+        // here directly instead, right after its taunt state is set.
+        if (pair.second->IsSuppressed() && !pair.second->ShouldBeSuppressed())
+        {
+            pair.second->_online = ThreatReference::ONLINE_STATE_ONLINE;
+            pair.second->HeapNotifyIncreased();
+        }
     }
 
     // taunt aura update also re-evaluates all suppressed states (retail behavior)
--- a/src/server/game/Combat/ThreatManager.cpp
+++ b/src/server/game/Combat/ThreatManager.cpp
@@ -532,13 +532,21 @@ void ThreatManager::TauntUpdate()
         else
             pair.second->UpdateTauntState();
 
-        // A taunting reference can never be suppressed, but EvaluateSuppressed below only
-        // refreshes what we threaten, not what threatens us, so one created suppressed would
-        // never be cleared. Do it here instead.
+        // Re-derive suppression for our own threat list. EvaluateSuppressed below only
+        // refreshes what we threaten, not what threatens us, so a reference created
+        // suppressed would never be cleared - which is what kept a taunt from landing.
+        // Both directions, because the usual reason it stops being suppressed is the taunt
+        // itself: leave out the second half and an immune tank stays on top after the taunt
+        // has already expired, holding the boss on a target it cannot damage.
         if (pair.second->IsSuppressed() && !pair.second->ShouldBeSuppressed())
         {
             pair.second->_online = ThreatReference::ONLINE_STATE_ONLINE;
             pair.second->HeapNotifyIncreased();
+        }
+        else if (pair.second->IsOnline() && pair.second->ShouldBeSuppressed())
+        {
+            pair.second->_online = ThreatReference::ONLINE_STATE_SUPPRESSED;
+            pair.second->HeapNotifyDecreased();
         }
     }
 


Chinoske:
Update:

CI caught a real problem with my first commit - thanks, this is exactly what that suite is for. `EvaluateSuppressed()` iterating `_threatenedByMe` isn't a bug, it's the deliberate contract for re-evaluating suppression from the *threatening* unit's own perspective (across everything it threatens), and `ThreatManagerIntegrationTest.EvaluateSuppressed_*` calls it exactly that way. My first fix swapped that list, which broke those 5 tests.

Pushed a correction: reverted `EvaluateSuppressed()` back to `_threatenedByMe`, and moved the actual fix into `TauntUpdate()` instead. `TauntUpdate()` calls `EvaluateSuppressed()` on the taunted creature's own manager, wanting to un-suppress a reference in *that* manager's `_myThreatListEntries` (the taunter) - not its outgoing `_threatenedByMe` (almost always empty for a mob). Now it checks and clears suppression directly in the loop that already sets taunt state, right after each reference's taunt state updates, instead of relying on `EvaluateSuppressed()` to do something it was never meant to do from that side.

All the GCC/Clang builds that failed before are green now, which includes the full test suite. Re-testing the original in-game repro against this corrected version next.

Chinoske:
Fresh re-review of this one. Re-traced the flow: ShouldBeSuppressed() special-cases IsTaunting() so a taunting reference can never legitimately stay suppressed, but EvaluateSuppressed() only walks _threatenedByMe (what the taunter threatens), never the list holding the reference that just started taunting - so a reference born suppressed (taunter was damage-immune at creation) stayed offline forever. The fix clears it exactly where the taunt state is set, mirroring what EvaluateSuppressed does for the other direction, including the HeapNotifyIncreased() so the victim re-sort happens. Also re-checked the earlier CI failure that shaped this version - the current approach doesn't touch EvaluateSuppressed's iteration at all, so that regression path is gone. Nothing further to change.

Chinoske:
**UPDATE.** Ran `/self-review` on this and it caught something I had missed, so there is one more commit.

The un-suppress only went one way. The reason a taunting reference stops being suppressed is `ShouldBeSuppressed()` returning false *because it is taunting* — a three second state — and nothing put it back afterwards. Traced the paths that could: `HandleModTaunt` ignores its `apply` flag so `TauntUpdate()` does run on removal, but by then the ref is ONLINE and the new branch no longer matched; `EvaluateSuppressed(true)` walks `_threatenedByMe`, which is empty for a boss; every other call site passes the default `canExpire = false` and only fires on aura apply; and `UpdateOffline()` early-returns for a non-offline ref.

So in the issue's own scenario: BoP goes up, tank taunts and it works, taunt expires three seconds later, and the reference stays ONLINE for the remaining seven seconds of BoP with the tank still immune. Top threat on a target the boss cannot damage, which is the thing SUPPRESSED exists to prevent. Worse with Divine Shield or Ice Block, and unbounded for a permanent immunity.

Worth being clear that master can already reach this through the un-suppress in `AddThreat`, gated on the same predicate — that is why the reporter's "two taunts works instantly" workaround works, since Hand of Reckoning deals damage. So this was not introduced here, but the one-way version made it the guaranteed outcome of every taunt rather than a rare accident. The mirror branch closes both.

Also rewrote the comment. It said "a taunting reference", but the condition has no taunt test in it and never did: it re-derives suppression for the whole incoming list. That is deliberate and matches the `EvaluateSuppressed(true)` line just below it, but the comment described something narrower than the code, which would have misled the next reader.

Two things I did not change and would rather hear an opinion on first:

The block probably belongs in `ThreatReference::UpdateTauntState`, where `_taunted` actually changes. It would early-return on `state == _taunted` instead of calling `ShouldBeSuppressed()` for every reference on every update, and `SetTauntStateForTesting` bypasses `TauntUpdate()` so the existing test harness cannot reach the code where it currently sits.

And this branch has a commit whose body is retracted by a later one, so it wants squashing before merge either way.

Happy to move it and add a test if that is the direction you would prefer.

Chinoske:
The `/self-review` report itself, which I should have posted with the update above rather than only the prose.

# Self-review — fix/taunt-suppressed-threat-reference → master

> **Outcome** 3 findings — 2 fixed, 1 left open for the maintainer — final round not re-reviewed
> **Intent** Taunt did not reliably put the taunter on top of the threat list. A threat reference created in a suppressed state stayed suppressed even while taunting, because the sweep that clears suppression only looks at what the creature threatens, not at what threatens it.
> **Reviewed** `289800514` (`fix/taunt-suppressed-threat-reference` vs `origin/master`, merge-base diff — 1 file, +17 −0)
> **By** Claude Code (VS Code extension, Claude Agent SDK), claude-opus-5 — self-review v0.1, 2026-08-09 — review by claude-opus-5, as reported by the reviewer subagent
> **Project rules** `.agents/docs/self-review-rules.md` present and applied. Note: `.agents/docs/cpp-guidelines.md` does **not** exist in this worktree — `.agents/` holds only `README.md` and `skills/`, so the style rules were checked against the self-contained `AGENTS.md` and `.editorconfig` instead.
> **In-game testing (author)** not recorded for this round. Suggested checks before merge are listed at the bottom; the one that matters most is watching the three seconds *after* a taunt expires while an immunity is still up.
> **Caveat** the two fixes landed in `289800514` were not put through a second review round, so they are unreviewed code. The reviewed SHA is the pushed head.

## Round 1 — `4dc5a589b`, 3 findings

1. `src/server/game/Combat/ThreatManager.cpp:538` — the transition ran one way only. `ShouldBeSuppressed()` returns false *because the reference is taunting*, a three-second state, and nothing put it back afterwards: `HandleModTaunt` ignores its `apply` flag so `TauntUpdate()` does run on removal, but by then the ref is ONLINE and the branch no longer matches; `EvaluateSuppressed(true)` walks `_threatenedByMe`, empty for a boss; every other call site passes the default `canExpire = false` and fires only on aura apply; `UpdateOffline()` early-returns for a non-offline ref. Concrete failure, the issue's own scenario: BoP up, tank taunts, taunt expires after 3 s, the reference stays ONLINE for the remaining ~7 s of BoP with the tank still immune — top threat on a target the boss cannot damage, which is what SUPPRESSED exists to prevent. Unbounded for a permanent immunity → **fixed** (mirror branch added: `IsOnline() && ShouldBeSuppressed()` → `ONLINE_STATE_SUPPRESSED` + `HeapNotifyDecreased()`)
2. `src/server/game/Combat/ThreatManager.cpp:535` — the comment said "a taunting reference can never be suppressed", but the condition contains no taunt test and re-derives suppression for the whole incoming list, on every taunt and detaunt, apply and removal. That is deliberate and consistent with the `EvaluateSuppressed(true)` line below it, but it overrides the rule stated at `:460` (*"SUPPRESSED threat states don't go back to ONLINE until threat is caused by them"*) which two existing tests encode, and a maintainer reading only the comment would mispredict the blast radius → **fixed** (comment rewritten to describe the whole-list behaviour and why both directions are needed)
3. `src/server/game/Combat/ThreatManager.cpp:538` — the block probably belongs in `ThreatReference::UpdateTauntState`, where `_taunted` actually changes. Three consequences of the current placement: `SetTauntStateForTesting` bypasses `TauntUpdate()`, so the existing test harness cannot reach this code; `ShouldBeSuppressed()` runs for every reference on every update instead of behind `UpdateTauntState`'s `state == _taunted` early return, and it walks the immunity list plus two aura-type lookups; and it is the only place both directions can be driven off a single `_taunted` change → **open**: moving it changes the shape of the file, so asked the maintainer for a direction first rather than reworking it twice. Offered to move it and add a test.

## Verified clean by the fresh-eyes review

- **The direct `_online` write is consistent with the file.** `ThreatManager` is a `friend` of `ThreatReference` (`ThreatManager.h:322`), no setter exists, and `EvaluateSuppressed` (`:381`, `:386`) and `AddThreat` (`:464`) write it the same way.
- **The heap notifications are correct in both directions.** `CompareReferencesLT` (`:699`) keys on `_online` first, higher is greater, so `SUPPRESSED(1) → ONLINE(2)` strictly increases and the reverse strictly decreases. No corruption path found; `_online` and `_taunted` always move together.
- **The pre-existing hole is acknowledged, not invented here.** `AddThreat`'s un-suppress at `:461` is gated on the same predicate, which is why the reporter's "two taunts works instantly" workaround works — Hand of Reckoning deals damage. The one-way version turned a rare accident into the guaranteed outcome of every taunt.
- **No iterator invalidation**: neither `ShouldBeSuppressed()` nor the heap notifications touch `_myThreatListEntries`.
- **An ONLINE reference that should be OFFLINE self-corrects** through `UpdateVictim()` → `ReselectVictim()` → `UpdateOffline()`.
- **Commit message claims check out.** `TauntUpdate()` is verbatim identical in TrinityCore's 3.3.5 branch, including `EvaluateSuppressed(true)` over `_threatenedByMe`; the existing tests do call it on the victim; issue #26775 exists and matches, workaround included.
- **Style**: added lines measured by hand, longest 93 columns, under the 120 limit; `apps/codestyle/codestyle-cpp.py` passes all checks. No debug prints, dead code, TODOs or stray files.

## History note

The branch carries a commit (`c4f22e30e`) whose body asserts that `EvaluateSuppressed` "looks at the wrong list", explicitly retracted by a later one (`797fcc734`). Both are on the branch, so the merged history would contain a commit body refuting its predecessor. Squash before merge.

## Still to test in game

- The issue's repro: a BoP'd player taunts, aggro should be instant.
- **The side effect worth the most attention**: keep watching the three seconds *after* the taunt expires while BoP is still up, and diff against master. This is finding 1.
- A BoP'd tank holding top threat: taunt, let it expire, confirm the boss switches to a killable target before BoP ends.
- Suppression stickiness (finding 2): a mage Ice Blocks and unblocks without acting, then an unrelated player taunts the same mob — does the mage's threat suddenly count again?
- Multi-taunt (Righteous Defense plus Hand of Reckoning): `TauntUpdate` assigns `TauntState(state++)` past the enum, so several references read as taunting at once.
- The detaunt path, which also routes through `TauntUpdate`.
- Breakable-CC targets (Blind, Repentance, Sap) on a mob while someone taunts it — they must stay SUPPRESSED while the CC holds.
- A boss whose whole threat list is offline must still evade.

Chinoske:

PR #26827, opened 2026-07-26
fix(Core/Creature): don't charge corpse decay time on forced despawn of a living creature

Nyeriah:
## Changes Proposed:

When a creature is despawned while still alive and no explicit respawn timer is supplied (`DespawnOrUnsummon()` with no second argument, or `SMART_ACTION_FORCE_DESPAWN` with `forceRespawnTimer = 0`, which is the vast majority of uses), the dynamic (non compatibility) spawn path charges the creature's corpse decay time on top of `spawntimesecs`, even though the creature is removed outright and never leaves a corpse. Affected spawns respawn 60s late (normal rank) or 300s late (rare, elite, rare elite).

Cause: `setDeathState(JustDied)` folds `m_corpseDelay` into `m_respawnTime` on the assumption that a corpse will exist. `ForcedDespawn` already compensates for this, but only when an explicit `forceRespawnTimer` is given. With no timer the stale value survives, because the dynamic mode branch of `RemoveCorpse` clamps with `std::max` and always picks it. Compatibility mode assigns the respawn time directly in `RemoveCorpse` and has never charged the decay in this situation, so the two modes diverge.

Fix: apply the existing "ignore corpse decay time" correction whenever the creature was alive at the moment of the forced despawn, not only when a forced timer was supplied. Since `m_respawnDelay` is already overridden with the forced timer earlier in the function, both cases reduce to the same assignment.

Not affected: normal deaths, `DespawnOnEvade` (always passes a respawn timer of at least 2s), any call with an explicit timer, temporary summons (routed to `TempSummon::UnSummon`), and compatibility mode, which recomputes the respawn time itself and is bit for bit unchanged. A creature that was already a corpse when force despawned also keeps its current behaviour, since a corpse did exist and charging the decay is correct there.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code (Fable 5)

## Issues Addressed:
- None. Behavioural divergence between dynamic and compatibility respawn mode introduced by the spawn system port (https://github.com/azerothcore/azerothcore-wotlk/pull/25206).

## SOURCE:

Validated against the compatibility mode code path (`Creature::RemoveCorpse`, legacy branch), which drops the corpse decay in the same situation. This PR restores parity between the two modes.

## Tests Performed:
This PR has been:

Codestyle checks pass.

## How to Test the Changes:

1. Ensure `Respawn.ForceCompatibilityMode = 0` (default) in worldserver.conf.
2. Pick a non-elite DB creature with a known `creature.spawntimesecs`, e.g. 30.
3. Give it a SmartAI script using `SMART_ACTION_FORCE_DESPAWN` with all action parameters at 0 (or despawn it via any script calling `DespawnOrUnsummon()` with no respawn timer).
4. Trigger the despawn and check the remaining respawn time:
   ```sql
   SELECT spawnId, respawnTime - UNIX_TIMESTAMP() AS seconds_left
   FROM creature_respawn WHERE spawnId = <spawnId>;
   ```
5. Before the fix `seconds_left` is `spawntimesecs + 60` (or + 300 for rare/elite). After the fix it matches `spawntimesecs`.
6. Regression checks: kill a creature normally (corpse decay must still delay the respawn as before), force despawn with an explicit respawn timer (timer must be honoured as before), and repeat the steps with `Respawn.ForceCompatibilityMode = 1` (behaviour must be unchanged).

- [ ] The same `std::max` clamp in the dynamic mode branch of `RemoveCorpse` also defeats `Respawn.DynamicRateCreature` scaling (it raises the shortened delay back to the full one). Out of scope here, no effect under the default configuration.

### DIFF
--- a/src/server/game/Entities/Creature/Creature.cpp
+++ b/src/server/game/Entities/Creature/Creature.cpp
@@ -2172,14 +2172,17 @@ void Creature::ForcedDespawn(Milliseconds timeMSToDespawn, Seconds forceRespawnT
     if (forceRespawnTimer > 0s)
         m_respawnDelay = forceRespawnTimer.count();
 
-    if (IsAlive())
+    bool const wasAlive = IsAlive();
+
+    if (wasAlive)
         setDeathState(DeathState::JustDied, true);
 
     // Xinef: Set new respawn time, ignore corpse decay time...
-    // After setDeathState, m_respawnTime includes m_corpseDelay which we don't
-    // want for a forced respawn. Override it so RemoveCorpse's max() picks ours.
-    if (forceRespawnTimer > 0s)
-        m_respawnTime = GameTime::GetGameTime().count() + forceRespawnTimer.count();
+    // setDeathState(JustDied) folds m_corpseDelay into m_respawnTime, but a creature
+    // despawned while alive never leaves a corpse, so the decay must not be charged.
+    // Recompute so RemoveCorpse's max() has nothing stale to pick up.
+    if (forceRespawnTimer > 0s || wasAlive)
+        m_respawnTime = GameTime::GetGameTime().count() + m_respawnDelay;
 
     RemoveCorpse(true);
 


Rorschach91 (approved):

Rorschach91:
Tested on various creatures with Force Despawn and with forceRespawnTimer = 0. Their respawn time is now exactly the same as their spawntimesecs parameter.

Issue #24216, opened 2025-12-26
Title: Players are randomly losing items (equipped + inventory) — getting worse


kambire:
### Current Behaviour

Hi AzerothCore team,

A few days ago I reported that some players were losing items. At the beginning it was rare and only happened to a small number of users, but now it’s becoming unsustainable.

Players are randomly losing 2, 3, or even 4 items, including items they have been wearing for a long time. It’s not limited to one type of item — it can be any item (equipped or from the inventory). The frequency seems to be increasing and it’s seriously affecting the server.

Sorry for the channel, but I really need help from someone if possible, because this is getting out of control.

### Expected Behaviour

no data

### Steps to reproduce the problem

no data

### AC rev. hash/commit

.server debug
AC> AzerothCore rev. 0bf023cb6e12+ 2025-12-24 21:03:57 -0300 (master branch) (Unix, Release, Static)
Connected players: 87. Characters in world: 82.
Connection peak: 90.
Server uptime: 10 hour(s) 19 minute(s) 15 second(s)
Update time diff: 2ms. Last 500 diffs summary:
|- Mean: 16ms
|- Median: 3ms
|- Percentiles (95, 99, max): 61ms, 72ms, 82ms
Using SSL version: OpenSSL 3.0.2 15 Mar 2022 (library: OpenSSL 3.0.2 15 Mar 2022)
Using Boost version: 1.74.0
Using CMake version: 3.22.1
Using MySQL version: 80407
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 5.15.0-163-generic
Worldserver listening connections on port 12500
Realmlist (Realm Id: 1) configured in port 12500
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
maps directory located in ./maps. Total size: 291014951 bytes
vmaps directory located in ./vmaps. Total size: 657798358 bytes
mmaps directory located in ./mmaps. Total size: 2192761336 bytes
Default DBC locale: enUS.
All available DBC locales: enUS
Using World DB: ACDB 335.14-dev
Latest LoginDatabase update: acore_cms_subscriptions.sql
Latest CharacterDatabase update: trasmorg.sql
Latest WorldDatabase update: trasm_world_VendorItems.sql
LoginDatabase queue size: 0
CharacterDatabase queue size: 0
WorldDatabase queue size: 0
List of enabled modules:
|- mod-account-mounts
[...]

PR #26834, opened 2026-07-27
fix(Core/Database): make player save statements idempotent

EricksOliveira:
Add `ON DUPLICATE KEY UPDATE` to four prepared statements used by the player save path in `CharacterDatabase.cpp`:

- `CHAR_INS_CHAR_SPELL` → `specMask`
- `CHAR_INS_CHAR_TALENT` → `specMask`
- `CHAR_INS_CHAR_SKILLS` → `value`, `max`
- `CHAR_INS_CHAR_ACTION` → `action`, `type`

No schema change, no behaviour change on the success path.

`Player::SaveToDB()` builds a single transaction covering inventory, position,
spells, talents, skills, actions and more. If any statement fails,
`MySQLConnection::ExecuteTransaction()` rolls back the entire transaction.
Only `1213` and `1205` are retried; `1062` is not, so the transaction is
silently discarded.

`_SaveSpells()`, `_SaveTalents()`, `_SaveSkills()` and `_SaveActions()` reset
their in-memory state to `UNCHANGED` at enqueue time, before knowing whether
the transaction committed. `_SaveSpells()` additionally erases
`PLAYERSPELL_REMOVED` entries from `m_spells`. After a rollback, memory and DB
are permanently out of sync, and nothing reconciles them.

Any subsequent re-learn — level up, spec switch, default skills, racials — then
finds the spell absent from memory and flags it `PLAYERSPELL_NEW`, producing a
plain `INSERT` over a row that still exists. That raises `1062`, which aborts
the transaction again. The failure is self-sustaining.

Because `_SaveInventory()` is appended to the same transaction *before*
`_SaveSpells()`, a duplicate key on `character_spell` discards the inventory
writes as well. `_SaveInventory()` clears `m_itemUpdateQueue` unconditionally
and `Item::SaveToDB()` resets item state, so the affected items are never
written by any later save either.

`SaveToDB()` also clears `m_additionalSaveTimer` / `m_additionalSaveMask`,
cancelling the pending partial `SaveInventoryAndGoldToDB()` save. The
lightweight safety net is disarmed by the same save that then fails.

Observed in production on a server with ~87 concurrent players: a duplicate
entry on `character_spell` causes the character to lose items acquired during
the session and to log back in at the position of the last successful save.

Related: #24216

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Opus5

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Enable SQL logging in `worldserver.conf`:
```
   Appender.SQL=2,1,7,SQL.log,w
   Logger.sql.sql=1,SQL
```
2. With a character online, insert a row for a spell it does not know:
   `INSERT INTO character_spell (guid, spell, specMask) VALUES (<guid>, 2018, 200);`
3. In game: `.learn 2018` — the spell is flagged `PLAYERSPELL_NEW`.
4. `.additem 6948`, then `.save` promptly (a delayed `.save` lets the partial
   inventory save commit first and masks the item loss; position can be used as
   the observable instead).
5. Without the patch: `1062` and `Transaction aborted` in `SQL.log`, and the
   item is absent from `character_inventory`.
   With the patch: no error, `specMask` updated to the in-memory value, item
   persisted.

- [ ]
- [ ]

### DIFF
--- a/src/server/database/Database/Implementation/CharacterDatabase.cpp
+++ b/src/server/database/Database/Implementation/CharacterDatabase.cpp
@@ -510,7 +510,8 @@ void CharacterDatabaseConnection::DoPrepareStatements()
     PrepareStatement(CHAR_UDP_CHAR_MONEY, "UPDATE characters SET money = ? WHERE guid = ?", CONNECTION_ASYNC);
     PrepareStatement(CHAR_UDP_CHAR_MONEY_ACCUMULATIVE, "UPDATE characters SET money = money + ? WHERE guid = ?", CONNECTION_ASYNC);
     PrepareStatement(CHAR_UPD_CHAR_REMOVE_GHOST, "UPDATE characters SET playerFlags = (playerFlags & (~16)) WHERE guid = ?", CONNECTION_ASYNC);
-    PrepareStatement(CHAR_INS_CHAR_ACTION, "INSERT INTO character_action (guid, spec, button, action, type) VALUES (?, ?, ?, ?, ?)", CONNECTION_ASYNC);
+    PrepareStatement(CHAR_INS_CHAR_ACTION, "INSERT INTO character_action (guid, spec, button, action, type) VALUES (?, ?, ?, ?, ?)"
+                     "ON DUPLICATE KEY UPDATE action = VALUES(action), type = VALUES(type)", CONNECTION_ASYNC);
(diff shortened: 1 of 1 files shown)


Fabian-Evolved:
I've tested this, and it seems to have stopped from happening, before this fix items would disappear quite consistently from players, after this fix, it hasn't happened since. So i assume it fixed it.

walkline (approved):

PR #24340, opened 2026-01-06
fix(Scripts/IcecrownCitadel): Fixed Blood Orb Game Object

EricksOliveira:
Scripted Blood Orb game object
Scripted trash mobs around him
Scripted spells anti-exploit of buffs

https: //github.com/TrinityCore/TrinityCore/pull/23397

## Changes Proposed:

This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [X] Scripts (bosses, spell scripts, creature scripts).
-  [X] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> While the use of AI tools when preparing pull requests is not prohibited, contributors must clearly disclose when such tools have been used and specify the model involved.
> 
> Contributors are also expected to fully understand the changes they are submitting and must be able to explain and justify those changes when requested by maintainers.

- [ ] AI tools (e.g. ChatGPT, Claude, or similar) were used entirely or partially in preparing this pull request. Please specify which tools were used, if any.

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/2176

## SOURCE:

The changes have been validated through:
- [X] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [X] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [X] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

## How to Test the Changes:

- [X] This pull request can be tested by following the reproduction steps provided in the linked issue
- [ ] This pull request requires further testing. Provide steps to test your changes. If it requires any specific setup e.g multiple players please specify it as well.

1.
2.
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

[...]

PR #26850, opened 2026-07-28
fix(Scripts/IcecrownCitadel): restore Crimson Hall door and darkfallen pack pull

### LINKED ISSUE #26821: Crimson Hall Door Not Open
### Current Behaviour

Crimson hall door not open!

### Expected Behaviour

I observe in game that after killing all mobs in the crimson hall in ice crown citadel the door wont open
it was ok the other day but now since the new update of the repo it wont open, please help fix it.
screenshot

Image

### Source

it will be temporarily fix using .gm on and u can unlock the door

just enter the raid instance

### Extra Notes

_No response_

### AC rev. hash/commit

n/a

### Operating system

Debian 13

### Custom changes or Modules

eluna
some eluna scripts

EricksOliveira:
Commit [ac6592a](https://github.com/azerothcore/azerothcore-wotlk/commit/ac6592ae3fd1599962a125e1170d7022dd6bb9fd) (PR [#24340](https://github.com/azerothcore/azerothcore-wotlk/pull/24340)) assigned ScriptName to the Crimson Hall darkfallen entries. In FactorySelector::SelectAI, ScriptName takes precedence over AIName, so SmartAI stopped being instantiated for those creatures and two DB-driven behaviours were silently lost. Two further problems in the same commit's C++ were masked by them.

1. Crimson Hall door never opens (issue [#26821](https://github.com/azerothcore/azerothcore-wotlk/issues/26821))

Four spawns of the entrance pack carried a per-GUID SAI (SMART_EVENT_DEATH -> SMART_ACTION_SET_INST_DATA, field 300) feeding BloodPrinceTrashCount in instance_icecrown_citadel.cpp. With SmartAI disabled the counter never reached 4 and GO_CRIMSON_HALL_DOOR, registered as DOOR_TYPE_PASSAGE for DATA_BLOOD_PRINCE_TRASH, stayed shut. Reimplemented in DarkFallenAI::JustDied, restricted to the same four spawn ids.

2. Pack no longer pulls as a group

npc_icc_orb_controller::SetGUID called minion->AI()->DoZoneInCombat(darkfallen). CreatureAI::DoZoneInCombat(Creature*) replaces me with the creature passed in, so the loop re-engaged the puller once per minion and never touched the minions. Replaced with EngageWithTarget on the puller's target, matching what CallOfHelpCreatureInRangeDo does for assistance.

3. Orb controller minion lookup too narrow

ICCOrbControllerMinionSearch was capped at 10.0f and the entrance pack's advisor (spawn 201479) was never registered in _minionGuids, so the group pull in npc_icc_orb_controller::SetGUID ran with an incomplete pack. The cause is cell granularity rather than distance: WorldObject::GetDistance subtracts both combat reaches, so all four entrance darkfallen are 6.58 to 9.26 yd effective and fit inside 10.0f, but Cell::VisitObjects adds no reach to the search radius and falls back to the standing cell alone when the area collapses to a single cell. At radius 10 that is cell (188, 214) around the orb, while 201479 sits in (187, 214). The searcher now takes a range parameter and the controller uses 15.0f, which spans x 187..188 and picks it up; the polymorph-ally lookup in DarkFallenAI keeps 10.0f. The two upper orbs were unaffected at either value (their packs are 5.16 to 8.11 and 5.74 to 7.73 effective). Headroom at 15.0f is 3.76 yd — the nearest non-pack creature is 22.23 yd raw, 18.76 after reaches — and the pack one floor up is only 7.7 to 8.8 yd away in 2D, excluded because GetDistance is 3D.

4. Call for help on aggro

SMART_ACTION_CALL_FOR_HELP was present on every darkfallen entry and on the four entrance spawns, in every case with param1 = 19. It is now issued from DarkFallenAI::JustEngagedWith with that same 19.0f radius, after the orb controller is notified: CallForHelp runs synchronously into the helpers' JustEngagedWith, so notifying first guarantees the creature that reaches npc_icc_orb_controller::SetGUID has a real victim and drives the pack pull, leaving the GetAnyTarget lookup as a fallback rather than the normal path.

SQL

The update removes the smart_scripts rows, which can no longer execute, and clears the leftover AIName. Entry 38463 (npc_icc_orb_controller) is included: it also received a ScriptName in [ac6592a](https://github.com/azerothcore/azerothcore-wotlk/commit/ac6592ae3fd1599962a125e1170d7022dd6bb9fd) and still carried AIName='SmartAI' plus four rows (spell hit 70293 / 70227) that the C++ replaced. The conditions rows added by [#24340](https://github.com/azerothcore/azerothcore-wotlk/pull/24340) (SourceTypeOrReferenceId=13, entries 70293 / 70299 / 72099) are untouched.

Note: dropping entry row 37763 ("Darkfallen Archmage - On Death - Set GO Flags") also removes an unlock that fired on the death of any archmage, not only the channeler. The C++ path narrows it to the channeler, matching TrinityCore. This is a deliberate behaviour change rather than a straight restoration of what SmartAI did.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Opus5

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/26821

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


ronxmag:
approve success!

sudlud:
The call for help radius should be 19.0f, not 10.0f. The deleted rows carry `action_type=39` (`SMART_ACTION_CALL_FOR_HELP`, param1 is the radius) with param1 = 19: `smart_scripts.sql:273, 277, 283, 290` for the four GUID rows and `37727, 37750, 37756, 37761, 37770` for the entry level ones. The description says "the same 10.0f radius those rows carried", and CodeRabbit's "required 10-yard range" has nothing behind it. At 10 yards the Archmage spawn 201482 sits 18.6 yards from its nearest pack mate and never chains in, though the 15 yard orb controller registration still covers it.

The description also says the call for help is issued after the orb controller is notified, but the code does it before. `CallForHelp` runs synchronously into the helpers' `JustEngagedWith`, so a helper normally reaches `npc_icc_orb_controller::SetGUID` first with no victim yet, and only the new `GetAnyTarget` fallback saves it. Swapping the two lines matches the stated intent.

Entry 38463 (`npc_icc_orb_controller`) has the same dead data and is missing from the cleanup: it got a ScriptName in ac6592a, still has `AIName='SmartAI'`, and still has four `smart_scripts` rows (spell hit 70293 / 70227) that the C++ replaced.

Three points on the changed lines: `Unit * target` should be `Unit* target`; `minion->EngageWithTarget(target);` is indented level with its guarding `if`, so it reads as if it were outside it; and `CALL_FOR_HELP_RADIUS` is a single use function local const while `ORB_CONTROLLER_MINION_RANGE` right next to it is file scope.

Not caused by this PR, but worth knowing: one of the deleted rows, "Darkfallen Archmage - On Death - Set GO Flags", has no C++ equivalent. The only path that clears `GO_FLAG_NOT_SELECTABLE` is `ACTION_SIPHON_INTERRUPTED`, which fires only on `AURA_REMOVE_BY_CANCEL`, not on the channeler's death. That behaviour has been gone since ac6592a and this PR does not restore it.

EricksOliveira:
Thanks for the review. I'll take a closer look at it tomorrow.

EricksOliveira:
All five points addressed. The 19.0f figure came from the deleted rows themselves; CodeRabbit's "required 10-yard range" had no source and I shouldn't have taken it. The siphon interrupt turned out to be a one-line inversion against TrinityCore rather than just an orphan row, so I restored the original condition here.

sudlud:
The Siphon Essence hunk should come out. TrinityCore has `== AURA_REMOVE_BY_CANCEL` today; commit `3b0743ea67` (2019) is the one that changed `!= AURA_REMOVE_BY_EXPIRE` into it, so ac6592a ported the fixed version rather than an inverted one.

My earlier note about the death path was wrong, sorry. 70299 is channeled, and `Unit::setDeathState` (`Unit.cpp:11354`) interrupts non-melee spells before removing auras on death, which routes through `Spell::cancel` and removes the aura with `AURA_REMOVE_BY_CANCEL` (`Spell.cpp:3724`). Master already unlocks the orb when the channeler dies.

What the wider condition adds is the evade path: `CreatureAI::_EnterEvadeMode` removes auras with `AURA_REMOVE_BY_DEFAULT` (`CreatureAI.cpp:403`, `Unit.cpp:5719`), and `npc_icc_orb_controller::DoAction(ACTION_EVADE)` only re-locks the orb while `_isInCombat` is true. If another pack member evades first, the channeler's evade clears `GO_FLAG_NOT_SELECTABLE` with nothing to re-set it, and the orb stays clickable with the pack alive at home. Keeping the null guard on `GetTarget()->GetAI()` is fine on its own.

Worth noting in the description: dropping entry row 37763 ("Darkfallen Archmage - On Death - Set GO Flags") also drops an unlock that fired on any archmage death, not just the channeler's. Narrowing it to the channeler matches TC, so I think it is correct, but it is a behaviour change beyond restoring what SmartAI did.

Everything else checks out. The 19 yd radius and the ordering match the deleted rows, 38463 is covered, and the 15 yd minion range is right for all three orbs, not just the entrance one: the packs sit at up to 13.26, 11.59 and 11.73 yards from their orb, and the next creature is 22.23 away.

EricksOliveira:
Thanks for the follow-up.

I was working from the 2019 TrinityCore diff and never checked the current file, so "inverted port" was my own mistake rather than something ac6592a did. Reverted the Siphon Essence hunk; the null guard on GetTarget()->GetAI() is kept.

I also tested the evade path before reverting, with logging on both npc_icc_orb_controller::DoAction and DarkFallenAI::DoAction. With the wider condition the orb ended up clickable after an evade, and with the reverted one the darkfallen unlock never fires at all, so the controller's re-lock is the last word and master's behaviour was already correct. Two pull/evade cycles, two re-locks, no unlocks.

Added the 37763 note to the description as a deliberate behaviour change, and updated the two description items that didn't match the code (radius and ordering).

sudlud:
Sorry, Claude went a bit off track on my first review here and I am still cleaning up after it.

Correction to my last comment. The geometry rationale I gave for the 15 yd range was wrong, and it has since gone into the description. The value is right, the reasoning is not.

The distance check never excluded any of the four. `WorldObject::GetDistance` subtracts both objects' combat reach (`Object.cpp:1301`), and the reaches here are 3.0 for the advisor, 2.4 for the blood knight and the archmage, 2.475 for the noble, plus 1.0 for the orb stalker (entry 38463 carries `flags_extra` 128, so `ObjectMgr::ChooseDisplayId` at `ObjectMgr.cpp:1651` always hands it the invisible model). Effective distances at the entrance orb are 6.58, 6.75, 6.89 and 9.26, all inside the old 10.0f.

What actually lost the advisor is cell granularity. `Cell::VisitObjects` forwards to the x/y/radius overload, so no reach is added to the search radius (`CellImpl.h:171-177`), and `Visit` falls back to the standing cell alone when the area collapses to one cell (`CellImpl.h:85-88`). Cells are 66.67 yd. At radius 10 the area around the entrance orb is the single cell (188, 214). Spawn 201479 sits in (187, 214) and was never visited at all. At radius 15 the area spans x 187..188 and it is found.

So three of the four registered at 10.0f rather than one, and the two upper orbs were never affected: their packs come out at 5.16 to 8.11 and 5.74 to 7.73 effective, well inside 10. Item 3 and the code comment should say the entrance advisor fell outside the visited cell, not that the pack was out of distance range.

The margin is also not what my "next creature is 22.23 away" implied. That is the exact distance. The check sees 22.23 - 2.475 - 1.0 = 18.76, so 3.76 yd of headroom over 15.0f. Worth keeping in the comment so nobody raises the constant to 17 or 18 later on the strength of the exact figure.

One thing worth stating next to the constant: the pack on the floor above sits 7.7 to 8.8 yd from the entrance orb in 2D and in the same grid cell. It is excluded only because `GetDistance` is 3D. That, not horizontal separation, is what makes 15.0f safe.

The "up to 13.3yd" figure in the comment is correct as written (13.26 measured from the orb), so nothing to change there.

Fabian-Evolved:
Can this be pushed? it's working with this patch.

sudlud:
Fixed the title, it was stored with a literal ellipsis and would have squash-merged that way. Text from your commit subject on 581d809.

One orphan left over: `conditions` row `(22, 6, 37664, 0, ...)` "Requires Empowering Orb Visual Stalker In Range" binds to smart_scripts id 5, not 6 (`GetConditionsForSmartEvent` looks up `event_id + 1`), and this update deletes that row.

```sql
DELETE FROM `conditions` WHERE `SourceTypeOrReferenceId` = 22 AND `SourceGroup` = 6 AND `SourceEntry` = 37664 AND `SourceId` = 0;
```

It is the only one, nothing else points at the deleted entries or guids.

Also: the advisor's deleted rows cast 72065 Shroud of Protection and 72066 Shroud of Spell Warding, `npc_darkfallen_advisor` only schedules 72066. Dead since ac6592a rather than here, but worth adding or noting as dropped like 37763.

Fabian-Evolved:
What's the status on this one? I've tested it, and it works, but not sure if only my test is good enough. But would love to be able to compile this tomorrow in the weekly compile as well for the live server.

Gultask (approved):

PR #26852, opened 2026-07-28
fix(Core/Player): delete invalid character_spell rows on login

walkline:
## Changes Proposed:

When `ValidateSkillLearnedBySpells` rejects a spell during `_LoadSpells`, the code only called `removeSpell()`. That is a no-op if the spell was never added to `m_spells` (validation runs *before* `addSpell`). The orphan `character_spell` row stayed in the database.

A later permanent re-learn of the same spell then issued a bare `INSERT`, hit MySQL **1062 Duplicate entry**, and aborted the entire character save transaction (inventory, position, gold, etc.). Memory had already been marked clean, so the loss became permanent on relog.

This PR always runs `DELETE FROM character_spell WHERE guid = ? AND spell = ?` when validation rejects a spell, so invalid rows are actually removed on login.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Grok (xAI) — investigation, root-cause analysis, and the code change

## Issues Addressed:
- Related to #24216 (orphan `character_spell` + MySQL 1062 aborting full player save)
- Related to #22370 (incomplete delete path for invalid race/class spells)

## SOURCE:

Code review of `_LoadSpells` / `CheckSkillLearnedBySpell` / `removeSpell` and production logs from #24216 (e.g. invalid spell 227 Staves on a Paladin GUID 347, then `INSERT character_spell` 1062).

## Tests Performed:
This PR has been:

Compiled successfully: `game` library (`PlayerStorage.cpp`) on macOS arm64 (RelWithDebInfo). Full `worldserver` link was not verified here due to an unrelated local readline/SDK issue. No in-game verification yet.

## How to Test the Changes:

1. Pick a class that cannot use Staves (e.g. Paladin). Confirm skill 136 is not valid for that class in `SkillRaceClassInfo`.
2. Insert an illegal permanent row:
   ```sql
   INSERT INTO character_spell (guid, spell, specMask) VALUES (<guid>, 227, 255);
   ```
3. Log in with that character. Server log should still report the invalid race/class spell (from `CheckSkillLearnedBySpell`).
4. After login (or after a short wait for the async delete), verify the row is gone:
   ```sql
   SELECT * FROM character_spell WHERE guid = <guid> AND spell = 227;
   ```
   Expect **0 rows** (with this PR). Without this PR the row remains forever.
5. While still online, force a permanent re-learn if possible (e.g. GM `.learn 227` if allowed), then wait for autosave / `.save`. Confirm no MySQL **1062** on `character_spell` and that normal inventory changes still persist across relog.
6. Regression: log in a normal Mage/Priest that legitimately has Staves skill/spells and confirm they still load and save correctly.

- [ ] This only fixes orphan rows from failed login validation. Full save idempotency for other `PLAYERSPELL_NEW` collisions (upsert / dirty-state after commit) is still a separate follow-up.
- [ ] In-game test not yet performed by the author.

### DIFF
--- a/src/server/game/Entities/Player/PlayerStorage.cpp
+++ b/src/server/game/Entities/Player/PlayerStorage.cpp
@@ -6610,7 +6610,16 @@ void Player::_LoadSpells(PreparedQueryResult result)
             if (CheckSkillLearnedBySpell(spellId))
                 addSpell(spellId, specMask, true);
             else
+            {
+                // Spell was never addSpell()'d, so removeSpell is often a no-op and would
+                // leave an orphan character_spell row (MySQL 1062 on later re-learn/save).
                 removeSpell(spellId, SPEC_MASK_ALL, false);
+
+                CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_DEL_CHAR_SPELL_BY_SPELL);
+                stmt->SetData(0, GetGUID().GetRawValue());
+                stmt->SetData(1, spellId);
+                CharacterDatabase.Execute(stmt);
+            }
         } while (result->NextRow());
     }
 }


sudlud:
`PlayerStorage.cpp:6619`: `character_spell.guid` holds the low guid, so this should be `GetGUID().GetCounter()`, like the identical DELETE in `_SaveSpells` and every other bind in the file. It only matches today because `HighGuid::Player` is 0.

Line 6618 is 128 chars, over the 120 limit from `.editorconfig`.

Worth coordinating with #26834, which attacks the same 1062 rollback from the other side by making `CHAR_INS_CHAR_SPELL` and friends idempotent. No conflict between them, but they overlap.

walkline:
@sudlud Nah, `GetGUID().GetRawValue()` is required now if we want to support crossrealm.
Regarding the long line - I still believe it is readable.

sudlud:
Thanks, that makes sense now. I hadn't looked at how ToCloud9 handles this. The mysqlreverseproxy reads the realm off the guid bind argument to pick the right characters DB, and `character_spell` is one of the columns it routes on, so raw really is the value you need there. Your fork binds `GetRawValue()` throughout, including both the DELETE and the INSERT in `_SaveSpells`, so it's consistent on your side.

Upstream it comes out identical anyway since `HighGuid::Player` is 0, so I'm happy to drop it. Only thing I'd mention is that upstream still binds the counter everywhere else, including the same DELETE at PlayerStorage.cpp:7848, so this one line reads a bit odd without the context.

sudlud (approved):

Fabian-Evolved:
Can this be merged please? as this actually fixes the missing items bug it seems. I haven't been able to reproduce it with this patch in place.

It would be very useful to being able to get this on the live server, as a lot of players are reporting missing items.

PR #26844, opened 2026-07-28
fix(Scripts/Ulduar): sync Mimiron's Rocket Strike marker and detonation with the missile

Nyeriah:
## Changes Proposed:
Mimiron's Rocket Strike sequence is currently desynced: the strike trigger (NPC 34047) is only summoned when the 63036 missile lands (~2.3s of travel at the DBC speed of 7 yd/s), so the ground marker appears only at touchdown (useless as a warning) and the 64064 fuse (5s duration, single tick triggering 63041) starts counting from touchdown. The result is the rocket visually slamming into the ground and sitting there for ~5 seconds before the explosion. On retail the marked spot is announced at launch and the rocket detonates on impact, with roughly a 5 second window from the DBM warning to the boom (measured frame by frame from https://youtu.be/F051-eHuxIE?t=138: warning renders at 141.4, detonation around 145.4-146.3).

This PR keeps the whole DBC chain and the existing missile animation, and only reorders the timing:

- When 63681 picks the victim, the strike trigger is spawned immediately at their position, so the 64064 warning visual shows from launch and the dodge window is the full 5 seconds.
- The mounted rocket fires 63036 after `5000ms - travel time` (computed per strike from the actual distance at 7 yd/s), so the missile lands as the fuse expires. Until then it stays visibly loaded on VX-001's shoulder; the display swap/reload behaviour is unchanged.
- A new spell script on 63036 pins the missile destination to the pre-spawned trigger (a moving player can no longer drag the impact point away from the marker) and suppresses its summon effect, since the trigger already exists.
- The trigger detonates 63041 exactly at the computed missile impact time. The 64064 periodic tick is suppressed via aura script because the aura timer and the client-side missile run on independent clocks; any offset between them produces a visible double explosion. Same approach as Kil'jaeden's Armageddon meteors ([spell_kiljaeden_armageddon_periodic_aura](https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/scripts/EasternKingdoms/SunwellPlateau/boss_kiljaeden.cpp), which also combines `PreventDefaultAction()` on `SPELL_AURA_PERIODIC_TRIGGER_SPELL` with a missile destination override).
- Removes the dead `SetData`/speed-accumulator leftovers in `npc_ulduar_mimiron_rocket` from an earlier lift-off attempt (see #25251).

[...]

PR #26858, opened 2026-07-29
fix(Scripts/Ulduar): Mimiron's Rocket Strike

PkllonG:
## Changes Proposed:
This will roll back #26844

https://github.com/user-attachments/assets/904c5c49-9a3e-471f-b417-45fbb5b21a1c

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:
- [ x ] Tested in-game by the author.

## How to Test the Changes:

1. Refer to the testing method in 26844

- [ ]
- [ ]


Nyeriah:
I’m sorry but it doesn’t seem anyone else is having issues with this boss but you. Could you recheck your setup? Queries/DB/core? Any custom changes?

PkllonG:
I just feel like doing it this way is simpler than before.
And doing this won't make you look like a hacker.

Nyeriah:
So adding a dbc hack in your opinion makes it not look like a hack?

PkllonG:
I think the reverted PR is a hacker, and it’s implemented in such a complicated way.

PkllonG:
Maybe you think that as long as it's complex enough, it's not hacking.

Nyeriah:
Thanks for taking the time to look at this encounter, but I don't think this PR should be merged in its current form. I want to lay out the reasoning in full, because the disagreement here is really about what counts as a hack and what doesn't, and I think that deserves more than a one-line answer.

**What this PR does, restated**

This reverts https://github.com/azerothcore/azerothcore-wotlk/pull/26844 (the pre-spawned strike trigger, the delayed missile fire computed from actual distance, the destination pinning on 63036, and the fuse tick suppression on 64064) and replaces all of it with a single data override in `SpellInfoCorrections.cpp` that forces `Speed = 0.0f` on 63036 (Summon Rocket Strike).

I'll grant up front that the timing outcome is broadly the same. With zero travel time the summon effect resolves instantly at target selection, so the strike trigger (NPC 34047) spawns immediately, the 64064 warning visual shows from launch, and the natural 5 second fuse tick fires 63041 at the right moment. And because the summon lands instantly at the position where the target was selected, the marker and the damage can't desync from a moving player either, so the destination pinning from 26844 is indeed not needed under this model. On pure timing, this works, and I understand the appeal of deleting ~110 lines of script.

**Why I still think this is the wrong trade**

1. **This fixes a script-layer sequencing problem by falsifying DBC data.** 63036 legitimately has a speed of 7 yd/s in the DBC. That is real client data, verifiable in sniffs, and it describes real retail behavior: the rocket travels. The problem 26844 addressed was never that the DBC value was wrong; it was that our script started the fuse at missile touchdown instead of at launch. That is an ordering bug in our code, and 26844 fixed it in our code while leaving the data intact. This PR instead makes the data wrong so that our naive script ordering happens to produce the right timing. On the "hack" question raised in the comments: complexity is not what makes something a hack. A hack is when you change something that is known to be correct in order to compensate for something else that is wrong. The DBC speed is correct. Overwriting it to 0 is the hack, regardless of how few lines it takes. `SpellInfoCorrections` exists to fix genuinely broken or missing spell data, not to simplify scripts by making authentic data inauthentic, and every entry added there for convenience makes it harder to trust the file for its actual purpose.

2. **It visibly regresses the encounter compared to both retail and current master.** With `Speed = 0` the client gets no missile travel at all: no rocket flying from VX-001's shoulder to the ground. The revert also restores `SetDisplayId(11686)` at target-select time, so the sequence a player sees is: the rocket silently vanishes from the shoulder, a ground marker appears, and five seconds later the ground explodes with nothing arriving. On retail (see the frame analysis in https://github.com/azerothcore/azerothcore-wotlk/pull/26844, based on https://youtu.be/F051-eHuxIE?t=138) the rocket visibly flies and detonates on impact, and syncing that impact with the explosion was the entire point of the rework. This PR doesn't simplify that behavior, it deletes it. "Simpler" is only a virtue when the output is equivalent, and here it isn't.

3. **The justification for a rollback is missing.** The PR description is "This will roll back #26844" plus a video. No issue is linked, and no concrete defect in 26844 is described: no repro steps, no expected vs. actual, nothing another tester could verify. 26844 was merged one day before this was opened, after review, and so far nobody else has reported a problem with it. If it genuinely misbehaves on your setup, please open an issue with reproduction steps first, and let's fix the actual defect in place; a targeted fix will almost certainly be smaller than a revert plus a DBC override. If the objection is purely that you find the merged implementation too complex, that is feedback that belongs on the implementation, and the answer to it can't be to reintroduce data corruption for the sake of line count.

**Smaller points, in case this moves forward anyway**

- The `spell_script_names` DELETE for `spell_mimiron_summon_rocket_strike` and `spell_mimiron_rocket_strike_aura` is correct and required given the script removal, so that part is fine.
- The pending SQL file should be named `rev_<timestamp>.sql` via `data/sql/updates/pending_db_world/create_sql.sh` rather than `boss_mimiron.sql`, and it currently starts with a blank line and a stray `--`.
- The CodeRabbit comment about time-based rocket acceleration refers to code that doesn't exist in this diff and can be ignored.

To be clear about where I land: the `Speed = 0` trick does reproduce the retail timing window, and I don't doubt the video shows the timing looking right. But it gets there by making correct DBC data incorrect and by dropping the missile flight that retail demonstrably has, and it reverts a merged, working fix without demonstrating a bug in it. That's a net loss on both authenticity and visuals. I'd ask that this either be closed, or reframed as an issue documenting whatever misbehavior you're seeing with 26844 so it can be fixed properly.

PR #26890, opened 2026-08-01
feat(Deps/libsidecar): Libsidecar (clustering) ABI versions check and portable shared-library loading

walkline:
Cluster mode needs a real libsidecar that matches the headers in this tree. It is easy to mix an old or wrong build with new headers (or the other way around). That used to show up late as confusing crashes or load errors.

This change makes the expected version obvious (1.0.0 / tc9_version.h) and checks it when cluster starts. If the loaded library does not match what worldserver was built for, it stops with a clear error: use headers and library from the same libsidecar build.

It also installs and finds the library in predictable places (build tree vs install prefix / next to the binary on Windows), so people are less likely to run with the “wrong file from somewhere on disk".

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Grok (xAI / Grok Build)

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Build and run with cluster mode.

### DIFF
--- a/deps/PackageList.txt
+++ b/deps/PackageList.txt
@@ -86,6 +86,6 @@ fkYAML (A C++ header-only YAML library)
   https://github.com/fktn-k/fkYAML
   Version: 721edb3e1a817e527fd9e1e18a3bea300822522e
 
-libsidecar (enables interaction with other components when the worldserver is running in cluster mode. It requires building a shared library first and putting it in the corresponding folder.)
-  https://github.com/walkline/ToCloud9/tree/master/game-server/libsidecar
-  Version: master
+libsidecar (enables interaction with other components when the worldserver is running in cluster mode. Place the shared library in deps/libsidecar for builds. UNIX: BUILD_RPATH finds deps/libsidecar; install copies to ${prefix}/lib; runtime also searches next to worldserver and ../lib. Windows: no rpath — build copies the DLL next to worldserver; install places it next to worldserver. Headers under deps/libsidecar/include — including tc9_version.h — must come from the same libsidecar build/release as the linked binary. worldserver checks ABI at cluster init.)
+  https://github.com/walkline/ToCloud9/tree/master/game-server/libsidecar-cpp
+  Version: 1.0.0 (TC9_VERSION in deps/libsidecar/include/tc9_version.h)
--- a/deps/PackageList.txt
+++ b/deps/PackageList.txt
@@ -86,6 +86,6 @@ fkYAML (A C++ header-only YAML library)
   https://github.com/fktn-k/fkYAML
   Version: 721edb3e1a817e527fd9e1e18a3bea300822522e
 
-libsidecar (enables interaction with other components when the worldserver is running in cluster mode. Place the shared library in deps/libsidecar for builds. UNIX: BUILD_RPATH finds deps/libsidecar; install copies to ${prefix}/lib; runtime also searches next to worldserver and ../lib. Windows: no rpath — build copies the DLL next to worldserver; install places it next to worldserver. Headers under deps/libsidecar/include — including tc9_version.h — must come from the same libsidecar build/release as the linked binary. worldserver checks ABI at cluster init.)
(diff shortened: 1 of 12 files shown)


sudlud:
Checked against upstream and the released artifact: the shipped `libsidecar-cpp-1.0.0` .so exports the three functions and the vendored headers match the release apart from comments. Premise holds.

**The check does not fail fast.** `sToCloud9Sidecar->Init()` runs at Main.cpp:403, after `StartDB()` (280), `SetInitialWorldSettings()` (310), `StartWorldNetwork()` (355), the realm being set online (374) and the ready log (388), so a mismatch aborts after the realm is already accepting connections. Both version functions return compile-time constants and could be checked right after config load. Main.cpp:466-468 already reads `Cluster.Enabled` straight from config, with a comment noting `Init()` has not run yet.

**Stub builds pass, then die anyway.** With the default `USE_REAL_LIBSIDECAR=OFF` (conf/dist/config.cmake:110) and `Cluster.Enabled = 1`, the stub answers from its own macros (stub/libsidecar.c:12-34, and only `stub/` is on the include path per deps/libsidecar/CMakeLists.txt:24-25): it logs `runtime 1.0.0 ... headers 1.0.0`, passes, then `TC9InitLib` panics and exits. That is the path most people trying cluster mode will actually hit. Naming the stub in that log line would fix it. Reporting a fake mismatch instead would swap a clear message for a misleading "copy matching headers + library" abort.

Two CMake points, both reproduced in a scratch project mirroring these files:

- src/server/apps/CMakeLists.txt:156 documents the install order backwards. The target inherits `CMAKE_INSTALL_RPATH` (root CMakeLists.txt:35) at creation and the code appends, so the installed binary gets `<prefix>/lib:$ORIGIN:$ORIGIN/../lib`, prefix first. The BUILD line above it is correct.
- deps/libsidecar/CMakeLists.txt:77-81. `IMPORTED_LOCATION_NOCONFIG` and `IMPORTED_SONAME_NOCONFIG` are never selected in any configuration AC supports (`$<TARGET_FILE:libsidecar>` gives the plain `IMPORTED_LOCATION` under Makefiles and Ninja for Release/Debug/RelWithDebInfo/MinSizeRel, and under Ninja Multi-Config). They could go while you are in this block.

Minor: showoptions.cmake:130 sits outside the `USE_REAL_LIBSIDECAR` conditional ending at 128, so stub builds also print `libsidecar required ABI : 1.0.0`, read from include/tc9_version.h while the stub compiles stub/tc9_version.h. And PackageList.txt could link the release page rather than the source tree, now that the tag ships prebuilt binaries.

On CodeRabbit's `server.sidecar`: the rest of that file uses `"server"`, but `TC9Sidecar/AsyncTask.h:65` already uses `"server.tc9"`, so if you change it, move the whole file.

walkline:
Hey @sudlud, I believe your feedback addressed, let me know if something is missing.

sudlud (approved):

sudlud:
a few things on `CheckLibsidecarAbi`:

`ABORT` here ends in a real crash (`Acore::Abort` does `*((volatile int*)nullptr) = 0`), so a wrong `.so` leaves operators with a core dump and a crash log. the other startup checks in `Main.cpp` just `return 1`, and the DBC version check uses `LOG_ERROR` + `exit(1)`. can we make this return a bool and `return 1` from main?

second one: in a stub build both sides of the comparison come from `stub/tc9_version.h`, so the check can only pass, and we die later in `Init()` at `Main.cpp:407` once the realm is already flagged online. since `TC9_LIBSIDECAR_IS_STUB` is right there, maybe bail out at that point with "rebuild with `-DUSE_REAL_LIBSIDECAR=ON`"? stub is the default so it's probably the common case.

and a small one on top: with `Log.Async.Enable = 1` both log lines only get posted to the io_context, which starts at `Main.cpp:257`, so they never print before we exit. `sLog->SetSynchronous()` or a plain stderr print would keep the message.

PR #26950, opened 2026-08-03
fix(Graveyard): use corpse location to select closest graveyard

MacWarrior:
When a player dies, the closest graveyard should be determined from the corpse location rather than the player's current position.

The current implementation uses the player's current zone and area during RepopAtGraveyard(). This can produce incorrect graveyard selection when the player is moved after death (teleport, transport, delayed relocation, or other server-side position changes before release).

This is the current retail process applied when using while dead :
image

This PR updates the graveyard lookup performed during repop so that the zone and area are resolved from the corpse coordinates instead of the player's current position. This ensures that the selected graveyard always matches the location where the player died.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

- [X] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: ChatGPT (GPT-5.5)

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Suicide somewhere having a graveyard (Let's say Stormwind)
2. Your spirif spawn just outside of stormwind
3. Move somewhere else in elwynn forest
4. Somehow call Player::RepopAtGraveyard()
Before edition, you'd be teleported to nearest graveyard of your current position
After edition, you'll be teleported to original graveyard, the nearest of your corpse position


src/server/game/Misc/GameGraveyard.cpp
```diff
@@ -117,7 +117,16 @@ GraveyardStruct const* Graveyard::GetClosestGraveyard(Player* player, TeamId tea
...
+
+    if (nearCorpse && player->HasCorpse())
+    {
+        Map* map = sMapMgr->CreateBaseMap(mapId);
+        map->GetZoneAndAreaId(PHASEMASK_NORMAL, zoneId, areaId, x, y, z);
+    }
+    else
+    {
+        player->GetZoneAndAreaId(zoneId, areaId);
+    }
```
sudlud: no braces around single line if / else statements

src/server/game/Misc/GameGraveyard.cpp
```diff
@@ -105,7 +105,7 @@ GraveyardStruct const* Graveyard::GetClosestGraveyard(Player* player, TeamId tea
 
     WorldLocation loc = player->GetWorldLocation();
 
-    if (nearCorpse)
+    if (nearCorpse && player->HasCorpse())
     {
         loc = player->GetCorpseLocation();
     }
```
sudlud: while you're here you could remove the braces

sudlud:
`Player::RepopAtGraveyard` isn't the only caller this changes, and two of the others look like regressions.

**Wintergrasp.** The battlefield branch at `Player.cpp:5025` keys off `GetZoneId()`, the ghost's current zone, while the fallback now resolves from the corpse's zone. A ghost who left WG with his corpse still inside skips `Battlefield::GetClosestGraveyard`, so the `gy->GetControlTeamId()` filter never runs, and the zone 4197 rows in `graveyard_zone` carry no control state at all. Can we end up sending him back into WG to a graveyard the other faction holds?

**Spirit healer.** `SendSpiritResurrect` already passes `HasCorpse()` (`NPCHandler.cpp:272`) and teleports when `corpseGrave != ghostGrave`. Until now both enumerated the ghost's zone and usually matched. Now: die in Stormwind, release, walk to Westfall, use the Westfall spirit healer, and you get pulled back to Stormwind. Is that intended?

**Rescue callers.** Same for the under-map rescue (`MovementHandler.cpp:525`), fatigue drowning (`Player.cpp:960`) and `.unstuck graveyard`. `_corpseLocation` is never cleared on release, so a ghost who left an instance gets sent back to its entrance graveyard from anywhere in the world. Worth a test pass at least.

**Description.** The corpse is created at release, not at death (`KillPlayer` skips it on purpose, `BuildPlayerRepop` calls `CreateCorpse`), so "player is moved after death before release" can't really happen, the corpse follows him. The real case is the one in your test steps, an already released ghost that moved. Since we squash, that text ends up as the commit message.

src/server/game/Misc/GameGraveyard.cpp
sudlud: and if we use `sMapMgr->GetZoneAndAreaId(PHASEMASK_NORMAL, zoneId, areaId, mapId, x, y, z)` here, the `if` branch collapses to one line too and both sides lose the braces

src/server/game/Misc/GameGraveyard.cpp
sudlud: also `nearCorpse && player->HasCorpse()` is evaluated again at 121, maybe hoist it into a local?

MacWarrior:
Hi @sudlud,
 
Thanks for your review. I've updated the PR to fix all these cases.

- Corpse-based repop is now explicitly requested only by release paths. 
- Rescue callers keep using the ghost's current location, while Spirit Healer keeps its previous zone lookup behavior.
- Wintergrasp now uses the corpse zone for Battlefield selection when performing a corpse-based repop, preserving its graveyard control rules.

PR #26830, opened 2026-07-26
fix(Core/Spells): Charge no longer drops you to your death mid-jump

Chinoske:
https://github.com/user-attachments/assets/8729b92e-3136-4386-b899-d99a0ecc5dfb

## Changes Proposed:

Charging while falling (e.g. jumping off Kologarn's walkway in Ulduar and Charging him mid-air) could carry the caster to their death. `Spell::CheckCast`'s charge-target validation builds a path via `PathGenerator`, but the "falling" shortcut in `PathGenerator::BuildPolyPath` produces a raw, unvalidated straight line whose endpoint is never confirmed to sit on real walkable navmesh. On top of that, the existing `PATHFIND_FARFROMPOLY_END` check alone isn't enough: while falling it's usually the *caster's own* mid-air position that registers as far from any polygon (`PATHFIND_FARFROMPOLY_START`), not the target's, so the unsafe case slipped through untouched.

Rather than rejecting the cast outright, this adds `PathGenerator::ShortenPathUntilSafeGround()`: when either end of the path was flagged far from a polygon, it walks the original (unsnapped) destination backward in small steps until it finds a point with confirmed real navmesh nearby, and charges only that far - landing the caster on the walkway in front of the target instead of falling through where the target stands. If no safe point exists anywhere along the path, the cast fails with `SPELL_FAILED_NOPATH` instead of moving the caster at all.

A new helper `PathGenerator::HasWalkableNavmeshAt()` does the actual check via `dtNavMeshQuery::findNearestPoly` with a tight search box, snapping to the polygon's own surface height rather than trusting the raw interpolated point.

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Claude Code (Opus 5)
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:
- Closes #26266

[...]

PR #26955, opened 2026-08-04
fix(Core/Spells): Charge now works up slopes instead of refusing to cast

### LINKED ISSUE #22331: [Warrior] Charge is not climbing terrain in wsg

Chinoske:
## Changes Proposed:

Charge is refused whenever the walking route to the target is long, even when the target is a few yards away up a slope. In WSG you can never charge anyone standing on either graveyard.

The cast runs the path through a length limit, `maxRange * 1.5f + objSize`, which `SetPathLengthLimit` turns into points of `SMOOTH_PATH_STEP_SIZE`. For Charge (8-25yd) that is 9 points, i.e. about 32 yards of walking. Measured at the Alliance graveyard with `.mmap path`, from a spot 21 yards from the target in a straight line:

```
Result: true - Length: 33 - Type: 1
StartPosition     (1069.0845, 1382.8339, 340.13882)
EndPosition       (1077.8861, 1389.0566, 322.01556)
ActualEndPosition (1078.0149, 1389.1853, 321.98935)
```

33 points, roughly 128 yards round the bank, against a limit of 9. `PATHFIND_SHORT` is the first of the three checks, so the cast is refused there.

When that limit is the only problem, this re-checks reachability without it and requires `PATHFIND_NORMAL`. That is what stops this reaching anywhere you could not walk to, and it has to be a fresh check: `BuildShortcut` overwrites the NORMAL/INCOMPLETE result before `CheckCast` sees it, so `PATHFIND_SHORT` on its own cannot answer the question.

The straight line then gets resampled onto the ground by a new `PathGenerator::SnapPathToGround`. A bare two-point line is not usable: across a convex slope it runs inside the hill, and the client drops you out of the world when the spline ends in there. Sampled against `GetHeight` along that same charge:

```
t=0.4  line 329.23   ground 329.13
t=0.5  line 331.06   ground 332.48   1.4 under
t=0.6  line 332.88   ground 335.84   3.0 under
t=0.7  line 334.70   ground 338.21   3.5 under
t=0.8  line 336.53   ground 338.98   2.5 under
t=0.9  line 338.35   ground 339.76   1.4 under
```

`SnapPathToGround` also returns false when the ground falls away below the line instead of rising to meet it, which is the difference between climbing a bank and clearing a gap onto a roof. Rooftop targets are still refused.

Two things worth knowing for review:

- The landing spot is worked out from the target's own footing rather than by backing off along the line. Backing off along a climb steps back down into the slope and drops the caster through it.
- `IsWalkableClimb` looked like the natural guard here and does not work: the bank measures about 74 degrees between sample points, which is why it has no walkable navmesh in the first place, so that check refuses the very case this fixes.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #22331
- Same report on the ChromieCraft tracker: chromiecraft/chromiecraft#8087

## SOURCE:

The issue carries WotLK footage of a warrior charging up the WSG graveyard: https://youtu.be/HrGcvxggyrM?t=212 and https://youtu.be/Z4174ce_62M?t=94

Two commenters on the issue describe it as generic rather than WSG-specific. @CheckYourFax:

> I vaguely remember doing this during WotLK to get up steep curves in more places than wsg (often to try and be lazy to not have to walk all the way around). I think its a generic thing related to the charge implementation.

Worth recording for anyone reading the issue thread: the analysis posted there points at `IsInvalidDestinationZ` and its 5.0f threshold. That is not what refuses the cast. Measured on the failing charge above, the gap between the target and the path's actual end is 0.026 yards against a threshold of 5.0, and `PATHFIND_SHORT` returns first anyway.

TrinityCore 3.3.5 has the same code, including the `// first try with raycast` comment above a call that never sets `SetUseRaycast`, so this is inherited rather than an AzerothCore regression.

On adding a method rather than extending `NormalizePath`: that one pins the points a path already has, and a shortcut only has two, the ends, so there is nothing in the middle to pin. What is needed here is generating the intermediate points first, which nothing in the codebase does. It is also called from six places while every path in the game is built, so widening it would change movement everywhere. `SnapPathToGround` is opt-in and only the charge shortcut asks for it. Its insides are the existing pieces: the same `UpdateAllowedPositionZ` call `NormalizePath` uses, `Movement::PointsArray`, and `SMOOTH_PATH_STEP_SIZE` passed in by the caller.

## Tests Performed:
This PR has been:

Local Windows build, level 80 warrior, each case from several approach angles:

| case | before | after |
|---|---|---|
| WSG graveyard bank | "no path available" | climbs the slope and lands on top |
| Target on a rooftop | "no path available" | still refused |
| Ordinary charge on flat ground | works | unchanged |
| Charging Kologarn mid-jump | fails safely | unchanged |
| Feral Charge - Bear, same bank | "no path available" | climbs the slope, same as the warrior |

The rooftop and Kologarn cases are the ones that matter for regressions: the first is what a straight-line charge could abuse, and the second is issue #26266.

Server-side numbers from the fixed run, showing the path following the surface and the landing on solid ground:

```
pt[0] = (1078.03 1388.78 321.94)     ground 321.94
pt[1] = (1076.49 1387.76 325.00)     ground 325.00
pt[2] = (1074.95 1386.73 329.13)     ground 329.13
pt[3] = (1073.41 1385.71 335.84)     ground 335.84
pt[4] = (1071.87 1384.69 338.98)     ground 338.98
pt[5] = (1070.33 1383.66 340.17)     ground 340.17
```

## How to Test the Changes:

1. `.tele WarsongGulchAlliance`, then `.npc add 31144` to leave a dummy up on the graveyard
2. Walk down off the bank, more than 8 yards away (Charge has a minimum range), out of combat
3. Charge the dummy. It should climb the bank and land next to it
4. Repeat on the Horde side
5. Put a dummy on a rooftop with nothing but line of sight to it and confirm the charge is still refused

Testing was done on a build that also carries #26830, which is still open and touches the same function. The change here is written against master without it; the two coexist, but that combination is not what upstream will build.

`MotionMaster::MoveCharge(PathGenerator const&)` says in a comment that movement is not started for `EVENT_CHARGE_PREPATH`. It is: `PointMovementGenerator::DoInitialize` has no early return for that event, so it reaches `init.MoveTo(i_x, i_y, i_z, true)` and launches a pathfound spline that the explicit `MovebyPath` right after then replaces. Two splines go out per charge. Not touched here, but it is worth a separate look.


Chinoske:
Both CodeRabbit findings were real, pushed in befbb004b.

The sample count now rounds up. It was truncating, so a path longer than one step but shorter than two came out as `steps = 1` and returned early without ever querying the terrain. Reachable in practice: Charge has an 8yd minimum, but the landing spot is backed off by the target's combat reach, which can leave the line around 5-6 yards. A short hop over a small gap was going through unchecked.

`SnapPathToGround` now takes only the two-point shortcut it documents, and is only called when the recalculated path is one. If the way to the landing spot does fit the length limit it is a real navmesh route and gets kept as is, instead of being flattened into a line.

On the Feral Charge - Bear question from the pre-merge check: tested, works. Druid in bear form up the same WSG graveyard bank, same result as the warrior. Nothing here is class-specific, it all sits in the generic `case SPELL_EFFECT_CHARGE`, but it was fair to ask for the test rather than take that on trust.

Retested after the change: graveyard bank still climbs, rooftop target still refused, ordinary charge on flat ground unchanged.

Also related, though it does not overlap: #26758 touches the same charge pre-path for oversized targets.

Chinoske:
added a note to the description answering the obvious question about `SnapPathToGround`: why a new method rather than extending `NormalizePath`.

short version, that one pins the points a path already has, and a shortcut has only two, the ends, so there is nothing in the middle to pin. what is needed is generating the intermediate points first. it is also called from six places while every path in the game is built, so widening it would change movement everywhere.

no code change. the red check here is the authserver startup step, which #27000 addresses, not this.

Chinoske:
Tested.

https://github.com/user-attachments/assets/0320bebf-8f9d-41f3-b4a6-267a1002c78e

Chinoske:
Ran `/self-review` on this. It found two things I want to fix before this goes any further, so treating it as **not ready** for now.

**1. The new `PATHFIND_SHORT` branch accepts the second path without validating it.** The only guard, `GetPath().size() < 2`, can never fire: every failing exit in `BuildPolyPath`/`BuildPointPath` goes through `BuildShortcut()`, which does `_pathPoints.resize(2)` exactly, and the only 1-point case needs `_slopeCheck`, which is false on `m_preGeneratedPath`. So cases that used to return `SPELL_FAILED_NOPATH` now go through as a raw un-snapped straight line into `MoveCharge` — the same straight line through the hill that drops the client out of the world.

Two concrete ones: `land` off the navmesh gives `PATHFIND_NORMAL | PATHFIND_NOT_USING_PATH` for a player caster over a 2-point shortcut, so no `PATHFIND_SHORT` and the snapping is skipped entirely. And `land` more than 7 yd from its poly gives `PATHFIND_INCOMPLETE | PATHFIND_FARFROMPOLY_END` with the endpoint relocated, so Charge "succeeds" and drops you at the foot of the bank with rage and cooldown spent.

The `bool` from the second `CalculatePath` is discarded too, so a rejected coordinate silently keeps the first call's stale path and stale type, and the snapping then runs on that.

**2. The "won't carry anyone onto a roof" guarantee isn't actually in the code.** `SnapPathToGround` only rejects ground *below* the chord (`point.z < lineZ - maxDrop`); there is no upper bound, and the probe starts at `max(point.z, previousZ) + stepSize * 2`, with `previousZ` only ever moving up. A sample can latch onto a deck or ledge and every later sample then probes from there.

And the case the guard exists for is the one where it cannot fire: `UpdateAllowedPositionZ` leaves `z` untouched when it finds no ground, so over a real void the sample keeps `lineZ + 8`, and `lineZ + 8 < lineZ - 4` is false — the point is accepted 8 yards in mid air, which is exactly the "hanging in mid air halfway along" the comment says it prevents.

Also worth doing: on the failure path this now runs up to 6 full path builds per attempt instead of 2, since the `reachable` probe uses the default 74-point limit and `CheckCast` runs twice per cast. That is the path players repeat by mashing the button.

Full report below.

# Self-review — fix/charge-climb-terrain-22331 → master

> **Outcome** 5 findings — 0 fixed, all open — 1 blocking
> **Intent** Warrior's Charge refused to cast at targets up a slope, or failed once cast. The change is meant to make Charge work uphill.
> **Reviewed** `dcdd23a11` (`fix/charge-climb-terrain-22331` vs `origin/master`, merge-base diff — 3 files, +89 −7)
> **By** Claude Code (VS Code extension, Claude Agent SDK), claude-opus-5 — self-review v0.1, 2026-08-09
> **Project rules** `.agents/docs/self-review-rules.md` found and applied
> **Caveat** no fixes applied yet — findings 1 and 2 need a rework rather than a patch
> **In-game testing** author reports: not recorded for this round

## Round 1 — `dcdd23a11`, 5 findings

1. `src/server/game/Spells/Spell.cpp:6281` — the `PATHFIND_SHORT` branch accepts the rebuilt path unchecked; `GetPath().size() < 2` is dead code, and the `bool` from `CalculatePath` is discarded, so a rejected `land` keeps the first call's stale path and type → **open, blocking**
2. `src/server/game/Movement/MovementGenerators/PathGenerator.cpp:1228` — `SnapPathToGround` has no upward bound and cannot detect "no ground found", so it accepts points in mid air over a void; the guarantee stated in the comment and commit message is not implemented → **open**
3. `src/server/game/Spells/Spell.cpp:6271` — the `reachable` probe uses the default 74-point limit and adds a third `CalculatePath`, so a failing Charge costs up to 6 path builds per attempt instead of 2 → **open**
4. `src/server/game/Spells/Spell.cpp:6290` — `SMOOTH_PATH_STEP_SIZE` reused as a vertical drop tolerance, and `stepSize * 2.0f` as a probe height, both unexplained magic numbers coupling unrelated tuning → **open**
5. `PathGenerator.cpp:1233` — `SnapPathToGround` rewrites `_pathPoints` but leaves `_type` advertising `PATHFIND_SHORTCUT | PATHFIND_SHORT`, so the object's type no longer matches its contents; harmless today since `MoveCharge` ignores the type and `m_preGeneratedPath` has no other consumer, but a trap for the next caller → **open**

## What came back clean

- **No regression to other pathing consumers.** `SnapPathToGround` is purely additive with exactly one caller; it does not touch `_type`, `_actualEndPosition` or `_polyLength`, and mutates `_pathPoints` only via `swap` after every early return, so a rejected snap leaves the path untouched. `NormalizePath`, `BuildShortcut`, `ShortenPathUntilDist` and `UpdateAllowedPositionZ` are unmodified. Creature chase, flee and waypoint movement, pets and every other consumer are unaffected.
- **The non-`PATHFIND_SHORT` path is behaviourally identical to master** — same three checks in the same order, only reflowed. Charges that worked before take the exact same route.
- **Style**: no added line over 120 columns (measured by hand), no tabs or trailing whitespace, codestyle passes, no long-lived raw unit pointers, no debug prints or TODOs.

## Unverified

- The WSG graveyard bank figures in the commit message and the comment at `Spell.cpp:6268` ("~21yd straight but ~130yd on foot") were taken on faith, not measured against the navmesh.
- Whether this actually makes Charge succeed on the reported slope. The 4-yard `maxDrop` means a **concave** slope, where the ground dips more than 4 yd below the chord, still fails the cast, so it does not cover all uphill geometry.

## Still to test in game

- The reported WSG graveyard case, plus a concave hillside, to see whether the cast still refuses.
- **This block is `SPELL_EFFECT_CHARGE` for any explicit-unit-target caster, so every NPC charge in the game gains this behaviour.** Pull a charging mob on broken terrain from somewhere close but a long walk away, and confirm it does not end up inside geometry.
- A target on an elevated walkway reachable only by a long route, with line of sight — where finding 2 predicts the charger is lifted up through the structure.
- A target across a chasm on an indoor map (Ulduar, ICC) — finding 2's "no ground found" case.
- A target on a bank edge where `land` falls off the navmesh — finding 1's case; watch for Charge eating rage and cooldown and dropping you short.
- Charge at a boss-sized target, since `land` is offset by `objSize` and sits further from the target.

Chinoske:

TheSCREWEDSoftware:
CC @walkline opnion on this PR since you're more familiar with the MMAPs

Also @Chinoske this PR has merge conflicts

Chinoske:
Merged master and fixed the conflicts, head is now `30f5f133`.

The clash was with the oversized combat reach block that landed after I opened this. I kept that flow intact and hung the `PATHFIND_SHORT` recovery inside the failure branch: `NOPATH` and `INCOMPLETE` still bail out right away, only a pure `PATHFIND_SHORT` gets the reachability re-check. `IsInvalidDestinationZ` and `ShortenPathUntilDist` stay under `cutPath`, same as before.

Retested in game on this commit: charge up the WSG graveyard bank works, targets on a roof are still refused, and Kologarn behaves the same as on master.

sudlud:
thanks for sorting the conflicts, and the two CodeRabbit points are properly fixed.

quick thing on the self-review first. the idea is to fix what it finds and rerun until a round comes back clean, rather than post it as is. the august report has two blocking findings and a "not ready" on it, and the only commit since is the merge, so both are still open at the head. they were good findings, which is the annoying part.

both still hold, plus a third:

`Spell.cpp:6316` the second `CalculatePath` result isn't checked. only `size() < 2` and `PATHFIND_SHORT` are tested, so `PATHFIND_NOPATH` (creature caster, `PathGenerator.cpp:291`) and `NORMAL | NOT_USING_PATH` (player caster, line 284) fall through with a raw two point shortcut and skip `SnapPathToGround`. that's the straight line through the hill. can we check the type there?

`Spell.cpp:6297` no `IsInvalidDestinationZ` in this branch anymore. with the target on the lip of a drop, `land` steps off the lip and resolves down to the ground below, the path there is flat so nothing snaps, and the charge ends at the foot of the bank with rage and cooldown spent. master refused that cast. worth testing with the dummy right on the edge.

`PathGenerator.cpp:1233` the guard only looks down, so nothing checks that consecutive samples are joined by anything climbable. and `UpdateAllowedPositionZ` leaves `z` alone when it finds no ground (`Object.cpp:1636`), so over a void the sample keeps `lineZ + 8` and the check never fires. that one needs a WMO only map or an ADT hole, outdoors `GetGridHeight` always finds a floor.

`IsWalkableClimb` is out, you're right about the 74 degrees. no obvious replacement, which is part of why this could use walkline's eyes.

minor:

- `Spell.cpp:6325` `SMOOTH_PATH_STEP_SIZE` as both sample spacing and drop tolerance, maybe give the drop its own constant?
- `PathGenerator.cpp:1240` `_type` still says `SHORTCUT | SHORT` after the points are replaced. harmless today, trap for the next caller.
- `PathGenerator.cpp:1209` the comment overstates it, 4 yard sampling still can't see a gap narrower than 4 yards.
- `Spell.cpp:6281` up to 4 path builds per `CheckCast`, and it runs twice on a cast that goes through. cooldown bounds it, just flagging.

also worth merging master in again: it gained the WorldObject caster refactor, so the block builds `m_preGeneratedPath` on `unitCaster` while the new `reachable` at 6307 still uses `m_caster`.

Chinoske:
the two CodeRabbit findings are already covered on the current head (`30f5f133`), they were raised against an earlier revision:

- `SnapPathToGround` bails out unless the path is exactly two points (`PathGenerator.cpp:1202`), so a real multi-point detour is never flattened.
- the sample count already rounds up with `std::ceil` (`PathGenerator.cpp:1212`).
- `CheckCast` only calls it when the recalculated path still comes back `PATHFIND_SHORT` (`Spell.cpp:6324`); a normal path is kept as it is.

Chinoske:
all three still held at the head, thanks for going through it again. fixed in `53d186b32`.

the second `CalculatePath` has its result checked now, and the type gate only lets through a plain `PATHFIND_NORMAL` route or the same `SHORTCUT | SHORT` shortcut. everything else goes back to `SPELL_FAILED_NOPATH` the way master had it. one thing worth knowing for that exact comparison: `BuildShortcut` assigns `_type = PATHFIND_SHORTCUT` rather than OR-ing it, so the FARFROMPOLY bits set just above in `BuildPolyPath` get wiped and the SHORT result is always exactly those two flags.

`IsInvalidDestinationZ` is back in the branch, so the lip of a drop refuses again.

on `SnapPathToGround`: the probe stops at one climb above the line instead of ratcheting up with no bound, which is also what keeps a sample off a roof, since the height query only ever reports a surface at or below where it starts. it reads `groundZ` back and bails when there's nothing under the sample, so the void case fails instead of sitting at `lineZ + 8`. the last segment into the landing spot gets the same climb check, which nothing was doing before. drop and climb are their own constants now.

on the perf point, the unlimited probe moved behind the cheap checks, so it only builds on a cast that's going to go through. it still runs twice per cast and I don't see a way around that short of restructuring `CheckCast`.

two things I left alone. `_type` still says `SHORTCUT | SHORT` after the snap and I think that's honest: the points follow the surface, but they come from height queries rather than the navmesh, and the walking route is still over the limit. I put it on the method contract instead. and the sampling comment no longer claims more than it can do, it says outright it can't see a hole narrower than the step.

`reachable` builds on `unitCaster` now, but I haven't merged master in yet and I haven't retested in game on this commit, so both still need a pass.

Issue #26984, opened 2026-08-05
[CI] Dashboard CI: the ubuntu-26.04 leg fails on any PR that recompiles the core


Chinoske:
### Summary

The `Build and Integration Test (ubuntu-26.04)` job of `Dashboard CI` fails on any pull request that actually recompiles core code. The failure is in libstdc++ headers, not in the code under review, so it is not something the PR author can fix. The `ubuntu-24.04` leg then goes red as collateral, because the matrix uses `fail-fast: true` and the sibling job is cancelled.

Master looks green, which hides this. Its own `Dashboard CI` runs are full ccache hits: the latest successful ubuntu-26.04 build and integration job on master finished in **4m35s** ([run 31005791258](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/31005791258)), where a real build takes around 45 minutes. Nothing is compiled there, so the compiler is never asked to parse the headers that break.

### Affected pull requests

All three currently open with a red check, all failing on the same job:

| PR | Failing run |
| :-- | :-- |
| #26735 | [job 92198476550](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30972100383/job/92198476550) |
| #26820 | [job 92198492917](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30972106279/job/92198492917) |
| #26955 | [job 92198506214](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30972110887/job/92198506214) |

### Cause: clang is compiling against GCC 16's libstdc++

On the ubuntu-26.04 runner, the default `clang` is clang-21 and the newest installed GCC toolchain is 16, so clang picks up GCC 16's libstdc++ headers and cannot parse them. Two fatal errors, both from `/usr/lib/gcc/x86_64-linux-gnu/16/`:

```
/usr/lib/gcc/x86_64-linux-gnu/16/../../../../include/c++/16/shared_mutex:528:33: fatal error: no member named '__to_timeout_timespec' in namespace 'std::chrono'
```

```
/usr/lib/gcc/x86_64-linux-gnu/16/../../../../include/c++/16/numeric:302:21: fatal error: no template named '__is_any_random_access_iter'; did you mean '__is_random_access_iter'?
```

[...]

PR #26735, opened 2026-07-22
fix(Core/Player): weapons can now be swapped while disarmed

Chinoske:
## Changes Proposed:

Disarmed players couldn't swap gear at all in the disarmed weapon slot - trying to equip or unequip anything there (e.g. going from a two-hander to a one-hand + shield combo mid-Disarm) failed with "You can't do that while disarmed." Blizzlike behaviour, per the evidence in the issue, is that you *can* re-gear a disarmed slot - you just can't use the new weapon until Disarm wears off.

`Player::CanEquipItem`/`Player::CanUnequipItem` (`PlayerStorage.cpp`) had an explicit guard added a while back that blocked any equip/unequip on a disarmed slot outright:

```cpp
// Xinef: dont allow to equip items on disarmed slot
if (!CanUseAttackType(GetAttackBySlot(eslot)))
    return EQUIP_ERR_NOT_WHILE_DISARMED;
```

Removed both instances of that guard (equip side and unequip side). On its own that would let a disarmed weapon's damage go live again just by re-equipping it, which is presumably what the guard was there to prevent, so `Player::EquipItem()` now suppresses the newly-equipped weapon's damage and dependent auras for as long as the slot is still disarmed - the same treatment `AuraEffect::HandleAuraModDisarm` gives whatever's equipped at the moment Disarm lands. `Player::RemoveItem()` restores them symmetrically right before an item leaves a disarmed slot, so the bookkeeping stays correct no matter how many times gear gets swapped during a single Disarm window, and there's nothing left to double up when Disarm eventually ends and hands stat control back to `HandleAuraModDisarm` for whatever weapon is in the slot by then.

**What doesn't change**: `Unit::CanUseAttackType()` - the thing that actually stops a disarmed character from swinging - is untouched. It still reads `UNIT_FLAG_DISARMED`/`UNIT_FLAG2_DISARM_OFFHAND`/`UNIT_FLAG2_DISARM_RANGED` regardless of which weapon occupies the slot, so the new weapon just sits there unusable until the effect ends, exactly like the video evidence in the issue shows.

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
[...]

PR #26985, opened 2026-08-05
fix(CI): build the Dashboard CI ubuntu-26.04 leg with gcc-15

Chinoske:
## Changes Proposed:

`Dashboard CI` is the only build workflow that does not pin a compiler. On ubuntu-26.04 that means it picks up whatever `clang` resolves to, which is clang 21, and clang 21 cannot parse the libstdc++ 16 headers it selects. It dies in `<shared_mutex>` and `<numeric>` before reaching any AzerothCore code:

```
/usr/lib/gcc/x86_64-linux-gnu/16/../../../../include/c++/16/shared_mutex:528:33: fatal error: no member named '__to_timeout_timespec' in namespace 'std::chrono'
```

```
/usr/lib/gcc/x86_64-linux-gnu/16/../../../../include/c++/16/numeric:302:21: fatal error: no template named '__is_any_random_access_iter'; did you mean '__is_random_access_iter'?
```

Every pull request that actually recompiles the core goes red on it, and since the matrix uses `fail-fast: true`, the ubuntu-24.04 leg is cancelled along with it and shows red too. Neither has anything to do with the code under review.

This pins both legs the way `core-build-pch.yml` and `core-build-nopch.yml` already do: clang-18 on 24.04 and gcc-15 on 26.04, two compilers those workflows run today. As a side effect the `runner.os` ccache restore-keys now line up with the ones the `linux-build` action writes on both legs, where before they only matched on 24.04.

Part of #26984, which also records a second and unrelated symptom on that workflow.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- None

## SOURCE:

Code inspection: `libstdc++-16-dev` is already installed on that runner by #26940, and the failing runs confirm it (`Setting up libstdc++-16-dev:amd64 (16-20260322-1ubuntu1)`), so this is not the missing-package problem that fixed. It is clang 21 being unable to consume that header set at all.

The reason this is not already visible on master is ccache: the ubuntu-26.04 legs of `Dashboard CI`, `pch-build` and `nopch-build` all finish in under 5 minutes there against roughly 45 for a real build, so nothing is compiled and the compiler is never asked to parse those headers. The same applies to the clang-21 legs of the core-build workflows, which are left untouched here since they are a separate call on whether to keep clang coverage on 26.04 or move it to a toolchain clang can handle.

`Detect compiler` already reads `${CC:-}` before falling back to `command -v clang`, so pinning through the job env is enough for both the build and the ccache key to follow.

## Tests Performed:
This PR has been:

Not run locally: this only takes effect on the GitHub runners, so the workflow run on this pull request is the test. What should be visible is the `Build and Integration Test (ubuntu-26.04)` job compiling and going green instead of stopping in `Run complete installation (deps, compile, database, client-data)`.

Evidence for the diagnosis comes from the failing runs on #26735 ([job 92198476550](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30972100383/job/92198476550)) and #26820 ([job 92198492917](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/30972106279/job/92198492917)), and from the job durations on the latest master run of each build workflow.

## How to Test the Changes:

Watch this pull request's own `Dashboard CI` run:

1. `Build and Integration Test (ubuntu-26.04)` should get through `Run complete installation (deps, compile, database, client-data)` rather than failing in it after roughly two minutes.
2. The `Echo cache key` step should print `gcc-15_g++-15` for that leg instead of `clang-21_clang++-21`.
3. `Build and Integration Test (ubuntu-24.04)` should stay green on `clang-18`.

Note that a green run here is not on its own proof, since a warm ccache can hide a compile break on this workflow. A leg that compiles for real takes tens of minutes, not a handful.

- None

### DIFF
--- a/.github/workflows/dashboard-ci.yml
+++ b/.github/workflows/dashboard-ci.yml
@@ -75,8 +75,20 @@ jobs:
       matrix:
         include:
           - os: ubuntu-24.04
+            compiler:
+              CC: clang-18
+              CXX: clang++-18
+          # gcc rather than clang here: the default clang on 26.04 is 21, which
+          # cannot parse the libstdc++ 16 headers it picks up and dies in
+          # <shared_mutex> and <numeric> before reaching any of our code
           - os: ubuntu-26.04
+            compiler:
+              CC: gcc-15
+              CXX: g++-15
     runs-on: ${{ matrix.os }}
+    env:
+      CC: ${{ matrix.compiler.CC }}
+      CXX: ${{ matrix.compiler.CXX }}
     if: github.repository == 'azerothcore/azerothcore-wotlk' && !github.event.pull_request.draft
     steps:
       - name: Checkout repository
@@ -149,9 +161,9 @@ jobs:
       # recent on any ref the run can reach, which is its own and the base
       # branch, and is how a first run inherits master's. Both must use
       # matrix.os to stay in the key's namespace.
-      # The runner.os entries below reach the linux-build action's caches, but
-      # only on the 24.04 leg: on 26.04 this workflow detects clang-21 while
-      # linux-build is passed clang-20, so nothing there lines up.
+      # The runner.os entries below reach the linux-build action's caches, which
+      # now line up on both legs: clang-18 on 24.04 and gcc-15 on 26.04 are both
+      # compilers the core-build workflows already run.
       - name: Restore ccache
         id: restore_ccache
         uses: actions/cache/restore@v4


sudlud:
hm the dashboard setup relies on the dashboard scripts to set up the toolchains / dependencies. so imo this pr is not the right way to address the issue

Chinoske:
you're right, wrong layer.

digging into it: the dashboard compiles with whatever `CCOMPILERC` points at, and `conf/dist/config.sh` defaults that to `/usr/bin/clang`, which on 26.04 is clang 21. it can't parse the libstdc++ 16 headers it picks up. so this isn't really a CI thing, anyone installing on 26.04 through the dashboard hits the same wall. CI was just the only place it surfaced, and only some of the time, since a warm ccache hides it.

two ways I can see to fix it in the scripts:

- point clang at the GCC 15 toolchain, something like `--gcc-install-dir=$(dirname "$(g++-15 -print-libgcc-file-name)")`. keeps clang, no hardcoded paths, and it drops out on its own once clang can handle libstdc++ 16. `libstdc++-15-dev` is already on the box since the installer pulls `g++`.
- or fall back to gcc on 26.04. that one is proven, this PR's own run compiled for real in about 21 minutes on gcc-15, but it gives up clang coverage on that release.

I'd lean to the first. happy to close this and redo it in the installer instead, which would you prefer?

Chinoske:
closing in favour of #26986, which does it in the dashboard scripts as you suggested. turns out it is not only CI: `conf/dist/config.sh` points the dashboard at `/usr/bin/clang`, so nobody can compile on 26.04 right now. that one probes the compiler and falls back to a GCC toolchain clang can parse, leaving working setups untouched. thanks for the steer.

PR #26997, opened 2026-08-06
fix(Scripts/Warrior): Sweeping Strikes no longer crashes the server on Execute

### LINKED ISSUE #26987: Crash: spell_warr_sweeping_strikes::HandleProc dereferences null _procTarget when Execute procs
### Steps to Reproduce

1. Warrior with the **Sweeping Strikes** talent active (spells 12328 / 18765 / 35429).
2. Cast **Execute** (spell 20647) on a target below 20% health, with at least one other enemy nearby so Sweeping Strikes selects a secondary target.
3. Occasionally — not on every cast — the worldserver segfaults.

**Intermittent.** Observed frequency on our realm: **4 crashes in ~12 hours** with only 2 warriors having the talent, ~50 concurrent players. Each crash took the whole server down.

No special configuration required. Default `worldserver.conf` apart from rates.

## Backtrace

```
Object::HasFlag (this=0x0, index=61, flag=2)
  #1  spell_warr_sweeping_strikes::HandleProc
        at src/server/scripts/Spells/spell_warrior.cpp:677
  ...
  #16 spell_warr_execute::HandleEffect
        at src/server/scripts/Spells/spell_warrior.cpp:426
```

`index 61` = `UNIT_FIELD_AURASTATE`, `flag 2` = `AURA_STATE_HEALTHLESS_20_PERCENT` — i.e. the `_procTarget->HasAuraState(AURA_STATE_HEALTHLESS_20_PERCENT)` call in the Execute branch of `HandleProc`.

## Cause

`_procTarget` is assigned in `CheckProc`:

```cpp
_procTarget = actor->SelectNearbyNoTotemTarget(eventInfo.GetProcTarget());
```

and read later in `HandleProc`. Between the two calls the selected target can be gone, so the cached raw pointer can be null when it is used.

## Expected behaviour

The extra attack is skipped when the secondary target no longer exists. No crash.

## Fix we are running

One null check, in production for 2 days now with no recurrence (previously 4 crashes/day):

```cpp
void HandleProc(AuraEffect const* aurEff, ProcEventInfo& eventInfo)
{
    PreventDefaultAction();

    if (!_procTarget)
        return;

    if (DamageInfo* damageInfo = eventInfo.GetDamageInfo())
    ...
```

## Environment

- AzerothCore rev `007b4464f5d2` (2026-07-22), branch master
- 3.3.5a, Linux (Docker), `RelWithDebInfo`
- ~50 concurrent players

Full core dump backtrace available on request.

## Related

We hit a second crash of the same shape three days later, in Eluna: a pointer that can legitimately be null, dereferenced without a check. Reported separately at ElunaLuaEngine/Eluna#537.

It may be worth auditing spell scripts that cache a raw `Unit*` in `CheckProc` and use it in `HandleProc`. We swept our copy: 48 candidates, and this was the only one that dereferences the cached pointer directly — the rest pass it to `CastSpell`, which null-checks in `SpellCastTargets::SetUnitTarget`.

### Operating System

Ubuntu 24.04 x64 (host) — worldserver running in Docker (official acore/ac-wotlk-worldserver image, Ubuntu-based)

### Dependencies & versions

Built from the official `apps/docker/Dockerfile`, so the versions are the ones that image pins:

- MySQL 8.0 (server, in a separate container)
- `default-libmysqlclient-dev` 21.2.46
- Boost: `libboost-all-dev`, linked statically (`-DBoost_USE_STATIC_LIBS=ON`)
- Compiler: clang / clang++ (as set by the Dockerfile), with `ccache`
- libstdc++ 13
- CMake: `cmake` from the image
- Build type: `RelWithDebInfo`
- OpenSSL: `libssl-dev` from the image

The only local modifications to the source are `apps/docker/Dockerfile` (enables `ccache`) and `MotdMgr.cpp` (custom message of the day) — neither touches spell scripts.

### Commit

007b4464f5d2

### Additional Context

_No response_

Chinoske:
## Changes Proposed:

Sweeping Strikes can take the whole worldserver down. `CheckProc` picks the secondary target and keeps it as a raw `Unit*`, then `HandleProc` dereferences it without checking, so if that target is gone by the time the proc is handled the server segfaults on `Object::HasFlag` with `this=0x0`. Reported in #26987 with a backtrace and a rate of four crashes in twelve hours on a fifty player realm, with only two warriors carrying the talent.

The fix is the one the codebase already uses for the same mechanic. Blade Flurry does exactly this, picks a nearby target on proc and swings at it, and it stores the guid and resolves it at the point of use. It even carries a note about the identical crash:

```cpp
// Xinef: no _procTarget but checkproc passed??
// Unit::CalculateAOEDamageReduction (this=0x0, damage=4118, schoolMask=1, caster=0x7ffdad089000)
Unit* procTarget = ObjectAccessor::GetUnit(*GetTarget(), _procTargetGUID);
```

Sweeping Strikes never got the same treatment, so it does the same here.

The reporter proposed a plain `if (!_procTarget) return;` instead. That covers today's crash but only half the failure: the cached pointer can end up dangling rather than null, and then the check passes and the dereference is worse to diagnose. Resolving a guid covers both, and it is what `AGENTS.md` asks for: *don't store a raw `Player*` / `Creature*` / `Unit*` past the current call/tick ... store the `ObjectGuid` and resolve at use time*.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- Closes #26987

## SOURCE:

Code inspection: verified against master. `CheckProc` assigns at `src/server/scripts/Spells/spell_warrior.cpp:667` and `HandleProc` dereferences at `:677`, which is the line in the reported backtrace.

Six spell scripts cache a `Unit*` selected in `CheckProc`, and only three touch it directly afterwards. `spell_rog_blade_flurry` converts to a guid immediately (`src/server/scripts/Spells/spell_rogue.cpp:133`) and resolves it in `HandleProc` (`:142`). `spell_mage.cpp:599` guards with a null check before use. `spell_warr_sweeping_strikes` was the only one dereferencing unguarded.

`spell_warr_vigilance` keeps the same pattern but hands the pointer to `CastSpell`, which null checks in `SpellCastTargets::SetUnitTarget`, so it is left alone here.

`ObjectAccessor` needs no new include: `spell_dk.cpp` calls `ObjectAccessor::GetUnit` with the same include set in this directory.

## Tests Performed:
This PR has been:

Not reproduced locally: the crash is a race, and the reporter saw it roughly four times in twelve hours on a populated realm without being able to trigger it on demand either.

What can be checked is that the ability still behaves the same. Sweeping Strikes should keep hitting a second nearby enemy on Whirlwind, Cleave and ordinary swings, and Execute above 20 percent should still land the normalized weapon damage variant rather than the mirrored damage one, since that branch is the one that used to dereference.

The C++ codestyle script passes.

## How to Test the Changes:

Normal behaviour, which is what needs checking since the crash itself is a race:

1. On a warrior with Sweeping Strikes active, attack a target with at least one other enemy next to it. The second enemy should take damage alongside the main one.
2. Repeat with Whirlwind and with Cleave.
3. Cast Execute on a target above 20 percent health with a second enemy nearby. The second enemy should take normalized weapon damage.
4. Cast Execute on a target below 20 percent health. The second enemy should take the mirrored damage instead.

For the crash itself, the reporter's realm saw it around four times in twelve hours with two warriors using the talent, so a populated realm running this for a day is the real test.

- None

### DIFF
--- a/src/server/scripts/Spells/spell_warrior.cpp
+++ b/src/server/scripts/Spells/spell_warrior.cpp
@@ -633,7 +633,7 @@ class spell_warr_sweeping_strikes : public AuraScript
 
     bool Load() override
     {
-        _procTarget = nullptr;
+        _procTargetGUID.Clear();
         return true;
     }
 
@@ -664,20 +664,34 @@ class spell_warr_sweeping_strikes : public AuraScript
             }
         }
 
-        _procTarget = actor->SelectNearbyNoTotemTarget(eventInfo.GetProcTarget());
-        return _procTarget != nullptr;
+        Unit* procTarget = actor->SelectNearbyNoTotemTarget(eventInfo.GetProcTarget());
+        if (procTarget)
+        {
+            _procTargetGUID = procTarget->GetGUID();
+        }
+
+        return procTarget != nullptr;
     }
 
     void HandleProc(AuraEffect const* aurEff, ProcEventInfo& eventInfo)
     {
         PreventDefaultAction();
+
+        // resolved rather than cached: the target picked in CheckProc can be gone by now,
+        // same crash spell_rog_blade_flurry already carries a note about
+        Unit* procTarget = ObjectAccessor::GetUnit(*GetTarget(), _procTargetGUID);
+        if (!procTarget)
+        {
+            return;
+        }
+
         if (DamageInfo* damageInfo = eventInfo.GetDamageInfo())
         {
             SpellInfo const* spellInfo = damageInfo->GetSpellInfo();
-            if (spellInfo && spellInfo->Id == SPELL_WARRIOR_EXECUTE && !_procTarget->HasAuraState(AURA_STATE_HEALTHLESS_20_PERCENT))
+            if (spellInfo && spellInfo->Id == SPELL_WARRIOR_EXECUTE && !procTarget->HasAuraState(AURA_STATE_HEALTHLESS_20_PERCENT))
             {
                 // If triggered by Execute (while target is not under 20% hp) deals normalized weapon damage
-                GetTarget()->CastSpell(_procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_2, aurEff);
+                GetTarget()->CastSpell(procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_2, aurEff);
             }
             else
             {
@@ -687,7 +701,7 @@ class spell_warr_sweeping_strikes : public AuraScript
                 }
 
                 int32 damage = damageInfo->GetUnmitigatedDamage();
-                GetTarget()->CastCustomSpell(_procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_1, &damage, 0, 0, true, nullptr, aurEff);
+                GetTarget()->CastCustomSpell(procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_1, &damage, 0, 0, true, nullptr, aurEff);
             }
         }
     }
@@ -699,7 +713,7 @@ class spell_warr_sweeping_strikes : public AuraScript
     }
 
 private:
-    Unit* _procTarget = nullptr;
+    ObjectGuid _procTargetGUID;
 };
 
 // 50720 - Vigilance
--- a/src/server/scripts/Spells/spell_warrior.cpp
+++ b/src/server/scripts/Spells/spell_warrior.cpp
@@ -688,7 +688,8 @@ class spell_warr_sweeping_strikes : public AuraScript
         if (DamageInfo* damageInfo = eventInfo.GetDamageInfo())
         {
             SpellInfo const* spellInfo = damageInfo->GetSpellInfo();
-            if (spellInfo && spellInfo->Id == SPELL_WARRIOR_EXECUTE && !procTarget->HasAuraState(AURA_STATE_HEALTHLESS_20_PERCENT))
+            if (spellInfo && spellInfo->Id == SPELL_WARRIOR_EXECUTE
+                && !procTarget->HasAuraState(AURA_STATE_HEALTHLESS_20_PERCENT))
             {
                 // If triggered by Execute (while target is not under 20% hp) deals normalized weapon damage
                 GetTarget()->CastSpell(procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_2, aurEff);
@@ -701,7 +702,8 @@ class spell_warr_sweeping_strikes : public AuraScript
                 }
 
                 int32 damage = damageInfo->GetUnmitigatedDamage();
-                GetTarget()->CastCustomSpell(procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_1, &damage, 0, 0, true, nullptr, aurEff);
+                GetTarget()->CastCustomSpell(procTarget, SPELL_WARRIOR_SWEEPING_STRIKES_EXTRA_ATTACK_1,
+                    &damage, 0, 0, true, nullptr, aurEff);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Chinoske:
@secoonday any chance you could run this one on your realm instead of the null check you have in?

yours is holding, but it only covers the pointer being null. the other end of the same race is it coming back dangling, and then the check passes and the crash reads a lot worse. this resolves the guid instead, so both ends land in the same place.

your realm is the only place this actually happens. four crashes in twelve hours beats anything I can force on a test server, so if it stays quiet there for a few days that is the real answer. worth an eye on the ability itself too while it is in: second target still getting hit on whirlwind, cleave and normal swings, and Execute above 20% still landing the normalized damage one.

no rush. and thanks for the writeup, the backtrace made this a five minute read.

Chinoske:
updated, the two touched lines are wrapped within 120 columns now. no behaviour change.

worth restating the short version for whoever picks this up: this is not a new idea, it is what `spell_rog_blade_flurry` already does for the same mechanic. same proc, same nearby target selection, same crash, and it has carried a note about it for years:

```cpp
// Xinef: no _procTarget but checkproc passed??
// Unit::CalculateAOEDamageReduction (this=0x0, damage=4118, schoolMask=1, caster=0x7ffdad089000)
Unit* procTarget = ObjectAccessor::GetUnit(*GetTarget(), _procTargetGUID);
```

Sweeping Strikes was the only one of the six scripts caching a `Unit*` from `CheckProc` that still dereferenced it unguarded. ready for a look whenever someone has a moment.

secoonday:
Thanks for picking this up, and for the `spell_rog_blade_flurry` pointer — the guid-resolve approach is clearly the right one and I've swapped our local null check for this patch.

One thing from re-running the sweep afterwards, offered in case it's useful rather than as a disagreement: I don't think `CastSpell` is a safe sink for these cached pointers, which would put a couple of the scripts we both classified as fine back on the list.

`Unit::CastSpell(Unit* victim, ...)` hands the raw pointer straight to `SpellCastTargets::SetUnitTarget`, and that null-checks and *then* dereferences:

```cpp
void SpellCastTargets::SetUnitTarget(Unit* target)
{
    if (!target)
        return;

    m_objectTarget = target;
    m_objectTargetGUID = target->GetGUID();   // <-- dereference
    m_targetMask |= TARGET_FLAG_UNIT;
}
```

So it's safe against a null pointer and not against a stale one — the same split that made the null check insufficient in Sweeping Strikes. "Doesn't touch it directly" ends up meaning "dereferences it one frame down".

That leaves these still holding a raw `Unit*` across the `CheckProc` → `HandleProc` boundary:

| Script | Cached in `CheckProc` | Used in `HandleProc` |
|---|---|---|
| `spell_warr_vigilance` | `GetCaster()` | `CastSpell(_procTarget, ...)` |
| `spell_dk_blood_gorged` | `GetTarget()->GetOwner()` | `CastCustomSpell(..., _procTarget, ...)` |
| `spell_pri_mana_leech` | `GetTarget()->GetOwner()` | `CastSpell(_procTarget, ...)` |
| `spell_mage_focus_magic` | `GetCaster()` | `CastSpell(_procTarget, ...)` |

On the mage one specifically: the `_procTarget && _procTarget->IsAlive()` at `spell_mage.cpp:599` is inside `CheckProc`, so it validates before the gap rather than at point of use. It's the same window Sweeping Strikes had, just with a nicer-looking guard in front of it.

I want to be careful about how much weight to put on this. None of them dereference in the script itself, and the cached object is a caster or a pet owner rather than an arbitrary nearby enemy, so the odds of it going away inside that window are clearly lower — we've never seen one of them crash. It may well be judged not worth the churn. But by the `AGENTS.md` rule they're the same shape, and if a stale pointer ever does reach `SetUnitTarget` the backtrace will be far less obvious than the one that started this.

Happy to open a PR converting those four to `ObjectGuid` in the same style as this one if you think it's worth having. Equally happy to leave it if you'd rather not touch working scripts on a theoretical risk.

For completeness on the other half of our sweep: the Eluna crash we hit is unrelated to this one and is being handled upstream at ElunaLuaEngine/Eluna#537 — their maintainers pushed back on guarding the call sites there, on the grounds that it would mask a state-lifecycle race, which I think is right.

sogladev (approved):

PR #27011, opened 2026-08-07
fix(Core/Build): the unit tests now build on Windows

Chinoske:
## Changes Proposed:

The unit tests do not compile with MSVC. `BUILD_TESTING` defaults to `0` and the Windows workflow builds through `./acore.sh compiler build`, so CI never turns them on and the breakage goes unnoticed. Three separate things stop the build:

- `src/test/CMakeLists.txt` was the only target in the tree not linking `acore-core-interface`, which is where `/utf-8` lives. Without it `fmt` stops the build on its own Unicode `static_assert`. The other targets all link it (`src/common/CMakeLists.txt:67`, `src/server/game/CMakeLists.txt:52`, `src/server/database/CMakeLists.txt:52`, `src/server/apps/CMakeLists.txt:128`), so this just follows them. The `modules` call below it had to move to the keyword form as well, since CMake refuses to mix both signatures on one target.
- The gcov coverage flags were set unconditionally and appended to the global `CMAKE_CXX_FLAGS`, so `cl.exe` received `-fprofile-arcs -ftest-coverage -fno-inline` on **every** target of the build, worldserver included. The `MATCHES GNU` guard right below only covered the two extra flags. The whole block now sits inside a `GNU|Clang` check.
- googletest defaults to the static CRT on MSVC while the core uses the dynamic one, which fails at link time with `LNK2038`. `gtest_force_shared_crt` is googletest's own option for exactly this.

Nothing changes for gcc or clang: the coverage flags are still set for them, and the other two edits are inert outside MSVC.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:
- None

## SOURCE:

## Tests Performed:
This PR has been:

Nothing to test in-game here, it is a build change, so the boxes above stay unchecked. What was done instead:

Built on Windows with Visual Studio 2026, `RelWithDebInfo`, `-DBUILD_TESTING=ON`, static scripts and modules. On master the build stops at the first of the three errors; with this branch it finishes clean and the binary runs:

```
[==========] 11394 tests from 83 test suites ran. (456 ms total)
[  PASSED  ] 5908 tests.
[  SKIPPED ] 5486 tests.
```

worldserver and authserver were rebuilt from the same tree afterwards, to confirm the coverage-flag change does not affect them. Both build and start.

## How to Test the Changes:

Configure on Windows with `-DBUILD_TESTING=ON` and build the `unit_tests` target. On current master it fails with the `fmt` Unicode `static_assert`; with this branch it builds and the binary runs.

`unit_tests.exe` needs the MySQL and OpenSSL DLLs on `PATH`, otherwise it exits with `0xC0000135` and no message.

Only the three build errors are addressed. The Windows CI still does not enable `BUILD_TESTING`, so it will not catch a regression of this on its own.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,16 +146,19 @@ endif()
 CU_RUN_HOOK("AFTER_SRC_LOAD")
 
 if (BUILD_TESTING AND BUILD_APPLICATION_WORLDSERVER)
-    # we use these flags to get code coverage
-    set(UNIT_TEST_CXX_FLAGS "-fprofile-arcs -ftest-coverage -fno-inline")
+    # we use these flags to get code coverage. gcov only, and they land in the global
+    # CMAKE_CXX_FLAGS, so MSVC would get them on every target of the build
+    if ( CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang" )
+        set(UNIT_TEST_CXX_FLAGS "-fprofile-arcs -ftest-coverage -fno-inline")
 
-    # enable additional flags for GCC.
-    if ( CMAKE_CXX_COMPILER_ID MATCHES GNU )
-        set(UNIT_TEST_CXX_FLAGS "${UNIT_TEST_CXX_FLAGS} -fno-inline-small-functions -fno-default-inline")
+        # enable additional flags for GCC.
+        if ( CMAKE_CXX_COMPILER_ID MATCHES GNU )
+            set(UNIT_TEST_CXX_FLAGS "${UNIT_TEST_CXX_FLAGS} -fno-inline-small-functions -fno-default-inline")
+        endif()
+
+        message("Unit tests code coverage: enabling ${UNIT_TEST_CXX_FLAGS}")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${UNIT_TEST_CXX_FLAGS}")
     endif()
-
-    message("Unit tests code coverage: enabling ${UNIT_TEST_CXX_FLAGS}")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${UNIT_TEST_CXX_FLAGS}")
 
     include(src/cmake/googletest.cmake)
     fetch_googletest(
--- a/src/cmake/googletest.cmake
+++ b/src/cmake/googletest.cmake
@@ -24,6 +24,10 @@ macro(fetch_googletest _download_module_path _download_root)
             ${_download_root}
     )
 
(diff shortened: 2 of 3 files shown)


Chinoske:
Good catch by CI, that was my mistake. `acore-core-interface` pulls `acore-warning-interface` with it, so the test target landed under `-Werror` for the first time and the build tripped over warnings that predate this PR (`OOC_LOCKED_GLYPH_SLOT`, the unused parameter in `UnitStub.h`).

Switched to `acore-default-interface`, which carries the compile options including `/utf-8` and nothing else. Putting the tests under the project warning level is a separate change and a bigger one.

Rebuilt on Windows after the switch, still clean, 5908 tests passing.

Chinoske:
For the record on what this was built against: the branch sits directly on master `0dda178` ("chore(DB): import pending files", 2026-08-07 11:02 UTC), no merge commit in between, and both checks were run from that same base.

- Windows, Visual Studio 2026, `RelWithDebInfo`, `BUILD_TESTING=ON`: builds clean, `unit_tests.exe` reports `AzerothCore rev. d7d70c7aaf0c+` and 5908 tests pass.
- Ubuntu 26.04, clang 21, with `-Werror -ferror-limit=1` as the CI passes them: the three files that failed earlier compile clean.

So nothing here was tested against an older tree.

CheckYourFax:
There's also another problem on the Windows build which has a simple workaround added to the error message:

```Run $ErrorActionPreference = 'Stop'
Invoke-WebRequest: D:\a\_temp\48c8cd00-8681-4efb-b78d-8b68496b688b.ps1:15
Line |
  15 |      Invoke-WebRequest -Uri 'https://master.dl.sourceforge.net/project …
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot follow an insecure redirection by default. Reissue the command specifying the -AllowInsecureRedirect
     | switch.
Error: Process completed with exit code 1.```

Chinoske:
Thanks, that one is the boost download at `windows_build.yml:67`, sourceforge redirects to plain http and PowerShell 7 refuses it now.

Different failure from what this PR fixes, so tell me which you prefer: I add `-AllowInsecureRedirect` here since it is one word and the Windows build is broken either way, or I keep this one to the test build and send that separately.

CheckYourFax:
Personally I'd make the insecure redirect flag a different PR for clarity and the ability to go back in history and find out why it was added.

Or just add an extra commit. Then if the actual redirect is fixed to only use HTTPS you can remove the flag for security reasons.

Chinoske:

PR #27069, opened 2026-08-09
feat(Core/Arena): add OnGetStartPersonalRating script hook

Helias:
## Changes Proposed:

This PR proposes changes to:

Adds an `ArenaScript` hook that lets a module change the personal rating a player receives when they join an arena team:

```cpp
virtual void OnGetStartPersonalRating(ArenaTeam* team, ObjectGuid playerGuid, uint32& personalRating) { }
```

It's called in `ArenaTeam::AddMember` right after the core computes the starting rating, so a module gets the value the player would have received and can adjust it before it's stored on the member and written to `arena_team_member`.

Today the only way to do that from a module is to let the join finish and then overwrite both the member and the row, which means deferring an update to the next world tick and leaving the roster showing the default rating in between.

The hook does nothing on its own. The default implementation is empty, so with no script registered the rating stays exactly what it is now.

I'm using it in [mod-arena-rating-memory](https://github.com/azerothcore/mod-arena-rating-memory), which gives a player back their old personal rating when they rejoin a team they used to be in, but the hook itself doesn't assume anything about that.

### AI-assisted Pull Requests

   - Tools/models used: Claude Code (Opus)

## Issues Addressed:

None.

## SOURCE:

Not applicable, this is a scripting hook and not a behaviour fix.

## Tests Performed:

Added a case to `ArenaHookDefaultsTest` covering the default: with no script registered the hook leaves `personalRating` untouched.

## How to Test the Changes:

1. Build with `-DBUILD_TESTING=1` and run `unit_tests`, `ArenaHookDefaultsTest.OnGetStartPersonalRatingKeepsCoreValue` should pass.
2. Without any module registering the hook, create an arena team and invite a player. Their personal rating in `arena_team_member` and in the arena frame is the same as before this PR (1000 on a team rated 1000+, 0 below that, or whatever `Arena.ArenaStartPersonalRating` is set to).
3. With a module implementing `OnGetStartPersonalRating` and setting a different value, the joining member gets that value in the arena frame and in `arena_team_member` right away, without a reload.

- [ ] Nothing pending.

### DIFF
--- a/src/server/game/Battlegrounds/ArenaTeam.cpp
+++ b/src/server/game/Battlegrounds/ArenaTeam.cpp
@@ -137,6 +137,8 @@ bool ArenaTeam::AddMember(ObjectGuid playerGuid)
     else if (GetRating() >= 1000)
         personalRating = 1000;
 
+    sScriptMgr->OnGetStartPersonalRating(this, playerGuid, personalRating);
+
     // xinef: sync query
     // Try to get player's match maker rating from db and fall back to config setting if not found
     CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_MATCH_MAKER_RATING);
--- a/src/server/game/Scripting/ScriptDefines/ArenaScript.cpp
+++ b/src/server/game/Scripting/ScriptDefines/ArenaScript.cpp
@@ -54,6 +54,11 @@ bool ScriptMgr::CanSaveArenaStatsForMember(ArenaTeam* team, ObjectGuid playerGui
     CALL_ENABLED_BOOLEAN_HOOKS(ArenaScript, ARENAHOOK_CAN_SAVE_ARENA_STATS_FOR_MEMBER, !script->CanSaveArenaStatsForMember(team, playerGuid));
 }
 
+void ScriptMgr::OnGetStartPersonalRating(ArenaTeam* team, ObjectGuid playerGuid, uint32& personalRating)
+{
+    CALL_ENABLED_HOOKS(ArenaScript, ARENAHOOK_ON_GET_START_PERSONAL_RATING, script->OnGetStartPersonalRating(team, playerGuid, personalRating));
+}
+
 ArenaScript::ArenaScript(char const* name, std::vector<uint16> enabledHooks)
     : ScriptObject(name, ARENAHOOK_END)
 {
--- a/src/server/game/Scripting/ScriptDefines/ArenaScript.h
+++ b/src/server/game/Scripting/ScriptDefines/ArenaScript.h
@@ -31,6 +31,7 @@ enum ArenaHook
     ARENAHOOK_ON_ARENA_START,
     ARENAHOOK_ON_BEFORE_TEAM_MEMBER_UPDATE,
     ARENAHOOK_CAN_SAVE_ARENA_STATS_FOR_MEMBER,
+    ARENAHOOK_ON_GET_START_PERSONAL_RATING,
     ARENAHOOK_END
 };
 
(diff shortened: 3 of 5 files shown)


Helias:
# Self-review — feat/arena-start-personal-rating-hook → master

> **Outcome** 0 findings — round 1 clean
> **Intent** Add an `ArenaScript` hook, `OnGetStartPersonalRating`, letting a module observe and modify the starting personal rating the core computed for a player joining an arena team, before it is stored on the member and persisted.
> **Reviewed** `6fe4f9fdc` (`feat/arena-start-personal-rating-hook` vs `upstream/master`, merge-base diff — 5 files, +23 −0)
> **By** Claude Code CLI, claude-opus-5[1m] — self-review v0.1, 2026-08-09
> **Project rules** `.agents/docs/self-review-rules.md` found and applied
> **Tested in-game** by the author — see below

## Round 1 — `6fe4f9fdc`, clean

All 5 changed files covered. Verified:

- **Call site** `ArenaTeam.cpp:140` — sits after every early-return guard and after the core's
  three-branch rating computation; nothing between it and `newMember.PersonalRating` / the
  `CHAR_INS_ARENA_TEAM_MEMBER` bind touches `personalRating`, so a module's edit is what gets
  stored and persisted.
- **No missed paths** — every join funnels through `ArenaTeam::AddMember`
  (`ArenaTeamHandler.cpp:191`, `PetitionsHandler.cpp:809`); `LoadMembersFromDB` reads an
  already-persisted value, correctly out of scope. `CONFIG_ARENA_START_PERSONAL_RATING` is read
  in exactly one place.
- **Plumbing** — enum appended before `ARENAHOOK_END`, no existing index shifts;
  `CALL_ENABLED_HOOKS` correct for a `void` out-param hook; naming mirrors the existing
  `OnGetPoints(ArenaTeam*, uint32, float&)`; declaration order matches across `ArenaScript.h`,
  `ArenaScript.cpp`, `ScriptMgr.h`. Tracing the sibling `CanSaveArenaStatsForMember` repo-wide
  found no registration list, module-facing header, changelog, or in-repo doc a new hook must
  also touch.
- **Codestyle** — `apps/codestyle/codestyle-cpp.py` passes; `-fsyntax-only` clean on both `.cpp`
  files with the project's real compile flags.
- **Test** — `ArenaHookDefaultsTest.cpp:95` matches its neighbours and asserts the real contract
  (value untouched with no script registered); auto-wired via `CollectSourceFiles`.
- **Leftovers** — none.

**Examined, not raised:** `ArenaScript.cpp:61` is 147 cols vs `.editorconfig`'s 120 — all five
sibling `CALL_ENABLED_HOOKS` lines are 131–187 cols and the linter does not check length.
`uint32&` vs `uint16 ArenaTeamMember::PersonalRating` truncates above 32767 — identical
pre-existing hazard on the `CONFIG_ARENA_START_PERSONAL_RATING` path, no new exposure.

## In-game testing

**Author reports having tested the change in-game.** Individual scenarios were not itemised.

The reviewer cannot test in-game, and this touches shared core. The coverage a reviewer would
want confirmed — all with **no module loaded**, confirming behaviour is unchanged:

1. Join a fresh arena team with `Arena.ArenaStartPersonalRating` set to a non-zero value → member
   starts at that value.
2. Pre-season-6 configuration → member starts at 1500.
3. Team rated ≥ 1000 with `Arena.ArenaStartPersonalRating = 0` → member starts at 1000; team
   rated < 1000 → starts at 0.
4. Both entry points: team creation (`ArenaTeamHandler`) **and** joining via petition
   (`PetitionsHandler`) — the second is the easy one to forget.
5. Verify the value actually landed in `arena_team_member.personalRating`, not just the UI.

Worth also checking with a module that implements the hook, that the modified value survives to
the DB row.

Issue #14094, opened 2022-12-08
Can't complete quest - Postponing the Inevitable


PerryCS:
### Current Behaviour

I stand on the edge and try to use the Interdimensional Refabricator and it doesn't find the rift.  I know where to stand, watched videos (in case I remembered wrong lol) and I also played this dungeon on retail hundreds of times across 20+ toons.

### Expected Blizzlike Behaviour

Quest completes, rift is found.

### Source

https://www.youtube.com/watch?v=XueEMYK-vag

### Steps to reproduce the problem

Go to the Nexus
Go past Anomalus
Go to edge, use quest item, it should work.  It doesn't find the rift.

### AC rev. hash/commit

AzerothCore rev. 721711d70f31 2022-12-02 08:59:47 +0000 (master branch) (Unix, RelWithDebInfo, Static)
Using SSL version: OpenSSL 1.1.1f  31 Mar 2020 (library: OpenSSL 1.1.1f  31 Mar 2020)
Using Boost version: 1.71.0
Using MySQL version: 80027
Using CMake version: 3.16.3
Compiled on: Linux 5.11.0-40-generic

### Operating system

Ubuntu 20.04.3 LTS

### Custom changes or Modules

None

PerryCS:
UPDATE: I was able to finish the quest after about 5 tries... but, not in the location where you expect to solve the quest... you need to be far back away from the ledge and keep trying all over the place.   Then it worked.   

I don't know if this is Blizz like behavior... we always went to the edge that I remember, BUT, the quest is solvable... just have to go almost into the middle of the platform where Anomalus was and do it from there.   Anywhere near the edge didn't work for me.

avarishd:
GameObject (SpellFocus) 188105, either new sniff for positioning or increase diameter to 20~ (from 15).
Position and diameter is the same for AC, Trinity and Mangos.

PR #27075, opened 2026-08-09
fix(Core/GameObject): use the full spell focus radius

### LINKED ISSUE #24177: Core/GameObject: SpellFocus distance
**Description:** GAMEOBJECT_TYPE_SPELL_FOCUS = 8 use data1: diameter (so radius*2) but according to tests in retail this is wrong data1= radius not diameter 

Proof:
image
Gamobject Forge: 173064 with data1=10

Also we get many cases where even if we know the position of the spell focus the result doesn't much retail behavior and we end up hacking the distance.

**Branch(es):** Both

**TC rev. hash/commit:** rev. 3181a1c87fdarev. 3181a1c87fda

Nyeriah:
## Changes Proposed:

`Data1` of a `GAMEOBJECT_TYPE_SPELL_FOCUS` template is the radius, not the diameter, but `GameObjectFocusCheck` halved it. Every spell focus object in the game was therefore usable only from half its intended range. The division was integer on top of that, so odd radii lost the remainder as well: a radius of 3 became 1 yard, and entry 186624 with a radius of 1 became 0.

That halving is why we keep running into spell focus objects that "need" a bigger range than the sniffed data gives them. Eight templates carry inflated `Data1` values as a result, seven inherited from the legacy import and one from https://github.com/azerothcore/azerothcore-wotlk/pull/4075. The second commit puts them back to their sniffed radius:

| entry | name | before | after |
|---|---|---|---|
| 20919 | Tuber Node | 100 | 50 |
| 184643 | Void Conduit Spell Focus | 90 | 50 |
| 184750 | Unguarded Summoning Site | 100 | 50 |
| 185144 | Bleeding Hollow Forge Spell Focus | 14 | 5 |
| 190033 | Alystros the Verdant Keeper Spell Focus | 80 | 50 |
| 190224 | Scourge Mummy Fire | 15 | 5 |
| 190731 | Scourgewagon | 15 | 10 |
| 192011 | Thane Ufrang the Mighty's Spell Focus | 17 | 5 |

Every spawn of those eight sits at the exact same coordinates TrinityCore uses, except one of the two 192011 spawns, so these radii are known to work at those positions.

This PR proposes changes to:

### AI-assisted Pull Requests

## Issues Addressed:

None open on our side.

## SOURCE:

Core change cherry-picked from https://github.com/TrinityCore/TrinityCore/commit/2de20560fe7caccd57fbfd1b17856fac43d5f921, committed with `--author` crediting Killyana. It closed https://github.com/TrinityCore/TrinityCore/issues/24177, which carries a retail screenshot of Forge 173064 (`Data1` = 10) being used at 10 yards. The fix has been in TrinityCore since February 2020 and is unchanged there today.

## Tests Performed:

This PR has been:

## How to Test the Changes:

The core change only widens a range check, so nothing that works today can stop working from the core commit alone. The DB commit is where the regression surface is, since three of those templates end up with a smaller effective radius than they have now.

1. Stand next to an anvil or forge and start moving away while trying to craft. Before the change it cuts out at 5 yards, after it at 10.
2. Bleeding Hollow Forge, Hellfire Peninsula (185144): effective radius goes from 7 to 5 yards. Check the forge is still usable from where you would naturally stand at it.
3. Thane Ufrang the Mighty's Spell Focus (192011), quest Shadow Vault Decree: effective radius goes from 8 to 5 yards. We spawn two of these 11 yards apart, so their circles no longer overlap. Confirm the quest still completes from both.
4. Scourge Mummy Fire (190224): effective radius goes from 7 to 5 yards.
5. Regression sweep on anything else that uses a spell focus and is not in the table above, since all of those gain range rather than lose it: moonwells, summoning sites, quest braziers, Lunar Festival spotlights.

- [ ] Entry 30 "Dalaran Forge Spell Focus" is ours, TrinityCore has no equivalent template, so there is no sniffed value to restore it to. It is left at `Data1` = 4 and goes from 2 to 4 effective yards. That is an improvement but probably still short of the real thing, and it may want a follow-up once someone can check it on a reference server.
- [ ] https://github.com/azerothcore/azerothcore-wotlk/pull/24518 cloned entry 30 with `Data1` = 16 to make the Forge of Fate reachable. It was closed unmerged, and with this fix in place it either becomes unnecessary or needs a much smaller bump.

### DIFF
--- a/data/sql/updates/pending_db_world/rev_1786298744183384100.sql
+++ b/data/sql/updates/pending_db_world/rev_1786298744183384100.sql
@@ -1,0 +1,11 @@
+--
+-- Spell focus radius is no longer halved by the core, so the inflated values that compensated for
+-- that go back to their sniffed radius.
+UPDATE `gameobject_template` SET `Data1` = 50 WHERE `entry` = 20919; -- Tuber Node
+UPDATE `gameobject_template` SET `Data1` = 50 WHERE `entry` = 184643; -- Void Conduit Spell Focus
+UPDATE `gameobject_template` SET `Data1` = 50 WHERE `entry` = 184750; -- Unguarded Summoning Site
+UPDATE `gameobject_template` SET `Data1` = 5 WHERE `entry` = 185144; -- Bleeding Hollow Forge Spell Focus
+UPDATE `gameobject_template` SET `Data1` = 50 WHERE `entry` = 190033; -- Alystros the Verdant Keeper Spell Focus
+UPDATE `gameobject_template` SET `Data1` = 5 WHERE `entry` = 190224; -- Scourge Mummy Fire
+UPDATE `gameobject_template` SET `Data1` = 10 WHERE `entry` = 190731; -- Scourgewagon
+UPDATE `gameobject_template` SET `Data1` = 5 WHERE `entry` = 192011; -- Thane Ufrang the Mighty's Spell Focus
--- a/src/server/game/Grids/Notifiers/GridNotifiers.h
+++ b/src/server/game/Grids/Notifiers/GridNotifiers.h
@@ -691,7 +691,7 @@ namespace Acore
             if (go->GetGOInfo()->spellFocus.focusId != i_focusId)
                 return false;
 
-            float dist = (float)((go->GetGOInfo()->spellFocus.dist) / 2);
+            float const dist = go->GetGOInfo()->spellFocus.dist;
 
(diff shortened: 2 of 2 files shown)


Gultask (approved):

Gultask:
Likely fixes or alleviates https://github.com/azerothcore/azerothcore-wotlk/issues/14094?

Nyeriah:
Yes as it’s a spell focus

PR #27076, opened 2026-08-09
fix(Core/Spells): Blood Tap brings back a spent blood rune again

Chinoske:
## Changes Proposed:

Closes #27073. Blood Tap should activate a spent blood rune and turn it into a death rune. With one blood rune ready and the other spent, it left the spent one alone and converted the one that was already available instead.

`Spell::EffectActivateRune` picked the rune by the type it is currently showing:

```cpp
if (player->GetRuneCooldown(j) && player->GetCurrentRune(j) == RuneType(m_spellInfo->Effects[effIndex].MiscValue))
{
    if (m_spellInfo->Id == 45529)
        if (player->GetBaseRune(j) != RuneType(m_spellInfo->Effects[effIndex].MiscValueB))
            continue;
```

For Blood Tap (45529) `MiscValue` is `RUNE_DEATH`, so a spent rune only matched once it had already been converted. A plain spent blood rune never did. The extra loop underneath was there to cover that, but it only fires when **both** runes of the pair are spent, which is why the reporter saw it work only with two death runes and one empty.

With nothing activated, effect 1 of the spell (the `SPELL_AURA_CONVERT_RUNE` aura) walks the runes looking for one that is off cooldown, so it converted the rune the player already had available. That is the "it overrides my available rune" in the report.

Blood Tap now picks by rune slot, the way TrinityCore does it, so both blood slots qualify whether or not they have turned into death runes yet:

```cpp
uint32 firstCooldown = player->GetRuneCooldown(0);
uint32 secondCooldown = player->GetRuneCooldown(1);

uint8 bloodRune = 0;
if (secondCooldown && (!firstCooldown || secondCooldown < firstCooldown))
    bloodRune = 1;
```

One ready and one spent activates the spent one; both spent activates whichever refreshes first. That covers what the removed loop did, and it also drops a couple of problems that loop had: it stepped one slot at a time while reading `l + 1`, so it compared slots (0,1), (1,2), (2,3) instead of the real rune pairs, and at `l = 5` it read index 6, past the end of the array.

A rune that is already up is left alone, so its grace period is not thrown away and the next one spent from that slot still refreshes on time.

### Known limitation

[...]

PR #27098, opened 2026-08-11
fix(Core/Spells): Blood Tap converts the rune it brings back

### LINKED ISSUE #27073: [DK]Blood tap
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9995

Triage Notes: 

reproduce:
- make dk level 80 and .learn all my class
- .go xyz -8722.548 355.6318 101.01977 to test on dummy
- use 2x Blood Strike , both Blood Runes become Death Runes on cooldown
- use Obliterate twice to consume your Frost Runes ,this is to isolate the Blood Rune situation
- wait for one Death Rune to become available (comes off cooldown),now you have: 1 available Death Rune + 1 Death Rune still on cooldown
- use Blood Tap and watch your rune bar
- observe: Blood Tap resets/overrides the available Death Rune instead of activating the cooldown one, expected: Blood Tap should activate the empty/cooldown Blood Rune slot, giving you 2 available Death Runes

video linked in og report

Original Post Below: 

---
### Current Behaviour

Blood tap is currently overriding a death rune if you have an available death rune and a blood rune on cooldown. It is activating an empty rune only when you have 2 death runes and one is empty.

https://github.com/user-attachments/assets/c22972a3-8ad0-4ae2-830b-4fce2a7ab9be

### Expected Behaviour

As the tooltip implies it should convert a "blood rune" into a death rune so it should convert an empty blood rune into an available death rune.

Don't have

Make a DK
teleport to a dummy
get talents that grant death runes upon using blood strike/pestilence
strike the dummy with 2 blood strikes, use obliterates to remove your active frost runes so that unbreakable armor falls into a death blood rune, use blood tap.

### Extra Notes

_No response_

### Customizations

_No response_

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/f8f667bff5ad84017c1921c09c4a71136b4396ee

### Operating system

Ubuntu 24.04

### Modules

### Server

ChromieCraft

Chinoske:
## Changes Proposed:

Blood Tap chooses which rune to refresh by the type the rune is currently showing, and the spell's
own DBC row makes that the wrong question to ask. Effects of 45529, read from `Spell.dbc`:

| # | Effect | MiscValue | MiscValueB |
| --- | --- | --- | --- |
| 0 | `SPELL_EFFECT_ACTIVATE_RUNE` | 3 (`RUNE_DEATH`) | 0 (`RUNE_BLOOD`) |
| 1 | `SPELL_EFFECT_APPLY_AURA`, `SPELL_AURA_CONVERT_RUNE` | 0 (`RUNE_BLOOD`) | 3 (`RUNE_DEATH`) |
| 2 | `SPELL_EFFECT_ENERGIZE` | 6 (`POWER_RUNIC_POWER`) | 0 |

Effect 0 asks for a death rune, so filtering on the current type only ever finds runes that are
already converted. A blood rune that is spent and still showing blood never matches. There is a
fallback loop below the main one that does look at `MiscValueB`, but it only fires when both runes
of the pair are spent and it `break`s as soon as one isn't, which is why the reporter sees it work
only with two death runes and one of them empty. On top of that it pairs slots by walking one at a
time, so it forms (1,2) and (3,4), and those cross rune types (slots are 0-1 blood, 2-3 unholy,
4-5 frost), and it reads `runes[6]` when the index reaches 5. `MAX_RUNES` is 6 and the accessors in
`Player.h` index `m_runes->runes[index]` with no bounds check.

Effect 1 is broken independently. It walks the slots from index 0 and converts the first blood rune
that isn't on cooldown, which has nothing to do with the rune effect 0 acted on. When effect 0
refreshes nothing, that's the rune the player already had up, which is the "it overwrites my
available rune" part of the report.

Both effects now live in a spell/aura script pair, the way TrinityCore handles this on their 3.3.5
branch. Nothing else changes about how the aura works: it is still the DBC's own
`SPELL_AURA_CONVERT_RUNE`, just aimed at the right slot.

The spell script takes effect 0. It selects by the slot's base rune type, taken from `MiscValueB` so
nothing about which slots are blood is hardcoded; `Player::IsBaseRuneSlotsOnCooldown` already walks
the slots the same way. It skips slots that are already up, so their grace period survives
(`SetGracePeriod` rides along with every `SetRuneCooldown(x, 0)`), and refreshes the one with the
lowest remaining cooldown. It honours the effect's value, which is 1 for Blood Tap today. Then it
hands that slot to the aura script.

The aura script takes effect 1 and converts exactly that slot, passing its own `AuraEffect` along.
That is what keeps the revert selective: `RemoveRunesByAuraEffect` only restores slots whose
`ConvertAura` matches, so when Blood Tap's 20 seconds are up it reverts its own conversion and
nothing else.

It converts unconditionally, which is worth spelling out because the common case looks like it
shouldn't. Reaping converts a blood rune the moment it is **spent**, so the rune Blood Tap brings
back is usually already a death rune owned by that talent, and Blood Tap takes that ownership over.
That is deliberate, and it is what upstream does. `RestoreBaseRune` turns a rune back to blood as
soon as it is spent when the owning aura is passive; the death rune talents are passive and Blood
Tap is not. Leave the talent the slot and Blood Tap's death rune dies on first use with its buff
still showing, which is not what the spell promises. The cost of taking it is the other direction: a
rune the player never spends reverts at Blood Tap's 20 seconds even if the talent had time left.
Since the talent resets its 30 second timer when the rune goes on cooldown, and a rune's cooldown is
10 seconds, the talent always has at least 20 seconds left at that point, so what is lost is only
the surplus.

`Spell::SetRuneState` is new. `SMSG_SPELL_GO` reports the rune mask from before the cast, and a
script taking over `SPELL_EFFECT_ACTIVATE_RUNE` has to record it itself.

Taking Blood Tap out of `Spell::EffectActivateRune` means its branch there is gone, and the fallback
loop with it. Empower Rune Weapon and every other spell reaching that function keep the generic loop
untouched, and its remaining spell id is now a named constant instead of a bare literal.

TrinityCore also fixed the effect 0 half separately on 3.3.5 in
[868dcb0](https://github.com/TrinityCore/TrinityCore/commit/868dcb0): `GetCurrentRune` to
`GetBaseRune` plus a bound on the loop. That keeps the "both runes spent" requirement, so on its own
it wouldn't cover the case in the issue.

Supersedes #27076, which I closed myself.

This PR proposes changes to:

### AI-assisted Pull Requests

   - Tools/models used: Claude Code with Claude Opus 5, for both the code change and this
     description. It ran a fixed pipeline of agent skills, one per phase, each leaving a document
     behind: `fetch-ticket` for the issue, `azerothcore-research` for the research below,
     `refine-ticket` for the requirements, `create-implementation-plan` for the plan, then the plan
     executed step by step, `generate-pr-description` for this text, and `self-review` before
     publishing, which put the commit through a fresh-context reviewer and produced the report
     below. Every piece of game data quoted here came from azerothMCP v2, a local MCP server that
     reads this server's own DBC files, world DB and source tree, so the spell values are what the
     server actually runs and not a wiki transcription.

## Issues Addressed:
- Closes #27073
- Escalated from [chromiecraft#9995](https://github.com/chromiecraft/chromiecraft/issues/9995)

## SOURCE:

- The script pair comes from TrinityCore [`d1dc0e2dc1`](https://github.com/TrinityCore/TrinityCore/commit/d1dc0e2dc1) ("Scripts/DK: correctly handle Blood Tap") by ccrs, credited with a `Co-authored-by` trailer on the follow-up commit. The rune selection is rewritten against AC.
- The spell's own `Spell.dbc` row (the effect table above), which is where the tooltip text comes
  from: "Instantly activates a blood rune and converts it into a death rune for the next 20 sec."
- [TrinityCore 868dcb0](https://github.com/TrinityCore/TrinityCore/commit/868dcb0), [PR 17132](https://github.com/TrinityCore/TrinityCore/pull/17132), the
  equivalent fix on their 3.3.5 branch.
- The video in the issue.

## Tests Performed:
This PR has been:

Tested in game on a level 80 Death Knight, all confirmed working:

- The scenario in the issue: two Blood Strikes, wait for one rune to come back, Blood Tap. It
  activates the rune that was still spent, and that same rune is the one that turns into a death
  rune.
- Both blood runes up: nothing is reset, one turns into a death rune, and it goes back to blood
  when the 20 seconds are up.
- With Reaping at 100%: spend a blood rune so the talent converts it, Blood Tap it back, then spend
  it before the aura expires. The rune stays a death rune while spent and comes back as one, which
  is the case that needs Blood Tap to own the conversion rather than the talent.
- All six runes spent: Blood Tap brings back one blood rune, nothing else moves.
- Regression, Empower Rune Weapon: still restores all six runes, with and without runes spent
  across types.
- Selection order: with both blood runes spent at different times, it picks the one closest to
  coming back, not the one with the lower index. Checked both ways round.

A rune that is already up keeping its grace period is not observable in game: the loop skips slots
with no cooldown before touching anything, so `SetGracePeriod` never runs on them.

worldserver builds clean and both `codestyle-cpp.py` and `codestyle-sql.py` pass.

## Self-review

**Outcome** 5 findings — 5 fixed — final round clean

**Reviewed** `573068b35` (`fix/dk-blood-tap-base-rune` vs `master`, merge-base diff — 4 files, +126 −24)

**By** Claude Code (Claude Agent SDK), claude-opus-5 — self-review v0.5, 2026-08-11

`codestyle-cpp.py` and `codestyle-sql.py` pass, worldserver builds clean, three rounds by a
fresh-context reviewer.

<details>

**Intent** Blood Tap (spell 45529) should instantly activate one spent blood rune and convert that
same rune into a death rune for 20 s. It failed to recognise a spent blood rune that had not been
converted yet, and its aura converted whichever blood rune happened to be available rather than the
one the spell acted on.

**Project rules** `.agents/docs/self-review-rules.md` found and applied

## Round 1 — 4 findings

1. Taking the effect over dropped the `CastSpell(47804)` the core handler ran at the end, which
   exists only to push the rune list to the client → **fixed**: cast from the script, same position.
2. `AddRuneByAuraEffect` rebinds the slot's `ConvertAura` unconditionally, so Blood Tap took over a
   slot Reaping owned → **fixed**, then revised in round 2.
3. `SPELL_DK_BLOOD_TAP` was added to the enum and never used → **fixed**: removed, and
   `SPELL_DK_BLOOD_TAP_VISUAL = 47804` took its place.
4. The SQL file used a date-prefixed name instead of the `rev_<timestamp>.sql` that `create_sql.sh`
   generates → **fixed**: renamed.

## Round 2 — 1 finding

5. The fix for finding 2 overshot. `RestoreBaseRune` reverts a rune to blood as soon as it is spent
   when the owning aura is passive, and the death rune talents are passive while Blood Tap is not,
   so leaving the talent the slot meant Blood Tap's death rune died on first use with its buff still
   showing → **fixed**: the guard was removed and the aura converts unconditionally, as upstream
   does. Accepted cost: a rune the player never spends reverts at Blood Tap's 20 s even if the
   talent had time left.

## Round 3 — clean

No findings. The reviewer separately confirmed that Blood Tap's aura cannot be force-removed by a
rune restore, that the talents' own revert paths are gated on spell icons Blood Tap does not match,
and that no `AuraEffect` pointer can dangle across the conversion.

</details>

## How to Test the Changes:

Level 80 Death Knight with Reaping (`.learn all my class`), hitting a dummy. Watch the rune bar
after each step.

1. Cast Blood Strike twice, wait for one blood rune to come back and convert to a death rune, then
   cast Blood Tap. Both runes should end up available.
2. With one blood slot up and the other spent, Blood Tap should refresh the spent one. Try it with
   slot 0 up and with slot 1 up.
3. With both blood slots spent on different cooldowns, it should refresh the one closest to coming
   back.
4. With both blood slots already up, Blood Tap should not reset anything, and the next rune you
   spend from those slots should still come back in its normal time.
5. With all six runes spent, Blood Tap should refresh a blood one.
6. With Reaping talented, spend a blood rune so the talent converts it, then Blood Tap it back and
   wait out the 20 seconds. The rune should stay a death rune: Blood Tap must not revert a
   conversion that belongs to the talent.
7. Cast Blood Tap on a blood rune that is not converted, wait out the 20 seconds, and the rune
   should go back to blood.
8. Regression: Empower Rune Weapon should still hand back every frost and death rune.
9. Regression: the client rune bar should match the server right after the cast, and the refreshed
   rune should be spendable straight away.

<details>
<summary>Research notes: spell data, rune system and prior art</summary>

### Spell 45529, from `Spell.dbc`

| Field | Value |
| --- | --- |
| Tooltip | "Instantly activates a blood rune and converts it into a death rune for the next 20 sec. This rune counts as a blood, frost or unholy rune." |
| RecoveryTime | 60000 ms |
| DurationIndex | 18, which is 20000 ms in `SpellDuration.dbc` |
| PowerType | -2, runes |
| SpellFamily | 15 DEATHKNIGHT, flags `[8, 0, 0]` |
| Cast | instant, self, no cost |

`damage` for effect 0 is `BasePoints + DieSides` = 1, so `count` starts at 1. One rune.

### The rune system, the part that matters

Six fixed slots, `MAX_RUNES` 6, base type wired in `runeSlotTypes` (`Player.cpp`): slots 0-1 blood,
2-3 unholy, 4-5 frost. Each slot carries `BaseRune`, which is permanent and survives every
conversion, `CurrentRune`, which is what it shows now, `Cooldown`, where 0 means up, and
`ConvertAura`, the aura that converted it so it can be reverted.

So a blood rune converted to death still has `BaseRune == RUNE_BLOOD`. Filtering on
`GetCurrentRune` loses those, `GetBaseRune` doesn't.

Reaping and Blood of the North are what turn blood into death, through `SPELL_AURA_CONVERT_RUNE`.
`spell_dk_death_rune::HandleProc` converts the rune the moment it goes on cooldown, one per proc,
and each conversion is owned by that talent's own `AuraEffect` on a 30 second timer. That ownership
is why the aura script has to leave an already-converted slot alone.

### The bug chain

1. The main loop requires `GetCurrentRune(j) == MiscValue`, and `MiscValue` is `RUNE_DEATH`. A spent
   blood rune that hasn't been converted yet never matches.
2. The fallback loop, the one that does look at `MiscValueB`, needs both runes of the pair spent and
   `break`s otherwise. Hence the report: it only activates an empty rune when you have two death
   runes and one of them is empty.
3. With nothing activated, effect 1 walks the slots from index 0 and converts the first one not on
   cooldown, which is the rune the player already had up.

Three more defects live in that same fallback loop, and all three go away with it: it reads
`runes[6]` when the index reaches 5 and the accessors don't bounds check, it pairs (1,2) and (3,4)
which cross rune types, and it can reset a slot that is already up, which throws away its grace
period since `SetGracePeriod` rides with every `SetRuneCooldown(x, 0)`.

### Prior art

[TrinityCore 868dcb0](https://github.com/TrinityCore/TrinityCore/commit/868dcb0), [PR 17132](https://github.com/TrinityCore/TrinityCore/pull/17132),
2016-05-15, on their 3.3.5 branch: `GetCurrentRune` to `GetBaseRune`, plus `l + 1 < MAX_RUNES` on
the loop. AzerothCore never ported it. That fix keeps the "both runes of the pair spent"
requirement, so on its own it wouldn't cover the case in this issue.

### Patch history

| When | What |
| --- | --- |
| 3.3.5a | Blood Tap as the DBC has it, activate a blood rune and convert it to death. This is the target. |
| Cataclysm on | Redesigned into charges and the rune system changed. Not applicable here. |
| 2016-05-15 | TrinityCore fixes the selection on their 3.3.5 branch. |
| Today | AzerothCore master still filters on `GetCurrentRune` with the unbounded loop. |

</details>

- [ ] Unrelated to this PR, but found while working on it and worth writing down somewhere:
      `Player::ResyncRunes` sends `uint8(255 - (GetRuneCooldown(i) * 51))`, a 0-5 scale formula fed
      milliseconds, so every rune still on cooldown would get a garbage byte. The call is commented
      out in master, so nothing triggers it today. The correct formula is already at
      `Spell.cpp:4903`. Worth noting for whoever picks this up: TrinityCore calls `ResyncRunes` from
      `Player::SetRuneCooldown`, so every rune state change carries it, which is why they never
      needed a loose call in `EffectActivateRune`.

### DIFF
--- a/src/server/game/Spells/SpellEffects.cpp
+++ b/src/server/game/Spells/SpellEffects.cpp
@@ -67,6 +67,11 @@
 //  see: https://github.com/azerothcore/azerothcore-wotlk/issues/9766
 #include "GridNotifiersImpl.h"
 
+enum DeathKnightSpells
+{
+    SPELL_DK_EMPOWER_RUNE_WEAPON    = 47568
+};
+
 pEffect SpellEffects[TOTAL_SPELL_EFFECTS] =
 {
     &Spell::EffectNULL,                                     //  0
@@ -5808,41 +5813,19 @@ void Spell::EffectActivateRune(SpellEffIndex effIndex)
 
     uint32 count = damage;
     if (count == 0) count = 1;
+
     for (uint32 j = 0; j < MAX_RUNES && count > 0; ++j)
     {
         if (player->GetRuneCooldown(j) && player->GetCurrentRune(j) == RuneType(m_spellInfo->Effects[effIndex].MiscValue))
         {
-            if (m_spellInfo->Id == 45529)
-                if (player->GetBaseRune(j) != RuneType(m_spellInfo->Effects[effIndex].MiscValueB))
-                    continue;
             player->SetRuneCooldown(j, 0);
             player->SetGracePeriod(j, player->IsInCombat()); // xinef: reset grace period
             --count;
         }
     }
 
-    // Blood Tap
-    if (m_spellInfo->Id == 45529 && count > 0)
-    {
-        for (uint32 l = 0; l < MAX_RUNES && count > 0; ++l)
-        {
-            // Check if both runes are on cd as that is the only time when this needs to come into effect
-            if ((player->GetRuneCooldown(l) && player->GetCurrentRune(l) == RuneType(m_spellInfo->Effects[effIndex].MiscValueB)) && (player->GetRuneCooldown(l + 1) && player->GetCurrentRune(l + 1) == RuneType(m_spellInfo->Effects[effIndex].MiscValueB)))
-            {
-                // Should always update the rune with the lowest cd
-                if (player->GetRuneCooldown(l) >= player->GetRuneCooldown(l + 1))
-                    l++;
-                player->SetRuneCooldown(l, 0);
-                player->SetGracePeriod(l, player->IsInCombat()); // xinef: reset grace period
-                --count;
-            }
-            else
-                break;
-        }
-    }
-
     // Empower rune weapon
-    if (m_spellInfo->Id == 47568)
+    if (m_spellInfo->Id == SPELL_DK_EMPOWER_RUNE_WEAPON)
     {
         // Need to do this just once
         if (effIndex != 0)
--- a/src/server/scripts/Spells/spell_dk.cpp
+++ b/src/server/scripts/Spells/spell_dk.cpp
@@ -44,6 +44,7 @@ enum DeathKnightSpells
     SPELL_DK_BLACK_ICE_R1                       = 49140,
     SPELL_DK_BLOOD_BOIL_TRIGGERED               = 65658,
     SPELL_DK_BLOOD_GORGED_HEAL                  = 50454,
+    SPELL_DK_BLOOD_TAP_VISUAL                   = 47804,
     SPELL_DK_BLOOD_PRESENCE                     = 48266,
     SPELL_DK_CORPSE_EXPLOSION_TRIGGERED         = 43999,
     SPELL_DK_CORPSE_EXPLOSION_VISUAL            = 51270,
@@ -456,6 +457,115 @@ class spell_dk_chains_of_ice_aura : public AuraScript
     {
         OnEffectUpdatePeriodic += AuraEffectUpdatePeriodicFn(spell_dk_chains_of_ice_aura::HandlePeriodic, EFFECT_1, SPELL_AURA_PERIODIC_DUMMY);
     }
+};
+
+// 45529 - Blood Tap
+class spell_dk_blood_tap_aura : public AuraScript
+{
+    PrepareAuraScript(spell_dk_blood_tap_aura);
+
+public:
+    void SetRuneIndex(uint8 index) { _runeIndex = index; }
+
+private:
+    void HandleApply(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/)
+    {
+        // The default handler walks the slots from 0, so it can convert a rune other than the
+        // one the spell just brought back. The spell script tells us which one it was.
+        PreventDefaultAction();
+
+        if (_runeIndex >= MAX_RUNES)
+            return;
+
+        Player* player = GetTarget()->ToPlayer();
+        if (!player || !player->IsClass(CLASS_DEATH_KNIGHT, CLASS_CONTEXT_ABILITY))
+            return;
+
+        // Taking ownership of the slot matters even when it already shows a death rune. Reaping
+        // and friends are passive, so RestoreBaseRune turns their runes back to blood as soon as
+        // they are spent, while ours survive being spent for the full duration.
+        player->AddRuneByAuraEffect(_runeIndex, RuneType(aurEff->GetMiscValueB()), aurEff);
+    }
+
+    void Register() override
+    {
+        OnEffectApply += AuraEffectApplyFn(spell_dk_blood_tap_aura::HandleApply, EFFECT_1, SPELL_AURA_CONVERT_RUNE, AURA_EFFECT_HANDLE_REAL);
+    }
+
+    uint8 _runeIndex = MAX_RUNES;
+};
+
+class spell_dk_blood_tap : public SpellScript
+{
+    PrepareSpellScript(spell_dk_blood_tap);
+
+    void HandleActivateRune(SpellEffIndex effIndex)
+    {
+        PreventHitDefaultEffect(effIndex);
+
+        Player* player = GetCaster()->ToPlayer();
+        if (!player || !player->IsClass(CLASS_DEATH_KNIGHT, CLASS_CONTEXT_ABILITY))
+            return;
+
+        GetSpell()->SetRuneState(player->GetRunesState());
+
+        // The DBC asks for a death rune here, so matching on the rune's current type misses a
+        // spent blood rune that has not been converted yet. Go by the slot's base type instead.
+        RuneType const baseType = RuneType(GetSpellInfo()->Effects[EFFECT_0].MiscValueB);
+        int32 count = GetEffectValue();
+        if (count <= 0)
+            count = 1;
+
+        while (count > 0)
+        {
+            uint8 spent = MAX_RUNES;
+            for (uint8 i = 0; i < MAX_RUNES; ++i)
+            {
+                // A rune that is already up keeps its grace period, so leave it alone
+                if (player->GetBaseRune(i) != baseType || !player->GetRuneCooldown(i))
+                    continue;
+                if (spent == MAX_RUNES || player->GetRuneCooldown(i) < player->GetRuneCooldown(spent))
+                    spent = i;
+            }
+
+            if (spent == MAX_RUNES)
+                break;
+
+            player->SetRuneCooldown(spent, 0);
+            player->SetGracePeriod(spent, player->IsInCombat()); // xinef: reset grace period
+            _runeIndex = spent;
+            --count;
+        }
+
+        // Nothing was spent, so fall back to what the default aura handler would have done and
+        // convert a rune that is already up
+        if (_runeIndex >= MAX_RUNES)
+            for (uint8 i = 0; i < MAX_RUNES; ++i)
+                if (player->GetBaseRune(i) == baseType && !player->GetRuneCooldown(i) && player->GetCurrentRune(i) == baseType)
+                {
+                    _runeIndex = i;
+                    break;
+                }
+
+        // Carries the rune list to the client, which otherwise keeps showing the rune we just
+        // handed back as still on cooldown. The core handler cast this too
+        player->CastSpell(player, SPELL_DK_BLOOD_TAP_VISUAL, true);
+    }
+
+    void SetRuneIndex(SpellEffIndex /*effIndex*/)
+    {
+        if (Aura* aura = GetHitAura())
+            if (spell_dk_blood_tap_aura* script = aura->GetScript<spell_dk_blood_tap_aura>("spell_dk_blood_tap"))
+                script->SetRuneIndex(_runeIndex);
+    }
+
+    void Register() override
+    {
+        OnEffectLaunch += SpellEffectFn(spell_dk_blood_tap::HandleActivateRune, EFFECT_0, SPELL_EFFECT_ACTIVATE_RUNE);
+        OnEffectHitTarget += SpellEffectFn(spell_dk_blood_tap::SetRuneIndex, EFFECT_1, SPELL_EFFECT_APPLY_AURA);
+    }
+
+    uint8 _runeIndex = MAX_RUNES;
 };
 
 // 50452 - Bloodworm
@@ -3024,6 +3134,7 @@ void AddSC_deathknight_spell_scripts()
     RegisterSpellScript(spell_dk_anti_magic_zone);
     RegisterSpellScript(spell_dk_blood_boil);
     RegisterSpellScript(spell_dk_blood_gorged);
+    RegisterSpellAndAuraScriptPair(spell_dk_blood_tap, spell_dk_blood_tap_aura);
     RegisterSpellScript(spell_dk_corpse_explosion);
     RegisterSpellScript(spell_dk_death_coil);
     RegisterSpellScript(spell_dk_death_gate);
--- a/data/sql/updates/pending_db_world/rev_17864847660000000.sql
+++ b/data/sql/updates/pending_db_world/rev_17864847660000000.sql
@@ -1,0 +1,5 @@
+-- Blood Tap (45529) is now handled by a spell/aura script pair, so that the rune the spell
+-- brings back is the same one its aura converts into a death rune.
+DELETE FROM `spell_script_names` WHERE `spell_id` = 45529 AND `ScriptName` = 'spell_dk_blood_tap';
+INSERT INTO `spell_script_names` (`spell_id`, `ScriptName`) VALUES
+(45529, 'spell_dk_blood_tap');
--- a/src/server/game/Spells/Spell.h
+++ b/src/server/game/Spells/Spell.h
@@ -580,6 +580,9 @@ public:
     void SetDelayStart(uint64 m_time) { m_delayStart = m_time; }
     uint64 GetDelayMoment() const { return m_delayMoment; }
     uint64 GetDelayTrajectory() const { return m_delayTrajectory; }
+    // SMSG_SPELL_GO reports the rune mask from before the cast, so a script taking over
+    // SPELL_EFFECT_ACTIVATE_RUNE has to record it itself
+    void SetRuneState(uint8 state) { m_runesState = state; }
 
     uint64 CalculateDelayMomentForDst() const;
     void RecalculateDelayMomentForDst();


sudlud:
thanks for sending the PR! looks like a very structured approach, thanks also for following up on the improvement suggestions.

just wondering - you did not tick the box at `Tested in-game by the author.` - is this a mistake or have you not tested the PR yet? it is also ok to tick both the `Tested in-game by the author.` box and `requires further testing`, this would show that you tested it but might ask for more testing of edge cases etc.

Chinoske:
Tested this in game on a level 80 DK, updating the description to match. What I covered:

The scenario from the issue: two Blood Strikes, wait for one rune to come back, Blood Tap. It brings back the one that was still spent, and that same rune is the one that turns into a death rune. Both runes end up available.

Both blood runes up: nothing gets reset, one turns into a death rune, and it goes back to blood when the 20 seconds are up.

With Reaping at 100%, the case I cared most about: spend a blood rune so the talent converts it, Blood Tap it back, then spend it straight away with Icy Touch. The rune stays a death rune while spent and comes back as one. That's the case that needs Blood Tap to own the conversion instead of the talent, since `RestoreBaseRune` reverts a rune the moment it's spent when the aura that converted it is passive, and the death rune talents are passive while Blood Tap isn't.

All six runes spent: Blood Tap brings back one blood rune and nothing else moves.

Selection order: spend both blood runes, let the first one come back, spend it again so the second is now the one closest to refreshing, then Blood Tap. It picks the second one, so it goes by remaining cooldown rather than by slot order.

Empower Rune Weapon still restores all six runes, which is the regression I most wanted to rule out since it shares `EffectActivateRune`.

The one thing I didn't test in game is a rune that's already up keeping its grace period, since there's nothing to observe: the loop skips slots with no cooldown before it touches anything, so `SetGracePeriod` never runs on them.

sudlud:
Thanks for testing! Also please try to avoid force pushes if possible, because it makes it harder to keep track of changes in a PR. just add new commits, they get squashed on merge anyways.

Chinoske:
I use all the recommended tools, yet people always criticize—ugh, never mind... If only they’d say what they’re looking for, what would be better, or how it ought to look—but they don't say a thing; they just criticize. They ought to explain how it should look and take the time to test it and see if improvements are possible, rather than just criticizing when the very same tools are being used.

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -458,6 +459,115 @@ class spell_dk_chains_of_ice_aura : public AuraScript
     }
 };
 
+// 45529 - Blood Tap
+class spell_dk_blood_tap_aura : public AuraScript
```
sudlud, line 463: this is TrinityCore's script pair, from `d1dc0e2dc1` ("Scripts/DK: correctly handle Blood Tap"), so it needs the upstream credit even with the selection logic rewritten. can you add a follow-up commit with this trailer?

Co-authored-by: ccrs <ccrs@users.noreply.github.com>

please don't amend or force push for this, a new commit on top is fine. worth ticking the cherry-pick box and naming that commit under SOURCE as well.

src/server/scripts/Spells/spell_dk.cpp
```diff
@@ -458,6 +459,115 @@ class spell_dk_chains_of_ice_aura : public AuraScript
...
+
+public:
+    void SetRuneIndex(uint8 index) { _runeIndex = index; }
+
+private:
+    void HandleApply(AuraEffect const* aurEff, AuraEffectHandleModes /*mode*/)
+    {
+        // The default handler walks the slots from 0, so it can convert a rune other than the
+        // one the spell just brought back. The spell script tells us which one it was.
+        PreventDefaultAction();
```
sudlud, line 475: maybe move `PreventDefaultAction()` below the `_runeIndex >= MAX_RUNES` check? an aura restored on login never runs the spell script, so it applies with the sentinel still set and converts nothing, where master converted a rune.

letting the default handler run in that case looks safe, effect 1 matches on `RUNE_BLOOD` and any slot the script skipped is either on cooldown or already a death rune.

src/server/game/Spells/SpellEffects.cpp
```diff
@@ -67,6 +67,11 @@
 //  see: https://github.com/azerothcore/azerothcore-wotlk/issues/9766
 #include "GridNotifiersImpl.h"
 
+enum DeathKnightSpells
+{
+    SPELL_DK_EMPOWER_RUNE_WEAPON    = 47568
```
sudlud, line 72: can we add `SPELL_DK_BLOOD_TAP_VISUAL = 47804` here too? it's still a bare literal at the bottom of `EffectActivateRune`, and you already named it in `spell_dk.cpp`.

Chinoske:
all three are in `83baeae3e`, new commit on top, no force push.

moved `PreventDefaultAction()` below the `_runeIndex >= MAX_RUNES` check. an aura restored on login now falls through to the default handler instead of applying with the sentinel set and converting nothing.

added `SPELL_DK_BLOOD_TAP_VISUAL = 47804` in `SpellEffects.cpp` and used it in `EffectActivateRune`.

credit trailer for `d1dc0e2dc1` is on that commit, and the cherry-pick box is ticked with the commit named under SOURCE.

Chinoske:
# Self-review — fix/dk-blood-tap-base-rune → master

> **Outcome** 3 findings — 2 fixed, 1 for whoever squashes this — nothing blocking
> **Reviewed** `8e2fc68e8` (merge-base diff vs `origin/master` — 4 files, +131 −25)
> **Project rules** `.agents/docs/self-review-rules.md`, `code-review.md` and `sql-guidelines.md`
> applied
> **In-game testing** the run reported on 2026-08-11 still stands: nothing since then changed
> behaviour, only the login fallback, the named visual constant, a file rename and two master
> merges. The login case itself is untested, see the bottom.

## Findings

1. `data/sql/updates/pending_db_world/` — the pending file was named by hand and only carried the
   seconds, seven digits short of what `create_sql.sh` produces with `date +%s%N`.
   **fixed** in `95b6c0955`, regenerated as `rev_1790121861505373800.sql`, contents untouched
2. branch was 64 commits behind master, and the one red check was master's own
   `mod-arena-3v3-solo-queue` build break, fixed upstream on the 19th. **fixed**, master merged
3. the `Co-authored-by: ccrs` trailer sits on `83baeae3e`, the follow-up commit, not on the one
   carrying the mirrored script pair. That is what was asked for here, so nothing to change, but
   whoever squashes this needs to keep the trailer in the squash message or the credit is lost.
   **open, for the merge**

## What came back clean

The hook order is the part worth checking, since the whole fix rests on the aura script seeing the
slot the spell script picked:

- `_RegisterForTargets()` calls `UpdateTargetMap(caster, false)`, and `SpellAuras.cpp:754` returns
  on `!apply` before applying effects, so the aura created at `Spell.cpp:3246` is registered but not
  applied.
- `Spell::HandleEffects` calls `CallScriptEffectHandlers` before the default handler, so on
  `EFFECT_1` the `OnEffectHitTarget` hook hands the slot over and only then does `EffectApplyAura`
  run and fire `OnEffectApply`.
- No double handling on `EFFECT_0`: `EffectActivateRune` returns unless
  `effectHandleMode == SPELL_EFFECT_HANDLE_LAUNCH`, and the prevented-default mask is read inside
  the same `CallScriptEffectHandlers` call, so `_InitHit()` clearing it later does not matter.

Game data, checked against `Spell.dbc` rather than taken from the description: 45529 `EFFECT_0` is
effect 146 with `MiscValue` 3 and `MiscValueB` 0, `EFFECT_1` is effect 6 with `AuraName` 249 and
`MiscValueB` 3. With `RUNE_BLOOD = 0` and `RUNE_DEATH = 3`, `Effects[EFFECT_0].MiscValueB` really is
the blood rune and `aurEff->GetMiscValueB()` really is the death rune.

The out of bounds read the description claims is real: `GetRuneCooldown(uint8 index)` indexes
`runes[MAX_RUNES]` with no bounds check, and the removed fallback called it with `l + 1` while `l`
ran to `MAX_RUNES - 1`.

Empower Rune Weapon cannot be affected: the removed guard sat inside `if (m_spellInfo->Id == 45529)`
and both the generic loop and the Empower block are unchanged.

`spell_dk_blood_tap` appears in no other SQL in the repo, so the `DELETE` is only the idempotency
guard the guidelines ask for. Both linters pass. `damage` on `EFFECT_0` is 1, so `count` is 1 and
the loop cannot leave a later slot overwriting an earlier one.

## Still to test in game

- Relog with the Blood Tap buff still up. That is the login fallback, and it is the one path in here
  nobody has watched run.
- Blood Tap with both blood runes up but already converted to death runes by Reaping, which is the
  case that leaves the slot unset and falls through to the default handler.
- Empower Rune Weapon once more on this head, since the last master merge touched the file.

PR #27141, opened 2026-08-14
fix(Core/Pet): stop new pets inheriting a recycled pet number's spells

Nyeriah:
## Changes Proposed:

A pet can inherit the spell book of a completely different pet, including one from another class. Reported at https://github.com/chromiecraft/chromiecraft/issues/10016 as a Succubus showing and casting hunter pet abilities.

Two things have to line up for it, and both are longstanding:

**Pet numbers are recycled.** `ObjectMgr::LoadPetNumber` seeds `_hiPetNumber` from `SELECT MAX(id) FROM character_pet` on every startup, and `GeneratePetNumber()` is just `++_hiPetNumber`. Nothing is persisted and no used-id list is kept, so whenever `MAX(id)` drops (a character is deleted, a hunter releases a pet, staff prune rows, a DB is restored) the next restart hands those numbers out again.

**`pet_spell` is the only pet table that doesn't self-heal on reuse.** `_SaveAuras` and `_SaveSpellCooldowns` both begin with `DELETE ... WHERE guid = ?` and then write the current set, which is why `pet_aura` and `pet_spell_cooldown` never suffer from this. `_SaveSpells` wrote per spell and skipped `PETSPELL_UNCHANGED`, so rows it didn't put there were never touched.

Put together: a new pet takes a number a dead pet used to own, and from its second summon onward `_LoadSpells` hands those leftover rows to `addSpell()`, which only checks that the spell exists in the store. No family check, no pet type check. Passives get cast on the pet and actives go on the action bar, so they genuinely work rather than just showing up in the book.

This PR makes `_SaveSpells` wipe and rewrite like its two sibling tables, so a recycled number is cleaned on the pet's first save no matter where the leftover rows came from. It also adds a characters DB update purging rows belonging to pets that no longer exist.

`CHAR_DEL_PET_SPELL_BY_SPELL` loses its last caller and is removed.

Worth knowing for review: this turns the spell save from additive into authoritative, so a save with an incomplete `m_spells` would now wipe the DB set rather than being roughly a no-op. The guard is the `m_loading` check at the top of `SavePetToDB`, the same one the other two tables already rely on. I traced every early return in `LoadPetFromDB` to confirm nothing saves mid-load.

TrinityCore's 3.3.5 branch has identical code in all of these places, so this isn't an AC regression and there's nothing to cherry-pick.

This PR proposes changes to:

### AI-assisted Pull Requests

## Issues Addressed:
- https://github.com/chromiecraft/chromiecraft/issues/10016

## SOURCE:

None of the boxes below apply. This is a data integrity bug rather than a behaviour parity claim, so there's no live or sniff source to cite. The evidence is the code paths above plus the reproduction below.

## Tests Performed:
This PR has been:

Reproduced the reported bug shape on a local realm, then confirmed the fix. Also confirmed the DB update applies through the normal updater and clears every orphaned set.

## How to Test the Changes:

The bug needs a recycled pet number carrying leftover rows, which you can set up directly with the server down:

1. Note `SELECT MAX(id) FROM character_pet`. The next two pets created will get that value plus 2 and plus 3.
2. Plant a hunter pet's spell set on both of those numbers, e.g. `INSERT INTO pet_spell (guid, spell, active) VALUES (<n>, 1742, 129), (<n>, 27047, 129), (<n>, 24453, 129);` (Cower, Claw, and a cat talent). Leave `character_pet` alone so they're orphans.
3. Start the server and log in a warlock that doesn't currently own a Succubus or a Voidwalker.
4. Summon a Succubus, then a Voidwalker. They take the two numbers you planted on.
5. Relog, so the current pet loads through `LoadPetFromDB` rather than being created fresh.
6. Open the spellbook and check the Demon tab.

Before the change the demon's book gains Claw and Cower and they can be cast. After it, the Demon tab holds only the demon's own abilities, and `pet_spell` for those numbers contains only the demon's spells.

For the DB update, run the server once with the pending file in place and check that `pet_spell`, `pet_aura` and `pet_spell_cooldown` no longer contain any guid missing from `character_pet`.

- [ ] The leak itself isn't closed. I could only find one core path that removes a `character_pet` row without cleaning its children ([the hunter slot delete in `SavePetToDB`](https://github.com/azerothcore/azerothcore-wotlk/blob/7d9027d73d47d08d9b8096518a4b16fb2da68c45/src/server/game/Entities/Pet/Pet.cpp#L551-L559)), and it can't account for the orphans I found, which were all warlock pets. Rows also arrive from outside the core entirely, via staff SQL and restores. Wiping on save makes the origin irrelevant, but the leak is still worth chasing separately.
- [ ] Pets that are already contaminated aren't repaired. Their rows hang off a live `character_pet` row, so they aren't orphans and neither change touches them. Fixing one is a `DELETE FROM pet_spell WHERE guid = <pet id>` with the character offline; the correct spells are rebuilt by `LearnPetPassives` and `InitLevelupSpellsForLevel` on the next summon. Healing them automatically would need validation in `_LoadSpells` against the pet's family and talent type, which I'd rather not bolt on here.
- [ ] Pet numbers still get burned without ever reaching the DB, which is what keeps forcing the counter to rewind. [Enslave Demon calls `GeneratePetNumber()`](https://github.com/azerothcore/azerothcore-wotlk/blob/7d9027d73d47d08d9b8096518a4b16fb2da68c45/src/server/game/Entities/Unit/Unit.cpp#L14783) purely to enable the stat window on a charmed creature, which never gets a `character_pet` row. Making pet numbers monotonic would be the deeper fix and is a separate change.


Nyeriah:
# Self-review — fix/pet-spell-recycled-pet-numbers → master

**Outcome** 0 findings — round 1 clean

**Reviewed** `10e0086a8d` (`fix/pet-spell-recycled-pet-numbers` vs `upstream/master`, merge-base diff — 4 files, +21 −34)

**By** Claude Code (Claude Agent SDK), Opus 5 — self-review v0.3, 2026-08-14

**Tested in-game** Local realm, warlock. Planted a hunter cat's saved spell set on the next two pet numbers to be issued, summoned a Succubus and a Voidwalker onto them, then relogged so the current pet loaded through `LoadPetFromDB`. Demon tab held only the demon's own abilities; `pet_spell` for both numbers held only their own spells. The SQL update applied through the normal DB updater and cleared every orphaned set.

**Not tested** Hunter pets. `_SaveSpells` is shared, so tame, stable/unstable and call-pet still want a pass, and pet talents in particular — they live in `pet_spell` and now depend on the rewrite preserving them across a relog.

**Verified** Release build clean; `codestyle-cpp.py` and `codestyle-sql.py` both pass.

<details>

**Intent** Prevent a newly created pet from inheriting spell rows left behind in `pet_spell` by a previous pet that held the same pet number, and purge already-orphaned pet rows from the characters database.

**Project rules** `.agents/docs/self-review-rules.md` found and applied — shared-core reach, SQL WHERE side effects, in-game testing recorded above.

**Coverage** All four changed files examined, none skipped.

## Round 1 — `10e0086a8d`, clean

Reviewer confirmed, with no findings:

- The mirrored pattern is itself sound — `_SaveAuras` and `_SaveSpellCooldowns` already delete-then-rewrite, all three reach the DB only via `SavePetToDB`, and its `!m_loading` gate enforces the same precondition for spells.
- Every path into `_SaveSpells`: single caller `SavePetToDB`, itself called from 9 sites; every creation path runs `InitPetCreateSpells()` before the first save, so none can reach the new unconditional DELETE with a partly populated `m_spells`.
- The one path that skips `_LoadSpells` and could rewrite an empty book over live rows (`isTemporarySummon` in `LoadPetFromDB`) is unreachable — no `SUMMON_PET` or `HUNTER_PET` gets a creating spell with a duration.
- `CHAR_DEL_PET_SPELL_BY_SPELL` has no remaining reference anywhere including `modules/`, and nothing keys off the numeric ordinals of `CharacterDatabaseStatements`.
- SQL joins are exact (PK to same-typed guid column written from `GetPetNumber()` in every insert path) and cannot catch a live pet's rows. The half-saved window in `SavePetToDB` isn't exposed to the updater, which runs in `StartDB()` before any socket opens.
- Conventions: correct pending directory, `create_sql.sh` header, no INSERTs so the DELETE-before-INSERT rule doesn't apply, LEFT JOIN purge idiom matches existing precedent in `db_characters/2026_07_18_00.sql`, commit message in `type(Scope): description` form.
- No leftovers — no debug output, commented-out code, TODOs or stray files.

Two non-blocking observations, raised as context rather than findings:

- The purge is one-shot; `CHAR_DEL_CHAR_PET_BY_SLOT` and `CHAR_DEL_CHAR_PET_BY_OWNER` still leak orphans, so they accumulate again. No longer a correctness issue once a recycled number is wiped on first save, but a hygiene gap.
- Every pet save now writes one DELETE plus N INSERTs instead of nothing when nothing changed. Same cost profile as the two sibling tables already have.

</details>

Gultask (approved):

PR #27147, opened 2026-08-14
fix(Scripts/IcecrownCitadel): make the Gas Cloud detonate Expunged Gas and keep hunting

### LINKED ISSUE #25849: [ICC] Professor Putricide - Red Ooze (Volatile Ooze) does not explode on player contact
### Current Behaviour

When the Red Ooze (Volatile Ooze) reaches its target player during the Professor Putricide encounter, it deals damage to the player but does not explode. The ooze should explode on contact, dealing AoE damage to nearby players.

### Expected Behaviour

When the Volatile Ooze reaches its targeted player, it should explode dealing  AoE damage to all players in range, then despawn. This is a core mechanic of the Professor Putricide encounter.

### Source

- Wowpedia encounter: [https://www.wowhead.com/wotlk/guide/raids/icecrown-citadel/professor-putricide-strategy](url)
- Wowhead - Volatile Ooze: [https://www.wowhead.com/wotlk/npc=37697/volatile-ooze](url)
- Ooze Eruption spell: [https://www.wowhead.com/wotlk/spell=70492/ooze-eruption](url)
- According to Wowpedia, when the Volatile Ooze reaches its target it should  trigger Ooze Eruption (spell 70347), exploding and dealing AoE damage to all nearby players.

1. Enter Icecrown Citadel - Plagueworks
2. Engage Professor Putricide
3. Wait for him to throw a Red Ooze (Volatile Ooze)
4. Allow the ooze to reach its target player
5. Ooze deals damage but does not explode and despawn

### Extra Notes

_No response_

### AC rev. hash/commit

0b30b3f99f8f

### Operating system

Ubuntu (Docker)

### Custom changes or Modules

None

EricksOliveira:
In the Professor Putricide encounter the Gas Cloud fixates a player with Gaseous Bloat
and chases them, but nothing ever cast Expunged Gas, so reaching the player did nothing
at all. The bloat now procs on the cloud's own auto attack against the player carrying
it and fires the detonation, scaled by the stacks left at contact.

Heads up on the issue text: it names the Volatile Ooze (green, 37697) and spell 70347,
and both are wrong. The broken add is the Gas Cloud (37562) with Expunged Gas (70701),
and 70347 is the Slime Puddle's Grow. The green ooze chain was already complete and
isn't touched here.

## Decisions worth knowing

- The cloud doesn't despawn after detonating any more, it fixates someone new and keeps
  going until killed. Its summon spell has no duration, the fixate aura is permanent,
  and Gaseous Bloat Protection only lasts 15s, which is pointless unless a new bloat can
  land soon after. The Wowhead strategy guide says the same. This needed no new code,
  the base ooze AI already handles it.
- Detonation damage is the whole bloat the target still had on them, summed off the
  periodic amount, which is already per stack times the stacks left and already carries
  the per difficulty scaling. The hardcoded 1250/1500 pair it replaces only separated 10
  from 25, so heroic raids were getting normal mode numbers.
- The damage is no longer split across everyone hit, so the `spell_custom_attr` row is
  gone. Ooze Eruption's tooltip says its damage is shared and its radius is 10 yards,
  while Expunged Gas says nothing about sharing and reaches the whole map.
- `SPELL_ATTR3_ALWAYS_HIT` for 70701 in `SpellInfoCorrections.cpp`: it's missing from the
  dbc, the detonation did miss in testing, and Ooze Eruption already carries it.
  Immunities still apply, since the immunity check runs first.
- No proc data ships with this. The four Gaseous Bloat ids already carry `ProcFlags` 8 in
  the dbc and their `spell_proc` rows fall back to it.

## Review guide

Worth real attention: the two new handlers on `spell_putricide_gaseous_bloat_aura` in
`boss_professor_putricide.cpp`. `CheckProc` is what stops other melee attackers on the
fixated player from setting the cloud off, and `HandleProc` removes the bloat before
casting, the same way the green ooze consumes its adhesive.

The rest is mechanical: the `Validate` line, the `SpellInfoCorrections.cpp` entry, and
dropping a dead `_newTargetSelectTimer` that shadowed the base class and was never read.

## Known gaps

- Nothing sniffed backs the detonation magnitude. 70701 carries no damage in the dbc, so
  the number has to come from the script. At a full 10 stacks it lands about 68.7k in 10N
  and 110k in 25H on every player, undivided. Worth a maintainer's eye.
- Needs a fresh in-game pass. The earlier 10N run tested the version that despawned the
  cloud, so it doesn't describe this one. Worst case to check is contact at 9 or 10 stacks.

## Changes Proposed:
This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Opus 5

## Issues Addressed:
- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/25849

## SOURCE:
- [x] The changes promoted by this pull request come partially or entirely from another project (cherry-pick).

## Tests Performed:
This PR has been:

## How to Test the Changes:

1. Engage Professor Putricide and wait for the Unstable Experiment that opens the red gas tank.
2. Let the Gas Cloud reach its fixated player. The raid should take Expunged Gas, the bloat should
   disappear from that player, and the cloud should fixate someone else instead of despawning.
3. On another cloud, kite it for the full 20s so the bloat ticks down to 0 stacks without contact.
   It should fixate a new player on its own.
4. Worst case: let a freshly spawned cloud reach its target at 9 or 10 stacks and check the raid
   survives the hit. Worth doing on both 10N and 25H.
5. Regression: the green ooze should still root its target, erupt for damage split within 10 yards,
   and pick a new target afterwards.


src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
```diff
@@ -860,18 +860,37 @@ class npc_gas_cloud : public CreatureScript
...
         }
 
         void CastMainSpell() override
         {
             me->CastCustomSpell(SPELL_GASEOUS_BLOAT, SPELLVALUE_AURA_STACK, 10, me, false);
         }
 
+        // the detonation consumes the cloud - unlike the Volatile Ooze it does not pick a new target
+        void SpellHitTarget(Unit* /*target*/, SpellInfo const* spell) override
+        {
```
PkllonG: +            npc_putricide_oozeAI::SpellHitTarget(target, spell);
If you can't catch the player, you should pick a new target.

PkllonG: Oh, sorry, my bad

EricksOliveira: I believe it is just a mistake in the comment. I will correct it to make it more informative and less confusing.

PkllonG:
Testing
image

src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
```diff
@@ -860,18 +860,37 @@ class npc_gas_cloud : public CreatureScript
...
+            if (_detonated || spell->Id != sSpellMgr->GetSpellIdForDifficulty(SPELL_EXPUNGED_GAS, me))
+                return;
+
+            _detonated = true;
+            me->AttackStop();
+            me->GetMotionMaster()->Clear();
+            me->StopMoving();
+            me->DespawnOrUnsummon(2s);    // lingers for a moment instead of blinking out with the damage
```
PkllonG: 2s is too long, I tested 1s it feels most like Blizzard

PkllonG (approved):

src/server/scripts/Northrend/IcecrownCitadel/boss_professor_putricide.cpp
EricksOliveira: I'm going to try it out. I did some research but couldn't find a precise time, since in most videos, they kill the Ooze before it reaches the target. Haha.

PkllonG:
Tested

PkllonG (approved):

imagionationeditor:
for this test u need 3 char in fight

mpfans:
If Gas Cloud does not die after catching up with the player's explosion, it will continue to choose new targets instead of disappearing

EricksOliveira:
From what I've researched, the green Ooze stays alive and acquires a new target until it is killed by the players. However, the red/orange one doesn't; as soon as it touches a player, it should disappear.

mpfans:
I will go find video evidence

mpfans:
> From what I've researched, the green Ooze stays alive and acquires a new target until it is killed by the players. However, the red/orange one doesn't; as soon as it touches a player, it should disappear.

https://www.wowhead.com/wotlk/guide/raids/icecrown-citadel/professor-putricide-strategy

```
Similarly to the[Volatile Ooze](https://www.wowhead.com/wotlk/npc=37697/volatile-ooze), the[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)is the raid's priority when it's alive.
The[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)will also fixate on a random player, casting a red beam on them, and start chasing them after inflicting the[ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70215/gaseous-bloat)debuff.[ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70672/gaseous-bloat)will start at 10 stacks and deal 1.5k nature and shadow damage (2k in heroic difficulty) per stack every 2 seconds, but it will also lose 1 stack after each tick. This means the first tick at 10 stacks will hit the target for 15k damage (20k in heroic difficulty) and the second tick, 2 seconds later, will hit for 13.5k (18k in heroic difficulty), all the way down to 0 stacks. If the[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)manages to catch up to its target, it will cast[ Expunged Gas](https://www.wowhead.com/wotlk/spell=70701/expunged-gas), dealing damage to the entire raid based on the number of[ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70672/gaseous-bloat)stacks remaining on its target. Since the target of the[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)is not immobilized this time, they can very easily kite the red ooze in the room and should never let it catch up.Nitro Boosts[ Nitro Boosts](https://www.wowhead.com/wotlk/item=41118/nitro-boosts)are, once again, a must-have for this encounter.
The[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)will pick a new target once the[ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70672/gaseous-bloat)stacks on its current target hit 0, so melees should run away from the red ooze when the kiter has only 3 stacks remaining to put some distance between them and the[Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud)should they be targeted next.
```



So if the target layer selected by the Poison Gas Cloud becomes 0 or explodes, it will cause a random target to be selected again

EricksOliveira:
The [Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud) will pick a new target once the [ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70672/gaseous-bloat) stacks on its current target hit 0, so melees should run away from the red ooze when the kiter has only 3 stacks remaining to put some distance between them and the [Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud) should they be targeted next.

mpfans:
> The [Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud) will pick a new target once the [ Gaseous Bloat](https://www.wowhead.com/wotlk/spell=70672/gaseous-bloat) stacks on its current target hit 0, so melees should run away from the red ooze when the kiter has only 3 stacks remaining to put some distance between them and the [Gas Cloud](https://www.wowhead.com/wotlk/npc=37562/gas-cloud) should they be targeted next.

I am working hard to find a video from 2010 to prove that our experience confirms that if the Poison Gas Cloud fails to catch up with the player for 20 seconds or if catching up with the player causes a lot of damage, both situations will randomly search for the target again

mpfans:
https://www.warcrafttavern.com/wotlk/guides/professor-putricide-10/#1

`40 seconds after the Volatile Ooze has been spawned, Putricide will open the red tank of gas, spawning a Gas Cloud. The ooze will target a random player with[Gaseous Bloat](https://wowclassicdb.com/wotlk/spell/70672)Gaseous Bloat, dealing higher damage the more stacks the target has. This debuff begins at 10 stacks, and every tick of damage will remove a stack. The gas will chase the target, and if they manage to reach them, will explode with Expunged Gas, damaging the raid based on the amount of stacks left. If[Gaseous Bloat](https://wowclassicdb.com/wotlk/spell/70672)Gaseous Bloatreaches 0 stacks, a new[Gaseous Bloat](https://wowclassicdb.com/wotlk/spell/70672)Gaseous Bloatwill be cast on a random player. Gas Clouds will need to be kited for their entire duration, and are the top priority to DPS while they are alive.`

When exploding, the AOE damage will be caused by multiplying the remaining target layers by the damage base, and the target's layers will be cleared to 0. Then, another target will be selected

mpfans:
https://bellumpax.rpg-board.net/t177-professor-putricide-killed#1274

`This orange-red slime add does not attack directly, and cannot be tanked. Immediately after spawning, the add chooses a random raid member, casts a 10-stack of Gaseous Bloat on them, and begins to move toward the player. If it reaches the debuffed player, it will cause the remaining stacks of Gaseous Bloat to fade, and Expunged Gas will hit everyone in the raid. At this point the Gas Cloud will pick a new target, and repeat the process until killed.`

mpfans:
https://www.youtube.com/watch?v=pllYT5du93I
Please watch from 3 minutes and 50 seconds

mpfans:
If you need more evidence, please let me know

EricksOliveira:
Judging by the video, I appear to be wrong. Sorry.

mpfans:
> Judging by the video, I appear to be wrong. Sorry.

I have already fixed it locally, but the explosion damage is not an official standard. I gave the damage based on experience,I haven't submitted a PR yet because this is a hack fix
﻿
SpellInfoCorrections.cpp
```
    // Professor Putricide, Expunged Gas - AoE explosion scaled by remaining Gaseous Bloat stacks
    ApplySpellFix({ 70701 }, [](SpellInfo* spellInfo)
    {
        spellInfo->Effects[EFFECT_0].TargetA = SpellImplicitTargetInfo(TARGET_UNIT_DEST_AREA_ENEMY); // raid-wide AoE
        spellInfo->Effects[EFFECT_0].DieSides = 0;   // random per-stack handled in script via 1218+irand(1,63)
        spellInfo->Attributes |= SPELL_ATTR0_NO_IMMUNITIES;
        spellInfo->AttributesEx3 |= SPELL_ATTR3_ALWAYS_HIT;
    });
```

boss_professor_putricide.cpp

```
        // Gas Cloud explosion on melee contact with Gaseous Bloat debuffed player
        if (_hitTargetSpellId == SPELL_EXPUNGED_GAS)
        {
            if (Unit* victim = me->GetVictim())
            {
                uint32 gaseousBloatId = sSpellMgr->GetSpellIdForDifficulty(SPELL_GASEOUS_BLOAT, me);
                if (victim->HasAura(gaseousBloatId) && me->IsWithinMeleeRange(victim))
                {
                    // Get remaining stacks for damage scaling
                    uint8 stacks = 0;
                    if (Aura* bloatAura = victim->GetAura(gaseousBloatId))
                        stacks = bloatAura->GetStackAmount();
                    victim->RemoveAurasDueToSpell(gaseousBloatId);

                    // Per-stack damage based on Gaseous Bloat BaseDice+DieSides from DBC per difficulty:
                    // 70672(10N): BaseDice=1218, DieSides=63 → 1219~1281 per stack
                    // 72455(25N): BaseDice=1462, DieSides=75 → 1463~1537 per stack
                    // 72832(10H): BaseDice=1462, DieSides=75 → 1463~1537 per stack
                    // 72833(25H): BaseDice=1949, DieSides=101 → 1950~2050 per stack
                    int32 perStackBase, perStackDieSides;
                    switch (gaseousBloatId)
                    {
                        case 70672: perStackBase = 1218; perStackDieSides = 63; break;   // 10N
                        case 72455: perStackBase = 1462; perStackDieSides = 75; break;   // 25N
                        case 72832: perStackBase = 1462; perStackDieSides = 75; break;   // 10H
                        case 72833: perStackBase = 1949; perStackDieSides = 101; break;  // 25H
                        default:    perStackBase = 1218; perStackDieSides = 63; break;   // fallback 10N
                    }
                    // Damage doubled per request, AoE on entire raid (TARGET_UNIT_DEST_AREA_ENEMY)
                    int32 damage = (stacks * 5 * (perStackBase + irand(1, perStackDieSides))) / 2;  // multiply by 2.5
                    me->CastCustomSpell(SPELL_EXPUNGED_GAS, SPELLVALUE_BASE_POINT0, damage, victim, TRIGGERED_FULL_MASK);
                    SelectNewTarget();
                }
            }
        }
```


If you need it, you can refer to it

EricksOliveira:
Thank you. I'm going to work on that case this afternoon.

EricksOliveira:
# Self-review — FixOozeRed → upstream/master

**Outcome** 5 findings across 4 rounds — 3 fixed, 2 dismissed — final round clean

**Caveat** not tested in-game at this SHA: the round 3 fixes changed the encounter's behaviour, so the 10N testing recorded for `b97829318` (which ended with the cloud despawning) no longer describes what this build does

**Reviewed** `f83497db3` (`FixOozeRed` vs `upstream/master`, merge-base diff — 2 files, +31 −10)

**By** Claude Code (Claude Agent SDK), claude-opus-5 — self-review v0.3, 2026-08-21

**Testing needed before merge** contact at 9–10 stacks on both 10N and 25H, to confirm the raid survives the largest detonation; bloat decaying to 0 without contact (20s of ticks against a 15s Gaseous Bloat Protection, so the same player may legitimately be re-picked); detonation while the target is immune (Divine Shield — `PROC_HIT_IMMUNE` is inside the proc's hit mask, so it should still fire); a Tear Gas phase transition landing between contact and re-target; and a Volatile Ooze regression pass. Build, lint and tests were not run by this review.

<details>
<summary>Review details (4 rounds)</summary>

**Intent** In the ICC Professor Putricide encounter, the Gas Cloud (37562) fixates a player with Gaseous Bloat (70672 and its difficulty variants) and chases them. On reaching that player it must detonate Expunged Gas (70701), damaging the raid scaled by the stacks remaining at contact, then fixate a new player and keep going until killed — it must not despawn. It must also fixate a new player if the bloat decays to zero stacks first. The Volatile Ooze (37697) must keep behaving as it does today.

**Project rules** `.agents/docs/self-review-rules.md` found and applied (pinned review, regression depth by reach, SQL predicate side effects, in-game testing recorded above).

**Excluded from the reviewers** `.agents/plans/` and this report.

**Caveat** the codestyle linters mandated by `AGENTS.md` could not run — no Python interpreter on this machine — so every round checked the style rules by hand.

## Round 1 — `ce5d7c1ff`, 1 finding

1. `rev_1786650055537512600.sql:5` — `UPDATE spell_proc SET ProcFlags = 8` is a no-op: the dbc already carries 8 on all four ids and `LoadSpellProcs` falls back to it while the column is 0 → **fixed** in `b97829318`

## Round 2 — `b97829318`, 2 findings

1. `boss_professor_putricide.cpp:1062` — `Is25ManRaid()` collapses 10N/10H and 25N/25H, so heroic got normal-mode detonation damage → dismissed at the time as unsourced, later **fixed** in `9e8bf90d1` by reading the amount off the aura effect, which already carries the per-difficulty scaling
2. `rev_1786650055537512600.sql:2-3` — comment reads as if Ooze Eruption were raid-wide too, while its radius is 10 yards → **dismissed**: the "exactly like" attaches to the damage split, which was accurate; moot since `9e8bf90d1` deletes the file

## Round 3 — `9e8bf90d1`, 2 findings

1. `boss_professor_putricide.cpp:872-881` — the new `UpdateAI` override duplicates the base class: Gaseous Bloat is channeled with no duration, so `UpdateChanneledTargetList` ends the channel when the target loses the aura and `npc_putricide_oozeAI::UpdateAI` already calls `SelectNewTarget()`; costs a `GetSpellIdForDifficulty` + `HasAura` per tick per cloud for nothing → **fixed** in `f83497db3`
2. `boss_professor_putricide.cpp:1050-1054` — Expunged Gas magnitude is unverifiable: at 9–10 stacks every raider takes ~12.5k (10N) to ~20k (25H) undivided → **dismissed**: 70701 carries no damage in the dbc (base points −1) and has no difficulty variants, so the value must be script-supplied and there is no ground truth; deriving it from the Gaseous Bloat periodic amount at least tracks Blizzard's own per-difficulty numbers, where any multiplier would be invented. Recorded as a required in-game test instead.

## Round 4 — `f83497db3`, clean

Round 4 also verified, without producing findings: the proc chain fires end to end without new SQL (dbc `ProcFlags` fallback in `LoadSpellProcs`, hit mask covering miss/dodge/parry/immune); the melee that triggers it is reachable during the channel; removing the aura inside `OnProc` is free of lifetime issues; both re-target paths work now that the override is gone (`SpellHitTarget` on 70701 after detonation, channel-end for stack decay) and no despawn code remains; the `GetAmount()` claim in the comment holds; the `ALWAYS_HIT` correction reaches only 70701, which nothing outside this encounter references, and has no DB-only alternative; the Volatile Ooze chain is untouched; no leftovers, and no style violations in the added lines.

</details>

EricksOliveira:
@mpfans Test now pls.

mpfans:
I used 3 characters for testing, one character is Professor Putricide's current tank, and the other two characters participate in the roll call test. The Poison Cloud will roll call two characters twice and then automatically disappear

EricksOliveira:
Thanks for testing this.

That's expected behavior when only 2 players are eligible. The cloud can't
pick anyone carrying Gaseous Bloat Protection (70812), and that aura lasts
15s and gets applied the moment the cloud fixates, not when it detonates
(effect 2 of 70672 force casts it on the fixated target). With the tank out
of the pool you're left with 2 candidates, so by the time the cloud looks
for a third target both are still protected and the list comes out empty.

The despawn itself isn't something this PR adds. It's in
`spell_putricide_ooze_channel::SelectTarget` on master already: when the
filtered list is empty it does `FinishCast(SPELL_FAILED_NO_VALID_TARGETS)`
and `DespawnOrUnsummon(1ms)`.

The Volatile Ooze has always worked the same way, it reselects after Ooze
Eruption and its protection (70530) is even longer at 25s. If you run the
same 3 character test with the green ooze on master, it should vanish on you
in exactly the same spot.

With a real raid of 10 or 25 the pool never runs dry, since the 15s window
only ever covers two or three fixates.

mpfans:
I think it can be merged now

sudlud:
the proc setup here comes from TrinityCore (`spell_putricide_gaseous_bloat::HandleProc` on their
3.3.5 branch), and the first commit still had their `1500 : 1250` mod and the
`for (i = 1; i <= stack; ++i)` loop verbatim. can we get a follow-up commit with
`Co-authored-by: Shauren <shauren.trinity@gmail.com>` naming the two upstream commits,
`240534ba2` (moved Expunged Gas to the proc system) and `8524faae9` (the damage formula)?
worth ticking the cherry-pick box in SOURCE too.

on the damage: `GetAmount()` is per stack times the current stacks, so it's one tick's worth,
about 12.5k at 10 stacks on 10N. TC deals the whole remaining DoT instead,
`mod * stacks * (stacks + 1) / 2`, so 68750 at the same point. you can keep your per difficulty
scaling and still land on that:

```cpp
uint8 const stacks = GetStackAmount();
int32 const damage = bloat ? bloat->GetAmount() * (stacks + 1) / 2 : 0;
```

both reads need to stay above the `RemoveAurasDueToSpell` call. that also avoids the 10H/25H
collapse you spotted in TC's version.

one more: the title reads better imperative per the commit guidelines, maybe
`fix(Scripts/IcecrownCitadel): make the Gas Cloud detonate Expunged Gas and keep hunting`

EricksOliveira:
all three done. added a follow-up commit with `Co-authored-by: Shauren` naming 240534ba2
and 8524faae9, and ticked the cherry-pick box in SOURCE.

good catch on the damage, i had missed that `GetAmount()` already multiplies by the stack
count. went with your `GetAmount() * (stacks + 1) / 2`, both reads above the
`RemoveAurasDueToSpell`. that puts it at 68750 at 10 stacks on 10N like TC, and heroic
still picks up its own scaling from the dbc.

title updated too, and i bumped the numbers in the known gaps section since they were
still the one tick figures.

Issue #25337, opened 2026-04-01
[Item] - "Glyph of Raise Dead" for Death Knight results in no cooldown being displaying and not following the normal or talent cooldown


TheSCREWEDSoftware:
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/9232

Triage Notes: 

- Make DK
- `.cast 50977`
- Take the portal
- `.modify money 1000g`
- `.level 25`
- `.additem 37201 20` // Corpse Dust
- `.additem 43673` // Glyph of the Dead thingy
- `.cheat cooldown` won't help for this testing, I recommend taking `Night of the Dead` 2 points (Unholy tree) so you can cast Raise Undead every 30 seconds.
- Cast `Raise Undead`
- Select it the ghoul`.damage 99999` or `.die`
- See that without the glyph you always get a cooldown displayed and it works after the given time you can cast it fine
- Equip the glyph and all of a sudden (with or without a reagent in your bag) the cooldown is not display (sometimes greyed out, sometimes fully / normal but `Spell not ready yet`, wait 30 seconds or 2 mins (depending if you have or not hte talent) see that even than it doesnt work, it does work after a while, just takes much longer

Original Post Below: 

---

### Current Behaviour

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2f90f669-651d-459e-a3c5-6005a0423931" />

Summon DK Pet keeps going greyed out, both when trying to summon it, and when it dies. 

Cooldown is not resetting and cant make a new one without relogging. 7th time this has happened

### Expected Behaviour

pet should reset cooldown when it dies

### Source

https://wotlk.evowow.com/?spell=46584

### Steps to reproduce the problem

summon pet
have it die
try summon another
+repeat 

### Extra Notes

im using Death Pact sometimes, maybe this is impacting it somehow? but its times it happens when i havent used DP too

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/8ea9252d764767dba2d9ddd7fd5c794a13fdc054

### Operating system

Ubuntu 22.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
[...]

PR #27159, opened 2026-08-14
fix(Core/Spells): sync Raise Dead cooldown with the client

### LINKED ISSUE #27081: [Death Knight] Unholy - Cannot Summon Pet in Dungeons/Raids sometimes- Raise Dead bug "Spell is not ready yet"
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9997

Triage Notes: 

repro
- make dk level 80 and .learn all my class and .learn all my talent
- .tele utgarde
- summon your ghoul, then .die on your self
- walk to the dungeon and try to summon the pet again 
- observe the icon appear as "castable" but says spell not ready 
https://dl.dropboxusercontent.com/scl/fi/abo9ujgrq2sbqy0ok9occ/Wow_jOygkb5Glk.mp4?rlkey=pniud53d8w6ekq9fo2pnyom39&dl=0

Original Post Below: 

---
### Current Behaviour

When you die in a dungeon or raid and enter again, the Unholy DK`s pet sometimes cannot be summoned. The icon of the speel goes dark and when you click it keeps saying ``Spell is not ready yet``. Sometimes teleporting out and in solves the issue. 

### Expected Behaviour

The pet summon ability should work after the cooldown ends.

https://www.wowhead.com/forums/topic/huge-problem-with-my-pet-157107

Summon unholy dk pet using Raise Dead.
Die inside a dungeon or raid.
Resurrect.
Unable to use summon the pet again, raise dead constantly gives the error "spell is not ready yet".

### Extra Notes

_No response_

### Customizations

_No response_

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/f8f667bff5ad84017c1921c09c4a71136b4396ee

### Operating system

Ubuntu 24.04

### Modules

### Server

ChromieCraft

Chinoske:
## Changes Proposed:

A Death Knight's Raise Dead (46584) ends up out of sync between client and server: the icon looks ready, but casting answers *"Spell is not ready yet"*. It clears up on relog, which is why it reads as random.

The ghoul is summoned by a triggered spell, 46585 without Master of Ghouls or 52150 with it. Neither is in the client's spellbook. When the ghoul dies, `Unit::RemoveMinion` calls `Player::SendCooldownEvent` with that triggered spell, so the `SMSG_COOLDOWN_EVENT` goes out with an id the client has never seen and gets dropped. Server side the 3 minute cooldown does land on 46584, through the category sibling loop in `AddSpellAndCategoryCooldowns`, with `needSendToClient = false`. Server blocked, client unaware.

The link the core was missing is the spell category. AzerothCore has no category cooldown map (TrinityCore resolves this through `SpellHistory::_categoryCooldowns`), so this adds a small helper that walks `sSpellsByCategoryStore` and answers "which spell in this category does the player actually have in the spellbook". `SendCooldownEvent` then sends a second event for that spell and flags its cooldown entry so it survives a relog. Everything else about the function is untouched.

Three more things fell out of that:

- `spell_dk_raise_dead` called `RemoveSpellCooldown(46584, true)` right after summoning. Server side that was a no-op, the only entry there was the infinity "pet is out" block, and the `SMSG_CLEAR_COOLDOWN` it sent desynced the client further. It also happened to be the only thing stopping a second guardian from being summoned. Gone.
- A pet that is only parked for a while (mounting, zoning, vehicles) no longer starts the real cooldown. The infinity block stays and the pet comes back to it, which is what retail does: the spell is blocked because the ghoul is out, not because it is recharging.
- The fake `SMSG_SPELL_GO` that `Pet::LoadPetFromDB` sends to restore the client-side block now carries the spellbook id when the create spell isn't in the book. Since that id is a real spell every client knows, on this path it goes to the owner alone instead of the whole visibility set, otherwise bystanders would see a Raise Dead cast that never happened. Every other class keeps master's broadcast.

Nothing else is affected. Every entry point is closed by filters that only Raise Dead crosses in 3.3.5a data:

| Guard | Where | What it rules out |
| --- | --- | --- |
| `HasActiveSpell(spellInfo->Id)` returns early | `Player::GetSpellbookSpellForCooldown` | every summon whose spell is already in the spellbook (warlock, hunter, mage, totems, summoned items) |
| `!itemId` | `Player::SendCooldownEvent` | potions and item-triggered spells, which always arrive with an item id |
| `IsCooldownStartedOnEvent()` | `SendCooldownEvent`, `RemoveMinion`, `RemovePet` | Summon Imp (688), Call Pet (883), Summon Water Elemental (31687), none of which carry the attribute |
| `ToPet()` non-null plus a recovery time plus a matching parked pet number | `Unit::RemoveMinion` | guardians, totems, minipets, and any pet summon without a DBC cooldown |
| same `SpellFamilyName` | `Player::GetSpellbookSpellForCooldown` | any other class sharing a category id |

### Evidence

Read from this server's DBC files through the AzerothCore MCP, not transcribed from a wiki:

| Id | Spell | Category | RecoveryTime | CategoryRecoveryTime | Family | FamilyFlags[0] | Attributes | `COOLDOWN_ON_EVENT` (0x02000000) |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 46584 | Raise Dead (spellbook) | 1245 | 0 | 180000 | 15 | 0x1000 | 0x82000000 | yes |
| 46585 | Raise Dead (guardian, 60s) | 1245 | 0 | 180000 | 15 | 0x1000 | 0x02010000 | yes |
| 52150 | Raise Dead (pet, Master of Ghouls) | 1245 | 0 | 180000 | 15 | 0x1000 | 0x0A810000 | yes (plus Attr7 0x4000) |
| 688 | Summon Imp | 0 | 0 | 0 | 5 | — | — | no |
| 883 | Call Pet | 0 | 0 | 0 | 0 | — | — | no |
| 31687 | Summon Water Elemental | 0 | 0 | 0 | 3 | — | — | no |

`SpellCategory.dbc` row 1245 has `flags = 0`, so the "cooldown starts on event" behaviour comes from the spell attribute, not the category. The DBC data is correct and untouched.

Related data that had to line up: Night of the Dead (55620 / 55623) applies `SPELLMOD_COOLDOWN` of -45s / -90s over ClassMask 0x1000, which covers 46584, 46585 and 52150 alike, so client and server compute the same number. Glyph of Raise Dead (60200) only applies `SPELL_AURA_NO_REAGENT_USE` and never touches cooldowns.

Before and after, per case:

| Case | master | this PR |
| --- | --- | --- |
| Ghoul dies (killed, Death Pact, guardian expiring, player death) | server starts 180s on 46584, client gets an event for 46585/52150 and drops it | both sides show the same 180s |
| Ghoul alive | server has nothing, the script wiped the block; a second guardian could be summoned | the "pet is out" block holds on both sides |
| Mount, zone, vehicle | real 180s starts even though the pet comes right back | the block is kept, no cooldown starts |
| Pet reloaded from DB | fake cast sent with 52150, which the client ignores | sent with 46584, to the owner only |
| Logout while the pet is parked | infinity block is never persisted, so Raise Dead came back free | the real cooldown starts before the save, so it is there on relog |

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code (Claude Agent SDK) running Opus 5 (`claude-opus-5`), for the code, this description, and the review below. It ran the phase pipeline from the AC agentic workflow: `azerothcore-research` for the research notes, `/refine-ticket` for the requirements, `/create-implementation-plan` for the plan, then implementation, then `/self-review` (which drives `fresh-eyes-review` in a separate context) and `/generate-pr-description`. All game data came from the AzerothCore MCP reading this server's own DBC files and source tree, so the values above are what the server actually runs.

## Issues Addressed:
- Closes #27081

It should also cover [#25337](https://github.com/azerothcore/azerothcore-wotlk/issues/25337) (Glyph of Raise Dead showing no cooldown), [chromiecraft/chromiecraft#9997](https://github.com/chromiecraft/chromiecraft/issues/9997) and [chromiecraft/chromiecraft#9232](https://github.com/chromiecraft/chromiecraft/issues/9232), which are the same desync reported from different angles. [chromiecraft/chromiecraft#9858](https://github.com/chromiecraft/chromiecraft/issues/9858) (ghoul on permanent cooldown after a knockback) looks like a different bug and is not claimed here.

## SOURCE:

- [Wowhead WotLK 46584](https://www.wowhead.com/wotlk/spell=46584/raise-dead) for the tooltip and the confirmation that Raise Dead and Raise Ally do not share a cooldown.
- Video repro of the bug in [chromiecraft/chromiecraft#9997](https://github.com/chromiecraft/chromiecraft/issues/9997).
- A 2009 Blizzard support case on the same symptom, [bluetracker](https://www.bluetracker.gg/wow/topic/eu-en/9520844585-solved-raise-dead-spell-not-ready-yet/).
- TrinityCore 3.3.5 solved the same problem with the same idea, the spell category as the link: [`SpellHistory::SendCooldownEvent`](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Spells/SpellHistory.cpp) and `spell_dk_raise_dead::OverrideCooldown` in [`spell_dk.cpp`](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/scripts/Spells/spell_dk.cpp). The code here is written for AzerothCore's own cooldown containers, not copied, so this is not a cherry-pick.

## Tests Performed:
This PR has been:

Builds clean on Windows with MSVC (RelWithDebInfo), no new warnings. `apps/codestyle/codestyle-cpp.py` passes.

Tested in-game on a level 80 Death Knight, with `.cheat cooldown off` so the server state is real:

- The reported repro: summon, `.die`, release, resurrect, zone into a dungeon, cast Raise Dead. The icon shows the real cooldown and the *"Spell is not ready yet"* mismatch is gone.
- Ghoul killed with `.damage 99999`: cooldown starts on both sides, and the resummon works when it expires, with no relog.
- Death Pact on the ghoul: same, 3 minutes.
- Guardian without Master of Ghouls, left to expire on its own after 60s: cooldown starts.
- Mount and dismount with Master of Ghouls: the ghoul is parked and comes back, the icon stays blocked and no 3 minute countdown appears. Pressing Raise Dead with the ghoul back is refused and no second ghoul appears.
- Zoning into a dungeon and out, and entering and leaving a vehicle: same as mounting.
- Respec out of Master of Ghouls while mounted: the pet does not come back and Raise Dead gets a normal cooldown that expires, not a permanent block.
- Logout mounted with the ghoul parked, then relog: the cooldown is there and running, no free ghoul.
- Relog with the cooldown partway through: the remaining time matches what it was.

Not tested yet: the non-DK regression pass on the shared functions (warlock and hunter across mount, zone, death and a battleground death while mounted, a shaman totem, a potion), Night of the Dead's reduced cooldown, and the Glyph of Raise Dead case.

## How to Test the Changes:

1. Level 80 Death Knight with Master of Ghouls (52143) and some Corpse Dust (`.additem 37201 20`). Turn cheats off with `.cheat cooldown off`, otherwise none of this measures anything.
2. Summon the ghoul, `.die`, release, resurrect, walk into a dungeon and press Raise Dead. The icon must show the real remaining cooldown, and once it expires the summon must work.
3. Summon again, kill the ghoul with `.damage 99999`. The cooldown must start immediately on the icon and the resummon must work when it runs out, without relogging.
4. With the ghoul alive, press Raise Dead again. It must be refused, with no second ghoul, both with and without Master of Ghouls.
5. Mount with the ghoul out. It gets parked, and the icon must stay blocked with no 3 minute countdown. Dismount and the same ghoul comes back.
6. Repeat step 5, and while still mounted log out to the character screen and back in. The cooldown must be running when you return, and you must not be able to summon a second ghoul.
7. Regression, since these functions are shared: warlock and hunter pets across mounting, zoning, dying and a battleground death while mounted; a shaman totem expiring; a potion with a category cooldown.

- [ ] The non-DK regression pass listed above still needs eyes, ideally from someone other than me.
- [ ] Two pre-existing bugs were found while reviewing and are deliberately **not** fixed here, to keep this PR to one concern. `Player::SendInitialSpells` announces `m_spellCooldowns.size()` and then skips every entry with `needSendToClient == false`, so `SMSG_INITIAL_SPELLS` carries a count higher than the number of entries written whenever anyone holds a category cooldown. In practice the client copes, relogging with a Raise Dead cooldown shows the right time left, but the count is still wrong. And `Aura::_UnapplyForTarget` runs the same `SendCooldownEvent` twice for a player caster, two identical blocks a few lines apart, so `AddSpellAndCategoryCooldowns` also runs twice on every aura unapply. Happy to open follow-ups for both.
- [ ] `SPELL_ATTR7_RECAST_ON_RESUMMON` (0x4000) is set on 52150, is defined in `SharedDefines.h` and is still unused. It was not used here because its documented meaning is "recast on resummon", not "do not start the cooldown", but it may be the cleaner gate if a maintainer prefers it.

## Self-review

A fresh-context agent reviewed the change as a maintainer would, three rounds, each on the commit produced by the previous one.

**3 findings, 3 fixed. Rounds stopped at the 3 round cap, so the round 3 fix was not re-reviewed; it is covered by in-game testing instead (the logout case above).**

Reviewed at `4770a05f0` against `origin/master` `7d9027d73`, 5 files, +107 −4. Reviewer: Claude Code, Opus 5, 2026-08-14. The rounds ran on the wider version of the change; the three cross-class edits they cleared were taken back out afterwards to keep this PR to the Death Knight, which is why two of them appear under Known Issues above.

**Round 1** — 1 finding.

1. `Pet::LoadPetFromDB` sends the fake `SMSG_SPELL_GO` to the whole visibility set. With the id translation it now carries 46584, a real spell for every client, so bystanders would see a Raise Dead cast that never happened every time a DK's ghoul is restored. **Fixed**: sent to the owner alone on the translated path, every other class keeps the broadcast.

**Round 2** — 1 finding.

2. `Player::RemovePet` cleared `m_temporaryUnsummonedPetNumber` for every class. A warlock or hunter dying mounted inside a battleground skips the reagent block that used to clear it, so the pet stopped coming back on dismount and a soul shard was wasted. **Fixed**: the clear only runs when a cooldown was actually converted, which is the ghoul's case alone.

**Round 3** — 1 finding.

3. Logging out mounted with the ghoul parked left Raise Dead with no cooldown at all: the new guard keeps the infinity block, `LogoutPlayer` calls `RemovePet(nullptr, PET_SAVE_AS_CURRENT)` which missed the conversion, and infinity cooldowns are never persisted. On relog the icon was free and a new ghoul cost nothing. **Fixed**: the logout path starts the real cooldown, and it runs before `SaveToDB` so it reaches `character_spell_cooldown`. Verified in-game afterwards.

Both reviewers covered every changed file, verified the DBC claims themselves rather than taking them from the description, and checked every caller of the touched functions. Neither reviewer had access to the planning documents.

## Research notes

**How "cooldown on event" works here.** `Player::m_spellCooldowns` maps spell id to `{ end, category, itemid, maxduration, needSendToClient, sendToSpectator }`. There is no per-category map, so category blocking is emulated by writing one entry per sibling spell in `AddSpellAndCategoryCooldowns`, and `HasSpellCooldown` is a plain `find(spellId)`. Siblings get `needSendToClient = false` when the cast spell is `IsCooldownStartedOnEvent()`, which is exactly our case, and `SendInitialSpells` skips those on login.

**The chain, with Master of Ghouls.** The DK casts 46584. `Spell::SendSpellCooldown` returns early because of the attribute, so no server cooldown, and the client blocks the icon on its own. The script summons 52150 and then wipes 46584's entry, sending `SMSG_CLEAR_COOLDOWN` to a client that was correctly blocked. The player dies, `Player::setDeathState` reaches `Unit::RemoveMinion`, which calls `SendCooldownEvent(52150)`: the 180s lands on 52150 and on 46584 through the sibling loop, and the packet goes out with 52150, which the client drops. The player resurrects, presses the button, and `Spell::CheckCast` finds the entry on 46584 and answers `SPELL_FAILED_NOT_READY`. Relogging resyncs both sides, hence the "sometimes it fixes itself".

**Prior art.** TrinityCore hit this in [TrinityCore/TrinityCore#15862](https://github.com/TrinityCore/TrinityCore/issues/15862) and fixed it with the same insight, the category as the link between the triggered spell and the spellbook one. TC starts an on-hold cooldown on 46584 at cast time and resolves the category in `SpellHistory::SendCooldownEvent`. AzerothCore already gets the on-hold half for free through `Unit::SetMinion`, which applies the infinity cooldown, so what was missing was only the lookup. VMaNGOS predates the mechanic and has nothing comparable.

**Patch history.**

| Patch | Change | Applies to 3.3.5a |
| --- | --- | --- |
| 3.0.2 | Raise Dead added: one ghoul, 1 minute, 3 minute cooldown, humanoid corpse or Corpse Dust | yes |
| 3.0.2 | Master of Ghouls (52143) makes the ghoul a permanent controllable pet (52150) | yes |
| 3.0.8 | Raise Ally added, and it does not share a cooldown with Raise Dead | yes |
| 3.1.0 | Ghoul scaling and ability adjustments | yes |
| 3.2.x | Night of the Dead reduces Raise Dead and Army of the Dead by the values now in 55620 / 55623 | yes, and it matches this server's DBC |
| 4.x Cataclysm | Reagent dropped, ghoul permanent for Unholy without the talent, cooldown changed | no, do not apply |

### DIFF
--- a/src/server/game/Entities/Pet/Pet.cpp
+++ b/src/server/game/Entities/Pet/Pet.cpp
@@ -392,14 +392,24 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petEntry, uint32 petnumber, bool c
     /// @todo pets should be summoned from real cast instead of just faking it?
     if (petInfo->CreatedBySpellId && spellInfo && (spellInfo->CategoryRecoveryTime > 0 || spellInfo->RecoveryTime > 0))
     {
+        // The death knight ghoul is summoned by a spell the client never learns, its spellbook entry is another one
+        uint32 castSpellId = petInfo->CreatedBySpellId;
+        uint32 spellbookSpellId = owner->GetSpellbookSpellForCooldown(spellInfo);
+        if (spellbookSpellId)
+            castSpellId = spellbookSpellId;
+
         WorldPacket data(SMSG_SPELL_GO, (8 + 8 + 4 + 4 + 2));
         data << owner->GetPackGUID();
         data << owner->GetPackGUID();
         data << uint8(0);
-        data << uint32(petInfo->CreatedBySpellId);
+        data << uint32(castSpellId);
         data << uint32(256); // CAST_FLAG_UNKNOWN3
         data << uint32(0);
-        owner->SendMessageToSet(&data, true);
+        // The translated id is a real spell for every client, so bystanders would see a cast that never happened
+        if (spellbookSpellId)
+            owner->SendDirectMessage(&data);
+        else
+            owner->SendMessageToSet(&data, true);
     }
 
     owner->SetMinion(this, true);
--- a/src/server/game/Entities/Player/Player.cpp
+++ b/src/server/game/Entities/Player/Player.cpp
@@ -9293,6 +9293,9 @@ void Player::RemovePet(Pet* pet, PetSaveMode mode, bool returnreagent)
             return;
     }
 
+    // Captured here because the reagent block below clears m_temporaryUnsummonedPetNumber
+    bool removedWhileTemporarilyUnsummoned = !pet && m_temporaryUnsummonedPetNumber;
+
     if (returnreagent && (pet || (m_temporaryUnsummonedPetNumber && (!m_session || !m_session->PlayerLogout()))) && !InBattleground())
     {
         //returning of reagents only for players, so best done here
@@ -9321,6 +9324,16 @@ void Player::RemovePet(Pet* pet, PetSaveMode mode, bool returnreagent)
 
     if (!pet)
     {
+        // The "pet is out" block is an infinity cooldown and those are never written to the DB, so it has
+        // to become the real one once the pet is gone for good or the session ends while it is away
+        SpellInfo const* petSpellInfo = nullptr;
+        if (removedWhileTemporarilyUnsummoned)
+        {
+            petSpellInfo = sSpellMgr->GetSpellInfo(m_oldpetspell);
+            if (petSpellInfo && !petSpellInfo->IsCooldownStartedOnEvent())
+                petSpellInfo = nullptr;
+        }
+
         if (mode == PET_SAVE_NOT_IN_SLOT && m_petStable && m_petStable->CurrentPet)
         {
             // Handle removing pet while it is in "temporarily unsummoned" state, for example on mount
@@ -9332,6 +9345,18 @@ void Player::RemovePet(Pet* pet, PetSaveMode mode, bool returnreagent)
 
             m_petStable->UnslottedPets.push_back(std::move(*m_petStable->CurrentPet));
             m_petStable->CurrentPet.reset();
+
+            if (petSpellInfo)
+            {
+                SendCooldownEvent(petSpellInfo);
+                // Nothing left to resummon, the pet is out of the slot now
+                m_temporaryUnsummonedPetNumber = 0;
+            }
+        }
+        else if (petSpellInfo && m_session && m_session->PlayerLogout())
+        {
+            // Runs before SaveToDB, so the cooldown reaches character_spell_cooldown
+            SendCooldownEvent(petSpellInfo);
         }
 
         return;
@@ -11277,11 +11302,65 @@ void Player::ModifySpellCooldown(uint32 spellId, int32 cooldown)
     SendDirectMessage(&data);
 }
 
+uint32 Player::GetSpellbookSpellForCooldown(SpellInfo const* spellInfo) const
+{
+    if (!spellInfo)
+        return 0;
+
+    // The spell itself is in the spellbook, so there is nothing to translate
+    if (HasActiveSpell(spellInfo->Id))
+        return 0;
+
+    uint32 category = spellInfo->GetCategory();
+    if (!category)
+        return 0;
+
+    SpellCategoryStore::const_iterator i_scstore = sSpellsByCategoryStore.find(category);
+    if (i_scstore == sSpellsByCategoryStore.end())
+        return 0;
+
+    for (auto i_scset = i_scstore->second.begin(); i_scset != i_scstore->second.end(); ++i_scset)
+    {
+        // Item categories share their ids with spell ones, they are not siblings
+        if (i_scset->first)
+            continue;
+
+        SpellInfo const* categorySpellInfo = sSpellMgr->GetSpellInfo(i_scset->second);
+        if (!categorySpellInfo || categorySpellInfo->SpellFamilyName != spellInfo->SpellFamilyName)
+            continue;
+
+        if (HasActiveSpell(i_scset->second))
+            return i_scset->second;
+    }
+
+    return 0;
+}
+
 void Player::SendCooldownEvent(SpellInfo const* spellInfo, uint32 itemId /*= 0*/, Spell* spell /*= nullptr*/, bool setCooldown /*= true*/)
 {
     // start cooldowns at server side, if any
     if (setCooldown)
         AddSpellAndCategoryCooldowns(spellInfo, itemId, spell);
+
+    // The spell triggering the event may not be in the client's spellbook (the death knight ghoul is
+    // summoned by one of those), so the client also needs the event for the spell it does know
+    if (!itemId && spellInfo->IsCooldownStartedOnEvent())
+    {
+        if (uint32 spellbookSpellId = GetSpellbookSpellForCooldown(spellInfo))
+        {
+            SpellCooldowns::iterator itr = m_spellCooldowns.find(spellbookSpellId);
+            if (itr != m_spellCooldowns.end())
+            {
+                // Otherwise SendInitialSpells skips it and the cooldown is lost on relog
+                itr->second.needSendToClient = true;
+            }
+
+            WorldPacket spellbookData(SMSG_COOLDOWN_EVENT, 4 + 8);
+            spellbookData << uint32(spellbookSpellId);
+            spellbookData << GetGUID();
+            SendDirectMessage(&spellbookData);
+        }
+    }
 
     // Send activate cooldown timer (possible 0) at client side
     WorldPacket data(SMSG_COOLDOWN_EVENT, 4 + 8);
--- a/src/server/game/Entities/Player/Player.h
+++ b/src/server/game/Entities/Player/Player.h
@@ -1820,6 +1820,7 @@ public:
     [[nodiscard]] bool HasSpellCooldown(uint32 spell_id) const override;
     [[nodiscard]] bool HasSpellItemCooldown(uint32 spell_id, uint32 itemid) const override;
     [[nodiscard]] uint32 GetSpellCooldownDelay(uint32 spell_id) const;
+    [[nodiscard]] uint32 GetSpellbookSpellForCooldown(SpellInfo const* spellInfo) const;
     void AddSpellAndCategoryCooldowns(SpellInfo const* spellInfo, uint32 itemId, Spell* spell = nullptr, bool infinityCooldown = false);
     void AddSpellCooldown(uint32 spell_id, uint32 itemid, uint32 end_time, bool needSendToClient = false, bool forceSendToSpectator = false) override;
     void _AddSpellCooldown(uint32 spell_id, uint16 categoryId, uint32 itemid, uint32 end_time, bool needSendToClient = false, bool forceSendToSpectator = false);
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -7945,7 +7945,21 @@ void Unit::SetMinion(Minion* minion, bool apply)
             SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(minion->GetUInt32Value(UNIT_CREATED_BY_SPELL));
             // Remove infinity cooldown
             if (spellInfo && spellInfo->IsCooldownStartedOnEvent())
-                ToPlayer()->SendCooldownEvent(spellInfo);
+            {
+                // A pet put away for a while comes back untouched, so its cooldown must not start yet: the
+                // infinity cooldown stays and Pet::LoadPetFromDB restores the client side block on resummon
+                bool temporarilyUnsummoned = false;
+                if (Pet* pet = minion->ToPet())
+                {
+                    CharmInfo* charmInfo = pet->GetCharmInfo();
+                    bool hasCooldown = spellInfo->CategoryRecoveryTime > 0 || spellInfo->RecoveryTime > 0;
+                    temporarilyUnsummoned = charmInfo && hasCooldown
+                        && ToPlayer()->GetTemporaryUnsummonedPetNumber() == charmInfo->GetPetNumber();
+                }
+
+                if (!temporarilyUnsummoned)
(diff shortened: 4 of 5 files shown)


src/server/game/Entities/Pet/Pet.cpp
```diff
@@ -392,14 +392,24 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petEntry, uint32 petnumber, bool c
     /// @todo pets should be summoned from real cast instead of just faking it?
     if (petInfo->CreatedBySpellId && spellInfo && (spellInfo->CategoryRecoveryTime > 0 || spellInfo->RecoveryTime > 0))
```
sogladev, line 393: > // The death knight ghoul is summoned by a spell the client never learns, its spellbook entry is another one

This is true. There's a potential branch that may  be missed where `spellInfo == nullptr` that does set the CD correctly.

```if (petInfo->CreatedBySpellId && spellInfo && (spellInfo->CategoryRecoveryTime > 0 || spellInfo->RecoveryTime > 0))``` used to be ``` if (petInfo->CreatedBySpellId)```
, but the extra condition fixed visual issues with pet summoning where a fake cast was sent to the client. This fake cast presumably could fix the ghoul summon spell.

Perhaps we revert the PR and introduce a branch so that the fake summon spell is sent for the summon ghoul, but not for the other pets.
 
- https://github.com/azerothcore/azerothcore-wotlk/pull/19277

sudlud:
the category-as-the-link idea comes from TrinityCore `ecbbcc8972` (Shauren, 2015-11-14, closes TC #15862), down to the reasoning in the comment. our AGENTS rule wants credit on a mirrored mechanism even when it's rewritten against AC, so could you add a follow-up commit with `Co-authored-by: Shauren <shauren.trinity@gmail.com>`? nothing already pushed needs rewriting.

and can you please address @sogladev's review on `Pet.cpp:393`, it's still open. fwiw #19277's `catrec > 0 || rec > 0` guard already limits the fake cast to the ghoul (every warlock and hunter summon has both at 0), so I think what he's asking for is what the PR already does. worth saying that on the thread.

non-blocking:

- `Player.cpp:11347`: the new block sits outside `if (setCooldown)`, so the two callers passing `false` (`Totem.cpp:165`, `Spell.cpp:4654`) would send the client event with no server cooldown behind it. not reachable with 3.3.5a data, but TC gates its equivalent on the same flag.
- `Player.cpp:11354`: the cooldown isn't lost on relog, `_SaveSpellCooldowns` writes the row either way and keeps the flag in `needSend`. only the display is skipped.
- your non-DK regression TODO is already covered by the data: 52150 is the only event-cooldown pet summon with a recovery time, and warlock and hunter summons have no cooldown at all.

Chinoske:
@sogladev I checked the spell data before deciding: `46584` (the spellbook entry) and `46585` (the one that actually summons) both sit in category 1245 with `CategoryRecoveryTime` 180000. So the guard added in #19277 is satisfied for the ghoul and this branch does run, which is why I'd rather not revert and gate the fake cast per summon.

the `spellInfo == nullptr` case you point at is a pet whose `CreatedBySpellId` isn't a real spell, and the ghoul is never that. still happy to add the branch if you'd prefer it explicit.

@sudlud added the `Co-authored-by: Shauren` trailer in `df2ab8bb8`, on a new commit, plus a source note on the helper itself. cherry-pick box is ticked now.

PR #27183, opened 2026-08-16
feat(Core/Config): make taxi flight speed configurable

moostigre:
## Changes Proposed:
Allow the taxi speed to be configurable so users can speed them if they want to

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


PkllonG:
After testing, the maximum value is 50.0f; anything higher has no effect.

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -1635,6 +1635,13 @@ DisableWaterBreath = 4
 
 AllFlightPaths = 0
 
+#
+#    TaxiFlightSpeed
+#        Description: Movement speed used while players travel along taxi flight paths.
+#        Default:     32.0
```
sogladev:
```suggestion
#        Default:     32.0 - (Blizzlike)
#                            50.0  - (Maximum)
```

sogladev (changes requested):

moostigre:
thanks for the test and suggestions ! 👍

src/server/game/World/WorldConfig.cpp
```diff
@@ -261,6 +261,7 @@ void WorldConfig::BuildConfigCache()
 
     SetConfigValue<uint32>(CONFIG_MAX_RECRUIT_A_FRIEND_BONUS_PLAYER_LEVEL_DIFFERENCE, "RecruitAFriend.MaxDifference", 4);
     SetConfigValue<bool>(CONFIG_ALL_TAXI_PATHS, "AllFlightPaths", false);
+    SetConfigValue<float>(CONFIG_TAXI_FLIGHT_SPEED, "TaxiFlightSpeed", 32.0f, ConfigValueCache::Reloadable::Yes, [](float const& value) { return value > 0.0f; }, "> 0");
```
sogladev:
```suggestion
    SetConfigValue<float>(CONFIG_TAXI_FLIGHT_SPEED, "TaxiFlightSpeed", 32.0f, ConfigValueCache::Reloadable::Yes, [](float const& value) { return value > 0.0f && value <= 50.0f; }, "> 0 && <= 50");
```

sogladev (dismissed):
Thanks you for your PR! I could push small stuff like this but `gh` is broken atm.

src/server/game/World/WorldConfig.cpp
moostigre: you want to enforce the value between 32 and 50 ? Guess its completely logical yes, I should have thought about it

src/server/game/World/WorldConfig.cpp
```diff
@@ -261,6 +261,7 @@ void WorldConfig::BuildConfigCache()
 
     SetConfigValue<uint32>(CONFIG_MAX_RECRUIT_A_FRIEND_BONUS_PLAYER_LEVEL_DIFFERENCE, "RecruitAFriend.MaxDifference", 4);
     SetConfigValue<bool>(CONFIG_ALL_TAXI_PATHS, "AllFlightPaths", false);
+    SetConfigValue<float>(CONFIG_TAXI_FLIGHT_SPEED, "TaxiFlightSpeed", 32.0f, ConfigValueCache::Reloadable::Yes, [](float const& value) { return value > 0.0f && value <= 50.0f; }, "> 0 && <= 50");
```
sudlud: `> 0` is a bit looser than the spline layer accepts: `MoveSplineInitArgs::Validate` requires `velocity > 0.01f`, so anything in (0, 0.01] passes this check and then `MoveSplineInit::Launch` bails out before initializing the spline. `DoReset` has already set `UNIT_STATE_IN_FLIGHT` and `UNIT_FLAG_DISABLE_MOVE` at that point and doesn't look at the return value, so the player ends up rooted in taxi state with no flight. Maybe raise the minimum to something sane?

Also, the `50` here is the same limit as the clamp in `MoveSplineInit::Launch`. Could we add a short comment pointing at it so the two don't drift?

src/server/apps/worldserver/worldserver.conf.dist
```diff
@@ -1635,6 +1635,14 @@ DisableWaterBreath = 4
 
 AllFlightPaths = 0
 
+#
+#    TaxiFlightSpeed
+#        Description: Movement speed used while players travel along taxi flight paths.
+#        Default:     32.0 - (Blizzlike)
+#                     50.0  - (Maximum)
```
sudlud: The file has a `Range:` field for this (see `CharactersPerRealm`), and it goes above `Default:`. Something like:

    #        Range:       0.01-50.0
    #        Default:     32.0 - (Blizzlike)

Worth knowing that a value above 50 is rejected and falls back to 32 rather than clamping to 50, so the current "Maximum" wording is slightly misleading.

src/server/game/World/WorldConfig.cpp
moostigre: i added a comment in the code and the config

src/server/apps/worldserver/worldserver.conf.dist
moostigre: i put a comment and added range 0.01 - 50

sogladev (approved):

PR #27195, opened 2026-08-17
fix(Scripts/ICC): keep the Raging Spirit summon on the Arthas Platform

### LINKED ISSUE #18974: ICC Raging spirit falls under the texture
### Current Behaviour

Raging spirit in P2.5 kicks player and then falls under the texture
(after ~4 minutes, it comes out from under the textures and performs its functions)
Video 1: https://youtu.be/dba8aBnHq58?si=B-LaKphcMyV2vxPC
Video 2 : https://youtu.be/b7POiVKbwwI?si=G3GduoDuVc1Tx9oP

https://youtu.be/_QcuB8zCpu4?si=wN1xrl2fj6kE_QZk
The Spirit must appear immediately and fulfill its functions

### Source

https://wowwiki-archive.fandom.com/wiki/Lich_King_(Icecrown_Citadel_tactics)#Phase_1.5_and_2.5

1.Go to Lich king and fight
2.IN P2.5 wait for spawn first Raging Spirit

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. <a href="https://github.com/azerothcore/azerothcore-wotlk/commit/15323bb9fbe47fd6a87f216437dbd8be3c21e0b5" target="_blank">15323bb</a> 2024-05-29 06:47:51 +0300 (TS branch) (Unix, RelWithDebInfo, Static)

### Operating system

Ubuntu 22.04.4 LTS.

### Custom changes or Modules

List enable modules:
- mod-cfbg
- mod-congrats-on-level
- mod-duel-reset
- mod-globalchat
- mod-individual-xp
- mod-transmog

EricksOliveira:
## What and why

In phase 2.5 the first Raging Spirit spawns under the Arthas Platform and stays there until it evades, which is https://github.com/azerothcore/azerothcore-wotlk/issues/18974. Spell 69201 summons it 5 yards in front of the player it hit, and when that spot has no collision under it the destination falls back to the terrain height of map 631, around 990 yards below the platform floor. This clamps the summon destination back onto the platform.

## Decisions worth knowing

An earlier version of this PR filtered the Lich King's target selection instead. That was dropped: the threat list already drops off-platform players once a second through `CanAIAttack`, so the filter only covered the gap between two threat updates. The fix now sits on the destination, in `spell_the_lich_king_summon_raging_spirit`.

When the destination is off the platform it goes back to the caster's own position rather than to a computed point on the platform, since a player standing there is proof that spot has ground.

`IsValidPlatformTarget` was split so the same platform test can run on a bare position.

## Review guide

`boss_the_lich_king.cpp`, the new `spell_the_lich_king_summon_raging_spirit`, is where the judgment is: whether `TARGET_DEST_CASTER_FRONT` is the only destination this spell can take, and whether falling back to the caster is acceptable given the 5 yard offset is blizzlike.

The `IsValidPlatformPosition` split and the `spell_script_names` row are mechanical.

## Known gaps

The rework has not been tested in game yet. The in-game testing recorded on this PR was done against the earlier target selector version.

There is no recorded repro pinning which of the two paths actually fires in practice: a player on the rim facing outward, or the window where the platform gameobject sits in its destroyed state during the second Remorseless Winter, which turns its collision off. Both end in the same fallback, and the clamp covers both.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/18974

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/boss_the_lich_king.cpp
@@ -455,6 +455,30 @@ private:
     Creature const* _source;
 };
 
+struct RagingSpiritTargetSelector
+{
+public:
+    RagingSpiritTargetSelector(Creature* source) : _source(source) { }
+    bool operator()(Unit const* target) const
+    {
+        if (!target)
+            return false;
+        if (!target->IsAlive())
+            return false;
+        if (!target->IsPlayer())
+            return false;
+        if (_source->GetExactDist(target) > 100.0f)
+            return false;
+        // the spirit is summoned at the target's own position, so the target must be on the platform
+        if (!IsValidPlatformTarget(target) || target->GetVehicle())
+            return false;
+        return true;
+    }
+
+private:
+    Creature const* _source;
+};
+
 class FrozenThroneResetWorker
 {
 public:
@@ -1084,7 +1108,7 @@ public:
                     events.ScheduleEvent(EVENT_SUMMON_ICE_SPHERE, 7500ms, EVENT_GROUP_ABILITIES);
                     break;
                 case EVENT_SUMMON_RAGING_SPIRIT:
-                    if (Unit* target = SelectTarget(SelectTargetMethod::Random, 0, 100.0f, true))
+                    if (Unit* target = SelectTarget(SelectTargetMethod::Random, 0, RagingSpiritTargetSelector(me)))
                         me->CastSpell(target, SPELL_RAGING_SPIRIT, false);
                     events.ScheduleEvent(EVENT_SUMMON_RAGING_SPIRIT, (!HealthAbovePct(40) ? 15s : 20s), EVENT_GROUP_ABILITIES);
                     break;
--- a/src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
@@ -1804,6 +1804,9 @@ public:
                     if (GameObject* platform = instance->GetGameObject(ArthasPlatformGUID))
                     {
                         platform->SetDestructibleState(GO_DESTRUCTIBLE_DESTROYED);
+                        // the destroyed state is a client side visual only - keep the server side floor,
+                        // or creatures summoned during this window spawn below the platform
+                        platform->EnableCollision(true);
                         Events.ScheduleEvent(EVENT_REBUILD_PLATFORM, 1500ms);
                     }
                     break;
--- a/src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
+++ b/src/server/scripts/Northrend/IcecrownCitadel/instance_icecrown_citadel.cpp
@@ -1804,9 +1804,6 @@ public:
                     if (GameObject* platform = instance->GetGameObject(ArthasPlatformGUID))
                     {
                         platform->SetDestructibleState(GO_DESTRUCTIBLE_DESTROYED);
-                        // the destroyed state is a client side visual only - keep the server side floor,
-                        // or creatures summoned during this window spawn below the platform
-                        platform->EnableCollision(true);
                         Events.ScheduleEvent(EVENT_REBUILD_PLATFORM, 1500ms);
                     }
                     break;


EricksOliveira:
# Self-review — fixRagingSpiritP2 → master

**Outcome** 3 findings — 1 fixed, 2 dismissed — final round yielded no fixes

**Caveat** The codestyle linters (`apps/codestyle/codestyle-cpp.py`) could not run — no Python interpreter on this machine. Style was verified by hand and with `git diff --check`.

**Reviewed** `b8ec876a8` (`fixRagingSpiritP2` vs `upstream/master` `77424db6d`, merge-base diff — 1 file, +25 −1)

**By** Claude Code (Claude Agent SDK), claude-opus-5 — self-review v0.3, 2026-08-17

**Tested in-game by the author** Raging Spirit spawns and stays on the platform in P2.5 and in P1.5; each spirit unroots after ~3s and immediately engages a valid target; platform still damages, destroys, rebuilds and resets correctly across a wipe and re-pull. The platform run predates the round-1 fix, which reverted `instance_icecrown_citadel.cpp` to its master state — that path is therefore unchanged from master.

<details>

**Intent** During The Lich King's transition phases (P1.5 after the first Remorseless Winter, P2.5 after the second), a summoned Raging Spirit must appear on the Arthas Platform and immediately perform its function — acquire and attack a valid target — instead of ending up below the platform, where it stays inert for minutes until an evade brings it home.

**Project rules** `.agents/docs/self-review-rules.md` found and applied — content-script weighting (blast radius contained to the Lich King encounter), shared instance script audited for other consumers of the state touched, in-game testing recorded above.

**Procedural** Both rounds ran in an isolated fresh context with the planning home and this report excluded. Governing docs run as a checklist against every changed file: `AGENTS.md`, `.agents/docs/cpp-guidelines.md`, `.agents/docs/cpp-scripts.md`, `.agents/docs/code-review.md`, `.editorconfig`. Round 1's fix is covered by round 2, which reviewed the changeset after the revert.

## Round 1 — `7ecd6e695`, 1 finding

1. `instance_icecrown_citadel.cpp:1809` — `EnableCollision(true)` is unreachable for its stated purpose: the no-collision window closes at 1500ms via `EVENT_REBUILD_PLATFORM` (`GO_DESTRUCTIBLE_REBUILDING` already re-enables collision) while the first Raging Spirit is summoned at 4s, and P1.5 never reaches the destroyed state at all; the call also re-enabled the *destroyed* model's collision, not the intact platform floor → **fixed**: line and comment removed, `instance_icecrown_citadel.cpp` now byte-identical to master

## Round 2 — `b8ec876a8`, 2 findings, no fixes

2. `7ecd6e695` and `b8ec876a8` (commit messages) — the first claims "two independent causes are addressed" and describes the collision change the second commit removes; the second's subject is a bare `fix`, off the project's `type(Scope): description` convention → **dismissed**: the author will rewrite the PR description when opening the PR

3. `boss_the_lich_king.cpp:470` — `_source->GetExactDist(target) > 100.0f` can never be true: with the caster at `CenterPosition`, `IsValidPlatformTarget` already bounds a passing target to ≈96.3 yd in 3D → **dismissed**: it preserves the 100 yd search radius that master's `SelectTarget(..., 100.0f, true)` expressed and matches the shape of the sibling selectors in this file; it regains effect if `IsValidPlatformTarget` is ever loosened

</details>

Francesco-Borzi:
@EricksOliveira I see you are using an older version of the `self-review` skill. It's ok for this PR, there is no need to re-execute it, but please update your `agent-toolkit` repo (just `git pull` inside it) to update it. I suggest doing this regularly.

EricksOliveira:
> @EricksOliveira I see you are using an older version of the `self-review` skill. It's ok for this PR, there is no need to re-execute it, but please update your `agent-toolkit` repo (just `git pull` inside it) to update it. I suggest doing this regularly.

Yes. I've updated it. Thanks for the feedback.

sudlud:
just wondering: does the selector change anything? `boss_the_lich_kingAI::CanAIAttack` is already `IsValidPlatformTarget(target) && !target->GetVehicle()`, and that's applied to the whole threat list, not just victim selection: `ThreatManager::Update` ticks every 1s, `ReselectVictim` runs `UpdateOffline` on every ref, `ShouldBeOffline` goes through `CanCreatureAttack` into `CanAIAttack`, and `SelectTargetList` skips offline refs. So master shouldn't be able to pick an off-platform player except in the <=1s gap before the next tick, which feels too narrow for something that hits the first spirit of every P2.5. Do you still have the repro? would be good to see one actually getting selected on master.

second thing, maybe the real cause: the spirit isn't summoned at the player's position. 69201 is `TARGET_DEST_CASTER_FRONT` with radius index 8 (5.0 yd), so `SelectImplicitCasterDestTargets` does `MovePositionToFirstCollision(pos, 5.0f, 0)` and it lands 5 yards in front of the player along their facing. `IsValidPlatformTarget` is basically `z > 840` against a floor at 840.86, so someone on the rim facing outward passes and the summon still goes past the edge. that'd explain why it's reliable, everyone spreads to the rim during Remorseless Winter. if so the fix belongs on the destination (an `OnDestinationTargetSelect` clamp on 69201), and either way the comment on the `IsValidPlatformTarget` line isn't accurate.

and ignore the coderabbit LOS comment, master never had one: `100.0f, true` hits the non-template overload where `true` is `playerOnly`, and `DefaultTargetSelector` has no `IsWithinLOSInMap`.

EricksOliveira:
you're right, the refs go offline every second through CanCreatureAttack into CanAIAttack and
SelectTargetList skips them, so the selector only covered the sub second window. and no, I don't
have a log from master, just my own runs in game.

on the dest: when UpdateAllowedPositionZ finds no ground there, CheckCollisionAndGetValidCoords
falls back to GetGridHeight, and the terrain under the Frozen Throne is around -149. so the spirit
lands ~990 yd below the 840.86 floor, which is the "falls under the texture" in the report.

dropped the selector and did the OnDestinationTargetSelect clamp on 69201: if the dest is off the
platform it goes back to the caster's position. that also covers the window where the platform GO
is destroyed during the second Remorseless Winter, same hole.

Gultask:
Wouldn't it be better to modify the summon's NPC script rather than hacking the spell? Is there a way to check if it's not pathing correctly on the creature's side and tp?
For what it's worth, it's common that creatures that can't find a path teleport to the player they're targeting. Although in this case I'm not sure if the player should be kiting the spirit or if the spirit oneshots player or wtv

EricksOliveira:
The summon is cast by the player, and npc_raging_spirit already gets them as the summoner in IsSummonedBy, so the fix fits there. I'll drop the spell script and the spell_script_names row, and have the spirit NearTeleportTo its summoner when it spawns off the platform.

About the mechanic: the spirit isn't a kite target and doesn't one-shot. It melees and Soul Shrieks, and the raid tanks and kills it, so a teleport to the player doesn't change the fight. I kept the check at spawn instead of making it a general "can't path, so teleport" in UpdateAI: the bug is only where it spawns, and players on the rim are already dropped by CanAIAttack.

Gultask (approved):

PR #27272, opened 2026-08-23
docs(agents): key the self-review rule to opening a PR

sudlud:
## Changes Proposed:

The "Mandatory reading per task" bullet for self-review currently reads:

> Self-reviewing a changeset before submission → also `.agents/docs/self-review-rules.md`

That only fires once you've already decided to self-review, which is circular: an agent
whose task is "fix X and open a PR" never classifies itself as self-reviewing, so it
never opens the doc. Naming the PR action in the trigger puts the obligation where the
work actually happens.

It also routes `code-review.md` for PR authors, not just reviewers, via the existing
`also` idiom. The rule about walking every existing comment lives there and applies
just as much when the PR is your own, so this adds the bot-reviews emphasis to that
rule rather than restating it in the routing line.

Found the hard way on #27255: I opened it without self-reviewing, and the review I then
ran caught a fix of mine that was completely inert. I had also not read the CodeRabbit
and Copilot comments before calling it ready, and one of them was a real finding.

This PR proposes changes to:

Agent docs only.

### AI-assisted Pull Requests

## Issues Addressed:
- Closes

## Tests Performed:

Docs only, nothing to run.

### DIFF
--- a/.agents/docs/code-review.md
+++ b/.agents/docs/code-review.md
@@ -26,10 +26,10 @@ line, not as background reading.
   call leaves a null result, not just what enables it.
 - Check the change is still needed against current `master`: the surrounding code may have moved,
   or another change may have landed the same fix.
-- On an existing PR, walk every discussion item one by one, bot reviews included: what was raised, whether it was
-  answered, and whether it still applies to the current head. Never skip one because it looks
-  resolved, old, or minor; this walk overrides any read-comments-lightly default of the reviewing
-  skill. `gh pr view` misses review bodies and inline threads; pull all three:
+- On an existing PR, walk every discussion item one by one, bot reviews included: what was raised,
+  whether it was answered, and whether it still applies to the current head. Never skip one
+  because it looks resolved, old, or minor; this walk overrides any read-comments-lightly default
+  of the reviewing skill. `gh pr view` misses review bodies and inline threads; pull all three:
 
   ```
   gh api repos/azerothcore/azerothcore-wotlk/issues/<N>/comments --paginate  # conversation comments
--- a/.agents/docs/code-review.md
+++ b/.agents/docs/code-review.md
@@ -26,7 +26,7 @@ line, not as background reading.
   call leaves a null result, not just what enables it.
 - Check the change is still needed against current `master`: the surrounding code may have moved,
   or another change may have landed the same fix.
-- On an existing PR, walk every discussion item one by one: what was raised, whether it was
+- On an existing PR, walk every discussion item one by one, bot reviews included: what was raised, whether it was
   answered, and whether it still applies to the current head. Never skip one because it looks
   resolved, old, or minor; this walk overrides any read-comments-lightly default of the reviewing
   skill. `gh pr view` misses review bodies and inline threads; pull all three:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,7 +19,8 @@ Read the matching doc(s) BEFORE starting the task:
 - Creating or modifying SQL → `.agents/docs/sql-guidelines.md`
   - SmartAI work (`smart_scripts` data) → also `.agents/docs/cpp-scripts.md`
 - Reviewing a changeset or PR → `.agents/docs/code-review.md`
-- Self-reviewing a changeset before submission → also `.agents/docs/self-review-rules.md`
+- Opening or updating a PR → `.agents/docs/code-review.md` + `.agents/docs/self-review-rules.md`:
+  self-review before submitting; read existing comments, bot reviews included
 - Touching a subsystem that has a doc in `.agents/docs/systems/` → read that doc too
 - Capturing a lesson or adding/updating agent docs → `.agents/docs/README.md`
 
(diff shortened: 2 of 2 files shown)


sudlud:
# Self-review — docs/agents-self-review-before-pr → master

**Outcome** 3 findings — 2 fixed, 1 dismissed — round 3 clean

**Reviewed** `f7e4ea1e4a` (`docs/agents-self-review-before-pr` vs `master`, merge-base diff — 2 files, +5 −5)

**By** Claude Code CLI, claude-opus-5 — self-review v0.7, 2026-08-23

**Verification** Docs only, nothing to run. The round 3 reviewer confirmed the final rewrap is word-identical to the version before it by tokenizing both and comparing (63 tokens each side, and identical hashes with all whitespace stripped), rather than by reading.

<details>

**Intent** `AGENTS.md`'s "Mandatory reading per task" bullet routed `.agents/docs/self-review-rules.md` for the task "self-reviewing a changeset before submission". That trigger is circular: an agent whose task is "fix X and open a PR" never classifies itself as self-reviewing, so it never reads the doc and skips self-review. Re-key the bullet so it also fires on opening or updating a PR.

**Project rules** `.agents/docs/self-review-rules.md` found and applied, with `.agents/docs/code-review.md` on top. Pinned-review rule honoured: every round ran on a commit, fixes committed before the next round.

**Diff** `e7ce5a487ec727000cf30a9fd2c6bb5d18545fda`

**Procedural** Reviewers ran as fresh-context subagents with `.agents/plans/` and this report excluded. `codestyle-cpp.py` / `codestyle-sql.py` are N/A. This review was run *after* the PR branch was pushed but *before* the PR was opened, which is the sequence the change itself is about.

## Round 1 — `6cb30cef2d`, 2 findings

1. `AGENTS.md:22-23` — re-keying the trigger solely to "opening or updating a PR" dropped the plain "self-review this branch" task, which then matched no bullet at all; `code-review.md` has no back-link to `self-review-rules.md`, so that path silently lost the pinned-commit rule and the regression-risk tiers → **fixed**, the bullet now names both triggers
2. `AGENTS.md:23` — the `read existing comments, bot reviews included` clause put substantive guidance in a routing line, duplicating `code-review.md`'s comment-walk rule at lower fidelity; the two would drift, and a summary in a routing line invites reading the summary instead of the doc → **fixed**, the bot-reviews emphasis moved into that rule in `code-review.md` and the routing line is routing again

Both fixes made the change smaller: `AGENTS.md` went from +10 words in round 1 to +2, while covering more.

## Round 2 — `0a6f384e69`, 1 finding

3. Commit bodies exceed the guidelines' 72-col limit (`6cb30cef2d` 5 lines at 77-78, `0a6f384e69` 4 at 73-75). Unlike the 50-char subject cap, which is not observed upstream (61 of the last 100 subjects exceed it), body width is: only 2 of the last 60 upstream body lines exceed 72 → **dismissed**: AC squash-merges, so the merge commit takes the PR title and description and these bodies never reach master; rewriting history for discarded text is churn

Round 2 also cleared both round 1 fixes, and noted that with the explicit `+ code-review.md` gone, the leading `also` no longer has an antecedent bullet for the PR-opening trigger. `code-review.md` still gets read, but via `self-review-rules.md`'s "The code-review.md rules apply on top" rather than from `AGENTS.md` directly. Working chain, no rule lost, so not a finding — but the guarantee now lives one hop away.

Separately fixed after round 2 without being a finding: the inserted clause had pushed `code-review.md:29` to 113 cols, past the ~98 the rest of that file wraps at. Rewrapped in `f7e4ea1e4a`.

## Round 3 — `f7e4ea1e4a`, clean

No findings. Verified the rewrap moved only line breaks, the bullet still renders as one list item with its fenced code block attached, all rewrapped lines are 91-98 cols, and nothing in the repo still refers to the old bullet wording.

**Noted, not a finding** `code-review.md:17` is 123 cols, over the 120 limit. It predates this change (`f806ca71ef`), is untouched, and its length comes from an unbreakable link URL.

</details>

.agents/docs/code-review.md
```diff
@@ -26,10 +26,10 @@ line, not as background reading.
...
 - Check the change is still needed against current `master`: the surrounding code may have moved,
   or another change may have landed the same fix.
-- On an existing PR, walk every discussion item one by one: what was raised, whether it was
-  answered, and whether it still applies to the current head. Never skip one because it looks
-  resolved, old, or minor; this walk overrides any read-comments-lightly default of the reviewing
-  skill. `gh pr view` misses review bodies and inline threads; pull all three:
+- On an existing PR, walk every discussion item one by one, bot reviews included: what was raised,
+  whether it was answered, and whether it still applies to the current head. Never skip one
+  because it looks resolved, old, or minor; this walk overrides any read-comments-lightly default
+  of the reviewing skill. `gh pr view` misses review bodies and inline threads; pull all three:
```
FrancescoBorzi, line 32: minor note that it is not related to this PR: I think we should get rid of the max length rule here, it just pollutes diffs like this

FrancescoBorzi (approved):

.agents/docs/code-review.md
sudlud: agreed, AI focusess way too much on it

PR #27275, opened 2026-08-23
fix(Core/Events): don't hold a map-store iterator across game event AI hooks

mybaixinyu:
## Changes Proposed:

`GameEventAIHookWorker::Visit` walks `Map::GetObjectsStore()` with a range-for and calls into creature/gameobject AI while still holding the iterator. A `SMART_EVENT_GAME_EVENT_START` script can erase its own creature from that store, which frees the hash node the loop is standing on; the next read of `__next_` returns garbage and the following dereference faults.

Confirmed chain (game event 89 "Leprithus", entry 572, `SMART_ACTION_RESPAWN_TARGET` on self):

```
GameEventAIHookWorker::Visit -> SmartScript::ProcessEventsFor -> ProcessAction
  -> Creature::Respawn -> Map::AddObjectToRemoveList
  -> Unit::CleanupBeforeRemoveFromMap -> Creature::RemoveFromWorld
  -> _objectsStore.Remove<Creature>()        // node freed
```

`Creature::Respawn()` reaches `AddObjectToRemoveList()` through its non-compatibility-mode branch, whose comment reads *"destroy and let ProcessRespawns() recreate"* — it assumes a safe call point, and `RunSmartAIScripts` calls it in the middle of iterating that very container.

The fix walks a snapshot of the guids and resolves each object through the store again, following the convention in `.agents/docs/cpp-guidelines.md`:

> **Long-lived references**: don't store a raw `Player*` / `Creature*` / `Unit*` past the current call/tick — the object can be removed (logout, despawn, instance unload) and the pointer dangles. Store the `ObjectGuid` and resolve at use time via `ObjectAccessor::FindPlayer(guid)`, `Map::GetCreature(guid)`, etc.

The new loop does not depend on knowing *which* script mutates the store: whatever is erased, the re-lookup returns `nullptr` and the entry is skipped.

Two other events ship the same pattern but effectively never fire (`occurence` ~10 years): event 85 (`Stitches`, 13 creatures) and event 87 (`Scarlet Oracle`, 2 creatures). Fixing the traversal covers those too.

This PR proposes changes to:

### AI-assisted Pull Requests

## Issues Addressed:

No AzerothCore issue. Same crash signature as TrinityCore#26687 (open since 2021), TrinityCore#17587 and TrinityCore#7235. The hypothesis there was a null `AI()` on a gameobject; the logging added in #26687 only ever caught a benign null on a transport (`Orgrim's Hammer`), and the reporter noted the crash happens "2 times in years", so it was never reproduced. The actual fault is iterator invalidation on the creature path.

## SOURCE:

- [x] Video evidence, knowledge databases or other public sources: crash report registers and disassembly from the faulting build, plus an instrumented run — both reproduced below.

<details>
<summary>Faulting instruction and registers (arm64, Release build)</summary>

```
exception: EXC_BAD_ACCESS (SIGSEGV)
KERN_INVALID_ADDRESS at 0x67d2c2ce13e9463b

x21 = 0x0000000000000059          ; eventId = 89
x27 = 0x67d2c2ce13e94623          ; hash node pointer, garbage
fault address = x27 + 0x18
```

Disassembly of `GameEventMgr::RunSmartAIScripts` around the fault:

```
<+332>: ldr    x27, [x27]         ; x27 = node->__next_
<+336>: cbz    x27, <+320>
<+340>: ldr    x8,  [x27, #0x18]  ; faults: node->second (Creature*)
<+344>: ldrb   w9,  [x8, #0x5d]   ; IsInWorld() inlined
```

`0x18` is the value offset of a libc++ `__hash_node` whose key is an 8-byte `ObjectGuid`, so `x27` is a node, and the garbage came from the previous node's `__next_` — that node had already been freed.

</details>

<details>
<summary>Instrumented run: the erase, caught in the act</summary>

Temporary probe comparing `size()` / `bucket_count()` around each AI call, plus a backtrace in `Creature::RemoveFromWorld()`:

```
[GEDIAG] creature Leprithus entry 572 mutated creature store during event 89:
         size 11985->11984  buckets 12853->12853
```

Size drops by one with the bucket count unchanged: an erase, not a rehash. `unordered_map::erase` frees the node, which is what leaves the garbage `__next_` above.

```
#01 Creature::RemoveFromWorld()
#02 Unit::CleanupBeforeRemoveFromMap(bool)
#03 Map::AddObjectToRemoveList(WorldObject*)
#04 Creature::Respawn(bool)
#05 SmartScript::ProcessAction(...)
#06 SmartScript::ProcessEventsFor(SMART_EVENT, ...)
#07 GameEventAIHookWorker::Visit(std::unordered_map<ObjectGuid, Creature*>&)
#08 GameEventMgr::RunSmartAIScripts(unsigned short, bool)
```

The probe was removed before this PR; the diff is the fix only.

</details>

## Tests Performed:

This PR has been:
- [ ] Tested in-game by other community members/someone else other than the author.

Built and run on a 3.3.5a realm (macOS/arm64, Release, ~100 bots online). The crash reproduced twice before the change — once from the scheduled 20:00 trigger, once from `.event start 89` — and the same trigger survived after it. The AI hook really ran on the fixed build rather than passing vacuously: Leprithus respawned and its `creature_respawn` row was consumed each time.

Edge cases I could not cover: the gameobject `Visit` path takes the same change but I have no reproducer for it, and I did not test a spawn group in `SPAWNGROUP_FLAG_COMPATIBILITY_MODE`, where `Creature::Respawn` takes the other branch and does not erase.

## How to Test the Changes:

Event 89 (`Leprithus`) runs daily at 20:00 for 600 minutes. Entry 572 carries `smart_scripts` id 3: `event_type 68` (`SMART_EVENT_GAME_EVENT_START`, param 89) -> `action_type 70` (`SMART_ACTION_RESPAWN_TARGET`, target self). `Creature::Respawn()` only reaches the erase when the creature is dead, which is why this is rare on a normal realm — Leprithus is a rare spawn in Duskwood on a 20 hour timer, so it is almost always alive when the event fires.

1. Kill Leprithus (Duskwood, entry 572, spawn 134021) and keep the grid loaded
2. `.event start 89`

Before: worldserver dies with `SIGSEGV` in `GameEventMgr::RunSmartAIScripts`.
After: the event applies, Leprithus respawns, the server keeps running.

- [ ] The `Snapshot()` copy is O(objects) per map per event transition. Measured on the same container sizes: 12 000 objects 28 µs -> 115 µs, 40 000 113 µs -> 391 µs, 150 000 615 µs -> 1806 µs. `RunSmartAIScripts` only runs on event start/stop, and `MinRecordUpdateTimeDiff` defaults to 100 ms, so this stays three orders of magnitude below the threshold at which a slow tick is logged. Still worth a second opinion if someone runs far larger maps.

### DIFF
--- a/src/server/game/Events/GameEventMgr.cpp
+++ b/src/server/game/Events/GameEventMgr.cpp
@@ -1875,14 +1875,16 @@ void GameEventMgr::SendWorldStateUpdate(Player* player, uint16 eventId)
 class GameEventAIHookWorker
 {
 public:
-    GameEventAIHookWorker(Map* map, uint16 eventId, bool activate) : _map(map), _eventId(eventId), _activate(activate) { }
+    GameEventAIHookWorker(Map* map, uint16 eventId, bool activate) :
+        _map(map), _eventId(eventId), _activate(activate) { }
 
     void Visit(std::unordered_map<ObjectGuid, Creature*>& creatureMap)
     {
         for (ObjectGuid const& guid : Snapshot(creatureMap))
         {
             Creature* creature = _map->GetCreature(guid);
-            if (!creature || !creature->IsInWorld() || creature->IsDuringRemoveFromWorld() || !creature->FindMap() || !creature->IsAIEnabled || !creature->AI())
+            if (!creature || !creature->IsInWorld() || creature->IsDuringRemoveFromWorld() ||
+                !creature->FindMap() || !creature->IsAIEnabled || !creature->AI())
                 continue;
 
             creature->AI()->sOnGameEvent(_activate, _eventId);
--- a/src/server/game/Events/GameEventMgr.cpp
+++ b/src/server/game/Events/GameEventMgr.cpp
@@ -1875,26 +1875,61 @@ void GameEventMgr::SendWorldStateUpdate(Player* player, uint16 eventId)
 class GameEventAIHookWorker
 {
 public:
-    GameEventAIHookWorker(uint16 eventId, bool activate) : _eventId(eventId), _activate(activate) { }
+    GameEventAIHookWorker(Map* map, uint16 eventId, bool activate) : _map(map), _eventId(eventId), _activate(activate) { }
 
     void Visit(std::unordered_map<ObjectGuid, Creature*>& creatureMap)
     {
-        for (auto const& p : creatureMap)
-            if (p.second->IsInWorld() && !p.second->IsDuringRemoveFromWorld() && p.second->FindMap() && p.second->IsAIEnabled && p.second->AI())
-                p.second->AI()->sOnGameEvent(_activate, _eventId);
+        for (ObjectGuid const& guid : Snapshot(creatureMap))
+        {
+            Creature* creature = _map->GetCreature(guid);
+            if (!creature || !creature->IsInWorld() || creature->IsDuringRemoveFromWorld() || !creature->FindMap() || !creature->IsAIEnabled || !creature->AI())
+                continue;
+
+            creature->AI()->sOnGameEvent(_activate, _eventId);
+        }
     }
 
     void Visit(std::unordered_map<ObjectGuid, GameObject*>& gameObjectMap)
     {
-        for (auto const& p : gameObjectMap)
-            if (p.second->IsInWorld() && p.second->FindMap() && p.second->AI())
-                p.second->AI()->OnGameEvent(_activate, _eventId);
+        for (ObjectGuid const& guid : Snapshot(gameObjectMap))
+        {
+            GameObject* gameObject = _map->GetGameObject(guid);
+            if (!gameObject || !gameObject->IsInWorld() || !gameObject->FindMap() || !gameObject->AI())
+                continue;
+
+            gameObject->AI()->OnGameEvent(_activate, _eventId);
+        }
     }
 
     template<class T>
     void Visit(std::unordered_map<ObjectGuid, T*>&) { }
 
 private:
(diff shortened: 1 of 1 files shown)


mybaixinyu:
Thanks for the review. Fixed the line lengths in 9ce3d5d — the constructor was 122 columns and the creature guard 160; `.editorconfig` sets `max_line_length = 120`, so both are now wrapped.

I left the brace style alone on purpose:

- Line 1905 (`void Visit(std::unordered_map<ObjectGuid, T*>&) { }`) is pre-existing upstream code that this PR does not touch — reformatting it would be scope creep in a crash fix.
- The one-line `) : ... { }` constructor form is the established idiom here: 207 occurrences under `src/server/game/`, and `apps/codestyle/codestyle-cpp.py` passes on it.

Happy to change either if a maintainer prefers otherwise.

sudlud:
Thanks for the PR!

- please use the PR template
- please follow these instructions for AI generated PRs:
- https://www.azerothcore.org/wiki/agentic-self-review

mybaixinyu:
Closing this in favour of #27291, which carries the full analysis and the reproducer.

@sudlud thanks for the pointers. I updated the body to the template, but I can't complete the mandatory `/self-review` step, so rather than leave a PR sitting here that doesn't meet the bar, I've filed the investigation as an issue. The patch is still on the branch (`mybaixinyu:fix/gameevent-ai-hook-iterator`) if anyone wants to pick it up, and I'm happy to reopen as a proper PR if a maintainer would prefer that route.

PR #27385, opened 2026-08-30
fix(Core/Pets): scope the pet-save delete to its owner

3kynox:
## Changes Proposed
SavePetToDB removes the previous row with a DELETE scoped only by pet number. Pet numbers are only as unique as their number source: setups where several worldservers or a merged/imported database share the character_pet table can hold colliding ids, and the unscoped DELETE then silently destroys another owner's pet. The delete is now scoped by owner as well (new prepared statement), so saving a pet can only ever touch the saving owner's rows.

## Issues Addressed
None filed — hit this live twice on a multi-worldserver setup sharing one characters DB, both occurrences proven by binlog: a colliding pet number on another account deleted an unrelated hunter's active pet.

## Tests Performed
Built and ran on 3.3.5a; pet taming, renaming, stabling, dismissal and relog all keep working (the pair id+owner is exactly what a legitimate save matched before). With two characters holding the same pet number on purpose, the save no longer deletes the other owner's row.

## How to Test the Changes
1. Tame a pet, play normally (stable, rename, dismiss, relog) — behavior unchanged.
2. To reproduce the original defect: give two characters rows in character_pet with the same id but different owner, then save the pet of one — without this patch the other character's row is deleted.


EricksOliveira:
# Maintainer review — PR #27385 `fix(Core/Pets): scope the pet-save delete to its owner`

**Verdict** Blocks the merge — the change is a no-op for the data loss it claims to fix

**Reviewed** `98e24cf5981654b3a64fc8f871f852f079fa9cba` (`3kynox:fix/pet-save-delete-owner-scope` vs `master`, merge-base `2fed8b96e46609df69edb5c189d87947379cc71c` — 3 files, +9 −1)

**Not run** The server was not built or executed. The blocking finding is MySQL `REPLACE` semantics against `PRIMARY KEY (id)`, settled from the schema. Open for the author: the binlog they cite — they should read the whole transaction, not the isolated `DELETE`.

**By** Claude Code (desktop), Opus 5 — fresh-eyes pass by Sonnet 5 — maintainer-review v0.2, 2026-09-06

<details>
<summary>Review details</summary>

**PR** https://github.com/azerothcore/azerothcore-wotlk/pull/27385 — author `3kynox`, opened 2026-08-30, base `master`, not a draft, labels `CORE` / `file-cpp`, `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`

**Diff** `96e602e52d3c354f350dcc549eefc56610dd8550`

**Checks** All green at this head: build matrix (gcc-15, clang-18/21, pch and nopch, with and without modules), C++ codestyle, bash scripts, e2e full, CodeFactor, triage.

**Reviews on record** No human review, no approval — ever, on any head. Two bot reviews (Copilot, CodeRabbit), one unresolved inline thread. The author has not replied to anything.

**Project docs applied** `AGENTS.md`, `.agents/docs/cpp-guidelines.md`, `.agents/docs/code-review.md`, `.editorconfig`, `apps/codestyle/codestyle-cpp.py`.

## What the PR does

`Pet::SavePetToDB` opens its "remove current data" block with a `DELETE FROM character_pet WHERE id = ?`. The PR adds a new prepared statement `CHAR_DEL_CHAR_PET_BY_ID_AND_OWNER` (`... WHERE id = ? AND owner = ?`) and switches that one call site to it, binding `ownerLowGUID` as the second parameter, plus a five-line explanatory comment.

The stated motivation: several worldservers, or a merged/imported database, sharing one `character_pet` table can hold colliding pet ids, and the unscoped `DELETE` then destroys another owner's pet. The author reports two live occurrences proven by binlog.

## Findings

### 1 — blocking · `src/server/game/Entities/Pet/Pet.cpp:552`

The scoped `DELETE` cannot prevent the described data loss.

`character_pet` is `PRIMARY KEY (id)`, with `owner` only a non-unique `KEY` (`data/sql/base/db_characters/character_pet.sql:41-43`). Twenty-three lines below the delete, the same function appends `CHAR_REP_CHAR_PET` — `REPLACE INTO character_pet (id, ...)` — binding the same pet number as `id` (`src/server/game/Entities/Pet/Pet.cpp:575`). Both statements go into one transaction, unconditionally, in that order. MySQL `REPLACE` deletes whatever row conflicts on the primary key and inserts the new one; it does not consider `owner`.

So in the colliding-id case: before the patch the `DELETE` destroyed the other owner's row; after the patch the `DELETE` matches nothing and the `REPLACE` destroys the other owner's row. The resulting database state is identical. The patch changes nothing observable.

### 2 — blocking, corollary of 1 · PR description

The reproduction in "How to Test" cannot be executed on the shipped schema: "give two characters rows in `character_pet` with the same id but different owner" is impossible under `PRIMARY KEY (id)`. The table only ever holds one row per pet number — which is exactly why finding 1 holds. The data loss in a multi-worldserver setup is real, but it is structural (one row per id), not a consequence of the `DELETE`'s scope.

### 3 — non-blocking · `src/server/game/Entities/Pet/Pet.cpp:552`

The `DELETE` is redundant on every path, before and after the patch. Inside `if (mode >= PET_SAVE_AS_CURRENT)` the transaction receives the `DELETE`, then the conditional `CHAR_DEL_CHAR_PET_BY_SLOT` (already owner-scoped), then the `REPLACE`. The `REPLACE` on the same primary key already removes the pet's own previous row. The patch therefore adds a prepared statement and an extra bind for a query with no observable effect.

### 4 — non-blocking · `src/server/game/Entities/Pet/Pet.cpp:609`

Even granting the premise, the fix is partial. `Pet::DeleteFromDB` still uses the unscoped `CHAR_DEL_CHAR_PET_BY_ID`, reached from `src/server/game/Entities/Player/Player.cpp:4207` (character deletion) and `src/server/scripts/Commands/cs_pet.cpp:188`.

Beyond that, `pet_aura` (PK `guid, casterGuid, spell, effectMask`), `pet_spell` and `pet_spell_cooldown` (PK `guid, spell`) are keyed by pet number alone and carry **no** `owner` column. The deletes in `_SaveAuras` / `_SaveSpells` / `_SaveSpellCooldowns` cannot be owner-scoped without a schema change.

### 5 — context on the premise · `src/server/game/Globals/ObjectMgr.cpp:8104`

`ObjectMgr::LoadPetNumber` seeds `_hiPetNumber` once at startup from `SELECT MAX(id) FROM character_pet`; `GeneratePetNumber` (`:8150`) only increments it in memory under a mutex. `ObjectMgr::SetHighestGuids` (`:7572`) uses the same pattern for player guids, item guids, auction ids, mail ids, arena team ids, equipment set guids and guild ids. Two worldservers sharing one characters database collide on all of those, not just on pet numbers. The premise describes a deployment the core does not support anywhere, so owner-scoping one statement does not make it safe.

### 6 — nit · `src/server/database/Database/Implementation/CharacterDatabase.cpp:588`, `src/server/game/Entities/Pet/Pet.cpp:552`

`AGENTS.md` and `.editorconfig` set a 120-column limit; the two new lines are 130 and 125. Context: the untouched neighbouring `PrepareStatement` lines in the same file are 133, 138 and 297 columns, and `apps/codestyle/codestyle-cpp.py` implements no line-length check, which is why the C++ job passes. A written rule, uniformly ignored in that file.

## Comment walk

Every conversation comment, review body and inline thread on the PR, with whether it was answered and whether it still holds at head `98e24cf`.

1. **`coderabbitai[bot]`, conversation comment, 2026-08-30 13:47** — walkthrough plus "Merge Risk: Moderate": other save operations still identify pet data by pet number alone, so the PR is not merge-ready until the full save path is owner-scoped. *Never answered by the author.* Still applies, with a correction: the diagnosis is right, the proposed remedy is not implementable on the child tables (finding 4).

2. **`copilot-pull-request-reviewer[bot]`, review COMMENTED, 2026-08-30 13:41, commit `98e24cf`** — describes the three changed files and states it generated no comments. Nothing raised, nothing to answer. Not an approval.

3. **`coderabbitai[bot]`, review body, outside-diff at `Pet.cpp:530` and `:536-537`** — asks for `_SaveAuras`, `_SaveSpells` and `_SaveSpellCooldowns` to be owner-scoped too. *Never answered.* Does not apply as written: `pet_aura`, `pet_spell` and `pet_spell_cooldown` have no `owner` column, so there is nothing to scope by without a schema change.

4. **`coderabbitai[bot]`, review body, outside-diff at `Pet.cpp:590`** — `CHAR_REP_CHAR_PET` uses `REPLACE INTO`, whose primary key is `id` while `owner` is not unique, so a colliding id lets the `REPLACE` delete the other owner's row after the owner-scoped delete. *Never answered.* Confirmed, and decisive — this is finding 1, the one that sinks the PR. The bot's line anchor is off: `GetPreparedStatement(CHAR_REP_CHAR_PET)` sits at `:575`.

5. **`coderabbitai[bot]`, inline thread on `CharacterDatabase.cpp:588`, unresolved, not outdated** — the 120-column limit. *Never answered.* Still applies literally; see finding 6 for why it is a nit here.

## Not verified

The server was not built or run; CI covers compilation on six toolchain combinations and the full e2e suite passed at this head. The binlog evidence the author cites is theirs to produce — and reading the full transaction rather than the isolated `DELETE` is the thing to check, since the `REPLACE` in the same transaction destroys the row either way.

## Recommended next action

Comment on the PR with finding 1 and ask the author to reconsider the approach: if the multi-worldserver data loss is worth chasing, the problem is the identity of `character_pet.id` (and of every other in-memory id generator), not the `WHERE` clause of one `DELETE`. Nothing has been posted, approved or labelled.

</details>

3kynox:
Thanks for the thorough review. You are right on the decisive point: the `REPLACE INTO character_pet` twenty lines below removes the other owner's row on the `id` primary key whatever the `DELETE` matched, so this change cannot prevent the loss it describes — I had read the isolated `DELETE` in my binlog, not the whole transaction. Points 2 to 4 follow from that.

On point 5, the deployment is one the core supports since #16832: with `Cluster.Enabled`, several worldservers of one realm share one characters database, and the core already takes character, item and instance guids from the guid service for exactly that reason (`deps/libsidecar/include/libsidecar.h`, `ObjectGuidGeneratorBase::GetClusterGuid`). `ObjectMgr::GeneratePetNumber` was left out: it still increments a per-process counter seeded from `SELECT MAX(id) FROM character_pet` at startup. Two worldservers booted against the same database hand out the same pet numbers, so the first player to tame or summon a pet on each of them gets the same `id`, and the second save's `REPLACE` destroys the first player's pet; `pet_spell` and `pet_aura`, keyed by pet number, get mixed the same way. That is the defect I hit, and it needs nothing beyond two worldservers in cluster mode.

The fix that holds is to allocate pet numbers from the guid service in cluster mode, the same way character and item guids are: a `PetNumber` guid type on the ToCloud9 side (guid service and libsidecar export), then a call in `GeneratePetNumber` guarded by `ClusterModeEnabled()`, modeled on `GetClusterGuid`. Since the sidecar interface this repository vendors is owned by the clustering maintainer, I will take both halves through that channel rather than here. Withdrawing this change.

PR #27426, opened 2026-09-01
fix(Core/Movement): Fix pet pathing and chase angle against oversized targets (e.g. Brain of Yogg-Saron)

### LINKED ISSUE #27125: [Ulduar] Yogg Saron - Pets not able to attack the Brain, unless you command them while standing in meele range aswell
### Current Behaviour

Pets not able to attack the Brain, unless you command them while standing in meele range aswell

### Expected Behaviour

Pets should be able to attack the boss normally.

### Source

Image

Steps to get rid off Sara "Anti-Cheat"
Flame Leviathan
- `.go c id 33113`
- `.damage 200000000` on FL
XT
- `.go c id 33293`
- `.damage 200000000` on XT
Vezax
`.go c id 33271`
- `.damage 200000000` on Vezax

- be level 80 with 2 characters (one of them being a hunter)
- `.cheat god`
- `.tele Yogg`
- engage the fight and bring fight to P2 
- `.aura 22663` on both characters to not lose sanity
- take a portal to brain room with a hunter and command pet to attack the brain
- observe it not attacking, unless you start the command to attack while in meele range

### Extra Notes

_No response_

### AC rev. hash/commit

AzerothCore rev. 9ad24248be9d 2026-08-13 00:10:27 +0000 (master branch) (Unix, RelWithDebInfo, Static)

### Operating system

Windows 11

### Custom changes or Modules

AlsoNotMehh:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests
- [x] AI tools were partially used in preparing this pull request.
   - Tools/models used: Claude / Gemini

### Problem & Blizzlike Mechanics Explained:

1. **Pet Rear Chase Angle Against Oversized Targets in Enclosed Geometry:**
   In `PetAI::DoAttack`, pets chasing an enemy are assigned a chase angle `ChaseAngle(M_PI, M_PI_4)` to attack from behind to avoid parries/blocks. For oversized targets (`CombatReach > NOMINAL_MELEE_RANGE`, e.g. Brain of Yogg-Saron with `CombatReach = 30.0f` or Kologarn), trying to calculate a contact point 30 yards behind the boss places the destination outside the navmesh inside the room's solid stone walls or off-mesh into the void. When pathfinding to this point failed, `cOwner->AttackStop()` was invoked, causing the pet to immediately abort the command and stand idle.
   * *Fix:* In `ChaseMovementGenerator<T>::DoUpdate`, clear the chase angle (`angle = Optional<ChaseAngle>()`) when targeting oversized creatures (`target->GetCombatReach() > NOMINAL_MELEE_RANGE`). Oversized raid bosses in enclosed rooms or fixed positions do not require rear flanking from pets, allowing path generation to seek the target directly along the walkable navmesh.

2. **Ground Projection for Non-Flying Chasers Against Elevated Targets:**
   Brain of Yogg-Saron floats at `Z = 265.0f`, whereas the room floor is at `Z = 243.5f` (a vertical difference of ~21.5 yards). In `ChaseMovementGenerator<T>::DispatchSplineToPosition`, `owner->UpdateAllowedPositionZ(x, y, z)` was previously only called if `owner->IsHovering()`. For walking pets (`!owner->CanFly()`), `CalculatePath` was receiving the target's elevated air Z coordinate directly, causing pathfinding failure against navmesh height limits.
   * *Fix:* Extended `owner->UpdateAllowedPositionZ(x, y, z)` to run whenever `owner->IsHovering() || !owner->CanFly()`, guaranteeing the destination coordinates project down to the floor mesh.

3. **Fallback Pathing Towards Target Center with Path Shortening (`cutPath`):**
   Commit `#26726` previously introduced a fallback for Kologarn by sampling the nearest point on the outer melee ring (`target->GetNearPoint2D()`). However, for Brain of Yogg-Saron, the entire chamber radius is ~25 yards, so the 30-yard outer melee ring in almost every direction extends into or behind the chamber walls.
   * *Fix:* Added a secondary fallback: if pathing to the outer melee ring fails, path directly toward the target's ground position and set `cutPath = true`. Recast navigates cleanly through the chamber door and hallway toward the center of the room, and `i_path->ShortenPathUntilDist` automatically cuts the path along the navmesh as soon as the chaser reaches combat range.

4. **Accurate 3D Distance Measurement for Elevated Targets:**
   When `cutPath` was active, `ShortenPathUntilDist` previously measured distance against the projected 2D/ground coordinates. For elevated bosses, this resulted in stopping at horizontal reach while remaining out of 3D melee reach.
   * *Fix:* `ShortenPathUntilDist` now measures against `GetTarget()->GetPosition()` (the boss's true 3D position), ensuring pets stop at true 3D melee combat range on the floor.

## Issues Addressed:

- Closes #27125

## SOURCE:

  - Retail WotLK: Hunter and Warlock pets commanded to attack the Brain of Yogg-Saron run through the doorway/room and attack the boss normally without requiring the owner to first stand in melee range.

## Tests Performed:

- [x] Built `game` and `worldserver` cleanly on Windows Release x64 with 0 errors and 0 warnings.
- [x] Verified pet command attack behavior against oversized targets from outside melee range.
- [x] Confirmed pet pathing from inside illusion chambers through the doorway into the Brain chamber.
- [x] Confirmed pet stops cleanly on the floor within 3D melee reach and auto-attacks continuously without calling `AttackStop()`.
- [x] Confirmed standard mobs (`CombatReach <= NOMINAL_MELEE_RANGE`) retain default rear chase angle positioning.

## How to Test the Changes:

1. Create a level 80 Hunter character with a pet.
2. Teleport to Yogg-Saron: `.tele Yogg`
3. Advance the encounter to Phase 2 (or spawn Brain of Yogg-Saron: `.npc add 33890`).
4. Stand at range (e.g. at the doorway, inside an illusion chamber, or 30-40 yards away in the room).
5. Command your pet to attack the Brain of Yogg-Saron.
6. Observe that the pet runs into the room along the floor, stops within melee attack range of the floating Brain, and attacks without aborting.

None.


AlsoNotMehh:
I'm still working on this

Rorschach91:
> I'm still working on this

I'm just adding the labels.

AlsoNotMehh:
@coderabbitai review

Nyeriah:
## Maintainer review

Reviewed head 558a2cb8 against master 42104bd7. One blocking finding and two non-blocking, all inline.

1. Blocking: the angle removal at line 262 applies to every melee chase against a target with combat reach above 5, about 900 templates including Onyxia, Marrowgar, Ragnaros and Sapphiron, not just Kologarn and the Brain.
2. Non-blocking: the floor-snapped destination can sit permanently outside 3D melee reach, and the pet then runs underneath and idles instead of aborting as it did before.
3. Nit: the target coordinates are spelled out six times.

Verified on my side: the elevated-z failure the description explains is real (PathGenerator flags the 21.5 yd gap as far from poly and pets reject the incomplete path), Brain reach 30 in the world DB, summon at z 265 in the boss script, and CodeRabbit's inline point is addressed in 558a2cb. Not run in-game. Worth exercising: a melee pet on Onyxia or Marrowgar for the positioning change, a pet sent at a creature hovering well above the floor, and Kologarn again since the primary path now goes for the center before the ring retry.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -243,6 +257,13 @@ bool ChaseMovementGenerator<T>::DoUpdate(T* owner, uint32 time_diff)
         mutualChase = true;
     }
 
+    // For oversized targets (e.g. Kologarn, Brain of Yogg-Saron), chasing behind the target
+    // in melee is invalid because the rear is off-mesh or inside geometry.
+    if (angle && (!_range || _range->MaxRange <= CONTACT_DISTANCE) && target->GetCombatReach() > NOMINAL_MELEE_RANGE)
```
Nyeriah: This drops the rear chase angle for every melee chase whose target has a combat reach above 5, not just Kologarn and the Brain. The world DB has 904 templates over that line, among them Onyxia (18), Marrowgar (20.2), Ragnaros (20), Sapphiron (22) and Malygos (40). Pets will now hit all of them from the front, where they get parried and blocked. #26726 kept the angle on purpose and listed "pet still takes its usual behind-the-target position on Onyxia" as a regression check, and the comment above claims the rear is off-mesh for all of these, which isn't true for open rooms.

I get why it's needed: PositionOkay rechecks the angle every 400 ms, so a pet placed by the ring or center fallback would repath forever. But that only applies when the fallback actually kicked in. Could you drop the angle only in that case, e.g. remember in the generator that the angled and ring destinations failed for this target, and use that both here and in the PositionOkay recheck? Otherwise this needs a source that retail pets don't flank big bosses, and the description should say the behaviour changes for all of them.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -137,6 +137,20 @@ bool ChaseMovementGenerator<T>::DispatchSplineToPosition(T* owner, float x, floa
         pathFailed = !i_path->CalculatePath(x, y, z, forceDest) || !isPathUsable();
         // the destination already lies on the melee ring, nothing to cut
         cutPath = false;
+
+        // If the nearest point on the melee ring is also unpathable (e.g. Brain of Yogg-Saron
+        // where the 30yd combat reach extends beyond the room's walls into geometry, but the
+        // floor directly underneath the target is walkable), fall back to pathing directly
+        // toward the target's ground position and cut the path once within combat range.
+        if (pathFailed)
+        {
```
Nyeriah, line 142: Snapping z to the floor can produce a destination the chase can never finish from. Melee range is 3D, so if the target sits higher above the floor than maxTarget, the snapped point is never in reach. CalculatePath still returns a normal path, SetCannotReachTarget() gets cleared below, and ShortenPathUntilDist finds nothing to cut. Before this change that case came back as an incomplete path and the pet aborted (the behaviour from #24330). Now it runs underneath and idles in attack state. Same thing for a Kologarn-shaped target if the ring retry ever fails: the pet paths down into the pit.

The Brain is fine since 21.5 yd is inside its ~32 yd maxTarget. A cheap guard for the fallback would be to keep pathFailed = true when std::abs(z - GetTarget()->GetPositionZ()) >= maxTarget. Either way, please test a melee pet against a creature hovering well above the floor, and Kologarn again since its primary path now goes for the center before the ring retry.

src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp
```diff
@@ -154,7 +168,7 @@ bool ChaseMovementGenerator<T>::DispatchSplineToPosition(T* owner, float x, floa
     }
 
     if (cutPath)
-        i_path->ShortenPathUntilDist(G3D::Vector3(x, y, z), maxTarget);
+        i_path->ShortenPathUntilDist(G3D::Vector3(GetTarget()->GetPositionX(), GetTarget()->GetPositionY(), GetTarget()->GetPositionZ()), maxTarget);
```
Nyeriah: Nit: the target's x/y/z are spelled out six times between here and the fallback above. GetTarget()->GetPosition(x, y, z) is already used further down in this file, or keep a G3D::Vector3 targetPos once and reuse it in both places.

Nyeriah (changes requested):

AlsoNotMehh:
### 1. Maintainer Review Findings & Resolutions

* **Finding 1 (Blocking): Unconditional angle removal on oversized targets (`CombatReach > 5.0f`)**
  * **Original Issue:** In the previous revision, clearing `angle` inside `DoUpdate` based solely on `target->GetCombatReach() > NOMINAL_MELEE_RANGE` stripped rear chase angles for over 900 creatures (including Onyxia, Marrowgar, Sapphiron, and Ragnaros), causing pets to attack from the front where they get parried and blocked.
  * **Fix:** Introduced `_fallbackPositioning` tracking in `ChaseMovementGenerator<T>`. Normal open-room targets (like Onyxia) continue to receive the primary angled destination behind the boss. The angle constraint is dropped only when an oversized target fallback (melee ring retry or center path) is actually required and succeeds. When the target moves, `_fallbackPositioning` is immediately reset to `false` to attempt the preferred rear angle first at the new position.

* **Finding 2 (Non-blocking): Ground projection on elevated targets breaking unreachable target aborts (#24330)**
  * **Original Issue:** If a creature is hovering high in the air above melee reach, unconditionally projecting Z to the floor produces a walkable destination directly underneath the target. A pet would path there, stop, and idle in attack state instead of aborting.
  * **Fix:**
    * Restored the top-level `if (owner->IsHovering())` check to match upstream master, preventing ground projection for normal paths against flying creatures.
    * Guarded both oversized fallbacks (ring retry and center path) with `std::abs(groundZ - targetPos.z) < maxTarget`. If the target sits higher above the floor than 3D melee reach, the fallbacks do not execute, leaving `pathFailed = true`. The pet cleanly triggers `SetCannotReachTarget` and `AttackStop()`.

* **Finding 3 (Nit): Repeated target coordinate lookups**
  * **Original Issue:** `GetTarget()->GetPositionX()`, `GetPositionY()`, and `GetPositionZ()` were queried multiple times throughout `DispatchSplineToPosition`.
  * **Fix:** Cached `G3D::Vector3 const targetPos = GetTarget()->GetPosition();` once at the entry of `DispatchSplineToPosition` and reused it across fallback calculations and `ShortenPathUntilDist`.

### 2. Scenario Behavior Breakdown

* **Brain of Yogg-Saron (`CombatReach = 30.0f`, floating at Z = 265.0f, floor at Z = 243.5f):**
  * Primary path to rear angle at Z = 265 fails (off-mesh inside geometry).
  * Fallback 1 (melee ring retry): 30yd ring sits inside chamber walls, fails.
  * Fallback 2 (center ground path): `std::abs(243.5 - 265.0) = 21.5 < maxTarget (30.5)`. Ground position inside the chamber is walkable. Path is created and shortened to true 3D melee reach via `ShortenPathUntilDist(targetPos, maxTarget)`.
  * `_fallbackPositioning` is set to `true`, preventing angle-recheck thrashing in `PositionOkay`. Pet attacks continuously from the chamber floor.

* **Onyxia / Marrowgar / Open-room bosses on the ground:**
  * Primary path calculates directly to the rear position.
  * `pathFailed` is `false`, `usedFallback` is `false`, `_fallbackPositioning` remains `false`.
  * Pets and melee units flank behind the target to avoid parry and block.

* **Kologarn (unwalkable void behind boss, platform in front):**
  * Primary path to rear angle fails (hangs over the pit).
  * Fallback 1 (melee ring retry): Finds the nearest point on the front ring on the walkable bridge (`std::abs(z - targetPos.z) < maxTarget`), succeeds.
  * `_fallbackPositioning` becomes `true`. Pet attacks from the front ring without falling into the void or endless repathing.

* **High-flying creatures (drakes or flying mounts > maxTarget in the air):**
  * Primary path fails (no poly in air).
  * Fallback 1 & 2 fail or skip due to `std::abs(z - targetPos.z) >= maxTarget`.
  * `pathFailed` remains `true`. Pet cleanly stops moving and executes `AttackStop()`, preserving PR #24330.

Nyeriah:
All three points from the last round are fixed, and the new height guard clears both shapes it has to: Kologarn's ring point sits on the walkway at essentially his own z, and the Brain's 21.5 yard drop fits inside its ~32 yard maxTarget.

Four things left, none blocking.

[TargetedMovementGenerator.cpp:141](https://github.com/azerothcore/azerothcore-wotlk/blob/ebd6751296baa4b63ead9d4f664ac93ed18ca644/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L141) and [:159](https://github.com/azerothcore/azerothcore-wotlk/blob/ebd6751296baa4b63ead9d4f664ac93ed18ca644/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L159) — `DistanceYourself` calls `DispatchSplineToPosition` with `maxTarget = 0.f`, so both new guards compare an absolute value against zero and can never pass. The oversized ring retry that master runs on that path is dead. It takes a SmartAI creature distancing from a non-player target with reach above 5 and an unpathable away point, so it is narrow, but it is a silent behaviour change. `maxTarget` means melee tolerance to one caller and nothing to the other: either skip the height check when `maxTarget <= 0.f`, or give the fallbacks their own tolerance.

[:342](https://github.com/azerothcore/azerothcore-wotlk/blob/ebd6751296baa4b63ead9d4f664ac93ed18ca644/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L342) — rebuilding `angle` from `mutualChase` and `_angle` throws away whatever the two anti-spinning guards above it did. It is correct today only because both of them also set `mutualChase = true`. Leaving line 262 as master had it and applying `_fallbackPositioning` only at the `PositionOkay` call in the periodic recheck gets the same behaviour without the re-derivation.

[:190](https://github.com/azerothcore/azerothcore-wotlk/blob/ebd6751296baa4b63ead9d4f664ac93ed18ca644/src/server/game/Movement/MovementGenerators/TargetedMovementGenerator.cpp#L190) — switching the shorten reference to `targetPos` also changes the ordinary cut-path case for a hovering owner, whose `z` was rewritten by the projection at line 107 before it got measured against. More correct, but it is a change outside the oversized path and it is not written down anywhere.

Description — point 2 still says `UpdateAllowedPositionZ` was extended to `IsHovering() || !owner->CanFly()`. That is gone, line 107 matches master. The `_fallbackPositioning` tracking and the two height guards are the substance of this revision and currently only exist in a comment, and the tests list still describes the old design. Worth updating before this lands, it is what ends up in the merge record.

The e2e red is two flaky tests in unrelated suites, Stratholme Timmy and a trade seed precondition. `combat/pets` and the Ulduar suite both pass.

Nyeriah (approved):

Issue #27291, opened 2026-08-24
Crash: GameEventMgr::RunSmartAIScripts — SmartAI respawn erases from the map object store mid-iteration


mybaixinyu:
### Steps to Reproduce

Deterministic, not intermittent — but it needs one precondition that is rare on a normal realm.

Game event 89 (`Leprithus`) runs daily at 20:00 for 600 minutes. Entry 572 carries `smart_scripts` id 3: `event_type 68` (`SMART_EVENT_GAME_EVENT_START`, param 89) → `action_type 70` (`SMART_ACTION_RESPAWN_TARGET`, target self). `Creature::Respawn()` only reaches the erase when the creature is dead, and Leprithus is a rare spawn in Duskwood on a 20 hour timer, so it is almost always alive when the event fires.

1. Kill Leprithus (Duskwood, entry 572, spawn 134021) and keep the grid loaded
2. `.event start 89`

worldserver dies with `SIGSEGV` in `GameEventMgr::RunSmartAIScripts`.

Two other events ship the same pattern but effectively never fire (`occurence` ~10 years): event 85 (`Stitches`, 13 creatures) and event 87 (`Scarlet Oracle`, 2 creatures).

### What is happening

`GameEventAIHookWorker::Visit` (`src/server/game/Events/GameEventMgr.cpp`) walks `Map::GetObjectsStore()` with a range-for and calls into creature AI while still holding the iterator:

```cpp
void Visit(std::unordered_map<ObjectGuid, Creature*>& creatureMap)
{
    for (auto const& p : creatureMap)
        if (p.second->IsInWorld() && !p.second->IsDuringRemoveFromWorld() && p.second->FindMap() && p.second->IsAIEnabled && p.second->AI())
            p.second->AI()->sOnGameEvent(_activate, _eventId);
}
```

The AI call erases its own creature from that map:

```
GameEventAIHookWorker::Visit -> SmartScript::ProcessEventsFor -> ProcessAction
  -> Creature::Respawn -> Map::AddObjectToRemoveList
  -> Unit::CleanupBeforeRemoveFromMap -> Creature::RemoveFromWorld
  -> _objectsStore.Remove<Creature>()        // node freed
```

[...]

PR #27429, opened 2026-09-01
fix(Core/Events): Snapshot map before AI hooks

nickk02:
## Changes Proposed:
This PR proposes changes to:

### AI-assisted Pull Requests

   - Tools/models used: Claude Sonnet 5 (Claude Code), for investigation, the fix itself, and this description. I reviewed and understand the change; I'm the one answering questions on it.

`GameEventAIHookWorker::Visit` iterates `creatureMap`/`gameObjectMap` by reference while dispatching `sOnGameEvent()`/`OnGameEvent()` callbacks. Those callbacks run arbitrary SmartAI actions. A same-map `SMART_ACTION_SUMMON_CREATURE` reaches `Creature::AddToWorld()`, which calls `GetMap()->GetObjectsStore().Insert<Creature>(...)` synchronously — inserting into the very `std::unordered_map` the range-based `for` is iterating. An insert can trigger a rehash, which invalidates every iterator into the map, including the one the loop holds. Same shape on the gameobject overload.

The fix copies the raw pointers into a local `std::vector` before dispatching, and iterates the copy instead. This is safe against use-after-free, not just iterator invalidation: object destruction in AzerothCore is deferred to `Map::RemoveAllObjectsInRemoveList()`, called only from `Map::DelayedUpdate()` — a different phase from the `World::Update() -> GameEventMgr::Update() -> RunSmartAIScripts()` path this code runs on. Nothing is freed mid-sweep, so the copied pointers can't dangle, and the existing `IsInWorld()`/`IsDuringRemoveFromWorld()`/`FindMap()` guards remain sufficient for anything logically removed during the sweep.

## Issues Addressed:
No open issue in this repo. Related, on TrinityCore (shared ancestor for this file): **TrinityCore#26687, "Crash GameEventMgr::RunSmartAIScripts"** — open since 2021-07-13, directly on point, cited here as evidence this class of defect is real and long-standing, not as something this PR closes there.

## SOURCE:

None of these quite fit — this isn't a content-accuracy fix, it's an engine memory-safety fix, reached by direct code reading (checked how object insertion and object removal both interact with the container this code iterates) rather than an external source. Root cause write-up: https://github.com/nickk02/azerothcore-deploy (private repo; happy to share specific excerpts if useful).

## Tests Performed:

More precisely than the checkboxes above allow: this fix has been running in production on two of my own self-hosted realms since 2026-08-01 (upstream tree) and 2026-08-07 (a Playerbots-fork build with the identical patch applied to both `Visit` overloads), with the compiled object verified by DWARF line info to confirm the new snapshot code is actually in the binary, not just present in source. Since deploying, the realm that runs 40 AI-controlled bots 24/7 (the traffic pattern most likely to hit this) has crossed roughly 7-8 daily event-boundary windows with zero SIGSEGVs, versus a prior history of a same-shaped crash recurring at that exact boundary.

That's meaningful field evidence the fix doesn't break anything and correlates with the crash disappearing, but I want to be precise about what it isn't: I never captured a backtrace with a rehash on this exact container as the confirmed trigger of the original crash. The hazard itself is real and provable by code inspection independent of that — this PR closes a genuine iterator-invalidation bug either way — but I can't claim a smoking-gun backtrace tying it to a specific prior incident, only that a real memory-safety hazard existed here and no longer does.

## How to Test the Changes:

1. Set up a realm with Playerbots or another source of sustained SmartAI activity around a game event (WorldState-driven events work too), where at least one participating creature's SmartAI can trigger `SMART_ACTION_SUMMON_CREATURE` onto the same map during the event's activation/deactivation hook.
2. Before this patch: under load, watch for a crash inside `GameEventAIHookWorker::Visit` / `RunSmartAIScripts` with a backtrace touching `std::unordered_map` rehash internals. This is timing-dependent and doesn't reproduce on demand — that's exactly why it's a real iterator-invalidation bug and not something a deterministic unit test can catch.
3. After this patch: the same activity should run without that crash. Also worth exercising a large game event with many participating creatures/gameobjects, to confirm normal event AI callbacks (not just the summon-during-event edge case) still fire during activation and deactivation, since the loop body is otherwise unchanged.

- [ ] No automated test covers this — the hazard is timing-dependent (whether a rehash happens to occur during the specific loop iteration) and not something a deterministic unit test reproduces cleanly. Open to suggestions if a maintainer has a pattern for this class of bug elsewhere in the codebase.
- [ ] Would appreciate a second look from someone who knows this subsystem better, particularly on whether any other code assumes `GameEventAIHookWorker::Visit` iterates the live container rather than a snapshot (I didn't find one, but I don't know this file's full history).

### DIFF
--- a/src/server/game/Events/GameEventMgr.cpp
+++ b/src/server/game/Events/GameEventMgr.cpp
@@ -1879,16 +1879,48 @@ public:
 
     void Visit(std::unordered_map<ObjectGuid, Creature*>& creatureMap)
     {
+        // Snapshot the pointers before dispatching anything.
+        //
+        // sOnGameEvent() runs arbitrary SmartAI actions for whichever creature
+        // the callback lands on. An action such as a same-map
+        // SMART_ACTION_SUMMON_CREATURE reaches Creature::AddToWorld(), which
+        // does GetMap()->GetObjectsStore().Insert<Creature>(...) synchronously
+        // -- inserting into this very container. Inserting into an
+        // std::unordered_map may rehash it, and a rehash invalidates every
+        // iterator into it, including the one a range-based for over
+        // creatureMap would still be holding. That is undefined behaviour and
+        // the observed crash signature.
+        //
+        // Copying the raw pointers is safe against use-after-free: object
+        // destruction is deferred to Map::RemoveAllObjectsInRemoveList(),
+        // which is called from Map::DelayedUpdate(), a different phase from
+        // the World::Update() -> GameEventMgr::Update() -> RunSmartAIScripts()
+        // path this runs on. Nothing is freed mid-sweep, so the per-element
+        // guards below remain sufficient for objects that were logically
+        // removed during the sweep.
+        std::vector<Creature*> creatures;
+        creatures.reserve(creatureMap.size());
         for (auto const& p : creatureMap)
-            if (p.second->IsInWorld() && !p.second->IsDuringRemoveFromWorld() && p.second->FindMap() && p.second->IsAIEnabled && p.second->AI())
-                p.second->AI()->sOnGameEvent(_activate, _eventId);
+            creatures.push_back(p.second);
+
+        for (Creature* creature : creatures)
+            if (creature->IsInWorld() && !creature->IsDuringRemoveFromWorld() && creature->FindMap() && creature->IsAIEnabled && creature->AI())
+                creature->AI()->sOnGameEvent(_activate, _eventId);
     }
 
     void Visit(std::unordered_map<ObjectGuid, GameObject*>& gameObjectMap)
     {
+        // Same hazard and same reasoning as the creature overload above:
+        // GameObjectAI::OnGameEvent() can reach GameObject::AddToWorld(), which
+        // inserts into this container synchronously.
+        std::vector<GameObject*> gameObjects;
+        gameObjects.reserve(gameObjectMap.size());
         for (auto const& p : gameObjectMap)
-            if (p.second->IsInWorld() && p.second->FindMap() && p.second->AI())
-                p.second->AI()->OnGameEvent(_activate, _eventId);
+            gameObjects.push_back(p.second);
+
+        for (GameObject* gameObject : gameObjects)
+            if (gameObject->IsInWorld() && gameObject->FindMap() && gameObject->AI())
+                gameObject->AI()->OnGameEvent(_activate, _eventId);
     }
 
     template<class T>


nickk02:
## Self-review

I don't have this repo's `/self-review` tooling available in my environment, so this is a manual pass covering the same ground the self-review rules ask for, rather than a generated `.SELF-REVIEW.md`.

**Regression risk.** `GameEventAIHookWorker` has exactly one call site in the whole tree: instantiated and visited once inside `GameEventMgr::RunSmartAIScripts` (called from the activate and deactivate paths, both in this same file). Nothing else references the class. The loop body itself is unchanged — same guards, same order, same callbacks — only the container being iterated changed from the live map to a snapshot copy of its pointers. I don't see anything else that could regress from this.

**What I actually tested, precisely.** Deployed to production on two of my own realms (2026-08-01 upstream tree, 2026-08-07 a Playerbots-fork build with the identical patch). Confirmed by DWARF line info that the new code is really in the compiled binary, not just the source. Since deploying, the bot-heavy realm has crossed about 7-8 of the daily event-boundary windows that previously reproduced a same-shaped crash, with zero recurrence. I did not capture a backtrace proving a rehash on this exact container caused that specific prior crash — the hazard is real and provable by code inspection on its own, but I want to be clear I'm not claiming a smoking-gun trace tying it to a named incident.

Happy to answer anything about the change.

EricksOliveira:
Change itself looks right to me, I traced the same path (`AddToWorld` -> `ObjectsStore::Insert<Creature>`) and the insert really does land in the container the loop is iterating.

Two things, both non-blocking.

The comment block is pretty long for a three line change. Can we cut it to a line or two on the invariant, something like "AI callbacks can insert into this container, so iterate a copy"? The call chain walkthrough and the crash signature part read well in the PR description, it just feels like a lot to carry in the file.

Maybe we should snapshot `ObjectGuid`s instead of raw pointers, and re-resolve with `map->GetCreature(guid)` / `map->GetGameObject(guid)` inside the loop? The cpp guidelines ask for that for anything held past the current call, and it means the fix stops depending on destruction staying deferred to `RemoveAllObjectsInRemoveList()`. It costs a hash lookup per object, which is nothing next to running the AI. Would need the `Map*` passed into the worker, `RunSmartAIScripts` already has it. wdyt?

One more for the description: with the snapshot, creatures or gameobjects spawned during the sweep no longer get the hook. Deterministic now instead of whatever the old loop happened to do, but it's still a behaviour change.

sudlud:
There's an open issue for this: #27291, same function, same defect, with a deterministic repro and an ASan run on plain master (in the comments, not the body). Can you link it so this closes it?

Worth reading, because it changes the comment block. What it proves isn't the rehash, it's a synchronous erase:

```
Creature::Respawn()                     Creature.cpp:2140
  -> Map::AddObjectToRemoveList()       Map.cpp:1793    calls obj->CleanupsBeforeDelete(false)
  -> Unit::CleanupBeforeRemoveFromMap   Unit.cpp:12580  calls RemoveFromWorld()
  -> Creature::RemoveFromWorld()        Creature.cpp:357
  -> GetObjectsStore().Remove<Creature>()
```

The delete is deferred, the store erase isn't, so the hash node the loop stands on gets freed. That's what "Nothing is freed mid-sweep" reads past. Can we document that path instead? Shipped data reaches it via entry 572 (Leprithus, event 89): `smart_scripts` id 3 is `SMART_EVENT_GAME_EVENT_START` -> `SMART_ACTION_RESPAWN_TARGET` on self.

The fix itself is fine, it covers both directions. @EricksOliveira is right that the block is long for a three line change, so trimming and correcting it can be one edit.

Same block: "the observed crash signature" doesn't match the description, where you say you never captured a backtrace. Maybe drop that clause.

The summon path isn't reachable from shipped data, no `smart_scripts` row pairs event_type 68/69 with action 12 or 50. It is reachable indirectly, `Creature::Respawn()` in compat mode calls `sPoolMgr->UpdatePool<Creature>` and pool spawn goes on to `AddToWorld()`. Worth a mention if you keep the insert path in.

TrinityCore#26687 I don't think is on point: the logging there only ever caught a null `AI()` on a transport (Orgrim's Hammer, empty AIName), and our loop already guards `p.second->AI()`. #27291 is the one that matches.

Seconding @EricksOliveira on the guids, we usually snapshot `ObjectGuid`s and re-resolve via `map->GetCreature(guid)` / `map->GetGameObject(guid)`. Raw pointers do work here, I traced it, every delete goes through `Map::RemoveAllObjectsInRemoveList()` which can't run inside this call, but the guid form is how we do it elsewhere and doesn't lean on that. #27275 has the shape, and if you end up picking anything up from it, add a trailer in a follow-up commit:

```
Co-authored-by: Bai Xinyu <mybaixinyu@live.com>
```

His third point deserves a line in the description: with the snapshot, objects added during the sweep don't get the hook. Fine by me, the old behaviour was undefined, but it's a change.

Could you also run the #27291 repro against this patch? The A/B on record was against #27275. Two things to check, no ASan report, and Leprithus still respawning so we know the hook ran rather than got skipped.

nickk02:
Closing this in favour of #27275, which is the same fix, filed first and proved properly.

@mybaixinyu opened #27291 on 24 August with the correct mechanism, and #27275 with the guid-snapshot patch. I opened this PR on 1 September without finding either one. That is my miss. I checked whether master still carried the vulnerable loop. I did not search the issue tracker.

Two corrections to what I wrote here.

**The comment block described the wrong mechanism.** I wrote it as a rehash on insert, with a summon during the hook inserting into the map being walked. @sudlud is right, and #27291 already proved it: the reachable path is a synchronous erase. `Creature::Respawn()` calls `Map::AddObjectToRemoveList()`, which reaches `Creature::RemoveFromWorld()` and erases from the object store on this same call stack. The delete is deferred. The store erase is not. My line "nothing is freed mid-sweep" is wrong about the hash node, which is the thing that actually gets freed.

**The TrinityCore#26687 citation does not apply.** Both #27291 and #27275 explain why. The logging added in that report only ever caught a null `AI()` on a transport, and this loop already guards `AI()`. #27291 is the matching report.

#27275 is also the better patch. It snapshots `ObjectGuid`s and re-resolves through the store, per `.agents/docs/cpp-guidelines.md`, so it does not depend on deletion staying deferred. My raw-pointer version does prevent the crash, as @sudlud traced, but it leans on that guarantee for nothing in return.

@sudlud @EricksOliveira: can #27275 be reopened? @mybaixinyu closed it only because they could not complete the `/self-review` step. That is a process blocker, not a technical one.

Two things I can add to that PR:

1. Field evidence. I have run an equivalent fix in production since 1 August on two realms, one with 40 Playerbots active continuously, which is the traffic pattern most likely to reach this loop. Same patch in both `Visit` overloads, on the upstream tree and on a mod-playerbots build. Nine days on the current build with zero coredumps. A glibc upgrade forced a restart this morning and both worldservers stopped and started clean.
2. The self-review. I can write it against #27275's patch and post it there, if a maintainer confirms that a review from someone other than the author satisfies that step.

If you would rather this PR carry the fix, say so and I will reopen it. I will take #27275's patch verbatim with a `Co-authored-by: Bai Xinyu <mybaixinyu@live.com>` trailer, correct the comment block, and post the self-review here.

PR #27443, opened 2026-09-02
chore(CI): skip full builds for docs/labeler-only changes

TheSCREWEDSoftware:
An example prior to this PR: https://github.com/azerothcore/azerothcore-wotlk/actions/runs/33690473318/job/100447826879?pr=27439

A change of a labeler it's running workflows that shouldn't be running there's no core changes all changes there were exclusively made for github.

From Claude:
This applies to three workflows: Dashboard CI, nopch-build, and pch-build — the ones that actually compile and test the server. Each gets a new `changes` job up front that checks which files a PR touched (via the GitHub API, no checkout needed, so it finishes in seconds). If every changed file is a `.md`, under `docs/**`, `.github/ISSUE_TEMPLATE/**`, `.github/FUNDING.yml`, or `.github/labeler.yml`, the actual build/test/e2e jobs in that workflow are skipped instead of running a full compile. Anything that touches real code, SQL, CMake, workflow files, etc. still runs exactly as before. The workflows still always trigger and still always post a status, so a required check can't get stuck waiting on something that never ran.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

Claude Antropic Sonnet 5

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/.github/workflows/core-build-nopch.yml
+++ b/.github/workflows/core-build-nopch.yml
@@ -35,11 +35,38 @@ permissions:
   contents: read
 
 jobs:
+  changes:
+    name: Detect relevant changes
+    runs-on: ubuntu-latest
+    if: github.repository == 'azerothcore/azerothcore-wotlk'
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant }}
+    steps:
+      # token mode: reads the changed-file list from the API, no checkout needed,
+      # so a PR that only touches docs/labeler config gets a fast, cheap "no" here
+      # instead of paying for a full build just to find out it wasn't needed.
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          filters: |
+            relevant:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+              - '!.github/FUNDING.yml'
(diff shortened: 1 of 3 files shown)


sudlud:
`!docs/**` doesn't match anything, the dir is `doc/`. maybe just remove the line, `!**/*.md` already covers everything in there except `changelog/pendings/create.sh`.

`dorny/paths-filter@v3` runs on node 20 and the job log already warns it's deprecated. can we go to v4? it also has `predicate-quantifier: some-with-excludes`, which is exactly what the `'**'` + `every` combo is doing here.

the `changes` job is the same ~30 lines in all three files and the exclusion list has to stay in sync by hand. maybe extract it into `.github/actions/` like `linux-build`?

just wondering about the gate job in general: `paths-ignore` on the trigger has the same semantics (skip when all changed files match) without the extra job or the third party action, and we already use workflow level `paths` in `codestyle.yml` and `core_modules_build.yml`. the description says we need the job so a required check can't get stuck, but master has no required status checks at the moment, the only ruleset is deletion/non-fast-forward/copilot review. what we'd lose is the "skipped" check showing up on the PR page. wdyt?

`needs.changes.outputs.relevant == 'true'` on the `e2e` job is redundant, `build-e2e` can't come back successful unless it's already true.

sudlud:
please address the review comment

PR #27018, opened 2026-08-07
feat(Core/Players): Add inventory arrival hook

Fuitad:
## Changes Proposed:

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

It adds a generic `PlayerScript` callback after an item has finished moving
into inventory. The dispatcher receives the surviving destination item, which
keeps the callback valid when the incoming item is consumed by a stack merge.

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: OpenAI Codex, GPT-5
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:

- Closes #27017

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

This is a new generic scripting API and is not based on external gameplay
research or cherry-picked code.

## Tests Performed:

This PR has been:
- [ ] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [x] This pull request requires further testing and may have edge cases to be tested.

Automated validation was performed on macOS:

- `cmake --build build-transmog-plus-hook --target unit_tests --parallel 4`
- `build-transmog-plus-hook/src/test/unit_tests --gtest_filter=PlayerInventoryHookTest.*` passed 1 test.
- The complete `unit_tests` executable passed 11,395 tests from 84 suites.
- `python3 apps/codestyle/codestyle-cpp.py` passed all C++ code style checks.

## How to Test the Changes:

[...]

PR #27447, opened 2026-09-03
feat(Core/Scripting): add OnPlayerAfterTakeItemFromMail player hook

r-o-b-o-t-o:
## Changes Proposed:

This PR proposes changes to:

Adds a `PlayerScript` hook that fires after a player takes an item out of a mailbox:

- `PLAYERHOOK_ON_AFTER_TAKE_ITEM_FROM_MAIL`
- `PlayerScript::OnPlayerAfterTakeItemFromMail(Player* player, Item* item, uint32 count)`

**Why:** ~~there is currently no way for a script to observe mail collection. `WorldSession::HandleMailTakeItem` calls `Player::MoveItemToInventory`, which calls `Player::StoreItem`, and none of the three invoke `sScriptMgr`. The nearby hooks do not cover it either:~~

**Update**: 2d0b8818bc93b6b9c12cc65e976cc1f303edb2bf adds an `OnPlayerAfterMoveItemToInventory` hook, called in `Player::MoveItemToInventory`. The new hook can be used to detect items taken from the mailbox, however it also fires for other sources of new items (trading, guild bank withdrawal). As such the `OnPlayerAfterTakeItemFromMail` hook from this PR still has its use due to its restricted scope.

The motivating case is a module that mails a character an upgraded item and wants to clean up the superseded one at the moment it is actually collected, rather than polling on a timer.

### `Player::MoveItemToInventory` now returns the stored item

`void` → `Item*`, returning the `pLastItem` it already computed from `StoreItem` and previously discarded.

This is required for the hook to be safe. Passing `HandleMailTakeItem`'s own `it` would be a use-after-free: on a full stack merge `Player::_StoreItem` marks that source item `ITEM_REMOVED` and returns the destination stack instead, and the `SaveInventoryAndGoldToDB` call that precedes the hook reaches `Player::_SaveInventory` → `Item::SaveToDB`, which does `delete this` for `ITEM_REMOVED`. A subscriber would have received freed memory.

`pLastItem` is either the original item (stored fresh, `ITEM_NEW`) or the merged-into stack (`ITEM_CHANGED`) — never `ITEM_REMOVED` — so it survives the save, and it is also the more useful value: the stack the player actually ends up holding.

Changing the return type from `void` is source-compatible. All nine existing callers (`Guild.cpp`, `MailHandler.cpp`, `TradeHandler.cpp` ×6, `TC9GrpcHandler.cpp`) ignore the return value and are unmodified.

### Other notes for reviewers

**Enum placement.** The new entry is appended immediately before `PLAYERHOOK_END`, so no existing hook id shifts for any other script.

**Call placement.** The call sits *after* `SendMailResult`, not before the transaction commit. Destroying the collected item is a plausible thing for a consumer to do, and doing that while the client is still awaiting its "item taken" acknowledgement invites a phantom item in the bag UI. Everything the core owes the client is done by that point.

### AI-assisted Pull Requests

   - Tools/models used: **Claude Code (Claude Opus 5)**

## Issues Addressed:

None — this adds a missing extension point rather than fixing a reported bug.

## SOURCE:

Not applicable: this is a scripting extension point, not a change to emulated game behaviour, so there is nothing to validate against live/sniff/Wowhead sources. The claims about existing behaviour in the description were verified by reading the current `master` sources cited inline (`HandleMailTakeItem`, `MoveItemToInventory`, `_StoreItem`, `_SaveInventory`, `Item::SaveToDB`).

## Tests Performed:

This PR has been:

Build and static checks:

- Full Debug build on Windows / MSVC is green (`worldserver.exe` links, zero compiler errors), with every `MoveItemToInventory` caller recompiled after the signature change.
- `python apps/codestyle/codestyle-cpp.py` passes all nine checks.

In-game on a local 3.3.5a realm, with a temporary `PlayerScript` subscribed to the hook. For every item taken it reads `GetEntry()`, `GetCount()` and `GetGUID()` through the `Item*`, then looks that GUID up with `player->GetItemByGuid(...)`. That lookup walks only the character's own slots, so it succeeds **only if** the hook really passed the stack the character ends up holding — the mail's own copy sits in no slot, and on the merge path has already been freed.

| # | Case | Result |
|---|------|--------|
| 1 | Fresh stack — mail 5 Linen Cloth to a character holding none | PASS · took 5, stack 5, held 5, fresh stack |
| 2 | **Partial merge — hold 3 Linen Cloth, mail 5 more** | **PASS · took 5, stack 8, held 8, merged into existing stack** |
| 3 | Merge into a full stack — hold 20, mail 5 more | PASS · took 5, stack 5, held 25, fresh stack (correctly did not merge) |
| 4 | Non-stackable item | PASS · took 1, stack 1, fresh stack |
| 5 | Three different items in one mail | PASS ×3, one per item taken, correct entry each time |

**Case 2 is the one that matters.** It is the path where `_StoreItem` marks the mail's item `ITEM_REMOVED` and `Item::SaveToDB` deletes it during `SaveInventoryAndGoldToDB`. With the fix the pointer resolves to the character's own 8-stack; on the first commit alone it would resolve to nothing, or crash the worldserver.

Regressions, all as expected:

- **Money-only mail:** hook does not fire, money credited.
- **COD mail:** hook fires once with the correct item, money deducted from the taker, sender paid.
- **Full bags:** hook does not fire, client shows "Inventory is full", item stays in the mailbox.
- **Return to sender / delete mail:** hook does not fire.

## How to Test the Changes:

1. Build a module or script that subclasses `PlayerScript`, enables `PLAYERHOOK_ON_AFTER_TAKE_ITEM_FROM_MAIL`, and overrides `OnPlayerAfterTakeItemFromMail` to log `player->GetName()`, `item->GetEntry()` and `count`.
2. Mail a character an item it is **not** already carrying, take it from the mailbox, and confirm the hook fires once with the correct entry and count.
3. Repeat with a **stackable** item the character already has a partial stack of — e.g. mail 5 Linen Cloth to a character holding 3. Take it from the mailbox and confirm the hook still fires with a valid `Item*`, and that the entry logged is correct. This is the merge path: before this PR's second commit the pointer here would have been freed memory.
4. Confirm no regression in ordinary mail handling: taking items, taking money, and COD mail all behave as before.

- [ ] Only `HandleMailTakeItem` fires the hook. Items that reach a character by other means are unaffected, by design.


Nyeriah (changes requested):
The "no way for a script to observe mail collection" rationale no longer holds. #27018 landed on master today (2d0b8818, your merge base) and `MoveItemToInventory` now fires `OnPlayerAfterMoveItemToInventory` with the same `pLastItem` you return here, so that hook already runs on every mail take.

As I read it, what this hook still adds over the generic one is:

- it fires after `CommitTransaction` and `SendMailResult`, not mid-transaction
- it is mail-only, the generic one also fires from trade, guild bank and TC9
- it passes `count`, which you can't get back from a merged stack

If that is the case you want to make, please rewrite the description around it and drop the "none of the three invoke sScriptMgr" part. As it stands the description argues for a hook that master already has, and that is what a reviewer will judge the PR on.

r-o-b-o-t-o:
> The "no way for a script to observe mail collection" rationale no longer holds. #27018 landed on master today ([2d0b881](https://github.com/azerothcore/azerothcore-wotlk/commit/2d0b8818bc93b6b9c12cc65e976cc1f303edb2bf), your merge base) and `MoveItemToInventory` now fires `OnPlayerAfterMoveItemToInventory` with the same `pLastItem` you return here, so that hook already runs on every mail take.
> 
> As I read it, what this hook still adds over the generic one is:
> 
> * it fires after `CommitTransaction` and `SendMailResult`, not mid-transaction
> * it is mail-only, the generic one also fires from trade, guild bank and TC9
> * it passes `count`, which you can't get back from a merged stack
> 
> If that is the case you want to make, please rewrite the description around it and drop the "none of the three invoke sScriptMgr" part. As it stands the description argues for a hook that master already has, and that is what a reviewer will judge the PR on.

I hadn't noticed that PR when making mine, only today when it was merged 😅 
I don't mind dropping mine in favour of the one already merged but I think it still brings something to the table and doesn't really hurt to have.
Either way, PR description is updated ✅

Nyeriah (approved):

Issue #23221, opened 2025-10-13
[Onyxia's Lair] - Onyxian Warders doesn't respawn when engaging Onyxia


TheSCREWEDSoftware:
### Triage

Issue linked from CC: https://github.com/chromiecraft/chromiecraft/issues/2957

Triage Notes: https://github.com/chromiecraft/chromiecraft/issues/2957#issuecomment-3396628306

Video of my testing: https://www.youtube.com/watch?v=ownRoyXjgTY&feature=youtu.be (more videos below)

Original Post Below: 

---

### What client do you play on?

enUS

### Faction

- [X] Alliance
- [ ] Horde

### Content Phase:

- [ ] Generic
- [ ] 1-19
- [ ] 20-29
- [ ] 30-39
- [ ] 40-49
- [ ] 50-59
- [X] 60

### Current Behaviour

When the boss is engaged in combat, previously killed warders remain dead

### Expected Blizzlike Behaviour

All Onyxia warders should respawn (but not engage) upon Onyxia being pulled so as to block exit from her room/lair.

### Steps to reproduce the problem

Kill all warders
Engage Onyxia

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/6f20b9ece71984f1a11469e65ede8accda05653e

### Operating system

Ubuntu 20.04

### Modules

- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-cfbg](https://github.com/azerothcore/mod-cfbg)
- [mod-chat-transmitter](https://github.com/azerothcore/mod-chat-transmitter)
- [mod-chromie-xp](https://github.com/azerothcore/mod-chromie-xp)
- [mod-desertion-warnings](https://github.com/azerothcore/mod-desertion-warnings)
- [mod-duel-reset](https://github.com/azerothcore/mod-duel-reset)
- [mod-eluna-lua-engine](https://github.com/azerothcore/mod-eluna-lua-engine)
- [mod-ip-tracker](https://github.com/azerothcore/mod-ip-tracker)
- [mod-low-level-arena](https://github.com/azerothcore/mod-low-level-arena)
- [mod-multi-client-check](https://github.com/azerothcore/mod-multi-client-check)
- [mod-pvp-titles](https://github.com/azerothcore/mod-pvp-titles)
- [mod-pvpstats-announcer](https://github.com/azerothcore/mod-pvpstats-announcer)
[...]

PR #27488, opened 2026-09-04
fix(Scripts/Onyxia): respawn warders on engage

Vinges541:
## Changes Proposed:
Draft implementation of the pull-triggered respawn proposed in #23221; the correct trigger and corpse/reset behavior remain unresolved. Loaded warders now respawn only after their corpse has decayed, preserving fresh corpses for looting and skinning. Despawned warders are scheduled through the map's normal respawn queue. Each GUID is validated against creature entry 12129 and the current map before it can be used.

This PR proposes changes to:
- [ ] Core (units, players, creatures, game systems).

### AI-assisted Pull Requests
- [x] AI tools were used to prepare this pull request.
   - Codex desktop assisted implementation and review; this follow-up used GPT-6 (exact backend snapshot unavailable).

## Issues Addressed:
- Addresses #23221

## SOURCE:
- [ ] Live research
- [ ] Sniffs
- [x] Video evidence, knowledge databases or other public sources
- [ ] The changes come from another project

The linked issue proposes a pull trigger, but its discussion distinguishes Vanilla/TBC footage and modern retail observations from Wrath-era behavior. I have not established whether Wrath uses a pull trigger, an evade trigger or a timer.

## Tests Performed:
- [ ] Tested in-game by another community member.
- [x] Further testing is required.

On current revision `f33529415`, the C++ codestyle check, C++ build matrix and dashboard integration passed. The [full e2e job](https://github.com/azerothcore/azerothcore-wotlk/actions/runs/34027519821/job/101471625826) failed in `TestAC_26997_SweepingStrikesExecuteNoCrash` because its Sweeping Strikes aura precondition was not met. This is outside the Onyxia diff; these checks do not verify the warder mechanic.

`git diff --check` and local C++ codestyle passed. Traced Creature::Respawn, SaveCreatureRespawnTime and ProcessRespawns in the PR base: dynamic respawns are no longer cancelled or invoked outside the queue. Gameplay verification and the trigger/lifetime decision remain outstanding.

## How to Test the Changes:
1. Kill the four entrance warders, let their corpses decay, then engage Onyxia. Confirm they return.
2. Repeat with fresh unlooted/skinnable corpses and confirm pulling Onyxia preserves them.
3. Repeat with living warders and confirm no duplicates appear.
4. Test both compatibility and dynamic respawn modes, including an unloaded warder grid.
5. In an isolated test DB, repoint one listed GUID to a different entry or map; confirm the encounter does not summon it.

Current limitation: these are persistent DB spawns, not entries in Onyxia's summons list. Restored warders remain after boss reset or death; this patch adds no despawn for them. The corpse guard skips fresh corpses entirely, so it does not fulfill an unconditional “all dead warders respawn on pull” rule. Both policies need evidence before changing them. Creature formation GroupAI supports respawn on evade, which would be appropriate only if evade is the confirmed trigger.

- [ ] In-game verification is still required.

### DIFF
--- a/src/server/scripts/Kalimdor/OnyxiasLair/boss_onyxia.cpp
+++ b/src/server/scripts/Kalimdor/OnyxiasLair/boss_onyxia.cpp
@@ -16,10 +16,15 @@
  */
 
 #include "CreatureScript.h"
+#include "Map.h"
 #include "Player.h"
 #include "ScriptedCreature.h"
 #include "SpellInfo.h"
 #include "onyxias_lair.h"
+
+#include <array>
+
+std::array<ObjectGuid::LowType, 4> constexpr OnyxianWarderSpawnIds = { 12550, 12551, 12552, 12553 };
 
 enum Spells
 {
@@ -195,6 +200,25 @@ public:
     {
         Talk(SAY_AGGRO);
         SetPhase(PHASE_GROUNDED);
+
+        for (ObjectGuid::LowType spawnId : OnyxianWarderSpawnIds)
+        {
+            bool hasAliveSpawn = false;
+            auto bounds = me->GetMap()->GetCreatureBySpawnIdStore().equal_range(spawnId);
+            for (auto itr = bounds.first; itr != bounds.second; ++itr)
+            {
+                if (!itr->second->IsAlive())
+                    itr->second->Respawn();
+                hasAliveSpawn |= itr->second->IsAlive();
+            }
+
+            // Dynamic respawns remove the dead object from the map, so force its database spawn as well.
+            if (!hasAliveSpawn)
+            {
+                me->GetMap()->RemoveCreatureRespawnTime(spawnId);
+                me->GetMap()->ProcessCreatureRespawn(spawnId);
+            }
+        }
 
         instance->DoStopTimedAchievement(ACHIEVEMENT_TIMED_TYPE_EVENT, ACHIEV_TIMED_START_EVENT); // just in case at reset some players already left the instance
         instance->DoStartTimedAchievement(ACHIEVEMENT_TIMED_TYPE_EVENT, ACHIEV_TIMED_START_EVENT);
--- a/src/server/scripts/Kalimdor/OnyxiasLair/boss_onyxia.cpp
+++ b/src/server/scripts/Kalimdor/OnyxiasLair/boss_onyxia.cpp
@@ -16,7 +16,9 @@
  */
 
 #include "CreatureScript.h"
+#include "GameTime.h"
 #include "Map.h"
+#include "ObjectMgr.h"
 #include "Player.h"
 #include "ScriptedCreature.h"
 #include "SpellInfo.h"
@@ -201,22 +203,28 @@ public:
         Talk(SAY_AGGRO);
         SetPhase(PHASE_GROUNDED);
 
+        Map* map = me->GetMap();
         for (ObjectGuid::LowType spawnId : OnyxianWarderSpawnIds)
         {
-            bool hasAliveSpawn = false;
-            auto bounds = me->GetMap()->GetCreatureBySpawnIdStore().equal_range(spawnId);
-            for (auto itr = bounds.first; itr != bounds.second; ++itr)
-            {
-                if (!itr->second->IsAlive())
-                    itr->second->Respawn();
-                hasAliveSpawn |= itr->second->IsAlive();
-            }
+            CreatureData const* data = sObjectMgr->GetCreatureData(spawnId);
+            if (!data || data->id != NPC_ONYXIAN_WARDER || data->mapid != map->GetId())
+                continue;
 
-            // Dynamic respawns remove the dead object from the map, so force its database spawn as well.
-            if (!hasAliveSpawn)
+            auto bounds = map->GetCreatureBySpawnIdStore().equal_range(spawnId);
+            if (bounds.first != bounds.second)
             {
-                me->GetMap()->RemoveCreatureRespawnTime(spawnId);
-                me->GetMap()->ProcessCreatureRespawn(spawnId);
+                for (auto itr = bounds.first; itr != bounds.second; ++itr)
+                {
+                    // Preserve fresh corpses for looting and skinning.
+                    if (itr->second->getDeathState() == DeathState::Dead)
+                        itr->second->Respawn();
+                }
+            }
+            else
+            {
+                // Despawned dynamic spawns must return through the normal respawn queue.
+                time_t respawnTime = GameTime::GetGameTime().count();
+                map->SaveCreatureRespawnTime(spawnId, respawnTime);
             }
         }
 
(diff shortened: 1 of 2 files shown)


Vinges541:
> Superseded by the 2026-09-06 follow-up: this report describes the earlier commit, not the current PR head. See the updated PR description for the current change and verification limits.

# Self-review — codex/fix-onyxian-warders → master

**Outcome** No merge-blocking findings.

**Caveat** Same-context review fallback because an independent review agent is unavailable in this session. In-game testing was not performed.

**Reviewed** `b08dd458f` (`codex/fix-onyxian-warders` vs `master`, merge-base diff — 1 file, +24 −0)

**By** Codex desktop, gpt-5.6-sol (reasoning xhigh) — self-review v0.7, 2026-09-04

**Verification** `git diff --check` and the AzerothCore C++ linter passed. Spawn ids 12550–12553 were checked against the repository database.

<details>

**Intent** Respawn the four permanent Onyxian Warders at the room entrance whenever Onyxia is engaged.

**Project rules** `.agents/docs/self-review-rules.md`, `code-review.md`, C++ guidelines, and script guidelines were applied.

**Diff** `425b9d9b7163c37668c1d37eacf5b1b49d6b349eead1cb8edf22dc678e1dfee1`

## Round 1 — `b08dd458f`, clean

Loaded dead warders use their normal `Respawn()` path. A warder whose dead object has already been removed from the map has its saved respawn time cleared and its persistent spawn processed. The four ids are limited to the entrance warders; the encounter's separately summoned Lair Guard is unchanged.

</details>

EricksOliveira:
covering both respawn modes is the right call, the compat/dynamic split is easy to miss. a few things on the fallback path though:

ProcessCreatureRespawn is the internal handler for the respawn queue, ProcessRespawns calls it right after popping the entry. calling it directly means the now + 5 re-queue it does when the grid isn't loaded lands in _respawnQueue with no matching entry in _creatureRespawnTimes, and a later SaveCreatureRespawnTime can't erase it anymore. maybe we can just call map->SaveCreatureRespawnTime(spawnId, now) and let ProcessRespawns pick it up next tick?

related: in dynamic mode Creature::Respawn() doesn't respawn in place, it saves the respawn time and adds the object to the remove list. so the loop schedules the respawn and RemoveCreatureRespawnTime right below cancels it. same suggestion covers both.

can we validate the spawn data before respawning? neither ProcessCreatureRespawn nor LoadCreatureFromDB checks the map id, so on a db where those guids got renumbered this spawns some unrelated creature inside the lair. a data->id != NPC_ONYXIAN_WARDER || data->mapid != me->GetMapId() skip would cover it.

12129 could go to onyxias_lair.h as NPC_ONYXIAN_WARDER while you're there.

just wondering: Respawn() also hits warders that are still fresh corpses, so pulling ony removes anything nobody looted or skinned yet. probably minor with the decay timer, but restricting it to DeathState::Dead would avoid it.

Vinges541:
Addressed in f33529415. Missing spawns are scheduled with SaveCreatureRespawnTime; the encounter no longer calls ProcessCreatureRespawn directly or cancels the timer established by Creature::Respawn. Existing objects are respawned only in DeathState::Dead, preserving fresh corpses. Spawn data must match NPC_ONYXIAN_WARDER (12129, now in onyxias_lair.h) and the current map. Full C++ lint and diff whitespace checks passed. Compatibility/dynamic respawn, unloaded-grid and corpse behavior still need in-game validation.

src/server/scripts/Kalimdor/OnyxiasLair/boss_onyxia.cpp
```diff
@@ -16,11 +16,18 @@
...
+#include "Map.h"
+#include "ObjectMgr.h"
 #include "Player.h"
 #include "ScriptedCreature.h"
 #include "SpellInfo.h"
 #include "onyxias_lair.h"
 
+#include <array>
+
+std::array<ObjectGuid::LowType, 4> constexpr OnyxianWarderSpawnIds = { 12550, 12551, 12552, 12553 };
```
moostigre: Are we sure we can hardcode ? Can use a database group instead ?

moostigre:
What happens if you pull and reset or when she dies ? There is no despawn of restored Wardens 

Preserving lootable corpses also contradicts the intended respawn on pull

Gultask:
Are you sure it isn't a simple timer? Or OnEvade?
creature_formations's GroupAI could cover this.

Vinges541:
The restored warders are existing DB spawns, not temporary summons registered in BossAI::summons. They therefore remain after reset/death; the summons.DespawnAll paths do not remove them. The fresh-corpse guard also skips those corpses entirely, so it does not implement unconditional respawn on pull. Those are real limitations of this proposal, now explicit in the description.

I converted this to draft while the trigger/lifetime policy is unresolved. The issue discussion does not give me enough Wrath-era evidence to choose pull vs timer vs evade, or to invent a despawn on boss death. GroupAI's respawn-on-evade flag is a viable data alternative if evade is confirmed, but it is not equivalent to the current pull hook. I have not tested these lifecycle cases in game.

AI-Autistic-Intelligence:
﻿@moostigre @Gultask Just to help clear the doubt on the 3.3.5 behavior and unblock this PR: the Onyxian Warders in the tunnel are actually *supposed* to respawn as soon as Onyxia is engaged in combat, and they despawn immediately when she dies. 

This mechanic was specifically designed by Blizzard to prevent players from hiding in the side tunnels or kiting the boss there to avoid mechanics. If a raid wipes, the Warders will remain alive in the tunnel during the run back. So using the engage hook (rather than a standard timer or evade hook) is indeed the correct blizzlike implementation. This PR is on the right track!

astgln:
Could you link the 3.2.2/3.3.5 footage, sniff, or other dated source for the pull/death/wipe behavior? In particular, we need to distinguish tunnel Warders (12129) from the Lair Guards summoned during the boss fight, and confirm what happens to fresh lootable/skinnable corpses.

The current draft preserves fresh corpses and leaves restored DB spawns alive after both a wipe and a kill, so it would still need changes if the behavior described here is confirmed. I am keeping it in draft until those lifecycle details are supported; the statement alone does not resolve the earlier source/version questions.

moostigre:
> @moostigre @Gultask Just to help clear the doubt on the 3.3.5 behavior and unblock this PR: the Onyxian Warders in the tunnel are actually _supposed_ to respawn as soon as Onyxia is engaged in combat, and they despawn immediately when she dies.
> 
> This mechanic was specifically designed by Blizzard to prevent players from hiding in the side tunnels or kiting the boss there to avoid mechanics. If a raid wipes, the Warders will remain alive in the tunnel during the run back. So using the engage hook (rather than a standard timer or evade hook) is indeed the correct blizzlike implementation. This PR is on the right track!

What happens if split your raid for a short period of time on pull and first nuke those spawned adds to free the tunnels so you can keep exploiting? Should they instantly respawn then?

PR #27507, opened 2026-09-04
fix(Core/Mail): announce every delayed mail, not just the first

r-o-b-o-t-o:
## Changes Proposed:

`Player::Update` cleared `m_nextMailDelivereTime` after delivering a mail and relied on the next mailbox open to recalculate it:

```cpp
SendNewMail();
++unReadMails;

// It will be recalculate at mailbox open (for unReadMails important
// non-0 until mailbox open, it also will be recalculated)
m_nextMailDelivereTime = time_t(0);
```

With two or more mails pending delivery, `AddNewMailDeliverTime` keeps only the earliest deadline. When that one fires the timer is zeroed and nothing re-arms it, so **every later mail arrives with no notification at all** until the player happens to open a mailbox (which runs `UpdateNextMailTimeAndUnreads` and rescues it) or relogs.

The auction house reaches this through **"sale successful"** mails, which are the ones actually sent on `MailDeliveryDelay`, so a seller with two auctions completing inside that window silently loses the notification for the second.

To be precise about the setting, since the original wording here was wrong: `SendAuctionSalePendingMail` passes **no** `deliver_delay`, so "sale pending" mail is delivered immediately — `MailDeliveryDelay` only shapes its *expiry*, through the `MAIL_AUCTION` branch in `MailDraft::SendMailTo`. Only `SendAuctionSuccessfulMail` passes the delay as a delivery delay. Thanks to @Nyeriah for catching it.

Recounting from the mailbox instead of clearing the timer makes mail that is still undelivered keep its own delivery time, so it is announced when it arrives. `UpdateNextMailTimeAndUnreads` also sets `unReadMails` to the true count, which replaces the blind `++unReadMails` — the just-delivered mail is unread and not expired, so the recount is always at least 1 and the notification state is preserved.

This PR proposes changes to:

### AI-assisted Pull Requests

## Issues Addressed:

None.

## SOURCE:

A/B tested with a server-side packet capture, same scenario both times — two item mails sent 32 seconds apart, mailbox untouched in between:

| | before | after |
|---|---|---|
| mails delivered (confirmed in `mail` table) | 2 | 2 |
| `SMSG_RECEIVED_MAIL` in capture | **1** | **2** |

An earlier run of the same test showed two notifications on the unpatched build, because the mailbox was opened between the two deliveries and `HandleGetMailList` re-armed the timer through the old path. Keeping the mailbox closed is what makes the defect visible.

## Tests Performed:

Tested on Windows, 3.3.5a client, with `MailDeliveryDelay` lowered to 60s to keep the cycle short. Also checked that a single delayed mail still notifies exactly once, and that instant mail (no attachment, same account) is unaffected — that path goes through `AddNewMailDeliverTime` and never sets the timer.

## How to Test the Changes:

1. Lower `MailDeliveryDelay` to 60 in worldserver.conf to keep the wait short.
2. From character A, mail an **item** to character B on a **different account** — attachments across accounts are what trigger the delivery delay. Keep B online throughout. (Two auctions completing within the delay window is the natural in-game equivalent; hand-sent item mail is just easier to time.)
3. Wait ~30 seconds, then send a second item mail from A to B.
4. **Do not open B's mailbox at any point** until both delays have elapsed. A single mailbox open re-arms the timer through the old path and hides the defect.
5. B should get two separate "new mail" notifications. On an unpatched build only the first arrives; the second mail lands silently and only appears after a mailbox open or relog.

Note that the minimap envelope stays lit from the first notification, so the second is not visually obvious — confirm with a packet capture (`PacketLogFile`) and count `SMSG_RECEIVED_MAIL`, or watch the second mail appear.

- [ ] No automated test. The e2e harness (AzerothGhost v1.0.8) has no mail opcode support, so this could not be encoded as a regression test per `.agents/docs/e2e-policy.md`.

---

### DIFF
--- a/src/server/game/Entities/Player/PlayerUpdates.cpp
+++ b/src/server/game/Entities/Player/PlayerUpdates.cpp
@@ -61,11 +61,10 @@ void Player::Update(uint32 p_time)
     if (m_nextMailDelivereTime && m_nextMailDelivereTime <= GameTime::GetGameTime().count())
     {
         SendNewMail();
-        ++unReadMails;
 
-        // It will be recalculate at mailbox open (for unReadMails important
-        // non-0 until mailbox open, it also will be recalculated)
-        m_nextMailDelivereTime = time_t(0);
+        // Recount from the mailbox instead of clearing the timer: any mail that is still
+        // undelivered keeps its own delivery time and gets announced when it arrives
+        UpdateNextMailTimeAndUnreads();
     }
 
     // Update cinematic camera (if needed)


Nyeriah:
The description says the auction house sends both "sale pending" and "sale successful" on MailDeliveryDelay, but only the successful mail is delayed. SendAuctionSalePendingMail passes no deliver_delay (https://github.com/azerothcore/azerothcore-wotlk/blob/4e80596cdaa21fa31830522f6f2d7ed8750bfffd/src/server/game/AuctionHouse/AuctionHouseMgr.cpp#L181), so it lands immediately; the setting only shapes its expiry in MailDraft::SendMailTo. The delay is only passed for the successful mail (https://github.com/azerothcore/azerothcore-wotlk/blob/4e80596cdaa21fa31830522f6f2d7ed8750bfffd/src/server/game/AuctionHouse/AuctionHouseMgr.cpp#L217).

The bug itself still reproduces with two sales completing in flight, so the fix stands. Could you reword the description and commit message to say "sale successful" mails, so the motivating case matches what the code does?

r-o-b-o-t-o:
You're right, and thanks for checking it — I had that backwards. `SendAuctionSalePendingMail` passes no `deliver_delay` at all, so that mail lands immediately; the `deliveryDelay` local there only feeds the body text telling the seller when the money arrives, and `MailDeliveryDelay` shapes that mail's *expiry* through the `MAIL_AUCTION && m_items.empty() && !m_money` branch in `MailDraft::SendMailTo`. Only `SendAuctionSuccessfulMail` passes it as a delivery delay.

Reworded both the commit message and the description to say "sale successful" mails, with a note on what the setting actually does to the pending one so the next reader doesn't repeat my mistake.

The repro is unchanged — two mails pending delivery at once, mailbox untouched between them — and the packet capture still shows one `SMSG_RECEIVED_MAIL` before the change and two after.

Nyeriah (approved):

PR #27512, opened 2026-09-05
feat(Core/Arena): swap OnGetStartPersonalRating for OnAddMember, add OnArenaWeekReset

Helias:
## Changes Proposed:

This PR proposes changes to:

Replaces `OnGetStartPersonalRating` (added in #27069, I'm its only consumer) with a hook that hands out the whole member, and adds a realm-wide one for the weekly arena point reset:

```cpp
virtual void OnAddMember(ArenaTeam* team, ArenaTeamMember& member) { }   // ArenaScript
virtual void OnArenaWeekReset() { }                                      // GlobalScript
```

`OnGetStartPersonalRating` only exposed the personal rating, and that turned out to be too narrow. `ArenaTeam::AddMember` hardcodes `WeekGames`, `WeekWins`, `SeasonGames` and `SeasonWins` to 0 and `CHAR_INS_ARENA_TEAM_MEMBER` wrote nothing but `(arenaTeamId, guid, personalRating)`, so a module had no way to seed those counters for a joining member.

`OnAddMember` is called with the fully built `ArenaTeamMember` right before it goes onto the team, so whatever a script leaves on the struct is what the team holds and what lands in `arena_team_member`. It can't veto the join, `CanAddMember` already does that.

The INSERT binds the counters from that struct now. With no script registered the row it writes is identical to the one we wrote before, since the counters are still 0 at that point and the columns already default to 0. No schema change.

`OnArenaWeekReset` fires once in `ArenaTeamMgr::DistributeArenaPoints`, after the loop that calls `FinishWeek()` and saves every team. It's realm-wide and takes no parameters, for modules that keep their own copy of week statistics and need to clear it when the core clears its own. It fires even if no team played that week.

Both hooks do nothing on their own, the default implementations are empty.

I'm using them in [mod-arena-rating-memory](https://github.com/azerothcore/mod-arena-rating-memory), which gives a player back their old rating and played games when they rejoin a team they used to be in, but neither hook assumes anything about that.

### AI-assisted Pull Requests

## Issues Addressed:

None.

## SOURCE:

Not applicable, these are scripting hooks and not a behaviour fix.

## Tests Performed:

Builds clean with `-DBUILD_TESTING=1 -DSCRIPTS=static -DMODULES=static` on Linux, GCC. `ArenaHookDefaultsTest.OnGetStartPersonalRatingKeepsCoreValue` is replaced by `OnAddMemberLeavesMemberUntouched`, which pre-fills a member with a rating and counters, calls the hook with no script registered and expects all five values unchanged.

## How to Test the Changes:

1. Build with `-DBUILD_TESTING=1` and run `unit_tests`, `ArenaHookDefaultsTest.OnAddMemberLeavesMemberUntouched` should pass.
2. Without any module registering the hook, create an arena team and invite a player. Their row in `arena_team_member` is the same as before this PR: rating 1000 on a team rated 1000+, 0 below that (or whatever `Arena.ArenaStartPersonalRating` is set to), and all four counters 0.
3. Let the team play a rated match and trigger the weekly distribution (short `Arena.AutoDistributeInterval`). Points are handed out and week stats reset exactly as before, `OnArenaWeekReset` has no listener.
4. With a module implementing `OnAddMember` and writing to the member, the joining player gets those values in `arena_team_member` right away, without a reload.

- [ ] Nothing pending.

### DIFF
--- a/src/server/database/Database/Implementation/CharacterDatabase.cpp
+++ b/src/server/database/Database/Implementation/CharacterDatabase.cpp
@@ -278,7 +278,7 @@ void CharacterDatabaseConnection::DoPrepareStatements()
 
     // Arena teams
     PrepareStatement(CHAR_INS_ARENA_TEAM, "INSERT INTO arena_team (arenaTeamId, name, captainGuid, type, rating, backgroundColor, emblemStyle, emblemColor, borderStyle, borderColor) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", CONNECTION_ASYNC);
-    PrepareStatement(CHAR_INS_ARENA_TEAM_MEMBER, "INSERT INTO arena_team_member (arenaTeamId, guid, personalRating) VALUES (?, ?, ?)", CONNECTION_ASYNC);
+    PrepareStatement(CHAR_INS_ARENA_TEAM_MEMBER, "INSERT INTO arena_team_member (arenaTeamId, guid, weekGames, weekWins, seasonGames, seasonWins, personalRating) VALUES (?, ?, ?, ?, ?, ?, ?)", CONNECTION_ASYNC);
     PrepareStatement(CHAR_DEL_ARENA_TEAM, "DELETE FROM arena_team WHERE arenaTeamId = ?", CONNECTION_ASYNC);
     PrepareStatement(CHAR_DEL_ARENA_TEAM_MEMBERS, "DELETE FROM arena_team_member WHERE arenaTeamId = ?", CONNECTION_ASYNC);
     PrepareStatement(CHAR_UPD_ARENA_TEAM_CAPTAIN, "UPDATE arena_team SET captainGuid = ? WHERE arenaTeamId = ?", CONNECTION_ASYNC);
--- a/src/server/game/Battlegrounds/ArenaTeam.cpp
+++ b/src/server/game/Battlegrounds/ArenaTeam.cpp
@@ -137,8 +137,6 @@ bool ArenaTeam::AddMember(ObjectGuid playerGuid)
     else if (GetRating() >= 1000)
         personalRating = 1000;
 
-    sScriptMgr->OnGetStartPersonalRating(this, playerGuid, personalRating);
-
     // xinef: sync query
     // Try to get player's match maker rating from db and fall back to config setting if not found
     CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_SEL_MATCH_MAKER_RATING);
@@ -177,6 +175,8 @@ bool ArenaTeam::AddMember(ObjectGuid playerGuid)
     newMember.MatchMakerRating = matchMakerRating;
     newMember.MaxMMR           = maxMMR;
 
+    sScriptMgr->OnAddMember(this, newMember);
+
     Members.push_back(newMember);
     sCharacterCache->UpdateCharacterArenaTeamId(playerGuid, GetSlot(), GetId());
 
@@ -184,7 +184,11 @@ bool ArenaTeam::AddMember(ObjectGuid playerGuid)
     stmt = CharacterDatabase.GetPreparedStatement(CHAR_INS_ARENA_TEAM_MEMBER);
     stmt->SetData(0, TeamId);
     stmt->SetData(1, playerGuid.GetCounter());
-    stmt->SetData(2, personalRating);
+    stmt->SetData(2, newMember.WeekGames);
+    stmt->SetData(3, newMember.WeekWins);
+    stmt->SetData(4, newMember.SeasonGames);
+    stmt->SetData(5, newMember.SeasonWins);
+    stmt->SetData(6, newMember.PersonalRating);
     CharacterDatabase.Execute(stmt);
 
     // Inform player if online
--- a/src/server/game/Battlegrounds/ArenaTeamMgr.cpp
+++ b/src/server/game/Battlegrounds/ArenaTeamMgr.cpp
@@ -268,6 +268,8 @@ void ArenaTeamMgr::DistributeArenaPoints()
         }
     }
 
+    sScriptMgr->OnArenaWeekReset();
+
     ChatHandler(nullptr).SendWorldText(LANG_DIST_ARENA_POINTS_TEAM_END);
 
     ChatHandler(nullptr).SendWorldText(LANG_DIST_ARENA_POINTS_END);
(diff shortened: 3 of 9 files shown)


Helias:
The module PR using both hooks: azerothcore/mod-arena-rating-memory#2. This one merges first.

Helias:
# Self-review — feat/arena-on-add-member → master

**Outcome** 4 findings — 2 fixed, 2 dismissed

**Caveat** the two fixes were not re-reviewed: they landed in `f9093ecd3` after the reviewed state and round 2 was skipped by the author. Both are comment/test-only.

**Reviewed** `614cc863a` (`feat/arena-on-add-member` vs `upstream/master`, merge-base diff — 9 files, +50 −17). Current head `f9093ecd3` (9 files, +53 −17) carries the fixes.

**By** Claude Code CLI, claude-opus-5[1m] — self-review v0.7, 2026-09-05

**Verification** Full build clean (GCC, `RelWithDebInfo`, `SCRIPTS=static MODULES=static BUILD_TESTING=ON`); `unit_tests` arena suites 11/11 pass; `codestyle-cpp.py` and `codestyle-sql.py` pass. CI green on `614cc863a` across GCC 15 and Clang 18/21, pch and nopch, with and without modules.

**Not tested in-game by the author.** PTR validation is planned by the tester who reported the original issue. The scenarios a reviewer would want confirmed are listed at the end.

<details>

**Intent** Replace the `ArenaScript` hook `OnGetStartPersonalRating` with `OnAddMember`, which hands a script the whole `ArenaTeamMember` right before it is added to the team and written to `arena_team_member`, so a script can seed the played-game counters and not just the rating; and add a realm-wide `GlobalScript` hook `OnArenaWeekReset`, fired once per weekly arena point distribution after every team's week statistics have been reset.

**Project rules** `.agents/docs/self-review-rules.md` found and applied.

**Diff** `91ad0c67216a5f04970efbc6bc50051278ea2ab2`

**Excluded from the reviewer** `.agents/plans/` and `*.SELF-REVIEW.md`. All 9 changed files covered, none skipped.

**e2e** `e2e full` failed on this commit with `TestAC_27095_FreyaAlliesSpawnRateReduction` (Ulduar). The same workflow also fails on plain `master` (run 33958806292) with a different test, `TestAC_26363_TimmyEmergesAfterSquareCleared`, and on an unrelated PR branch. No arena-team e2e coverage exists — the only "arena" hits in the log are a Nagrand Arena pad borrowed by a charm combat test. Attributed to pre-existing flakiness, not this change.

## Round 1 — `614cc863a`, 4 findings

1. `ArenaTeam.cpp:995` (`FinishWeek`) — the early return `if (Stats.WeekGames == 0) return false;` skips the member-counter reset. Once a script can seed `member.WeekGames` on a team whose `Stats.WeekGames` is 0, the counter survives the flush in memory and in `arena_team_member`; the next time that team plays 10 games, `UpdateArenaPointsHelper` grants a full week of arena points on the stale count. Before this change the invariant held, because member counters only moved in `MemberWon`/`MemberLost`, which run in the same match flow as `FinishGame`'s `Stats.WeekGames += 1`. → **dismissed**: author's position, verbatim — "I don't think that the PR core needs any fixes", "I don't think this change C1 is necessary". In the ordinary flow the module's `ForgetWeek` keeps both sides in step, since remembered week counters can only exist for a team that played that week; the paths that break it (module disabled across a flush, the clearing write lost to a shutdown) are documented operator hazards with a documented remedy.
2. `ArenaScript.h:64-66` — the contract comment promised that whatever the hook leaves on the member is stored, true only for the five bound columns. `AddMember` never persists `MatchMakerRating`/`MaxMMR` (only the next `SaveToDB` does, and no join path calls it), so a seeded MMR is lost on a restart before the member's next rated match and silently falls back to `CONFIG_ARENA_START_MATCHMAKER_RATING`; `Guid`/`Name`/`Class` are identity, and the INSERT and `UpdateCharacterArenaTeamId` both use the local `playerGuid`, so rewriting them desyncs the member from its row and the character cache. Independently raised by CodeRabbit ("limit the fields `OnAddMember` can change", Major) and Copilot. → **fixed** in `f9093ecd3`: comment now names what is persisted at join and what is not. Signature deliberately not narrowed — core hands scripts whole mutable structs throughout (`OnScaleAuraUnitAdd(…, TargetInfo&)`, `OnConditionCheck(…, ConditionSourceInfo&)`, `OnPlayerBeforeFillQuestLootItem(Player*, LootItem&)`, `OnPatchValuesUpdate(…, ByteBuffer&, …)`), and narrowing would defeat the hook's purpose.
3. `ArenaHookDefaultsTest.cpp:102` — `ArenaTeamMember member;` left `Class`, `MatchMakerRating`, `MaxMMR` indeterminate. No live UB (the test reads only what it sets, and the fresh-eyes reviewer declined to raise it), but free to foreclose. Raised by Copilot. → **fixed** in `f9093ecd3`: `ArenaTeamMember member{};`.
4. `CharacterDatabase.cpp:281` — changed line is 211 columns vs `.editorconfig`'s 120. Raised by CodeRabbit. → **dismissed**: 401 of that file's 663 lines already exceed 120 and the neighbours run 121–240; the statement was already 153 columns before this change; the `C++ Codestyle` workflow runs only `codestyle-cpp.py` and `cppcheck`, neither of which checks length, and it passed on this exact commit. Wrapping one statement would make it the only wrapped one in the file. Same disposition as the identical case on `ArenaScript.cpp:61` in the predecessor PR's review.

**Verified as claimed, no finding** — counter columns already exist in `arena_team_member` with `DEFAULT 0`, so no schema change and no SQL file touched (AGENTS.md immutability respected); with no script registered the written row is byte-identical to before; the hook cannot veto the join (`void` return, `CanAddMember` still guards); `OnArenaWeekReset` fires outside the per-team loop, so it runs even when no team played, on the scheduled, manual (`cs_misc.cpp`) and `BattlegroundMgr` paths alike; no stale references to the removed hook remain; `ARENAHOOK_*` values unchanged by the in-place replacement; no leftovers.

</details>

## In-game coverage a reviewer would want confirmed

With **no module loaded**, confirming nothing changed:

1. Join a fresh arena team — the `arena_team_member` row matches pre-PR behaviour: rating 1000 on a team rated 1000+, 0 below, or `Arena.ArenaStartPersonalRating`; all four counters 0.
2. Both join paths — invite/accept (`ArenaTeamHandler`) and petition turn-in (`PetitionsHandler`). The second is the easy one to forget.
3. Play a rated match, then trigger the weekly distribution (short `Arena.AutoDistributeInterval`, or `.flusharenapoints`) — points paid and week stats reset exactly as before.

With a module implementing the hook:

4. A seeded member's rating and counters land in `arena_team_member` at join, with no relog and no match.
5. The dismissed finding 1 scenario, if you want to probe it: seed a member's week counters on a team that then sits out a whole week, and check whether the counters survive the flush.

PR #27241, opened 2026-08-20
fix(DB/SAI): Restore Timmy the Cruel

moostigre:
## Changes Proposed:

This PR restores Timmy the Cruel as a usable encounter in Stratholme:

- Keeps Timmy as a persistent spawn, hidden and immune on respawn in SmartAI phase 1.
- Every 10 seconds, checks that no living Crimson Guardsman, Conjuror, Initiate, or Gallant remains within the verified 70-yard activation area.
- Once the area is clear, reveals Timmy, removes his immunities, casts Emerge, and starts the repeating sniffed route.
- Moves Timmy to the WotLK-sniffed position, casts the ghoul Emerge visual (50142), then immediately starts the repeating sniffed path with SmartAI Action 232, matching the captured movement.
- Adds the complete 23-point looping walking route captured on an official server.
- Applies passive Thrash (12787), casts Ravenous Claw (17470), and triggers Enrage (8599) below 50% health.
- Keeps the TIMMY! yell on engage.

This PR proposes changes to:
- [ ] Core (units, players, creatures, game systems).
- [ ] Scripts (bosses, spell scripts, creature scripts).
- [x] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

- [x] AI tools were used partially to prepare this pull request.
  - Tools/models used: Codex GPT-5.6
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:

- Closes #26363

## SOURCE:

The changes have been validated through:
- [x] Live research from an official TBC Anniversary 2.5.6.69110 Ymir sniff supplied by the author. The capture contains the full repeated path, engage yell, aura and combat spell evidence.
- [x] WotLK sniff data supplied by the reviewer for the spawn position and CreateObject value.
- [x] Video evidence from the linked issue: https://youtu.be/BnYU-HsIU6Q?t=219
- [x] The linked 2008 Wowhead comment identifies the minimum retail trigger as the Scarlet creatures closest to the door. PTR testing showed that limiting the gate to that formation lets Timmy emerge into other hostile Scarlet packs along his route, so the trigger covers every Scarlet spawn on the square.

The target is WotLK. The official TBC Anniversary capture is used for behavioral evidence; the WotLK spawn data is retained, and TrinityCore 3.3.5 independently corroborates Thrash spell 12787 and the Ravenous Claw scheduling used here.

## Tests Performed:

- [x] AzerothCore SQL codestyle check.
[...]

Issue #26363, opened 2026-06-27
[DB/Creature][Stratholme - Living] Timmy not spawning after clearing Crusaders' Square


LoturCC:
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/9800

Triage Notes: 
Reproduce:
- .go xyz 3641.1216 -3190.08 127.96045 329
- wait for them to kill the ghoul and observe he won't spawn
test image: 
https://i.imgur.com/oUWyhg8.png
https://www.wowhead.com/wotlk/npc=10808/timmy-the-cruel#comments

Original Post Below: 

---
### Current Behaviour

After clearing the mobs from Crusaders' Square on the living side of Stratholme [Timmy the Cruel](https://wowgaming.altervista.org/aowow/?npc=10808) isn't spawning like he should do.

I encountered this issue yesterday with one of my alts and retested it again with a level 80 character on live after the server restart happened, but there was no change in behaviour.

### Expected Behaviour

After clearing the mobs on Crusaders' Square (the 3 or 4 mobs in front of the door), Timmy should emerge and roam the square.

### Official Sources (that prove or disprove) the issue being reported

A Wow Classic Video showing that he is already spawning after the group cleared almost everything aside some lone crusader fighting a skeleton: https://youtu.be/BnYU-HsIU6Q?t=219

There is also a rather old comment explaining how to get him to spawn here: https://www.wowhead.com/wotlk/npc=10808/timmy-the-cruel#comments
<img width="1302" height="122" alt="Image" src="https://github.com/user-attachments/assets/76c620f6-8479-49be-b33e-05c9aceed8c4" />

### Steps to reproduce the problem

1. Enter Stratholme from the Living Side and fight your way to Crusaders' Square
2. Kill all mobs on Crusaders' Square
3. Timmy not spawning :'(

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/f5b21bb21e08bfdcfb93984ed1ae5b13beab19ec

### Operating system

Ubuntu 24.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
[...]

PR #27514, opened 2026-09-05
test(e2e): review, flakes, no unsolicited e2e

walkline:
Follow-up to #27158 for Yehonal's review, plus a few flake fixes that showed up while re-running the suite.

Job timeout goes from 480 to 90 (the only real ceiling, since `go test -timeout` is per package). `E2E_RUNS_ON` is documented as ephemeral only. The `github.repository` guard comment now says what it actually does: fork PRs still run in upstream CI, the check only skips the job inside a fork's own CI.

`e2e/` is `gofmt`-clean and `e2e-live.yml` fails on drift. `spawnAndBoardSteed` no longer returns the unused `boarded` GUID.

Agent docs no longer tell agents to add e2e unless asked.

Flakes:
- Stormwind Steed boards on Argent Tournament so Armistice is not stripped by `spell_area`.
- `TestTrade_StackableMerge` uses the same inventory wait as the dual-accept test.
- Timmy resets leftover instance state before the oracle.
- Freya enters Ulduar as GM, re-acquires after `CombatReady`, and fails a missed wave in 90s instead of 4m.

## Changes Proposed:
This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Grok 4.6 (xAI). The agent wrote the code changes and this description.

## Issues Addressed:
- Closes

## SOURCE:

## Tests Performed:
This PR has been:

Local live auth+world, each as `go test -count=1`:
- `TestTrade_StackableMerge` 10/10
- `TestVehicles_EnterExitStormwindSteed` 10/10
- `TestAC_26363_TimmyEmergesAfterSquareCleared` 10/10
- `TestAC_27095_FreyaAlliesSpawnRateReduction` 10/10 (one run at a time)

## How to Test the Changes:

1. On this PR, wait for `nopch-build` `ubuntu-24.04-clang-18-nopch` and the following `e2e full` job. Confirm `Check gofmt` runs and the job `timeout-minutes` is 90.
2. Locally, with auth+world+MySQL up and `E2E_*` from `e2e/.env.example`:
   `cd e2e && go test -tags=e2e ./suites/social/trade ./suites/combat/vehicles ./suites/instances/classic/stratholme ./suites/instances/northrend/ulduar -count=1 -timeout 15m -parallel 1 -v`
3. Timmy expects the current master SAI from #27241. Freya is about 2 minutes per pass (60s add waves). A Lasher-first pull waits for a third wave and takes closer to 3 minutes.

- Freya cannot be a short smoke. The oracle needs two real add waves.


sudlud (approved):

FrancescoBorzi (approved):

Yehonal (commented):
Thanks for the follow up. The flake work here is solid and most of it I'd take as is.

Two things I'd like to sort out first.

## The Stratholme suite

I think we should keep this one. As far as I can see it's the only e2e test covering #26363, and #27241 added it exactly as a regression for that DB fix. If we remove the file we lose that coverage completely, and the inventory in this same PR marks it manual-only.

The policy you're adding in this same PR says to classify and fix, or quarantine with a ticket, and not to remove a green regression test just because the bug is closed. I'd rather follow that here.

Can we restore it with a deterministic instance reset? If that's not quick, maybe we keep it compiling in quarantine, with a ticket and a condition to re-enable it.

One thing on the diagnosis: every CI job I looked at runs `setup-db` on an ephemeral runner, and the same SHA alternates between pass and "Timmy already present". That doesn't look like leftover instance state to me. Did you rule out a race between grid load, the 10s SmartAI tick and the object cache?

## `TestTrade_StackableMerge` can pass on a partial transfer

`waitInv` returns the last count it saw when the timeout expires, it doesn't fail (`trade_e2e_test.go:88`). At lines 244 and 245 we pass the predicates `n == a0-5` and `n == b0+5`, but we never assert they were reached.

After that we only check that the total is conserved and that `b1 > b0`. So a transfer of 4 instead of 5 would still pass.

`waitInvAtLeast` just above already does the right thing with `Preconditionf`. Could we do the same here, or assert both values explicitly?

## Description

It still says "Timmy resets leftover instance state before the oracle", and "How to Test" asks to run `./suites/instances/classic/stratholme`, but that package isn't there anymore. Could you please refresh it on the final SHA? It would also help to say that the vehicle test now applies Armistice via GM.

## A few questions, not blockers

The job timeout: the interface recommends `>=120m`, the runner caps at 90m, and `go test` gets 180m per package. The 19 packages run serially and the recent jobs I checked land between 9 and 13 minutes, so 90m is plenty for normal CI. It does cut off configurations the workflow says are valid though. Can we align the three?

Freya: the comment says we re-acquire a living Freya, but the GUID is only replaced when the object disappears from the cache. If it stays there with zero HP the loop just times out. Is that intentional?

The vehicle test: should it cover the real `spell_area` precondition, or only enter, exit and disconnect? If it's the second one the GM fallback is fine, but then applying Armistice by hand would hide an area regression.

On the agent policy: the gate on writing new tests makes sense to me. As written though it also seems to stop an agent from reporting a missing oracle. Could we split "don't implement without a request" from "always report the gap and ask"?

## What I like

The `gofmt` gate is simple and fail closed, that's the kind of check I want.

Moving the vehicle test where Armistice survives is a real fix, not a workaround.

And replacing the immediate reads in trade with bounded polling is the right approach to that race. We just need the polling result to be binding.

Yehonal (changes requested):
Just to be explicit about what blocks this for me: `TestTrade_StackableMerge` can pass on a partial transfer, so I'd like that one sorted before merge.

On the Stratholme suite I'd still prefer we restore it, but quarantine plus a ticket works for me if that's faster.

walkline:
Hey @Yehonal I believe feedback should be addressed.

As for this:
> On the agent policy: the gate on writing new tests makes sense to me. As written though it also seems to stop an agent from reporting a missing oracle. Could we split "don't implement without a request" from "always report the gap and ask"?

With showing the coverage gap, it would encourage users to write e2e tests. But given that a big chunk of our active reviewers will currently ignore PRs with e2e tests, I think it makes more sense to discourage adding them to regular PRs for now, to keep the review process simpler.

Yehonal (approved):
Both sorted, thanks. `waitInvEqual` fails properly now, and the Stratholme suite is back.

The four questions from my first review aren't blockers. Worth a look at some point, but not here.

PR #27517, opened 2026-09-06
feat(Codestyle/SQL): require both id and guid in spawn DELETEs

sudlud:
## Changes Proposed:
The SQL codestyle script now requires every ``DELETE FROM `creature` `` and ``DELETE FROM `gameobject` `` in a pending update to filter on both `id` and `guid`.

A guid-only delete wipes whatever spawn happens to own that guid at import time, and an id-only one wipes every spawn of that entry in the world. Asking for both keeps the delete bounded and makes it obvious what it removes.

Two things follow from that and are worth calling out, because contributors will hit them:

- Only `=`, `IN` and `BETWEEN` count as filters. ``WHERE `id` = 5 AND `guid` > 0`` is rejected, because `> 0` bounds nothing.
- `OR` is refused outright, even when both columns are filtered. Judging ``WHERE `id` = 1 OR `guid` = 2`` properly needs real boolean parsing, so the check does not guess: use `IN`, or split it into one statement per spawn. No spawn delete in the current tree uses `OR`.

The check only runs on `pending_db_*`, so nothing already merged is affected and CI stays green today. One thing to be aware of: a deliberate mass-despawn like ``DELETE FROM `creature` WHERE `id` IN (...) AND `map` = 603;`` will now fail and has to spell out the guids. That is intentional, but say so if you'd rather have an escape hatch.

`.agents/docs/sql-guidelines.md` gets the rule added to its list of linted conventions.

This PR proposes changes to:

### AI-assisted Pull Requests

   - Tools/models used: Claude Opus 5 (Claude Code)

## Issues Addressed:
None.

## SOURCE:

## Tests Performed:
This PR has been:

No in-game testing, this only touches CI tooling.

Fixtures in `pending_db_world/`, run through the real script. Accepted: both orders of `id`/`guid`, `BETWEEN` with `@GUID` variables, parenthesised `IN` lists, a statement split over three lines with the two filters on different lines, an inline ``/* comment */`` between the table and the `WHERE`, a quoted value containing `;`, an `UPDATE` whose text happens to mention `DELETE FROM creature`, `creature_addon`, and `--`, `#` and `/* */` comments. Rejected: guid-only, id plus `map`, the pre-rename `id1` column, no `WHERE` at all, `OR`, `>` and `!=`, a quoted `'id = 1 AND guid = 2'` masquerading as a filter, an aliased ``DELETE FROM `creature` `c` ``, and a second delete sharing a line with a valid one. `creature_template` still produces only the existing "should not be deleted" message.

Parser run over the whole tree, 7449 files across `updates/`, `base/`, `archive/` and `old/`: no crashes, no statement mis-parses, and of the 54 spawn deletes merged since the `id1` → `id` rename, 14 would be rejected. Every one of those is an intentional id-only or map-bounded delete, which is the category the rule is meant to catch. That corpus scan is the real evidence here; a run against the empty `pending_db_*` directories proves nothing, since it checks no files.

## How to Test the Changes:

1. Put a file in `data/sql/updates/pending_db_world/` with e.g. ``DELETE FROM `creature` WHERE `guid` IN (132304, 132305);``
2. Run `python apps/codestyle/codestyle-sql.py` from the repo root.
3. It should report the missing `id` filter and exit 1. Add ``AND `id` = <entry>`` and it passes.

- [ ]


heyitsbench:
If the people won't listen, Sudlud will *make* them listen. 👍

Nyeriah (approved):

TheSCREWEDSoftware (approved):

PR #27528, opened 2026-09-07
fix(Core/Item): null-owner guard in CheckSoulboundTradeExpire

Celandriel:
Playerbots fix. Guard from invalid owner. `GetOwner()` was dereferenced unconditionally and returns null when the owner isn't resolvable, which crashed the map update thread (crashlog in mod-playerbots#127). Returns false so the item stays tracked and the check retries instead.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Fable: Breakup of playerbot core

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Item/Item.cpp
+++ b/src/server/game/Entities/Item/Item.cpp
@@ -1286,10 +1286,15 @@ void Item::ClearSoulboundTradeable(Player* currentOwner)
 
 bool Item::CheckSoulboundTradeExpire()
 {
-    // called from owner's update - GetOwner() MUST be valid
-    if (GetUInt32Value(ITEM_FIELD_CREATE_PLAYED_TIME) + 2 * HOUR < GetOwner()->GetTotalPlayedTime())
+    // we have to check the owner for mod_playerbots since bots programically call methods like DestroyItem,
+    // MoveItemToMail, DestroyItemCount which do not handle soulboundTradeable clearing.
+    Player* owner = GetOwner();
+    if (!owner)
+        return true; // remove from tradeable list
+
+    if (GetUInt32Value(ITEM_FIELD_CREATE_PLAYED_TIME) + 2 * HOUR < owner->GetTotalPlayedTime())
     {
-        ClearSoulboundTradeable(GetOwner());
+        ClearSoulboundTradeable(owner);
         return true; // remove from tradeable list
     }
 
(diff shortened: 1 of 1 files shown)


sudlud:
`CheckSoulboundTradeExpire` returning `true` here erases the item from `m_itemSoulboundTradeable`, but nothing cleared `ITEM_FIELD_FLAG_BOP_TRADEABLE` or the `item_soulbound_trade_data` row, so it stays tradeable and unwatched until the next login re-adds it. can we return `false` instead? a null owner means we can't tell yet whether the 2 hours are up, so retrying on the next tick seems closer to what we want.

the comment above it still says `GetOwner()` MUST be valid, which is the opposite of what the code now does. can we drop it?

this is a pick of `58e3316c`, `befa0158` and `7aaefb76` by bashermens on the Playerbot branch of mod-playerbots/azerothcore-wotlk, so worth ticking the cherry-pick box and naming the source. the `Co-authored-by` on the commit already covers the credit.

on the reason that came with it: `MoveItemToMail` doesn't exist in core or in mod-playerbots. the closest module paths, `GiveItemAction` and `SendMailAction`, both go through `MoveItemFromInventory`, which calls `RemoveTradeableItem` for you, and `DestroyItem` clears and removes as well (`PlayerStorage.cpp:3121`), with both `DestroyItemCount` overloads routed through it. `UpdateSoulboundTradeItems` also erases entries whose `GetOwnerGUID() != GetGUID()` before it ever calls this, so a reassigned owner guid can't reach the deref. that leaves one shape, the updated player itself not being in world, since `GetOwner()` is `FindPlayer` and that returns null on `!IsInWorld()`. both `Player::Update` call sites gate on `IsInWorld()` (`Map.cpp:443` and `:484`) and the Playerbot branch doesn't change that, so I can't get there from either tree. can you post the crashlog behind `58e3316c`? happy to be wrong about the mechanism, but right now the description asserts one I can't find.

sogladev (approved):

Celandriel:
https://github.com/mod-playerbots/azerothcore-wotlk/pull/127

Celandriel:
Set to false.

sudlud (approved):

PR #27533, opened 2026-09-07
feat(Core/Session): support client-less (headless) sessions

Celandriel:
Playerbots and other headless cases need sessions without a client socket. This adds:

- an IsHheadless check for socketless conections to WorldSession. 
- NextQueuedPacket for better ownership of packet-queue accessor

- `LoginQueryHolder` moves to `WorldSession.h` so a module can build and queue the holder itself.
- `HandlePlayerLoginFromDB` is reachable by modules.
- `CharacterCreateInfo` gets an explicit constructor taking the character fields.

- Guild: `HandleSetEmblem(EmblemInfo const&)`, `HandleSetRankInfo(rankId, name, rights, moneyPerDay)` taking optionals in the same order as the session overload and returning before the broadcast when all of them are empty, plus read accessors `HasRankRight`, `GetRankRights` and `MemberHasTabRights`. `EmblemInfo` gets an explicit valued constructor.
- ArenaTeam: `SetEmblem` persists through a new `CHAR_UPD_ARENA_TEAM_EMBLEM` statement, mirroring how `SetCaptain` and `SetName` write theirs, and sits behind the same `CanSaveToDB` check as `SaveToDB` so temp teams stay out of the database. `SetRatingForAll` sets the team and member ratings and saves with `forceMemberSave`, because members without games this week are otherwise skipped.
- `GraveyardStruct const* GetClosestGraveyard(mapId, x, y, z, teamId, areaId, zoneId, isDeathKnight)`, a coordinate-based overload of the existing lookup.
- Movement: `MotionMaster::MovePointBackwards`, and an `allowClientControlled` parameter on `MoveKnockbackFrom`, default `false`. `PointMovementGenerator` takes a `reverseOrientation` flag to support the former.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Opus/Fable use is just in chopping Playerbot core up into manageable pieces and responding to reviews. 

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


src/server/game/Battlegrounds/ArenaTeam.cpp
```diff
@@ -1111,3 +1111,22 @@ std::unordered_map<uint8, uint8> ArenaTeam::ArenaReqPlayersForType =
...
+}
+
+void ArenaTeam::SetRatingForAll(uint32 rating)
+{
+    Stats.Rating = rating;
+
+    for (MemberList::iterator itr = Members.begin(); itr != Members.end(); ++itr)
+    {
+        itr->PersonalRating = rating;
+    }
```
sudlud: no braces around single line for / if statements please

src/server/game/Guilds/Guild.cpp
```diff
@@ -1408,6 +1408,29 @@ void Guild::HandleSetRankInfo(WorldSession* session, uint8 rankId, std::string_v
...
 }
 
+void Guild::HandleSetRankInfo(uint8 rankId, uint32 rights, std::string_view name, uint32 moneyPerDay)
+{
+    if (RankInfo* rankInfo = GetRankInfo(rankId))
+    {
+        if (!name.empty())
+        {
+            rankInfo->SetName(name);
+        }
```
sudlud: same here and further below

src/server/game/Movement/MotionMaster.cpp
```diff
@@ -955,6 +955,45 @@ void MotionMaster::MoveRotate(uint32 time, RotateDirection direction)
...
 
+void MotionMaster::MoveKnockbackFromForPlayer(float srcX, float srcY, float speedXY, float speedZ)
+{
+    if (speedXY <= 0.1f)
+        return;
+
+    Position dest = _owner->GetPosition();
+    float moveTimeHalf = speedZ / Movement::gravity;
+    float dist = 2 * moveTimeHalf * speedXY;
+    float max_height = -Movement::computeFallElevation(moveTimeHalf, false, -speedZ);
```
sudlud: can make these float const if possible

src/server/game/Server/WorldSession.h
```diff
@@ -30,6 +30,7 @@
 #include "DatabaseEnv.h"
 #include "Duration.h"
 #include "GossipDef.h"
+#include "QueryHolder.h"
 #include "Packet.h"
```
sudlud: alphabetical include ordering please

src/server/game/Server/WorldSession.h
```diff
@@ -290,6 +290,20 @@ enum CharterTypes
     ARENA_TEAM_CHARTER_5v5_TYPE                   = 5
 };
 
+class LoginQueryHolder : public CharacterDatabaseQueryHolder
+{
+    private:
+        uint32 m_accountId;
+        ObjectGuid m_guid;
```
sudlud: please check AC codestyle:
https://www.azerothcore.org/wiki/cpp-code-standards

`m_` prefix must be removed

src/server/game/World/World.cpp
```diff
@@ -69,6 +69,7 @@
 #include "ObjectMgr.h"
 #include "Opcodes.h"
 #include "OutdoorPvPMgr.h"
+#include "QueryHolder.h"
 #include "PetitionMgr.h"
```
sudlud: alphabetical include ordering

sudlud:
thanks for the fast turnaround on the last round, build's green and the graveyard hook looks good.

one thing we try to keep out of core is naming that points at a specific module, so could we rename `_isBot` / `IsBot()`? what core actually cares about here is "this session has no client socket", and the ctor already knows that from `sock`, so maybe we can derive it and skip the parameter:

```cpp
// WorldSession.h
[[nodiscard]] bool IsHeadless() const { return _headless; }
...
bool const _headless;
```

```cpp
// WorldSession.cpp, ctor init list
    _headless(!sock)
```

then the new ctor parameter can go away, and `else if (isBot) m_Address = "bot";` turns into a plain `else` with something neutral like `"headless"`, or nothing at all since it only feeds log lines.

nice side effect: `IntegrationTestFixture.h:101`, `GmVisibleCommandTest.cpp:91` and `OmenOfClarityGlyphLockTest.cpp:126` already build sessions with a null socket and no `isBot`, so they'd pick up the same guard for free.

while you're in the destructor anyway, should the `totaltime` write get the same treatment? only `online` sits behind the check right now.

`GetPacketQueue()` might not be needed either, `WorldSession::QueuePacket` is already public and does the same `_recvQueue.add`. wdyt?

and would you mind dropping the two "Public accessors for the playerbots module" comments in `Guild.h` and `Guild.cpp`? the three accessors read fine without them.

last one, `MoveKnockbackFromForPlayer` is quite similar to `MoveKnockbackFrom`. maybe a parameter on the existing one would do:

```cpp
void MoveKnockbackFrom(float srcX, float srcY, float speedXY, float speedZ, bool allowClientControlled = false);
```

that'd also keep the fleeing/confused guard from ac071f8f5d.

sudlud:
just wondering about the `temp` guid in `Item::CreateItem`: is a temp item ever meant to reach a save path?

if one ever does, `Item::SaveToDB` writes `GetGUID().GetCounter()` straight into `item_instance` with no guard, so the row lands at guid 4294967295. next startup `SetHighestGuids` runs `SELECT MAX(guid) FROM item_instance` and `Set(... + 1)`, which wraps to 0 on uint32, and the four cleanup queries right below it turn into:

```
DELETE FROM character_inventory WHERE item      >= '0'
DELETE FROM mail_items          WHERE item_guid >= '0'
DELETE FROM auctionhouse        WHERE itemguid  >= '0'
DELETE FROM guild_bank_item     WHERE item_guid >= '0'
```

so every inventory, mail attachment, auction and guild bank on the realm, and it wouldn't surface until the restart after the item was saved.

nothing in core passes `temp = true` today so it can't fire from here, but the parameter only exists for callers outside the tree, which is what makes shipping the sentinel as is a bit scary. could we leave `temp` out of this PR for now, or have `SaveToDB` refuse a sentinel guid, or reserve a range that `SetHighestGuids` skips? whichever fits your use case best.

sudlud:
rest of my notes on this one. first one isn't a nit, the others are.

the part I'm still unsure about: how does a headless session get its packets dispatched? `WorldSession::Update` only drains `_recvQueue` while `m_Socket` is set, and returns false right below that when it's null, so `WorldSessionMgr::UpdateSessions` erases and deletes the session on the next tick. does the module keep its sessions out of `_sessions` and pump them itself?

asking because it also decides whether `World` is the right home for `AddQueryHolderCallback`. `WorldSession` already has one, and the session-owned processor drops its pending callbacks when the session is destroyed, while the `World` one outlives every session. so a login holder whose `AfterComplete` captures the session would fire on freed memory once that session is gone.

`ArenaTeam::SetEmblem` sets the fields but nothing persists them. `SaveToDB` only writes `CHAR_UPD_ARENA_TEAM_STATS` and `CHAR_UPD_ARENA_TEAM_MEMBER`, and the emblem columns are only ever written by `CHAR_INS_ARENA_TEAM` in `Create`, so it's back to the old values after a restart. is that intended?

in the new `HandleSetRankInfo`, `_BroadcastEvent(GE_RANK_UPDATED, ...)` still runs when all three optionals are empty. maybe put it behind a check that something actually changed?

`CharacterCreateInfo`'s new ctor defaults `gender` to 0, which overrides the `Gender = GENDER_NONE` initialiser on the member. nothing in core notices since `HandleCharCreateOpcode` fills it from the packet right after, but `gender = GENDER_NONE` would keep the sentinel intact.

`CHAR_SEL_CHARS_BY_ACCOUNT_ID` picked up `class` and `race`, and the only core caller, `AccountMgr::DeleteAccount`, reads column 0. could that be a separate statement so account deletion doesn't pull columns it never uses?

`EmblemInfo` and `CharacterCreateInfo` are both implicit converting ctors now, `EmblemInfo e = 5u;` compiles. worth marking them `explicit`?

and could you tick the Core box and fill in the test boxes in the description? plus the AC agentic engineering guidelines one, since the AI box is checked.

Celandriel (commented):
Let me work through the comments and test it. It could take a bit.

Celandriel (commented):
> the part I'm still unsure about: how does a headless session get its packets dispatched? `WorldSession::Update` only drains `_recvQueue` while `m_Socket` is set, and returns false right below that when it's null, so `WorldSessionMgr::UpdateSessions` erases and deletes the session on the next tick. does the module keep its sessions out of `_sessions` and pump them itself?

So the module drains the packets every tick in `void PlayerbotHolder::HandleBotPackets(WorldSession* session)`. I made the changes changes based on some of the reviews and changed GetPacketQueue to NextQueuedPacket, with unique_ptr it should be a little bit more safe to run. 

> asking because it also decides whether `World` is the right home for `AddQueryHolderCallback`. `WorldSession` already has one, and the session-owned processor drops its pending callbacks when the session is destroyed, while the `World` one outlives every session. so a login holder whose `AfterComplete` captures the session would fire on freed memory once that session is gone.

Here I dont really know. 

> `ArenaTeam::SetEmblem` sets the fields but nothing persists them. `SaveToDB` only writes `CHAR_UPD_ARENA_TEAM_STATS` and `CHAR_UPD_ARENA_TEAM_MEMBER`, and the emblem columns are only ever written by `CHAR_INS_ARENA_TEAM` in `Create`, so it's back to the old values after a restart. is that intended?
>
> in the new `HandleSetRankInfo`, `_BroadcastEvent(GE_RANK_UPDATED, ...)` still runs when all three optionals are empty. maybe put it behind a check that something actually changed?
> 
> `CharacterCreateInfo`'s new ctor defaults `gender` to 0, which overrides the `Gender = GENDER_NONE` initialiser on the member. nothing in core notices since `HandleCharCreateOpcode` fills it from the packet right after, but `gender = GENDER_NONE` would keep the sentinel intact.
> 
> `CHAR_SEL_CHARS_BY_ACCOUNT_ID` picked up `class` and `race`, and the only core caller, `AccountMgr::DeleteAccount`, reads column 0. could that be a separate statement so account deletion doesn't pull columns it never uses?
> 
> `EmblemInfo` and `CharacterCreateInfo` are both implicit converting ctors now, `EmblemInfo e = 5u;` compiles. worth marking them `explicit`?
> 
> and could you tick the Core box and fill in the test boxes in the description? plus the AC agentic engineering guidelines one, since the AI box is checked.

Resolved.

Celandriel (commented):
> one thing we try to keep out of core is naming that points at a specific module, so could we rename _isBot / IsBot()? what core actually cares about here is "this session has no client socket", and the ctor already knows that from sock, so maybe we can derive it and skip the parameter: 
> ...
then the new ctor parameter can go away, and else if (isBot) m_Address = "bot"; turns into a plain else with something neutral like "headless", or nothing at all since it only feeds log lines.
>
> nice side effect: IntegrationTestFixture.h:101, GmVisibleCommandTest.cpp:91 and OmenOfClarityGlyphLockTest.cpp:126 already build sessions with a null socket and no isBot, so they'd pick up the same guard for free.
> while you're in the destructor anyway, should the totaltime write get the same treatment? only online sits behind the check right now.

Renamed to headless, and implemented as suggested. 

> GetPacketQueue() might not be needed either, WorldSession::QueuePacket is already public and does the same _recvQueue.add. wdyt?

See other comment. 

> and would you mind dropping the two "Public accessors for the playerbots module" comments in Guild.h and Guild.cpp? the three accessors read fine without them.
>
> last one, MoveKnockbackFromForPlayer is quite similar to MoveKnockbackFrom. maybe a parameter on the existing one would do:
>
>just wondering about the temp guid in Item::CreateItem: is a temp item ever meant to reach a save path? ...

Resolved all of these, Removed comments, merged the knockbacks, and restored the temp CreateItem.

src/server/game/Guilds/Guild.cpp
```diff
@@ -1330,15 +1330,20 @@ void Guild::HandleSetEmblem(WorldSession* session, EmblemInfo const& emblemInfo)
...
-        m_emblemInfo.SaveToDB(m_id);
+        HandleSetEmblem(emblemInfo);
 
         SendSaveEmblemResult(session, ERR_GUILDEMBLEM_SUCCESS); // "Guild Emblem saved."
 
         HandleQuery(session);
     }
 }
 
+void Guild::HandleSetEmblem(EmblemInfo const& emblemInfo)
```
Celandriel, line 1341: @sudlud I realized this override was missing previously.

sudlud:
thanks for working through all of that, the rename, the emblem persistence, the optionals and the item temp revert all look good.

still open from last round: `WorldSession::Update` returns false whenever `m_Socket` is null (`WorldSession.cpp:616`) and `WorldSessionMgr` then deletes the session, with no `_headless` check anywhere on that path. does the module keep its sessions out of `_sessions` and pump them itself?

asking because it decides the home for `AddQueryHolderCallback`: `World::_queryHolderProcessor` outlives every session, so a login holder capturing the session fires on freed memory once it's gone. `WorldSession::AddQueryHolderCallback` is already public, only `ProcessQueryCallbacks()` isn't, so maybe make that public and drop the `World` copy?

`ArenaTeam::SetEmblem` executes `CHAR_UPD_ARENA_TEAM_EMBLEM` unconditionally, but every other write in the class goes through `SaveToDB`, which returns early on `!sScriptMgr->CanSaveToDB(this)` (`:944`). that's the hook that keeps temp teams from `CreateTempArenaTeam` out of the db. same check here?

`SetRatingForAll` only sets the fields, and `SaveToDB()` skips members with `WeekGames == 0` unless `forceMemberSave` (`:965`), so a rating set on a team that hasn't played this week is lost. save inside, or note it on the method?

`CHAR_UPD_CHAR_OFFLINE` has no callers. module using it, or left over?

could you refresh the description? it still has `is_bot`, `GetPacketQueue`, the `Item::CreateItem` temp mode and `MoveKnockbackFromForPlayer`, and is missing the `HandleSetEmblem` overload and the `CharacterCreateInfo` ctor. we squash with it.

nits:

`m_playerLoading = true` in `HandlePlayerLoginFromDB` (`CharacterHandler.cpp:798`) is already set at `:789`.

new `HandleSetRankInfo` takes `rights` before `name`, the session one the other way round. match it?

Celandriel (commented):
> still open from last round: `WorldSession::Update` returns false whenever `m_Socket` is null (`WorldSession.cpp:616`) and `WorldSessionMgr` then deletes the session, with no `_headless` check anywhere on that path. does the module keep its sessions out of `_sessions` and pump them itself?
> 
> asking because it decides the home for `AddQueryHolderCallback`: `World::_queryHolderProcessor` outlives every session, so a login holder capturing the session fires on freed memory once it's gone. `WorldSession::AddQueryHolderCallback` is already public, only `ProcessQueryCallbacks()` isn't, so maybe make that public and drop the `World` copy?

So this requires a bit of a change how the module was handling sessions (it basically let it live in the player object and referenced as needed). But I changed it as suggested and it appears to still work with the simpler approach. 

> `ArenaTeam::SetEmblem` executes `CHAR_UPD_ARENA_TEAM_EMBLEM` unconditionally, but every other write in the class goes through `SaveToDB`, which returns early on `!sScriptMgr->CanSaveToDB(this)` (`:944`). that's the hook that keeps temp teams from `CreateTempArenaTeam` out of the db. same check here?
>
> `SetRatingForAll` only sets the fields, and `SaveToDB()` skips members with `WeekGames == 0` unless `forceMemberSave` (`:965`), so a rating set on a team that hasn't played this week is lost. save inside, or note it on the method?
> 
> `CHAR_UPD_CHAR_OFFLINE` has no callers. module using it, or left over?
>
> new `HandleSetRankInfo` takes `rights` before `name`, the session one the other way round. match it?

Added arena check, added save on set rating (with forced update) reorganized the calls for guild, and removed CHAR_UPD_CHAR_OFFLINE
 
> could you refresh the description? it still has `is_bot`, `GetPacketQueue`, the `Item::CreateItem` temp mode and `MoveKnockbackFromForPlayer`, and is missing the `HandleSetEmblem` overload and the `CharacterCreateInfo` ctor. we squash with it.
> 
> nits:
> 
> `m_playerLoading = true` in `HandlePlayerLoginFromDB` (`CharacterHandler.cpp:798`) is already set at `:789`.

Here we call HandlePlayerLoginFromDB directly bypassing HandlePlayerLoginOpcode. I think we need to keep both.

sudlud (approved):

PR #27544, opened 2026-09-07
feat(Core/Database): enable modules to own their database

Celandriel:
Revives #24525.

Modules can now bring a complete database of their own without core-side registration:

- ModuleDatabasePool: a non-templated pool a module subclasses with its own MySQLConnection type, carrying prepared statements, synchronous query/execute, transactions and keep-alive (revives PR #24525 with all review items addressed; async execution is deliberately deferred)
- DatabaseUpdatePool + DatabaseWorkerPoolAdapter: the DBUpdater now operates on an interface, so core pools and module pools share it
- ModuleDBUpdater: the DBUpdater<T> bodies refactored into shared implementations with a non-template entry point taking runtime metadata, so modules reuse create/populate/update instead of duplicating the mysql-CLI apply logic
- DatabaseScript lifecycle hooks (OnModuleDatabasesLoading/KeepAlive/Closing and OnDatabaseGetDBRevision) called from the worldserver's DB startup/shutdown/keep-alive path; .server info reports module DB revisions through the same hook

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Fable: Resolve comments in original PR. 

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


src/server/database/Updater/DBUpdater.h
```diff
@@ -79,23 +79,38 @@ class AC_DATABASE_API DBUpdater
...
     static bool IsEnabled(uint32 const updateMask);
     static BaseLocation GetBaseLocationType();
-    static bool Create(DatabaseWorkerPool<T>& pool);
-    static bool Update(DatabaseWorkerPool<T>& pool, std::string_view modulesList = {});
-    static bool Update(DatabaseWorkerPool<T>& pool, std::vector<std::string> const* setDirectories);
-    static bool Populate(DatabaseWorkerPool<T>& pool);
+    static bool Create(DatabaseUpdatePool& pool);
+    static bool Update(DatabaseUpdatePool& pool, std::string_view modulesList = {});
+    static bool Update(DatabaseUpdatePool& pool, std::vector<std::string> const* setDirectories);
+    static bool Populate(DatabaseUpdatePool& pool);
```
sudlud: Changing `Create/Update/Populate` from `DatabaseWorkerPool<T>&` to `DatabaseUpdatePool&` breaks every existing module that calls them, and `ubuntu-24.04-clang-18-nopch-modules` is red on it:

```
modules/mod-progression-system/src/ProgressionSystem.cpp:48:17: fatal error: no matching function for call to 'Update'
```

Is the break actually needed? `DatabaseWorkerPoolAdapter<T>` is right there, so we could keep the three `DatabaseWorkerPool<T>&` overloads and have them build the adapter and forward. `DBUpdater<T>` is only ever instantiated for the three core connection types anyway.

If you'd rather keep the break, can you prepare the mod-progression-system PR alongside this one? I don't see one open yet.

src/server/database/Database/ModuleDatabasePool.cpp
```diff
@@ -0,0 +1,232 @@
...
+                i + 1, _synchThreads, _connectionInfo.database, result);
+            Close();
+            return result;
+        }
+
+        _connections.push_back(std::move(conn));
+    }
+
+    return 0;
+}
```
sudlud, line 70: `Open()` returns 0 (success) when `_synchThreads` is 0, and then every call hits an `_connections.empty()` guard and quietly does nothing.

That gets nasty with `ModuleDBUpdater::Update`. `UpdateFetcher` applies the actual .sql files through the mysql CLI so those land, but the `INSERT INTO updates` bookkeeping goes through `pool.DirectExecute` and disappears, and the `SHOW TABLES LIKE 'updates'` probe returns null. So every start re-applies the module's whole update history against a live db, with nothing logged.

Can we reject 0 in `SetConnectionInfo()` or `Open()`?

src/server/game/Server/WorldSession.cpp
```diff
@@ -854,8 +858,8 @@ void WorldSession::LogoutPlayer(bool save, bool redirecting)
         //! Since each account can only have one online character at any given time, ensure all characters for active account are marked as offline
         if (!redirecting)
         {
-            CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CHAR_UPD_ACCOUNT_ONLINE);
-            stmt->SetData(0, GetAccountId());
+            CharacterDatabasePreparedStatement* stmt = CharacterDatabase.GetPreparedStatement(CharacterDatabaseStatements(statementIndex));
```
sudlud: This takes whatever index the hook wrote, and `GetPreparedStatement` is `new PreparedStatement<T>(index, _preparedStatementSize[index])` with an unchecked `operator[]`. A module returning an out-of-range index is an OOB read, and a zero-param statement trips the `SetData` assert. Startup stays clean, it blows up at logout.

src/server/game/Server/WorldSession.cpp
```diff
@@ -833,6 +833,10 @@ void WorldSession::LogoutPlayer(bool save, bool redirecting)
         LOG_INFO("entities.player", "Account: {} (IP: {}) Logout Character:[{}] ({}) Level: {}",
             GetAccountId(), GetRemoteAddress(), _player->GetName(), _player->GetGUID().ToString(), _player->GetLevel());
 
+        uint32 statementIndex = CHAR_UPD_ACCOUNT_ONLINE;
+        uint32 statementParam = GetAccountId();
+        sScriptMgr->OnDatabaseSelectIndexLogout(_player, statementIndex, statementParam);
```
sudlud: This fires even when `redirecting` is true, where the statement below is never used. Also only one `uint32` param is expressible, so any replacement statement is stuck with that shape.

Stepping back though, letting a module swap the core's "mark this account's characters offline" write has nothing to do with a module owning its own database. Can we split it into its own PR?

src/server/game/Scripting/ScriptDefines/DatabaseScript.cpp
```diff
@@ -19,6 +19,21 @@
...
+        return !script->OnDatabasesLoading();
+    });
+
+    if (ret && *ret)
+    {
+        return false;
+    }
+
+    return true;
+}
```
sudlud: The six new hooks go through `ExecuteScript<>` / `IsValidBoolScript<>` and add no `DATABASEHOOK_*` entries, so they ignore `ScriptRegistry::EnabledHooks` entirely. A module that opts into a subset via the `enabledHooks` ctor param still gets all six. No hook should be addable that can't be disabled properly, so these need the enum entries plus `CALL_ENABLED_HOOKS`, same as everything that landed this week (#27552, #27447, #27018).

`OnDatabasesLoading` also reimplements `ReturnValidBool(ret)`. `CALL_ENABLED_BOOLEAN_HOOKS` covers the whole function, and that also takes care of the braces around the single-line `if` at `:29`.

Separate thing, the name: it runs after `loader.Load()` and sits about 40 lines above the existing `OnAfterDatabasesLoaded`. Two near-identical names at two near-identical points will confuse module authors, worth renaming while it's cheap.

src/server/apps/worldserver/Main.cpp
```diff
@@ -446,6 +446,11 @@ bool StartDB()
     if (!loader.Load())
         return false;
 
+    if (!sScriptMgr->OnDatabasesLoading())
```
sudlud: Braces around a single-line `if` body, `.agents/docs/cpp-guidelines.md` says no braces there. It also sits three lines under an unbraced `if (!loader.Load())`.

src/server/database/Updater/DBUpdater.cpp
```diff
@@ -220,7 +248,7 @@ bool DBUpdater<T>::Create(DatabaseWorkerPool<T>& pool)
 
     try
     {
-        DBUpdater<T>::ApplyFile(pool, pool.GetConnectionInfo()->host, pool.GetConnectionInfo()->user, pool.GetConnectionInfo()->password,
+        ApplyFile(pool, pool.GetConnectionInfo()->host, pool.GetConnectionInfo()->user, pool.GetConnectionInfo()->password,
```
sudlud: The continuation line kept the old member-function indent, so the wrapped params are misaligned now. Same at `:485` and `:489`.

src/server/database/Updater/DBUpdater.cpp
```diff
@@ -255,16 +282,16 @@ bool DBUpdater<T>::Update(DatabaseWorkerPool<T>& pool, std::string_view modulesL
...
     auto CheckUpdateTable = [&](std::string const& tableName)
     {
-        auto checkTable = DBUpdater<T>::Retrieve(pool, Acore::StringFormat("SHOW TABLES LIKE '{}'", tableName));
+        auto checkTable = Retrieve(pool, Acore::StringFormat("SHOW TABLES LIKE '{}'", tableName));
         if (!checkTable)
         {
             LOG_WARN("sql.updates", "> Table '{}' not exist! Try add based table", tableName);
 
-            Path const temp(GetBaseFilesDirectory() + tableName + ".sql");
+            Path const temp(info.baseFilesDirectory + tableName + ".sql");
```
sudlud: nit: this builds the path with `info.baseFilesDirectory + tableName + ".sql"`, so the field has to end in a separator, but `:414` uses the same field as a plain directory. Core is fine since `GetBaseFilesDirectory()` returns a trailing slash, but the doc comment at `DBUpdater.h:101` doesn't say so. `Path(info.baseFilesDirectory) / (tableName + ".sql")` works for both. Same at `:359`.

src/server/database/Database/ModuleDatabasePool.h
```diff
@@ -0,0 +1,120 @@
...
+    QueryResult Query(std::string_view sql, Args&&... args)
+    {
+        if (sql.empty())
+            return QueryResult(nullptr);
+
+        return Query(std::string_view(Acore::StringFormat(sql, std::forward<Args>(args)...)));
+    }
+
+    //! Prepared statements. The index space is defined by the module's connection
+    //! class (DoPrepareStatements); parameter counts are recorded during Open().
```
sudlud: nit: param counts are recorded in `PrepareStatements()`, not `Open()`. A module author following this gets 0 back from `GetPreparedStatementParamCount` (`ModuleDatabasePool.cpp:173` returns 0 for an unpopulated vector, no assert), builds a zero-param statement and asserts on first use.

`Close()` also leaves `_preparedStatementSize` filled, so a close/reopen reports stale counts.

src/server/database/Database/ModuleDatabasePool.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+bool ModuleDatabasePool::PrepareStatements()
```
sudlud, line 72: nit: `DatabaseWorkerPool<T>::PrepareStatements` does `LockIfReady()`/`Unlock()` around this and calls `Close()` before returning false. Without that a failed prepare leaves a half-prepared open pool.

src/server/database/Database/ModuleDatabasePool.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    return PreparedQueryResult(result);
+}
+
+uint8 ModuleDatabasePool::GetPreparedStatementParamCount(uint32 index) const
+{
+    return index < _preparedStatementSize.size() ? _preparedStatementSize[index] : 0;
+}
+
+void ModuleDatabasePool::DirectCommitTransaction(std::shared_ptr<TransactionBase> transaction)
```
sudlud, line 193: nit: core calls `transaction->Cleanup()` on the failure path here.

src/server/database/Database/DatabaseWorkerPoolAdapter.h
```diff
@@ -0,0 +1,54 @@
...
+#define DATABASE_WORKER_POOL_ADAPTER_H
+
+#include "DatabaseUpdatePool.h"
+#include "DatabaseWorkerPool.h"
+
+template <class T>
+class DatabaseWorkerPoolAdapter : public DatabaseUpdatePool
+{
+public:
+    DatabaseWorkerPoolAdapter(DatabaseWorkerPool<T>& pool) : _pool(pool) {}
```
sudlud, line 28: nit: this forwards to the async `Execute`, while `ModuleDatabasePool::Execute` is synchronous, and nothing in `UpdaterImpl` calls either. Maybe just remove `Execute` from the interface?

src/server/database/Updater/DBUpdater.cpp
```diff
@@ -186,8 +204,18 @@ BaseLocation DBUpdater<T>::GetBaseLocationType()
     return LOCATION_REPOSITORY;
 }
 
-template<class T>
-bool DBUpdater<T>::Create(DatabaseWorkerPool<T>& pool)
+namespace UpdaterImpl
```
sudlud: nit: these are namespace-scope non-static with no header declaring them, can they go in an anonymous namespace?

src/server/scripts/Commands/cs_server.cpp
```diff
@@ -216,6 +217,11 @@ class server_commandscript : public CommandScript
 
         handler->PSendSysMessage("Using World DB: {}", sWorld->GetDBVersion());
 
+        std::string moduleDBRevision;
```
sudlud: nit: one `std::string&` goes to every script, so two modules with their own db overwrite each other and the last one wins.

sudlud (dismissed):

src/server/database/Database/ModuleDatabasePool.cpp
Celandriel: Add guard against no threads.

Celandriel (commented):
Addressed review comments.

sudlud:
five things left, none of them blocking. left the coderabbit spin-loop thread open, `DatabaseWorkerPool::GetFreeConnection` does the same thing without even the yield, so that's core's problem and not this PR's.

`ModuleDatabasePool::PrepareStatements()`: a module starting from `WorldDatabaseConnection::DoPrepareStatements` gets `CONNECTION_ASYNC` on most of its statements, and `MySQLConnection::PrepareStatement` resets those to null on a sync connection. startup stays clean and the first use hits `ASSERT(m_mStmt)`. can we say so in the class comment, or log the indices that came back null?

`ModuleDatabasePool.cpp:121` `DirectExecute`: core returns early on `sql.empty()`, this one hands it to `mysql_query("")` and logs `[1065] Query was empty`. the formatting overloads already guard it, so `Execute(std::string{})` and `Execute("{}", "")` behave differently.

`OnDatabaseWarnAboutSyncQueries`: what's a module meant to hang off this? the pool is sync-only, so anything wired to it warns on every query. maybe leave it out until async lands?

the description still lists `OnDatabaseSelectIndexLogout` and `OnDatabasesLoading`, and that text becomes the squash commit message. worth a quick edit before this goes in.

nit: `ModuleDatabasePool.h:49` says `Open()` returns the mysql error code, the zero-threads path returns `1`.

sudlud (approved):

PR #27563, opened 2026-09-09
fix(Core/Unit): clear moving flags before sending root

espkk:
## Changes Proposed:

This PR proposes changes to:

For a unit that is not rooted, the client consumes the elapsed delta and derives the displacement from it. `MOVEMENTFLAG_ROOT` (0x800) sends the step down a fallback path that reports zero consumed time, while the loop driving that step only terminates once the consumed time reaches the step target. So `MOVEMENTFLAG_ROOT` combined with any bit of `0xC0100F` (the moving bits, turn and pitch excluded, as those never reach that step) gives a loop that can never make progress and never exits. Every client that receives such movement info for any unit hangs, not just the client of the affected unit.

The guard lived in `Unit::SetRooted()` since AzerothCore's first commit. 
#23147 moved the in-place flag application into the new Unit::SendMoveRoot() and, in the move, replaced `MOVEMENTFLAG_MASK_MOVING` with `MOVEMENTFLAG_MASK_MOVING_FLY`.

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Entities/Unit/Unit.cpp
+++ b/src/server/game/Entities/Unit/Unit.cpp
@@ -14500,7 +14500,13 @@ void Unit::SendMoveRoot(bool apply)
     {
         if (apply)
         {
-            m_movementInfo.RemoveMovementFlag(MOVEMENTFLAG_MASK_MOVING_FLY);
+            // MOVEMENTFLAG_ROOT must not be sent alongside MOVEMENTFLAG_MASK_MOVING.
+            // The 3.3.5a client's movement step returns zero consumed time as soon
+            // as it sees ROOT, while the loop driving it only terminates once that
+            // time reaches the step target - so a moving flag alongside ROOT spins
+            // forever and freezes every client that receives this unit's movement
+            // info. Turn and pitch bits are inert here, they never reach that step.
+            m_movementInfo.RemoveMovementFlag(MOVEMENTFLAG_MASK_MOVING | MOVEMENTFLAG_MASK_MOVING_FLY);
             m_movementInfo.AddMovementFlag(MOVEMENTFLAG_ROOT);
             if (!client)
                 StopMoving();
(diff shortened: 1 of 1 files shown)


sudlud:
@killerwife thoughts

sudlud (approved):

killerwife:
Sendmoveroot should never do this, it should be manipulated in parent func which does game logic. This should be a const only func for network send.

sudlud:
@espkk can you please address killerwife's feedback

espkk:
> Sendmoveroot should never do this, it should be manipulated in parent func which does game logic. This should be a const only func for network send.

but your own code added removal of `MOVEMENTFLAG_MASK_MOVING_FLY` inside `SendMoveRoot`. how is it different from removing `MOVING` mask there?
it also already adds/removes root flag itself. and calls other functions that mutate the state. should those be moved outside too?

espkk:
also looks like mangos does exactly same (+ turn flags) https://github.com/cmangos/mangos-wotlk/blob/2448954ee7611b2f580119f789bfd14116922250/src/game/Entities/Unit.cpp#L741
but from my memory allowing turning is correct, and in my tests those flags were benign in the context of client freezes

espkk:
PS just to clarify, I'm happy to adjust this, just genuinely confused

killerwife:
Correct. I never said its fully conceptualized in cmangos and fully correct. Im trying to.give advice based on acore context, not my or cmangos past mistakes. It should be in cmangos inside SetImmobilizedState just like SetHover and the other ones. And then send move root should just literally do the sending. (And be const otherwise)

espkk:
> Correct. I never said its fully conceptualized in cmangos and fully correct. Im trying to.give advice based on acore context, not my or cmangos past mistakes. It should be in cmangos inside SetImmobilizedState just like SetHover and the other ones. And then send move root should just literally do the sending. (And be const otherwise)

Got it now, thanks!
And it totally makes sense.
But I was trying to make a minimal change that would resolve client crashes that people are experiencing since it was changed, while minimizing the chances to break something else.
Rewriting this is a bit of a pain as it's basically not covered by any kind of tests.
It still definitely makes sense to do, but do you think it's worth merging it like this and attempt to refactor it properly as a follow up?

sudlud:
@espkk sounds like a plan

PR #27578, opened 2026-09-09
feat(Core/DBC): load CharSections and EmotesTextSound, re-add area helpers

Celandriel:
Server-side consumers (modules generating character appearances or reacting to emotes) need CharSections.dbc and EmotesTextSound.dbc loaded, plus FindTextSoundEmoteFor and the area-ID helper family.

## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

## Issues Addressed:

- Closes 

## SOURCE:

Playerbots Core
## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


sudlud:
the area lookup rewrite looks right, thanks.

CI is still red on the two new `LOAD_DBC` lines though. they pass a world DB override table, and `LoadDBC` runs `SELECT * FROM <table>` whenever that argument is set, so the worldserver aborts at startup with `[1146] Table 'acore_world.charsections_dbc' doesn't exist`. can we add a `pending_db_world` update that creates both? something like:

```sql
CREATE TABLE IF NOT EXISTS `charsections_dbc` (
  `ID` int NOT NULL DEFAULT '0',
  `RaceID` int NOT NULL DEFAULT '0',
  `SexID` int NOT NULL DEFAULT '0',
  `BaseSection` int NOT NULL DEFAULT '0',
  `TextureName_1` varchar(100) DEFAULT NULL,
  `TextureName_2` varchar(100) DEFAULT NULL,
  `TextureName_3` varchar(100) DEFAULT NULL,
  `Flags` int NOT NULL DEFAULT '0',
  `VariationIndex` int NOT NULL DEFAULT '0',
  `ColorIndex` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`ID`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

CREATE TABLE IF NOT EXISTS `emotestextsound_dbc` (
  `ID` int NOT NULL DEFAULT '0',
  `EmotesTextID` int NOT NULL DEFAULT '0',
  `RaceID` int NOT NULL DEFAULT '0',
  `SexID` int NOT NULL DEFAULT '0',
  `SoundID` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`ID`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

two things there that bite if they drift: the column count has to match the format string exactly, `x` fields included, and the index column has to be named `ID` because the loader does ``ORDER BY `ID` DESC``. leaving the tables empty is fine, the DBC file still wins.

also `"emotetextsound_dbc"` is missing an s. `EmotesText.dbc` maps to `emotestext_dbc`, so this one should be `emotestextsound_dbc`.

sudlud:
`CharSectionType` names the values of `GenType`, not of the member called `Type`. in the dbc, `GenType` is the 0 to 4 section type and `Type` is the variation index, so `section->Type == SECTION_TYPE_HAIR` would never match. maybe rename `GenType` to `Type`, `Type` to `VariationIndex` and `Color` to `ColorIndex`? that would also line the struct up with the column names in the override table.

`#include "RaceMgr.h"` isn't used in `DBCStores.cpp`, can we drop it? we've had a run of PRs cleaning useless includes out, so it'd be odd to add one back.

last thing, and more of a question: nothing in the core calls any of this yet. which module needs the area helpers? and could `FindTextSoundEmoteFor` be wired into `HandleTextEmoteOpcode` so the emote sounds actually play? easier to keep an api honest when something in tree uses it.

Celandriel (commented):
Sql to add is prepared and I stripped out the unused stuff. 

Right now the general approach is to find the sound and play it directly. 

TC has an example we can implement. 
```
     void HandleScriptEffect(SpellEffIndex effIndex)
     {
         PreventHitDefaultEffect(effIndex);
         if (Unit* target = GetHitUnit())
        {
             target->HandleEmoteCommand(EMOTE_ONESHOT_TRAIN);
            if (EmotesTextSoundEntry const* soundEntry =
                FindTextSoundEmoteFor(TEXT_EMOTE_TRAIN, target->getRace(), target->getGender()))
                target->PlayDistanceSound(soundEntry->SoundId);
        }
     }
```

sudlud:
thanks, the sql and the renames look good, and the area lookup is right now.

two things left.

the cherry-pick credit is still missing the original authors. the `Co-authored-by` trailers on the first commit credit the playerbots fork, but the code is trinitycore's: `CharSections` (fmt, struct, both enums) is arks in 6d4c672fb1, `EmotesTextSound` (fmt, struct, map and `FindTextSoundEmoteFor`) is treeston in 8c7cd03d08, and the `find` + ternary body we have here is Shauren in a2344fba67. no need to rewrite anything, can you add these to the next commit you push?

```
Co-authored-by: arks <ariel.silva305@gmail.com>
Co-authored-by: treeston <treeston.mmoc@gmail.com>
```

`GetAreaEntryByAreaID` is just `sAreaTableStore.LookupEntry(area_id)` and that store is already exported from `DBCStores.h`. can we remove it? on the module side the only call site is in `PlayerbotMgr.cpp`, where `GetAreaEntryByAreaID(3459)` becomes `sAreaTableStore.LookupEntry(3459)`, same as the module already does everywhere else. the two stores need to be in the core since a module can't hook `LoadDBCStores`, but this one we'd carry for nothing. the title would need updating then.

Celandriel (commented):
Fixed

sudlud (approved):

PR #27579, opened 2026-09-09
feat(Core/Misc): Add read-only module accessors

Celandriel:
## Changes Proposed:

This PR proposes changes to:

This enables
- Read-only getters for battleground node/capture-point state
- Const accessors for group target icons and rolls
- Small helpers: ItemTemplate::IsWeapon, static GetFactionReactionTo
- ChannelMgr::GetChannels, ObjectMgr::GetAllAreaTriggerTeleports, public LFGMgr::GetLFGDungeon, ARENA_TYPE_NONE

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used:

## Issues Addressed:

- Closes 

## SOURCE:

Playerbots
## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]

### DIFF
--- a/src/server/game/Battlegrounds/Battleground.h
+++ b/src/server/game/Battlegrounds/Battleground.h
@@ -205,6 +205,7 @@ struct BattlegroundObjectInfo
 
 enum ArenaType : uint8
 {
+    ARENA_TYPE_NONE                 = 0,
     ARENA_TYPE_2v2                  = 2,
     ARENA_TYPE_3v3                  = 3,
     ARENA_TYPE_5v5                  = 5
--- a/src/server/game/Battlegrounds/Zones/BattlegroundAB.h
+++ b/src/server/game/Battlegrounds/Zones/BattlegroundAB.h
@@ -246,6 +246,18 @@ protected:
     uint32 BasesDefended = 0;
 };
 
+struct CaptureABPointInfo
+{
+    CaptureABPointInfo() : _ownerTeamId(TEAM_NEUTRAL), _iconNone(0), _iconCapture(0), _state(BG_AB_NODE_STATE_NEUTRAL), _captured(false) {}
+
+    TeamId _ownerTeamId;
+    uint32 _iconNone;
+    uint32 _iconCapture;
+    uint8  _state;
+
+    bool _captured;
+};
+
 class AC_GAME_API BattlegroundAB : public Battleground
 {
 public:
@@ -270,6 +282,9 @@ public:
     bool IsTeamScores500Disadvantage(TeamId teamId) const { return _teamScores500Disadvantage[teamId]; }
 
     TeamId GetPrematureWinner() override;
+
+    [[nodiscard]] CaptureABPointInfo const& GetCapturePointInfo(uint32 node) const { return _capturePointInfo[node]; }
+
 private:
     void PostUpdateImpl(uint32 diff) override;
 
@@ -279,21 +294,7 @@ private:
     void NodeOccupied(uint8 node);
     void NodeDeoccupied(uint8 node);
 
-    struct CapturePointInfo
-    {
-        CapturePointInfo() : _ownerTeamId(TEAM_NEUTRAL), _iconNone(0), _iconCapture(0), _state(BG_AB_NODE_STATE_NEUTRAL), _captured(false)
-        {
-        }
-
-        TeamId _ownerTeamId;
-        uint32 _iconNone;
-        uint32 _iconCapture;
-        uint8 _state;
-
-        bool _captured;
-    };
-
-    CapturePointInfo _capturePointInfo[BG_AB_DYNAMIC_NODES_COUNT];
+    CaptureABPointInfo _capturePointInfo[BG_AB_DYNAMIC_NODES_COUNT];
     EventMap _bgEvents;
     uint32 _honorTics;
     uint32 _reputationTics;
--- a/src/server/game/Battlegrounds/Zones/BattlegroundEY.h
+++ b/src/server/game/Battlegrounds/Zones/BattlegroundEY.h
@@ -346,6 +346,25 @@ protected:
     uint32 FlagCaptures = 0;
 };
 
+struct CaptureEYPointInfo
+{
+    CaptureEYPointInfo() : _ownerTeamId(TEAM_NEUTRAL), _barStatus(BG_EY_PROGRESS_BAR_STATE_MIDDLE), _areaTrigger(0)
+    {
+        _playersCount[TEAM_ALLIANCE] = 0;
+        _playersCount[TEAM_HORDE] = 0;
+    }
+
+    Player* player = nullptr;
+    TeamId _ownerTeamId;
+    int8 _barStatus;
+    uint32 _areaTrigger;
+    int8 _playersCount[PVP_TEAMS_COUNT];
+
+    bool IsUnderControl(TeamId teamId) const { return _ownerTeamId == teamId; }
+    bool IsUnderControl() const { return _ownerTeamId != TEAM_NEUTRAL; }
+    bool IsUncontrolled() const { return _ownerTeamId == TEAM_NEUTRAL; }
+};
+
 class AC_GAME_API BattlegroundEY : public Battleground
 {
 public:
@@ -385,6 +404,8 @@ public:
     bool AllNodesConrolledByTeam(TeamId teamId) const override;
     TeamId GetPrematureWinner() override;
 
+    [[nodiscard]] CaptureEYPointInfo const& GetCapturePointInfo(uint32 node) const { return _capturePointInfo[node]; }
+
 private:
     void PostUpdateImpl(uint32 diff) override;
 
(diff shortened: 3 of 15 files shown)


src/server/game/Battlegrounds/Zones/BattlegroundEY.h
```diff
@@ -346,6 +346,25 @@ struct BattlegroundEYScore final : public BattlegroundScore
...
 
+struct CaptureEYPointInfo
+{
+    CaptureEYPointInfo() : _ownerTeamId(TEAM_NEUTRAL), _barStatus(BG_EY_PROGRESS_BAR_STATE_MIDDLE), _areaTrigger(0)
+    {
+        _playersCount[TEAM_ALLIANCE] = 0;
+        _playersCount[TEAM_HORDE] = 0;
+    }
+
+    Player* player = nullptr;
```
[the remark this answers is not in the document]
Celandriel: Resolved to pointer.

src/server/game/Groups/Group.h
```diff
@@ -311,6 +312,8 @@ class Group
     void EndRoll(Loot* loot);
     void RemovePlayerFromRolls(ObjectGuid guid);
 
+    std::vector<Roll const*> GetRolls() const { return { RollId.begin(), RollId.end() }; }
```
sudlud, line 316: the `const` sorts the mutation side, thanks. one thing left: we delete these `Roll` objects ourselves (`Group.cpp:1529` and `1747`), so the snapshot is only good for the current call. could we note that on the method? anything caching it across ticks would be sitting on freed pointers.

src/server/game/Groups/Group.h
```diff
@@ -270,6 +270,7 @@ class Group
     void SetTargetIcon(uint8 id, ObjectGuid whoGuid, ObjectGuid targetGuid);
     void SetGroupMemberFlag(ObjectGuid guid, bool apply, GroupMemberFlags flag);
     void RemoveUniqueGroupMemberFlag(GroupMemberFlags flag);
+    ObjectGuid const GetTargetIcon(uint8 id) const { return m_targetIcons[id]; }
```
sudlud: `SetTargetIcon` guards this with `if (id >= TARGETICONCOUNT) return;` (`Group.cpp:1819`), can we do the same here? right now an out of range `id` reads past `m_targetIcons`.

same for `GetCapturePointInfo`, `GetICNodePoint`, `GetAVNodeInfo`, `GetMineOwner` and `IsCaptainAlive`, they all index fixed arrays with whatever the caller passes.

(leading `const` on a by value `ObjectGuid` return doesn't do much either, could drop it.)

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -7281,6 +7281,11 @@ ReputationRank Unit::GetFactionReactionTo(FactionTemplateEntry const* factionTem
         }
     }
 
+    return GetFactionReactionTo(factionTemplateEntry, targetFactionTemplateEntry);
+}
+
+ReputationRank Unit::GetFactionReactionTo(FactionTemplateEntry const* factionTemplateEntry, FactionTemplateEntry const* targetFactionTemplateEntry)
```
sudlud, line 7287: the overload above checks both entries for null before it reaches this body (lines 7248 and 7252), but this one is public and derefs both straight away. `sFactionTemplateStore.LookupEntry()` returns null for an unknown id, which is exactly what a module doing its own lookup would hand it. can we add the same two guards?

sudlud (commented):
thanks for moving the EY player over to a guid, and for the roll snapshot.

few things left before this can go in, most of them inline.

the title needs the AC format, and right now it names three of the eight areas the PR touches. maybe `feat(Core/Misc): Add read-only module accessors`?

src/server/game/Chat/Channels/ChannelMgr.h
```diff
@@ -40,6 +40,7 @@ class ChannelMgr
 
     Channel* GetJoinChannel(std::string const& name, uint32 channel_id);
     Channel* GetChannel(std::string const& name, Player* p, bool pkt = true);
+    ChannelMap const& GetChannels() const { return channels; }
```
sudlud, line 43: `ChannelMap` is declared above the first `public:` (line 30), so it's private and callers can't name `ChannelMgr::ChannelMap`. `auto const&` still binds, but that's all you get. can we move the typedef into the public section?

worth noting the map holds `Channel*`, so this one isn't really read only.

src/server/game/Battlegrounds/Zones/BattlegroundAB.h
```diff
@@ -246,6 +246,18 @@ struct BattlegroundABScore final : public BattlegroundScore
     uint32 BasesDefended = 0;
 };
 
+struct CaptureABPointInfo
```
sudlud: could we keep the struct nested and just make it public instead? something like:

```cpp
public:
    struct CapturePointInfo
    {
        ...
    };

    [[nodiscard]] CapturePointInfo const& GetCapturePointInfo(uint32 node) const { return _capturePointInfo[node]; }

private:
    CapturePointInfo _capturePointInfo[BG_AB_DYNAMIC_NODES_COUNT];
```

`Group::MemberSlot` does the same. the struct stays where it is, no rename needed, and the AB/EY clash goes away on its own since `BattlegroundMgr.cpp` includes both headers.

src/server/game/Battlegrounds/Zones/BattlegroundEY.h
```diff
@@ -346,6 +346,25 @@ struct BattlegroundEYScore final : public BattlegroundScore
...
 
+struct CaptureEYPointInfo
+{
+    CaptureEYPointInfo() : _ownerTeamId(TEAM_NEUTRAL), _barStatus(BG_EY_PROGRESS_BAR_STATE_MIDDLE), _areaTrigger(0)
+    {
+        _playersCount[TEAM_ALLIANCE] = 0;
+        _playersCount[TEAM_HORDE] = 0;
+    }
+
+    TeamId _ownerTeamId;
```
sudlud: same nesting question as in `BattlegroundAB.h`.

on the names: these are public members now, and the [standards](https://www.azerothcore.org/wiki/cpp-code-standards) use `_` for private members (`uint32 _mapEntry;`) and no prefix for public ones (`uint32 ShadowBoltTimer;`). keeping the struct nested means none of them have to change.

src/server/game/Entities/Unit/Unit.cpp
```diff
@@ -15189,6 +15195,14 @@ void Unit::SendPlaySpellVisual(uint32 id)
     SendMessageToSet(&data, true);
 }
 
+void Unit::SendPlaySpellVisual(ObjectGuid guid, uint32 id)
```
sudlud: `SendMessageToSet` goes to whoever can see `this`, not whoever is near `guid`, so `caster->SendPlaySpellVisual(target->GetGUID(), id)` plays the visual on the target but only reaches people watching the caster.

`target->SendPlaySpellVisual(id)` already covers what this looks like it's for. what did you have in mind? if it's for when you only have a guid, maybe resolve it and call the existing one.

src/server/game/Entities/Vehicle/VehicleDefines.h
```diff
@@ -45,6 +45,7 @@ enum VehicleFlags
     VEHICLE_FLAG_CUSTOM_PITCH                    = 0x00000040,           // If set use pitchMin and pitchMax from DBC, otherwise pitchMin = -pi/2, pitchMax = pi/2
     VEHICLE_FLAG_ADJUST_AIM_ANGLE                = 0x00000400,           // Lua_IsVehicleAimAngleAdjustable
     VEHICLE_FLAG_ADJUST_AIM_POWER                = 0x00000800,           // Lua_IsVehicleAimPowerAdjustable
+    VEHICLE_FLAG_FIXED_POSITION                  = 0x00200000            // Used for cannons, when they should be rooted (mod_playerbots, not sure this is still valid since its only used my playerbot)
```
sudlud: `VEHICLE_FLAG_FIXED_POSITION` isn't used anywhere in the core, only defined here. can we leave it out and let the module define its own?

if it has to stay, could we reword the comment? naming a module and saying we're not sure it's still valid is a bit odd in a core header. (`my` should be `by` too.)

src/server/game/Battlegrounds/Zones/BattlegroundIC.h
```diff
@@ -922,6 +922,9 @@ class AC_GAME_API BattlegroundIC : public Battleground
     bool AllNodesConrolledByTeam(TeamId teamId) const override;  // overwrited
     bool IsResourceGlutAllowed(TeamId teamId) const;
     void DoAction(uint32 action, ObjectGuid guid) override;
+
+    [[nodiscard]] ICNodePoint const& GetICNodePoint(uint8 index) { return nodePoint[index]; }
```
sudlud: missing `const`, so this can't be called through a `BattlegroundIC const*`:

```cpp
[[nodiscard]] ICNodePoint const& GetICNodePoint(uint8 index) const { return nodePoint[index]; }
```

`GetNodeState` just above has it.

src/server/game/Entities/Item/ItemTemplate.h
```diff
@@ -818,6 +818,8 @@ struct ItemTemplate
     [[nodiscard]] bool IsWeaponVellum() const { return Class == ITEM_CLASS_TRADE_GOODS && SubClass == ITEM_SUBCLASS_WEAPON_ENCHANTMENT; }
     [[nodiscard]] bool IsArmorVellum() const { return Class == ITEM_CLASS_TRADE_GOODS && SubClass == ITEM_SUBCLASS_ARMOR_ENCHANTMENT; }
     [[nodiscard]] bool IsConjuredConsumable() const { return Class == ITEM_CLASS_CONSUMABLE && HasFlag(ITEM_FLAG_CONJURED); }
+    [[nodiscard]] bool IsWeapon() const { return Class == ITEM_CLASS_WEAPON; }
+    [[nodiscard]] bool IsRangedWeapon() const { return IsWeapon() && (InventoryType == INVTYPE_RANGED || InventoryType == INVTYPE_THROWN || InventoryType == INVTYPE_RANGEDRIGHT); }
```
sudlud, line 822: just checking: are wands meant to count here? `INVTYPE_RANGEDRIGHT` covers guns, crossbows and wands, so this comes back true for every wand. we exclude them in the ammo path (`Spell.cpp:5387`). happy either way, just want to make sure it's deliberate.

src/server/game/Battlegrounds/Battleground.h
```diff
@@ -205,6 +205,7 @@ struct BattlegroundObjectInfo
 
 enum ArenaType : uint8
 {
+    ARENA_TYPE_NONE                 = 0,
```
sudlud, line 208: `ARENA_TYPE_NONE` isn't used anywhere, just flagging it. harmless since nothing switches on `ArenaType`.

src/server/game/Entities/Item/ItemTemplate.h
Celandriel: Yes this is intended to include that.

sudlud (approved):

sudlud:
Can you prepare a module PR for the module failing in module ci please

Celandriel (commented):
> Can you prepare a module PR for the module failing in module ci please 

https://github.com/azerothcore/mod-arena-3v3-solo-queue/pull/67

PR #27612, opened 2026-09-11
chore(Core/Unit): Name all hitinfo flags seen in sniffs.

ratkosrb:
## Changes Proposed:

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

- [X] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Claude was used to help with data analysis. All conclusions were verified and refined by humans against the data.

## SOURCE:

A dataset containing 343082 parsed SMSG_ATTACKERSTATEUPDATE rows with additional data from the update fields of both attacker and victim at the time the packet was received in order to facilitate the analysis. Sniffs from patch 3.3.2 were used to create the dataset.

Parsed data can be downloaded here - https://github.com/ratkosrb/attacking_state_update_data

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Parse original wotlk sniffs.
2. Verify conclusions against sniff data.

**HITINFO_DEBUG** - This flag is read by the client when parsing the packet. It can be verified in IDA or Ghidra, and it's used to include extra debug info about the chance to miss, crit, dodge, parry, etc at the time of the attack.
**HITINFO_ROLLED_DODGE** - This flag is only set victim is able to dodge and the dodge check in RollMeleeOutcomeAgainst was reached. It's never set on Miss because the Miss check happens before the dodge check.
**HITINFO_ROLLED_PARRY** - This flag is only set victim is able to parry and the parry check in RollMeleeOutcomeAgainst was reached. It's never set on Miss or Dodge because those checks happens before the dodge check.
**HITINFO_ROLLED_BLOCK** - This flag is only set victim is able to block and the block check in RollMeleeOutcomeAgainst was reached. It's never set on Miss, Dodge, Parry or Glancing Blow because those checks happens before the block check.
**HITINFO_BLOOD_SPURT** - This flag is used by client to spray extra blood, and server only sets it on Player victims in cases where damage is between 20% and 100% of victim max health.
**HITINFO_PVP** - There is a 100% correlation between this flag and the attacker and victim both being players or a player pet (UNIT_FLAG_PLAYER_CONTROLLED). You will see it in any sniff that includes battlegrounds, arenas, duels, or open world pvp.
**HITINFO_DUEL** - This flag is again only set in cases where both the attacker and victim are players or player pets, but it's never set inside battlegrounds or arenas, and commonly set in fights between players inside sanctuary areas, indicating it is a duel flag.
**HITINFO_CRITICAL_BLOCK** - This flag is only set in case of block, and only for warriors, which is the only class that has the Critical Block talent. The blocked amount when this flag is set is double the amount the same victim blocks without it.

- [ ] Actually implement use of the flags. (Not in this PR)

### DIFF
--- a/src/server/game/Entities/Unit/Unit.h
+++ b/src/server/game/Entities/Unit/Unit.h
@@ -96,29 +96,31 @@ enum VictimState
 enum HitInfo
 {
     HITINFO_NORMALSWING         = 0x00000000,
-    HITINFO_UNK1                = 0x00000001,               // req correct packet structure
+    HITINFO_DEBUG               = 0x00000001,               // packet includes extra debug data such as chance to crit, miss, dodge, etc
     HITINFO_AFFECTS_VICTIM      = 0x00000002,
     HITINFO_OFFHAND             = 0x00000004,
-    HITINFO_UNK2                = 0x00000008,
+    HITINFO_UNK2                = 0x00000008,               // not seen in sniffs
     HITINFO_MISS                = 0x00000010,
     HITINFO_FULL_ABSORB         = 0x00000020,
     HITINFO_PARTIAL_ABSORB      = 0x00000040,
     HITINFO_FULL_RESIST         = 0x00000080,
     HITINFO_PARTIAL_RESIST      = 0x00000100,
     HITINFO_CRITICALHIT         = 0x00000200,               // critical hit
-    HITINFO_UNK10               = 0x00000400,
-    HITINFO_UNK11               = 0x00000800,
-    HITINFO_UNK12               = 0x00001000,
-    HITINFO_BLOCK               = 0x00002000,               // blocked damage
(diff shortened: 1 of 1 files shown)


ratkosrb:
Note for future implementation: The "ROLLED" on the name of some of the flags does not imply a successful Dodge, Parry, Block, it means the chance is above 0 and the server rolled to check whether victim will Dodge, Parry, Block or not. It's not tied to the roll succeeding. You can see appropriate place to apply the flag in vmangos: https://github.com/vmangos/core/blob/448df9ba06d2b2b1678fcfd782a77e1e3ebf26b9/src/game/Objects/Unit.cpp#L2316

Gultask (approved):

Gultask:
Unit.cpp:7000:31 needs to change to  HITINFO_DEBUG :P

PR #27621, opened 2026-09-12
fix(Core/Spells): stop forced casts from inheriting the caster as their destination

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

`Spell::EffectForceCast` always handed the forced cast the original caster as its unit target. When the triggered spell takes no unit target but does need a destination, `Spell::InitExplicitTargets` takes its "needs a dst, none set" branch and reads the *parameter* `targets` rather than the already-sanitised `m_targets`, so the original caster becomes the forced cast's destination. The forced cast then resolves against the original caster instead of against the unit that was forced to cast it.

`Spell::EffectSummonType` runs in `SPELL_EFFECT_HANDLE_HIT` mode and only ever sees `destTarget`, so an implicit `TARGET_UNIT_CASTER` on a summon effect cannot override that inherited destination.

The fix passes the unit target only when the triggered spell's explicit target mask accepts one - the same `TARGET_FLAG_UNIT_MASK | TARGET_FLAG_CORPSE_MASK` test `InitExplicitTargets` itself applies a few lines earlier before discarding it. All three force-cast effects (`SPELL_EFFECT_FORCE_CAST`, `SPELL_EFFECT_FORCE_CAST_WITH_VALUE`, `SPELL_EFFECT_FORCE_CAST_2`) route through this handler, so all three are covered.

### Reach

Twelve spells in the client data force-cast a spell that needs a destination but takes no unit target: 42073, 48759, 52187, 57838, 58566, 62207, 62301, 62921, 64088, 64598, 69839 and 70882.

Seven of them place their summon somewhere new: 48759, 52187, 57838, 58566, 62207, 62921 and 64088. Their triggers summon either at the destination itself (`TARGET_UNIT_CASTER`, which contributes no destination of its own, so the inherited one stands) or at a fixed offset from it (`TARGET_DEST_DEST_BACK` / `_LEFT` / `_RIGHT`). That is why the inherited destination showed up as a placement error instead of being overwritten.

The other five do not move:

- 42073 force-casts on itself, so the original and the forced caster are the same unit.
- 69839's force-cast effect is prevented by `spell_rotface_unstable_ooze_explosion_init` and never reaches this code.
- 70882's trigger 70883 carries `TARGET_DEST_CASTER` on `TargetB`, and `SelectImplicitCasterDestTargets` overwrites the destination with the forced caster's own position either way.
- 62301 and 64598 land correctly on master only because of the workaround below.

Of those seven, two are reachable on master today: 48759, cast by creature 27445's SmartAI when a player walks into its line of sight, so High Abbot Landgren now appears behind the player instead of behind the bell bunny; and 52187, Har'koa's gossip option, whose kitten now appears beside the player. 57838, 58566 and 70882 are cast by nothing in the world DB or in `src/`. 62207, 62921 and 64088 are what the accompanying Elder Brightleaf script PR starts casting: without this core fix every Unstable Sun Beam stacks on the elder instead of one landing at the elder and one under each player in range.

### Removed workaround

The `SpellInfoCorrections.cpp` entry for 62293 (Algalon's Cosmic Smash) forced `Effects[EFFECT_0].TargetB` to `TARGET_DEST_CASTER`. 62293 inherited the wrong destination like every other spell here, but that implicit target made `SelectImplicitCasterDestTargets` overwrite it with the forced caster afterwards (`m_targets.SetDst(dest)`), so the craters landed correctly in spite of the bug. The entry existed only to route around what this PR fixes, so it goes with it.

Removing it changes nothing in-game. `TARGET_DEST_CASTER` returns the caster's exact position, which is also what `InitExplicitTargets` now falls back to, so 33104 spawns in the same spot either way. TrinityCore carries the same correction, but there it is still load-bearing.

Worth knowing when reading the new test: because the correction covers Algalon on master, its subtest passes there too. It guards the fix in the merged state, where the correction is gone. The 62207 and 48759 subtests fail on master as it stands.

### AI-assisted Pull Requests

## Issues Addressed:

- Closes

## SOURCE:

A sniff of the Elder Brightleaf encounter shows the second Unstable Sun Beam's summon being cast by the *player*, not by the elder, and landing at that player's own position. 62207's client data says the same: effect 0 summons at the caster, effect 1 force-casts 62221 on every player within 100 yards, uncapped and `ONLY_ON_PLAYER`.

Not a cherry-pick. TrinityCore carries the same behaviour in both places, `unitTarget->CastSpell(m_caster, …)` in `EffectForceCast` and the identical destination fallback in `InitExplicitTargets`, so there is nothing upstream to credit here.

## Tests Performed:

This PR has been:

Verified against the Elder Brightleaf encounter on a live stack: before the change both sun beams land on the elder, after it one lands on the elder and one under the player.

`TestEffects_ForceCastDestination` in `e2e/suites/spells/effects/` covers the core behaviour on a live stack with three oracles, driving 62207, 62301 and 48759 from a hostile NPC onto the bot: the summon must land on the unit that was forced to cast (62221, 62293) or at its documented offset from that unit (48757, 3 yards behind).

Green twice on the fixed core: 62207's beam 0.0 yards from the forced caster and 15.0 from the original, 62301's crater 0.0, 48757's summon 3.1 and 18.0. With the one-liner reverted and everything else identical, all three fail as assertions with every summon on the original caster instead.

## How to Test the Changes:

1. Walk up to the bell bunny in Howling Fjord (creature 27445, part of "A Fall from Grace"): High Abbot Landgren must appear behind you, not behind the bunny.
2. Take Har'koa's gossip option in Zul'Drak (creature 28401): her kitten must appear beside you, not beside her.
3. Algalon's Cosmic Smash: the crater marker must still land on the players it targets and the damage must still hit that spot 4 seconds later. Placement should be unchanged, but this is the path the removed `SpellInfoCorrections.cpp` entry used to cover, so it is worth a look.
4. With the accompanying Elder Brightleaf script PR applied, pull the elder in the Conservatory of Life and watch where each Unstable Sun Beam lands: one at his feet, one under each player in range. Without this core fix they all stack on him.

- [ ] None.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -149,7 +149,7 @@ go test -tags=e2e ./local/... -count=1 -v -timeout 30m -parallel 1
 | combat/vehicles | spellclick steed enter/exit | P2 | covered | — |
 | spells/aura | apply/query; CC broken by damage; mount persist | P1 | covered (`TestAC_26130_*`) | #26130 |
 | spells/cast | Charge on dummy; fail path; stance; Raise Dead + ghoul | P1 | covered (`TestAC_27061_*`) | #27061 |
-| spells/effects | Charge / grounding totem / Sweeping Strikes Execute | P1 | covered (`TestAC_26997_*`); dummy-summon `blocked-harness` (engineering dummy lifetime) | #26774 #26997 |
+| spells/effects | Charge / grounding totem / Sweeping Strikes Execute; forced cast summons at the forced caster, not at the unit that forced it | P1 | covered (`TestAC_26997_*`, `TestEffects_ForceCastDestination`); dummy-summon `blocked-harness` (engineering dummy lifetime) | #26774 #26997 #27621 |
 | social/group | form / leave / leader / loot method / disband | P2 | covered | — |
 | social/loot | need/greed / master loot; below-half kill | P1 | covered (`TestAC_26862_*`); chest mid-roll `blocked-harness` (GO 194821 UseGameObject); pass-on-loot delete `blocked-harness` (item-survive after ALL_PASSED) | #26894 #26862 #22000 |
 | social/trade | item+gold accept; cancel; walk-OOR TARGET_TO_FAR | P1 | covered | #25723 |
--- a/e2e/suites/spells/effects/effects_e2e_test.go
+++ b/e2e/suites/spells/effects/effects_e2e_test.go
@@ -3,6 +3,10 @@
 package effects_test
 
 import (
+	"fmt"
+	"math"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


Rorschach91 (approved):

Nyeriah:
# Self-review — fix/forcecast-destination → master

**Outcome** 8 findings — all 8 fixed — final round clean

**Caveat** findings 3, 4 and the description half of finding 5 were fixed in the PR description, which was applied after the final round and so was never re-reviewed.

**Reviewed** `7fef5ff7dd` (`fix/forcecast-destination` vs `master`, merge-base diff — 4 files, +210 −8), matching the pushed head

**By** Claude Code (Claude Agent SDK), claude-opus-5 — self-review v0.10, 2026-09-12

**Tested in-game by the author** Elder Brightleaf on a live stack, before and after, plus the e2e harness. Nothing else.

**Verified this review** `TestEffects_ForceCastDestination` green twice on a Release build of this branch (62207's beam 0.0y from the forced caster and 15.0y from the original, 62301's crater 0.0y, 48757's summon 3.1y and 18.0y). With the one-liner reverted and nothing else changed, all three fail as `assert:` with every summon on the original caster. The rest of `suites/spells/...` was run once on the same build: four unrelated tests failed on setup (character creation, dummy spawn, `DamageToFraction` overkill, Cyclone expiry); no baseline was taken, so they are not established as pre-existing.

<details>

**Intent** Forced casts must stop inheriting the original caster as the triggered spell's destination: when the forced spell takes no unit target but needs a destination, its effects should resolve against the unit that was forced to cast, not against the unit that forced it.

**Project rules** `.agents/docs/self-review-rules.md` found and applied (pinned review, regression-risk depth for shared core code, upstream attribution, in-game testing recorded).

**Diff** `e9fb96abfe387efdd6c4fe72b4295fac05dc00a2`

**Metadata** `21225cf00f686a153c5bc08992e9927c2955cc21` (the rewritten description; the rounds ran against `4900bb0eabf81c973edca7c7cf32824352e23728`, the version it replaced)

**Procedural** The branch was force-pushed over `5428a82883` with a lease and the rewritten description applied, both after round 3; the diff content is unchanged from what round 3 reviewed. One untracked `data/sql/updates/pending_db_world/rev_*.sql` from another task was present during rounds 1–2 and excluded from every diff. A concurrent session switched the checkout during round 3; that round reviewed from the commit object instead of the working tree, and the branch state was re-verified afterwards.

**Author-requested work folded in** the now-redundant 62293 `SpellInfoCorrections.cpp` entry was removed after an audit of all twelve affected spells found it to be the only per-spell workaround tied to them, and live-stack e2e coverage was added for both trigger shapes.

**Still to test in-game** Har'koa's gossip kitten (52187, creature 28401) and the bell bunny's High Abbot Landgren (48759, creature 27445) are the only two changed paths reachable on master, and neither was covered by the Brightleaf run. Algalon's Cosmic Smash should place craters and deal damage exactly as before — the removal is a no-op by analysis, so any difference there is the finding that matters most. Worth probing the branches the Brightleaf run does not take: a force-cast whose trigger *does* accept a unit target (the guard's true branch), and a multi-player pull, where 62207 fans out to every player in range.

## Round 1 — `5428a82883`, full changeset, 5 findings

1. `src/server/game/Spells/SpellEffects.cpp:1048` — comment cites Algalon's Cosmic Smash, whose 62293 correction made the example misleading, and calls implicit targets 79/80/81 "caster-relative" when they are `TARGET_DEST_DEST_BACK`/`_RIGHT`/`_LEFT` → **fixed**: wording corrected and the correction removed, which makes the example exact. (The reviewer's stated reason — that the correction zeroed 62293's explicit target mask — was itself wrong; round 2 finding 6.)
2. Commit message reach list omitted 69839 and 70882, and "wrong for all of them and right for none" was false for five of the twelve → **fixed**: rewritten with the seven-that-move split.
3. PR description overstated the live effect (only 48759 and 52187 are reachable on master) and its test step 2 named 57838/58566, which nothing casts → **fixed** in the drafted description.
4. PR description dropped `pull_request_template.md`'s non-removable "How to Test AzerothCore PRs" block → **fixed** in the drafted description.
5. Found during the walk: PR description and commit message both described dest-relative offsets (79/80/81) as caster-relative → **fixed** in both texts.

## Round 2 — `5c3035a717`, round 1 fixes, 3 findings

6. Commit message claimed the 62293 correction "zeroes the explicit target mask and kept the spell off this branch entirely" — false: `_InitializeExplicitTargetMask` seeds `targetMask` from the raw DBC `Targets` field (`0x40`) and only ORs, so 62293 always took this branch; the correction overwrote the destination afterwards → **fixed**.
7. Commit message claimed the crater moves from a combat reach in front to on the player — false: `TARGET_DEST_CASTER` breaks out before the distance-clamping branch, so both paths use the caster's exact position and the removal is behaviourally neutral → **fixed**: message and description now say the correction was dead code and that nothing moves in-game.
8. `e2e/README.md:153` added a second `spells/effects` inventory row where every other category appears once with oracles joined by `;` → **fixed**: merged into the existing row.

## Round 3 — `7fef5ff7dd`, round 2 fixes, clean

Independent reimplementation of `_InitializeExplicitTargetMask`, `GetExplicitTargetMask` and `GetMissingTargetMask` against the shipped `Spell.dbc` confirmed the twelve-spell list, the seven-that-move split, and both corrected Algalon claims. No force-cast trigger needs a destination *and* accepts a unit target, so the partition has no borderline case. Serverside `spell_dbc` adds three force casts, none qualifying. Upstream attribution checked in all three rounds against a local TrinityCore 3.3.5 checkout: TC still carries the unfixed `EffectForceCast` and still keeps its own 62293 correction, so nothing here is mirrored and no `--author` credit is owed.

</details>

PR #27486, opened 2026-09-04
fix(Core/Groups): finalize timed-out loot votes

Vinges541:
## Changes Proposed:

When a group loot timer expires, `EndRoll` resolves the item without recording PASS for unanswered votes. The existing result packets already include those players, but the issue reports that their roll window remains visible.

This PR converts only `NOT_EMITED_YET` votes to PASS, updates the pass count, and broadcasts a timeout PASS before the existing result logic. The packet uses `roll->itemGUID`, matching the roll ID sent in `SMSG_LOOT_START_ROLL` and the existing automatic-pass broadcasts. This gives the client a PASS event tied to the particular roll; closing the stock-client window still needs runtime verification.

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Codex desktop, GPT
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:

- Closes #27299

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [x] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

The linked issue includes reproduction video from clients with and without addons. The timeout packet uses the roll GUID from the existing automatic-pass paths in `GroupLoot`, `NeedBeforeGreed`, and `MasterLoot`.

## Tests Performed:

[...]

Issue #27299, opened 2026-08-24
[CORE][BossLoot] Window for rolling on loot will not vanish if you pass by timeout.


LoturCC:
### Triage

Issue linked from CC: 
https://github.com/chromiecraft/chromiecraft/issues/10059

Triage Notes: 
repro:
- group with an alt and kill a boss like 16028 and observe loot window after not pressing any option

test: https://dl.dropboxusercontent.com/scl/fi/q39xn5zpf0ek8vq4j7tri/wow_Lm3hLF79ZL.mp4?rlkey=k3gyyioeb2oqfywg3kcgmi72w&dl=0

Original Post Below: 

---
### Current Behaviour

So passing on items without actioning the rolling popup. Will leave the popup on the screen for all eternity. 

<img width="817" height="215" alt="Image" src="https://github.com/user-attachments/assets/1ce791d2-3fc6-4eea-bf60-16fd7ff97729" />
As you can see here Oczza run this roll but no message about me passing and the thing stays on the screen and looks actionable. 

Asked Epiria to test as well on next boss, as she got no addons at all. She could reproduce. 

Video showing the menu is fully actionable:

https://github.com/user-attachments/assets/70bdd91e-435a-42fe-98ba-c1d85c987f23

### Expected Behaviour

It should show that "xxxxx passed on item" when time runs out and then the rolling window vanish. 

### Official Sources (that prove or disprove) the issue being reported

Very default behaviour. No source needed. 

### Steps to reproduce the problem

1. Party with someone. 
2. go to instance
3. Kill boss
4. Loot
5. Someone action their popup. 
6. You dont just wait until time vanish

### Customizations

Standard HD Patch for 3.3.5a

These are additional patches to improve the graphics of 3.3.5a and get a more modern look.

https://chromiecraft.com/en/downloads/

### AC rev. hash/commit

https://github.com/chromiecraft/azerothcore-wotlk/commit/ca8e6d78dee182fbe4c840dbc24a577c25844593

### Operating system

Ubuntu 24.04

### Modules

- [mod-1v1-arena](https://github.com/azerothcore/mod-1v1-arena)
- [mod-ah-bot](https://github.com/azerothcore/mod-ah-bot)
- [mod-anticheat](https://github.com/azerothcore/mod-anticheat)
- [mod-arena-3v3-solo-queue](https://github.com/azerothcore/mod-arena-3v3-solo-queue)
[...]

PR #27631, opened 2026-09-13
fix(Core/Groups): identify the finished roll in SMSG_LOOT_ROLL_WON

Nyeriah:
## Changes Proposed:

`Group::CountTheRoll` ends a group loot roll with `SendLootRollWon(ObjectGuid::Empty, ...)`, so the packet that announces the end of a roll never says which roll ended. The client tracks a roll by the guid it got in `SMSG_LOOT_START_ROLL`, which here is the roll's item guid, and it only closes a roll frame for a roll id it recognises. `SMSG_LOOT_ALL_PASSED` already sends that guid. `SMSG_LOOT_ROLL_WON` did not.

This is a follow-up to #27486. That PR gave the members who never answered a roll their missing "passed on" line, which is merged. This is the other half of the same report: a roll that ends with a winner is now identifiable, so the roll window of a member who never voted can be cancelled instead of staying on screen.

Measured on a live stack before and after, two players in a GROUP_LOOT party, one rolling NEED and one never answering, 60 second timer:

| | before | after |
|---|---|---|
| roll id in `SMSG_LOOT_ROLL_WON` | `0x0` | matches `SMSG_LOOT_START_ROLL` |
| roll still pending once the winner is announced | yes | no |

The all-passed path was already correct and is untouched, and that is what made the asymmetry visible: same timeout, same path into `CountTheRoll`, and the roll only terminated when the packet carried the id.

This PR proposes changes to:

### AI-assisted Pull Requests

   - Tools/models used: Claude Code, Opus 5

## Issues Addressed:

- Related to #27299, which #27486 closed. The "passed on" line from that report is fixed; this covers the roll window that stays on screen.

## SOURCE:

cMaNGOS puts the roll id in every group loot roll packet, `SendRoll`, `SendLootRollWon` and `SendAllPassed` alike, and never sends an empty one. Committed with `--author` crediting Cyberium, who wrote that code: https://github.com/cmangos/mangos-wotlk/commit/ffdf9a05d67a04c3c0304e9b021807fa5b867583

The reproduction video is on #27299.

## Tests Performed:

This PR has been:

Run on a live auth + world + MySQL stack, with a new e2e case added in this PR, `TestAC_27299_TimedOutRollTerminates` under `e2e/suites/social/loot/`. Two bots form a GROUP_LOOT party at Uncommon threshold, kill the group loot fixture, and let the 60 second roll timer expire, once with a NEED winner and once with nobody voting. It asserts that the packet ending the roll carries the id from `SMSG_LOOT_START_ROLL`, that the roll stops being pending, and that the member who never voted gets exactly one PASS line and not an automatic one.

Before this change the NEED case fails on the first assertion with roll id `0x0`; after it, both cases pass. The all-passed case passes either way, which is the control.

## How to Test the Changes:

1. Two players in a party set to Group Loot, threshold Uncommon or higher.
2. Kill something that drops an item above the threshold and open the corpse.
3. One player rolls Need. The other leaves the popup alone.
4. When the timer runs out, the player who never answered should see the pass message and the roll window should go away.
5. Repeat with neither player answering, which goes through the all-passed path and should behave as it did before.

- [ ] Out of scope here, but noticed nearby: in `Group::GroupLoot`, a player with "pass on loot" enabled gets `vote = PASS` without `totalPass` being incremented, so the `if (r->totalPass)` guard can suppress their pass line entirely and `CountRollVote`'s early-finish threshold is short by one. `NeedBeforeGreed` misses the same increment but broadcasts unconditionally, so only the counter is off there.

---


Gultask (approved):

Nyeriah:
# Self-review — fix/loot-roll-won-guid → master

**Outcome** No findings — single round, clean; one out-of-scope observation deferred

**Reviewed** `0e94ee3c9c` (`fix/loot-roll-won-guid` vs `master`, merge-base diff — 3 files, +231 −3)

**By** Claude Code, Opus 5 — self-review v0.10, 2026-09-13

**Verification** `TestAC_27299_TimedOutRollTerminates` run against a live auth + world + MySQL stack, both subtests green. Measured before and after on the same fixture: roll id in `SMSG_LOOT_ROLL_WON` goes from `0x0` to the id from `SMSG_LOOT_START_ROLL`, and the roll stops being pending once the winner is announced. `apps/codestyle/codestyle-cpp.py` clean. The e2e file was vetted with `GOWORK=off` against the `v1.0.8` harness pinned in `e2e/go.mod`, so it builds on the CI path and not only against a local AzerothGhost clone.

<details>

**Intent** A group loot roll that ends with a winner must name the roll in `SMSG_LOOT_ROLL_WON`, using the same roll id the client received in `SMSG_LOOT_START_ROLL`, so a member who never voted can have their roll window closed. Adds live-stack e2e coverage for group loot rolls that end on the timer rather than on a vote.

**Project rules** `.agents/docs/self-review-rules.md` found and applied — pinned review, regression-risk depth for shared core C++, upstream attribution, in-game testing.

**Diff** `42d6e7ed51ba4264a1921c687d2d88cdbbce2ee6`

**Metadata** `20347c4814d481aaac6f5a2232604b3434af0fd3`

Coverage: all three changed files reviewed in full, plus the unchanged code the change relies on and the harness APIs the new test binds to. Governing docs run as a checklist: `AGENTS.md`, `cpp-guidelines.md`, `code-review.md`, `self-review-rules.md`, `e2e-policy.md`, `e2e/README.md`, `.editorconfig`. No file in scope was skipped.

## Round 1 — `0e94ee3c9c`, full changeset, 0 findings

Blast radius traced rather than assumed: `SendLootRollWon` has exactly two callers, both in `CountTheRoll` and both changed, and `SMSG_LOOT_ROLL_WON` is built in one place only, with no caller in scripts or `modules/`. `roll->itemGUID` is the value every recipient already received in `SMSG_LOOT_START_ROLL`, set once at `Roll` construction and never mutated. Recipients that never got a start roll for that guid, the auto-passers at `Group.cpp:1189`, `:1348` and `:1418`, previously received `0x0` and now receive an id they do not track, which is a no-op either way. All `SendLootRoll` result lines are emitted before `SendLootRollWon` in both branches, so nothing arrives after the client would cancel the frame.

Upstream attribution checked against the rule: `d90759d13` is authored `Cyberium <cybermomo@techemail.com>`, matching the author of the cited cMaNGOS commit, and the template's cherry-pick box is checked. The mirrored thing is the invariant that every group loot roll packet names the roll; cMaNGOS uses the loot source guid where AzerothCore uses the per-roll item guid, and the PR body says so.

e2e checklist: `//go:build e2e`, consumer suite, blank MySQL driver import, `meta.Begin` serial, prefix within the 7-character limit, `TestAC_<issue>_<Short>` naming with the issue URL in the comment, `Preconditionf` for setup and `Assertf` for the oracle, a unique oracle against the five existing tests in the suite, and the inventory row updated without dropping existing links. The oracle is not tautological: the harness drops a pending roll only when both roll id and slot match, so both assertions fail before the fix.

No leftovers, no debug logging, no stray files, no SQL.

### Deferred

1. `src/server/game/Groups/Group.cpp:1606` — `CountTheRoll` early-returns when the `Loot` was destroyed before the roll timer fired, deleting the roll without sending `ROLL_WON` or `ALL_PASSED`, which leaves the same stuck roll window on a third path → **deferred**: pre-existing and outside this change's intent, author files it as a follow-up issue. Raised by the reviewer as an observation, not as a finding against this diff.

</details>

PR #26152, opened 2026-06-11
feat(Core/Spells): WorldObject spell casters - GameObjects cast their own spells

blinkysc:
## Changes Proposed:

This PR ports TrinityCore's WorldObject spell caster system: `Spell::m_caster` becomes `WorldObject*`, GameObjects (traps, goobers) cast their spells **directly**, and the old hack in `GameObject::CastSpell` that summoned an invisible WORLD_TRIGGER creature to cast on the GO's behalf is removed.

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [x] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

**Core:**
- `WorldObject::CastSpell` entry point with `CastSpellExtraArgs`/`CastSpellTargetArg`; `SpellScript::GetGObjCaster`/`AuraScript::GetGObjCaster`
- WorldObject-level spell helpers (moved from Unit, TC structure): `SpellHitResult`/`MagicSpellHitResult`, `CalculateSpellDamage`, `CalcSpellDuration`, `ModSpellDuration`, `ModSpellCastTime`, `SendSpellMiss`, `SendSpellNonMeleeDamageLog`
- Original-caster fallbacks so GameObject casts work end-to-end: hit results, aura application (GO-cast auras apply with cleared caster GUID, sniff-verified vs spell 22247), `EffectSummonType`, **diminishing returns** (trap CC now diminishes in PvP through the trap owner) and **spell reflection** (reflected trap spells bounce onto the trap owner instead of fizzling)
- `GameObject::getLevelForTarget`: owner level → trap template level → match target (ownerless traps used to resolve as level 1 = ~1% hit chance vs level 80)
- `Spell::finish` sets `SPELL_STATE_FINISHED` before the non-unit-caster early return (otherwise `SpellEvent` re-executed GO casts every world tick — infinite summon/aura loops)
- Null-caster guards across `Spell.cpp`/`SpellEffects.cpp`; caster-less auras excluded from character DB save/load
- Despawned GameObjects rejected in the use opcode handlers (they remain visible/clickable for GMs and could be re-fired endlessly)

**Scripts** — audit of every spell cast by a GameObject with a script attached (`SpellScript::GetCaster()` returns nullptr for GO casts):
[...]

PR #27632, opened 2026-09-13
fix(Scripts/Feralas): summon the Gordunni trap mound through its own spells

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

`spell_gordunni_trap` summoned gameobject 144064 directly, so the junk mound the trap can also yield (177681) never appeared, and the chest was the only loot a Gordunni Trap could ever give. The script now casts 11756 or 19394 on the player who disturbed the trap and lets the summon spells place the mound.

One behaviour change worth reviewer attention: placement now follows the summon spells' own `TARGET_DEST_CASTER_FRONT` from the player, rather than the trap's position. That matches upstream, and the trap-anchored form upstream used before that commit is not expressible here, since `SpellScript::GetGObjCaster()` resolves through `m_spell->GetCaster()` and AzerothCore's caster for 19395 is the world trigger summoned on the trap.

The handler moves from `OnCast` to the dummy effect, where the hit target is available.

### AI-assisted Pull Requests

## SOURCE:

Mirrored from TrinityCore https://github.com/TrinityCore/TrinityCore/commit/9d6c4e3d931875008cd9ca15754b3edb1aef2a01, committed with `--author` crediting offl, and `Co-authored-by` for Giacomo Pozzoni. The `Co-authored-by: blinkysc` trailer credits https://github.com/azerothcore/azerothcore-wotlk/pull/26152, which touches the same script and is where this change started.

No SQL accompanies this. TrinityCore shipped `2021_12_23_02_world.sql` with that commit, setting `RangeIndex`, `EffectImplicitTargetA` and `EffectRadiusIndex` on 11756 and 19394; AzerothCore already carries those values in `data/sql/base/db_world/spell_dbc.sql`.

## Tests Performed:

This PR has been:

Verified on a live realm with the e2e test included here, in both directions:

| Core | Result |
|---|---|
| With this change | PASS at `-count=2`, distinct gameobject guids per run, rolls 3/1 and 1/4 |
| Without it | FAIL — chest 12/12 across 12 distinct summons, junk mound 0 |

## How to Test the Changes:

1. Find a Gordunni Trap (gameobject 144050) in Feralas and disturb it, or spawn one with `.gobject add 144050` and click it.
2. A dirt mound appears in front of you. Loot it.
3. Repeat several times: roughly half the triggers should give the junk mound (177681, junk loot) and half the chest mound (144064, Gordunni Cobalt) plus its cobalt visual. Before this change only the chest ever appeared.

Alternatively, run the included test against a live stack:

```
cd e2e && make e2e-sub C=spells/effects
```

- [ ] A mounted player gets no mound. The cast is untriggered, as upstream wrote it, so `Spell::CheckCast` refuses it with `SPELL_FAILED_NOT_MOUNTED`. This does not happen on TrinityCore, which dismounts for any object whose `IsUsableMounted()` is false; AzerothCore skips that dismount for `GAMEOBJECT_TYPE_TRAP` ("workaround for GO casting", `GameObject.cpp`). Left at parity with upstream deliberately rather than deviating inside a cherry-pick. The AzerothCore-side carve-out looks like the better place to close the gap, since it would cover every GO-casting trap. The same applies to a stunned or feared player, via `CheckCasterAuras`.


EricksOliveira:
# Maintainer review — PR #27632 `fix(Scripts/Feralas): summon the Gordunni trap mound through its own spells`

**Verdict** Blocks. The new cast is untriggered, so a player who is mounted or crowd controlled when the trap fires now gets no mound at all, where the old code always produced one. Separately, the Known Issues note in the description describes a leak that does not exist in AzerothCore, and the e2e test repeats it in a comment.

**Reviewed** `2ff49d2cb54dfe26e3a3cf62779fcd9fb5889195` (`Nyeriah:fix/gordunni-trap-mounds` vs `master`, merge-base `9c416aaacb5537636abb13c80f55a88947838e33`, 3 files, +161 −14)

**Not run** The server was not built or executed. Everything below comes from the sources, the world DB and the upstream patch. Finding 1 is traced statically and still wants a live check, see Not verified.

**By** Claude Code (desktop), Opus 5, fresh-eyes pass in an isolated subagent, maintainer-review v0.6, 2026-09-14

<details>
<summary>Review details</summary>

**PR** https://github.com/azerothcore/azerothcore-wotlk/pull/27632, author `Nyeriah`, opened 2026-09-13, base `master`, not a draft, labels `Script` / `file-cpp` / `e2e`, `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`

**Diff** `6f74f1d86fe2257c02554367b11d153680be4619f117a20aa2741b54f79207aa`

**Checks** 13 green, `e2e full / e2e (native / full)` among them at 18m, so the new test built and passed on a fixed core. CodeRabbit's golangci-lint step reported a typecheck error on `e2e/suites/spells/effects`, which is the `e2e` build tag excluding the files, not a code problem.

**Reviews on record** Two bot reviews, both COMMENTED on the current head. No human review, no approval, no human comment. Both inline threads are unresolved and neither has a reply.

**Project docs applied** `AGENTS.md`, `.agents/docs/cpp-guidelines.md`, `.agents/docs/cpp-scripts.md`, `.agents/docs/code-review.md`, `.agents/docs/self-review-rules.md`, `.agents/docs/e2e-policy.md`, `.editorconfig`.

## What the PR does

Gordunni Trap is gameobject 144050, a `GAMEOBJECT_TYPE_TRAP` whose `trap.spellId` is 19395. Using it runs `GameObject::Use` (`GameObject.cpp:1521-1531`), which calls `GameObject::CastSpell` (`:2083`). 19395 is not caster referenced, so that summons a world trigger at the trap and has it cast 19395 on the player. `spell_gordunni_trap` is bound to 19395 in `spell_script_names.sql:371`.

The old script summoned gameobject 144064 itself from `OnCast`, anchored on the caster, which is that world trigger. The junk mound 177681 could therefore never appear, and the chest was the only loot a trap could ever yield.

The new script moves the handler to `OnEffectHitTarget` on `EFFECT_0` and has the hit unit cast 11756 (chest) or 19394 (junk) on itself. Both spells are `SPELL_EFFECT_SUMMON_OBJECT_WILD` with `ImplicitTargetA 47`, so placement now follows `TARGET_DEST_CASTER_FRONT` from the player instead of the trap's own position. The PR flags that change itself for reviewer attention.

## Findings

### 1 — blocking · `src/server/scripts/Kalimdor/zone_feralas.cpp:41`

`target->CastSpell(target, <spellId>)` with two arguments is an untriggered cast, so the whole `Spell::CheckCast` pipeline now stands between the trap and the mound.

Only one overload is viable with two arguments: `Unit.h:1680`, whose `TriggerCastFlags triggerFlags` defaults to `TRIGGERED_NONE`. The `bool triggered` overload right above it at `:1679` has no default for that parameter.

11756 and 19394 both carry `Attributes = 384` (`spell_dbc.sql:359` and `:489`), which does not include `SPELL_ATTR0_ALLOW_WHILE_MOUNTED` (`0x01000000`). So a mounted player hits `Spell.cpp:6201` and gets `SPELL_FAILED_NOT_MOUNTED`. A stunned, feared or confused one hits `CheckCasterAuras` at `Spell.cpp:6244-6248` and is refused there. In both cases the trap fires, the player gets the red error on screen, and no mound spawns.

Nothing dismounts the player on this path: `IsUsableMounted()` is false for `GAMEOBJECT_TYPE_TRAP` (`GameObjectData.h:411-426`), and `HandleGameObjectUseOpcode` (`SpellHandler.cpp:329-347`) only range checks. Feralas is a zone people ride through, and the trap sits among Gordunni ogres, so both states are ordinary.

Passing `true` as the third argument fixes it. That maps to `TRIGGERED_FULL_MASK`, which carries both `TRIGGERED_IGNORE_CASTER_MOUNTED_OR_ON_VEHICLE` and `TRIGGERED_IGNORE_CASTER_AURAS`.

Upstream is byte identical here, so this came in with the mirror rather than from the port. It is still what happens on AzerothCore.

### 2 — blocking on the record · PR description, Known Issues, and `e2e/suites/spells/effects/gordunni_trap_e2e_test.go:27-32`

The claim that these mounds are never despawned is wrong on AzerothCore, and the same text is baked into the test as the stated reason for the ring and jitter machinery.

`EffectSummonObjectWild` calls `SetSpellId` at `SpellEffects.cpp:3998`, and `GameObject::SetSpellId` (`GameObject.h:178-182`) clears `m_spawnedByDefault`. `GameObject::Update` zeroes `m_respawnTime` at `GameObject.cpp:621`, reaches `if (!m_spawnedByDefault)` at `:653-658` and sets `GO_JUST_DEACTIVATED`. From there the `isSummonedAndExpired` case at `:844-846` skips the branch that keeps chests alive, and `:863-866` deletes the mound. Both mounds are `type=3` (CHEST) in `gameobject_template.sql:5457` and `:9111`, which is exactly the type that case guards.

The machinery in the test is still needed, since the mounds live for the spell's duration (`DurationIndex 4`). Only the reason is wrong. Copilot raised this and it has no reply.

### 3 — non-blocking · `e2e/suites/spells/effects/gordunni_trap_e2e_test.go:75`

`seen` starts empty and is only filled after each trip's own search at `:117`, so a mound already standing nearby is never excluded before the first `GameObjectUse` at `:100`. A leftover from an earlier run can satisfy this one.

Both bots asked for the same thing and neither got an answer. The odds are low, since the ring centre moves by up to 150 yards per run, but recording nearby mound guids into `seen` before the first trip is a couple of lines.

### 4 — non-blocking · `e2e/suites/spells/effects/gordunni_trap_e2e_test.go:45`

`maxTrips = 12` over a 50/50 roll leaves a 2·(1/2)¹² chance, roughly 0.05%, of failing on a correct core, and `-count=2` doubles the exposure. `.agents/docs/e2e-policy.md` sends nondeterminism to manual testing with an optional deterministic slice. Copilot raised this as a suppressed comment, unanswered.

### 5 — nit · `src/server/scripts/Kalimdor/zone_feralas.cpp:40`

`GetHitUnit()` is dereferenced with no guard, where the old code had `if (Unit* caster = GetCaster())`. `OnEffectHitTarget` fires with `unitTarget` null on the gameobject and item target paths (`Spell.cpp:3439` and `:3460`).

19395 is a client DBC spell, absent from `spell_dbc`, so I could not read its implicit targets. A `SPELL_EFFECT_DUMMY` with unit targets never gets entries in those target lists, so this is not reachable here, and upstream is identical. Raised as defensive only: 159 of the roughly 195 `GetHitUnit()` sites under `src/server/scripts/` use the guarded form.

### 6 — minor · `e2e/suites/spells/effects/gordunni_trap_e2e_test.go:117`

`seen[chest], seen[junk] = true, true` writes `seen[0] = true` whenever one of the two is zero. Harmless, guid 0 is never looked up. The `switch` below prefers chest over junk, which only bites when a stray mound of the other type sits in range, so it is a symptom of finding 3 rather than its own bug.

### 7 — minor · `e2e/suites/spells/effects/gordunni_trap_e2e_test.go`

`.agents/docs/e2e-policy.md` asks for the issue or PR URL in a comment, and the neighbouring test does exactly that at `effects_e2e_test.go:239` (`// PR: https://github.com/azerothcore/azerothcore-wotlk/pull/27621`). This one cites only the TrinityCore commit, so the AzerothCore PR that produced the fix is not findable from the test.

## What checked out

The mirror is faithful. I pulled the upstream patch and the C++ body is a literal copy, `urand(0, 1)` included, which also matches the `Random.h` helper the C++ guidelines ask for.

The no-SQL claim holds. TrinityCore shipped `2021_12_23_02_world.sql` with that commit setting `RangeIndex=12`, `EffectImplicitTargetA` 47 and `EffectRadiusIndex` 15 on 11756 and 19394, and `spell_dbc.sql` already carries all of those. Target 47 is `TARGET_DEST_CASTER_FRONT` (`SpellInfo.cpp:260`), so the placement account in the description is accurate.

Attribution is in order: `82bde945` is authored `offl` with `Co-authored-by` for Giacomo Pozzoni, which is what `AGENTS.md` asks of a cherry-pick. There is also a `Co-authored-by: blinkysc` the description never mentions, worth a line so the record explains itself.

Codestyle matches the file and `apps/codestyle/codestyle-cpp.py` passes, as does the green C++ job on the head. The script binding at `spell_script_names.sql:371` already exists, so no binding SQL is owed.

I did not raise prefer data over code from `code-review.md`. A 50/50 roll between two summon spells driven off a dummy effect is not expressible in SmartAI or conditions here.

## Comment walk

Every conversation comment, review body and inline thread, with whether it was answered and whether it still holds at `2ff49d2c`.

1. **`coderabbitai[bot]`, conversation comment, 2026-09-13 19:16** — walkthrough plus a Low merge risk verdict, and one recommendation: record pre-existing mound guids before tripping the trap. No reply. Still holds, it is finding 3. The Low risk verdict itself did not survive: the bot never looked at the cast flags.

2. **`coderabbitai[bot]`, review body, 2026-09-13 19:16** — run configuration and file list only, nothing to answer.

3. **`coderabbitai[bot]`, inline thread on `gordunni_trap_e2e_test.go:74`, 2026-09-13 19:16** — the random offset can reuse an earlier run's ground, so a standing mound can create a false pass. Unresolved, no reply. Still holds, finding 3.

4. **`copilot-pull-request-reviewer[bot]`, review body, 2026-09-13 19:16** — "Changes recommended", three points, two of them suppressed. No reply to any. All three still hold, see the next three items.

5. **`Copilot`, inline thread on `gordunni_trap_e2e_test.go:75`, 2026-09-13 19:16** — same stale guid concern, and it correctly notes that `SetSpellId` makes these mounds temporary rather than permanent. Unresolved. Both halves still hold, findings 3 and 2.

6. **`Copilot`, suppressed comment on `gordunni_trap_e2e_test.go:45`** — `maxTrips` leaves a 0.049% failure chance on a correct core, against the e2e policy on intermittent failures. Unanswered, still holds, finding 4.

7. **`Copilot`, suppressed comment on `gordunni_trap_e2e_test.go:30`** — the comment documents a known issue that does not exist, since `EffectSummonObjectWild` sets the spell id after the duration. I traced it and the bot is right. Unanswered, still holds, finding 2.

## Not verified

Nothing was run. The author reports testing in game plus the e2e test passing in both directions, with and without the fix, and CI backs the second half.

What that leaves open is finding 1, because both the manual test and the bot run from an unmounted, uncontrolled player. Worth asking the author to trip the trap while mounted and while stunned or feared by a Gordunni ogre, and to say what came out. The other side effect of the change is placement: the mound now lands in front of the player rather than at the trap, so it is worth looking at where it lands from a few different facings, and on a slope.

## Recommended next action

Comment on the PR asking for `true` on the `CastSpell` (finding 1) and for the Known Issues note to be corrected in both the description and the test comment (finding 2), and point at the four bot threads still open. Nothing has been posted, approved or labelled. Do not apply `To Be Merged`.

</details>

Nyeriah:
Thanks for the reviews. Answering the open threads together.

**Corrected — the mounds are not permanent.** `EffectSummonObjectWild` calls `SetRespawnTime` and then `SetSpellId`, and `GameObject::SetSpellId` clears `m_spawnedByDefault`, so the mounds despawn after the summon spell's 120s duration. The Known Issues note claiming a permanent object was wrong, and the same wrong reason was baked into the test comment. Both are fixed. The ring and jitter machinery stays, because a 120s lifetime still outlasts a run and an earlier trip's mound would otherwise satisfy a later one — only the stated reason changes.

**Fixed — stale mound GUIDs.** The test now banks any mound already standing inside the search radius into `seen` before tripping the trap, so only a fresh summon can satisfy a trip. It also no longer writes `seen[0]` when one of the two lookups comes back empty.

**Fixed — the roll's failure probability.** `maxTrips` goes from 12 to 20, taking the chance of missing a face on a correct core from about 0.05% to about 0.0002%. The loop still exits as soon as both faces appear, which takes three trips on average, so the headroom is free in the normal case. The ring radius grows with it so the extra points stay more than a search radius apart.

**Added** the PR URL to the test comment, alongside the upstream commit.

**Not changing — the untriggered cast.** This is deliberate parity with the upstream commit this mirrors. You are right about the mechanism: AzerothCore skips the dismount for `GAMEOBJECT_TYPE_TRAP` in `GameObject::Use` ("workaround for GO casting"), where TrinityCore dismounts for any object whose `IsUsableMounted()` is false, so a mounted player reaches the self-cast still mounted here and gets `SPELL_FAILED_NOT_MOUNTED`. Deviating from the upstream script is a call for a maintainer rather than something to slip into a cherry-pick, and if the difference is worth closing, the AzerothCore-side trap carve-out looks like the better place to do it, since it would cover every GO-casting trap rather than this one script. Happy to follow whichever way you prefer.

**Not changing — the unguarded `GetHitUnit()`.** Same reasoning, and by your own analysis the null paths are the gameobject and item target lists, which a unit-targeted `SPELL_EFFECT_DUMMY` never populates. Upstream is identical here.

On the `Co-authored-by: blinkysc` trailer: this started as an extraction from #26152, which touches the same script, and the trailer credits that work. Worth stating explicitly rather than leaving it to be inferred, so it is now in the description.

EricksOliveira:
Thanks, the four fixes check out on `5a796055`. I also confirmed the dismount difference you describe: AC carves traps out at `GameObject.cpp:1492-1494`, TC dismounts before the trap branch.

On the untriggered cast, I'd still lean towards passing `true`. `AGENTS.md` asks a mirror for attribution rather than literal fidelity ("even when rewritten against AC"), and the mounted player loses a mound they used to get as soon as this merges. It's one character and it doesn't block fixing the carve-out properly later. If you'd rather keep parity, can we open the `GameObject::Use` carve-out as its own issue instead of leaving it in Known Issues?

On the new baseline: `FindGameObjectByEntry` returns a single guid, so if a banked stale mound and the fresh one are both inside the 8yd radius, can it keep handing back the stale one? That would fail the trip on a correct core. Do you know how it picks?

Gultask (approved):

PR #27651, opened 2026-09-15
fix(Scripts/Ulduar): stop Freya's Ward adds despawning on the spell timer

Nyeriah:
## Changes Proposed:

This PR proposes changes to:

Flame Leviathan's Tower of Life adds despawn seconds after they spawn, so raids just ignore them.

`npc_freya_ward_summon` raises its summons to `TEMPSUMMON_MANUAL_DESPAWN` so the adds stay up until the encounter clears them, but it sets that from `IsSummonedBy`, which runs *inside* the summon call. `Map::SummonCreature` fires `InitSummon()`, and with it `JustSummoned` and `IsSummonedBy`, before it returns, and both `WorldObject::SummonCreature` and `Spell::EffectSummonType` then re-apply the duration-derived type on the way out. The override never survives the call, so the adds fall back to the summon spells' own durations:

| spell | summons | `SpellDuration` | despawns after |
|---|---|---|---|
| 62947 | 33387 Writhing Lasher | 1 | 10s |
| 62906 -> 62907 | 34275 Ward of Life | 27 | 3s |

This is a regression from #27567. Before it, `npc_freya_ward::UpdateAI` applied `TEMPSUMMON_MANUAL_DESPAWN` on a later world tick, after the summon call had already returned, so there was nothing left to overwrite it. That PR moved the same call into `IsSummonedBy`, into the window the two assignments clobber.

The fix defers the override by a tick so it lands once the summon call has finished. It goes through the creature's event processor rather than a flag checked in `UpdateAI`, because `Creature::Update` skips `UpdateAI` entirely while the creature is in evade mode, whereas `WorldObject::Update` runs `m_Events.Update(diff)` unconditionally and first.

34275 shares this AI, so the same line covers the Wards of Life.

### AI-assisted Pull Requests

## Issues Addressed:

No tracked issue. Reported on Discord against a 10-man run and reproduced on a 25-man test realm.

## SOURCE:

Nothing external to cite: this restores the add lifetime the script already intended and had before #27567. The 10s and 3s figures above are the `SpellDuration` rows the two summon spells carry in `Spell.dbc`.

## Tests Performed:

This PR has been:

Covered by a new live-stack e2e regression test, `TestUlduar_FreyaWardLasherOutlivesSummonDuration` in `e2e/suites/instances/northrend/ulduar/`. It spawns a Freya's Ward Targetting Reticle and a Freya's Ward, lets `npc_freya_ward` run its own cast, then pins the first Writhing Lasher's GUID and asserts it is still in the world 25s later. The GUID has to be pinned because the ward keeps summoning every 29s, so "some lasher is present" stays true even on a broken core.

Before the fix:

```
Writhing Lasher 0xF13000826B0005C4 summoned by ward 0xF1300082570005C3
E2E_FAIL assert: Writhing Lasher 0xF13000826B0005C4 left the world 10.024s after it was summoned
--- FAIL: TestUlduar_FreyaWardLasherOutlivesSummonDuration (15.83s)
```

10.024s is 62947's `SpellDuration` to the millisecond.

After the fix, green, and stable over `-count=2`:

```
PASS Writhing Lasher 0xF13000826B0005C4 still up 25s after summon (hp=64260/64260)
--- PASS: TestUlduar_FreyaWardLasherOutlivesSummonDuration (31.70s)
```

## How to Test the Changes:

1. Enter Ulduar with the Tower of Life left standing and pull Flame Leviathan.
2. Wait for Freya's Ward to spawn its adds, roughly 30s into the fight.
3. Leave a Writhing Lasher alive and time it. It should stay up and keep attacking the vehicles instead of vanishing ~10s after it spawned, and the Ward of Life should not vanish after ~3s.
4. Kill the adds and finish the encounter. They should still be cleaned up on wipe and on kill, since `boss_flame_leviathan` clears them through `ACTION_DESPAWN_ADDS`.

Worth probing beyond the main path: wipe with lashers up, and kill Flame Leviathan with lashers up. Both go through `summons.DoAction(ACTION_DESPAWN_ADDS)` followed by `summons.DespawnAll()`, and a `TEMPSUMMON_MANUAL_DESPAWN` summon now has no timer of its own to fall back on if that path ever misses one.

- [ ] The underlying core behaviour is untouched. `WorldObject::SummonCreature` and `Spell::EffectSummonType` still re-apply the duration-derived summon type after `Map::SummonCreature` has fired `JustSummoned` and `IsSummonedBy`, so any script that sets a summon type from either hook has it silently discarded. `npc_frosthound::JustSummoned` in `zone_storm_peaks.cpp` does exactly that today. Worth a separate look, but it is a much wider change than this fix needs.


Nyeriah:
# Self-review — fix/temp-summon-type-override → master

**Outcome** 6 findings — 3 fixed, 1 fixed as a drafted PR-body edit, 2 deferred — round 3 confirmed the last fix holds and raised only a deferred finding

**Caveat** Fixes not re-reviewed: the round 1 finding 3 draft was never applied to the PR description, so no round judged the corrected text.

**Reviewed** `f468e13d17` (`fix/temp-summon-type-override` vs `upstream/master`, merge-base diff — 3 files, +95 −2)

**By** Claude Code (desktop app), Claude Opus 5 — self-review v0.10, 2026-09-15

`python apps/codestyle/codestyle-cpp.py` passes; `gofmt -l` and `go vet -tags=e2e` clean in every round; `worldserver` builds clean (MSVC, Release).

The oracle was demonstrated both ways. On the unfixed core the pinned Writhing Lasher left the world 10.024s after it was summoned, matching 62947's `SpellDuration` to the millisecond. On the fixed core at the reviewed state `f468e13d17` it was still up at 25s, twice over `-count=2`, and the cleanup cleared the ward plus both adds it had summoned.

In-game testing by the author: not yet recorded.

<details>

**Intent** Flame Leviathan's Tower of Life adds (Writhing Lasher 33387, Ward of Life 34275) despawn on their summon spell's own duration instead of staying up until the encounter clears them. Make the AI's `TEMPSUMMON_MANUAL_DESPAWN` override take effect, and add a live-stack e2e regression test for it.

**Project rules** `.agents/docs/self-review-rules.md` found and applied (pinned review — every round ran on a commit and each round's fixes were committed before the next; upstream attribution checked: nothing mirrored from another core, no `--author` needed; regression risk treated as content-script scope, plus the core mechanism the script relies on)

**Diff** `1ee1e313cda881b9d39d011780b26d366caaaf36`

**Metadata** `bf10a16754e427ddc3f0771a407d83b9bdb9644a`

Each round ran in an isolated subagent context with `.agents/plans/**` and all `*.SELF-REVIEW*`/`*.HANDOVER*` files excluded. Every round reported full coverage of its scope.

## Round 1 — `14b5b9b84e`, full changeset, 3 findings

1. `e2e/suites/instances/northrend/ulduar/ulduar_e2e_test.go:1100` — `bot.CheatGod(t)` is dead: `ScenarioBot.CombatReady` already runs `CombatReadyOpts{God: true}` → **fixed**
2. `ulduar_e2e_test.go:1124` — cleanup killed only the pinned lasher, so every other add the ward summoned stayed permanent, which this PR is what causes → **fixed**
3. submission metadata, Known Issues — names `npc_frosthound::JustSummoned` as the only other script hitting the core trap; `boss_yoggsaron_cloud::JustSummoned` does too (62979 → 33136, SummonProperties 714, same WILD/Type 0 branch, 600s timer) → **fixed** as a drafted replacement bullet, not applied to the PR

## Round 2 — `9fdde9f4e5`, round 1 fixes + changed metadata, 2 findings

1. `ulduar_e2e_test.go:1128` — the sweep was one snapshot taken ~25s into the ward's 29s cycle, leaving ~4s for the kills, the probe and the spawn cleanup before another pair was summoned outside it → **fixed**
2. `ulduar_e2e_test.go:329,717` — the same redundant `CombatReady` + `CheatGod` pair stands in two other tests in this file → **deferred**: pre-existing and outside the changeset, editing unrelated oracles would widen a three-file PR; noted for a follow-up e2e tidy-up, not filed

## Round 3 — `f468e13d17`, round 2 fix + changed metadata, 1 finding

1. `ulduar_e2e_test.go:1128-1136` — the kill and sweep are on the success path, so an abort leaves the adds permanent; asymmetric, since on an unfixed core they self-despawn and only a failure on a fixed core leaks → **deferred**: same follow-up note, bites only a run that has already failed, in a per-run throwaway instance

Round 3 also verified the load-bearing claims independently: `Creature::Update` reaches `UpdateAI` only in the `Alive` branch, so a dead ward runs no further cycles; `.damage` reaches a `NOT_SELECTABLE | NON_ATTACKABLE` trigger because `HandleDamageCommand` calls `Unit::DealDamage` with no attackability check; and `.npc delete` resolves by spawn id, so it still works on a corpse.

</details>

Gultask (approved):

PR #27646, opened 2026-09-14
fix(Core/Events): do not count a holiday as active during its building stage

sudlud:
## Changes Proposed:

`IsHolidayActive()` matched game events by holiday id alone and ignored `game_event.holidayStage`.
Brewfest and the Darkmoon Faire both have a building stage that carries the same holiday id as the
event itself (Brewfest: events 70 and 91 on stage 1, event 24 on stage 2), so everything gated on
the holiday went live the moment the building stage started, days before the holiday itself.

That is why Coren Direbrew turns up in the random dungeon finder while Brewfest Building is running.
The same went for Brewfest holiday items, the holiday achievement criteria, and the Brewfest music
object checks.

`GameEventMgr` now records the highest `holidayStage` any event declares per holiday while loading
`game_event`, and `IsHolidayActive()` only counts an active event sitting on that stage. Holidays
with a single stage keep their current behaviour, and so do the events that set a holiday id with
stage 0.

This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Claude Code with Claude Opus 5, for the change and this description
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:

- Closes chromiecraft/chromiecraft#10170

## SOURCE:

The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

## Tests Performed:

This PR has been:
- [ ] Tested in-game by the author.
[...]

PR #9099, opened 2021-11-10
feat(Core/Disables): Implement DISABLE_TYPE_GAME_EVENT

Nyeriah:
## Changes Proposed:
- Allow disabling game events.
- Stop events if they're disabled while running during .reload disables.
- Disabled events can still be forced to start using .event start.

## Issues Addressed:

- Closes 

## Tests Performed:

- Tested in-game.
- 

## How to Test the Changes:

```sql
DELETE FROM disables WHERE sourcetype = 9 AND entry = 26;
INSERT INTO `disables` (`sourceType`, `entry`, `flags`, `params_0`, `params_1`, `comment`) VALUES
(9, 26, 0, 0, 0, 'Disable Pilgrim\'s Bounty');
```

1. https://www.nirsoft.net/utils/run_as_date.html worldserver - 25th Nov or so
2. Check if the event starts
3.

## Known Issues and TODO List:

- [ ]
- [ ]

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.

PR #27655, opened 2026-09-15
fix(Core/Events): match a holiday to its own event, not its building stage

sudlud:
## Changes Proposed:

Follow-up to #27646 and #27652.

`GameEventMgr::GetHolidayEventId()` returned whichever event carrying the holiday came first by id
and ignored `holidayStage`. Brewfest and the Darkmoon Faire each have building stage events sharing
their holiday id (70 and 91 for Brewfest, 23, 71 and 77 for the faire), so the lookup can hand back
a building stage event. Its only caller, `WorldSession::HandleCalendarGetCalendar()`, feeds that id
to the `DISABLE_TYPE_GAME_EVENT` check, so a `disables` row meant to hide a holiday from the
calendar silently stops working as soon as a building stage event sorts below the event proper.

Nothing changes with the shipped data, where the ids happen to order the right way (24 before 70
and 91, 3, 4 and 5 before 23, 71 and 77). Event 91 was added recently near the top of the id range
and could as easily have taken a free low id.

The lookup now matches the holiday's main stage, the rule #27646 introduced for `IsHolidayActive()`.

This PR proposes changes to:

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Claude Code with Claude Opus 5, for the change and this description

## Issues Addressed:

- Closes 

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Open the in-game calendar and confirm the holidays are listed as before.
2. `INSERT INTO disables (sourceType, entry, flags, params_0, params_1, comment) VALUES (9, 24, 0, '', '', 'test');` to hide Brewfest from the calendar.
3. `UPDATE game_event SET holiday = 372, holidayStage = 1 WHERE eventEntry = 22;` to fake a building stage event sorting below Brewfest's own event 24 (22 is AQ War Effort, which carries no holiday).
4. Restart the worldserver and open the calendar. Brewfest is absent. Before this change it is still listed, because the disable check looked up event 22 instead of 24.
5. Undo both rows.

- [ ]
- [ ]


sudlud:
# Self-review — fix/holiday-event-id-stage → master

**Outcome** no findings — clean round; three non-blocking notes recorded below

**Reviewed** `a08659a10e` (`fix/holiday-event-id-stage` vs `upstream/master`, merge-base diff — 1 file, +2 −1)

**By** Claude Code, Opus 5 (1M context) — self-review v0.8, 2026-09-15

**In-game testing** none performed by the author so far; the PR keeps its "requires further testing" box ticked

<details>

**Intent** `GameEventMgr::GetHolidayEventId()` should return the event that represents the holiday itself, rather than whichever event carrying that holiday id happens to have the lowest event id, so its caller's `disables` lookup targets the right event.

**Project rules** `.agents/docs/self-review-rules.md` found and applied (pinned review, shared-core reach check, upstream attribution, in-game testing asked of the author)

**Diff** `4f174ed7a7f41bf1b909893f6f95c664464b08d5`

**Metadata** `893a5e1c46b537cc88bc12c82709151080aa7684`

## Round 1 — `a08659a10e`, full changeset, no findings

Verified rather than trusted: the returned event id is identical for every shipped row, the four holidays with more than one event (372, 374, 375, 376) all put the event proper on the higher stage, and that matches the `Holidays.dbc` durations; a holiday absent from `game_event`, one whose events all sit on stage 0, and `holidayId` 0 against the default-constructed vector slots all behave exactly as before; the function has a single caller, `HandleCalendarGetCalendar()`, with no consumers in `modules/`, `src/test/` or `e2e/`; no attribution is owed, since TrinityCore has no equivalent function and the AC one originates in #9099. Codestyle passes.

Non-blocking notes:

1. PR test step 3 makes event 22 (AQ War Effort) a live holiday event, whose length and start `SetHolidayEventTime()` then rewrites from the DBC, so its content can spawn while the fixture is in place → **dismissed**: step 5 reverts it
2. `GameEventMgr.cpp:2035` and the same rule at `:2064` are data-shaped: the Call to Arms holidays define a stage 2 that is the off period, not a later main stage. Shipped `game_event` carries only their stage-1 rows, so nothing breaks; a DB change adding the off-period rows would make both treat the gap as the holiday. DB-side concern, not changed here
3. A custom row with a holiday, a stage above 0 and length 0, as the sole event for its holiday, hits the `continue` at `GameEventMgr.cpp:376` before main-stage registration, so the lookup returns 0 where it used to return that event id. Such an event is invalid and can never start

</details>

Gultask (approved):

PR #27199, opened 2026-08-17
fix(Scripts/ICC): restore Rocket Artillery in the Gunship Battle

EricksOliveira:
Skybreaker Mortar Soldiers and Kor'kron Rocketeers never fired Rocket Artillery during the Icecrown Gunship Battle. Two calls to WorldObject::GetMapHeight still passed a phase mask as the first argument.

That function takes (x, y, z, vmap, distanceToSearch) and applies the object's own phase internally, so the extra argument doesn't just duplicate work, it shifts every other one: x got the phase mask, y got posX, z got posY, and vmap got posZ narrowed to a bool. It compiles because those are all implicit conversions.

The height that came back was meaningless, so the < 465.0f guard in IgbArtilleryCheck rejected every candidate, SelectRandomTarget cleared the list, and the spell resolved with no targets. The spell the NPC casts is instant and has no visual, so in game the rocketeers just stood there. No cast bar, no missile, nothing in the log. This came in with the 2021 pathfinding rework, which rewrote GetMap()->GetHeight(phase, x, y, z) into GetMapHeight(...) without dropping the phase argument.

## Changes Proposed:

This PR proposes changes to:
-  [ ] Core (units, players, creatures, game systems).
-  [x] Scripts (bosses, spell scripts, creature scripts).
-  [ ] Database (SAI, creatures, etc).

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: Claude Code (Opus 5)
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/10670

## SOURCE:

The changes have been validated through:
- [x] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [ ] Sniffs (remember to share them with the open source community!)
- [x] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
[...]

PR #4220, opened 2021-01-06
feat(Core/Movement): Improved pathfinding, collisions and movements

Yehonal:
**Authors**: @Yehonal & @Footman 

Thanks to all the testers and reviewer @pklloveyou, @wizzymore , @Helias , @FrancescoBorzi , @nicoaravena, @Kitzunu , @lineagedr , @55Honey 

Thanks also to TC devs where we got some of code/ideas.

#### Npc positioning
- Implemented slope check to avoid unwanted climbing for some kind of movements (backwards, repositioning etc.)
- Implemented backwards movement
- Re-implemented circle repositioning algorithm (smartest than retail, but with the same feeling)
- Fixed random position of summoned minions
- Improved pet following movement. Also, they attack NPC from behind now. Thanks to @Footman 

#### Swimming creatures
- Fixed max_z coordinate for swimming creatures. Now only part of their body is allowed to be out of the water level
- Fixed pathfinder for swimming creatures creating shortcuts for specific segments, now they swim underwater to reach the seashore instead of flying above the water level.
- Creatures with water InhabitType but no swimming flag now, when not in combat, will walk on water depth instead of swimming. Thanks @jackpoz for the original code
- UNIT_FLAG_SWIMMING in UpdateEnvironmentIfNeeded to show the swimming animation correctly when underwater 
- Implemented HasEnoughWater check to avoid swimming creatures to go where the water level is too low but also to properly enable swimming animation only when a creature has enough water to swim.

#### Walking creatures
- Extended the DetourNavMeshQuery adding area cost based on walkability (slope angle + source height) to find better paths at runtime instead of completely remove them from mmaps
- improve Z height in certain conditions (see https://github.com/azerothcore/azerothcore-wotlk/issues/4205, https://github.com/azerothcore/azerothcore-wotlk/issues/4203, https://github.com/azerothcore/azerothcore-wotlk/issues/4247  )

#### Flying creatures
- Rewriting of the hover system
- Removed hacks and improved the UpdateEnvironmentIfNeeded. Now creatures can properly switch from flying to walk etc.

#### Spells
- LOS on spell effect must be calculated on CollisionHeight and HitSpherePoint instead of position coords. 
- Improved position for object/creature spawned via spells
- Improved checks for Fleeing movements (fear spells)

#### Other improvements
- Implemented method to calculate the CollisionWidth from dbc (used by repositioning algorithm etc.)
- Improved raycast and collision checks

## Issues Addressed:

[...]

PR #27661, opened 2026-09-15
fix(Scripts/ICC): spawn Keleseth Shadow Resonance adds away from him

### LINKED ISSUE #27654: [ICC] [Blood Prince Council] Shadow Resonance spawns on top of Keleseth
### Current Behaviour

Prince Keleseth's Shadow Resonance adds spawn on top of him instead of out in the room.

`spell_blood_council_summon_shadow_resonance::SetDest` (`boss_blood_prince_council.cpp:1676`) keeps the rolled destination only when it's more than 35 yards out and inside the room. Otherwise it walks outward from Keleseth looking for a spot with line of sight and a sane ground height, and falls back to his own position if it doesn't find one.

That search never finds anything, because of this line:

```cpp
float destZ = summoner->GetMapHeight(summoner->GetPhaseMask(), destX, destY, summoner->GetPositionZ());
```

`WorldObject::GetMapHeight` takes `(x, y, z, vmap, distanceToSearch)` and reads the object's own phase internally, it never took a phase mask. So every argument shifts by one: `x` gets the phase mask, `y` gets `destX`, `z` gets `destY`, and `vmap` gets `destZ` narrowed to a bool. It compiles because those are all implicit conversions.

The height that comes back is meaningless, so `fabs(destZ - summoner->GetPositionZ()) < 10.0f` never passes, every candidate gets rejected, and the fallback at the bottom of the loop relocates the destination onto Keleseth himself.

Same root cause as #27199, which fixed the two occurrences in the gunship battle. This is the last one left in the repo. It came in with the 2021 pathfinding rework (c8f43d858, #4220), which rewrote `GetMap()->GetHeight(phaseMask, x, y, z)` into `GetMapHeight(...)` without dropping the phase argument.

### Expected Behaviour

Shadow Resonance should spawn away from Keleseth, at the offset position the search loop is meant to pick.

### Source

Code inspection. `WorldObject::GetMapHeight` is declared at `src/server/game/Entities/Object/Object.h:716` and reads `GetPhaseMask()` internally at `Object.cpp:3145`.

1. Pull the Blood Prince Council.
2. Wait for Keleseth's Shadow Resonance, it goes off every 10 to 15 seconds for the whole fight.
3. Watch where the add appears.

### Extra Notes

Fix is dropping the first argument, same as #27199:

```cpp
float destZ = summoner->GetMapHeight(destX, destY, summoner->GetPositionZ());
```

Not tested in game, this came out of reading the code while reviewing #27199.

### AC rev. hash/commit

5c94774263

### Operating system

n/a, code issue

### Custom changes or Modules

None

EricksOliveira:
## Changes Proposed:

This PR proposes changes to:

## What and why

On the Blood Prince Council, the Shadow Resonance adds Keleseth summons every 10 to 15 seconds always appeared right on top of him instead of out in the room.

The script picks a spot by walking outward from the caster and checking the ground height there, but the height lookup was still passing a phase mask as its first argument. `WorldObject::GetMapHeight` takes `(x, y, z, vmap, distanceToSearch)` and reads the object's phase itself, so every argument slid one slot over: the phase mask landed in `x`, the candidate X in `y`, the candidate Y in `z`. It compiles because those are all implicit conversions. The height that came back was meaningless, the `fabs(destZ - casterZ) < 10.0f` check never passed, and the loop always fell through to the fallback that drops the add on the caster.

This has been broken since the 2021 pathfinding rework in https://github.com/azerothcore/azerothcore-wotlk/pull/4220, which converted the sibling call in this same file correctly and missed this one.

## Decisions worth knowing

Only the extra argument goes. The 35 yard early return, the room bounds, the line of sight check, the search order, the 10 yard Z tolerance and the fallback onto the caster all stay as they are, so an add landing 10 yards from Keleseth is still a valid result.

Same shape as https://github.com/azerothcore/azerothcore-wotlk/pull/27199, which fixed the same bug at the two gunship call sites. This was the last one left in the tree.

The caster's raw Z goes in untouched, since `GetMapHeight` adds the collision height on its own.

No SQL needed. Spell 71943 already has its script binding and no difficulty rows.

## Review guide

One line in `spell_blood_council_summon_shadow_resonance::SetDest`, in `src/server/scripts/Northrend/IcecrownCitadel/boss_blood_prince_council.cpp`. The judgment call isn't the line itself, it's whether the search can now return a spot that's worse than the old always on Keleseth behaviour, so it's worth reading the surrounding loop with the room bounds and the Z tolerance in mind.

## Known gaps

Tested on normal difficulty, where the adds now spawn away from him. 25 heroic not tested yet.

The east side of the room behind the throne is the one direction the bounds don't cap, so a candidate there rests on the line of sight check alone. Not tested with Keleseth pulled over that way.

Spell 71943's radius comes from DBC data, so how often the 35 yard early return fires instead of the search is unknown either way.

### AI-assisted Pull Requests

> [!IMPORTANT]
>

   - Tools/models used: Opus5

## Issues Addressed:

- Closes https://github.com/azerothcore/azerothcore-wotlk/issues/27654

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1.
2.
3.

- [ ]
- [ ]


EricksOliveira:
# Self-review — fixShadowResonanceSpawns → master

**Outcome** clean — 1 round, 0 findings

**Reviewed** `edfcaaf0695b0cdb29db4dd9edf4e317013761f4` (`fixShadowResonanceSpawns` vs `upstream/master`, merge-base diff — 1 file, +1 −1)

**By** Claude Code, Opus 5 — self-review v0.11, 2026-09-15

**Tested in-game** Blood Prince Council on normal difficulty: Keleseth's Shadow Resonance adds spawn away from him instead of on his position. Edge cases not covered yet: 25-man heroic, and Keleseth pulled to the east side of the room behind the throne, where the destination search leans on the line-of-sight check alone.

<details>

**Intent** On the Icecrown Citadel Blood Prince Council, the Shadow Resonance adds that Prince Keleseth summons should appear out in the encounter room away from him, instead of always landing on his exact position.

**Project rules** `.agents/docs/self-review-rules.md` found and applied

**Diff** `8baadb30375edf3dda2fe8f5a823a9deb0a9e6c7`

**Metadata** `fa69d4edb77eb891bcb160252a8e2cea757876df`

## Round 1 — `edfcaaf06`, 0 findings

Reviewer verified the call now matches `WorldObject::GetMapHeight(x, y, z, vmap, distanceToSearch)`, that the implementation raises Z and reads the object's own phase itself, and that the old shifted call fed `Map::GetHeight` junk coordinates returning `INVALID_HEIGHT`, so the Z guard never passed and every summon hit the relocate-onto-caster fallback.

No destination worse than the old behaviour was constructible: the room bounds plus the ±10 yard Z tolerance exclude the upper ring (z 384.8) and the west corridor (z 351.2), checked against the real spawns in `data/sql/base/db_world/creature.sql`.

The line-of-sight call on the line above does take a phase mask in its own signature, so it is not a second shifted call. The sibling `GetMapHeight` at `boss_blood_prince_council.cpp:1400` and the gunship precedent (`5c9477426`) were both already correct, so the change does not extend a broken pattern.

Sweep of `src/` and `modules/` found no remaining `GetMapHeight` call passing a phase mask. Spell 71943 is already bound in `spell_script_names` and has no `spelldifficulty_dbc` rows, so no SQL or doc counterpart is missing.

Governing docs checked as a checklist: `codestyle-cpp.py` clean, `.editorconfig` respected, commit scope and subject match repo convention. Upstream attribution does not apply — the defect came from AzerothCore's own `c8f43d858`, nothing is mirrored from another core. No leftovers.

</details>

Gultask (approved):

PR #27669, opened 2026-09-16
fix(Core/Calendar): count only the holidays actually sent

sudlud:
## Changes Proposed:

`WorldSession::HandleCalendarGetCalendar()` writes `ModifiedHolidays.size()` as the holiday count
and only afterwards skips any holiday whose event is disabled through `disables`. The count then
promises the client more holiday records than the packet carries, and every byte the client reads
past the last real record belongs to nothing.

The same function already handles this twice, for instance binds and for raid reset times: fill a
`ByteBuffer`, count while filling, then write the count and append the buffer. The holiday block now
does the same.

Only servers that put a `disables` row of sourceType 9 on a game event carrying a holiday are
affected. The code goes back to #9099.

Spotted in the review of #27655, which has since merged.

This PR proposes changes to:

### AI-assisted Pull Requests

   - Tools/models used: Claude Code with Claude Opus 5, for the change and this description

## Issues Addressed:

- None. Spotted in the review of #27655.

## SOURCE:

## Tests Performed:

This PR has been:

## How to Test the Changes:

1. Open the in-game calendar on an unmodified world DB and confirm the holidays are listed as before.
2. `INSERT INTO disables (sourceType, entry, flags, params_0, params_1, comment) VALUES (9, 24, 0, '', '', 'test');` to hide Brewfest.
3. Restart the worldserver and open the calendar. Brewfest is absent and the remaining holidays still list correctly. Before this change the packet claims one holiday more than it carries.
4. Remove the row again.

- [ ] The `/// @todo` above the holiday block still stands: nothing decides which holidays belong in the packet, only how many of them are written.


sudlud:
# Self-review — fix/calendar-holiday-count → master

**Outcome** no findings — clean round

**Reviewed** `aa645bd169` (`fix/calendar-holiday-count` vs `master`, merge-base diff — 1 file, +17 −10)

**By** Claude Code, Opus 5 (1M context) — self-review v0.8, 2026-09-16

**In-game testing** none performed by the author, and the PR keeps its "requires further testing" box ticked. CI builds pass on all eight compiler configurations, so the change compiles; e2e was still running at review time.

<details>

**Intent** `HandleCalendarGetCalendar()` wrote the holiday count before the loop that can skip individual holidays, so the count could exceed the number of records written. The packet's holiday count should equal the records it carries, without changing which holidays are sent or the bytes of each record.

**Project rules** `.agents/docs/self-review-rules.md` found and applied (pinned review, shared-core reach check, upstream attribution, in-game testing asked of the author)

**Diff** `c1cf5574921076f4a69b2c653a95b4df348e1ca4`

**Metadata** `9c8149a0615a8aad75c96122a60a9a7cc655e5dc`

## Round 1 — `aa645bd169`, full changeset, no findings

Verified rather than trusted: with nothing disabled the packet is byte-for-byte what master produces, since `WorldPacket` derives from `ByteBuffer` and overrides no `operator<<`, every field keeps its overload and order, and the block still lands after the raid-reset append. `ByteBuffer::clear()` resets `_storage`, `_rpos` and `_wpos`, matching the two existing reuses of the same buffer in this function, and `append()` guards on `wpos()`, so a server with every holiday disabled writes a count of 0 and no records; the counter is `uint32` into a `uint32` field.

`SMSG_CALENDAR_SEND_CALENDAR` is built only here, and the handler's one non-opcode caller, the broadcast at `InstanceSaveMgr.cpp:658`, is unaffected; no unit test or e2e reads the packet. Nothing has touched `CalendarHandler.cpp` on master since the merge base, so the change is still needed and applies cleanly. Codestyle passes and there are no leftovers.

Not a mirror of another core, so no `--author` credit is owed: the skipped-but-still-counted holiday is AC's own, introduced by #9099 (`855f098a57`), and the buffer-and-count pattern comes from this same function.

Description claims checked against the code rather than accepted: the two existing buffer-and-count blocks are there; `DISABLE_TYPE_GAME_EVENT` is 9 and that `continue` is the loop's only skip path; #9099 introduced the mismatch; `disables` entry 24 is Brewfest, which `GetHolidayEventId(372)` resolves to through the main-stage match.

</details>

Gultask (approved):

